PR #375, opened 2015-11-13
Fix setting ContentRange to inclusive positions

### LINKED ISSUE #373: DefaultStorageRpc.read setting wrong range
In [DefaultStorageRpc.read](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java#L418) we set the content range as:
```java
public byte[] read(StorageObject from, Map<Option, ?> options, long position, int bytes)
    throws StorageException {
  ...
  downloader.setContentRange(position, (int) position + bytes);
  ...
}
```

Accorting to [here](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35) positions are inclusive, so we should probably do: `downloader.setContentRange(position, (int) position + bytes - 1);`

mziccard:
This fixes #373

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
@@ -60,6 +60,7 @@ import com.google.api.services.storage.model.StorageObject;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import com.google.common.primitives.Ints;
 import com.google.gcloud.storage.StorageException;
 import com.google.gcloud.storage.StorageOptions;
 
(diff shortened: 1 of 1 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
```diff
@@ -415,7 +415,7 @@ public void onFailure(GoogleJsonError e, HttpHeaders responseHeaders) {
           .setIfGenerationMatch(IF_GENERATION_MATCH.getLong(options))
           .setIfGenerationNotMatch(IF_GENERATION_NOT_MATCH.getLong(options));
       MediaHttpDownloader downloader = req.getMediaHttpDownloader();
-      downloader.setContentRange(position, (int) position + bytes);
+      downloader.setContentRange(position, (int) position + bytes - 1);
```
aozarov: Shall we do:
[Ints.checkedCast](http://docs.guava-libraries.googlecode.com/git/javadoc/src-html/com/google/common/primitives/Ints.html#line.87)(position + bytes - 1) until the http-api-client bug is fixed?

mziccard: Sure, it should make it easier to identify errors. Fixed in place.

PR #376, opened 2015-11-13
Simplify package-info examples

### LINKED ISSUE #358: Standardize code snippets in package-info
Right now, the example in the Datastore's and Storage's package-info.java files explicitly sets the service options' project IDs but not the auth credentials.  The user will probably specify both or none, depending on whether they run the code from App/Compute Engine or not.

I think we should have two examples in each package info, one with both left unset and the other with both explicitly set.

ajkannan:
Fixes #358 

Instead of including two examples (one from App/Compute Engine, the other not), I added links to the "Specifying a Project ID" and "Authentication" sections of the README.

### DIFF
--- a/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/package-info.java
+++ b/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/package-info.java
@@ -50,8 +50,7 @@
  * href="https://github.com/GoogleCloudPlatform/gcloud-java#specifying-a-project-id">specify a
  * project ID</a> and
  * <a href="https://github.com/GoogleCloudPlatform/gcloud-java#authentication">provide
- * authentication</a>.
- *
+ * credentials</a>.
  * @see <a href="https://cloud.google.com/datastore/">Google Cloud Datastore</a>
  */
 package com.google.gcloud.datastore;
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/package-info.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/package-info.java
@@ -17,7 +17,7 @@
 /**
  * A client to Google Cloud Storage.
  *
- * <p>Here's a simple usage example for when running on App/Compute Engine:
+ * <p>Here's a simple usage example for using gcloud-java from App/Compute Engine:
  * <pre>{@code
  * Storage storage = StorageOptions.defaultInstance().service();
  * BlobId blobId = BlobId.of("bucket", "blob_name");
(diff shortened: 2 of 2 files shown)


gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/package-info.java
```diff
@@ -47,6 +46,12 @@
  * }
  * } </pre>
  *
+ * <p>When using gcloud-java from outside of App/Compute Engine, you have to <a
+ * href="https://github.com/GoogleCloudPlatform/gcloud-java#specifying-a-project-id">specify a
+ * project ID</a> and
+ * <a href="https://github.com/GoogleCloudPlatform/gcloud-java#authentication">provide
```
mziccard: Maybe "setup authentication"?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/package-info.java
```diff
@@ -35,6 +34,11 @@
  *   channel.close();
  * }}</pre>
  *
+ * When using gcloud-java from outside of App/Compute Engine, you have to <a
+ * href="https://github.com/GoogleCloudPlatform/gcloud-java#specifying-a-project-id">specify a
+ * project ID</a> and
+ * <a href="https://github.com/GoogleCloudPlatform/gcloud-java#authentication">provide
```
mziccard: Also here, "setup authentication"?

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/package-info.java
```diff
@@ -47,6 +46,12 @@
...
+ * authentication</a>.
+ *
```
mziccard: Nit, is this extra space needed?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/package-info.java
```diff
@@ -17,10 +17,9 @@
 /**
  * A client to Google Cloud Storage.
  *
- * <p>A simple usage example:
+ * <p>Here's a simple usage example for when running on App/Compute Engine:
```
mziccard: "for using gcloud-java from App/Compute Engine"

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/package-info.java
aozarov: or "provide credentials"?

ajkannan:
Suggestions incorporated

mziccard:
Looks good to me

PR #377, opened 2015-11-13
Update Releasing docs to add "wait for artifact deployment" step

ajkannan:


RELEASING.md
```diff
@@ -10,13 +10,15 @@ This script takes an optional argument denoting the new version.  By default, if
...
 
-3. Create a release on Github manually.
+3. Before moving on, verify that the artifacts have successfully been pushed to the Maven Central Repository.  To do this, we need to check the Travis CI logs since the artifacts take a couple hours to appear on the Maven Central Repository's website.  Open Travis CI, click the ["Build History" tab](https://travis-ci.org/GoogleCloudPlatform/gcloud-java/builds), and open the second build's logs for Step 2's PR.  Be sure that you are not opening the "Pull Request" build logs; you should not be under the "Pull Requests" tab.  When the build finishes, scroll to the end of the log and verify that the artifacts were successfully staged and deployed.  If the deployment didn't succeed because of a flaky test, the artifact may still have been pushed by the next Travis build, "Updating READMEs."  If both builds failed to deploy the artifacts, rerun the build.
```
aozarov: Can you check the logs and pom if we are actually pushing to [this](https://oss.sonatype.org/content/repositories/releases/com/google/gcloud/) maven repository? If so, that would be another way to check...

ajkannan: According to the pom and the logs, we push to a staging repository first, i.e. "oss.sonatype.org/service/local/staging/deploy/maven2".  Looking at the timestamps on the URL you posted, it looks like the artifacts appear there hours after the build.

aozarov: I see, I always was able to find the *just* pushed release by searching for 'gcloud-java' [here](https://oss.sonatype.org/#welcome).

PR #380, opened 2015-11-17
Update batch delete to return false on not found

### LINKED ISSUE #372: Storage.delete returns false on NotFound, BatchRequest.delete shows failure
If we try to delete a non-existing blob `storage.delete(nonExistingBlob)` returns `false`.
If we try to delete a non-existing blob with a batch request:
```java
BatchRequest batchRequest = BatchRequest.builder()
    .delete(nonExistingBlob)
    .build();
...
batchResponse.deletes().get(0).failed(); // This is true
```
The deletion of a non-existing-blob is considered a failure in `BatchResponse`.
Wouldn't it be better to make this two ways of deleting a blob exhibit the same behavior?

mziccard:
This fixes #372 and also adds better javadoc to delete methods.

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
@@ -30,6 +30,7 @@ import static com.google.gcloud.spi.StorageRpc.Option.PREDEFINED_ACL;
 import static com.google.gcloud.spi.StorageRpc.Option.PREDEFINED_DEFAULT_OBJECT_ACL;
 import static com.google.gcloud.spi.StorageRpc.Option.PREFIX;
 import static com.google.gcloud.spi.StorageRpc.Option.VERSIONS;
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 
 import com.google.api.client.googleapis.batch.json.JsonBatchCallback;
 import com.google.api.client.googleapis.json.GoogleJsonError;
@@ -94,7 +95,8 @@ public class DefaultStorageRpc implements StorageRpc {
 
   private static StorageException translate(IOException exception) {
     StorageException translated;
-    if (exception instanceof GoogleJsonResponseException) {
+    if (exception instanceof GoogleJsonResponseException
+        && ((GoogleJsonResponseException) exception).getDetails() != null) {
       translated = translate(((GoogleJsonResponseException) exception).getDetails());
     } else {
       translated = new StorageException(0, exception.getMessage(), false);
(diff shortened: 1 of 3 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
```diff
@@ -368,7 +368,11 @@ public void onSuccess(Void ignore, HttpHeaders responseHeaders) {
 
           @Override
           public void onFailure(GoogleJsonError e, HttpHeaders responseHeaders) {
-            deletes.put(tuple.x(), Tuple.<Boolean, StorageException>of(null, translate(e)));
+            if (e.getCode() == 404) {
```
aozarov: Sorry, my mess but I wonder if we can clean it in this PR or create another one for it:
We deal with 404 in both places `DefaultStorageRpc` for deletes and in `StorageImpl` for gets.
I think it is fine (and probably simpler) to move all of it to `DefaultStorageRpc` and it would be nice 
to mention the "return null" case in `StorageRpc`.

Also, we could use `java.net.HttpURLConnection.HTTP_NOT_FOUND` instead of the numerical value `404`.

mziccard: Sure I'll move all the `NOT_FOUND` checks to `DefaultStorageRpc`.

I am fine with documenting the "return null" case in `DefaultStorageRpc`, do you think we should do that in javadoc format or simple line/block comments to avoid adding javadoc to what we wanted to leave undocumented?

aozarov: > Sure I'll move all the NOT_FOUND checks to DefaultStorageRpc.

Thanks, sounds good.

> I am fine with documenting the "return null" case in DefaultStorageRpc, do you think we should do that in javadoc format or simple line/block comments to avoid adding javadoc to what we wanted to leave undocumented?

I don't have objection on having javadoc in the SPI layer, it was just not a priority.

mziccard:
Comments addressed.

gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
```diff
@@ -190,6 +192,12 @@ public Bucket get(Bucket bucket, Map<Option, ?> options) {
           .setIfMetagenerationNotMatch(IF_METAGENERATION_NOT_MATCH.getLong(options))
           .setFields(FIELDS.getString(options))
           .execute();
+    } catch (GoogleJsonResponseException ex) {
+      GoogleJsonError jsonError = ex.getDetails();
```
aozarov: Though slightly less "optimize" I prefer the style that we did for `delete`.

```java
  } catch (IOException ex) {
      StorageException serviceException = translate(ex);
      if (serviceException.code() == 404) {
        return false;
      }
      throw serviceException;
    }
```

Regardless of my preference, can you make them consistent and also replace the `delete` 404 with HTTP_NOT_FOUND?

aozarov:
One more nit.

mziccard:
I totally missed those more 404, fixed in place.

Issue #342, opened 2015-11-10
in the README the datastore sample does not compile.


ludoch:
For example, Datastore datastore = DatastoreOptions.getDefaultInstance().service();
does not seem to work with 0.0.10

ajkannan:
Yes, the README examples got updated too soon.  This example will work once we release version 0.0.11, which should be coming very soon (pending some testing in core).  The old way should still work, for example: 

`StorageOptions options = StorageOptions.builder().projectId("my_project_id").build();
 Storage storage = StorageFactory.instance().get(options);`

ajkannan:
Also the javadoc (available on the [website](http://googlecloudplatform.github.io/gcloud-java/0.0.10/apidocs/index.html) and also packaged with version 0.0.10) should have examples that are in sync with the release version.

ludoch:
Got it, thanks!

aozarov:
@ajkannan just released 0.0.11 so maven should be consistent with site again.

ludoch:
0.0.11 does not work with:
Datastore datastore = DatastoreOptions.getDefaultInstance().service();

mziccard:
@ludoch Sorry for this, the method `getDefaultInstance` is not part of the `DatastoreOptions` interface, I don't know how it happened to be in the README. To get a Datastore service instance with default options try:
```java
Datastore datastore = DatastoreOptions.builder().build().service();
```
I'll take care of fixing this, either by adding the `getDefaultInstance` method or by updating the READMEs

mziccard:
@ludoch In #357 we add back a `defaultInstance` method to `StorageOptions` and `DatastoreOptions` and update both READMEs and website's landing page (will be available on the next release).
Do you think we can close this?

aozarov:
0.0.12 which includes adding back the `defaultInstance` methods is out but I think we need to merge #389 before closing this issue.

mziccard:
#357 and #389 are merged so I am closing this. Feel free to reopen it if need be.

PR #389, opened 2015-11-18
fix readmes for storage

aozarov:


aozarov:
I think we can close #342 after merging this PR.

mziccard:
LGTM

PR #390, opened 2015-11-18
Make BlobReadChannel should fail if content is updated

### LINKED ISSUE #290: Fix int casting of lastBytePos in DefaultStorageRpc.read
This depends on https://github.com/google/google-api-java-client/issues/937. `MediaHttpDownloader.setContentRange` takes `firstBytePos` as a long while `lastBytePos` as an int. As soon as this gets fixed we need to remove `int` casting.

### LINKED ISSUE #359: BlobReadChannel should fail if content comes from different generations.
Feedback from @Capstan:

If someone updates the object when the client is in the middle of a read and hadn’t specified a precondition, you will simply read from the new object. You need to record the generation upon the first read and then choose what the behavior is for future reads… if the generation still exists (e.g., it’s now a history object), do you still allow reads or do you balk (because the original request is talking about the “current” object which has changed)?

mziccard:
This PR fixed #359. If a blob changes its `etag` while being read the `BlobReadChannel.read` method may throw `StorageException`. 
Main changes are:
- Fix `BlobWriteChannel` and `BlobReadChannel` javadoc
- Change `StorageRpc.read` to return a pair (etag, bytes)
- Keep track the last read etag in `BlobReadChannelImpl` to detect changes
- Throw `StorageException` if new etag does not match the last one
- Document this behavior in `Storage.reader` javadoc

This also fixes #290 as it no longer uses `MediaHttpDownloader.setContentRange`.

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
@@ -1408,7 +1408,11 @@ public interface Storage extends Service<StorageOptions> {
   /**
    * Return a channel for reading the blob's content. The blob's latest generation is read. If the
    * blob changes while reading (i.e. {@link BlobInfo#etag()} changes), subsequent calls to
-   * {@link BlobReadChannel#read(ByteBuffer)} may throw {@link StorageException}.
+   * {@code blobReadChannel.read(ByteBuffer)} may throw {@link StorageException}.
+   *
+   * <p>The {@link BlobSourceOption#generationMatch(long)} option can be provided to ensure that
+   * {@code blobReadChannel.read(ByteBuffer)} calls will throw {@link StorageException} if blob`s
+   * generation differs from the expected one.
    *
    * @throws StorageException upon failure
    */
@@ -1418,8 +1422,13 @@ public interface Storage extends Service<StorageOptions> {
    * Return a channel for reading the blob's content. If {@code blob.generation()} is set
    * data corresponding to that generation is read. If {@code blob.generation()} is {@code null}
    * the blob's latest generation is read. If the blob changes while reading (i.e.
-   * {@link BlobInfo#etag()} changes), subsequent calls to {@link BlobReadChannel#read(ByteBuffer)}
+   * {@link BlobInfo#etag()} changes), subsequent calls to {@code blobReadChannel.read(ByteBuffer)}
    * may throw {@link StorageException}.
+   *
+   * <p>The {@link BlobSourceOption#generationMatch()} and
+   * {@link BlobSourceOption#generationMatch(long)} options can be used to ensure that
+   * {@code blobReadChannel.read(ByteBuffer)} calls will throw {@link StorageException} if the
+   * blob`s generation differs from the expected one.
    *
    * @throws StorageException upon failure
    */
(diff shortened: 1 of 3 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1405,14 +1406,20 @@ private static void checkContentType(BlobInfo blobInfo) throws IllegalArgumentEx
   BatchResponse apply(BatchRequest batchRequest);
 
   /**
-   * Return a channel for reading the blob's content.
+   * Return a channel for reading the blob's content. The blob's latest generation is read. If the
```
aozarov: Also here BlobSourceOptions.generationMatch(long) can be used to pin request to specific generation.

gcloud-java-storage/src/test/java/com/google/gcloud/storage/BlobReadChannelImplTest.java
```diff
@@ -167,15 +167,41 @@ public void testReadClosed() {
...
+    replay(storageRpcMock);
+    reader.read(firstReadBuffer);
+    try {
+      reader.read(secondReadBuffer);
+      fail("Expected BlobReadChannel read to throw StorageException");
+    } catch (StorageException ex) {
+      StringBuilder messageBuilder = new StringBuilder();
+      messageBuilder.append("Blob ").append(blobId).append(" was updated while reading");
+      assertEquals(messageBuilder.toString(), ex.getMessage());
+      // expected
```
aozarov: Nit, I think we typically put the "// expected" comment in empty catch blocks.

gcloud-java-storage/src/test/java/com/google/gcloud/storage/ITStorageTest.java
```diff
@@ -731,6 +732,39 @@ public void testReadChannelFail() throws IOException {
...
+        byte[] newContent = new byte[blobSize];
+        random.nextBytes(newContent);
+        int numWrittenBytes = writer.write(ByteBuffer.wrap(newContent));
+        assertEquals(blobSize, numWrittenBytes);
+      }
+      readBytes = ByteBuffer.allocate(chunkSize);
+      reader.read(readBytes);
+      fail("StorageException was expected");
+    } catch(StorageException ex) {
+      // expected
```
aozarov: Maybe we should add the same error message verification such as the one you had for `BlobReadChannelImplTest.testReadGenerationChanged` just to make sure the failure is not from something else (e.g. the write operations)?

aozarov:
Looks great. Few nits.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
mziccard: Are you suggesting that we should set the storage object's generation with the value from `BlobSourceOptions.generationMatch(long)` option?

aozarov: No, I described it incorrectly.

Basically I was trying to suggest that if one uses `BlobSourceOptions.generationMatch(long)` or `BlobSourceOptions.generationMatch()` when `BlobId` has a generation value then in effect that
also guaranty to either read from that version or fail. Though if one does not specify the generation as part of the `BlobId` the generation that will be used is latest.

I am not sure how to phrase that level of detail without confusing. If you can think of something, great. If not, lets merge it as is.

mziccard:
Comments addressed, please @aozarov check whether the javadoc I added is clear enough for you.

aozarov:
excellent!

PR #401, opened 2015-11-20
Limit batch deletes to 100

### LINKED ISSUE #365: Limit google-api-client batch deletes to ~100.
Suggestion made by @Capstan 

We should change [`DefaultStorageRpc`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java) and potentially invoke multiple sequential batch requests to avoid sending more than 100 delete requests per google-api-client batch.

mziccard:
This fixes #365 by sending separate batch requests if deletes are more than 100.

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
@@ -35,7 +35,6 @@ import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import com.google.api.client.googleapis.batch.json.JsonBatchCallback;
 import com.google.api.client.googleapis.json.GoogleJsonError;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
-import com.google.api.client.googleapis.media.MediaHttpDownloader;
 import com.google.api.client.http.ByteArrayContent;
 import com.google.api.client.http.GenericUrl;
 import com.google.api.client.http.HttpHeaders;
@@ -59,9 +58,10 @@ import com.google.api.services.storage.model.ComposeRequest.SourceObjects.Object
 import com.google.api.services.storage.model.Objects;
 import com.google.api.services.storage.model.StorageObject;
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import com.google.common.primitives.Ints;
 import com.google.gcloud.storage.StorageException;
 import com.google.gcloud.storage.StorageOptions;
 
@@ -69,6 +69,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -82,6 +83,7 @@ public class DefaultStorageRpc implements StorageRpc {
   // see: https://cloud.google.com/storage/docs/concepts-techniques#practices
   private static final Set<Integer> RETRYABLE_CODES = ImmutableSet.of(504, 503, 502, 500, 429, 408);
   private static final long MEGABYTE = 1024L * 1024L;
+  private static final int MAX_BATCH_DELETES = 100;
 
   public DefaultStorageRpc(StorageOptions options) {
     HttpTransport transport = options.httpTransportFactory().create();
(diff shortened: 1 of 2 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1398,7 +1398,8 @@ private static void checkContentType(BlobInfo blobInfo) throws IllegalArgumentEx
   byte[] readAllBytes(BlobId blob, BlobSourceOption... options);
 
   /**
-   * Send a batch request.
+   * Send a batch request. If the {@code batchRequest} parameter contains more than 100 deletes the
```
aozarov: I don't think we should document that, nor force it on other `StorageRpc` providers.
I see this change as a necessary internal implementation to avoid an existing known issue (which is likely to be
fixed in the future). Do you think otherwise?

gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
```diff
@@ -361,6 +363,37 @@ public StorageObject compose(Iterable<StorageObject> sources, StorageObject targ
...
+    Map<StorageObject, Tuple<Boolean, StorageException>> deletes =
+        Maps.newHashMapWithExpectedSize(request.toDelete.size());
+    Map<StorageObject, Tuple<StorageObject, StorageException>> updates =
+        Maps.newHashMapWithExpectedSize(request.toUpdate.size());
+    Map<StorageObject, Tuple<StorageObject, StorageException>> gets =
+        Maps.newHashMapWithExpectedSize(request.toGet.size());
+    Iterator<List<Tuple<StorageObject, Map<Option, ?>>>> iterator = partitionedToDelete.iterator();
+    BatchRequest chunkRequest = new BatchRequest(iterator.hasNext() ? iterator.next() :
+        ImmutableList.<Tuple<StorageObject, Map<Option, ?>>>of(), request.toUpdate, request.toGet);
+    BatchResponse chunkResponse = batchChunk(chunkRequest);
```
aozarov: Fine as is. An alternative (without `mergeBatchResults`) could be:

```java
public BatchResponse batch(BatchRequest request) throws StorageException {
  List<List<Tuple<StorageObject, Map<Option, ?>>>> partitionedToDelete =
      Lists.partition(request.toDelete, MAX_BATCH_DELETES);
  Iterator<List<Tuple<StorageObject, Map<Option, ?>>>> iterator = partitionedToDelete.iterator();
  BatchRequest chunkRequest = new BatchRequest(iterator.hasNext() ? iterator.next() :
      ImmutableList.<Tuple<StorageObject, Map<Option, ?>>>of(), request.toUpdate, request.toGet);
  BatchResponse response = batchChunk(request);
  Map<StorageObject, Tuple<Boolean, StorageException>> deletes =
      Maps.newHashMapWithExpectedSize(request.toDelete.size());
  deletes.putAll(response.deletes);
  while (iterator.hasNext()) {
    BatchResponse deleteBatch = batchChunk(new BatchRequest(iterator.next(), null, null));
    deletes.putAll(deleteBatch.deletes);
  }
  return new BatchResponse(deletes, response.updates(), response.gets());
}
```

aozarov:
Looks good. One comment about the javadoc and the visibility of this chunking and another style suggestion.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
mziccard: My rationale here was to be transparent with the user on the amount of traffic we generate (as it might mean $ :). I agree that this is very tied to the implementation but nevertheless this is how the API behaves now, if it changes we can drop the comment.

BTW I am not attached to this and suspected you would have asked me to drop it :) I'll remove this!

gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
mziccard: This is much better.

mziccard:
Fixed in place

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
aozarov: Yes, please lets take it out.

I think that in general we should not commit in our API to the number of RPCs that we are actually going to perform per API call and in fact it may even changed per platform. (e.g. to support AE URL request [limits] (https://cloud.google.com/appengine/docs/java/urlfetch/#Java_Quotas_and_limits)).

I think one of the goals of our API it to perform a user request in an optimal way and applying
this chunking is doing exactly that.

PR #411, opened 2015-11-24
Avoid throwing exception when checking isLocalHost in Datastore

ajkannan:
DefaultDatastoreRpc uses sockets to determine whether a host is the localhost or not.  In appengine, the socket API isn't enabled by default and you have to enable billing to use sockets.  As a result, using an app without billing enabled (within the free Datastore usage bracket) will cause the following exception: 

`com.google.apphosting.api.ApiProxy$FeatureNotEnabledException: The Socket API will be enabled for this application once billing has been enabled in the admin console.`

This fix catches that exception.

### DIFF
--- a/gcloud-java-datastore/src/main/java/com/google/gcloud/spi/DefaultDatastoreRpc.java
+++ b/gcloud-java-datastore/src/main/java/com/google/gcloud/spi/DefaultDatastoreRpc.java
@@ -42,9 +42,7 @@ import org.json.JSONObject;
 import org.json.JSONTokener;
 
 import java.net.InetAddress;
-import java.net.MalformedURLException;
 import java.net.URL;
-import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


gcloud-java-datastore/src/main/java/com/google/gcloud/spi/DefaultDatastoreRpc.java
```diff
@@ -95,7 +93,7 @@ private static boolean isLocalHost(String host) {
         }
         InetAddress hostAddr = InetAddress.getByName(new URL(normalizedHost).getHost());
         return hostAddr.isAnyLocalAddress() || hostAddr.isLoopbackAddress();
-      } catch (UnknownHostException | MalformedURLException e) {
+      } catch (Exception e) {
```
aozarov: What exactly is using the sockets api? is it `InetAddress.getByName`?

ajkannan: Yes, that showed up in the stack trace.

PR #423, opened 2015-11-30
Add TableInfo and related model classes

mziccard:
This is a very big PR (but most of it is javadocs and tests) that adds `TableInfo` model class and all related classes:
- `TableSchema`
- `FieldSchema`
- `CsvOptions`
- `ExternalDataConfiguration`
- `UserDefinedFunction`

Unit tests are also added.

**Remarks**
I added the `TableSchema` class to wrap schema information as it also appears elsewhere (Query results). Schema is just a list of `FieldSchema` objects so if you prefer we can drop the class and just use `List<FieldSchema>`. I have no strong preferences.

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
@@ -19,8 +19,6 @@ package com.google.gcloud.bigquery;
 import com.google.api.services.bigquery.model.Table;
 import com.google.common.base.MoreObjects;
 
-import java.util.Objects;
-
 /**
  * A Google BigQuery Table information. A BigQuery table is a standard, two-dimensional table with
  * individual records organized in rows, and a data type assigned to each column (also called a
@@ -49,9 +47,12 @@ public class TableInfo extends BaseTableInfo {
       this.streamingBuffer = tableInfo.streamingBuffer;
     }
 
-    @Override
-    protected Builder self() {
-      return this;
+    protected Builder(Table tablePb) {
+      super(tablePb);
+      this.location = tablePb.getLocation();
+      if (tablePb.getStreamingBuffer() != null) {
+        this.streamingBuffer = StreamingBuffer.fromPb(tablePb.getStreamingBuffer());
+      }
     }
 
     Builder location(String location) {
@@ -119,7 +120,7 @@ public class TableInfo extends BaseTableInfo {
   }
 
   /**
-   * Returns a builder for the {@code ExternalTableInfo} object.
+   * Returns a builder for the {@code TableInfo} object.
    */
   @Override
   public Builder toBuilder() {
@@ -134,12 +135,6 @@ public class TableInfo extends BaseTableInfo {
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof TableInfo
-        && Objects.equals(toPb(), ((TableInfo) obj).toPb());
-  }
-
-  @Override
   Table toPb() {
     Table tablePb = super.toPb();
     tablePb.setLocation(location);
@@ -147,5 +142,10 @@ public class TableInfo extends BaseTableInfo {
       tablePb.setStreamingBuffer(streamingBuffer.toPb());
     }
     return tablePb;
+  }
+
+  @SuppressWarnings("unchecked")
+  static TableInfo fromPb(Table tablePb) {
+    return new Builder(tablePb).build();
   }
 }
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BaseTableInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BaseTableInfo.java
@@ -22,10 +22,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.api.client.util.Data;
 import com.google.api.services.bigquery.model.Streamingbuffer;
 import com.google.api.services.bigquery.model.Table;
-import com.google.api.services.bigquery.model.ViewDefinition;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
-import com.google.common.collect.Lists;
 
 import java.io.Serializable;
 import java.math.BigInteger;
@@ -38,7 +36,7 @@ import java.util.Objects;
  *
  * @see <a href="https://cloud.google.com/bigquery/docs/tables">Managing Tables</a>
  */
-public class BaseTableInfo implements Serializable {
+public abstract class BaseTableInfo implements Serializable {
 
   static final Function<Table, BaseTableInfo> FROM_PB_FUNCTION =
       new Function<Table, BaseTableInfo>() {
@@ -167,7 +165,7 @@ public class BaseTableInfo implements Serializable {
   private final Long expirationTime;
   private final Long lastModifiedTime;
 
-  public static class Builder<T extends BaseTableInfo, B extends Builder<T, B>> {
+  public static abstract class Builder<T extends BaseTableInfo, B extends Builder<T, B>> {
 
     private String etag;
     private String id;
@@ -199,6 +197,28 @@ public class BaseTableInfo implements Serializable {
       this.creationTime = tableInfo.creationTime;
       this.expirationTime = tableInfo.expirationTime;
       this.lastModifiedTime = tableInfo.lastModifiedTime;
+    }
+
+    protected Builder(Table tablePb) {
+      this.type = Type.valueOf(tablePb.getType());
+      this.tableId = TableId.fromPb(tablePb.getTableReference());
+      if (tablePb.getSchema() != null) {
+        this.schema(Schema.fromPb(tablePb.getSchema()));
+      }
+      if (tablePb.getLastModifiedTime() != null) {
+        this.lastModifiedTime(tablePb.getLastModifiedTime().longValue());
+      }
+      if (tablePb.getNumRows() != null) {
+        this.numRows(tablePb.getNumRows().longValue());
+      }
+      this.description = tablePb.getDescription();
+      this.expirationTime = tablePb.getExpirationTime();
+      this.friendlyName = tablePb.getFriendlyName();
+      this.creationTime = tablePb.getCreationTime();
+      this.etag = tablePb.getEtag();
+      this.id = tablePb.getId();
+      this.numBytes = tablePb.getNumBytes();
+      this.selfLink = tablePb.getSelfLink();
     }
 
     @SuppressWarnings("unchecked")
@@ -288,12 +308,9 @@ public class BaseTableInfo implements Serializable {
     }
 
     /**
-     * Creates a {@code BaseTableInfo} object.
+     * Creates an object.
      */
-    @SuppressWarnings("unchecked")
-    public T build() {
-      return (T) new BaseTableInfo(this);
-    }
+    public abstract T build();
   }
 
   protected BaseTableInfo(Builder builder) {
@@ -408,11 +425,9 @@ public class BaseTableInfo implements Serializable {
   }
 
   /**
-   * Returns a builder for the {@code BaseTableInfo} object.
+   * Returns a builder for the object.
    */
-  public Builder toBuilder() {
-    return new Builder(this);
-  }
+  public abstract Builder toBuilder();
 
   protected MoreObjects.ToStringHelper toStringHelper() {
     return MoreObjects.toStringHelper(this)
@@ -458,6 +473,7 @@ public class BaseTableInfo implements Serializable {
     if (schema != null) {
       tablePb.setSchema(schema.toPb());
     }
+    tablePb.setType(type.name());
     tablePb.setCreationTime(creationTime);
     tablePb.setDescription(description);
     tablePb.setEtag(etag);
@@ -471,71 +487,16 @@ public class BaseTableInfo implements Serializable {
 
   @SuppressWarnings("unchecked")
   static <T extends BaseTableInfo> T fromPb(Table tablePb) {
-    Builder builder;
-    TableId tableId = TableId.fromPb(tablePb.getTableReference());
-    Schema schema = tablePb.getSchema() != null ? Schema.fromPb(tablePb.getSchema()) : null;
-    if (Objects.equals(tablePb.getType(), Type.VIEW.name()) || tablePb.getView() != null) {
-      ViewDefinition viewPb = tablePb.getView();
-      ViewInfo.Builder viewBuilder = ViewInfo.builder(tableId, viewPb.getQuery());
-      if (tablePb.getView().getUserDefinedFunctionResources() != null) {
-        viewBuilder.userDefinedFunctions(Lists.transform(viewPb.getUserDefinedFunctionResources(),
-            UserDefinedFunction.FROM_PB_FUNCTION));
-      }
-      builder = viewBuilder;
-    } else if (Objects.equals(tablePb.getType(), Type.EXTERNAL.name())
-        || tablePb.getExternalDataConfiguration() != null) {
-      ExternalTableInfo.Builder externalBuilder = ExternalTableInfo.builder(tableId,
-          ExternalDataConfiguration.fromPb(tablePb.getExternalDataConfiguration()));
-      if (tablePb.getStreamingBuffer() != null) {
-        externalBuilder.streamingBuffer(StreamingBuffer.fromPb(tablePb.getStreamingBuffer()));
-      }
-      builder = externalBuilder;
-    } else if (Objects.equals(tablePb.getType(), Type.TABLE.name()) || schema != null) {
-      TableInfo.Builder tableBuilder = TableInfo.builder(tableId, schema);
-      if (tablePb.getLocation() != null) {
-        tableBuilder.location(tablePb.getLocation());
-      }
-      if (tablePb.getStreamingBuffer() != null) {
-        tableBuilder.streamingBuffer(StreamingBuffer.fromPb(tablePb.getStreamingBuffer()));
-      }
-      builder = tableBuilder;
-    } else {
-      // This is for incomplete tables returned by bigquery.listTables
-      builder = new Builder().tableId(tableId);
+    switch (Type.valueOf(tablePb.getType())) {
+      case TABLE:
+        return (T) TableInfo.fromPb(tablePb);
+      case VIEW:
+        return (T) ViewInfo.fromPb(tablePb);
+      case EXTERNAL:
+        return (T) ExternalTableInfo.fromPb(tablePb);
+      default:
+        // never reached
+        throw new IllegalArgumentException("Format " + tablePb.getType() + " is not supported");
     }
-    if (schema != null) {
-      builder.schema(schema);
-    }
-    if (tablePb.getDescription() != null) {
-      builder.description(tablePb.getDescription());
-    }
-    if (tablePb.getExpirationTime() != null) {
-      builder.expirationTime(tablePb.getExpirationTime());
-    }
-    if (tablePb.getFriendlyName() != null) {
-      builder.friendlyName(tablePb.getFriendlyName());
-    }
-    if (tablePb.getLastModifiedTime() != null) {
-      builder.lastModifiedTime(tablePb.getLastModifiedTime().longValue());
-    }
-    if (tablePb.getNumRows() != null) {
-      builder.numRows(tablePb.getNumRows().longValue());
-    }
-    if (tablePb.getCreationTime() != null) {
-      builder.creationTime(tablePb.getCreationTime());
-    }
-    if (tablePb.getEtag() != null) {
-      builder.etag(tablePb.getEtag());
-    }
-    if (tablePb.getId() != null) {
-      builder.id(tablePb.getId());
-    }
-    if (tablePb.getNumBytes() != null) {
-      builder.numBytes(tablePb.getNumBytes());
-    }
-    if (tablePb.getSelfLink() != null) {
-      builder.selfLink(tablePb.getSelfLink());
-    }
-    return (T) builder.build();
   }
 }
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalTableInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalTableInfo.java
@@ -21,8 +21,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.api.services.bigquery.model.Table;
 import com.google.common.base.MoreObjects;
 
-import java.util.Objects;
-
 /**
  * Google BigQuery External Table information. BigQuery's external tables are tables whose data
  * reside outside of BigQuery but can be queried as normal BigQuery tables. External tables are
@@ -51,9 +49,15 @@ public class ExternalTableInfo extends BaseTableInfo {
       this.streamingBuffer = tableInfo.streamingBuffer;
     }
 
-    @Override
-    protected Builder self() {
-      return this;
+    protected Builder(Table tablePb) {
+      super(tablePb);
+      if (tablePb.getExternalDataConfiguration() != null) {
+        this.configuration =
+            ExternalDataConfiguration.fromPb(tablePb.getExternalDataConfiguration());
+      }
+      if (tablePb.getStreamingBuffer() != null) {
+        this.streamingBuffer = StreamingBuffer.fromPb(tablePb.getStreamingBuffer());
+      }
     }
 
     /**
@@ -83,7 +87,7 @@ public class ExternalTableInfo extends BaseTableInfo {
 
   private ExternalTableInfo(Builder builder) {
     super(builder);
-    this.configuration = checkNotNull(builder.configuration);
+    this.configuration = builder.configuration;
     this.streamingBuffer = builder.streamingBuffer;
   }
 
@@ -122,12 +126,6 @@ public class ExternalTableInfo extends BaseTableInfo {
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof ExternalTableInfo
-        && Objects.equals(toPb(), ((ExternalTableInfo) obj).toPb());
-  }
-
-  @Override
   Table toPb() {
     Table tablePb = super.toPb();
     tablePb.setExternalDataConfiguration(configuration.toPb());
@@ -155,5 +153,10 @@ public class ExternalTableInfo extends BaseTableInfo {
    */
   public static ExternalTableInfo of(TableId table, ExternalDataConfiguration configuration) {
     return builder(table, configuration).build();
+  }
+
+  @SuppressWarnings("unchecked")
+  static ExternalTableInfo fromPb(Table tablePb) {
+    return new Builder(tablePb).build();
   }
 }
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ViewInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ViewInfo.java
@@ -25,7 +25,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
 import java.util.List;
-import java.util.Objects;
 
 /**
  * Google BigQuery View Table information. BigQuery's views are logical views, not materialized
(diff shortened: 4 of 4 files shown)


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/CsvOptions.java
```diff
@@ -0,0 +1,267 @@
...
+     * data. The default value is 0. This property is useful if you have header rows in the file
+     * that should be skipped.
+     */
+    public Builder skipLeadingRows(Integer skipLeadingRows) {
+      this.skipLeadingRows = skipLeadingRows;
+      return this;
+    }
+
+    /**
+     * Creates an {@code ExternalDataConfiguration} object.
```
aozarov: s/ExternalDataConfiguration/CsvOptions/

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/CsvOptions.java
```diff
@@ -0,0 +1,267 @@
...
+    private String quote;
+    private Integer skipLeadingRows;
+
+    private Builder() {}
+
+    /**
+     * Set whether BigQuery should accept rows that are missing trailing optional columns. If
+     * {@code true}, BigQuery treats missing trailing columns as null values. If {@code false},
+     * records with missing trailing columns are treated as bad records, and if there are too many
+     * bad records, an invalid error is returned in the job result. By default, rows with missing
```
aozarov: As for "too many bad records" it would be nice to have a link to `ExternalDataConfiguration.maxBadRecords`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/CsvOptions.java
```diff
@@ -0,0 +1,267 @@
...
+      this.allowQuotedNewLines = allowQuotedNewLines;
+      return this;
+    }
+
+    /**
+     * Sets the character encoding of the data. The supported values are UTF-8 or ISO-8859-1. The
+     * default value is UTF-8. BigQuery decodes the data after the raw, binary data has been split
+     * using the values set in {@link #quote(String)} and {@link #fieldDelimiter(String)}.
+     */
+    public Builder encoding(String encoding) {
```
aozarov: Fine as is. Alternative (or addition) could be to accept a `Charset` and then use its `name()`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
```diff
@@ -0,0 +1,419 @@
...
+ * Google BigQuery configuration for tables backed by external data. Objects of this class describe
+ * the data format, location, and other properties of a table stored outside of BigQuery.
+ * By defining these properties, the data source can then be queried as if it were a standard
+ * BigQuery table. Support for external tables is experimental and might be subject to changes or
+ * removed.
+ *
+ * @see <a href="https://cloud.google.com/bigquery/federated-data-sources">Federated Data Sources
+ *     </a>
+ */
+public class ExternalDataConfiguration implements Serializable {
```
aozarov: Would you have something similar for a data configuration of a load job?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/UserDefinedFunctionTest.java
```diff
@@ -0,0 +1,7 @@
+package com.google.gcloud.bigquery;
+
+/**
+ * Created by mziccardi on 11/30/15.
+ */
+public class UserDefinedFunctionTest {
+}
```
aozarov: implement or add TODO or remove.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
```diff
@@ -0,0 +1,419 @@
...
+    /**
+     * Sets the schema for the external data.
+     */
+    public Builder schema(TableSchema schema) {
+      this.schema = checkNotNull(schema);
+      return this;
+    }
+
+    /**
+     * Sets the source format of the external data.
```
aozarov: Maybe we could copy some of the text from the link?
BTW, https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load.sourceFormat seems to include more options...

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
```diff
@@ -0,0 +1,419 @@
...
+    public Builder compression(String compression) {
+      this.compression = compression;
+      return this;
+    }
+
+    /**
+     * Sets additional properties to be used to parse CSV data (used when
+     * {@link #sourceFormat(String)} is set to CSV).
+     */
+    public Builder csvOptions(CsvOptions csvOptions) {
```
aozarov: What is your plan to support future format specific options?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
```diff
@@ -0,0 +1,287 @@
...
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Google Bigquery schema for a Table field. Several data types and modes are supported.
```
aozarov: I find this javadoc a bit confusing. what does it mean "Table field". 
What is the purpose of the "Several data types and modes are supported" description and if that is only a subset where can I get the complete list.

aozarov: Ok, I don't find the first part ("Table field") as confusing now, but few suggestions:
(1) we should be consistent with name style Bigquery -> BigQuery
(2) Maybe we should say something like "Google BigQuery field schema" (schema or description) ", which is part of a TableSchema".
(3) did you consider these names TableSchema -> Table, FieldSchema -> Field  or do you think that may collide with the values options? Another possibility is Schema (for TableSchema) and Field or FieldSchema.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
```diff
@@ -0,0 +1,287 @@
...
+    public Builder type(Type type) {
+      this.type = checkNotNull(type);
+      return this;
+    }
+
+    /**
+     * Sets the mode of the field. By default {@link Mode#NULLABLE} is used.
+     */
+    public Builder mode(Mode mode) {
+      this.mode = mode;
```
aozarov: Should we consider null values for patch operation or is that not supported?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
```diff
@@ -0,0 +1,287 @@
...
+    Builder fields(Iterable<FieldSchema> fields) {
+      this.fields = fields != null ? ImmutableList.copyOf(fields) : null;
+      return this;
+    }
+
+    /**
+     * Creates an {@code FieldSchema} object.
+     */
+    public FieldSchema build() {
+      FieldSchema fieldSchema = new FieldSchema(this);
```
aozarov: Alternative: 

```java
checkNotNull(fieldSchema.name);
checkNotNull(fieldSchema.type);
return new FieldSchema(this);
```

Also, should you check that fields is only populated when type is RECORD?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
```diff
@@ -0,0 +1,287 @@
...
+
+  private static final long serialVersionUID = -8154262932305199256L;
+
+  /**
+   * Data Types for a BigQuery Table field.
+   *
+   * @see <a href="https://cloud.google.com/bigquery/preparing-data-for-bigquery#datatypes">
+   *     Data Types</a>
+   */
+  public enum Type {
```
aozarov: Another option that works nicer with RECORDS (and fields) is to make that a class with private constructor.
The class could have an an inner enum `Value` with the same set  of values.
Class has a factory method for each type `string()`, `integer()`, ... record(FieldSchema...)`.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
```diff
@@ -0,0 +1,287 @@
...
+    public Builder description(String description) {
+      this.description = description;
+      return this;
+    }
+
+    /**
+     * Sets the nested schema fields if {@link #type(Type)} is set to {@link Type#RECORD}.
+     */
+    Builder fields(Iterable<FieldSchema> fields) {
+      this.fields = fields != null ? ImmutableList.copyOf(fields) : null;
```
aozarov: Is it valid to have a record with no fields?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableSchema.java
```diff
@@ -0,0 +1,151 @@
...
+   */
+  public List<FieldSchema> fields() {
+    return fields;
+  }
+
+  /**
+   * Returns a builder for the {@code TableSchema} object.
+   */
+  public Builder toBuilder() {
+    Builder builder = new Builder();
```
aozarov: Alternative:
```java
return builder().fields(fields);
```

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/UserDefinedFunction.java
```diff
@@ -0,0 +1,158 @@
...
+        && Objects.equals(toPb(), ((UserDefinedFunction) obj).toPb());
+  }
+
+  public abstract com.google.api.services.bigquery.model.UserDefinedFunctionResource toPb();
+
+  /**
+   * Creates a Google Cloud BigQuery user-defined function given a code blob.
+   */
+  public static UserDefinedFunction inline(String functionDefinition) {
+    return new InlineFunction(functionDefinition);
```
aozarov: should we make the specific constructors package scope?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/UserDefinedFunctionTest.java
mziccard: Ooops, must have forgot to re-add after implementing. Weird. Will fix.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
mziccard: The approach I adopted so far for *pick-one* fields is to have all fields in the top level class and design the builder and factory methods so that only one is initialized at a time. It would be something like:
```java
public static ExternalDataConfiguration of(String sourceUri, TableSchema schema, CsvOptions options) {
  return builder(sourceUri, schema, options).format("CSV").build();
}

public static ExternalDataConfiguration of(String sourceUri, TableSchema schema, JsonOptions options) {
  return builder(sourceUri, schema, options).format("NEWLINE_DELEMITED_JSON").build();
}
```

This should enforce a "good" construction.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
mziccard: The only [supported changes](https://cloud.google.com/bigquery/docs/tables#updateschema) to the schema of an existing table are:
- Adding NULLABLE or REPEATED fields at the end of the schema
- Making REQUIRED fields NULLABLE

Even though "unsetting" a REQUIRED mode should set it to the default (NULLABLE) I think requiring explicit values here is better.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
mziccard: Yes, a load job has its own configuration (which is slightly different). This class is also used to define [external data sources to a query job](https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.query.tableDefinitions), which are query-scoped federated tables (described [here](https://cloud.google.com/bigquery/federated-data-sources#cli))

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
mziccard: Yes I noticed that, and also an extract job supports [different destination formats](https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.extract.destinationFormat).

mziccard:
Fixed the commends, main changes are:
- `TableSchema` renamed to `Schema` and `FieldSchema` renamed to `Field`
- Changed `Type` enum to class in `Field` with factory methods for all types
- Add setter and builder method that take fields as a vararg for `Schema`
- Fixed all javadocs comments and updated tests according to changes

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
aozarov: Yes, that helps but one could also have access and use the `Builder` directly.

Did you consider a base class for `FormatOptions` with `type()` returns enum or string of possible types? then one need to set the formatOptions (probably instead of supplying `sourceFormat`) and you can use the enum in the toPb...?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
aozarov: Honestly, I think if we are going to use patch for updating a [`Table`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/bigquery/gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/BigQueryRpc.java#L153) we should probably
not relax the implementation based on current restriction as that might get relaxed in the future
and break the code.

If you agree, we don't have to do it now but rather create an issue for it.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Field.java
```diff
@@ -0,0 +1,370 @@
...
+    /**
+     * Sets the field description. The maximum length is 16K characters.
+     */
+    public Builder description(String description) {
+      this.description = description;
+      return this;
+    }
+
+    /**
+     * Creates an {@code Field} object.
```
aozarov: s/ an / a / ?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Field.java
```diff
@@ -0,0 +1,370 @@
...
+     * @see <a href="https://cloud.google.com/bigquery/preparing-data-for-bigquery#datatypes">
+     *     Data Types</a>
+     */
+    public Builder type(Type type) {
+      this.type = checkNotNull(type);
+      return this;
+    }
+
+    /**
+     * Sets the mode of the field. By default {@link Mode#NULLABLE} is used.
```
aozarov: Fine as is (and I don't remember what we typically say, so preference is for consistency) but another way to say it would be "Sets the mode of the field. When not specified {@lilnk Mode#Nullable} will be used."

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Field.java
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  /**
+   * Returns the field description.
+   */
+  public String description() {
+    return description;
+  }
+
+  /**
```
aozarov: Though could be considered as a "convenient" method my preference would be not to have it.
Feel free to leave as is if you think otherwise.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Schema.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+    return tableSchemaPb;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static Schema of(Iterable<Field> fields) {
+    checkNotNull(fields);
```
aozarov: will be checked in `Builder.fields` setter.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/UserDefinedFunction.java
aozarov: Also, in its current state I am not sure there is even a value in exposing the specific sub-classes 'UriFunction` and `UserDefinedFunction`.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/UserDefinedFunction.java
```diff
@@ -0,0 +1,151 @@
...
+  public Type type() {
+    return type;
+  }
+
+  /**
+   * Returns function's definition. If {@link #type()} is {@link Type#INLINE} this method returns
+   * a code blob. If {@link #type()} is {@link Type#FROM_URI} this method returns a Google Cloud
+   * Storage URI (e.g. gs://bucket/path).
+   */
+  public String functionDefinition() {
```
aozarov: How about `content` considering the scope `UserDefinedFunction` and the fact that it may actually not be a function definition but rather a reference.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/UserDefinedFunction.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  public Type type() {
+    return type;
+  }
+
+  /**
+   * Returns function's definition. If {@link #type()} is {@link Type#INLINE} this method returns
+   * a code blob. If {@link #type()} is {@link Type#FROM_URI} this method returns a Google Cloud
+   * Storage URI (e.g. gs://bucket/path).
```
aozarov: Is externalDataConfiguration.sourceUris also using the same format "gs://bucket/path" ?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
```diff
@@ -0,0 +1,741 @@
...
+          return tableInfo.toPb();
+        }
+      };
+
+  private static final long serialVersionUID = -7679032506430816205L;
+
+  /**
+   * The table type.
+   */
+  public enum Type {
```
aozarov: Did you consider using sub-classes?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
```diff
@@ -0,0 +1,741 @@
...
+     * @see <a href="https://cloud.google.com/bigquery/federated-data-sources">Federated Data
+     *     Sources</a>
+     */
+    public Builder externalConfiguration(ExternalDataConfiguration externalConfiguration) {
+      this.externalConfiguration = externalConfiguration;
+      return this;
+    }
+
+    /**
+     * Sets a user-friendly name for the dataset.
```
aozarov: s/dataset/table/

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
```diff
@@ -0,0 +1,741 @@
...
+      return this;
+    }
+
+    Builder numRows(Long numRows) {
+      this.numRows = numRows;
+      return this;
+    }
+
+    /**
+     * Sets the table's schema. Providing a schema is not necessary when {@link #viewQuery} is
```
aozarov: I assume that this is also not necessary for EXTERNAL (when ExternalDataConfiguration is provided)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
```diff
@@ -0,0 +1,741 @@
...
+   */
+  public TableId tableId() {
+    return tableId;
+  }
+
+  /**
+   * Returns a user-friendly name for the table.
+   */
+  public String friendlyName() {
+    return friendlyName;
```
aozarov: ``` return Data.isNull(friendllyName) ? null : friendlyName;``` here and in other places where
we use `Data.nullOf`

aozarov:
Looks good. Few more comments and suggestions.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
mziccard: I thought about that but in general I don't like requiring the user to do casting to access some data:
```java
FormatOptions options = externalDataConfiguration.formatOptions();
// to access real options:
CsvOptions csvOptions = (CsvOptions) options;
```
However, if you think having `FormatOptions` is better I am happy to change.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
mziccard: Do you think there's a need for this?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
mziccard: Why do you think updates will break the code? Here we are simply not allowing a user to unset a field's mode, which does not make sense at the moment. Am I missing something?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExternalDataConfiguration.java
mziccard: In the end I gave this a try in 79001cf046ca1f33427231655dd457cd41015e09, what do you think?

aozarov: I like it! I added few further suggestions.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
aozarov: I feel that using a standard Java inheritance would be a nicer user experience than having a flat type with all possible setter/getter (merged from all types TABLE, EXTERNAL and VIEW).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableInfo.java
aozarov: I think this is still not fixed.

aozarov:
See my comment regarding `CsvOptions` (I like the change you made and suggested to take it slightly further), the possible bug in `TableInfo` when nulls are passed in the builder and later retrieved by the built instance and also my suggestion to use inheritance for TableInfo as well (if not inheritence maybe we can consider something similar to what you did with `FormatOptions`).

Also, I feel like an open comment regarding `patch` and `ExternalDataConfiguration` got lost somehow.

mziccard:
I see where you are going with the hierarchies. The fact is that i really don't like that the user has to cast values to get to real data. It might be me but I really consider casting an unnecessary pain in the neck. The main point is that if we have:
```java
public <F extends FormatOptions> F formatOptions() {
  return (F) formatOptions;
}
```
To use this method a user must not only read its javadoc but he needs also to be aware of the `FormatOptions` hierarchy. If he does't know the expected type, in fact, he has to do type-checking to figure out how to use the return type. Avoiding type-checking and casting is the reason why I provided both `format` and `csvOptions` getters.

The same reasoning (even worse maybe) applies to `TableInfo` type. We might want to have something like:
```java
public abstract class TableType {

  static class Table extends TableType {
    // ...
  }
  static class View {
    // ...
  }
  static class External {
    // ...
  }
  public static Table of(Schema schema);
  public static View of (String query, List<UserDefinedFunction> functions);
  public static External of(ExternalDataConfiguration configuration);
}

public class TableInfo {
  // ...
  public static Builder builder(TableType type);
  public static ViewInfo of(TableType type);
}
```
This is ok but if you ask me I *would not* remove all `schema()`, `query()`, `userDefinedFunctions()` and `externalConfiguration()` getters from `TableInfo` if favor of `public <T extends ViewType>type()` for the same reasoning I did above for `CsvOptions`. Even worse, here the information wrapped in types is not disjoint, the table's schema in fact is returned by the BigQuery service for all 3 types of tables.

To sum things up: I am in favor of using hierarchies to make it easier for the user to build an object right, I am not in favor of having getters that require type-checking. I am open for discussion :)

> Also, I feel like an open comment regarding patch and ExternalDataConfiguration got lost somehow.

It got lost in fact, my answer is [here](https://github.com/GoogleCloudPlatform/gcloud-java/pull/423#discussion_r46421030).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldSchema.java
aozarov: My comment was more generic rather than specific. I agree that it may not be that useful
to support reseting a mode to default (NULLABLE) but if Builder allows setting `null` why not
respect it in patch operations? Same applies to description...

aozarov:
I think javadoc, which shoes class hierarchies, and making the set of format type constants more public (maybe not enum, but there are other ways can do it) could help with knowing which one should be used.

Also, I would expect the user to know it as (similar to the schema) it probably set it.
I only mentioned that method as a "convenient" getter that avoid explicit casting (which
we also do in other places).

>Avoiding type-checking and casting is the reason why I provided both format and csvOptions getters.

If someone was calling `csvOptions` it is because it already knew it was a `csvOptions`. Yes,
one could call all xxxOptions and check for not `null` but this does not look better to me than
`instanceof`.

If we assume that more and more formats are going to be supported (which, I think, is the reason we don't use enum for `type`), than in that case it would mean adding more and more `xxxOptions` get methods on `ExternalDataConfiguration`. Is that desired?

As for `TableInfo` I think here we should not expect proliferation of types (if any), so
having the, currently 3, specific getters is not a problem (but I don't see that as an improvement) however flattening *all* options (such as `userDefinedFunctions`) in `TableInfo` does not seem
right to me (and force the user to know what belongs to what).  In my mind it is prefered to have common stuff in common class and specific stuff in its specific place. BTW, did you consider to make `TableInfo` abstract and have 3 concrete implementation for it?

mziccard:
I obey :) Kidding, I don't completely agree but could discuss this for days and there's no point in stalling this any longer.
I hope I caught all comments, let me recap:
- f84a3b1: Fixes the [comment](https://github.com/GoogleCloudPlatform/gcloud-java/pull/423#discussion_r46231753) on unsetting `Field` mode and description
- bd58ee5 Removes [redundant getters](https://github.com/GoogleCloudPlatform/gcloud-java/pull/423#commitcomment-14750946) from ExternalDataConfiguration. Makes `FormatOptions.csv` return a `CsvOptions` object (comment [here](https://github.com/GoogleCloudPlatform/gcloud-java/pull/423#commitcomment-14750781))
- 66972df: Adds the `TableType` class with 2 subclasses `View` and `ExternalTable`. This class is used in `TableInfo` [to wrap table's fields that depends on the type](https://github.com/GoogleCloudPlatform/gcloud-java/pull/423#discussion_r46366010).

aozarov:
I like this one better! (but still have few suggestions).

Next time when you do not completely agree I suggest you create an issue (marked as question) and we can continue to discuss it there (and solicit more opinions) while keeping the code as is.

mziccard:
Some more changes:
- ae2a484: Removes TableType and creates a hierarchy of `TableInfo`. If you are wondering why `BaseTableInfo` is not abstract it's because `listTables` return partial information on tables that does not allow to say whether they are normal tables, views or external tables. So I thought to use `BaseTableInfo` for them
- 338ede1: Fixes the `mode` issue and another minor comment

aozarov:
I really like it!! Few nits.

PR #424, opened 2015-11-30
Run tests for PRs to branches other than master

### LINKED ISSUE #420: Run unit tests for non-master branch PRs
Right now, we only run unit tests for PRs that are on the master branch.  We should run the unit tests.  We can optionally include a coverage report, but since PRs in other branches tend to be "work in progress" it's not as vital.

mziccard:
This should fix #420.
Notice that #420 will be fixed only when this is merged/rebased in all other branches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ before_install:
 install: mvn install -DskipTests=true -Dgpg.skip=true
 script: 
 - utilities/verify.sh
-branches:
-  only:
-  - master
 after_success:
 - utilities/after_success.sh
 env:
--- a/utilities/verify.sh
+++ b/utilities/verify.sh
@@ -4,7 +4,7 @@ source ./utilities/integration_test_env.sh
 # This script is used by Travis-CI to run tests.
 # This script is referenced in .travis.yml.
 
-if [ "${TRAVIS_BRANCH}" == "master" -a "${TRAVIS_PULL_REQUEST}" == "false" ]; then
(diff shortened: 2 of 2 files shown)


jgeewax:
Just out of curiosity, if someone submits a PR (that none of us review) that echo's our encrypted keys in plaintext, what's the point of encrypting the keys ?

Should we have to approve it before the tests are run by OUR Travis instance?

ajkannan:
@jgeewax Right now, we only decrypt the keys in non-PR Travis builds (see [here](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/utilities/verify.sh#L9)), so I don't think that particular scenario can happen right now.  If we happen to change the CI build code, we should be sure to decrypt the keys only when merging in a PR. 

@mziccard Do you think we should remove the check for whether the branch is the `master` branch [here](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/utilities/verify.sh#L7).

jgeewax:
Cool - so long as we're not opening up a gaping security hole :)

mziccard:
@jgeewax Everything that we encrypt/decrypt needs travis encryption variables that are only available for non-PR builds (AFAIK). So decrypting the keys in a PR should not be possible.

@ajkannan I thought to leave integration tests only for master but you are right we should run them (if any) also in other branches. I'll fix this!

aozarov:
Yes @mziccard is right. see [this](https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests).

PR #427, opened 2015-11-30
Default spi layer implementation

ajkannan:


gcloud-java-resourcemanager/src/main/java/com/google/gcloud/spi/DefaultResourceManagerRpc.java
```diff
@@ -52,37 +56,68 @@ private static ResourceManagerException translate(GoogleJsonError exception) {
...
 
   @Override
   public Tuple<String, Iterable<Project>> list(Map<Option, ?> options)
       throws ResourceManagerException {
-    // TODO(ajaykannan): fix me!
-    return null;
+    try {
+      ListProjectsResponse response = resourceManager.projects()
+          .list()
+          .setFilter(FIELDS.getString(options))
```
aozarov: `setFields`

PR #430, opened 2015-12-01
Remove http initializer from AuthCredentials

### LINKED ISSUE #403: Remove the HttpIntializer dependency from AuthCredentials
After we resolve #334 We should also change `AuthCredentials` and replace `protected abstract HttpRequestInitializer httpRequestInitializer` with `public abstract Credentials credentials()` returning `com.google.auth.Credentials.Credentials`.

We should also change `ServiceOptions.httpRequestInitializer` to get the credentials from `AuthCredentials` and wrap them inside `HttpCredentialsAdapter`.

If we can't find a way to deal with the deferred scoping, we can consider returning `GoogleCredentials`
instead of `credentials`.

ajkannan:
Fixes #403.  This PR also removes the dependency on com.google.api.client.googleapis.auth.oauth2.GoogleCredential

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
@@ -177,7 +177,7 @@ public abstract class AuthCredentials implements Restorable<AuthCredentials> {
     }
 
     @Override
-    public GoogleCredentials credentials() {
+    public ServiceAccountCredentials credentials() {
       return new ServiceAccountCredentials(null, account, privateKey, null, null);
     }
 
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java
@@ -562,8 +562,7 @@ final class StorageImpl extends BaseService<StorageOptions> implements Storage {
     }
     ServiceAccountAuthCredentials authCred =
         (ServiceAccountAuthCredentials) optionMap.get(SignUrlOption.Option.SERVICE_ACCOUNT_CRED);
-    ServiceAccountCredentials cred =
-        (ServiceAccountCredentials) (authCred != null ? authCred.credentials() : null);
+    ServiceAccountCredentials cred = authCred != null ? authCred.credentials() : null;
     if (authCred == null) {
       checkArgument(
           this.options().authCredentials() != null


gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
```diff
@@ -271,9 +315,17 @@ public static ServiceAccountAuthCredentials createFor(String account, PrivateKey
    */
   public static ServiceAccountAuthCredentials createForJson(InputStream jsonCredentialStream)
       throws IOException {
-    GoogleCredential tempCredentials = GoogleCredential.fromStream(jsonCredentialStream);
-    return new ServiceAccountAuthCredentials(tempCredentials.getServiceAccountId(),
-        tempCredentials.getServiceAccountPrivateKey());
+    GoogleCredentials tempCredentials = GoogleCredentials.fromStream(jsonCredentialStream);
```
aozarov: It would be nicer to change this method to return AuthCredentials and to also support `UserCredentials(String clientId, String clientSecret, String refreshToken) `.

gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
```diff
@@ -42,8 +38,66 @@
   private static class AppEngineAuthCredentials extends AuthCredentials {
 
     private static final AuthCredentials INSTANCE = new AppEngineAuthCredentials();
```
aozarov: We can't have the INSTANCE anymore as we also maintain a state (`scopesRequired` and `scopes`).
What you need to do is to change `capture()` to return an instance of `AppEngineAuthCredentialsState`
that store both variables and use them to create a new `AppEngineAuthCredentials` instance in `restore()`.

gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
```diff
@@ -42,8 +38,66 @@
...
+    private static class AppEngineCredentials extends GoogleCredentials {
+
+      private final Object appIdentityService;
+      private final Collection<String> scopes;
+      private final boolean scopesRequired;
+
+      AppEngineCredentials() {
+        try {
+          Class<?> factoryClass =
+              Class.forName("com.google.appengine.api.appidentity.AppIdentityServiceFactory");
```
aozarov: I want you to keep a reference here for `appIdentityService` and `getAccessTokenResult` and `getAccessToken`.

ajkannan:
@aozarov comments incorporated

Added a don't merge label because this still has to be tested on all the platforms.

aozarov:
LGTM, feel free to merge (and remove label) after you verified.

ajkannan:
Test app runs on all platforms without problems, will merge in now.

PR #438, opened 2015-12-03
Refactor service exceptions

### LINKED ISSUE #413: Move some exception translation logic to BaseServiceException
In `DefaultStorageRpc` (and also other services as `DefaultBigQueryRpc`) we have code handling the translation of `IOException` to service specific exceptions:
```java
private static BigQueryException translate(IOException exception) {
  BigQueryException translated;
  if (exception instanceof GoogleJsonResponseException
      && ((GoogleJsonResponseException) exception).getDetails() != null) {
    translated = translate(((GoogleJsonResponseException) exception).getDetails());
  } else {
    translated =
        new BigQueryException(BigQueryException.UNKNOWN_CODE, exception.getMessage(), false);
  }
  translated.initCause(exception);
  return translated;
}

private static BigQueryException translate(GoogleJsonError exception) {
  boolean retryable = RETRYABLE_CODES.contains(exception.getCode());
  return new BigQueryException(exception.getCode(), exception.getMessage(), retryable);
}
```
Part of this logic should be moved to `BaseServiceException`

mziccard:
This does a bit of refactoring of our service exceptions and fixes #413. A lot of files are changed. This has very low priority :) I really don't expect this PR to be merged as is but might serve as a good starting point for discussion.

Main changes are:
- Add `idempotent` boolean to `BaseServiceException`
- Move translation logic from `StorageException` to `BaseServiceException`
- Remove `DatastoreRpcException` class
- Remove `DatastoreError` class
- Remove structures for mapping datastore's reason to status code
- Throw `IllegalState/IllegalArgument/UnsupportedOperation` instead of `DatastoreException` is serveral cases where the error is on the user's side
- Update code and tests

@ajkannan Could you have a look at what I did to `StorageException` and `BaseServiceException` in this PR and tell me if your `ResourceManagerException` could be refactored similarly? `BigQueryException` should benefit from `BaseServiceException`'s new methods.

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
@@ -16,26 +16,80 @@
 
 package com.google.gcloud;
 
+import com.google.api.client.googleapis.json.GoogleJsonError;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+
+import java.io.IOException;
+import java.net.SocketTimeoutException;
+import java.util.Collections;
+import java.util.Set;
+
 /**
  * Base class for all service exceptions.
  */
 public class BaseServiceException extends RuntimeException {
 
   private static final long serialVersionUID = 5028833760039966178L;
+  public static final int UNKNOWN_CODE = 0;
 
   private final int code;
   private final boolean retryable;
+  private final boolean idempotent;
 
-  public BaseServiceException(int code, String message, boolean retryable) {
-    super(message);
-    this.code = code;
-    this.retryable = retryable;
+  public BaseServiceException(IOException exception, boolean idempotent) {
+    super(exception.getMessage(), exception);
+    this.code = statusCode(exception);
+    this.idempotent = idempotent;
+    this.retryable = idempotent && isRetryable(exception);
   }
 
-  public BaseServiceException(int code, String message, boolean retryable, Exception cause) {
+  public BaseServiceException(GoogleJsonError error, boolean idempotent) {
+    super(error.getMessage());
+    this.code = error.getCode();
+    this.idempotent = idempotent;
+    this.retryable = idempotent && isRetryable(error);
+  }
+
+  public BaseServiceException(int code, String message, boolean retryable, boolean idempotent) {
+    super(message);
+    this.code = code;
+    this.idempotent = idempotent;
+    this.retryable = idempotent && retryable;
+  }
+
+  public BaseServiceException(int code, String message, boolean retryable, boolean idempotent,
+      Exception cause) {
     super(message, cause);
     this.code = code;
-    this.retryable = retryable;
+    this.idempotent = idempotent;
+    this.retryable = idempotent && retryable;
+  }
+
+  protected Set<Integer> retryableCodes() {
+    return Collections.emptySet();
+  }
+
+  protected int statusCode(IOException exception) {
+    if (exception instanceof GoogleJsonResponseException) {
+      if (((GoogleJsonResponseException) exception).getDetails() != null) {
+        return ((GoogleJsonResponseException) exception).getDetails().getCode();
+      }
+    }
+    return UNKNOWN_CODE;
+  }
+
+  protected boolean isRetryable(GoogleJsonError error) {
+    return error != null && retryableCodes().contains(error.getCode());
+  }
+
+  protected boolean isRetryable(IOException exception) {
+    if (exception instanceof GoogleJsonResponseException) {
+      return isRetryable(((GoogleJsonResponseException) exception).getDetails());
+    }
+    if (exception instanceof SocketTimeoutException) {
+      return true;
+    }
+    return false;
   }
 
   /**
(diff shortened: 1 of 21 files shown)


ajkannan:
Yes, I think ResourceManagerException could be refactored similary to how you refactor Storage.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -16,26 +16,80 @@
...
+  public BaseServiceException(int code, String message, boolean retryable, boolean idempotent,
+      Exception cause) {
     super(message, cause);
     this.code = code;
-    this.retryable = retryable;
+    this.idempotent = idempotent;
+    this.retryable = idempotent && retryable;
+  }
+
+  protected Set<Integer> retryableCodes() {
```
ajkannan: Nit (and maybe not relevant if this doesn't particular PR isn't meant to be merged in), but would it be viable to make this method (and the class) abstract?

aozarov: +1 for making the class abstract (to discourage throwing it directly).
However, I am fine with a default empty set of retryable codes.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseDatastoreBatchWriter.java
```diff
@@ -64,10 +67,8 @@ public final void addWithDeferredIdAllocation(FullEntity<?>... entities) {
 
   private void addInternal(FullEntity<Key> entity) {
     Key key = entity.key();
-    if (toAdd.containsKey(key) || toUpdate.containsKey(key) || toPut.containsKey(key)) {
-      throw newInvalidRequest("Entity with the key %s was already added or updated in this %s",
-          entity.key(), name);
-    }
+    checkArgument(!toAdd.containsKey(key) && !toUpdate.containsKey(key) && !toPut.containsKey(key),
```
aozarov: This is actually changing the possible set of exceptions thrown by the Datastore.

Before it was DatastoreException with DatastoreError.FAILED_PRECONDITION as error code
and now IllegalArgumentExcetpion.

Though I think this change is reasonable (and aligned with other services) I am not sure it must
be done in this PR (I would rather have a separate issue for such change).
In any case, when we change that we should also change the "@throws" sets of the relevant methods.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -16,26 +16,80 @@
...
+    return Collections.emptySet();
+  }
+
+  protected int statusCode(IOException exception) {
+    if (exception instanceof GoogleJsonResponseException) {
```
aozarov: I think the changes you made to DatastoreException and DatastoreRpcException are reasonable however I wonder what do you think if we apply these changes instead:

1) include a notion of a reason (String) to base service exception using `GoogleJsonError.getMessage()`.

2) replace `retryableCodes` with a pair of (code, reason).

3) retryable is true only when there is a match of the pair, however null value means *any*.
  - it should be possible to have e.g. (null, "InternalError") [null code but with a reason] to
    support the `StorageException` case.

4) change DatastoreRpcException to throw GoogleJsonResponseException (translate `com.google.api.services.datastore.client.DatastoreException` to `GoogleJsonResponseException` in `DefaultDatastoreRpc` and change signature of the DatastoreRpc to throw IOExcetion instead of DatastoreRpcException.


* I would still hold off in changing the `Datastore` api layer to throw `IllegalArgumentException` instead of `DatastoreException` and do it in a separate PR (or change the api javadoc in this one).

aozarov:
I have mentioned few ideas that I think we should discuss.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mziccard: 1) 2) and 3) What holds me from including (and relying on) the notion of *reason* is that reasons are not well documented across services. For instance there seem to be no naming convention: Datastore uses underscore separated, upper-case strings. Storage instead uses camel-case strings (not even sure about this)

1) Isn't the reason stored in `GoogleJsonError.getErrors().get(0).getReason()`?

4) This is does not seem to be what we do in `Storage`, isn't it? Do you see a benefit in doing this?

> I would still hold off in changing the Datastore api layer to throw IllegalArgumentException instead of DatastoreException and do it in a separate PR (or change the api javadoc in this one).

This PR was meant for discussion so we can trash it a start from scratch.

aozarov:
Using this change could we use one common exception handler Interceptor?

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
aozarov: > 1) 2) and 3) What holds me from including (and relying on) 

Though consistency for reason naming across services would be nice it should not be a problem as we treat them just as string and it is the service that defines which strings indicates a retryable exception.

> 1) Isn't the reason stored in `GoogleJsonError.getErrors().get(0).getReason()`?

Yes, that is correct we should use that one instead of `GoogleJsonError.getMessage()`

> 4) This is does not seem to be what we do in Storage

I agree, consistency would be prefered and we can actually take the extra step
and throw the DatastoreException from the RPC layer (initially for Datastore there was a complete separation of the API and SPI including using API from SPI [which in most parts should be OK).

mziccard:
Looking a bit more into this I noticed that in storage we compare the message to "InternalError" to tell if an exception is retryable:
```java
"InternalError".equals(exception.getMessage())
```
Looking [here](https://cloud.google.com/storage/docs/json_api/v1/status-codes) I noticed that while `InternalError` is the error's name the messages seem to have the following format: `Login Required` (notice the space in the middle).

> 2) replace retryableCodes with a pair of (code, reason).
3) retryable is true only when there is a match of the pair, however null value means any.

If we follow @aozarov suggestion we should compare reasons, not messages. Again looking at the example [here](https://cloud.google.com/storage/docs/json_api/v1/status-codes) reasons seem to have a different format: for a 404 error in fact the message is "Not Found" and the reason is "notFound".

Am I missing something? Do we have a comprehensive list of reasons?

aozarov:
@mziccard, this can be closed, right?

mziccard:
Closed as already addressed in #554

PR #441, opened 2015-12-06
Initial commit of LocalResourceManagerHelper and tests

ajkannan:


mziccard:
I added a bunch of minor comments, this is really an amazing and much-much-needed job!

ajkannan:
Thanks for the first pass @mziccard!  I've pushed another commit taking into account the comments.  A few notes on my changes:

* I left in `changeCreateTime` and `changeProjectNumber` in case the user wants to have some autogenerated values and some non-autogenerated (explained more fully in a response to the original comment).
* I decided not to use project's `clone` function because it creates a deep copy, which is problematic when users use `ImmutableMap` (because `ImmutableMap` doesn't have a public constructor).
* I agree that the testing filename may need to change, but I haven't decided what to call the test file yet.  It's half rpc-level tests and half `LocalResourceManagerHelper` tests, but other modules don't have dedicated spi layer tests.  Once I merge in the api-level `ResourceManagerImpl` (in a future PR), I could potentially change the direct RPC calls and then I could split the tests in this file up to be `ResourceManagerImplTest` and `LocalResourceManagerHelperTest`.  To get at how the mock handles the different requests, I think I'll inevitably be testing either the rpc layer or the api layer as a by-product because a lot of the functionality is done by private methods.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+        String requestBody = decodeContent(exchange.getRequestHeaders(), exchange.getRequestBody());
+        response = replace(jsonFactory.fromString(requestBody, Project.class));
+      }
+      if (response == null) {
+        throw new UnsupportedOperationException("Request not recognized.");
+      }
+      exchange.getResponseHeaders().set("Content-type", "application/json; charset=UTF-8");
+      exchange.sendResponseHeaders(response.code(), response.body().length());
+      OutputStream outputStream = exchange.getResponseBody();
+      outputStream.write(response.body().getBytes());
```
aozarov: `response.body().getBytes(StandardCharsets.UTF_8)`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
+package com.google.gcloud.resourcemanager.testing;
```
aozarov: file contains many trailing whitespaces.
see [here](http://stackoverflow.com/questions/1043433/how-to-auto-remove-trailing-whitespace-in-eclipse) for fixiing suggestions.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/package-info.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+/**
+ * A testing helper for Google Cloud Resource Manager.
+ *
+ * <p>A simple usage example:
+ * <p>Before the test:
+ * <pre> {@code
+ * LocalResourceManagerHelper resourceManagerHelper = LocalResourceManagerHelper.create();
+ * // TODO(ajaykannan): implement the following line when ResourceManagerImpl is checked in.
```
aozarov: I am not sure these TODOs are relevant in this file but if you decide to keep them make sure we fix that later.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/package-info.java
```diff
@@ -0,0 +1,35 @@
...
+ * ResourceManager resourceManager = resourceManagerHelper.options().service();
+ * implement this in the next PR
+ * resourceManager.list();
```
aozarov: We should either do something with the list or not have it here

gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/LocalResourceManagerHelperTest.java
```diff
@@ -0,0 +1,511 @@
...
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+public class LocalResourceManagerHelperTest {
+
+  private static final int PORT = 8080;
```
aozarov: We should not use a fixed port as that could be used when running the test.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+import java.util.Random;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.Logger;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Utility to create a local Resource Manager mock for testing.
+ *
+ * The mock runs in a separate thread, listening to port 8080 for HTTP requests.
```
aozarov: We should make the port configurable

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+    private final Response response;
+
+    Error(int code, String domain, String message, String reason, String status) {
+      this.response = new Response(code, toJson(code, domain, message, reason, status));
+    }
+
+    private static String toJson(
+        int code, String domain, String message, String reason, String status) {
+      Map<String, Object> args = new HashMap<>();
+      Map<String, Object> nestedArgs = new HashMap<>();
```
aozarov: why not call this variable errors and put it directly as `args.put("errors", ImmutableList.of(errors))`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  private class RequestHandler implements HttpHandler {
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
+      // see https://cloud.google.com/resource-manager/reference/rest/
+      String path = exchange.getRequestURI().getPath();
+      String requestMethod = exchange.getRequestMethod();
+      Response response = null;
+      if (requestMethod.equals("POST") && path.startsWith("/v1beta1/projects")) {
```
aozarov: Would be nice to make v1beta1 a constant

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+      } else if (requestMethod.equals("GET") && path.startsWith("/v1beta1/projects/")) {
+        response = get(projectIdFromURI(path), parseFields(exchange.getRequestURI().getQuery()));
+      } else if (requestMethod.equals("GET")) {
+        response = list(parseListOptions(exchange.getRequestURI().getQuery()));
+      } else if (requestMethod.equals("PUT")) {
+        String requestBody = decodeContent(exchange.getRequestHeaders(), exchange.getRequestBody());
+        response = replace(jsonFactory.fromString(requestBody, Project.class));
+      }
+      if (response == null) {
+        throw new UnsupportedOperationException("Request not recognized.");
```
aozarov: How is that going to be communicated to the client? Shouldn't it be something like 404 bad request...

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+        } else {
+          String requestBody =
+              decodeContent(exchange.getRequestHeaders(), exchange.getRequestBody());
+          response = create(jsonFactory.fromString(requestBody, Project.class));
+        }
+      } else if (requestMethod.equals("DELETE")) {
+        response = delete(projectIdFromURI(path));
+      } else if (requestMethod.equals("GET") && path.startsWith("/v1beta1/projects/")) {
+        response = get(projectIdFromURI(path), parseFields(exchange.getRequestURI().getQuery()));
+      } else if (requestMethod.equals("GET")) {
```
aozarov: This could have been combined with the prior GET (and inside check for path match).

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+      outputStream.close();
+    }
+  }
+
+  private static String decodeContent(Headers headers, InputStream inputStream) throws IOException {
+    List<String> contentEncoding = headers.get("Content-encoding");
+    byte[] bytes;
+    if (contentEncoding != null && contentEncoding.size() > 0
+        && contentEncoding.get(0).contains("gzip")) {
+      bytes = ByteStreams.toByteArray(new GZIPInputStream(inputStream));
```
aozarov: `input = new GZIPInputStream(inputStream)` and then you can share the code of toByteArray.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+    } else {
+      bytes = ByteStreams.toByteArray(inputStream);
+    }
+    return new String(bytes, StandardCharsets.UTF_8);
+  }
+
+  private static String projectIdFromURI(String path) {
+    String[] pathSplit = path.split("/");
+    if (pathSplit.length < 4) {
+      throw new IllegalArgumentException("This path doesn't have a project ID");
```
aozarov: Maybe display the given path?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+    }
+    return path.split("/")[3].split(":")[0];
```
aozarov: `pathSplit[3].split(":")[0]` ?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  private static String[] parseFields(String query) {
+    return query != null ? query.split("=")[1].split(",") : null;
```
aozarov: Are you guaranteed to have at least one query param?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+          case "pageSize":
+            // support pageSize when Cloud Resource Manager supports this (#421)
+            break;
+        }
+      }
+    }
+    return options;
+  }
+
+  private static final boolean isValidProject(Project project) {
```
aozarov: Nit, probably more useful would have been to return the error string as the message for the error response.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+    if (value.length() > 0 && (!Character.isLetter(value.charAt(0)) || value.endsWith("-"))) {
+      return false;
+    }
+    return value.length() >= minLength && value.length() <= maxLength;
+  }
+
+  Response create(Project project) throws IOException {
+    project.setLifecycleState("ACTIVE");
+    project.setProjectNumber(Math.abs(PROJECT_NUMBER_GENERATOR.nextLong()));
+    project.setCreateTime(ISODateTimeFormat.dateTime().print(new Date().getTime()));
```
aozarov: you can replace `new Date().getTime()` with `System.currentTimeMillis()`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+    }
+    if (value.length() > 0 && (!Character.isLetter(value.charAt(0)) || value.endsWith("-"))) {
+      return false;
+    }
+    return value.length() >= minLength && value.length() <= maxLength;
+  }
+
+  Response create(Project project) throws IOException {
+    project.setLifecycleState("ACTIVE");
+    project.setProjectNumber(Math.abs(PROJECT_NUMBER_GENERATOR.nextLong()));
```
aozarov: Nit, that conceptually can still return a negative value (though extremely unlikely - nextLong returns Long.MIN_VALUE)

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,588 @@
...
+    project.setCreateTime(ISODateTimeFormat.dateTime().print(new Date().getTime()));
+    Response response;
+    if (!isValidProject(project)) {
+      response = Error.INVALID_ARGUMENT.response;
+    } else if (projects.containsKey(project.getProjectId())) {
+      response = Error.ALREADY_EXISTS.response;
+      log.info(
```
aozarov: Same comment, we should prefer communicating request level errors via the error response rather than using the logs.

aozarov:
Half way through. Will complete the review tomorrow.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/package-info.java
ajkannan: I removed the TODOs

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/package-info.java
ajkannan: removed `list`

gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/LocalResourceManagerHelperTest.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Added a check for empty string

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

ajkannan:
@aozarov added a commit to incorporate your comments thus far

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    }
+    return response;
+  }
+
+  private LocalResourceManagerHelper() {
+    InetSocketAddress addr = new InetSocketAddress(0);
+    try {
+      server = HttpServer.create(addr, 0);
+      port = server.getAddress().getPort();
+      server.createContext("/", new RequestHandler());
```
aozarov: Any reason not to specify `"/" + VERSION + "/projects/"` as the context?

Also, I would hope that in that case getting the projectId would be simple if
`getRequestURI().getPath()` would return a relative path to the context.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+
+  private class RequestHandler implements HttpHandler {
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
+      // see https://cloud.google.com/resource-manager/reference/rest/
+      String path = exchange.getRequestURI().getPath();
+      String requestMethod = exchange.getRequestMethod();
+      Response response = null;
+      if (requestMethod.equals("POST") && path.startsWith("/" + VERSION + "/projects")) {
+        if (path.contains("undelete")) {
```
aozarov: Nit, should it `endsWith(":undelete")`?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    InputStream input = inputStream;
+    if (contentEncoding != null && contentEncoding.size() > 0
+        && contentEncoding.get(0).contains("gzip")) {
+      input = new GZIPInputStream(inputStream);
+    }
+    return new String(ByteStreams.toByteArray(input), StandardCharsets.UTF_8);
+  }
+
+  private static String projectIdFromURI(String path) throws IOException {
+    String[] pathSplit = path.split("/");
```
aozarov: Consider my suggestion for the context which would hopefully simplify this.

If not, another option would be to have as a static constant BASE_URI with ("/v1beta1/projects/" created using the version constant) and you can use URI [function](http://docs.oracle.com/javase/6/docs/api/java/net/URI.html#relativize(java.net.URI)) to get the extra.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    Response response(String message) {
+      try {
+        return new Response(code, toJson(message));
+      } catch (IOException e) {
+        return Error.INTERNAL_ERROR.response("Error when generating JSON error response");
+      }
+    }
+
+    private String toJson(String message) throws IOException {
+      Map<String, Object> args = new HashMap<>();
```
aozarov: Nit, move to before first use.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      args.put("code", code);
+      args.put("message", message);
+      args.put("status", status);
+      return jsonFactory.toString(ImmutableMap.of("error", args));
+    }
+  }
+
+  private class RequestHandler implements HttpHandler {
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
```
aozarov: I suggest you remove the `throws IOException` and make sure you handle all exception that may be thrown by the individual calls. all failures should be translated into a json error response if possible.

You will probably need to catch the exception from the OutputStream.write though (and log it as ERROR).

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      // see https://cloud.google.com/resource-manager/reference/rest/
+      String path = exchange.getRequestURI().getPath();
+      String requestMethod = exchange.getRequestMethod();
+      Response response = null;
+      if (requestMethod.equals("POST") && path.startsWith("/" + VERSION + "/projects")) {
```
aozarov: did you try switch on requestMethod?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      } else if (requestMethod.equals("DELETE")) {
+        response = delete(projectIdFromURI(path));
+      } else if (requestMethod.equals("GET")) {
+        if (path.startsWith("/" + VERSION + "/projects/")) {
+          response = get(projectIdFromURI(path), parseFields(exchange.getRequestURI().getQuery()));
+        } else {
+          response = list(parseListOptions(exchange.getRequestURI().getQuery()));
+        }
+      } else if (requestMethod.equals("PUT")) {
+        String requestBody = decodeContent(exchange.getRequestHeaders(), exchange.getRequestBody());
```
aozarov: Nit, you should probably check that request path has the same projectId as in the body.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      OutputStream outputStream = exchange.getResponseBody();
+      outputStream.write(response.body().getBytes(StandardCharsets.UTF_8));
+      outputStream.close();
+    }
+  }
+
+  private static String decodeContent(Headers headers, InputStream inputStream) throws IOException {
+    List<String> contentEncoding = headers.get("Content-encoding");
+    InputStream input = inputStream;
+    if (contentEncoding != null && contentEncoding.size() > 0
```
aozarov: Nit, s/ `contentEncoding.size() > 0` / `!contentEncoding.isEmpty()` /

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+        && contentEncoding.get(0).contains("gzip")) {
```
aozarov: Nit, should be "gzip" or ""x-gzip". see [here](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.5). You don't to try it on "bzip". Any other compression should be rejected with "Unsupported".

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+        }
+      }
+    }
+    return null;
+  }
+
+  private static final boolean isValidIdOrLabel(String value, int minLength, int maxLength) {
+    for (char c : value.toCharArray()) {
+      if (c != '-' && !Character.isDigit(c)
+          && (!Character.isLetter(c) || !Character.isLowerCase(c))) {
```
aozarov: isn't lowerCase covered by isLetter?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      }
+    }
+    if (value.length() > 0 && (!Character.isLetter(value.charAt(0)) || value.endsWith("-"))) {
+      return false;
+    }
+    return value.length() >= minLength && value.length() <= maxLength;
+  }
+
+  Response create(Project project) {
+    project.setLifecycleState("ACTIVE");
```
aozarov: Nit, this 3 lines could probably go inside the `else` part.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    project.setProjectNumber(Math.abs(PROJECT_NUMBER_GENERATOR.nextLong() % Long.MAX_VALUE));
+    project.setCreateTime(ISODateTimeFormat.dateTime().print(System.currentTimeMillis()));
+    Response response;
+    String customErrorMessage = checkForProjectErrors(project);
+    if (customErrorMessage != null) {
+      response = Error.INVALID_ARGUMENT.response(customErrorMessage);
+    } else if (projects.containsKey(project.getProjectId())) {
```
aozarov: this check and the put should be synchronized or better use `putIfAbsent` and complain if return value is not `null`.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      } catch (IOException e) {
+        response =
+            Error.INTERNAL_ERROR.response("Error serializing project " + project.getProjectId());
+      }
+    }
+    return response;
+  }
+
+  Response delete(String projectId) {
+    Project project = projects.get(checkNotNull(projectId));
```
aozarov: We should probably not generate NPE if projectId is missing but rather complain about bad request.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    Response response;
+    if (project == null) {
+      // when possible, change this to 404 (#440)
+      response = Error.PERMISSION_DENIED.response(
+          "Error when deleting " + projectId + " because the project was not found.");
+    } else if (!project.getLifecycleState().equals("ACTIVE")) {
```
aozarov: Nit, this check and change status should be synchronized on the Project.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+          "Error when deleting " + projectId + " because the lifecycle state was not ACTIVE.");
+    } else {
+      project.setLifecycleState("DELETE_REQUESTED");
+      response = new Response(HttpURLConnection.HTTP_OK, "{}");
+    }
+    return response;
+  }
+
+  Response get(String projectId, String[] fields) {
+    if (!projects.containsKey(checkNotNull(projectId))) {
```
aozarov: we should avoid checkNotNull here and rather detect it and response accordingly.

Throwing an NPE which will probably (and hopefully) be caught by the HttpServer and translated to the client as some error (probably 500) is probably not what the client will see when using the real service.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      // when possible, change this to 404 (#440)
+      return Error.PERMISSION_DENIED.response("Project " + projectId + " not found.");
+    }
+    Project project = projects.get(projectId);
+    try {
+      return new Response(
+          HttpURLConnection.HTTP_OK, jsonFactory.toString(extractFields(project, fields)));
```
aozarov: Nit, if you static import HTTP_OK you may avoid the need for few line wraps.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    } catch (IOException e) {
+      return Error.INTERNAL_ERROR.response(
+          "Error when serializing project " + project.getProjectId());
+    }
+  }
+
+  Response list(final Map<String, Object> options) {
```
aozarov: why final?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      Boolean includeProject = includeProject(p, (String[]) options.get("filter"));
+      if (includeProject) {
+        try {
+          projectsSerialized.add(
+              jsonFactory.toString(extractFields(p, (String[]) options.get("fields"))));
+        } catch (IOException e) {
+          return Error.INTERNAL_ERROR.response(
+              "Error when serializing project " + p.getProjectId());
+        }
+      } else if (includeProject == null) {
```
aozarov: Don't you just skip the project in that case? Do you have a test for it?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      } else {
+        return null;
+      }
+    }
+    return true;
+  }
+
+  private static boolean satisfiesFilter(String projectValue, String filterValue) {
+    if (projectValue == null) {
+      return false;
```
aozarov: should you return `false` even if the filterValue was '*'? (I am thinking about name for example).

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+      return Error.INTERNAL_ERROR.response(
+          "Error when serializing project " + project.getProjectId());
+    }
+  }
+
+  Response list(final Map<String, Object> options) {
+    // Use pageSize and pageToken options when Cloud Resource Manager does so (#421)
+    List<String> projectsSerialized = new ArrayList<>();
+    for (Project p : projects.values()) {
+      Boolean includeProject = includeProject(p, (String[]) options.get("filter"));
```
aozarov: Nit, get the filter before the loop.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
aozarov: Actually I see now what you are doing. You were using the `null` as a third value to indicate bad request. I think that is confusing and should be avoided. Plus, I am sure there is no test for it as I would expect the unboxing in line 321 "if (includeProject)" to trigger NPE in that case.

As suggested above get the filter before the loop, run a check on it and then change includeProject to return `boolean`.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+          return Error.INTERNAL_ERROR.response(
+              "Error when serializing project " + p.getProjectId());
+        }
+      } else if (includeProject == null) {
+        return Error.INVALID_ARGUMENT.response("Could not parse the filter.");
+      }
+    }
+    StringBuilder responseBody = new StringBuilder();
+    responseBody.append("{\"projects\": [");
+    responseBody.append(Joiner.on(",").join(projectsSerialized));
```
aozarov: More optimized version would be: Joiner.on(",").[appendTo](http://guava-libraries.googlecode.com/svn-history/r14/trunk/javadoc/com/google/common/base/Joiner.html#appendTo(A, java.lang.Iterable))(responseBody, projectsSerialized);

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    } catch (IOException e) {
+      return Error.INTERNAL_ERROR.response(
+          "Error when serializing project " + project.getProjectId());
+    }
+  }
+
+  Response list(final Map<String, Object> options) {
+    // Use pageSize and pageToken options when Cloud Resource Manager does so (#421)
+    List<String> projectsSerialized = new ArrayList<>();
+    for (Project p : projects.values()) {
```
aozarov: does the production service return projects with non active state?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+
+  /**
+   * Stops the thread that runs the mock Resource Manager server.
+   */
+  public void stop() {
+    server.stop(1);
+  }
+
+  /**
+   * Utility method to add a project.
```
aozarov: I am not sure I understand the purpose of the "Utility" methods inside the helper.
Why can't any "setup" method to add/remove project be done via the service or rpc layer?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -0,0 +1,593 @@
...
+    try {
+      return new Response(HttpURLConnection.HTTP_OK, jsonFactory.toString(project));
+    } catch (IOException e) {
+      return Error.INTERNAL_ERROR.response(
+          "Error when serializing project " + project.getProjectId());
+    }
+  }
+
+  Response undelete(String projectId) {
+    Project project = projects.get(checkNotNull(projectId));
```
aozarov: NPE.

aozarov:
Finished with reviewing the other half.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: After checking the behavior on API explorer, I think Cloud Resource Manager actually doesn't return an error if the project IDs don't match.  Instead, they use the project ID in the request header and ignore the project ID in the body.  I changed the mock to mimic this behavior, but it probably doesn't really make too much of a difference for our purposes because in the spi layer we only allow the users to specify one project ID.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Yes if they're in `DELETE_REQUESTED` state and probably yes `DELETE_IN_PROGRESS`.  The latter is hard to test because I only got notifications that my projects were going to be deleted (after calling delete() to put them in the `DELETE_REQUESTED` state) a couple days after deleting them, so it's hard to reproduce an instance and check it at the right time.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: I removed all utility functions except `changeLifecycleState` (important for testing when a project goes `ACTIVE` -> `DELETE_REQUESTED` -> `DELETE_IN_PROGRESS`) and `removeProject` so that a project can truly be deleted (which happens in the Cloud Resource Manager after the `DELETE_IN_PROGRESS` state)

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: yes, fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Made an issue to deal with synchronization because I want to think a little bit more about any potential deadlocks (since there's already a lock on the concurrentHashMap)

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Removed (it initially was there when I was using `options` inside an `Iterables.transform` which was since taken out)

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: fixed

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Yes, according to the [public documentation](https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/list).  So for example `name:*` only returns projects that have a non-null name.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Done

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Fixed

ajkannan:
@aozarov incorporated your comments, ready for another look

PR #443, opened 2015-12-09
Add InsertAllRequest, InsertAllResponse and model classes for table data

mziccard:
This PR will probably require some discussion :) I am eager to receive any feedback.
It adds:
- `InsertAllRequest` and `InsertAllResponse` to insert data using [TableData: insertAll](https://cloud.google.com/bigquery/docs/reference/v2/tabledata/insertAll).
- `TableRow` and `FieldValue` to hold the results of [TableData: list](https://cloud.google.com/bigquery/docs/reference/v2/tabledata/list) and [query](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query)

Remarks:
- We discussed with @aozarov about adding more information to [FieldValue.Kind](https://github.com/mziccard/gcloud-java/blob/20dfb487637d8d31903a352e443867e505c3ac06/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java#L30) as `REPEATED_OF_PRIMITIVE` and `REPEATED_OF_RECORD` but I realized that an empty repeated field is returned as an empty array and there's no way to say whether it is of primitive or record type.

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
@@ -31,8 +31,8 @@ import java.util.Objects;
 
 /**
  * Google BigQuery Table Field Value class. Objects of this class represent values of a BigQuery
- * Table Field. A list of values forms a {@link TableRow}. Tables rows can be gotten as the result
- * of a query or when listing table data.
+ * Table Field. A list of values forms a table row. Tables rows can be gotten as the result of a
+ * query or when listing table data.
  */
 public class FieldValue implements Serializable {
 
@@ -99,7 +99,8 @@ public class FieldValue implements Serializable {
   }
 
   /**
-   * Returns this field's value as an {@link Object}.
+   * Returns this field's value as an {@link Object}. If {@link #isNull()} is {@code true} this
+   * method returns {@code null}.
    */
   public Object value() {
     return value;
@@ -112,9 +113,11 @@ public class FieldValue implements Serializable {
    * {@link Field.Type#timestamp()}).
    *
    * @throws ClassCastException if the field is not a primitive type
+   * @throws NullPointerException if {@link #isNull()} returns {@code true}
    */
   @SuppressWarnings("unchecked")
   public String stringValue() {
+    checkNotNull(value);
     return (String) value;
   }
 
@@ -155,7 +158,6 @@ public class FieldValue implements Serializable {
   @SuppressWarnings("unchecked")
   public boolean booleanValue() {
     String stringValue = stringValue();
-    checkNotNull(stringValue);
     checkState(stringValue.equalsIgnoreCase("true") || stringValue.equalsIgnoreCase("false"),
         "Field value is not of boolean type");
     return Boolean.parseBoolean(stringValue);
@@ -183,9 +185,11 @@ public class FieldValue implements Serializable {
    * {@link Attribute#REPEATED}).
    *
    * @throws ClassCastException if the field has not {@link Field.Mode#REPEATED} mode
+   * @throws NullPointerException if {@link #isNull()} returns {@code true}
    */
   @SuppressWarnings("unchecked")
   public List<FieldValue> repeatedValue() {
+    checkNotNull(value);
     return (List<FieldValue>) value;
   }
 
@@ -195,9 +199,11 @@ public class FieldValue implements Serializable {
    * is {@link Attribute#RECORD}).
    *
    * @throws ClassCastException if the field is not a {@link Field.Type#record(Field...)} type
+   * @throws NullPointerException if {@link #isNull()} returns {@code true}
    */
   @SuppressWarnings("unchecked")
   public List<FieldValue> recordValue() {
+    checkNotNull(value);
     return (List<FieldValue>) value;
   }
 
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
@@ -50,8 +50,7 @@ public class InsertAllRequest implements Serializable {
    * A Google Big Query row to be inserted into a table. Each {@code RowToInsert} has an associated
    * id used by BigQuery to detect duplicate insertion requests on a best-effort basis.
    *
-   * <p>
-   * Example usage of creating a row to insert:
+   * <p>Example usage of creating a row to insert:
    * <pre>    {@code
    *   List<Long> repeatedFieldValue = Arrays.asList(1L, 2L);
    *   Map<String, Object> recordContent = new HashMap<String, Object>();
@@ -128,7 +127,6 @@ public class InsertAllRequest implements Serializable {
     /**
      * Creates a row to be inserted without associated id.
      *
-     * @param id id of the row, used to identify duplicates
      * @param content the actual content of the row
      */
     public static RowToInsert of(Map<String, Object> content) {
@@ -176,8 +174,7 @@ public class InsertAllRequest implements Serializable {
     /**
      * Adds a row to be inserted with associated id.
      *
-     * <p>
-     * Example usage of adding a row with associated id:
+     * <p>Example usage of adding a row with associated id:
      * <pre>    {@code
      *   InsertAllRequest.Builder builder = InsertAllRequest.builder(tableId);
      *   List<Long> repeatedFieldValue = Arrays.asList(1L, 2L);
@@ -198,8 +195,7 @@ public class InsertAllRequest implements Serializable {
     /**
      * Adds a row to be inserted without an associated id.
      *
-     * <p>
-     * Example usage of adding a row without an associated id:
+     * <p>Example usage of adding a row without an associated id:
      * <pre>    {@code
      *   InsertAllRequest.Builder builder = InsertAllRequest.builder(tableId);
      *   List<Long> repeatedFieldValue = Arrays.asList(1L, 2L);
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
@@ -35,7 +35,7 @@ import java.util.Objects;
 /**
  * Google Cloud BigQuery insert all response. Objects of this class possibly contain errors for an
  * {@link InsertAllRequest}. If a row failed to be inserted, the non-empty list of errors associated
- * to that row's index can be obtained with {@link InsertAllResponse#errorsFor(Long)}.
+ * to that row's index can be obtained with {@link InsertAllResponse#errorsFor(long)}.
  * {@link InsertAllResponse#insertErrors()} can be used to return all errors caused by a
  * {@link InsertAllRequest} as a map.
  */


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,207 @@
...
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Google BigQuery Table Field Value class. Objects of this class represent values of a BigQuery
+ * Table Field. A list of values forms a {@link TableRow}. Tables rows can be get as the result of
```
ajkannan: s/can be get/can be gotten/g

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/TableRowTest.java
```diff
@@ -0,0 +1,46 @@
+package com.google.gcloud.bigquery;
```
ajkannan: Nit: license statement missing

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,207 @@
+package com.google.gcloud.bigquery;
```
ajkannan: Nit: license statement missing

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,119 @@
...
+  private final Map<Long, List<BigQueryError>> insertErrors;
+
+  InsertAllResponse(Map<Long, List<BigQueryError>> insertErrors) {
+    this.insertErrors = insertErrors != null ? ImmutableMap.copyOf(insertErrors)
+        : ImmutableMap.<Long, List<BigQueryError>>of();
+  }
+
+  /**
+   * Returns all insertion errors as a map whose keys are indexes of rows that failed to insert. To
+   * each failed row index a non-empty list of {@link BigQueryError} is associated.
```
ajkannan: Nit: I think this sentence would sound slightly better as: "Each failed row index is associated with a non-empty list of {@link BigQueryError}s"

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Google Cloud BigQuery insert all response. Objects of this class possibly contain errors for an
+ * {@link InsertAllRequest}. If a row failed to be inserted, the non-empty list of errors associated
+ * to that row's index can be obtained with {@link InsertAllResponse#errorsFor(Long)}.
+ * {@link InsertAllResponse#insertErrors()} can be used to return all errors caused by a
+ * {@link InsertAllRequest}, as a map.
```
ajkannan: nit: can remove comma

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Returns errors for the provided row index. If no error exists returns {@code null}.
+   */
+  public List<BigQueryError> errorsFor(Long index) {
+    return insertErrors.get(index);
+  }
+
+  /**
+   * Returns {@code true} if no row insertion failed, false otherwise. If {@code true}
```
ajkannan: At the end of this line, should "true" be "false"?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * Return this field's value as an {@link Object}.
+   */
+  public Object value() {
+    return value;
+  }
+
+  /**
+   * Return this field's value as an {@link String}.
+   */
```
ajkannan: Should this javadoc have: "This method should only be used if the corresponding field has {@link Field.Type#string}" warning and `@throws` statement?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
```diff
@@ -0,0 +1,400 @@
...
+      this.content = checkNotNull(content);
+    }
+
+    /**
+     * Creates a row to be inserted.
+     *
+     * @param id id of the row, used to identify duplicates
+     * @param content the actual content of the row
+     */
+    public RowToInsert(String id, Map<String, Object> content) {
```
ajkannan: Based off previous code, it looks like `gcloud-java` has the precedent of using `of` in this type of situation rather than expose the constructor.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/FieldValueTest.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final Map<String, String> FLOAT_FIELD = ImmutableMap.of("v", "1.5");
+  private static final Map<String, String> STRING_FIELD = ImmutableMap.of("v", "string");
+  private static final Map<String, String> TIMESTAMP_FIELD = ImmutableMap.of("v", "42");
+  private static final Map<String, Object> REPEATED_FIELD =
+      ImmutableMap.<String, Object>of("v", ImmutableList.<Object>of(INTEGER_FIELD, INTEGER_FIELD));
+  private static final Map<String, Object> RECORD_FIELD =
+      ImmutableMap.<String, Object>of("f", ImmutableList.<Object>of(FLOAT_FIELD, TIMESTAMP_FIELD));
+
+  @Test
+  public void testFromPb() {
```
ajkannan: Should there be a check for a FieldValue with value `null`?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/FieldValueTest.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import org.junit.Test;
+
+import java.util.Map;
+
+public class FieldValueTest {
+
+  private static final TableCell BOOLEAN_FIELD = new TableCell().setV("false");
+  private static final Map<String, String> INTEGER_FIELD = ImmutableMap.of("v", "1");
```
ajkannan: Any reason these aren't `new TableCell().setV(...)`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableRow.java
```diff
@@ -0,0 +1,104 @@
...
+  public int hashCode() {
+    return Objects.hash(values);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof TableRow && Objects.equals(values, ((TableRow) obj).values);
+  }
+
+  static TableRow fromPb(com.google.api.services.bigquery.model.TableRow rowPb) {
```
ajkannan: I think we should add a test for this method in TableRowTest

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/TableRowTest.java
```diff
@@ -0,0 +1,46 @@
...
+public class TableRowTest {
+
+  private static final FieldValue BOOL_VALUE = new FieldValue(FieldValue.Kind.PRIMITIVE, "false");
+  private static final FieldValue INT_VALUE = new FieldValue(FieldValue.Kind.PRIMITIVE, "1");
+  private static final FieldValue FLOAT_VALUE = new FieldValue(FieldValue.Kind.PRIMITIVE, "1.5");
+  private static final FieldValue STRING_VALUE = new FieldValue(FieldValue.Kind.PRIMITIVE, "str");
+  private static final FieldValue TIME_VALUE = new FieldValue(FieldValue.Kind.PRIMITIVE, "42");
+  private static final List<FieldValue> ROW_VALUES =
+      ImmutableList.of(BOOL_VALUE, INT_VALUE, FLOAT_VALUE, STRING_VALUE, TIME_VALUE);
+  private static final TableRow TABLE_ROW1 = new TableRow.Builder()
```
ajkannan: Could there theoretically be `FieldValue`s with value `null` in a TableRow?  If so I think we should add a `NULL_VALUE`

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/InsertAllRequestTest.java
```diff
@@ -0,0 +1,161 @@
...
+    assertEquals(IGNORE_UNKNOWN_VALUES, INSERT_ALL_REQUEST3.ignoreUnknownValues());
+    assertEquals(IGNORE_UNKNOWN_VALUES, INSERT_ALL_REQUEST4.ignoreUnknownValues());
+    assertEquals(IGNORE_UNKNOWN_VALUES, INSERT_ALL_REQUEST5.ignoreUnknownValues());
+    assertEquals(IGNORE_UNKNOWN_VALUES, INSERT_ALL_REQUEST6.ignoreUnknownValues());
+    assertEquals(IGNORE_UNKNOWN_VALUES, INSERT_ALL_REQUEST7.ignoreUnknownValues());
+    assertEquals(IGNORE_UNKNOWN_VALUES, INSERT_ALL_REQUEST8.ignoreUnknownValues());
+  }
+
+  @Test
+  public void testOf() {
```
ajkannan: nit: I think two of the `of` methods with varargs in the method signature aren't tested

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/InsertAllRequestTest.java
```diff
@@ -0,0 +1,161 @@
...
+  private static final Map<String, Object> CONTENT2 =
+      ImmutableMap.<String, Object>of("key", "val2");
+  private static final List<InsertAllRequest.RowToInsert> ROWS =
+      ImmutableList.of(new InsertAllRequest.RowToInsert(CONTENT1),
+          new InsertAllRequest.RowToInsert(CONTENT2));
+  private static final List<InsertAllRequest.RowToInsert> ROWS_WITH_ID =
+      ImmutableList.of(new InsertAllRequest.RowToInsert("id1", CONTENT1),
+          new InsertAllRequest.RowToInsert("id2", CONTENT2));
+  private static final TableId TABLE_ID = TableId.of("dataset", "table");
+  private static final boolean SKIP_INVALID_ROWS = true;
```
ajkannan: Nit: for the Boolean values, I think it would be nice to vary the values.  Say someone in the future accidentally always sets SKIP_INVALID_ROWS to true or IGNORE_UNKNOWN_VALUES to false, then this test wouldn't catch that.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/InsertAllRequestTest.java
```diff
@@ -0,0 +1,161 @@
...
+  private static final boolean IGNORE_UNKNOWN_VALUES = false;
+  private static final InsertAllRequest INSERT_ALL_REQUEST1 = InsertAllRequest.builder(TABLE_ID)
```
ajkannan: Nit: one of the 8 builders used here could use the `BaseTableInfo` parameter

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/FieldValueTest.java
mziccard: I am trying to test different types that can be passed to `fromPb`. First level field values (in a row) are passed as a `TableCell` object, while field values nested in a record are just an `ArrayMap` hence I am using `instance of Map` in `fromPb`.

mziccard:
Thanks @ajkannan for the pass. I fixed the comments in place.

mziccard:
Should we make all value getters in `FieldValue` throw the same exception? Right now they either throw `ClassCastException`, `IllegalStateException`, `NumberFormatException` or `NullPointerException` depending on the type of conversion being asked. What do you think?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,232 @@
...
+  private static final int MICROSECONDS = 1000000;
+  private static final long serialVersionUID = 469098630191710061L;
+
+  private final Kind kind;
+  private final Object value;
+
+  /**
+   * The field value's kind, giving information on the field's content type.
+   */
+  public enum Kind {
```
aozarov: I wish we could find some other name (as kind is used for a different purpose in the underlying layer ). I am not sure what though. `Type`, `TypeAttributes`, ..

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,232 @@
...
+  @SuppressWarnings("unchecked")
+  public String stringValue() {
+    return (String) value;
+  }
+
+  /**
+   * Returns this field's value as a {@link Long}. This method should only be used if the
+   * corresponding field has {@link Field.Type#integer()} type.
+   *
+   * @throws NumberFormatException if the field's value could not be converted to {@link Integer}
```
aozarov: here and below - could also throw ClassCastException

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
mziccard: I did not pick `Type` to avoid confusion with `Field.Type` but you are right that kind is also overloaded. I couldn't come up with a better idea though

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
mziccard: Fixed in 1674c9c

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
aozarov: `Attribute`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,236 @@
...
+    return value;
+  }
+
+  /**
+   * Return this field's value as a {@link String}. This method should only be used if the
+   * corresponding field has primitive type ({@link Field.Type#bool()}, {@link Field.Type#string()},
+   * {@link Field.Type#floatingPoint()}, {@link Field.Type#integer()},
+   * {@link Field.Type#timestamp()}).
+   *
+   * @throws ClassCastException if the field has not primitive type
```
aozarov: here and below - is it "has not" or "is not a"?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/TableRowTest.java
ajkannan: Looking a bit more, it seems like TableRows can in fact contain null FieldValues.  Am I wrong in my understanding (or was not adding a `NULL_VALUE` to this table row an accidental leave-out)?  Also it might be nice to add a `repeatedValue` and a `recordValue` as well.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,236 @@
...
+   * Returns this field's value as a {@link Boolean}. This method should only be used if the
+   * corresponding field has {@link Field.Type#bool()} type.
+   *
+   * @throws ClassCastException if the field has not primitive type
+   * @throws IllegalStateException if the field's value could not be converted to {@link Boolean}
+   */
+  @SuppressWarnings("unchecked")
+  public boolean booleanValue() {
+    String stringValue = stringValue();
+    checkState(stringValue.equalsIgnoreCase("true") || stringValue.equalsIgnoreCase("false"),
```
aozarov: I assume because we are using the json format it is always going to be "true" or "false" and this class will never be used for a CSV format which also supports 1 or 0?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  public Kind kind() {
+    return kind;
+  }
+
+  /**
+   * Return this field's value as an {@link Object}.
+   */
+  public Object value() {
+    return value;
```
aozarov: what about handling nulls?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
mziccard: Yes, in JSON format booleans should only be "true" or "false"

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
mziccard: Mmmm do you think there's a big difference here? I don't mind changing

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/TableRowTest.java
mziccard: You are right, it can contain null values as well as repeated and record, but what's the point of adding them too? I mean: we are unit testing `TableRow` which is merely a collection of field values and we already unit-tested `FieldValue` so I don't think this is really needed. But I'll add them anyway.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
ajkannan: I think "is not a" sounds better, I support changing it

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
mziccard: I'm on the fence on this so if `Attribute` is your preference I am more than ok to go with it.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
mziccard: What do you mean with *handling nulls*?

aozarov: I assume `value` could be `null` for primitive value or record.

If so, should we have `isNull` method and should we indicate that NPE will be thrown
from the xxxValue methods if `isNull` is `true`?

mziccard: Oh sure I can add that.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
ajkannan: I think I would slightly prefer `Attribute`.  Just throwing out another option: we could change `Field.Type` to `Field.PrimitiveType`/ `Field.DataType` and then changed `Kind` to `FieldType`.  Having Field.FieldType is a little repetitive, but I couldn't really think of anything better.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Returns this field's value as a {@link Long}, representing a timestamp in microseconds. This
+   * method should only be used if the corresponding field has {@link Field.Type#timestamp()} type.
+   *
+   * @throws ClassCastException if the field has not primitive type
+   * @throws NumberFormatException if the field's value could not be converted to {@link Long}
+   */
+  @SuppressWarnings("unchecked")
+  public long timestampValue() {
+    return new Double(((Double.valueOf(stringValue())) * MICROSECONDS)).longValue();
```
aozarov: So timestamp are always returned in json as a long (time since epoch in seconds)?

mziccard: They are returned as a double value since epoch in the form `1408452095.22` where the integer part is **seconds**: `1408452095.22 == 2014-08-19 07:41:35.220 -05:00`

aozarov: I see. Thanks for the clarification.

I think we should add this description as a comment in the code and also specify in the javadoc that this is "unix time" (or epoch time) .

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Returns this field's value as a {@link Long}. This method should only be used if the
+   * corresponding field has {@link Field.Type#integer()} type.
+   *
+   * @throws ClassCastException if the field has not primitive type
+   * @throws NumberFormatException if the field's value could not be converted to {@link Integer}
+   */
+  @SuppressWarnings("unchecked")
+  public long longValue() {
+    return Long.valueOf(stringValue());
```
aozarov: Any reason you use `valueOf` (which returns `Long`) instead of `parseLong`?
Same for below.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
```diff
@@ -0,0 +1,236 @@
...
+
+  private final Kind kind;
+  private final Object value;
+
+  /**
+   * The field value's kind, giving information on the field's content type.
+   */
+  public enum Kind {
+    /**
+     * A primitive field value. A {@code FieldValue} has type {@link #PRIMITIVE} when the
```
aozarov: Nit,  `{@link #PRIMITIVE}` link to itself

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
```diff
@@ -0,0 +1,399 @@
...
+
+  private final TableId table;
+  private final List<RowToInsert> rows;
+  private final Boolean skipInvalidRows;
+  private final Boolean ignoreUnknownValues;
+
+  /**
+   * A Google Big Query row to be inserted into a table. Each {@code RowToInsert} has an associated
+   * id used by BigQuery to detect duplicate insertion requests on a best-effort basis.
+   *
```
aozarov: Nit, here and below remove empty line before `<p>` tag.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
```diff
@@ -0,0 +1,399 @@
...
+      return this;
+    }
+
+    /**
+     * Adds a row to be inserted with associated id.
+     *
+     * <p>
+     * Example usage of adding a row with associated id:
+     * <pre>    {@code
+     *   InsertAllRequest.Buider builder = InsertAllRequest.builder(tableId);
```
aozarov: s/Buider/Builder/g

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
```diff
@@ -0,0 +1,399 @@
...
+  public static class RowToInsert implements Serializable {
+
+    private static final long serialVersionUID = 8563060538219179157L;
+
+    private final String id;
+    private final Map<String, Object> content;
+
+    RowToInsert(String id, Map<String, Object> content) {
+      this.id = id;
+      this.content = content;
```
aozarov: Should we copy the content?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    /**
+     * Creates a row to be inserted.
+     *
+     * @param id id of the row, used to identify duplicates
+     * @param content the actual content of the row
+     */
+    public static RowToInsert of(String id, Map<String, Object> content) {
+      return new RowToInsert(checkNotNull(id), checkNotNull(content));
```
aozarov: I thought `id` can be `null` or did you plan to provide another version of `of(Map<String, Object> content)` without id?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this).add("insertErrors", insertErrors).toString();
+  }
+
+  TableDataInsertAllResponse toPb() {
+    TableDataInsertAllResponse responsePb = new TableDataInsertAllResponse();
+    if (insertErrors.size() > 0) {
```
aozarov: Nit, `!insertError.isEmpty()`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Objects;
+
+/**
+ * Google Cloud BigQuery insert all response. Objects of this class possibly contain errors for an
+ * {@link InsertAllRequest}. If a row failed to be inserted, the non-empty list of errors associated
+ * to that row's index can be obtained with {@link InsertAllResponse#errorsFor(Long)}.
+ * {@link InsertAllResponse#insertErrors()} can be used to return all errors caused by a
+ * {@link InsertAllRequest} as a map.
+ */
+public class InsertAllResponse implements Serializable {
```
aozarov: missing `serialVersionUID`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,119 @@
...
+   * Each failed row index is associated with a non-empty list of {@link BigQueryError}.
+   */
+  public Map<Long, List<BigQueryError>> insertErrors() {
+    return insertErrors;
+  }
+
+  /**
+   * Returns errors for the provided row index. If no error exists returns {@code null}.
+   */
+  public List<BigQueryError> errorsFor(Long index) {
```
aozarov: why `Long`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,119 @@
...
+    return obj instanceof InsertAllResponse
+        && Objects.equals(insertErrors, ((InsertAllResponse) obj).insertErrors);
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this).add("insertErrors", insertErrors).toString();
+  }
+
+  TableDataInsertAllResponse toPb() {
```
aozarov: is that for testing only?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableRow.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Google BigQuery Table Row class. Objects of this class contain a list of {@link FieldValue}, one
+ * for each field in the table.
+ */
+public class TableRow implements Serializable {
```
aozarov: I am not totally against but I see no reason (or a potential benefit) for making users go an extra step (and call `values()`) for getting the values. Do you?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
mziccard: That's the one time I do something to make checkstyle happy:) It was complaining with:
> \<p\> tag should be precede with empty line.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
mziccard: My idea was that if `RowToInsert` is used an `id` *must* be provided. If the user wants to insert a row without id `addRow(Map<String, Object> content)` can be used.
However, I don't mind adding a new factory method if you prefer.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
mziccard: Yes, the conversions from and to `TableDataInsertAllResponse` are more complex than usual so I added this for easier testing. I could remove it though.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableRow.java
mziccard: There are not practical benefit. As there were no particular ones in having the `Schema` class. I added this just to wrap each row into its own class.
If we remove this class we will have:
```java
Page<List<FieldValue>> results = bigquery.listTableData(tableId);
```
which is ok to me.

mziccard: I removed the `TableRow` class in 50d381f

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
mziccard: Fixed in d2f0a82

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
mziccard: Fixed in d2f0a82

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
mziccard: Fixed in d2f0a82

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
mziccard: Added the missing factory method in d2f0a82

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
mziccard: Done in d2f0a82

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
mziccard: Fixed in d2f0a82

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
aozarov: Do you think we should add a checkNotNull in `stringValue()` (and remove in other places such as `booleanValue`) to make it consistent (get value methods throw NPE if `isNull() == true`)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/FieldValue.java
aozarov: Nit, but I think the same applies to the description part.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
aozarov: You are right! I am not sure why my IDE complaint about it. I guess it was configured with Sun style (which seems to avoid the [empty line](http://www.oracle.com/technetwork/articles/java/index-137868.html))?

Also, when looking at guava code it seems that `<p>` is indeed preceded by an empty line however never after (the paragraph starts on the same line).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllRequest.java
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    /**
+     * Creates a row to be inserted without associated id.
+     *
+     * @param id id of the row, used to identify duplicates
```
aozarov: extra `param`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/InsertAllResponse.java
```diff
@@ -0,0 +1,121 @@
...
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Google Cloud BigQuery insert all response. Objects of this class possibly contain errors for an
+ * {@link InsertAllRequest}. If a row failed to be inserted, the non-empty list of errors associated
+ * to that row's index can be obtained with {@link InsertAllResponse#errorsFor(Long)}.
```
aozarov: `{@link InsertAllResponse#errorsFor(long)}`

aozarov:
Looks great! Few nits.

mziccard:
Fixed all comments in 43eb31b. I added `checkNotNull` to `stringValue`, `repeatedValue` and `recordValue` and documented the NPE. I also documented that `value()` can return `null`. 43eb31b also fixes some other javadoc errors.

PR #446, opened 2015-12-10
Fix checkstyle issues and warnings

ajkannan:


gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
```diff
@@ -71,6 +71,7 @@ public PageImpl(NextPageFetcher<T> pageFetcher, String cursor, Iterable<T> resul
     this.results = results;
   }
 
+  @SuppressWarnings("unchecked")
```
aozarov: I think this may not be necessary if we change `Collections.EMPTY_LIST` to `Collections.emptyList()`

gcloud-java-datastore/src/test/java/com/google/gcloud/datastore/ValueTest.java
```diff
@@ -60,11 +60,13 @@
 
   private ImmutableMap<ValueType, Value<?>> typeToValue;
 
+  @SuppressWarnings("rawtypes")
   private class TestBuilder extends Value.BaseBuilder<Set, Value<Set>, TestBuilder> {
     TestBuilder() {
       super(ValueType.LIST);
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
```
aozarov: is "rawtypes" needed (here and below) once you defined it in the class level?

gcloud-java-storage/src/test/java/com/google/gcloud/storage/BlobWriteChannelImplTest.java
```diff
@@ -234,6 +234,7 @@ public void testStateEquals() {
         .times(2);
     replay(storageRpcMock);
     writer = new BlobWriteChannelImpl(options, BLOB_INFO, EMPTY_RPC_OPTIONS);
+    @SuppressWarnings("resource") // avoid closing when you don't want partial writes to GCS
```
aozarov: partial writes to GCS upon failure?

gcloud-java-storage/src/test/java/com/google/gcloud/storage/ITStorageTest.java
```diff
@@ -77,7 +75,7 @@ public static void beforeClass() {
 
   @AfterClass
   public static void afterClass()
-      throws ExecutionException, TimeoutException, InterruptedException {
+throws ExecutionException, InterruptedException {
```
aozarov: bad indentation.

gcloud-java-storage/src/test/java/com/google/gcloud/storage/SerializationTest.java
```diff
@@ -127,6 +128,7 @@ public void testWriteChannelState() throws IOException, ClassNotFoundException {
         .projectId("p2")
         .retryParams(RetryParams.defaultInstance())
         .build();
+    @SuppressWarnings("resource") // avoid closing when you don't want partial writes to GCS
```
aozarov: same comment as above.

aozarov:
Great. Thanks for doing that. Few nits.

gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java
```diff
@@ -53,6 +53,7 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+@SuppressWarnings("rawtypes")
```
eamonnmcmanus: I think it is a bad idea to apply this to a whole class, especially one as big as this one. The fact that you feel the need says to me that there is something wrong with the class. Do we really need all the type parameters?

ajkannan: Good point, I changed it so that I don't suppress warnings on the entire file.  `Service` was the only problem originally, and I added a type to it now.

ajkannan:
Added a commit to address all the comments thus far

PR #452, opened 2015-12-11
Add query request and query response classes and tests

mziccard:
This PR adds classes for Query Requests and Query Responses:
- `QueryRequest` is used to run queries "on the fly" [Jobs: query](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query)
- `QueryResponse` is used to both hold the result of queries run on the fly ([Jobs: query](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query)) and to hold the result of query jobs ([Jobs: getQueryResults](https://cloud.google.com/bigquery/docs/reference/v2/jobs/getQueryResults))

**Remarks**
`QueryResponse` has no `fromPb` method as it is initialized through the builder inside the `BigQueryServiceImpl` class. This is so because:
- we need service and service options to initialize the field `Page<List<FieldValue>> rows`
- `QueryResponse` is created from 2 pb classes: [QueryResponse](https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/java/latest/com/google/api/services/bigquery/model/QueryResponse.html) and [GetQueryResultsResponse](https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/java/latest/com/google/api/services/bigquery/model/GetQueryResultsResponse.html)
  I preferred to handle the creation inside `BigQueryServiceImpl` rather than having a `fromPb` method with more parameters.

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
@@ -27,7 +27,7 @@ import java.util.Objects;
  * Google Cloud BigQuery Query Request. This class can be used to run a BigQuery SQL query and
  * return results if the query completes within a specified timeout. The query results are saved to
  * a temporary table that is deleted approximately 24 hours after the query is run. The query is run
- * through a BigQuery Job whose identity can be accessed via {@link QueryResponse#job()}.
+ * through a BigQuery Job whose identity can be accessed via {@link QueryResponse#jobId()}.
  *
  * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/query">Query</a>
  * @see <a href="https://cloud.google.com/bigquery/query-reference">Query Reference</a>
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
@@ -17,6 +17,7 @@
 package com.google.gcloud.bigquery;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
 
 import java.io.Serializable;
 import java.util.List;
@@ -30,10 +31,11 @@ import java.util.Objects;
  * <pre>    {@code
  *    QueryResponse response = bigquery.query(request);
  *    while (!response.jobComplete()) {
- *      response = bigquery.getQueryResults(response.job());
+ *      response = bigquery.getQueryResults(response.jobId());
  *      Thread.sleep(1000);
  *    }
  *    List<BigQueryError> executionErrors = response.executionErrors();
+ *    // look for errors in executionErrors
  *    QueryResult result = response.result();
  *    Iterator<List<FieldValue>> rowIterator = result.iterateAll();
  *    while(rowIterator.hasNext()) {
@@ -52,7 +54,7 @@ public class QueryResponse implements Serializable {
 
   private final QueryResult result;
   private final String etag;
-  private final JobId job;
+  private final JobId jobId;
   private final boolean jobComplete;
   private final List<BigQueryError> executionErrors;
 
@@ -60,7 +62,7 @@ public class QueryResponse implements Serializable {
 
     private QueryResult result;
     private String etag;
-    private JobId job;
+    private JobId jobId;
     private boolean jobComplete;
     private List<BigQueryError> executionErrors;
 
@@ -76,8 +78,8 @@ public class QueryResponse implements Serializable {
       return this;
     }
 
-    Builder job(JobId job) {
-      this.job = job;
+    Builder jobId(JobId jobId) {
+      this.jobId = jobId;
       return this;
     }
 
@@ -99,9 +101,10 @@ public class QueryResponse implements Serializable {
   private QueryResponse(Builder builder) {
     this.result = builder.result;
     this.etag = builder.etag;
-    this.job = builder.job;
+    this.jobId = builder.jobId;
     this.jobComplete = builder.jobComplete;
-    this.executionErrors = builder.executionErrors;
+    this.executionErrors = builder.executionErrors != null ? builder.executionErrors
+      : ImmutableList.<BigQueryError>of();
   }
 
   /**
@@ -123,8 +126,8 @@ public class QueryResponse implements Serializable {
    * Returns the identity of the BigQuery Job that was created to run the query. This field will be
    * present even if the original request timed out.
    */
-  public JobId job() {
-    return job;
+  public JobId jobId() {
+    return jobId;
   }
 
   /**
@@ -135,6 +138,15 @@ public class QueryResponse implements Serializable {
    */
   public boolean jobComplete() {
     return jobComplete;
+  }
+
+  /**
+   * Returns whether errors and warnings occurred during the execution of the job. If this method
+   * returns {@code true} it does not necessarily mean that the job has completed or was
+   * unsuccessful.
+   */
+  public boolean hasErrors() {
+    return !executionErrors.isEmpty();
   }
 
   /**
@@ -150,7 +162,7 @@ public class QueryResponse implements Serializable {
     return MoreObjects.toStringHelper(this)
         .add("result", result)
         .add("etag", etag)
-        .add("job", job)
+        .add("jobId", jobId)
         .add("jobComplete", jobComplete)
         .add("executionErrors", executionErrors)
         .toString();
@@ -158,7 +170,7 @@ public class QueryResponse implements Serializable {
 
   @Override
   public int hashCode() {
-    return Objects.hash(job);
+    return Objects.hash(jobId);
   }
 
   @Override
@@ -173,7 +185,7 @@ public class QueryResponse implements Serializable {
     return jobComplete == response.jobComplete
         && Objects.equals(etag, response.etag)
         && Objects.equals(result, response.result)
-        && Objects.equals(job, response.job)
+        && Objects.equals(jobId, response.jobId)
         && Objects.equals(executionErrors, response.executionErrors);
   }
 
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResult.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResult.java
@@ -72,17 +72,17 @@ public class QueryResult extends PageImpl<List<FieldValue>> {
     Builder pageFetcher(QueryResultsPageFetcher pageFetcher) {
       this.pageFetcher = pageFetcher;
       return this;
-    };
+    }
 
     Builder cursor(String cursor) {
       this.cursor = cursor;
       return this;
-    };
+    }
 
     Builder results(Iterable<List<FieldValue>> results) {
       this.results = results;
       return this;
-    };
+    }
 
     QueryResult build() {
       return new QueryResult(this);
@@ -108,7 +108,7 @@ public class QueryResult extends PageImpl<List<FieldValue>> {
   }
 
   /**
-   * Returns the schema of the results.
+   * Returns the schema of the results. This is present only when the query completes successfully.
    */
   public Schema schema() {
     return schema;
--- a/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/QueryResponseTest.java
+++ b/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/QueryResponseTest.java
@@ -19,6 +19,7 @@ package com.google.gcloud.bigquery;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 import com.google.common.collect.ImmutableList;
 
@@ -62,7 +63,7 @@ public class QueryResponseTest {
       .build();
   private static final QueryResponse QUERY_RESPONSE = QueryResponse.builder()
       .etag(ETAG)
-      .job(JOB_ID)
+      .jobId(JOB_ID)
       .jobComplete(JOB_COMPLETE)
       .executionErrors(ERRORS)
       .result(QUERY_RESULT)
@@ -72,9 +73,10 @@ public class QueryResponseTest {
   public void testBuilder() {
     assertEquals(ETAG, QUERY_RESPONSE.etag());
     assertEquals(QUERY_RESULT, QUERY_RESPONSE.result());
-    assertEquals(JOB_ID, QUERY_RESPONSE.job());
+    assertEquals(JOB_ID, QUERY_RESPONSE.jobId());
     assertEquals(JOB_COMPLETE, QUERY_RESPONSE.jobComplete());
     assertEquals(ERRORS, QUERY_RESPONSE.executionErrors());
+    assertTrue(QUERY_RESPONSE.hasErrors());
   }
 
   @Test
@@ -82,9 +84,10 @@ public class QueryResponseTest {
     QueryResponse queryResponse = QueryResponse.builder().jobComplete(false).build();
     assertNull(queryResponse.etag());
     assertNull(queryResponse.result());
-    assertNull(queryResponse.job());
+    assertNull(queryResponse.jobId());
     assertFalse(queryResponse.jobComplete());
-    assertNull(queryResponse.executionErrors());
+    assertEquals(ImmutableList.<BigQueryError>of(), queryResponse.executionErrors());
+    assertFalse(queryResponse.hasErrors());
   }
 
   @Test
@@ -96,8 +99,9 @@ public class QueryResponseTest {
     assertEquals(expected, value);
     assertEquals(expected.etag(), value.etag());
     assertEquals(expected.result(), value.result());
-    assertEquals(expected.job(), value.job());
+    assertEquals(expected.jobId(), value.jobId());
     assertEquals(expected.jobComplete(), value.jobComplete());
     assertEquals(expected.executionErrors(), value.executionErrors());
+    assertEquals(expected.hasErrors(), value.hasErrors());
   }
 }
--- a/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/SerializationTest.java
+++ b/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/SerializationTest.java
@@ -203,7 +203,7 @@ public class SerializationTest {
       .build();
   private static final QueryResponse QUERY_RESPONSE = QueryResponse.builder()
       .etag(ETAG)
-      .job(JOB_ID)
+      .jobId(JOB_ID)
       .jobComplete(true)
       .result(QUERY_RESULT)
       .build();


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Google Cloud BigQuery Query Request. This class can be used to run a BigQuery SQL query and
+ * return results if the query completes within a specified timeout. The query results are saved to
+ * a temporary table that is deleted approximately 24 hours after the query is run. Query is run
+ * through a BigQuery Job whose identity can be accessed via {@link QueryResponse#job()}.
+ *
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/query">Query</a>
+ */
+public class QueryRequest implements Serializable {
```
ajkannan: Looks like the two classes added are both serializable, they should be added to `SerializationTests`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Google Cloud BigQuery Query Request. This class can be used to run a BigQuery SQL query and
+ * return results if the query completes within a specified timeout. The query results are saved to
+ * a temporary table that is deleted approximately 24 hours after the query is run. Query is run
```
ajkannan: Nit: I think the sentence "Query is run..." would sound slightly better with a "The" at the beginning.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+     * to complete and read the results. The default value is 10000 milliseconds (10 seconds).
+     */
+    public Builder maxWaitTime(Long maxWaitTime) {
+      this.maxWaitTime = maxWaitTime;
+      return this;
+    }
+
+    /**
+     * Sets whether the query has to be dry run or not. If set, the query is not executed: if the
+     * query is valid statistics are returned on how many bytes would be processed, if the query is
```
ajkannan: Nit: I think this comment would sound slightly better split up into multiple sentences:

"Sets whether the query will be a dry run or not.  If set, the query is not executed.  If the query is valid, statistics are returned on how many bytes would have been processed.  If the query is invalid, an error is returned."

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+      return this;
+    }
+
+    /**
+     * Sets how long to wait for the query to complete, in milliseconds, before the request times
+     * out and returns. Note that this is only a timeout for the request, not the query. If the
+     * query takes longer to run than the timeout value, the call returns without any results and
+     * with the {@link QueryResponse#jobComplete()} set to {@code false}. You can call
+     * {@link BigQuery#getQueryResults(JobId, BigQuery.QueryResultsOption...)} to wait for the query
+     * to complete and read the results. The default value is 10000 milliseconds (10 seconds).
```
ajkannan: The note about the default value is nice.  I think it'd be good to add the default behavior for the other optional fields if applicable (i.e. `useQueryCache`)

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+
+  /**
+   * Returns the hash of the {@code QueryResponse} resource or {@code null} if not set.
+   */
+  public String etag() {
+    return etag;
+  }
+
+  /**
+   * Returns the schema of the results when the query completed successfully. Returns {@code null}
```
ajkannan: s/ when / if

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+import com.google.gcloud.Page;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Google Cloud BigQuery Query Response. This class contains the results of a Query Job or of a
+ * Query Request.
+ */
```
ajkannan: Could add `@see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/query">Query</a>` here as well. The anchor `#response` also works on that URL, so you can add it if you think it'd be helpful.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/QueryRequestTest.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Test
+  public void testToBuilderIncomplete() {
+    QueryRequest queryRequest = QueryRequest.of(QUERY);
+    compareQueryRequest(queryRequest, queryRequest.toBuilder().build());
+  }
+
+  @Test
+  public void testBuilder() {
```
ajkannan: Nit: could add a test where you try to build a QueryRequest with a null query string and expect a NPE.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+import java.util.Objects;
+
+/**
+ * Google Cloud BigQuery Query Request. This class can be used to run a BigQuery SQL query and
+ * return results if the query completes within a specified timeout. The query results are saved to
+ * a temporary table that is deleted approximately 24 hours after the query is run. Query is run
+ * through a BigQuery Job whose identity can be accessed via {@link QueryResponse#job()}.
+ *
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/query">Query</a>
+ */
```
aozarov: Should we also include a link to the SQL syntax [reference](https://cloud.google.com/bigquery/query-reference)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+    private Boolean dryRun;
+    private Boolean useQueryCache;
+
+    private Builder() {}
+
+    /**
+     * Sets the BigQuery query to be executed.
+     */
+    public Builder query(String query) {
+      this.query = query;
```
aozarov: `checkNotNull`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+      return this;
+    }
+
+    public QueryRequest build() {
+      return new QueryRequest(this);
+    }
+  }
+
+  private QueryRequest(Builder builder) {
+    query = checkNotNull(builder.query);
```
aozarov: I think we should do the checks in the `Builder` so failure happens sooner than later (and  may miss some context).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
```diff
@@ -0,0 +1,272 @@
...
+      this.defaultDataset = defaultDataset;
+      return this;
+    }
+
+    /**
+     * Sets how long to wait for the query to complete, in milliseconds, before the request times
+     * out and returns. Note that this is only a timeout for the request, not the query. If the
+     * query takes longer to run than the timeout value, the call returns without any results and
+     * with the {@link QueryResponse#jobComplete()} set to {@code false}. You can call
+     * {@link BigQuery#getQueryResults(JobId, BigQuery.QueryResultsOption...)} to wait for the query
```
aozarov: Are BigQuery#getQueryResults and BigQuery.QueryResultsOption coming in the next PR?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+        && Objects.equals(totalBytesProcessed, response.totalBytesProcessed)
+        && Objects.equals(jobComplete, response.jobComplete)
+        && Objects.equals(errors, response.errors)
+        && Objects.equals(cacheHit, response.cacheHit);
+  }
+
+  static Builder builder() {
+    return new Builder();
+  }
+}
```
aozarov: Nit, add new line.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Google Cloud BigQuery Query Response. This class contains the results of a Query Job or of a
+ * Query Request.
+ */
+public class QueryResponse implements Serializable {
```
aozarov: I think it would be nice to add a code snippet to show how to use this class and check for jobComplete or get the results.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+  /**
+   * Returns the query result as a paginated list of rows, if the query completed successfully.
+   * Returns {@code null} otherwise.
+   */
+  public Page<List<FieldValue>> rows() {
+    return rows;
+  }
+
+  /**
+   * Returns the total number of bytes processed for the query.
```
aozarov: Any reason for not including "If this query was a dry run, this is the number of bytes that would be processed if the query were run."?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+   */
+  public Long totalBytesProcessed() {
+    return totalBytesProcessed;
+  }
+
+  /**
+   * Returns whether the job running the query has completed or not. If {@link #rows()} and
```
aozarov: I know that this is a copy/paste from [here](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query), but I think we can phrase it better. (use null instead of present) and make it clear this method should be used to check if results should be available.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+   */
+  public Boolean jobComplete() {
+    return jobComplete;
+  }
+
+  /**
+   * Returns errors and warnings encountered during the running of the job, if any. Errors here do
+   * not necessarily mean that the job has completed or was unsuccessful.
+   */
+  public List<BigQueryError> errors() {
```
aozarov: would `executionErrors` make it more consistent?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+    return errors;
+  }
+
+  /**
+   * Returns whether the query result was fetched from the query cache.
+   *
+   * @see <a href="https://cloud.google.com/bigquery/querying-data#querycaching">Query Caching</a>
+   */
+  public Boolean cacheHit() {
+    return cacheHit;
```
aozarov: can this ever be null?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+    return totalBytesProcessed;
+  }
+
+  /**
+   * Returns whether the job running the query has completed or not. If {@link #rows()} and
+   * {@link #totalRows()} are present, this method will always return {@code true}. If this method
+   * returns {@code false}, {@link #totalRows()} will not be available.
+   */
+  public Boolean jobComplete() {
+    return jobComplete;
```
aozarov: can this ever be `null`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,233 @@
...
+   */
+  public Page<List<FieldValue>> rows() {
+    return rows;
+  }
+
+  /**
+   * Returns the total number of bytes processed for the query.
+   */
+  public Long totalBytesProcessed() {
+    return totalBytesProcessed;
```
aozarov: can this ever be `null`?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/QueryResponseTest.java
```diff
@@ -0,0 +1,113 @@
...
+  private static final Long TOTAL_ROWS = 42L;
+  private static final PageImpl.NextPageFetcher<List<FieldValue>> FETCHER =
+      new PageImpl.NextPageFetcher<List<FieldValue>>() {
+        @Override
+        public Page<List<FieldValue>> nextPage() {
+          return null;
+        }
+      };
+  private static final Page<List<FieldValue>> ROWS =
+      new PageImpl<List<FieldValue>>(FETCHER, "cursor", ImmutableList.<List<FieldValue>>of());
```
aozarov: Nit, `new PageImpl<>`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
mziccard: Yes, this is {@code null} when the query has not completed yet.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
mziccard: Yes, this is {@code null} when the query has not completed yet.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
mziccard: In the tests I run this seems to always be set. I'll use `boolean` instead.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryRequest.java
mziccard: Yes but I better remove these links and add them later to avoid merging commits that break the build

mziccard:
Ehi guys, thanks for the feedback. I fixed all comments and answered the open questions.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/getQueryResults">Get Query
+ *     Results</a>
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/query">Query</a>
+ */
+public class QueryResponse implements Serializable {
+
+  private static final long serialVersionUID = 3549226764825005655L;
+
+  private final String etag;
```
aozarov: Do you know why etag is only returned by getQueryResult and not by `query` (even if completed)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,251 @@
...
+ *    }
+ *    List<BigQueryError> executionErrors = response.executionErrors();
+ *    Page<List<FieldValue>> rows = response.rows();
+ * }</pre>
+ *
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/getQueryResults">Get Query
+ *     Results</a>
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs/query">Query</a>
+ */
+public class QueryResponse implements Serializable {
```
aozarov: The public [docs](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query) for query suggests that one could also use `bigquery.tabledata.list(table_id)` to query the temporary table that is created for the associated request. Where do you get the tableId from?

mziccard: I am about to ask that to the BigQuery team. The table is not part of the query response. You can still identify the table from the UI/by listing all tables but this is worth asking.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,251 @@
...
+
+  private static final long serialVersionUID = 3549226764825005655L;
+
+  private final String etag;
+  private final Schema schema;
+  private final JobId job;
+  private final Long totalRows;
+  private final Page<List<FieldValue>> rows;
```
aozarov: As `schema`, `totalRows`, `totalBytesProcessed`, `cacheHit`, and `rows` are only available when `jobCompleted` did you consider moving them to something like a `QueryResult` class that would extends `PageImpl` and add these accessors (`rows` would be part of the standard page)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
mziccard: This is something else worth asking, I see no reason why this should be missing (even if not completed yet).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
mziccard: Do you think this would be more understandable/usable? I didn't want to mix non-pageable properties with pageable ones.

aozarov: I slightly leaning toward having a separate class that will have all values (so can use primitive) when result is available rather than having one class with all values but subset of them are not available based on a condition. 

We could create QueryResult that returns rows as a page or have that class extends Page. 

I don't feel strongly about it, so if you think current experience is better we can defer that decision.

mziccard: I'll give this a try now so we can see how it looks like.

mziccard:
@ajkannan I fixed the typo in 9756dff
@aozarov I added the `QueryResult` class in ea54699

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/SerializationTest.java
```diff
@@ -212,7 +233,7 @@ public void testModelAndRequests() throws Exception {
         TABLE_SCHEMA, TABLE_INFO, VIEW_INFO, EXTERNAL_TABLE_INFO, INLINE_FUNCTION, URI_FUNCTION,
         JOB_STATISTICS, EXTRACT_STATISTICS, LOAD_STATISTICS, QUERY_STATISTICS, BIGQUERY_ERROR,
         JOB_STATUS, JOB_ID, COPY_JOB, EXTRACT_JOB, LOAD_JOB, QUERY_JOB, INSERT_ALL_REQUEST,
-        INSERT_ALL_RESPONSE, FIELD_VALUE};
+        INSERT_ALL_RESPONSE, FIELD_VALUE, QUERY_REQUEST, QUERY_RESPONSE};
```
ajkannan: `QUERY_RESULT` is created above but not included in this list.

ajkannan: Actually not necessary, ignore that comment

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,183 @@
...
+  private final String etag;
+  private final JobId job;
+  private final boolean jobComplete;
+  private final List<BigQueryError> executionErrors;
+
+  static final class Builder {
+
+    private QueryResult result;
+    private String etag;
+    private JobId job;
```
aozarov: Nit, why not call it `jobId`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,183 @@
...
+ * Query Request.
+ *
+ * <p>Example usage of a query response:
+ * <pre>    {@code
+ *    QueryResponse response = bigquery.query(request);
+ *    while (!response.jobComplete()) {
+ *      response = bigquery.getQueryResults(response.job());
+ *      Thread.sleep(1000);
+ *    }
+ *    List<BigQueryError> executionErrors = response.executionErrors();
```
aozarov: Nit, we should either use `executionErrors` or add a comment suggesting a possible check.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,183 @@
...
+ */
+public class QueryResponse implements Serializable {
+
+  private static final long serialVersionUID = 3549226764825005655L;
+
+  private final QueryResult result;
+  private final String etag;
+  private final JobId job;
+  private final boolean jobComplete;
+  private final List<BigQueryError> executionErrors;
```
aozarov: I am curious. is there an easy way for a user to tell that query completely failed (e.g. bad syntax)?
Does such validation happen before job starts (and sending the query request will fail)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
```diff
@@ -0,0 +1,183 @@
...
+  public boolean jobComplete() {
+    return jobComplete;
+  }
+
+  /**
+   * Returns errors and warnings encountered during the running of the job, if any. Errors here do
+   * not necessarily mean that the job has completed or was unsuccessful.
+   */
+  public List<BigQueryError> executionErrors() {
+    return executionErrors;
```
aozarov: can this ever be `null` or does `isEmpty` indicate no errors? My preference would be the latter (and possible add `hasErrors`).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResult.java
```diff
@@ -0,0 +1,176 @@
...
+   * Returns whether the query result was fetched from the query cache.
+   *
+   * @see <a href="https://cloud.google.com/bigquery/querying-data#querycaching">Query Caching</a>
+   */
+  public boolean cacheHit() {
+    return cacheHit;
+  }
+
+  /**
+   * Returns the schema of the results.
```
aozarov: I assume this could still be `null`? "Present only when the query completes **successfully**."

aozarov:
Looks great! I have few more questions/nits.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
mziccard: Query is validated before starting the job. The query request fails with error 400.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResult.java
mziccard: Yes

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryResponse.java
mziccard: Yes if there are no errors this is null, I can update the constructor to set this to an empty list.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/QueryResultTest.java
```diff
@@ -0,0 +1,91 @@
...
+      .pageFetcher(FETCHER)
+      .results(ImmutableList.<List<FieldValue>>of())
+      .cacheHit(CACHE_HIT)
+      .build();
+  private static final QueryResult QUERY_RESULT_INCOMPLETE = QueryResult.builder()
+      .totalBytesProcessed(TOTAL_BYTES_PROCESSED)
+      .build();
+
+  @Test
+  public void testBuilder() {
```
ajkannan: Should there be an `assertEquals` for that the `pageFetcher`?

mziccard: I don't have access to it, this is why I am testing nextPage and nextPageCursor instead.

mziccard:
c6e53c8 fixes latest comments and makes checkstyle happy

PR #454, opened 2015-12-15
Add BigQuery service implementation

mziccard:
This class implements the `BigQuery` service, main changes are:
- Connect `DefaultBigQueryRpc` and `BigQueryImpl` to `BigQueryOptions`
- Add `defaultInstance` method to `BigQueryOptions`
- Add `Option` base class for `BigQuery` operation options and test
- Add methods and options to the `BigQuery` interface
- Add `BigQueryImpl` service implementation
- Add `BigQueryImplTest` with unit tests

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
@@ -16,7 +16,10 @@
 
 package com.google.gcloud.bigquery;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import com.google.common.base.Function;
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
@@ -24,8 +27,8 @@ import com.google.gcloud.Page;
 import com.google.gcloud.Service;
 import com.google.gcloud.spi.BigQueryRpc;
 
-import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * An interface for Google Cloud BigQuery.
@@ -34,11 +37,17 @@ import java.util.List;
  */
 public interface BigQuery extends Service<BigQueryOptions> {
 
+  /**
+   * Fields of a BigQuery Dataset resource.
+   *
+   * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/datasets#resource">Dataset
+   *     Resource</a>
+   */
   enum DatasetField {
     ACCESS("access"),
     CREATION_TIME("creationTime"),
     DATASET_REFERENCE("datasetReference"),
-    DEFAULT_TABLE_EXPIRATION_MS("defaultTableLifetime"),
+    DEFAULT_TABLE_EXPIRATION_MS("defaultTableExpirationMsS"),
     DESCRIPTION("description"),
     ETAG("etag"),
     FRIENDLY_NAME("friendlyName"),
@@ -58,15 +67,21 @@ public interface BigQuery extends Service<BigQueryOptions> {
     }
 
     static String selector(DatasetField... fields) {
-      HashSet<String> fieldStrings = Sets.newHashSetWithExpectedSize(fields.length + 1);
+      Set<String> fieldStrings = Sets.newHashSetWithExpectedSize(fields.length + 1);
       fieldStrings.add(DATASET_REFERENCE.selector());
       for (DatasetField field : fields) {
         fieldStrings.add(field.selector());
       }
-      return com.google.common.base.Joiner.on(',').join(fieldStrings);
+      return Joiner.on(',').join(fieldStrings);
     }
   }
 
+  /**
+   * Fields of a BigQuery Table resource.
+   *
+   * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/tables#resource">Table
+   *     Resource</a>
+   */
   enum TableField {
     CREATION_TIME("creationTime"),
     DESCRIPTION("description"),
@@ -97,16 +112,22 @@ public interface BigQuery extends Service<BigQueryOptions> {
     }
 
     static String selector(TableField... fields) {
-      HashSet<String> fieldStrings = Sets.newHashSetWithExpectedSize(fields.length + 2);
+      Set<String> fieldStrings = Sets.newHashSetWithExpectedSize(fields.length + 2);
       fieldStrings.add(TABLE_REFERENCE.selector());
       fieldStrings.add(TYPE.selector());
       for (TableField field : fields) {
         fieldStrings.add(field.selector());
       }
-      return com.google.common.base.Joiner.on(',').join(fieldStrings);
+      return Joiner.on(',').join(fieldStrings);
     }
   }
 
+  /**
+   * Fields of a BigQuery Job resource.
+   *
+   * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs#resource">Job Resource
+   *     </a>
+   */
   enum JobField {
     CONFIGURATION("configuration"),
     ETAG("etag"),
@@ -128,13 +149,13 @@ public interface BigQuery extends Service<BigQueryOptions> {
     }
 
     static String selector(JobField... fields) {
-      HashSet<String> fieldStrings = Sets.newHashSetWithExpectedSize(fields.length + 2);
+      Set<String> fieldStrings = Sets.newHashSetWithExpectedSize(fields.length + 2);
       fieldStrings.add(JOB_REFERENCE.selector());
       fieldStrings.add(CONFIGURATION.selector());
       for (JobField field : fields) {
         fieldStrings.add(field.selector());
       }
-      return com.google.common.base.Joiner.on(',').join(fieldStrings);
+      return Joiner.on(',').join(fieldStrings);
     }
   }
 
@@ -184,7 +205,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
 
     /**
      * Returns an option to specify the dataset's fields to be returned by the RPC call. If this
-     * option is not provided all dataset's fields are returned. {@code DatasetOption.fields}) can
+     * option is not provided all dataset's fields are returned. {@code DatasetOption.fields} can
      * be used to specify only the fields of interest. {@link DatasetInfo#datasetId()} is always
      * returned, even if not specified.
      */
@@ -228,6 +249,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
      * Returns an option to specify the maximum number of tables to be returned.
      */
     public static TableListOption maxResults(long maxResults) {
+      checkArgument(maxResults >= 0);
       return new TableListOption(BigQueryRpc.Option.MAX_RESULTS, maxResults);
     }
 
@@ -252,7 +274,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
 
     /**
      * Returns an option to specify the table's fields to be returned by the RPC call. If this
-     * option is not provided all table's fields are returned. {@code TableOption.fields}) can be
+     * option is not provided all table's fields are returned. {@code TableOption.fields} can be
      * used to specify only the fields of interest. {@link BaseTableInfo#tableId()} and
      * {@link BaseTableInfo#type()} are always returned, even if not specified.
      */
@@ -276,6 +298,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
      * Returns an option to specify the maximum number of rows to be returned.
      */
     public static TableDataListOption maxResults(long maxResults) {
+      checkArgument(maxResults >= 0);
       return new TableDataListOption(BigQueryRpc.Option.MAX_RESULTS, maxResults);
     }
 
@@ -291,6 +314,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
      * data.
      */
     public static TableDataListOption startIndex(long index) {
+      checkArgument(index >= 0);
       return new TableDataListOption(BigQueryRpc.Option.START_INDEX, index);
     }
   }
@@ -314,14 +338,14 @@ public interface BigQuery extends Service<BigQueryOptions> {
     }
 
     /**
-     * Returns an option to list only jobs that match the provided filters.
+     * Returns an option to list only jobs that match the provided state filters.
      */
     public static JobListOption stateFilter(JobStatus.State... stateFilters) {
       List<String> stringFilters = Lists.transform(ImmutableList.copyOf(stateFilters),
           new Function<JobStatus.State, String>() {
             @Override
             public String apply(JobStatus.State state) {
-              return state.toString().toLowerCase();
+              return state.name().toLowerCase();
             }
           });
       return new JobListOption(BigQueryRpc.Option.STATE_FILTER, stringFilters);
@@ -331,6 +355,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
      * Returns an option to specify the maximum number of jobs to be returned.
      */
     public static JobListOption maxResults(long maxResults) {
+      checkArgument(maxResults >= 0);
       return new JobListOption(BigQueryRpc.Option.MAX_RESULTS, maxResults);
     }
 
@@ -343,7 +368,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
 
     /**
      * Returns an option to specify the job's fields to be returned by the RPC call. If this option
-     * is not provided all job's fields are returned. {@code JobOption.fields()}) can be used to
+     * is not provided all job's fields are returned. {@code JobOption.fields()} can be used to
      * specify only the fields of interest. {@link JobInfo#jobId()}, {@link JobStatus#state()},
      * {@link JobStatus#error()} as well as type-specific configuration (e.g.
      * {@link QueryJobInfo#query()} for Query Jobs) are always returned, even if not specified.
@@ -370,7 +395,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
 
     /**
      * Returns an option to specify the job's fields to be returned by the RPC call. If this option
-     * is not provided all job's fields are returned. {@code JobOption.fields}) can be used to
+     * is not provided all job's fields are returned. {@code JobOption.fields()} can be used to
      * specify only the fields of interest. {@link JobInfo#jobId()} as well as type-specific
      * configuration (e.g. {@link QueryJobInfo#query()} for Query Jobs) are always returned, even if
      * not specified.
@@ -395,21 +420,23 @@ public interface BigQuery extends Service<BigQueryOptions> {
      * Returns an option to specify the maximum number of rows to be returned.
      */
     public static QueryResultsOption maxResults(long maxResults) {
+      checkArgument(maxResults >= 0);
       return new QueryResultsOption(BigQueryRpc.Option.MAX_RESULTS, maxResults);
     }
 
     /**
-     * Returns an option to specify the page token from which to start listing query results.
+     * Returns an option to specify the page token from which to start getting query results.
      */
     public static QueryResultsOption startPageToken(String pageToken) {
       return new QueryResultsOption(BigQueryRpc.Option.PAGE_TOKEN, pageToken);
     }
 
     /**
-     * Returns an option that sets the zero-based index of the row from which to start listing query
+     * Returns an option that sets the zero-based index of the row from which to start getting query
      * results.
      */
-    public static QueryResultsOption startIndex(Long startIndex) {
+    public static QueryResultsOption startIndex(long startIndex) {
+      checkArgument(startIndex >= 0);
       return new QueryResultsOption(BigQueryRpc.Option.START_INDEX, startIndex);
     }
 
@@ -418,7 +445,8 @@ public interface BigQuery extends Service<BigQueryOptions> {
      * before returning. Default is 10 seconds. If the timeout passes before the job completes,
      * {@link QueryResponse#jobComplete()} will be {@code false}.
      */
-    public static QueryResultsOption maxWaitTime(Long maxWaitTime) {
+    public static QueryResultsOption maxWaitTime(long maxWaitTime) {
+      checkArgument(maxWaitTime >= 0);
       return new QueryResultsOption(BigQueryRpc.Option.TIMEOUT, maxWaitTime);
     }
   }
@@ -460,7 +488,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
 
   /**
    * Lists the project's datasets. This method returns partial information on each dataset
-   * ({@link DatasetInfo#datasetId()} ()}, {@link DatasetInfo#friendlyName()} and
+   * ({@link DatasetInfo#datasetId()}, {@link DatasetInfo#friendlyName()} and
    * {@link DatasetInfo#id()}). To get complete information use either
    * {@link #getDataset(String, DatasetOption...)} or
    * {@link #getDataset(DatasetId, DatasetOption...)}.
@@ -592,15 +620,16 @@ public interface BigQuery extends Service<BigQueryOptions> {
   JobInfo getJob(JobId jobId, JobOption... options) throws BigQueryException;
 
   /**
-   * Lists the dataset's tables.
+   * Lists the jobs.
    *
    * @throws BigQueryException upon failure
    */
   Page<JobInfo> listJobs(JobListOption... options) throws BigQueryException;
 
   /**
-   * Sends a job cancel request. This call will return immediately, and the client will need to poll
-   * for the job status to see if the cancel completed successfully.
+   * Sends a job cancel request. This call will return immediately. The job status can then be
+   * checked using either {@link #getJob(JobId, JobOption...)} or
+   * {@link #getJob(String, JobOption...)}).
    *
    * @return {@code true} if cancel was requested successfully, {@code false} if the job was not
    *     found
@@ -609,9 +638,9 @@ public interface BigQuery extends Service<BigQueryOptions> {
   boolean cancel(String jobId) throws BigQueryException;
 
   /**
-   * Sends a job cancel request. This call will return immediately. The client will need to poll
-   * for the job status using either {@link #getJob(JobId, JobOption...)} or
-   * {@link #getJob(String, JobOption...)}) to see if the cancel operation completed successfully.
+   * Sends a job cancel request. This call will return immediately. The job status can then be
+   * checked using either {@link #getJob(JobId, JobOption...)} or
+   * {@link #getJob(String, JobOption...)}).
    *
    * @return {@code true} if cancel was requested successfully, {@code false} if the job was not
    *     found
@@ -620,14 +649,14 @@ public interface BigQuery extends Service<BigQueryOptions> {
   boolean cancel(JobId tableId) throws BigQueryException;
 
   /**
-   * Runs the query associated to the request.
+   * Runs the query associated with the request.
    *
    * @throws BigQueryException upon failure
    */
   QueryResponse query(QueryRequest request) throws BigQueryException;
 
   /**
-   * Returns results of the query associated to the provided job.
+   * Returns results of the query associated with the provided job.
    *
    * @throws BigQueryException upon failure
    */
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
@@ -25,7 +25,6 @@ import com.google.api.services.bigquery.model.Job;
 import com.google.api.services.bigquery.model.Table;
 import com.google.api.services.bigquery.model.TableDataInsertAllRequest;
 import com.google.api.services.bigquery.model.TableDataInsertAllRequest.Rows;
-import com.google.api.services.bigquery.model.TableDataInsertAllResponse;
 import com.google.api.services.bigquery.model.TableReference;
 import com.google.api.services.bigquery.model.TableRow;
 import com.google.common.base.Function;
@@ -300,7 +299,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
   }
 
   @Override
-  public boolean delete(final String datasetId, final String tableId) throws BigQueryException {
+  public boolean delete(String datasetId, String tableId) throws BigQueryException {
     return delete(TableId.of(datasetId, tableId));
   }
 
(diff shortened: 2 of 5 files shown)


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
  */
 public interface BigQuery extends Service<BigQueryOptions> {
 
-  // TODO(mziccard) add missing methods
+  enum DatasetField {
+    ACCESS("access"),
+    CREATION_TIME("creationTime"),
+    DATASET_REFERENCE("datasetReference"),
+    DEFAULT_TABLE_EXPIRATION_MS("defaultTableLifetime"),
```
aozarov: defaultTableExpirationMs ?

mziccard: Where did I get that from?:D Will fix!

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+    DatasetField(String selector) {
+      this.selector = selector;
+    }
+
+    public String selector() {
+      return selector;
+    }
+
+    static String selector(DatasetField... fields) {
+      HashSet<String> fieldStrings = Sets.newHashSetWithExpectedSize(fields.length + 1);
```
aozarov: Just thinking out loud. If enum fields implemented a "Field" interface with a `String selector()` method then you can have only one static selector method `static String selector(Selector first, Selector... other)` for all...

mziccard: Yes and we should probably reuse that same interface in `Storage` and other services that use field selection. Do you mind if we make this part of a separate PR once bigquery is merged? That would make easier to put the interface in `gcloud-java-core` and modify also `gcloud-java-storage` (no conflicts on merge)

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+      fieldStrings.add(DATASET_REFERENCE.selector());
+      for (DatasetField field : fields) {
+        fieldStrings.add(field.selector());
+      }
+      return com.google.common.base.Joiner.on(',').join(fieldStrings);
```
aozarov: Any reason for this fully qualified reference (and avoiding the import)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
aozarov: Nit, `Set<String> fieldStrings`

aozarov: Separate PR sounds fine to me.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
mziccard: No actual reason

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
  */
 public interface BigQuery extends Service<BigQueryOptions> {
 
-  // TODO(mziccard) add missing methods
+  enum DatasetField {
```
ajkannan: Should we have javadoc for the enums `DatasetField`, `TableField`, etc?

mziccard: We don't have it for storage (but this is no good reason for not adding it). I think that the names are quite self-explanatory. I also think that adding javadoc for each enum value would mean adding too much "obvious" comments. I'm +1 to adding a javadoc comment on top of the enum type with a link to a dataset/table/job resource definition.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    private static final long serialVersionUID = -7166083569900951337L;
+
+    private DatasetDeleteOption(BigQueryRpc.Option option, Object value) {
+      super(option, value);
+    }
+
+    /**
+     * Returns an option to delete a dataset even if non-empty. If not provided, attempting to
+     * delete a non-empty dataset will result in a {@link BigQueryException} being thrown.
```
aozarov: Is there a way to tell from the exception that the delete failed for that reason?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
aozarov: I wasn't bothered by it because I also think that name level comments are redundant. I didn't mind much about the class level as their usage is always part of a specific `Option` and they have a nice javadoc for the `fields` method however it would be nice to have a class level javadoc for consistency and for lulling checkstyle and a like about it.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+   * {@link #getJob(String, JobOption...)}) to see if the cancel operation completed successfully.
+   *
+   * @return {@code true} if cancel was requested successfully, {@code false} if the job was not
+   *     found
+   * @throws BigQueryException upon failure
+   */
+  boolean cancel(JobId tableId) throws BigQueryException;
+
+  /**
+   * Runs the query associated to the request.
```
ajkannan: Nit: replace "to" with "with"

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+   */
+  boolean delete(DatasetId datasetId, DatasetDeleteOption... options) throws BigQueryException;
+
+  /**
+   * Deletes the requested table.
+   *
+   * @return {@code true} if table was deleted, {@code false} if it was not found.
+   * @throws BigQueryException upon failure
+   */
+  boolean delete(String datasetId, String tableId) throws BigQueryException;
```
aozarov: There is no support for conditional delete based on etag?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+  /**
+   * Lists the dataset's tables.
+   *
+   * @throws BigQueryException upon failure
+   */
+  Page<JobInfo> listJobs(JobListOption... options) throws BigQueryException;
+
+  /**
+   * Sends a job cancel request. This call will return immediately, and the client will need to poll
+   * for the job status to see if the cancel completed successfully.
```
ajkannan: Nit: we could add the links like in the comment below

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+    public static TableDataListOption startPageToken(String pageToken) {
+      return new TableDataListOption(BigQueryRpc.Option.PAGE_TOKEN, pageToken);
+    }
+
+    /**
+     * Returns an option that sets the zero-based index of the row from which to start listing table
+     * data.
+     */
+    public static TableDataListOption startIndex(long index) {
+      return new TableDataListOption(BigQueryRpc.Option.START_INDEX, index);
```
aozarov: Should we `checkArgument(index >= 0)`? If so, we could also apply it for `maxResults > 0` and other similar places.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    /**
+     * Returns an option to list only jobs that match the provided filters.
+     */
+    public static JobListOption stateFilter(JobStatus.State... stateFilters) {
+      List<String> stringFilters = Lists.transform(ImmutableList.copyOf(stateFilters),
+          new Function<JobStatus.State, String>() {
+            @Override
+            public String apply(JobStatus.State state) {
+              return state.toString().toLowerCase();
```
aozarov: nit, `name()` is slightly better than `toString()` as the latter could be overridden.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    /**
+     * Returns an option to specify the job's fields to be returned by the RPC call. If this option
+     * is not provided all job's fields are returned. {@code JobOption.fields()}) can be used to
+     * specify only the fields of interest. {@link JobInfo#jobId()}, {@link JobStatus#state()},
+     * {@link JobStatus#error()} as well as type-specific configuration (e.g.
+     * {@link QueryJobInfo#query()} for Query Jobs) are always returned, even if not specified.
+     * {@link JobField#SELF_LINK} and {@link JobField#ETAG} can not be selected when listing jobs.
+     */
+    public static JobListOption fields(JobField... fields) {
```
aozarov: Do you think this is useful (considering the possible return values and the minimum that we specify)? I am not sure about it, and not having it would be consistent with the other listing operations that do not provide a field selector.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    /**
+     * Returns an option to specify the maximum number of rows to be returned.
+     */
+    public static QueryResultsOption maxResults(long maxResults) {
+      return new QueryResultsOption(BigQueryRpc.Option.MAX_RESULTS, maxResults);
+    }
+
+    /**
+     * Returns an option to specify the page token from which to start listing query results.
```
aozarov: is "listing" the right term here?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+     */
+    public static QueryResultsOption startPageToken(String pageToken) {
+      return new QueryResultsOption(BigQueryRpc.Option.PAGE_TOKEN, pageToken);
+    }
+
+    /**
+     * Returns an option that sets the zero-based index of the row from which to start listing query
+     * results.
+     */
+    public static QueryResultsOption startIndex(Long startIndex) {
```
aozarov: Why using `Long` (here and below).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    private static final long serialVersionUID = 1674133909259913250L;
+
+    private DatasetOption(BigQueryRpc.Option option, Object value) {
+      super(option, value);
+    }
+
+    /**
+     * Returns an option to specify the dataset's fields to be returned by the RPC call. If this
+     * option is not provided all dataset's fields are returned. {@code DatasetOption.fields}) can
```
ajkannan: Nit: remove ')' after `fields}`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    private static final long serialVersionUID = -1723870134095936772L;
+
+    private TableOption(BigQueryRpc.Option option, Object value) {
+      super(option, value);
+    }
+
+    /**
+     * Returns an option to specify the table's fields to be returned by the RPC call. If this
+     * option is not provided all table's fields are returned. {@code TableOption.fields}) can be
```
ajkannan: Nit: there's an extra ')' after '.fields}'

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    /**
+     * Returns an option to list all jobs, even the ones issued by other users.
+     */
+    public static JobListOption allUsers() {
+      return new JobListOption(BigQueryRpc.Option.ALL_USERS, true);
+    }
+
+    /**
+     * Returns an option to list only jobs that match the provided filters.
```
ajkannan: Should we specify that the filter is on the job's state?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+    /**
+     * Returns an option to specify the page token from which to start listing jobs.
+     */
+    public static JobListOption startPageToken(String pageToken) {
+      return new JobListOption(BigQueryRpc.Option.PAGE_TOKEN, pageToken);
+    }
+
+    /**
+     * Returns an option to specify the job's fields to be returned by the RPC call. If this option
+     * is not provided all job's fields are returned. {@code JobOption.fields()}) can be used to
```
ajkannan: extra `)` after @code block

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+    private static final long serialVersionUID = -3111736712316353665L;
+
+    private JobOption(BigQueryRpc.Option option, Object value) {
+      super(option, value);
+    }
+
+    /**
+     * Returns an option to specify the job's fields to be returned by the RPC call. If this option
+     * is not provided all job's fields are returned. {@code JobOption.fields}) can be used to
```
ajkannan: extra ')'

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+  /**
+   * Returns the requested dataset or {@code null} if not found.
+   *
+   * @throws BigQueryException upon failure
+   */
+  DatasetInfo getDataset(DatasetId datasetId, DatasetOption... options) throws BigQueryException;
+
+  /**
+   * Lists the project's datasets. This method returns partial information on each dataset
+   * ({@link DatasetInfo#datasetId()} ()}, {@link DatasetInfo#friendlyName()} and
```
ajkannan: Extra `()}` after `datasetId`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Option.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.common.base.MoreObjects;
+import com.google.gcloud.spi.BigQueryRpc;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Base class for BigQuery operation option.
+ */
+class Option implements Serializable {
```
ajkannan: Should we add `Option`'s subclasses to serialization tests?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/OptionTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Test
+  public void testOption() {
+    Option option = new Option(BigQueryRpc.Option.PAGE_TOKEN, "token");
+    assertEquals(BigQueryRpc.Option.PAGE_TOKEN, option.rpcOption());
+    assertEquals("token", option.value());
+  }
+
+  @Test(expected=NullPointerException.class)
+  public void testIndexOutOfBoundsException() {
```
ajkannan: testIndexOutOfBoundsException -> testNullOption?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+  private BigQueryOptions options;
+  private BigQueryRpcFactory rpcFactoryMock;
+  private BigQueryRpc bigqueryRpcMock;
+  private BigQuery bigquery;
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @Before
+  public void setUp() throws IOException, InterruptedException {
```
ajkannan: Nit: we can probably get rid of `throws ...`

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+      .build();
+
+  // Empty BigQueryRpc options
+  private static final Map<BigQueryRpc.Option, ?> EMPTY_RPC_OPTIONS = ImmutableMap.of();
+
+  // Dataset options
+  private static final BigQuery.DatasetOption DATASET_OPTION_FIELDS =
+      BigQuery.DatasetOption.fields(BigQuery.DatasetField.ACCESS, BigQuery.DatasetField.ETAG);
+
+  // Dataset list options
```
ajkannan: We could add the "max results" option here and set it to something low for testing purposes

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
```diff
@@ -0,0 +1,723 @@
...
+          return bigQueryRpc.deleteDataset(datasetId.dataset(), optionsMap);
+        }
+      }, options().retryParams(), EXCEPTION_HANDLER);
+    } catch (RetryHelper.RetryHelperException e) {
+      throw BigQueryException.translateAndThrow(e);
+    }
+  }
+
+  @Override
+  public boolean delete(final String datasetId, final String tableId) throws BigQueryException {
```
ajkannan: Nit: I believe the `final` is unnecessary here

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+  private static final BigQuery.DatasetDeleteOption DATASET_DELETE_CONTENTS =
+      BigQuery.DatasetDeleteOption.deleteContents();
+  private static final Map<BigQueryRpc.Option, ?> DATASET_DELETE_OPTIONS = ImmutableMap.of(
+      BigQueryRpc.Option.DELETE_CONTENTS, true);
+
+  // Table options
+  private static final BigQuery.TableOption TABLE_OPTION_FIELDS =
+      BigQuery.TableOption.fields(BigQuery.TableField.SCHEMA, BigQuery.TableField.ETAG);
+
+  // Table list options
```
ajkannan: Could add the "startIndex" option here.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+      BigQuery.TableDataListOption.startPageToken("cursor");
+  private static final Map<BigQueryRpc.Option, ?> TABLE_DATA_LIST_OPTIONS = ImmutableMap.of(
+      BigQueryRpc.Option.MAX_RESULTS, 42L,
+      BigQueryRpc.Option.PAGE_TOKEN, "cursor");
+
+  // Job options
+  private static final BigQuery.JobOption JOB_OPTION_FIELDS =
+      BigQuery.JobOption.fields(BigQuery.JobField.USER_EMAIL);
+
+  // Job list options
```
ajkannan: Could test the "max results" and "start page token" options here too

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+      BigQuery.JobListOption.fields(BigQuery.JobField.STATISTICS);
+  private static final BigQuery.JobListOption JOB_LIST_ALL_USERS =
+      BigQuery.JobListOption.allUsers();
+  private static final BigQuery.JobListOption JOB_LIST_STATE_FILTER =
+      BigQuery.JobListOption.stateFilter(JobStatus.State.DONE, JobStatus.State.PENDING);
+  private static final Map<BigQueryRpc.Option, ?> JOB_LIST_OPTIONS = ImmutableMap.of(
+      BigQueryRpc.Option.ALL_USERS, true,
+      BigQueryRpc.Option.STATE_FILTER, ImmutableList.of("done", "pending"));
+
+  // Query Results options
```
ajkannan: If you think it's useful, we could add the max results and start page token options here

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+    String selector = (String) capturedOptions.getValue().get(TABLE_OPTION_FIELDS.rpcOption());
+    assertTrue(selector.contains("tableReference"));
+    assertTrue(selector.contains("schema"));
+    assertTrue(selector.contains("etag"));
+    assertEquals(31, selector.length());
+    assertEquals(updatedTableInfoWithProject, table);
+  }
+
+  @Test
+  public void testInsertAll() {
```
ajkannan: Do you think it'd be worthwhile to test that `skipInvalidRows` and `ignoreUnknownValues` are being handled properly in this method?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+    assertTrue(bigquery.cancel(JobId.of(PROJECT, JOB)));
+  }
+
+  @Test
+  public void testQueryRequest() {
+    JobId queryJob = JobId.of(PROJECT, JOB);
+    com.google.api.services.bigquery.model.QueryResponse responsePb =
+        new com.google.api.services.bigquery.model.QueryResponse()
+            .setJobReference(queryJob.toPb())
+            .setJobComplete(false);
```
ajkannan: Do you think we should also test the "true" case to check that the response is being setup correctly in that case?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -0,0 +1,982 @@
...
+            .setJobComplete(true)
+            .setCacheHit(false)
+            .setPageToken("cursor")
+            .setTotalBytesProcessed(42L)
+            .setTotalRows(BigInteger.valueOf(1L));
+    EasyMock.expect(bigqueryRpcMock.getQueryResults(JOB, EMPTY_RPC_OPTIONS)).andReturn(responsePb);
+    EasyMock.replay(bigqueryRpcMock);
+    bigquery = options.service();
+    QueryResponse response = bigquery.getQueryResults(queryJob);
+    assertEquals(queryJob, response.jobId());
```
ajkannan: Should we assert that `etag` is as expected as well?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
mziccard: I guess you mean to add id to table data options.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+  /**
+   * Returns the requested job or {@code null} if not found.
+   *
+   * @throws BigQueryException upon failure
+   */
+  JobInfo getJob(JobId jobId, JobOption... options) throws BigQueryException;
+
+  /**
+   * Lists the dataset's tables.
```
aozarov: Lists the jobs

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+   * @return {@code true} if cancel was requested successfully, {@code false} if the job was not
+   *     found
+   * @throws BigQueryException upon failure
+   */
+  boolean cancel(String jobId) throws BigQueryException;
+
+  /**
+   * Sends a job cancel request. This call will return immediately. The client will need to poll
+   * for the job status using either {@link #getJob(JobId, JobOption...)} or
+   * {@link #getJob(String, JobOption...)}) to see if the cancel operation completed successfully.
```
aozarov: what does it mean "completed **successfully**" for cancel?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -25,5 +34,602 @@
...
+
+  /**
+   * Runs the query associated to the request.
+   *
+   * @throws BigQueryException upon failure
+   */
+  QueryResponse query(QueryRequest request) throws BigQueryException;
+
+  /**
+   * Returns results of the query associated to the provided job.
```
aozarov: s/ to / with / sounds better to me.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
```diff
@@ -0,0 +1,723 @@
...
+import com.google.gcloud.bigquery.InsertAllRequest.RowToInsert;
+import com.google.gcloud.spi.BigQueryRpc;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuery {
+
+  private static final Interceptor EXCEPTION_HANDLER_INTERCEPTOR = new Interceptor() {
```
aozarov: After your #438 we could probably handle this in a common way...

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
```diff
@@ -0,0 +1,723 @@
...
+    });
+    requestPb.setRows(rowsPb);
+    try {
+      return InsertAllResponse.fromPb(runWithRetries(new Callable<TableDataInsertAllResponse>() {
+        @Override
+        public TableDataInsertAllResponse call() {
+          return bigQueryRpc.insertAll(tableId.dataset(), tableId.table(), requestPb);
+        }
+      }, options().retryParams(), EXCEPTION_HANDLER));
+    } catch (RetryHelper.RetryHelperException e) {
```
aozarov: One problem with retries here is that we can get duplicates if `insertId` is missing.

Some options:
1) document that when retries are configured make sure you provide id to avoid duplicates.

2) We should add a `UUID.randomUUID().toString()` when `insertId` is missing (before the `runWithRetries`) call.

3) Don't use retry handler here.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
```diff
@@ -0,0 +1,723 @@
...
+            new Function<TableRow, List<FieldValue>>() {
+              @Override
+              public List<FieldValue> apply(TableRow rowPb) {
+                return Lists.transform(rowPb.getF(), FieldValue.FROM_PB_FUNCTION);
+              }
+            }));
+  }
+
+  @Override
+  public JobInfo getJob(final String jobId, JobOption... options) throws BigQueryException {
```
aozarov: redundant `final`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
mziccard: I think it is useful for instance do avoid getting `statistics` if we do not need them. It becomes even more useful now that [queryPlan](https://cloud.google.com/bigquery/docs/reference/v2/jobs#statistics.query.queryPlan) is included in query statistics.

We do not provide a selector for list operation for the moment, but as soon as lists return complete resources I think we should add it.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
mziccard: We better go with 3) now, open an issue and consider 2)

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
```diff
@@ -0,0 +1,723 @@
...
+      return builder.build();
+    } catch (RetryHelper.RetryHelperException e) {
+      throw BigQueryException.translateAndThrow(e);
+    }
+  }
+
+  @Override
+  public QueryResponse getQueryResults(JobId job, QueryResultsOption... options)
+      throws BigQueryException {
+    final Map<BigQueryRpc.Option, ?> optionsMap = optionMap(options);
```
aozarov: no need for final.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
aozarov: unanswered.

mziccard:
@aozarov I removed the remaining HashSet in place.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
aozarov: Sounds good. did you create the issue?

aozarov:
Looks good to me. @ajkannan do you have any more comments?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
mziccard: `code` is `400` and `message` is `Dataset gcloud-devel:dataset is still in use`. Once we tie `BigQueryError` to `BigQueryException` the exception will also have the `reason` information, which is `resourceInUse`.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
mziccard: Just opened #470

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
aozarov: Ok, it would be nice to include that in the Javadoc (though can wait until you tie BigQueryError to the exception).

ajkannan:
Looks good to me as well

PR #457, opened 2015-12-15
ResourceManagerImpl + docs

### LINKED ISSUE #360: Add example to Resource Manager's README
Also address the other TODOs mentioned in the README.  This should be done when the package is complete and ready for use.

ajkannan:
Also fixes #360

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,6 +1,10 @@
 ## gcloud-java tools for testing
 
-This library provides tools to help write tests for code that uses gcloud-java services.
+This library provides tools to help write tests for code that uses the following gcloud-java services:
+
+-  [Datastore] (#testing-code-that-uses-datastore)
+-  [Storage] (#testing-code-that-uses-storage)
+-  [Resource Manager] (#testing-code-that-uses-resource-manager)
 
 ### Testing code that uses Datastore
 
@@ -85,8 +89,7 @@ You can test against a temporary local Resource Manager by following these steps
 2. In your program, create and use a Resource Manager service object whose host is set to `localhost` at the appropriate port.  For example:
 
   ```java
-  ResourceManager resourceManager = ResourceManagerOptions.builder()
-      .host("http://localhost:" + helper.port()).build().service();
+  ResourceManager resourceManager = LocalResourceManagerHelper.options().service();
   ```
 
 3. Run your tests.
@@ -98,18 +101,6 @@ You can test against a temporary local Resource Manager by following these steps
   ```
 
   This method will block until the server thread has been terminated.
-
-#### On a remote machine
-
-You can test against a remote Resource Manager emulator as well.  To do this, set the host to the hostname of the remote machine, like the example below.
-
-  ```java
-  ResourceManager resourceManager = ResourceManagerOptions.builder()
-      .host("http://<hostname of machine>:<port>").build();
-  ```
-
-Note that the remote Resource Manager emulator must be running before your tests are run.
-
 
 
 [cloud-platform-storage-authentication]:https://cloud.google.com/storage/docs/authentication?hl=en#service_accounts
--- a/gcloud-java-resourcemanager/README.md
+++ b/gcloud-java-resourcemanager/README.md
@@ -17,8 +17,22 @@ Java idiomatic client for [Google Cloud Resource Manager] (https://cloud.google.
 
 Quickstart
 ----------
-This library is currently under development and will be available soon!
-<!--TODO(ajaykannan): add in pom.xml snippet once gcloud-java-resourcemanager becomes available on maven -->
+If you are using Maven, add this to your pom.xml file
+```xml
+<dependency>
+  <groupId>com.google.gcloud</groupId>
+  <artifactId>gcloud-java-resourcemanager</artifactId>
+  <version>0.1.0</version>
+</dependency>
+```
+If you are using Gradle, add this to your dependencies
+```Groovy
+compile 'com.google.gcloud:gcloud-java-resourcemanager:jar:0.1.0'
+```
+If you are using SBT, add this to your dependencies
+```Scala
+libraryDependencies += "com.google.gcloud" % "gcloud-java-resourcemanager" % "0.1.0"
+```
 
 <!-- TODO(ajaykannan): once the API becomes usable, make an example application
 Example Application
@@ -38,7 +52,7 @@ Google [Cloud Resource Manager][cloud-resourcemanager] provides a programmatic w
 * Create new projects.
 * Update existing projects.
 * Delete projects.
-* Undelete, or recover, projects that you don't want to delete.
+* Undelete projects that you don't want to delete.
 
 Google Cloud Resource Manager is currently in beta and may occasionally make backwards incompatible changes.
 
@@ -74,7 +88,7 @@ All you need to create a project is a globally unique project ID.  You can also 
 import com.google.gcloud.resourcemanager.ProjectInfo;
 ```
 
-Then add the following code to create a project (be sure to change `myProjectId` to be something unique).
+Then add the following code to create a project (be sure to change `myProjectId` to your own unique project ID).
 
 ```java
 String myProjectId = "my-globally-unique-project-id"; // Change to a unique project ID.
@@ -121,7 +135,7 @@ while (projectIterator.hasNext()) {
 
 #### Complete source code
 
-Here we put together all the code shown above into one program.  This program assumes that you are running from your own desktop.
+Here we put together all the code shown above into one program.  This program assumes that you are running from your own desktop and used the Google Cloud SDK to authenticate yourself.
 
 ```java
 import com.google.gcloud.resourcemanager.ProjectInfo;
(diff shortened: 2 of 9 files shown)


mziccard:
(just being annoying here, feel free to skip this paragraph) In general it would be cool to avoid unnecessary merge commits: c992427 could have been avoided by rebasing your local `resource-manager-impl` on top of `resource-manager` before working on ca53a03 and sending the PR.

@aozarov If you have funnier stuff to do today I would be happy to give a first pass to this tomorrow morning.

aozarov:
Thanks @mziccard that helps!

ajkannan:
Rebased to remove merge commit

ajkannan:
@mziccard thanks for the pass!  I added a commit to include all your feedback thus far.

mziccard:
Only one more minor comment on my side. And an [open discussion](https://github.com/GoogleCloudPlatform/gcloud-java/pull/457#commitcomment-15024785) not necessarily related to this PR.

TESTING.md
```diff
@@ -65,5 +65,51 @@ Here is an example that clears the bucket created in Step 3 with a timeout of 5
...
+
+4. Stop the Resource Manager emulator.
+
+  ```java
+  helper.stop();
+  ```
+
+  This method will block until the server thread has been terminated.
+
+#### On a remote machine
```
aozarov: current emulator only works for localhost, right?

gcloud-java-resourcemanager/README.md
```diff
@@ -27,27 +27,145 @@ Example Application
...
+```
+
+#### Creating a project
+All you need to create a project is a globally unique project ID.  You can also optionally attach a non-unique name and labels to your project. Read more about naming guidelines for project IDs, names, and labels [here](https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects). To create a project, add the following import at the top of your file:
+
+```java
+import com.google.gcloud.resourcemanager.ProjectInfo;
+```
+
+Then add the following code to create a project (be sure to change `myProjectId` to be something unique).
```
aozarov: to your own unique project id.

gcloud-java-resourcemanager/README.md
```diff
@@ -27,27 +27,145 @@ Example Application
...
+To edit a project, create a new `ProjectInfo` object and pass it in to the `ResourceManager.replace` method.
+
+For example, to add a label for the newly created project to denote that it's launch status is "in development", add the following code:
+
+```java
+ProjectInfo newProjectInfo = resourceManager.replace(projectFromServer.toBuilder()
+    .addLabel("launch-status", "in-development").build());
+```
+
+Note that the values of the project you pass in to `replace` overwrite the server's values for non-read-only fields, namely `projectName` and `labels`. For example, if you create a project with `projectName` "some-project-name" and subsequently call replace using a `ProjectInfo` object that didn't set the `projectName`, then the server will unset the project's name. The server ignores any attempted changes to the read-only fields `projectNumber`, `lifecycleState`, and `createTime`. The `projectId` cannot change.
```
aozarov: BTW, does it retain the original createTime?

gcloud-java-resourcemanager/README.md
```diff
@@ -27,27 +27,145 @@ Example Application
...
+
+#### Listing all projects
+Suppose that we want a list of all projects for which we have read permissions. Add the following import:
```
aozarov: Does it actually work for user with only read permissions? (considering that our auth is specifying admin scope).

gcloud-java-resourcemanager/README.md
```diff
@@ -27,27 +27,145 @@ Example Application
...
+Iterator<ProjectInfo> projectIterator = resourceManager.list().iterateAll();
+System.out.println("Projects I can view:");
+while (projectIterator.hasNext()) {
+  System.out.println(projectIterator.next().projectId());
+}
+```
+
+#### Complete source code
+
+Here we put together all the code shown above into one program.  This program assumes that you are running from your own desktop.
```
aozarov: Shall we mention something like "and used the gcloud sdk to authenticate yourself"?

gcloud-java-resourcemanager/pom.xml
```diff
@@ -27,6 +27,12 @@
       <artifactId>google-api-services-cloudresourcemanager</artifactId>
       <version>v1beta1-rev6-1.19.0</version>
       <scope>compile</scope>
+      <exclusions>
```
aozarov: is that still an issue with the new api client?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerOptions.java
```diff
@@ -29,14 +30,14 @@
   private static final long serialVersionUID = 538303101192527452L;
   private static final String GCRM_SCOPE = "https://www.googleapis.com/auth/cloud-platform";
   private static final Set<String> SCOPES = ImmutableSet.of(GCRM_SCOPE);
+  private static final String DEFAULT_HOST = "https://cloudresourcemanager.googleapis.com";
```
aozarov: I am curious why this services uses a special host rather than the default "https://www.googleapis.com"?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerImpl.java
```diff
@@ -0,0 +1,235 @@
...
+              return resourceManagerRpc.get(projectId, optionsMap);
+            }
+          }, options().retryParams(), EXCEPTION_HANDLER);
+      return answer == null ? null : ProjectInfo.fromPb(answer);
+    } catch (RetryHelperException e) {
+      throw ResourceManagerException.translateAndThrow(e);
+    }
+  }
+
+  private abstract static class BasePageFetcher<T extends Serializable>
```
aozarov: This is a common base class for different services.

Not for this PR (so probably creating an issue) but we should move it to core and use Generics for the specific Options type `O` and provide an abstract method to return the `PAGE_TOKEN` option.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ProjectInfo.java
```diff
@@ -324,7 +323,7 @@ public Builder toBuilder() {
       projectPb.setLifecycleState(state.toString());
     }
     if (createTimeMillis != null) {
-      projectPb.setCreateTime(ISODateTimeFormat.dateTime().print(createTimeMillis));
+      projectPb.setCreateTime(ISODateTimeFormat.dateTime().withZoneUTC().print(createTimeMillis));
```
aozarov: `ProjetInfoTest` didn't need to change?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ProjectInfo.java
```diff
@@ -31,7 +31,6 @@
 
 /**
  * A Google Cloud Resource Manager project metadata object.
- *
```
aozarov: If you want to make checkstyle happy add a <p> after a new line (no new line after the <p>).
Also, State's ACTIVE and LIFECYCLE_STATE_UNSPECIFIED are missing a '.' in their javadoc.

Maybe after this PR run google checkstyle on all module files.

aozarov:
Looks good. Few nits and I didn't see the changes we discussed verbally.

TESTING.md
ajkannan: Yup, removing the "remote" section in next commit

gcloud-java-resourcemanager/README.md
ajkannan: Yup, it retains the original createTime

gcloud-java-resourcemanager/pom.xml
ajkannan: The cloudresourcemanager artifact uses version 1.19 of google-api-client, which pulls some version of guava-jdk5 (maybe 13, but I'd have to double check).  guava-jdk5 version 13 had an old version of clock, which caused build issues.  The latest api client verison is 1.21, which also pulls guava-jdk5, but it doesn't specify a specific version and I didn't check compatibility with clock.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ProjectInfo.java
ajkannan: It was using our local time zone, so converting to and from milliseconds wasn't an issue.  I'll add a test that makes sure that the serialized time zone ends with a 'Z' for UTC

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerOptions.java
ajkannan: Unsure, I got that URL from the docs: https://cloud.google.com/resource-manager/reference/rest/

gcloud-java-resourcemanager/README.md
ajkannan: Yes, I can view read-only projects using "get" and "list" even though the oauth scope is read/write

ajkannan:
Note: the latest commit addresses comments and also handles the issue of "Unnamed" being returned by "list" as the project name for projects with a null name.

PR #460, opened 2015-12-16
Logging API. Add generated files.

pongad:
These are the generated files for Logging API. I have made the PR against pubsub-alpha, since that is where the new branch is based off of. Ideally, this commit should go into ''logging-alpha'' branch of the repo.


pongad:
Will redo this against the right branch.

Issue #382, opened 2015-11-17
In ServiceApi classes, provide a mechanism to determine if an api method call can be retried on failure


garrettjonesgoogle:
An api method call should be retryable if it is idempotent, not retryable if not idempotent. We need to figure out the best way to surface this, whether in an exception, the ApiCallable, the ClientCall, some registry, or somewhere else.

garrettjonesgoogle:
This is done for pubsub, but not logging yet.

garrettjonesgoogle:
This is now supported:

* Retryable codes can be customized (but come with defaults), so that the built in retry system will retry only on those codes
* The ApiCallable chain now has https://github.com/googleapis/gax-java/blob/master/src/main/java/com/google/api/gax/grpc/ExceptionTransformingCallable.java , which will convert a StatusRuntimeException into an ApiException, which holds the knowledge of whether the exception is retryable, based on the retryable codes that were set. 

@aozarov , I am closing out this issue now.

PR #463, opened 2015-12-16
Add Logging APIs generated by VKit.

pongad:
Some tests are written but definitely not completed yet.
It shows, however, that the API is not dead on arrival.

/cc @anthmgoogle @garrettjonesgoogle

### DIFF
--- a/gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
+++ b/gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
@@ -1,0 +1,533 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/*
+ * EDITING INSTRUCTIONS
+ * This file was generated from the file
+ * https://github.com/google/googleapis/blob/master/google/logging/v2/logging_config.proto
+ * and updates to that file get reflected here through a refresh process.
+ * For the short term, the refresh process will only be runnable by Google engineers.
+ * Manual additions are allowed because the refresh process performs
+ * a 3-way merge in order to preserve those manual additions. In order to not
+ * break the refresh process, only certain types of modifications are
+ * allowed.
+ *
+ * Allowed modifications - currently these are the only types allowed:
+ * 1. New methods (these should be added to the end of the class)
+ * 2. New imports
+ * 3. Additional documentation between "manual edit" demarcations
+ *
+ * Happy editing!
+ */
+package com.google.gcloud.logging.spi.v2;
+
+import com.google.logging.v2.ConfigServiceV2Grpc;
+import com.google.logging.v2.CreateSinkRequest;
+import com.google.logging.v2.DeleteSinkRequest;
+import com.google.logging.v2.GetSinkRequest;
+import com.google.logging.v2.ListSinksRequest;
+import com.google.logging.v2.ListSinksResponse;
+import com.google.logging.v2.LogSink;
+import com.google.logging.v2.UpdateSinkRequest;
+import com.google.protobuf.Empty;
+import io.gapi.gax.grpc.ApiCallable;
+import io.gapi.gax.grpc.PageDescriptor;
+import io.gapi.gax.grpc.ServiceApiSettings;
+import io.gapi.gax.internal.ApiUtils;
+import io.gapi.gax.protobuf.PathTemplate;
+import io.grpc.ManagedChannel;
+import java.io.IOException;
+
+// Manually-added imports: add custom (non-generated) imports after this point.
+
+
+
+// AUTO-GENERATED DOCUMENTATION AND SERVICE - see instructions at the top of the file for editing.
+/**
+ * See //google/logging/v2/logging.proto for documentation
+ *
+ * <!-- manual edit -->
+ * <!-- end manual edit -->
+ */
+@javax.annotation.Generated("by API code generation")
+public class ConfigServiceV2Api implements AutoCloseable {
+
+  // =========
+  // Constants
+  // =========
+
+  /**
+   * The default address of the service.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String SERVICE_ADDRESS = "logging.googleapis.com";
+
+  /**
+   * The default port of the service.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final int DEFAULT_SERVICE_PORT = 443;
+
+
+  private static final ApiCallable<ListSinksRequest, ListSinksResponse>
+      LIST_SINKS = ApiCallable.create(ConfigServiceV2Grpc.METHOD_LIST_SINKS);
+  private static final ApiCallable<GetSinkRequest, LogSink>
+      GET_SINK = ApiCallable.create(ConfigServiceV2Grpc.METHOD_GET_SINK);
+  private static final ApiCallable<CreateSinkRequest, LogSink>
+      CREATE_SINK = ApiCallable.create(ConfigServiceV2Grpc.METHOD_CREATE_SINK);
+  private static final ApiCallable<UpdateSinkRequest, LogSink>
+      UPDATE_SINK = ApiCallable.create(ConfigServiceV2Grpc.METHOD_UPDATE_SINK);
+  private static final ApiCallable<DeleteSinkRequest, Empty>
+      DELETE_SINK = ApiCallable.create(ConfigServiceV2Grpc.METHOD_DELETE_SINK);
+
+  private static PageDescriptor<ListSinksRequest, ListSinksResponse, LogSink> LIST_SINKS_PAGE_DESC =
+      new PageDescriptor<ListSinksRequest, ListSinksResponse, LogSink>() {
+        @Override
+        public Object emptyToken() {
+          return "";
+        }
+        @Override
+        public ListSinksRequest injectToken(
+            ListSinksRequest payload, Object token) {
+          return ListSinksRequest
+            .newBuilder(payload)
+            .setPageToken((String) token)
+            .build();
+        }
+        @Override
+        public Object extractNextToken(ListSinksResponse payload) {
+          return payload.getNextPageToken();
+        }
+        @Override
+        public Iterable<LogSink> extractResources(ListSinksResponse payload) {
+          return payload.getSinksList();
+        }
+      };
+
+  private static String ALL_SCOPES[] = {
+    "https://www.googleapis.com/auth/logging.read",
+    "https://www.googleapis.com/auth/logging.write",
+    "https://www.googleapis.com/auth/logging.admin"
+  };
+
+  /**
+   * A PathTemplate representing the fully-qualified path to represent
+   * a project_name resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  private static final PathTemplate PROJECT_NAME_PATH_TEMPLATE =
+      PathTemplate.create("/projects/{project}");
+  /**
+   * A PathTemplate representing the fully-qualified path to represent
+   * a sink_name resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  private static final PathTemplate SINK_NAME_PATH_TEMPLATE =
+      PathTemplate.create("/projects/{project}/sinks/{sink}");
+
+  // ========
+  // Members
+  // ========
+
+  private final ManagedChannel channel;
+  private final ServiceApiSettings settings;
+
+  // ===============
+  // Factory Methods
+  // ===============
+
+  /**
+   * Constructs an instance of ConfigServiceV2Api with default settings.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static ConfigServiceV2Api create() throws IOException {
+    return create(new ServiceApiSettings());
+  }
+
+  /**
+   * Constructs an instance of ConfigServiceV2Api, using the given settings. The channels are created based
+   * on the settings passed in, or defaults for any settings that are not set.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static ConfigServiceV2Api create(ServiceApiSettings settings) throws IOException {
+    return new ConfigServiceV2Api(settings);
+  }
+
+  /**
+   * Constructs an instance of ConfigServiceV2Api, using the given settings. This is protected so that it
+   * easy to make a subclass, but otherwise, the static factory methods should be preferred.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  protected ConfigServiceV2Api(ServiceApiSettings settings) throws IOException {
+    ServiceApiSettings internalSettings = ApiUtils.populateSettings(settings,
+        SERVICE_ADDRESS, DEFAULT_SERVICE_PORT, ALL_SCOPES);
+    this.settings = internalSettings;
+    this.channel = internalSettings.getChannel();
+  }
+
+  // ==============================
+  // Resource Name Helper Functions
+  // ==============================
+
+  /**
+   * Creates a string containing the fully-qualified path to represent
+   * a project_name resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String createProjectNamePath(String project) {
+    return PROJECT_NAME_PATH_TEMPLATE.instantiate(
+        "project", project);
+  }
+
+  /**
+   * Creates a string containing the fully-qualified path to represent
+   * a sink_name resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String createSinkNamePath(String project, String sink) {
+    return SINK_NAME_PATH_TEMPLATE.instantiate(
+        "project", project,"sink", sink);
+  }
+
+
+  /**
+   * Extracts the project from the given fully-qualified path which
+   * represents a projectName resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String extractProjectFromProjectNamePath(String projectNamePath) {
+    return PROJECT_NAME_PATH_TEMPLATE.parse(projectNamePath).get("project");
+  }
+
+  /**
+   * Extracts the project from the given fully-qualified path which
+   * represents a sinkName resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String extractProjectFromSinkNamePath(String sinkNamePath) {
+    return SINK_NAME_PATH_TEMPLATE.parse(sinkNamePath).get("project");
+  }
+
+  /**
+   * Extracts the sink from the given fully-qualified path which
+   * represents a sinkName resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String extractSinkFromSinkNamePath(String sinkNamePath) {
+    return SINK_NAME_PATH_TEMPLATE.parse(sinkNamePath).get("sink");
+  }
+
+
+  // =============
+  // Service Calls
+  // =============
+
+  // ----- listSinks -----
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Lists sinks.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public Iterable<LogSink> listSinks(String projectName) {
+    ListSinksRequest request =
+        ListSinksRequest.newBuilder()
+        .setProjectName(projectName)
+        .build();
+    return listSinks(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Lists sinks.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param request The request object containing all of the parameters for the API call.
+   */
+  public Iterable<LogSink> listSinks(ListSinksRequest request) {
+    return listSinksStreamingCallable()
+        .iterableResponseStreamCall(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Lists sinks.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public ApiCallable<ListSinksRequest, LogSink> listSinksStreamingCallable() {
+    return listSinksCallable().pageStreaming(LIST_SINKS_PAGE_DESC);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Lists sinks.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public ApiCallable<ListSinksRequest, ListSinksResponse> listSinksCallable() {
+    return ApiUtils.prepareIdempotentCallable(LIST_SINKS, settings).bind(channel);
+  }
+
+  // ----- getSink -----
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Gets a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param sinkName The resource name of the sink to return.
+   * Example: `"projects/my-project-id/sinks/my-sink-id"`.
+   */
+  public LogSink getSink(String sinkName) {
+    GetSinkRequest request =
+        GetSinkRequest.newBuilder()
+        .setSinkName(sinkName)
+        .build();
+
+    return getSink(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Gets a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param request The request object containing all of the parameters for the API call.
+   */
+  public LogSink getSink(GetSinkRequest request) {
+    return getSinkCallable().call(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Gets a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public ApiCallable<GetSinkRequest, LogSink> getSinkCallable() {
+    return ApiUtils.prepareIdempotentCallable(GET_SINK, settings).bind(channel);
+  }
+
+  // ----- createSink -----
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Creates a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param projectName The resource name of the project in which to create the sink.
+   * Example: `"projects/my-project-id"`.
+   *
+   * The new sink must be provided in the request.
+   * @param sink The new sink, which must not have an identifier that already
+   * exists.
+   */
+  public LogSink createSink(String projectName, LogSink sink) {
+    CreateSinkRequest request =
+        CreateSinkRequest.newBuilder()
+        .setProjectName(projectName)
+        .setSink(sink)
+        .build();
+
+    return createSink(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Creates a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param request The request object containing all of the parameters for the API call.
+   */
+  public LogSink createSink(CreateSinkRequest request) {
+    return createSinkCallable().call(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Creates a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public ApiCallable<CreateSinkRequest, LogSink> createSinkCallable() {
+    return CREATE_SINK.bind(channel);
+  }
+
+  // ----- updateSink -----
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Creates or updates a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param sinkName The resource name of the sink to update.
+   * Example: `"projects/my-project-id/sinks/my-sink-id"`.
+   *
+   * The updated sink must be provided in the request and have the
+   * same name that is specified in `sinkName`.  If the sink does not
+   * exist, it is created.
+   * @param sink The updated sink, whose name must be the same as the sink
+   * identifier in `sinkName`.  If `sinkName` does not exist, then
+   * this method creates a new sink.
+   */
+  public LogSink updateSink(String sinkName, LogSink sink) {
+    UpdateSinkRequest request =
+        UpdateSinkRequest.newBuilder()
+        .setSinkName(sinkName)
+        .setSink(sink)
+        .build();
+
+    return updateSink(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Creates or updates a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param request The request object containing all of the parameters for the API call.
+   */
+  public LogSink updateSink(UpdateSinkRequest request) {
+    return updateSinkCallable().call(request);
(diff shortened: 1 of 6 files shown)


gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    logs.remove(request.getLogName());
+    responseObserver.onNext(Empty.getDefaultInstance());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void writeLogEntries(WriteLogEntriesRequest request,
+      StreamObserver<WriteLogEntriesResponse> responseObserver) {
+    List<LogEntry> entries = logs.get(request.getLogName());
+    if (entries == null) {
```
aozarov: this is not thread-safe.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
```diff
@@ -0,0 +1,97 @@
...
+import com.google.protobuf.Empty;
+
+import io.grpc.stub.StreamObserver;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class LocalLoggingImpl implements LoggingServiceV2Grpc.LoggingServiceV2 {
```
aozarov: I am not sure why we would want to make it public (and Helper could return the interface rather impl).

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    entries.addAll(request.getEntriesList());
+    // Response is empty
+    responseObserver.onNext(WriteLogEntriesResponse.newBuilder().build());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void listLogEntries(ListLogEntriesRequest request,
+      StreamObserver<ListLogEntriesResponse> responseObserver) {
+    List<LogEntry> entries = new ArrayList<>();
```
aozarov: Do you plan to support `filter`, `order_by` and respect `page_size`?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    for (ByteString proj : request.getProjectIdsList().asByteStringList()) {
+      String prefix = "projects/" + proj.toStringUtf8() + "/";
+      for (Map.Entry<String, List<LogEntry>> entry : logs.entrySet()) {
+        if (entry.getKey().startsWith(prefix)) {
+          entries.addAll(entry.getValue());
+        }
+      }
+    }
+    responseObserver.onNext(ReadLogEntriesResponse.newBuilder()
```
aozarov: This one method is streaming, is it expected to have multiple onNext calls rather than one big onNext callback?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/testing/LocalLoggingHelper.java
```diff
@@ -0,0 +1,77 @@
...
+import io.grpc.netty.NettyChannelBuilder;
+import io.grpc.netty.NettyServerBuilder;
+import io.netty.channel.local.LocalAddress;
+import io.netty.channel.local.LocalChannel;
+import io.netty.channel.local.LocalServerChannel;
+
+import java.io.IOException;
+import java.net.SocketAddress;
+
+public class LocalLoggingHelper {
```
aozarov: javadoc

gcloud-java-logging/src/main/java/com/google/gcloud/logging/testing/LocalLoggingHelper.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private final SocketAddress address;
+  private final Server server;
+  private final LocalLoggingImpl loggingImpl;
+
+  /**
+   * Constructs a new LocalPubsubHelper. The method start() must
+   * be called before it is used.
+   */
+  public LocalLoggingHelper(String addressString) {
```
aozarov: can you just pick a random name for address?

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Test
+  public void testWriteLog() throws Exception {
+    String logName = LoggingServiceV2Api.createLogNamePath("my-project", "my-log");
+    MonitoredResource resource = MonitoredResource.newBuilder().build();
+    List<LogEntry> entries = new ArrayList<>();
+    entries.add(LogEntry.newBuilder()
+        .setLogName(logName)
+        .setTextPayload("foobar")
+        .build()
```
aozarov: I am not 100% about the google-style about it but I feel that `.build());` is prefered.

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
```diff
@@ -0,0 +1,132 @@
...
+    );
+    loggingApi.writeLogEntries(logName, resource, Collections.<String, String>emptyMap(), entries);
+
+    Iterable<LogEntry> gotEntries = loggingApi.listLogEntries(
+        Collections.singletonList("my-project"), "", ""
```
aozarov: I think  `loggingApi.listLogEntries(Collections.singletonList("my-project"), "", "");` will fit in one line
and therefore preferred wrapping would be after the ' ='  sign.

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
```diff
@@ -0,0 +1,132 @@
...
+    );
+    List<LogEntry> gotList = new ArrayList<>();
+    for (LogEntry e : gotEntries) {
+      gotList.add(e);
+    }
+    Assert.assertEquals(entries, gotList);
+  }
+
+  @Test
```
aozarov: what is throwing the 'Exception`?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/MetricsServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/*
+ * EDITING INSTRUCTIONS
+ * This file was generated from the file
+ * https://github.com/google/googleapis/blob/master/google/logging/v2/logging_metrics.proto
```
aozarov: This link is broken. I don't think logging_metrics.proto is in the googleapis repository yet.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+
+
+// AUTO-GENERATED DOCUMENTATION AND SERVICE - see instructions at the top of the file for editing.
+/**
+ * See //google/logging/v2/logging.proto for documentation
+ *
+ * <!-- manual edit -->
+ * <!-- end manual edit -->
+ */
+@javax.annotation.Generated("by the veneer generator")
```
aozarov: Should we use "veneer" publically?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+import io.grpc.ManagedChannel;
+import java.io.IOException;
+
+// Manually-added imports: add custom (non-generated) imports after this point.
+
+
+
+// AUTO-GENERATED DOCUMENTATION AND SERVICE - see instructions at the top of the file for editing.
+/**
+ * See //google/logging/v2/logging.proto for documentation
```
aozarov: Should be a public reference.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+        @Override
+        public Iterable<LogSink> extractResources(ListSinksResponse payload) {
+          return payload.getSinksList();
+        }
+      };
+
+  private static String ALL_SCOPES[] = {
+    "https://www.googleapis.com/auth/logging.read",
+    "https://www.googleapis.com/auth/logging.write",
+    "https://www.googleapis.com/auth/logging.admin"
```
aozarov: Nit, this one includes the 2 above.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
aozarov: Do you plan to also implement `ConfigServiceV2`?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+
+// AUTO-GENERATED DOCUMENTATION AND SERVICE - see instructions at the top of the file for editing.
+/**
+ * See //google/logging/v2/logging.proto for documentation
+ *
+ * <!-- manual edit -->
+ * <!-- end manual edit -->
+ */
+@javax.annotation.Generated("by the veneer generator")
+public class ConfigServiceV2Api implements AutoCloseable {
```
aozarov: This class is missing test.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+
+  // ==============================
+  // Resource Name Helper Functions
+  // ==============================
+
+  /**
+   * Creates a string containing the fully-qualified path to represent
+   * a project_name resource.
+   *
+   * <!-- manual edit -->
```
aozarov: Any reason for having this manual edit comment in a javadoc of member variables or methods.

If you really support manual edit of javadoc how do you handle tags?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+  // ----- listSinks -----
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Lists sinks.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public Iterable<LogSink> listSinks(String projectName) {
```
aozarov: Any updates on adding @throws and supporting an "is retryable" flag?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+  /**
+   * Initiates an orderly shutdown in which preexisting calls continue but new calls are immediately
+   * cancelled.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  @Override
+  public void close() {
+    // Manually-added shutdown code
```
aozarov: Did you consider methods that one could override (no-op) by default for handling such interceptions?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
```diff
@@ -0,0 +1,583 @@
...
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String extractProjectFromLogNamePath(String logNamePath) {
+    return LOG_NAME_PATH_TEMPLATE.parse(logNamePath).get("project");
+  }
+
+  /**
+   * Extracts the log from the given fully-qualified path which
```
aozarov: I am not sure what it means "extracts the log" ...

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
```diff
@@ -0,0 +1,583 @@
...
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Writes log entries to Cloud Logging.
+   * All log entries in Cloud Logging are written by this method.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param logName Optional. A default log resource name for those log entries in `entries`
```
aozarov: what happens when no logName is provided and some entries do not specify their own `logName`.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
```diff
@@ -0,0 +1,583 @@
...
+   * that do not specify their own `logName`.  Example:
```
aozarov: Here and in other places. I think using {@code logName} would be better than `logName`.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
```diff
@@ -0,0 +1,533 @@
...
+  // ----- updateSink -----
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Creates or updates a sink.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   *
+   * @param sinkName The resource name of the sink to update.
```
aozarov: The service definition is not consistent with `Required` and `Optional`.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
```diff
@@ -0,0 +1,583 @@
...
+        .addAllEntries(entries)
+        .build();
+
+    return writeLogEntries(request);
+  }
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Writes log entries to Cloud Logging.
+   * All log entries in Cloud Logging are written by this method.
```
aozarov: I think having "All log entries in Cloud Logging are written by this method." is strange when you have 3 of them saying the same.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
```diff
@@ -0,0 +1,583 @@
...
+    return WRITE_LOG_ENTRIES.bind(channel);
+  }
+
+  // ----- listLogEntries -----
+
+  // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
+  /**
+   * Lists log entries.
+   * Use this method to examine log entries from Cloud Logging.
+   * See [Exporting Logs](/logging/docs/export) for other ways to copy
```
aozarov: We should find a way to replace "See [Exporting Logs](/logging/docs/export)" with a "@see and a valid doc URL".

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
anthmgoogle: I was not expecting to see this pattern in the API yet. I suggest we turn off the StreamObersver pattern for now. We can discuss bringing it back when the design and implementation pattern are finalized, including making it thread safe.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
anthmgoogle: Probably not. Suggest something vague like "api code generation".

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
anthmgoogle: This may not be practical to fix, because this is automated, and the knowledge of scope containment lives elsewhere. This should not do any harm and is not part of the public surface.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
anthmgoogle: I don't think that part is odd, although I don't think the "all" part is clear. A bug in the underlying protodoc we should track. We have been avoiding making local edits to the proto-doc so far.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: We will track this in #471 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: Done.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: Track in #472 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: I'm not sure how this is related. If the tags are manually entered, they can be placed inside the block.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: I think it is consistent though potentially confusing. The sink must be provided in the request. However, if the sink doesn't already exist on Logging, it is created. It is possible to "update" a nonexistent sink. Tracking in #473 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
pongad: logNamePath contains two parameters: project and log. This method extracts the parameter log from the path. I agree that this is confusing. Tracking in #474 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
pongad: protodoc issue. Tracking in #475

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
pongad: Tracking in #476

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
pongad: Tracking in #471 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
pongad: Both of this tracked in #472 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: If method is idempotent, the client library will take care of retrying calls for you.
As for the throws tag, #477 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
pongad: Will resolve as part of #472 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
pongad: Part of #472 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/testing/LocalLoggingImpl.java
pongad: Part of #472 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/testing/LocalLoggingHelper.java
pongad: Part of #472 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/testing/LocalLoggingHelper.java
pongad: Part of #472

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
pongad: Part of #472 .

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
pongad: Part of #472.

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
pongad: Part of #472.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/testing/LocalLoggingHelper.java
aozarov: I don't think `LocalLoggingHelper` (and its associated comments) is part of #472.
#472 is for providing an alternative for a local fake implementation (`LocalLoggingImpl`).
Like in other services we provide the user (and also use it in our testing) a helper (like this one)
to *get* a local services (which could come from SDK, public location or local implementation).

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
aozarov: I don't expect you to put much efforts on the local fake if you think we can get it from a different source (#472), however as long as we have this test we should make sure it adheres the public google style (and the repository style). Same for bellow. Did you try run the [public google formatter](https://github.com/google/styleguide/) tool on the code?

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
aozarov: As noted above, this should not be part of #472 (unless you plan to remove this code and wait until that issue is resolved).

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/MetricsServiceV2Api.java
aozarov: the comment above was not addressed.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
aozarov: I think the scope of #472 is too wide. I assume that it is for using a cloud emulator (once provided) instead of the local fake implementation. If you plan to defer writing tests for the spi layer until then create a separate issue for it and make it depends on #472.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
aozarov: This "" marker is on a method that is not marked as "auto generated".
This method also does not specify the "@param" tag. 

What will happen if a user will specifies it as part of the "manual edit" and then later you also
specify it (or add another tag e.g. "@see")?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
aozarov: I think we should propagate this information (is failure transient and retriable) to the user (or the hand-crafted layer) so they can decide if they want to retry it themselves or take any further actions. see #382.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
aozarov: comment?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
aozarov: creating an issue?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/testing/LocalLoggingHelper.java
pongad: Understood. Done.

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
pongad: Apologies for this. I ran the formatter, but did not include the change in the commit. This should be fixed now along with other less obvious stylistic problems.

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
pongad: Done.

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
pongad: As the above, change not included in the commit. Done. Sorry for the back-and-forth :(

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/MetricsServiceV2Api.java
pongad: This line is generated. Since the change will affect everyone, let's track this in #479 for now.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: I understand your point. This will affect all APIs, tracking in #480

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: Makes perfect sense. Will track this in the issue you mention. Since these methods are all generated, they should automatically get the upgrade.

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
pongad: I am not sure if this is worth it. The manual sections are there so that if the people who generate the API need a more sophisticated shutdown sequence, they can add it. Delegating this to another method, in my humble opinion, just adds more methods with no real gain.

pongad: @garrettjonesgoogle ?

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2Api.java
pongad: Tracking in #481 .

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
garrettjonesgoogle: Our general approach has been to have modifications done through modifying this class directly. Having modifications through overrides is a different approach that we didn't design for.

gcloud-java-logging/src/test/java/com/google/gcloud/logging/spi/v2/LoggingServiceV2ApiTest.java
```diff
@@ -0,0 +1,117 @@
...
+    MonitoredResource resource = MonitoredResource.newBuilder().build();
+    List<LogEntry> entries = new ArrayList<>();
+    entries.add(LogEntry.newBuilder().setLogName(logName).setTextPayload("foobar").build());
+    loggingApi.writeLogEntries(logName, resource, Collections.<String, String>emptyMap(), entries);
+
+    loggingApi.deleteLog(logName);
+
+    Iterable<LogEntry> gotEntries =
+        loggingApi.listLogEntries(Collections.singletonList("my-project"), "", "");
+    List<LogEntry> gotList = new ArrayList<>();
```
aozarov: Nit, alternative:
`assertTrue(Iterables.isEmpty(goEntries))`
or (if, like current code, you want to *show* the extra entries)
`assertTrue(goEntries.toString(), Iterables.isEmpty(goEntries))`

BTW, for converting `Iterable` to `List` you could use `Lists.newArrayList(goEntries)`

gcloud-java-logging/src/main/java/com/google/gcloud/logging/spi/v2/ConfigServiceV2Api.java
aozarov: Ok, your call, but let me explain my suggestion (at least for the record).

So far it looked to me that your general approach for modifying the class directly by a user is for adding new functionality or documentation rather than modifying  the methods themselves (this is also what you mention in the "EDITING INSTRUCTIONS").

The other approach would be to override, which is also indirectly documented.
 
The "close" method seems to get a "special" treatment (which violates the general editing instruction). I am not sure if that is worth it  (and if so, better documentation would be useful). If not worth it, then one could just achieve
the same by overriding and delegating (between "before" and "after" logic).
If "close" was critical and you would not want people to "mess it up", then
you could make it final and provide the 2 before and after methods to be
overridden.

garrettjonesgoogle: I should mention what drove me to put this here in the first place. When I was implementing the buffered publisher for pubsub, I needed to have the buffered publisher closed when the wrapper was closed (since it has threads that are running). At that point in time, the buffered publisher was going to be a handwritten modification to the service wrapper, so I needed to add these markers to support the manual additions to close(). Now that the buffered publisher is going to be auto-generated, there is not a current use case that needs the manual markers.

Now if we were going to support before & after close logic, I might actually prefer to use a compositional approach instead of an inheritance-based approach. It would look like this: There would be an interface for the close operation (CloseOperation), and clients could prepend and append close operations to the wrapper. The wrapper would store a list of these operations, and would run them in sequence at close time. Now that I think about it, I might implement this for the buffered publisher, because then it would already support future flexibility for other use cases.

aozarov: I am not sure I 100% understand your suggestion. Are you suggesting adding some close listener approach? In any case, do you think we should create an issue for it?

garrettjonesgoogle: Yes, you can see it as a listener. File issue https://github.com/GoogleCloudPlatform/gcloud-java/issues/483 .

Issue #440, opened 2015-12-05
Cloud Resource Manager error message swap


ajkannan:
403 forbidden (permission denied) is returned instead of 404 not found when a project doesn't exist.  For the time being, we will assume 403 means the project wasn't found.  We should change back our implementation once this is fixed.

Also, 400 bad request is returned when attempting to perform actions on a project that is in an incorrect lifecycle state.  This may make more sense with error code 412 (precondition failed) according to the [public documentation](https://cloud.google.com/resource-manager/v1/errors/core_errors#PRECONDITION_FAILED), since the error message contains the message "precondition check failed."

ajkannan:
The 403 vs 404 is intended behavior, and a check for 404 is added in #491, and the 400 vs 412 isn't a pressing or obvious issue, so I'm closing this issue.

PR #491, opened 2015-12-18
Add more retryable exceptions + cleanup

### LINKED ISSUE #429: Check retryable exceptions
There are some exceptions (i.e. 403, with reason "rateLimitExceeded") currently unlisted which may be retryable.  We should check best practices with the Cloud Resource Manager team.

ajkannan:
Closes #429

Included in the cleanup is a check for error code 404 not found when using "get" (see #440).

### DIFF
--- a/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/package-info.java
+++ b/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/package-info.java
@@ -18,7 +18,7 @@
  * A testing helper for Google Cloud Resource Manager.
  *
  * <p>A simple usage example:
- * <p>Before the test:
+ * Before the test:
  * <pre> {@code
  * LocalResourceManagerHelper resourceManagerHelper = LocalResourceManagerHelper.create();
  * ResourceManager resourceManager = resourceManagerHelper.options().service();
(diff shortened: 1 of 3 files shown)


gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -31,7 +31,7 @@
  */
 public interface ResourceManager extends Service<ResourceManagerOptions> {
 
-  public static final String DEFAULT_CONTENT_TYPE = "application/octet-stream";
+  String DEFAULT_CONTENT_TYPE = "application/octet-stream";
```
mziccard: I might be missing some background here but why did this become an instance variable?

aozarov: This is an interface.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/spi/DefaultResourceManagerRpc.java
```diff
@@ -82,7 +88,7 @@ public Project get(String projectId, Map<Option, ?> options) throws ResourceMana
           .execute();
     } catch (IOException ex) {
       ResourceManagerException translated = translate(ex);
-      if (translated.code() == HTTP_FORBIDDEN) {
+      if (translated.code() == HTTP_FORBIDDEN || translated.code() == HTTP_NOT_FOUND) {
```
mziccard: Just out of curiosity and related to #440: is 403 returned for both the not found and permission denied case? Is this done not to disclose info on existing projects?

ajkannan: Yup, 403 is returned for both not found and permission denied, and that is indeed the reason why.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
mziccard: Looks like one :)

PR #494, opened 2015-12-21
Remove RemoteGcsHelper.create(String, String) and other minor fixes

mziccard:
This PR does the following:
- Remove `RemoteGcsHelper.create(String, String)` (2808d95)
- Fix storage example in `testing/package-info.java` (2808d95)
- Fix functional classes tests (2808d95)
- Change storage code to make checkstyle happy (6edd748)

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
@@ -479,8 +479,8 @@ public class DefaultStorageRpc implements StorageRpc {
   }
 
   @Override
-  public void write(String uploadId, byte[] toWrite, int toWriteOffset, StorageObject dest,
-      long destOffset, int length, boolean last) throws StorageException {
+  public void write(String uploadId, byte[] toWrite, int toWriteOffset, long destOffset, int length,
+      boolean last) throws StorageException {
     try {
       GenericUrl url = new GenericUrl(uploadId);
       HttpRequest httpRequest = storage.getRequestFactory().buildPutRequest(url,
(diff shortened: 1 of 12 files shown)


aozarov:
Looks good to me. @ajkannan good finding. Some more cleanup items (could be done in a separate PR):

* BlobInfo#580 - Inefficient keySet iterator (Use entrySet).
* Storage interface has redundant "public static final" modifiers for DEFAULT_CONTENT_TYPE.
* BlobInfo.Builder#metadata is missing generic type on the right side.
* ITStorageTest is missing generic type for metadata in testListBlobsSelectedFields and testListBlobsSelectedFields
* StorageRpc.write - we don't use the dest param (I think was added for version check, but not used because the feature is missing in the JSON API). Should we
remove it (or add a comment suggesting why we have it)?
*  Bucket.BlobPageFetcher#nextPage, right side assignment could use the diamond shortcut
* BlobTest.testLoadFromId - should we add assertNotNull(loadedBlob) before using it? (Same comment for BucketTest.testLoad).
* StorageImplTest - multiple cases of redundant generic type declaration when
  using captured options - > assignment could use Capture.newInstance()

* Spelling issues:
Acl.java - 'Creats' 
BlobTest - 'delele'
DefaultStorageRpc - 'Reponse'



BTW, I keep seeing this message for the maven (applies to all modules):
"Definition of groupId is redundant, because it's inherited from the parent "

mziccard:
@ajkannan I fixed your comments in place
@aozarov I added efb1ae6 to fix your comments

I will look into maven's complaint!

aozarov:
I think you missed this one:
"ITStorageTest is missing generic type for metadata in testListBlobsSelectedFields and testListBlobsSelectedFields"

mziccard:
Changed to assert and fixed missing comments in place

aozarov:
> I will look into maven's complaint!

Hey @mziccard did you have a chance to look at that, "Definition of groupId is redundant", complaint?

mziccard:
Actually I forgot :) I'll look into this.

Issue #451, opened 2015-12-11
Synchronize LocalResourceManagerHelper


ajkannan:
Right now, the mock works fine if there's only one connection to it.  However, if there are multiple connections, there are potential race conditions in some of the handler methods (such as `delete`).

ajkannan:
Fixed by #497

PR #497, opened 2015-12-21
Synchronize LocalResourceManagerHelper

ajkannan:
Fix for #451


gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -324,12 +322,9 @@ Response create(Project project) {
     }
   }
 
-  Response delete(String projectId) {
+  synchronized Response delete(String projectId) {
```
aozarov: This is a very coarse synchronization which could be done more efficient but I agree that for a test helper it is not worth the trouble.

aozarov: At the moment list (and maybe get in the future) don't have a filter that applies on life-cycle.
If that will change you may have a problem where the output and the check may not match (as you change the same object). If you want you can avoid it by always replacing the Project (like in replace method).

aozarov:
Looks good to me. One suggestion that does not have any immediate benefit but could make future changes simpler. Feel free to merge as is if you like.

PR #498, opened 2015-12-22
Make protobuf methods package protected

### LINKED ISSUE #100: API includes protobuf references
Explicitly, or implicitly. For example, [StructuredQuery](http://googlecloudplatform.github.io/gcloud-java/apidocs/com/google/gcloud/datastore/StructuredQuery.html) has protected methods `fromPb`, `populatePb`, and `toPb` which should not appear in the API.

ajkannan:
Fixes #100

### DIFF
--- a/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/Value.java
+++ b/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/Value.java
@@ -80,9 +80,9 @@ public abstract class Value<V> extends Serializable<DatastoreV1.Value> {
       return builder.build();
     }
 
-    abstract V getValue(DatastoreV1.Value from);
+    protected abstract V getValue(DatastoreV1.Value from);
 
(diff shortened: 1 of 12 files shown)


gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/Value.java
```diff
@@ -80,9 +80,9 @@ public final B fromProto(DatastoreV1.Value proto) {
       return builder.build();
     }
 
-    protected abstract V getValue(DatastoreV1.Value from);
```
aozarov: Because Marshallers implementations are package scope I don't think it has an impact on the Javadoc.

ajkannan: True, I'll revert this and related changes.

ajkannan:
@aozarov added a commit to revert the unnecessary changes to BaseMarshaller

PR #501, opened 2015-12-22
Remove javadoc refs to non-public classes

### LINKED ISSUE #103: API documentation exposes non-public types
For example, [KeyFactory](http://googlecloudplatform.github.io/gcloud-java/apidocs/com/google/gcloud/datastore/KeyFactory.html) has a bunch of methods that return `B`, which is actually a type parameter of the non-public parent class. But nothing indicates that in the API documentation. We should eliminate cases like this, which might sometimes involve overriding methods just to specify a concrete return type and return `super.whatever()`.

ajkannan:
Fixes #103

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BaseTableInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BaseTableInfo.java
@@ -94,6 +94,12 @@ public abstract class BaseTableInfo implements Serializable {
   private final Long expirationTime;
   private final Long lastModifiedTime;
 
+  /**
+   * Base builder for tables.
+   *
+   * @param <T> the table type.
+   * @param <B> the table builder.
+   */
   public abstract static class Builder<T extends BaseTableInfo, B extends Builder<T, B>> {
 
     private String etag;
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobInfo.java
@@ -32,6 +32,8 @@ import java.util.Objects;
  * a table. Use {@link QueryJobInfo} for a job that runs a query.
  *
  * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs">Jobs</a>
+ *
+ * @param <S> the statistics type.
  */
 public abstract class JobInfo<S extends JobStatistics> implements Serializable {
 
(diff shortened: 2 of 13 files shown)


mziccard:
@ajkannan If you don't mind I will go through this tomorrow morning (my morning:))

ajkannan:
Yeah, that works fine, no rush!

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BlobValue.java
```diff
@@ -57,6 +57,32 @@ private Builder() {
     }
 
     @Override
+    public Blob get() {
```
aozarov: I think you have another get() in 1BlobValue1 itself.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BooleanValue.java
```diff
@@ -57,6 +57,32 @@ private Builder() {
     }
 
     @Override
+    public Boolean get() {
```
mziccard: Here and elsewhere (possibly not related to this PR) do you think that getters should be in the builder? I think that elsewhere in `gcloud-java` we don't have them.

This reasoning also applies to other getters: `getMeaning`, `getIndexed`, `getValueType`.

mziccard: Also, if we decide to keep getters, is it really needed to override `get`?

mziccard:
This is a general question that might stem from the fact that I haven't fully understood issue #103 but, why are we doing this?

Let me take the example of `Value.BaseBuilder`, all implementations in this PR are public static classes in public classes. Why do we need to override all setters given that `Value.BaseBuilder` is generic and setters return `B`?

If we decide that we need to override setters then are the `self()` method and the type parameter `B` still necessary?

mziccard:
To add to my previous comment: if we decide not to have generic methods inside builders (i.e. to avoid relying on inherited methods with the signature `B set(V value)` but rather have `Builder set(Boolean value)`) then we should do that for all classes across `gcloud-java`.

This involves changing a number of classes (as `ServiceOptions.Builder` for instance) and adding a bunch of boilerplate code as @ajkannan had to do in this PR. I am not sure these changes are worth it and fear they lead to a less maintainable code. Opinions are welcome:)

aozarov:
@mziccard You definitely have a point!

I think the problem is with classes/interfaces that uses generic types (in some cases to the extreme - such as `ValueMarshaller` and `ValueBuilder`) without documenting the types (so all their subclasses has an undocumented generic type in their javadoc).

In some cases we can fix that by documenting the generic types, though in other cases it may be simpler to override the method to the specific type. 

I suggest to tackle this issue using a different approach and change the specific datastore interfaces/classes in this PR to package scope (as it is not likely that new types will be created outside of the library - we can revisit that in the future).

As for `ServiceOptions` and its Builder, lets add `@param <..>` type definition to their javadoc.

ajkannan:
Changed this PR to document the types.

PR #503, opened 2015-12-22
Merge gcloud-java-bigquery into master

mziccard:
This PR adds support for BigQuery to master.
No surprises here, just commits we already reviewed before merging into the bigquery branch.

Let's just hope integration tests won't hang on load jobs!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ This client supports the following Google Cloud Platform services:
 
 -  [Google Cloud Datastore] (#google-cloud-datastore)
 -  [Google Cloud Storage] (#google-cloud-storage)
+-  [Google Cloud Resource Manager] (#google-cloud-resource-manager)
+-  [Google Cloud BigQuery] (#google-cloud-bigquery)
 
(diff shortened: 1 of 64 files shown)


mziccard:
All contributors already signed the CLA, removing label.

ajkannan:
I wonder if a rebase might get rid of the merge commits and pack all the commits together in a block in the master's history.  I did a rebase onto master for resource manager, and that seemed to work well.  If you think the work-to-reward ratio is small for switching to a rebase, we can merge as is.

mziccard:
@ajkannan I would love to do that. But I did more than one fetch from master in bigquery, removing all merges would require me to do a lot of history rewrite which is a pain in the neck and might not even lead to a very clean result.

ajkannan:
Ah I see, I'll merge as is.  Congrats, good work on BigQuery!

aozarov:
Great Job!! Thanks @mziccard 

We also need example, is there an issue for it?

PR #510, opened 2015-12-28
Speed up resource manager tests

### LINKED ISSUE #504: Speed up ResourceManager tests on Travis
Tests that use the mock resource manager helper take a relatively long time to complete on Travis.  `LocalResourceManagerHelperTest` takes ~100 seconds and `ResourceManagerImplTest` takes ~45 seconds.  When running locally, these tests only tend to take 3-5 seconds.  This is likely because the `HttpServer` library used for the mock spins up threads and also uses sockets, and Travis probably handles these operations more slowly.  We should think of a way to speed up these tests.

ajkannan:
Fixes #504


mziccard:
Execution times seem to be much more reasonable now. If you are ready we can merge!

ajkannan:
Great, I'll go ahead and merge it in then.

PR #506, opened 2015-12-23
Add BigQuery module to gcloud-java pom, add BigQuery example

mziccard:
This PR does the following:
- Make `FormatOptions.of(format)` return `CsvOptions` if `format` is `CSV`
- Some minor style fixes to `DatastoreExample` and `StorageExample`
- Adds the BigQuery module to `gcloud-java`'s pom
- Adds `BigQueryExample` and updates the READMEs accordingly

PR #512, opened 2015-12-29
Fix bigquery example and related issues

mziccard:
This PR fixes issues related to `BigQueryExample` discussed in #506.

### DIFF
--- a/gcloud-java-examples/README.md
+++ b/gcloud-java-examples/README.md
@@ -41,7 +41,17 @@ To run examples from your command line:
 
   Here's an example run of `BigQueryExample`.
 
-  Before running the example, go to the [Google Developers Console][developers-console] to ensure that BigQuery API is enabled.
+  Before running the example, go to the [Google Developers Console][developers-console] to ensure
+  that BigQuery API is enabled. You can upload a CSV file `my_csv_file` to the `my_bucket` bucket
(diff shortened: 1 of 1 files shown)


gcloud-java-examples/README.md
```diff
@@ -39,6 +39,17 @@ To run examples from your command line:
 
 4. Run an example using Maven from command line.
 
+  Here's an example run of `BigQueryExample`.
+
+  Before running the example, go to the [Google Developers Console][developers-console] to ensure that BigQuery API is enabled.
```
ajkannan: I think it'd be good to add a note here that mentions that the user should:
* create a test csv file with some strings, one string per line.
* upload the csv file Cloud Storage (using the Developers Console)

ajkannan:
Looks good, just one docs comment from me.

gcloud-java-examples/README.md
aozarov: Yes, it would be good to mention the option of uploading the file using the GCS [web browser](https://pantheon.corp.google.com/storage/browser).

As for the direction for the CSV I wonder if we could actually include a sample file?

mziccard:
Added more details on CSV file in 6af6db4

ajkannan:
Looks good!

PR #516, opened 2015-12-30
Add functional classes for BigQuery datasets jobs and tables

### LINKED ISSUE #490: Adding functional Dataset, Table and Job to bigquery
We should add functional classes (`Dataset`, `Table` and `Job`) to gcloud-java-bigquery, to make using the library easier. Functional classes should allow something like:
```java
// get an existing dataset
Dataset dataset = Dataset.load(bigquery, "my_dataset_id");
// create a table
Table table = dataset.create("my_table_id", Schema.of(field1, field2));
// copy a table
Job copyJob = table.copy(TableId.of("my_dataset_id", "my_copy_table_id");
// cancel a job
copyJob.cancel();
// load data into the table
Job loadJob = table.load(FormatOptions.csv(), ImmutableList.of("gs://my-bucket/my-file.csv"));
// wait for job to complete
while (!loadJob.isDone()) {
  loadJob = loadJob.reload();
  Thread.sleep(1000);
}
```

Comments are welcome:)
/cc @jtigani

mziccard:
This PR adds functional classes to BigQuery and fixes #490 
- Add `Dataset`, `Job` and `Table` functional classes
- Add unit tests

### DIFF
--- a/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
+++ b/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
@@ -33,9 +33,12 @@ import com.google.gcloud.PageImpl;
 
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import java.util.Iterator;
+import java.util.List;
 
 public class DatasetTest {
 
@@ -48,7 +51,12 @@ public class DatasetTest {
       ExternalTableInfo.builder(TableId.of("dataset", "table2"),
           ExternalDataConfiguration.of(ImmutableList.of("URI"), Schema.of(), FormatOptions.csv()))
           .build());
+  private static final UserDefinedFunction FUNCTION1 = UserDefinedFunction.inline("inline");
+  private static final UserDefinedFunction FUNCTION2 = UserDefinedFunction.inline("gs://b/f");
+  private static final List<UserDefinedFunction> FUNCTIONS = ImmutableList.of(FUNCTION1, FUNCTION2);
 
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
   private BigQuery bigquery;
   private Dataset dataset;
 
@@ -66,6 +74,12 @@ public class DatasetTest {
   @Test
   public void testInfo() throws Exception {
     assertEquals(DATASET_INFO, dataset.info());
+    replay(bigquery);
+  }
+
+  @Test
+  public void testBigQuery() throws Exception {
+    assertSame(bigquery, dataset.bigquery());
     replay(bigquery);
   }
 
@@ -119,7 +133,28 @@ public class DatasetTest {
     expect(bigquery.update(updatedInfo)).andReturn(updatedInfo);
     replay(bigquery);
     Dataset updatedDataset = dataset.update(updatedInfo);
-    assertSame(bigquery, dataset.bigquery());
+    assertSame(bigquery, updatedDataset.bigquery());
+    assertEquals(updatedInfo, updatedDataset.info());
+  }
+
+  @Test
+  public void testUpdateWithDifferentId() throws Exception {
+    DatasetInfo updatedInfo = DATASET_INFO.toBuilder()
+        .datasetId(DatasetId.of("dataset2"))
+        .description("Description")
+        .build();
+    replay(bigquery);
+    thrown.expect(IllegalArgumentException.class);
+    dataset.update(updatedInfo);
+  }
+
+  @Test
+  public void testUpdateWithOptions() throws Exception {
+    DatasetInfo updatedInfo = DATASET_INFO.toBuilder().description("Description").build();
+    expect(bigquery.update(updatedInfo, BigQuery.DatasetOption.fields())).andReturn(updatedInfo);
+    replay(bigquery);
+    Dataset updatedDataset = dataset.update(updatedInfo, BigQuery.DatasetOption.fields());
+    assertSame(bigquery, updatedDataset.bigquery());
     assertEquals(updatedInfo, updatedDataset.info());
   }
 
@@ -151,6 +186,27 @@ public class DatasetTest {
   }
 
   @Test
+  public void testListWithOptions() throws Exception {
+    BigQueryOptions bigqueryOptions = createStrictMock(BigQueryOptions.class);
+    PageImpl<BaseTableInfo> tableInfoPage = new PageImpl<>(null, "c", TABLE_INFO_RESULTS);
+    expect(bigquery.listTables(DATASET_INFO.datasetId(), BigQuery.TableListOption.maxResults(10L)))
+        .andReturn(tableInfoPage);
+    expect(bigquery.options()).andReturn(bigqueryOptions);
+    expect(bigqueryOptions.service()).andReturn(bigquery);
+    replay(bigquery, bigqueryOptions);
+    Page<Table> tablePage = dataset.list(BigQuery.TableListOption.maxResults(10L));
+    Iterator<BaseTableInfo> tableInfoIterator = tableInfoPage.values().iterator();
+    Iterator<Table> tableIterator = tablePage.values().iterator();
+    while (tableInfoIterator.hasNext() && tableIterator.hasNext()) {
+      assertEquals(tableInfoIterator.next(), tableIterator.next().info());
+    }
+    assertFalse(tableInfoIterator.hasNext());
+    assertFalse(tableIterator.hasNext());
+    assertEquals(tableInfoPage.nextPageCursor(), tablePage.nextPageCursor());
+    verify(bigqueryOptions);
+  }
+
+  @Test
   public void testGet() throws Exception {
     BaseTableInfo info = TableInfo.builder(TableId.of("dataset", "table1"), Schema.of()).build();
     expect(bigquery.getTable(TableId.of("dataset", "table1"))).andReturn(info);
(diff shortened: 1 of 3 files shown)


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  /**
+   * Fetches current dataset's latest information.
+   *
+   * @param options dataset options
+   * @return a {@code Dataset} object with latest information.
+   * @throws BigQueryException upon failure
+   */
+  public Dataset reload(BigQuery.DatasetOption... options) {
```
aozarov: Should this delegate to `load`? Any reason `load` does not accept options?
What should this method do if bigquery.getDataset returns `null`?

aozarov: Just noticed that `Storage` has the same issue and that `ResourceManager` does not
check for `null` in `Project.load` (and therefore does not documented it - /cc @ajkannan).

aozarov: Once fixed we should also include tests that call load, delete and then reload in that order.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -0,0 +1,285 @@
...
+    this.info = checkNotNull(info);
+  }
+
+  /**
+   * Creates a {@code Dataset} object for the provided dataset's user-defined id. Performs an RPC
+   * call to get the latest dataset information.
+   *
+   * @param bigquery the BigQuery service used for issuing requests
+   * @param dataset dataset's user-defined id
+   * @return the {@code Dataset} object or {@code null} if not found.
```
aozarov: Nit, though consistent I wonder where it comes from (the '.' for `@return`).
The [Javastyle](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@return) guide suggests it should be treated like the `@param` tag.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Job.java
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Fetches current job's latest information.
+   *
+   * @param options job options
+   * @return a {@code Job} object with latest information.
+   * @throws BigQueryException upon failure
+   */
+  public Job reload(BigQuery.JobOption... options) {
+    return new Job(bigquery, bigquery.getJob(info.jobId(), options));
```
aozarov: Same issue of a possible `null`.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Job.java
```diff
@@ -0,0 +1,117 @@
...
+   * @throws BigQueryException upon failure
+   */
+  public boolean exists() {
+    return bigquery.getJob(info.jobId(), BigQuery.JobOption.fields()) != null;
+  }
+
+  /**
+   * Checks if this job has completed its execution, either failing or succeeding.
+   *
+   * @return {@code true} if this job is in {@link JobStatus.State#DONE} state, {@code false} if the
```
aozarov: I think this is can lead to a bad code that loop forever due to a  `while(!job.isDone())` when job
no longer exists (delete or never existed). The very least we should make that clear in the javadoc
and suggest using `while(job.exists() && !job.isDone())`...

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
mziccard: > Should this delegate to load? Any reason load does not accept options?
What should this method do if bigquery.getDataset returns null?

Yes we should delegate this to `load` (also in `Storage`) and add options to `load`. If we do this the method will return `null` if not found (which is reasonable to me) but should be documented.

mziccard: > Once fixed we should also include tests that call load, delete and then reload in that order.

You mean integration or unit tests? I am not sure that adding integration tests for functional classes is really necessary given that we have integration tests for the service

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
mziccard: You are right, I think we added them for consistency but we should remove them (I opened #519 and will take care of removing the periods from this PR).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Job.java
mziccard: A job can not be deleted. It can be cancelled but the job's resource will still be returned with `state = DONE`. I agree that we can suggest to use `while(job.exists() && !job.isDone())`.

mziccard:
d75f1f4 fixed issues with bigquery while cbe2634 fixes storage.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+  @Test
+  public void testGetNull() throws Exception {
+    expect(bigquery.getTable(TableId.of("dataset", "table1"))).andReturn(null);
+    replay(bigquery);
+    assertNull(dataset.get("table1"));
+  }
+
+  @Test
+  public void testCreateTable() throws Exception {
```
ajkannan: Do you think we should test the `create` functions that have `UserDefinedFunction` in their signature as well?  We could also add a `TableOption` as well.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
```diff
@@ -0,0 +1,226 @@
...
+    expect(bigquery.getDataset(DATASET_ID.dataset(), BigQuery.DatasetOption.fields()))
+        .andReturn(updatedInfo);
+    replay(bigquery);
+    Dataset updatedDataset = dataset.reload(BigQuery.DatasetOption.fields());
+    assertSame(bigquery, updatedDataset.bigquery());
+    assertEquals(updatedInfo, updatedDataset.info());
+  }
+
+  @Test
+  public void testUpdate() throws Exception {
```
ajkannan: Should we add a test where we try to call `update` with a `DatasetInfo` that has a different ID?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
```diff
@@ -0,0 +1,226 @@
...
+    dataset = new Dataset(bigquery, DATASET_INFO);
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    verify(bigquery);
+  }
+
+  @Test
+  public void testInfo() throws Exception {
```
ajkannan: Nit: if we have a test for this, should we also have a test for the function `bigquery()`?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+  @Test
+  public void testLoadNull() throws Exception {
+    expect(bigquery.getDataset(DATASET_INFO.datasetId().dataset())).andReturn(null);
+    replay(bigquery);
+    assertNull(Dataset.load(bigquery, DATASET_INFO.datasetId().dataset()));
+  }
+
+  @Test
+  public void testLoadWithOptions() throws Exception {
```
ajkannan: For this test (and similar tests in `JobTest` and `TableTest`), I believe this test would still pass if the options were completely ignored.  Maybe there should be another `DatasetInfo` with an optional extra field that gets stripped away?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Table.java
```diff
@@ -0,0 +1,279 @@
...
+   * Updates the table's information. Dataset's and table's user-defined ids cannot be changed. A
+   * new {@code Table} object is returned.
+   *
+   * @param tableInfo new table's information. Dataset's and table's user-defined ids must match the
+   *     ones of the current table
+   * @param options dataset options
+   * @return a {@code Table} object with updated information
+   * @throws BigQueryException upon failure
+   */
+  public Table update(BaseTableInfo tableInfo, BigQuery.TableOption... options) {
```
ajkannan: I think this function doesn't have an associated test.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
ajkannan: Actually, ignore this comment.  I missed that the `expect` checks that the options are specified.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -156,16 +156,18 @@ public Blob(Storage storage, BlobInfo info) {
 
   /**
    * Creates a {@code Blob} object for the provided bucket and blob names. Performs an RPC call to
-   * get the latest blob information.
+   * get the latest blob information. Returns {@code null} if the blob does not exist.
```
ajkannan: Nit: we should probably standardize this javadoc and the javadoc for load below (either include the "Returns {@code null}..." sentence here and in the second `load` function, or just leave it in the `@return` statement like in the second `load` function.

ajkannan:
Looks good, just a couple test/docs comments from me.

mziccard:
@ajkannan thanks for the suggestions, I fixed the one on storage in place and I added d58930e to catch all comments on bigquery tests.

ajkannan:
LGTM

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -0,0 +1,287 @@
...
+          nextInfoPage.nextPageCursor(), new LazyTableIterable(options, nextInfoPage.values()));
+    }
+  }
+
+  private static class LazyTableIterable implements Iterable<Table>, Serializable {
+
+    private static final long serialVersionUID = 3312744215731674032L;
+
+    private final BigQueryOptions options;
+    private Iterable<BaseTableInfo> infoIterable;
```
aozarov: Nit, could be final.

aozarov:
Looks good. one minor comment + few other unrelated:

Typo in `ITBigQueryTest` -> Udpate -> Update
self reference in Javadoc for `QueryJobInfo.Priority` enum values.

PR #517, opened 2015-12-30
Refactor BasePageFetcher

### LINKED ISSUE #485: Refactor BasePageFetcher
Some page-based cursor services, such as `gcloud-java-storage` and `gcloud-java-resourcemanager`, currently have duplicated code for the `BasePageFetcher`.  We should refactor the `BasePageFetcher` code to be in `gcloud-java-core`.  We can use Generics for the specific Options type O and provide an abstract method to return the PAGE_TOKEN option.

ajkannan:
Fixes #485

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
@@ -38,6 +38,7 @@ import com.google.gcloud.ExceptionHandler;
 import com.google.gcloud.ExceptionHandler.Interceptor;
 import com.google.gcloud.Page;
 import com.google.gcloud.PageImpl;
+import com.google.gcloud.PageImpl.NextPageFetcher;
 import com.google.gcloud.RetryHelper;
 import com.google.gcloud.bigquery.InsertAllRequest.RowToInsert;
 import com.google.gcloud.spi.BigQueryRpc;
@@ -69,36 +70,17 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
   static final ExceptionHandler EXCEPTION_HANDLER = ExceptionHandler.builder()
       .abortOn(RuntimeException.class).interceptor(EXCEPTION_HANDLER_INTERCEPTOR).build();
 
-  private abstract static class BasePageFetcher<T> implements PageImpl.NextPageFetcher<T> {
+  private static class DatasetPageFetcher implements NextPageFetcher<DatasetInfo> {
 
-    private static final long serialVersionUID = -338124488600215401L;
-
-    protected final Map<BigQueryRpc.Option, ?> requestOptions;
-    protected final BigQueryOptions serviceOptions;
-
-    BasePageFetcher(BigQueryOptions serviceOptions, String cursor,
-        Map<BigQueryRpc.Option, ?> optionMap) {
-      this.serviceOptions = serviceOptions;
-      ImmutableMap.Builder<BigQueryRpc.Option, Object> builder = ImmutableMap.builder();
-      if (cursor != null) {
-        builder.put(BigQueryRpc.Option.PAGE_TOKEN, cursor);
-      }
-      for (Map.Entry<BigQueryRpc.Option, ?> option : optionMap.entrySet()) {
-        if (option.getKey() != BigQueryRpc.Option.PAGE_TOKEN) {
-          builder.put(option.getKey(), option.getValue());
-        }
-      }
-      this.requestOptions = builder.build();
-    }
-  }
-
-  private static class DatasetPageFetcher extends BasePageFetcher<DatasetInfo> {
-
-    private static final long serialVersionUID = 3030824397616608646L;
+    private static final long serialVersionUID = -3057564042439021278L;
+    private final Map<BigQueryRpc.Option, ?> requestOptions;
+    private final BigQueryOptions serviceOptions;
 
     DatasetPageFetcher(BigQueryOptions serviceOptions, String cursor,
         Map<BigQueryRpc.Option, ?> optionMap) {
-      super(serviceOptions, cursor, optionMap);
+      this.requestOptions =
+          PageImpl.nextRequestOptions(BigQueryRpc.Option.PAGE_TOKEN, cursor, optionMap);
+      this.serviceOptions = serviceOptions;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
```diff
@@ -38,6 +40,38 @@
     Page<T> nextPage();
   }
 
+  /**
+   * Base class for page fetchers.
+   *
+   * @param <T> the value type that the page holds. {@code T} should implement
+   *     {@code java.io.Serializable}.
```
mziccard: Nit: remove "period" at the end, here and in the following `@param`

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
```diff
@@ -38,6 +40,38 @@
...
+   * @param <ServiceOptionsT> the service options type.
+   */
+  @SuppressWarnings("rawtypes")
```
mziccard: Just wondering, is this necessary?

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
```diff
@@ -38,6 +40,38 @@
...
+  public abstract static class BasePageFetcher<T, ServiceOptionsT extends ServiceOptions>
+      implements NextPageFetcher<T> {
+
+    private static final long serialVersionUID = -3333056232729661990L;
+    protected final Map<?, ?> requestOptions;
+    protected final ServiceOptionsT serviceOptions;
+
+    protected BasePageFetcher(ServiceOptionsT serviceOptions, Enum<?> pageTokenOption,
+        String cursor,
+        Map<?, ?> optionMap) {
```
mziccard: I am not entitled to give style suggestions:) But why newline?

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
```diff
@@ -38,6 +40,38 @@
...
 
+  /**
+   * Base class for page fetchers.
+   *
+   * @param <T> the value type that the page holds. {@code T} should implement
+   *     {@code java.io.Serializable}.
+   * @param <ServiceOptionsT> the service options type.
+   */
+  @SuppressWarnings("rawtypes")
+  public abstract static class BasePageFetcher<T, ServiceOptionsT extends ServiceOptions>
```
mziccard: We should keep naming consistent, that is using either `T` and `O` or `ItemT` (just guessing) and `ServiceOptionsT`.

mziccard:
@ajkannan great job, I just added some comments!

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
ajkannan: Eclipse gives me a warning that `ServiceOptions` (in the method signature) is a raw type, so I added it in.

ajkannan:
Thanks for taking a look @mziccard!  I made the fixes in place.

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
mziccard: I might be wrong so please correct me but if I remove the annotation (also from `ServiceFactory`) and compile with java 7 I don't get any warning.

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
```diff
@@ -38,6 +40,37 @@
...
   }
 
+  /**
+   * Base class for page fetchers.
+   *
+   * @param <T> the value type that the page holds. {@code T} should implement
+   *     {@code java.io.Serializable}
+   * @param <O> the service options type
+   */
+  @SuppressWarnings("rawtypes")
```
aozarov: is that needed?

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
```diff
@@ -38,6 +40,37 @@
...
+  public abstract static class BasePageFetcher<T, O extends ServiceOptions>
+      implements NextPageFetcher<T> {
+
+    private static final long serialVersionUID = -3333056232729661990L;
+    protected final Map<?, ?> requestOptions;
+    protected final O serviceOptions;
+
+    protected BasePageFetcher(O serviceOptions, Enum<?> pageTokenOption,
```
aozarov: Fine as is, though to make it more generic we could make `pageTokenOption` an `Object` and compare bellow using `Objects.equals`

gcloud-java-core/src/main/java/com/google/gcloud/PageImpl.java
aozarov: Oh, I guess that is needed because` O` extends the rawtype for `ServiceOptions`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
```diff
@@ -69,108 +70,93 @@ public RetryResult beforeEval(Exception exception) {
...
-    private static final long serialVersionUID = 3030824397616608646L;
+    private static final long serialVersionUID = 8049465230453759547L;
 
     DatasetPageFetcher(BigQueryOptions serviceOptions, String cursor,
         Map<BigQueryRpc.Option, ?> optionMap) {
-      super(serviceOptions, cursor, optionMap);
+      super(serviceOptions, BigQueryRpc.Option.PAGE_TOKEN, cursor, optionMap);
     }
 
+    @SuppressWarnings("unchecked")
```
aozarov: I wonder if instead of implementing the common code in the base class `BasePageFetcher` (and suppressing its rawtype warning) and then needing to cast and suppress warning in each subclass we would not be better of with putting the common logic in a static method.
```java
  /**
   * Utility method to construct the next options map.
   *
   * @param <T> the value type that the page holds. Instances of {@code T} should be
   *     {@code Serializable}
   * @param pageTokenOption the key for the
   * @param cursor the cursor for the next page
   * @param optionMap the previous options map
   */
  public static <T> Map<T, Object> nextRequestOptions(T pageTokenOption, String cursor,
      Map<T, ?> optionMap) {
    ImmutableMap.Builder<T, Object> builder = ImmutableMap.builder();
    if (cursor != null) {
      builder.put(pageTokenOption, cursor);
    }
    for (Map.Entry<T, ?> option : optionMap.entrySet()) {
      if (!Objects.equals(option.getKey(), pageTokenOption)) {
        builder.put(option.getKey(), option.getValue());
      }
    }
    return builder.build();
  }
```

and In the implementations of `NextPageFetcher` constructors we would instead call that static method to get the next requestOptions (and keep a reference to it as well as to the specific service options which was not needed by the common code).

e.g.
```java
  private static class ProjectPageFetcher implements NextPageFetcher<ProjectInfo> {

    private static final long serialVersionUID = -533306655445189098L;
    private final Map<ResourceManagerRpc.Option, ?> requestOptions;
    private final ResourceManagerOptions serviceOptions;

    ProjectPageFetcher(ResourceManagerOptions serviceOptions, String cursor,
        Map<ResourceManagerRpc.Option, ?> optionMap) {
      this.requestOptions =
          PageImpl.nextRequestOptions(ResourceManagerRpc.Option.PAGE_TOKEN, cursor, optionMap);
      this.serviceOptions = serviceOptions;
    }

    @Override
    public Page<ProjectInfo> nextPage() {
      return listProjects(serviceOptions, requestOptions);
    }
  }
```

ajkannan:
Added a commit to use the static method as @aozarov suggested.

PR #520, opened 2016-01-04
Check for null in Project.load()

ajkannan:
From discussion in #516

### DIFF
--- a/gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/ProjectTest.java
+++ b/gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/ProjectTest.java
@@ -81,6 +81,13 @@ public class ProjectTest {
   }
 
   @Test
+  public void testLoadNull() {
+    expect(resourceManager.get(PROJECT_INFO.projectId())).andReturn(null);
+    replay(resourceManager);
+    assertNull(Project.load(resourceManager, PROJECT_INFO.projectId()));
+  }
+
+  @Test
   public void testReloadDeletedProject() {
     expect(resourceManager.get(PROJECT_INFO.projectId())).andReturn(PROJECT_INFO);
     expect(resourceManager.get(PROJECT_INFO.projectId())).andReturn(null);


gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
```diff
@@ -42,12 +42,13 @@ public Project(ResourceManager resourceManager, ProjectInfo projectInfo) {
   /**
    * Constructs a Project object that contains project information loaded from the server.
    *
-   * @return Project object containing the project's metadata
+   * @return Project object containing the project's metadata or
+   *     null if the project doesn't exist
```
mziccard: Nit, couldn't null fit on the previous line? Also we better use `{@code null}`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
```diff
@@ -67,7 +68,8 @@ public ResourceManager resourceManager() {
   /**
    * Returns a Project object with updated project information.
    *
-   * @return Project object containing the project's updated metadata
+   * @return Project object containing the project's updated metadata or null if the project no
+   *     longer exists
```
mziccard: `null` could also be returned if the user feeds to the Project class infos of a non-existing project, this is unlikely to happen but I was thinking about phrasing this as: `or {@code null} if not found`

ajkannan:
@mziccard fixed in place

mziccard:
@ajkannan What do you think about adding tests for `Project`'s `load` and `reload` for the cases when the service returns `null`? Something like:
```java
@Test
public void testLoadNull() {
  expect(resourceManager.get(PROJECT_INFO.projectId())).andReturn(null);
  replay(resourceManager);
  assertNull(Project.load(resourceManager, PROJECT_INFO.projectId()));
}
```

ajkannan:
I added a test for `reload` already, and I think that covers `load` since `reload` uses `load`.  But I can still add a test like the one you listed in case `reload` doesn't use `load` in the future.

ajkannan:
@mziccard Added the `testLoadNull` test

mziccard:
Looks good to me!

Issue #515, opened 2015-12-30
testInsertAllWithSuffix is failing


ajkannan:
The BigQuery integration test for `insertAllWithSuffix` gives the following error in Travis CI: 
```
Running com.google.gcloud.bigquery.ITBigQueryTest
Tests run: 24, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 101.531 sec <<< FAILURE! - in com.google.gcloud.bigquery.ITBigQueryTest
testInsertAllWithSuffix(com.google.gcloud.bigquery.ITBigQueryTest)  Time elapsed: 0.418 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertNotNull(Assert.java:712)
	at org.junit.Assert.assertNotNull(Assert.java:722)
	at com.google.gcloud.bigquery.ITBigQueryTest.testInsertAllWithSuffix(ITBigQueryTest.java:508)
Results :
Failed tests: 
  ITBigQueryTest.testInsertAllWithSuffix:508 null
Tests run: 24, Failures: 1, Errors: 0, Skipped: 0
```

aozarov:
Looks like this is a flaky test.
I managed to get it pass and fail (using different account and on my desktop).

@jtigani Is there an expected delay between the automatic creation of suffixed table and to be able to get information about it?

mziccard:
Mmmm this is unpleasant. Sorry for pushing it, I did not experience the error locally, I run the test 10 times now and it failed eventually. Should we remove the integration test while we clarify this?

aozarov:
Did you manage to understand why it failed? Did it fail due to some delay or a failure to insert?
Yes, we should make the test pass (by adding a delay or a retry if helps and if not then disable it for now.)

mziccard:
I checked and it seems that eventually the table is created. My guess is that this happen with some delay. I could not find any reference to this delay in the docs. I will put the failing code inside a loop+delay and that should do it until we figure out what is going on.

jcondit:
Will summon the appropriate experts...

seanchen-google:
In general table creation should happen O(seconds) after the first insert, but occasionally transient errors will result in O(minutes) delay.

If you're consistently seeing more than a few seconds of delay, then there's likely a bug somewhere. Do you have any statistics on what you're seeing?

[...]

PR #521, opened 2016-01-04
Wait for table to be created in testInsertAllWithSuffix

mziccard:
This is a workaround for #515. This PR adds a loop+delay to check for the creation of a suffix table. If the table is never created the test will eventually time-out. 

I suggest we keep #515 open until we figure out if this delay is expected or not (and document our library accordingly).

### DIFF
--- a/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/ITBigQueryTest.java
+++ b/gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/ITBigQueryTest.java
@@ -476,7 +476,7 @@ public class ITBigQueryTest {
   }
 
   @Test
-  public void testInsertAllWithSuffix() {
+  public void testInsertAllWithSuffix() throws InterruptedException {
     String tableName = "test_insert_all_with_suffix_table";
     BaseTableInfo tableInfo = TableInfo.of(TableId.of(DATASET, tableName), TABLE_SCHEMA);
     assertNotNull(bigquery.create(tableInfo));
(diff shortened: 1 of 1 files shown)


gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/ITBigQueryTest.java
```diff
@@ -505,7 +505,12 @@ public void testInsertAllWithSuffix() {
     assertFalse(response.hasErrors());
     assertEquals(0, response.insertErrors().size());
     String newTableName = tableName + "_suffix";
-    assertNotNull(bigquery.getTable(DATASET, newTableName, TableOption.fields()));
+    BaseTableInfo suffixTable = bigquery.getTable(DATASET, newTableName, TableOption.fields());
+    while (suffixTable == null) {
+      Thread.sleep(1000L);
+      suffixTable = bigquery.getTable(DATASET, newTableName, TableOption.fields());
+    }
+    assertNotNull(suffixTable);
```
ajkannan: Nit: this line is made unnecessary by the loop

ajkannan:
Looks good.  One nit comment, and feel free to merge after that's fixed.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/ITBigQueryTest.java
mziccard: I left it just to remember what is the expected result if we ever remove the loop. If you think redundancy is evil (which in general is) I am happy to remove.

ajkannan: Sounds reasonable

aozarov: I vote to remove it and a comment to explain why we loop would be nice.

PR #523, opened 2016-01-05
Add support for BigQuery query plan explanation

### LINKED ISSUE #465: Add support for BigQuery query plan explanation
Based on the recent [BigQuery update](http://googlecloudplatform.blogspot.com/2015/12/BigQuery-cost-controls-now-let-you-set-a-daily-maximum-for-query-costs.html) we should add support for query plan explanation. Plan explanation is contained in the [queryPlan field](https://cloud.google.com/bigquery/docs/reference/v2/jobs#statistics.query.queryPlan) inside query job's statistics.

mziccard:
This PR adds support for [Query Plan Explanation](https://cloud.google.com/bigquery/query-plan-explanation) to job's statistics and fixes #465

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobStatistics.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobStatistics.java
@@ -339,10 +339,10 @@ public class JobStatistics implements Serializable {
     }
 
     /**
-     * Returns the query plan as a list of stages. Each stage involves a number of steps that read
-     * from data sources, perform a series of transformations on the input, and emit an output to a
-     * future stage (or the final result). The query plan is available for a completed query job and
-     * is retained for 7 days.
+     * Returns the query plan as a list of stages or {@code null} if a query plan is not available.
+     * Each stage involves a number of steps that read from data sources, perform a series of
+     * transformations on the input, and emit an output to a future stage (or the final result). The
+     * query plan is available for a completed query job and is retained for 7 days.
      *
      * @see <a href="https://cloud.google.com/bigquery/query-plan-explanation">Query Plan</a>
      */
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
@@ -78,18 +78,27 @@ public class QueryStage implements Serializable {
         };
     private static final long serialVersionUID = 8663444604771794411L;
 
-    private final String kind;
+    private final String name;
     private final List<String> substeps;
 
-    QueryStep(String kind, List<String> substeps) {
-      this.kind = kind;
+    QueryStep(String name, List<String> substeps) {
+      this.name = name;
       this.substeps = substeps;
     }
 
-    public String kind() {
-      return kind;
+    /**
+     * Returns a machine-readable name for the operation.
+     *
+     * @see <a href="https://cloud.google.com/bigquery/query-plan-explanation#steps_metadata">Steps
+     *     Metadata</a>
+     */
+    public String name() {
+      return name;
     }
 
+    /**
+     * Returns a list of human-readable stage descriptions.
+     */
     public List<String> substeps() {
       return substeps;
     }
@@ -97,14 +106,14 @@ public class QueryStage implements Serializable {
     @Override
     public String toString() {
       return MoreObjects.toStringHelper(this)
-          .add("kind", kind)
+          .add("name", name)
           .add("substeps", substeps)
           .toString();
     }
 
     @Override
     public int hashCode() {
-      return Objects.hash(kind, substeps);
+      return Objects.hash(name, substeps);
     }
 
     @Override
@@ -113,20 +122,16 @@ public class QueryStage implements Serializable {
         return false;
       }
       QueryStep other = (QueryStep) obj;
-      return Objects.equals(kind, other.kind)
-          && Objects.equals(substeps, other.substeps);
+      return Objects.equals(name, other.name) && Objects.equals(substeps, other.substeps);
     }
 
     ExplainQueryStep toPb() {
-      return new ExplainQueryStep().setKind(kind).setSubsteps(substeps);
+      return new ExplainQueryStep().setKind(name).setSubsteps(substeps);
     }
 
     static QueryStep fromPb(com.google.api.services.bigquery.model.ExplainQueryStep stepPb) {
-      List<String> substeps = null;
-      if (stepPb.getSubsteps() != null) {
-        substeps = ImmutableList.copyOf(stepPb.getSubsteps());
-      }
-      return new QueryStep(stepPb.getKind(), substeps);
+      return new QueryStep(stepPb.getKind(), ImmutableList.copyOf(stepPb.getSubsteps() != null
+          ? stepPb.getSubsteps() : ImmutableList.<String>of()));
     }
   }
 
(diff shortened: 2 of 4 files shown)


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobStatistics.java
```diff
@@ -325,13 +338,26 @@ public Long totalBytesProcessed() {
...
+    /**
+     * Returns the query plan as a list of stages. Each stage involves a number of steps that read
+     * from data sources, perform a series of transformations on the input, and emit an output to a
+     * future stage (or the final result). The query plan is available for a completed query job and
+     * is retained for 7 days.
+     *
+     * @see <a href="https://cloud.google.com/bigquery/query-plan-explanation">Query Plan</a>
+     */
+    public List<QueryStage> queryPlan() {
+      return queryPlan;
```
aozarov: Should we return `null` when none is available and if so, should that documented?

aozarov: Should we mentioned this method in `QueryJobInfo.dryRun`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
```diff
@@ -0,0 +1,431 @@
...
+    recordsWritten = builder.recordsWritten;
+    steps = builder.steps;
+    waitRatioAvg = builder.waitRatioAvg;
+    waitRatioMax = builder.waitRatioMax;
+    writeRatioAvg = builder.writeRatioAvg;
+    writeRatioMax = builder.writeRatioMax;
+  }
+
+  /**
+   * Returns the relative amount of time the average shard spent on CPU-bound tasks.
```
aozarov: Nit, I am not sure if adding "relative" to each description is actually adding clarity.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
```diff
@@ -0,0 +1,431 @@
...
+
+    private final String kind;
+    private final List<String> substeps;
+
+    QueryStep(String kind, List<String> substeps) {
+      this.kind = kind;
+      this.substeps = substeps;
+    }
+
+    public String kind() {
```
aozarov: Can we come with a better name? Maybe "step" or "name"?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
```diff
@@ -0,0 +1,431 @@
...
+          public ExplainQueryStep apply(QueryStep stage) {
+            return stage.toPb();
+          }
+        };
+    private static final long serialVersionUID = 8663444604771794411L;
+
+    private final String kind;
+    private final List<String> substeps;
+
+    QueryStep(String kind, List<String> substeps) {
```
aozarov: How would QueryStep be mapped when the output is the one mentioned at https://cloud.google.com/bigquery/query-plan-explanation#the_bigquery_api_and_query_plan_explanation?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
```diff
@@ -0,0 +1,431 @@
...
+      this.kind = kind;
+      this.substeps = substeps;
+    }
+
+    public String kind() {
+      return kind;
+    }
+
+    public List<String> substeps() {
+      return substeps;
```
aozarov: Should we return empty list for `null`?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
```diff
@@ -0,0 +1,431 @@
...
+
+  /**
+   * Returns the relative amount of time the slowest shard spent reading input.
+   */
+  public double readRatioMax() {
+    return readRatioMax;
+  }
+
+  /**
+   * Returns the number of records read into the stage.
```
aozarov: Nit, I think "read by" sounds better than "read into"

aozarov:
Nice! one question and few nits and suggestions.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
mziccard: My guess is that that example is not well-formed. In the [Job Resource Docs](https://cloud.google.com/bigquery/docs/reference/v2/jobs#statistics.query.queryPlan.steps) there is no mention of `field.description`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
mziccard: I like `name` better

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
mziccard: Better return an empty list in this case.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
mziccard: Javadoc is taken from [here](https://cloud.google.com/bigquery/docs/reference/v2/jobs#resource). In my understanding "relative" is meant to suggest that the user should not expect absolute execution time values. If you think this creates confusion I don't mind removing it.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobStatistics.java
mziccard: > Should we return null when none is available and if so, should that documented?

We do return `null` if not available, I'll add documentation for this

> Should we mentioned this method in QueryJobInfo.dryRun?

Why do you think we should mention this in `QueryJobInfo.dryRun`? When the job is dry run no `queryPlan` is available.

aozarov: Oh, that sounds strange to me. Indeed for dry run not all values may be available (especially time related - if can't be estimated) however I thought that could be helpful to estimate how expensive a query might be and how it would be performed (similar to the SQL explain plan which does not really run the query, hence "dry run").

/cc @tigani

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
aozarov: Ah, the descriptions in https://cloud.google.com/bigquery/query-plan-explanation#available_metadata seemed better to me. Feel free to change or to keep
based on your preference.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobStatistics.java
mziccard: I checked earlier today (as you can see also in #525) but I might still be wrong. Maybe @jcondit can help us.

jcondit: Query plan is not expected to be available via dry run right now. We may add it in the future, but you should definitely handle the case where it's absent.

mziccard: @jcondit thanks for the clarification!

PR #526, opened 2016-01-06
Add BigQueryError field to BigQueryException and tests

### LINKED ISSUE #444: Tie BigQueryError to BigQueryException
`BigQueryError`s are returned as part of a job state, as the result of a [query operation](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query) and as a response to [insert all](https://cloud.google.com/bigquery/docs/reference/v2/tabledata/insertAll). In all the above cases a `BigQueryException` is not thrown.

When a request fails the service returns an error in the form:
```json
{
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "notFound",
    "message": "Not found: Table gcloud-devel:dataset.view"
   }
  ],
  "code": 404,
  "message": "Not found: Table gcloud-devel:dataset.view"
 }
}
```
So we could add a `BigQueryError` field to `BigQueryException` containing the error that caused the exception.

mziccard:
This PR ties BigQueryError to BigQueryException and adds some integration tests (fixes #444).

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryException.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryException.java
@@ -31,8 +31,23 @@ public class BigQueryException extends BaseServiceException {
   private static final long serialVersionUID = -5504832700512784654L;
   public static final int UNKNOWN_CODE = -1;
 
+  private final BigQueryError error;
+
   public BigQueryException(int code, String message, boolean retryable) {
+    this(code, message, retryable, null);
+  }
+
+  public BigQueryException(int code, String message, boolean retryable, BigQueryError error) {
     super(code, message, retryable);
+    this.error = error;
+  }
+
+  /**
+   * Returns the {@link BigQueryError} that caused this exception. Returns {@code null} if none
+   * exists.
+   */
+  public BigQueryError error() {
+    return error;
   }
 
   /**
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
@@ -49,6 +49,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 
+import com.google.gcloud.bigquery.BigQueryError;
 import com.google.gcloud.bigquery.BigQueryException;
 import com.google.gcloud.bigquery.BigQueryOptions;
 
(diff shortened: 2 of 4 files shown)


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryException.java
```diff
@@ -31,10 +31,25 @@
   private static final long serialVersionUID = -5504832700512784654L;
   public static final int UNKNOWN_CODE = -1;
 
+  private BigQueryError error;
```
ajkannan: nit: could this be made final?

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
```diff
@@ -91,7 +92,14 @@ private static BigQueryException translate(IOException exception) {
 
   private static BigQueryException translate(GoogleJsonError exception) {
     boolean retryable = RETRYABLE_CODES.contains(exception.getCode());
-    return new BigQueryException(exception.getCode(), exception.getMessage(), retryable);
+    BigQueryError bigqueryError = null;
+    if (exception.getErrors() != null  && !exception.getErrors().isEmpty()) {
+      GoogleJsonError.ErrorInfo error = exception.getErrors().get(0);
```
ajkannan: Do you know whether it's possible to get multiple errors in the same `GoogleJsonError` (and if so, would it be helpful to return a list of errors)?  I realize I also only [take the first error](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/spi/DefaultResourceManagerRpc.java#L60) in Resource Manager as well

ajkannan:
Looks good, one nit and one question from me.

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
mziccard: And if I am not mistaken this is also the case in Datastore.
But no, I don't know whether there are some cases in which the `errors` array contains more than one value. Maybe @aozarov has more information than I do?

aozarov: I think the api client had it as a list for flexibility but so far I am not aware of services using that and prefered
to keep the API simple in that case.

I think It should not be hard to nest the Exceptions later on if needed.

aozarov:
Looks good to me too (after fixing the nit @ajkannan mentioned).

mziccard:
Fixed comment in place!

PR #534, opened 2016-01-08
Add support for streaming inserts with resumable upload

mziccard:
This PR adds support to BigQuery for inserts that use [resumable upload](https://cloud.google.com/bigquery/loading-data-post-request#resumable).
- Add `LoadConfiguration` class for data needed by load jobs and resumable uploads
- Update `LoadJobInfo` and related classes to use `LoadConfiguration` instead
- Add `status` and `write` methods to `BigQueryRpc` to support resumable insert
- Add `insertAll(LoadConfiguration, byte[])` method to `BigQuery`
- Add `insertAll(LoadConfiguration. SeekableByteChannel)` method to `BigQuery`
- Add `Table.insert` methods that use resumable upload

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
@@ -27,6 +27,7 @@ import com.google.gcloud.Page;
 import com.google.gcloud.Service;
 import com.google.gcloud.spi.BigQueryRpc;
 
+import java.nio.channels.SeekableByteChannel;
 import java.util.List;
 import java.util.Set;
 
@@ -589,6 +590,33 @@ public interface BigQuery extends Service<BigQueryOptions> {
    * @throws BigQueryException upon failure
    */
   InsertAllResponse insertAll(InsertAllRequest request) throws BigQueryException;
+
+  /**
+   * Sends a resumable insert request given a seekable channel containing the rows to be inserted.
+   * This method does not close the channel so you should take care of closing it.
+   *
+   * <p>Example usage of inserting data from a local file:
+   * <pre> {@code
+   * LoadConfiguration config = LoadConfiguration.of(TableId.of("my_dataset_id", "my_table_id"));
+   * try(FileChannel channel = FileChannel.open(Paths.get("/path/to/file"))) {
+   *   bigquery.insertAll(config, channel);
+   * }}</pre>
+   *
+   * @throws BigQueryException upon failure
+   * @see <a href="https://cloud.google.com/bigquery/loading-data-post-request#resumable">Resumable
+   *     Upload</a>
+   */
+  void insertAll(LoadConfiguration configuration, SeekableByteChannel channel)
+      throws BigQueryException;
+
+  /**
+   * Sends a resumable insert request given a byte array containing the rows to be inserted.
+   *
+   * @throws BigQueryException upon failure
+   * @see <a href="https://cloud.google.com/bigquery/loading-data-post-request#resumable">Resumable
+   *     Upload</a>
+   */
+  void insertAll(LoadConfiguration configuration, byte[] content) throws BigQueryException;
 
   /**
    * Lists the table's rows.
--- a/gcloud-java-examples/src/main/java/com/google/gcloud/examples/BigQueryExample.java
+++ b/gcloud-java-examples/src/main/java/com/google/gcloud/examples/BigQueryExample.java
@@ -33,6 +33,7 @@ import com.google.gcloud.bigquery.FormatOptions;
 import com.google.gcloud.bigquery.JobId;
 import com.google.gcloud.bigquery.JobInfo;
 import com.google.gcloud.bigquery.JobStatus;
+import com.google.gcloud.bigquery.LoadConfiguration;
 import com.google.gcloud.bigquery.LoadJobInfo;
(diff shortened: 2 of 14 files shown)


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -591,6 +592,33 @@ public static QueryResultsOption maxWaitTime(long maxWaitTime) {
   InsertAllResponse insertAll(InsertAllRequest request) throws BigQueryException;
 
   /**
+   * Sends a resumable insert request given a seekable channel containing the rows to be inserted.
+   * This method does not close the channel so you should take care of closing it.
+   *
+   * <p>Example usage of inserting data from a local file:
+   * <pre> {@code
+   * LoadConfiguration config = LoadConfiguration.of(TableId.of("my_dataset_id", "my_table_id"));
+   * try(FileChannel channel = FileChannel.open(Paths.get("/path/to/file"))) {
```
aozarov: Nit, "try ("

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -591,6 +592,33 @@ public static QueryResultsOption maxWaitTime(long maxWaitTime) {
...
+   *   bigquery.insertAll(config, channel);
+   * }}</pre>
+   *
+   * @throws BigQueryException upon failure
+   * @see <a href="https://cloud.google.com/bigquery/loading-data-post-request#resumable">Resumable
+   *     Upload</a>
+   */
+  void insertAll(LoadConfiguration configuration, SeekableByteChannel channel)
```
aozarov: You don't think we could have done it similar to storage were we return to the caller a Writer?

aozarov: Also, if we keep as is, any reason not to get InputStream directly (you could wrap it in BufferedInputStream or a like to support mark/reset for retries).

gcloud-java-examples/src/main/java/com/google/gcloud/examples/BigQueryExample.java
```diff
@@ -33,6 +33,7 @@
 import com.google.gcloud.bigquery.JobId;
```
aozarov: Nit, Unrelated to the PR. s/suceeded/succeeded/

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
mziccard: > You don't think we could have done it similar to storage were we return to the caller a WriteChannel?

Yes that could be done, what discouraged me is this sentence from [the docs](https://cloud.google.com/bigquery/loading-data-post-request#chunking)

> With resumable uploads, you can break a file into chunks and send a series of requests to upload each chunk in sequence. This is not the preferred approach since there are performance costs associated with the additional requests, and it is generally not needed. 

However I understand chunking might be useful so I am happy to change.

> you could wrap it in BufferedInputStream or a like to support mark/reset for retries

If I am not mistaken BufferedInputStream supports mark/reset by buffering all data into a byte array, for large data this might become a problem.

aozarov: > Yes that could be done, what discouraged me is this sentence from the docs

Actually, in addition to being consistent, that section mention 2 more reasons to support it.
Running on platforms that limit bytes/time per request and supporting frameworks that run
in chunks/slices (e.g. MR).

> However I understand chunking might be useful so I am happy to change.

We could support both if you want. One similar to `Storage.create` with `InputStream` and
one similar to `Storage.write`.

> If I am not mistaken BufferedInputStream supports mark/reset by buffering all data into a byte array, for large data this might become a problem.

It does not necessarily buffer *all* data but rather a configurable size or when mark/reset is used it will buffer up-to the limit. So, it is true that if we unbound the limit that would not be good but if we write in chunks it could potentially used. 
Regardless, I think that you are right not using it considering current implementation
and using status to get the last position and I would keep it this way if we decide to provide both ways.

mziccard: Yes I oversimplified the behavior of BufferedInputStream :) What I intended to say is that in the current implementation we cannot afford to use it.

I will implement the `WriteChannel` approach, we can always add this direct method later if need be.

I think the `WriteChannel` is good enough

aozarov: > I will implement the WriteChannel approach, we can always add this direct method later if need be.

Sounds good. And I would even say that when we do so, we should consider adding it
to `Storage.create()` as well (instead `InputStream` or overload it) as GCS supports
the same kind of resumable concept.

lesv:
ping - is this PR obsolete?  Should it be deleted?  Or what?

mziccard:
This PR became obsolete but contains an implementation of upload that could be useful for both the BigQuery and Storage modules. Due to lack of time and low priority this remained behind. If this bothers you we can close, I will reopen a new one when ready.

sduskis:
Given the age of this PR, I'm closing it.

PR #536, opened 2016-01-08
Fix failing ITBigQueryTest.testUpdateNonExistingTable

mziccard:
There's a test failing in Travis (my bad). The failing test is `ITBigQueryTest.testUpdateNonExistingTable`

``` java
java.lang.AssertionError: null
    at org.junit.Assert.fail(Assert.java:86)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at org.junit.Assert.assertNotNull(Assert.java:712)
    at org.junit.Assert.assertNotNull(Assert.java:722)
    at com.google.gcloud.bigquery.ITBigQueryTest.testUpdateNonExistingTable(ITBigQueryTest.java:465)
```

The line failing is the following:

``` java
assertNotNull(error.debugInfo());
```

It seems that `debugInfo` is not available. Noticeably this error never happened to me locally. This PR should fix this error even though I am unsure why it even happens.


aozarov:
Debug information is only provided by the service internally (IP, users,...).

PR #539, opened 2016-01-09
Create simple forceDelete that can be used on App Engine

ajkannan:
This method avoids spawning a thread to delete the bucket (which can be a problem on some platforms).

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/testing/RemoteGcsHelper.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/testing/RemoteGcsHelper.java
@@ -65,7 +65,10 @@ public class RemoteGcsHelper {
 
   /**
    * Deletes a bucket, even if non-empty. Objects in the bucket are listed and deleted until bucket
-   * deletion succeeds or {@code timeout} expires.
+   * deletion succeeds or {@code timeout} expires. To allow for the timeout, this method uses a
+   * separate thread to send the delete requests. Use
+   * {@link #forceDelete(Storage storage, String bucket)} if spawning an additional thread is
+   * undesirable, such as in the App Engine production runtime.
    *
    * @param storage the storage service to be used to issue requests
    * @param bucket the bucket to be deleted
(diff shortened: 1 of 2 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/storage/testing/RemoteGcsHelper.java
```diff
@@ -89,6 +89,23 @@ public static Boolean forceDelete(Storage storage, String bucket, long timeout,
   }
 
   /**
+   * Deletes a bucket, even if non-empty. Objects in the bucket are listed and deleted until bucket
+   * deletion succeeds. This method can be used to delete buckets from within App Engine.  Note that
```
aozarov: Instead, I would mentioned in the other method that it uses a separate thread to allow the timeout
and if that is a problem (such as in AE) one should use this method instead (with a "@link" reference). Also, instead of "Note that ..." I would just say "This method blocks until the deletion competed or failed".

gcloud-java-storage/src/main/java/com/google/gcloud/storage/testing/RemoteGcsHelper.java
```diff
@@ -89,6 +89,23 @@ public static Boolean forceDelete(Storage storage, String bucket, long timeout,
...
+   *
+   * @param storage the storage service to be used to issue requests
+   * @param bucket the bucket to be deleted
+   * @throws StorageException if an exception is encountered during bucket deletion
+   */
+  public static void forceDelete(Storage storage, String bucket) throws StorageException {
+    try {
+      new DeleteBucketTask(storage, bucket).call();
+    } catch (Exception e) {
+      throw (StorageException) e;
```
aozarov: This could trigger a `ClassCastException`.

We should either change `DeleteBucketTask.call` method signature and remove the `throws Exception` (not needed as StorageException is a runtime exception if we also change `Thread.sleep` to use guava's [`Uninterruptibles.sleepUninterruptibly`](http://docs.guava-libraries.googlecode.com/git/javadoc/src-html/com/google/common/util/concurrent/Uninterruptibles.html#line.267)) or use
guava's `Throwables.propagateIfInstanceOf(e, StorageException.class)` (and a line
after that add `throw Throwables.propagate(e)`)

ajkannan:
@aozarov added a commit to address those comments.

PR #540, opened 2016-01-11
Add support for BigQuery resumable uploads via a write channel

mziccard:
This PR adds support to BigQuery for inserts that use [resumable upload](https://cloud.google.com/bigquery/loading-data-post-request#resumable) using a write channel interface (different from #534).

Main changes are:
- Move `BlobWriteChannel` and `BlobReadChannel` to core module (the latter for consistency) 
- Rename `BlobWriteChannel` and `BlobReadChannel` to `WriteChannel` and `ReadChannel`
- Add abstract class `BaseWriteChannel`implementing entity-agnostic channel functionality
- Add `BlobWriteChannel` and `BlobReadChannel` implementations to `gcloud-java-storage`
- Add `LoadConfiguration` and modify `LoadJobInfo` to take configuration as a parameter
- Add `BigQuery.writer` method to return a write channel given a `LoadConfiguration`
- Add `BigQueryRpc.open` and `.write` methods to implement write channel
- Add `TableDataWriteChannel` class to support bigquery resumable streaming inserts
- Update bigquery example with load-data action


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
```diff
@@ -0,0 +1,381 @@
...
+import com.google.gcloud.bigquery.JobInfo.CreateDisposition;
+import com.google.gcloud.bigquery.JobInfo.WriteDisposition;
+
+import java.io.Serializable;
+import java.nio.channels.SeekableByteChannel;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Google BigQuery Configuration for a Load operation. A Load configuration can be used to build a
```
ajkannan: Nit: should "Load" --> "load?  It looks like "load" isn't capitalized in the BigQuery docs (e.g. in step 2 [here](https://cloud.google.com/bigquery/loading-data-into-bigquery#loaddatagcs))

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
```diff
@@ -0,0 +1,381 @@
...
+     * already exists, or if you're loading data from Google Cloud Datastore.
+     */
+    public Builder schema(Schema schema) {
+      this.schema = schema;
+      return this;
+    }
+
+    /**
+     * Sets whether BigQuery should allow extra values that are not represented in the table schema.
+     * If {@code true}, the extra values are ignored. If {@code true}, records with extra columns
```
ajkannan: Should the second {@code true} be "false" instead?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
```diff
@@ -0,0 +1,381 @@
...
+   *     Jobs: Load Configuration</a>
+   */
+  public CreateDisposition createDisposition() {
+    return this.createDisposition;
+  }
+
+  /**
+   * Returns the action that should occur if the destination table already exists.
+   *
+   * @see <a href="https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load">
```
ajkannan: Nit: I think it would slightly clearer to link to the write disposition entry directly (#configuration.load.writeDisposition).  Same for `createDisposition()` and both methods' equivalents in the builder.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
```diff
@@ -0,0 +1,381 @@
...
+   * Returns the schema for the destination table, if set. Returns {@code null} otherwise.
+   */
+  public Schema schema() {
+    return schema;
+  }
+
+  /**
+   * Returns the format of the data files.
+   */
+  public String format() {
```
ajkannan: Should this be `formatOptions`?  Also I like the javadoc in the builder for this functionality, can we copy it over for this method?

gcloud-java-core/src/main/java/com/google/gcloud/WriteChannel.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Sets the minimum size that will be written by a single RPC.
+   * Written data will be buffered and only flushed upon reaching this size or closing the channel.
+   */
+  void chunkSize(int chunkSize);
+
+  /**
+   * Captures the write channel state so that it can be saved and restored afterwards. The original
+   * {@code WriteChannel} and the restored one should not both be used. Closing one channel
+   * causes the other channel to close, subsequent writes will fail.
```
ajkannan: Nit: replace the comma with a period or semicolon

gcloud-java-core/src/main/java/com/google/gcloud/BaseWriteChannel.java
```diff
@@ -0,0 +1,218 @@
...
+
+    /**
+     * Base builder for a write channel's state. Users are not supposed to access this class
+     * directly.
+     *
+     * @param <ServiceOptionsT> the service options used by the channel to issue RPC requests
+     * @param <EntityT> the entity this channel writes data to. Possibly with additional
+     *     configuration
+     */
+    public abstract static class Builder<
```
ajkannan: I don't think this is tested (correct me if I'm wrong).

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/BigQueryRpc.java
```diff
@@ -185,4 +186,26 @@ GetQueryResultsResponse getQueryResults(String jobId, Map<Option, ?> options)
...
+   */
+  String open(JobConfigurationLoad configuration) throws BigQueryException;
+
+  /**
+   * Uploads the provided data to the resumable upload session at the specified position.
+   *
+   * @param uploadId the resumable upload session URI
+   * @param toWrite a byte array of data to upload
+   * @param toWriteOffset offset in the {@code toWrite} param to start writing from
+   * @param destOffset offset in the destition where to upload data to
```
ajkannan: s/destition/destination

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/BigQueryRpc.java
```diff
@@ -185,4 +186,26 @@ GetQueryResultsResponse getQueryResults(String jobId, Map<Option, ?> options)
...
+   * @param length the number of butes to upload
```
ajkannan: s/butes/bytes

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/BigQueryRpc.java
```diff
@@ -185,4 +186,26 @@ GetQueryResultsResponse getQueryResults(String jobId, Map<Option, ?> options)
...
+   * @param last if {@code true} indicates that the last chunk if being uploaded
```
ajkannan: The second "if" (near the end of the line) should be an "is".  This also might be clearer if worded as "true indicates that the ..." (removing the "if")

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
```diff
@@ -64,6 +76,9 @@
   public static final String DEFAULT_PROJECTION = "full";
   // see: https://cloud.google.com/bigquery/troubleshooting-errors
   private static final Set<Integer> RETRYABLE_CODES = ImmutableSet.of(500, 502, 503, 504);
+  private static final String BASE_RESUMABLE_URI =
+      "https://www.googleapis.com/upload/bigquery/v2/projects/";
+  private static final int HTTP_RESUME_INCOMPLETE = 308;
```
ajkannan: Do you mind adding a link for this error code as a comment?  It doesn't see to be on the same table as the others, and it'd be good to have it documented.

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
```diff
@@ -417,4 +432,69 @@ public QueryResponse query(QueryRequest request) throws BigQueryException {
...
+  @Override
+  public void write(String uploadId, byte[] toWrite, int toWriteOffset, long destOffset, int length,
+      boolean last) throws BigQueryException {
+    try {
+      GenericUrl url = new GenericUrl(uploadId);
+      HttpRequest httpRequest = bigquery.getRequestFactory().buildPutRequest(url,
+          new ByteArrayContent(null, toWrite, toWriteOffset, length));
+      long limit = destOffset + length;
+      StringBuilder range = new StringBuilder("bytes ");
+      range.append(destOffset).append('-').append(limit - 1).append('/');
```
ajkannan: question: why is the last '/' necessary?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
mziccard: This method returns the format as a string, not the options. To get options you can use `csvOptions()` with returns a `CsvOptions` object (which is the only format options supported for the moment).

mziccard:
@ajkannan Thanks for the feedback, I fixed the issues in place

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
```diff
@@ -417,4 +433,69 @@ public QueryResponse query(QueryRequest request) throws BigQueryException {
...
+        HttpResponse response = httpRequest.execute();
+        code = response.getStatusCode();
+        message = response.getStatusMessage();
+      } catch (HttpResponseException ex) {
+        exception = ex;
+        code = ex.getStatusCode();
+        message = ex.getStatusMessage();
+      }
+      if (!last && code != HTTP_RESUME_INCOMPLETE
+          || last && !(code == HTTP_OK || code == HTTP_CREATED)) {
```
ajkannan: It seems like there's a [very small] chance for throwing an exception when the upload actually worked (under step 2 in the [example](https://cloud.google.com/bigquery/loading-data-post-request#resume-upload)).  Do you think this is worth documenting?

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
```diff
@@ -417,4 +433,69 @@ public QueryResponse query(QueryRequest request) throws BigQueryException {
...
+      httpRequest.getHeaders().set("X-Upload-Content-Value", "application/octet-stream");
+      HttpResponse response = httpRequest.execute();
+      return response.getHeaders().getLocation();
+    } catch (IOException ex) {
+      throw translate(ex);
+    }
+  }
+
+  @Override
+  public void write(String uploadId, byte[] toWrite, int toWriteOffset, long destOffset, int length,
```
ajkannan: Is it possible to overwrite bytes with a request?  I would guess yes, but I figured I'd ask anyways just because I didn't see that in any example on the resumable upload docs.

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
mziccard: I am missing how this can happen. The request for status must have content range of the form `Content-Range: bytes */<limit>`. If I am not mistaken this can never occur as we always set a range.

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
mziccard: Yes you should be able to do that. The upload mechanism is the exact same used in storage.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/LoadConfigurationTest.java
```diff
@@ -0,0 +1,123 @@
...
+    configuration = LoadConfiguration.of(TABLE_ID, CSV_OPTIONS);
+    assertEquals(TABLE_ID, configuration.destinationTable());
+    assertEquals(FORMAT, configuration.format());
+    assertEquals(CSV_OPTIONS, configuration.csvOptions());
+  }
+
+  @Test
+  public void testToBuilderIncomplete() {
+    LoadConfiguration configuration = LoadConfiguration.of(TABLE_ID);
+    compareLoadConfiguration( configuration, configuration.toBuilder().build());
```
ajkannan: Nit: extra space after '('

gcloud-java-bigquery/src/main/java/com/google/gcloud/spi/DefaultBigQueryRpc.java
ajkannan: You're right (I confused myself earlier); ignore my original comment.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableDataWriteChannel.java
```diff
@@ -0,0 +1,109 @@
...
+      private Builder(BigQueryOptions options, LoadConfiguration configuration, String uploadId) {
+        super(options, configuration, uploadId);
+      }
+
+      public RestorableState<WriteChannel> build() {
+        return new StateImpl(this);
+      }
+    }
+
+    static Builder builder(BigQueryOptions options, LoadConfiguration blobInfo, String uploadId) {
```
ajkannan: s/blobInfo/configuration

ajkannan:
@mziccard I took one more pass, and LGTM except for one small variable name change.  Feel free to merge after fixing that.

gcloud-java-core/src/main/java/com/google/gcloud/BaseWriteChannel.java
```diff
@@ -0,0 +1,218 @@
...
+ * Base implementation for a {@link WriteChannel}.
+ *
+ * @param <ServiceOptionsT> the service options used by the channel to issue RPC requests
+ * @param <EntityT> the entity this channel writes data to. Possibly with additional configuration
+ */
+public abstract class BaseWriteChannel<
+    ServiceOptionsT extends ServiceOptions<?, ?, ServiceOptionsT>,
+    EntityT extends Serializable> implements WriteChannel {
+
+  private static final int MIN_CHUNK_SIZE = 256 * 1024;
```
aozarov: At some point this may need to be overridden by subclasses. current value is specific to GCS.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
```diff
@@ -0,0 +1,381 @@
...
+     * By default no bad record is ignored.
+     */
+    public Builder maxBadRecords(Integer maxBadRecords) {
+      this.maxBadRecords = maxBadRecords;
+      return this;
+    }
+
+    /**
+     * Sets the schema for the destination table. The schema can be omitted if the destination table
+     * already exists, or if you're loading data from Google Cloud Datastore.
```
aozarov: from Google Cloud Datastore backup (DATASTORE_BACKUP format option)?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    /**
+     * Sets which entity properties to load into BigQuery from a Cloud Datastore backup. This field
+     * is only used if the source format is set to {@code DATASTORE_BACKUP}. Property names are case
+     * sensitive and must be top-level properties. If no properties are specified, BigQuery loads
+     * all properties. If any named property isn't found in the Cloud Datastore backup, an invalid
+     * error is returned in the job result.
+     */
+    public Builder projectionFields(List<String> projectionFields) {
```
aozarov: shouldn't it be part of the DATASTORE_BACKUP FormatOptions? I mean via a subclasses, similar to CSV?

aozarov:
Looks good! I would like to give it another pass tomorrow. Until then few questions/comments. Also,
I think there are still unaddressed comments made by @ajkannan.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadConfiguration.java
mziccard: You are right, I opened a new issue (#552) as this PR is already big enough :)

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/BigQueryImplTest.java
```diff
@@ -41,10 +41,12 @@
 import com.google.common.collect.Lists;
 import com.google.gcloud.Page;
 import com.google.gcloud.RetryParams;
+import com.google.gcloud.WriteChannel;
 import com.google.gcloud.bigquery.InsertAllRequest.RowToInsert;
 import com.google.gcloud.spi.BigQueryRpc;
 import com.google.gcloud.spi.BigQueryRpc.Tuple;
 import com.google.gcloud.spi.BigQueryRpcFactory;
+import com.google.gcloud.storage.BlobInfo;
```
aozarov: Not needed.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/ITBigQueryTest.java
```diff
@@ -16,6 +16,7 @@
 
 package com.google.gcloud.bigquery;
 
+import static com.google.gcloud.bigquery.BigQuery.*;
```
aozarov: replace with specific.

gcloud-java-core/src/main/java/com/google/gcloud/BaseWriteChannel.java
```diff
@@ -0,0 +1,226 @@
...
+  private static final int DEFAULT_CHUNK_SIZE = 8 * MIN_CHUNK_SIZE;
+
+  protected final ServiceOptionsT options;
```
aozarov: To prevent bad practice or mishandling in sub-classes I suggest to have all members private with protected accessors.
This is not an issue for the final members but I would favor consistency.

gcloud-java-core/src/main/java/com/google/gcloud/BaseWriteChannel.java
```diff
@@ -0,0 +1,226 @@
...
+
+  protected int minChunkSize() {
+    return MIN_CHUNK_SIZE;
+  }
+
+  protected int defaultChunkSize() {
+    return DEFAULT_CHUNK_SIZE;
+  }
+
+  protected abstract void write(int length, boolean last);
```
aozarov: javadoc would be nice. Also maybe name should be more specific (e.g. `writeBuffer`, `flushBuffer`, ...).

gcloud-java-core/src/main/java/com/google/gcloud/BaseWriteChannel.java
```diff
@@ -0,0 +1,226 @@
...
+
+  protected BaseWriteChannel(ServiceOptionsT options, EntityT entity, String uploadId) {
+    this.options = options;
+    this.entity = entity;
+    this.uploadId = uploadId;
+  }
+
+  protected void flush() {
```
aozarov: I would make anything that is not need to be overriden by the subclasses either package scope, private or final (if must be public). Especially methods that would easily get it wrong and require deep understanding on the base class internals.

gcloud-java-core/src/main/java/com/google/gcloud/ReadChannel.java
```diff
@@ -0,0 +1,57 @@
...
+  void chunkSize(int chunkSize);
+
+  /**
+   * Captures the read channel state so that it can be saved and restored afterwards.
+   *
+   * @return a {@link RestorableState} object that contains the read channel state and can restore
+   *     it afterwards.
+   */
+  @Override
+  RestorableState<ReadChannel> capture();
```
aozarov: I think this could be implemented in `BaseWriteChannel` using an abstract protected method that would return a newly created `BaseState.Builder` instance.

aozarov:
Ok, finished with my pass. Looks good. Few more suggestion and some cleanup.
Also, do you understand why the coverage was dropped in this PR by almost 1 percent?

mziccard:
@aozarov I fixed your comments and rebased.

> do you understand why the coverage was dropped in this PR by almost 1 percent?

I have been trying to figure this out but have not so many ideas. A decrease might be caused by the fact that this adds code to the Rpc layer which is only tested in ITs that are (I think) not considered by *coveralls* report. However this ~1% seems a bit high to me.

PR #546, opened 2016-01-11
Fixed #478 - Make DatastoreException informative.

### LINKED ISSUE #478: Make DatastoreException message more informative
In `DefaultDatastoreRpc.translate(DatastoreException exception)`, we lose helpful error messages because we only throw a DatastoreRpcException based on the associated `Reason`.  For example, in a malformed query request, the user gets an exception with the message "Invalid request".  The original exception thrown and translated by the method mentioned above may include a message such as "must specify all group by orderings before any non group by orderings", which is far more helpful for the user.

mderka:
Makes DatastoreException message more informative by preserving a cause.
Fixes #478


ajkannan:
LGTM

PR #549, opened 2016-01-14
Populate cursorAfter in datastore v1beta2

### LINKED ISSUE #548: Lazy loading of entities when running a query
I just want to scroll all entities of a specific kind in a lazy way. I have tried to add a limit but as intended I have only the number of specified results and not the entire dataset.

```java
StructuredQuery<Entity> query = Query.entityQueryBuilder()
    .kind("user")
    .limit(100)
    .build();

QueryResults<Entity> results = datastore.run(query);
```

How to do it ? Did you plan to add a `fetchSize()` parameter ?

Without `limit()` I have a DatastoreException: Server returned an error INTERNAL 500

ajkannan:
Fixes #548

### DIFF
--- a/gcloud-java-datastore/src/test/java/com/google/gcloud/datastore/DatastoreTest.java
+++ b/gcloud-java-datastore/src/test/java/com/google/gcloud/datastore/DatastoreTest.java
@@ -477,12 +477,11 @@ public class DatastoreTest {
           .andReturn(responses.get(i));
     }
     EasyMock.replay(rpcFactoryMock, rpcMock);
-    Datastore mockDatastore =
-        options.toBuilder()
-            .retryParams(RetryParams.defaultInstance())
-            .serviceRpcFactory(rpcFactoryMock)
-            .build()
-            .service();
+    Datastore mockDatastore = options.toBuilder()
+        .retryParams(RetryParams.defaultInstance())
+        .serviceRpcFactory(rpcFactoryMock)
+        .build()
+        .service();
     int limit = 2;
     int totalCount = 0;
     StructuredQuery<Entity> query = Query.entityQueryBuilder().limit(limit).build();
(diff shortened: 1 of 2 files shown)


gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/QueryResults.java
```diff
@@ -35,8 +35,9 @@
   Class<?> resultClass();
 
   /**
-   * Returns the Cursor for point after the value returned in the last {@link #next} call.
-   * Not currently implemented (depends on v1beta3).
+   * Returns the Cursor for the point after the value returned in the last {@link #next} call. Until
+   * v1beta3 is supported, this field should only be used if you have set a limit and the number of
```
aozarov: I don't think `v1beta3` says much to people. Instead I think we should just say "currently cursorAfter returns null in all cases but the last result" or something like that.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/QueryResultsImpl.java
```diff
@@ -100,6 +100,6 @@ protected T computeNext() {
   @Override
   public Cursor cursorAfter() {
     //return new Cursor(cursor); // only available in v1beta3
-    return null;
+    return new Cursor(queryResultBatchPb.getEndCursor());
```
aozarov: I think we should set it to non-null only after last result is consumed.

gcloud-java-datastore/src/test/java/com/google/gcloud/datastore/DatastoreTest.java
```diff
@@ -463,6 +466,95 @@ public void testRunStructuredQuery() {
...
+    EasyMock.expect(rpcFactoryMock.create(EasyMock.anyObject(DatastoreOptions.class)))
+        .andReturn(rpcMock);
+    List<RunQueryResponse> responses = buildResponsesForQueryPaginationWithLimit();
+    for (int i = 0; i < responses.size(); i++) {
+      EasyMock.expect(rpcMock.runQuery(EasyMock.anyObject(RunQueryRequest.class)))
+          .andReturn(responses.get(i));
+    }
+    EasyMock.replay(rpcFactoryMock, rpcMock);
+    Datastore mockDatastore =
+        this.options.toBuilder()
```
aozarov: Nit "is "this." necessary?

ajkannan:
Added commit to address comments

ajkannan:
@mziccard In [663e54b](https://github.com/ajkannan/gcloud-java/commit/663e54b356162b6890fd8175c9cc17a4a70dd097) I added a little bit more to the documentation for `QueryResults`.  Do you think there's anything else that should be mentioned to make it clearer?

aozarov:
Looks good. @mziccard I am merging it but please comment and we should open a separate PR for any further doc changes.

PR #551, opened 2016-01-14
Add link to bookshelf example app

ajkannan:


ajkannan:
It doesn't look like there's that much documentation on how to run the app yet, so we can choose to wait on adding it to the list if that's preferable.

PR #554, opened 2016-01-15
Refactor service exceptions

mziccard:
This PR follows the discussion in #438 and makes a better job (I think) at refactoring service exceptions.
066a78f also reduces exceptions handlers to one.

This PR required a bunch of changes across all modules so guys give your best review :)

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
@@ -58,6 +58,16 @@ public class BaseServiceException extends RuntimeException {
       return reason;
     }
 
+    boolean isRetryable(Set<Error> retryableErrors) {
+      for (Error retryableError : retryableErrors) {
+        if ((retryableError.code() == null || retryableError.code().equals(this.code()))
+            && (retryableError.reason() == null || retryableError.reason().equals(this.reason()))) {
+          return true;
+        }
+      }
+      return false;
+    }
+
     @Override
     public String toString() {
       return MoreObjects.toStringHelper(this).add("code", code).add("reason", reason).toString();
@@ -83,7 +93,7 @@ public class BaseServiceException extends RuntimeException {
       Error error = error(((GoogleJsonResponseException) exception).getDetails());
       this.code = error.code;
       this.reason = error.reason;
-      this.retryable = isRetryable(error);
+      this.retryable = error.isRetryable(retryableErrors());
     } else {
       this.code = UNKNOWN_CODE;
       this.reason = null;
@@ -110,7 +120,7 @@ public class BaseServiceException extends RuntimeException {
     this.code = code;
     this.reason = reason;
     this.idempotent = idempotent;
-    this.retryable = idempotent && isRetryable(new Error(code, reason));
+    this.retryable = idempotent && new Error(code, reason).isRetryable(retryableErrors());
   }
 
   protected Set<Error> retryableErrors() {
@@ -118,7 +128,7 @@ public class BaseServiceException extends RuntimeException {
   }
 
   protected boolean isRetryable(GoogleJsonError error) {
-    return error != null && isRetryable(error(error));
+    return error != null && error(error).isRetryable(retryableErrors());
   }
 
   protected boolean isRetryable(IOException exception) {
@@ -126,16 +136,6 @@ public class BaseServiceException extends RuntimeException {
       return isRetryable(((GoogleJsonResponseException) exception).getDetails());
     }
     return exception instanceof SocketTimeoutException;
-  }
-
-  protected boolean isRetryable(Error error) {
-    for (Error retryableError : retryableErrors()) {
-      if ((retryableError.code() == null || retryableError.code().equals(error.code()))
-          && (retryableError.reason() == null || retryableError.reason().equals(error.reason()))) {
-        return true;
-      }
-    }
-    return false;
   }
 
   /**


gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -16,26 +16,115 @@
...
+import java.util.Collections;
+import java.util.Set;
+
 /**
  * Base class for all service exceptions.
  */
 public class BaseServiceException extends RuntimeException {
 
-  private static final long serialVersionUID = 5028833760039966178L;
+  protected static class Error {
```
aozarov: Shall we make it `final` and `Serializable`?

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -16,26 +16,115 @@
...
+
+    private final Integer code;
+    private final String reason;
+
+    public Error(Integer code, String reason) {
+      this.code = code;
```
aozarov: Do we accept `null`? If not, I would prefer to use primitive int.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -16,26 +16,115 @@
...
+  public BaseServiceException(GoogleJsonError error, boolean idempotent) {
+    super(error.getMessage());
+    this.code = error.getCode();
+    this.reason = reason(error);
+    this.idempotent = idempotent;
+    this.retryable = idempotent && isRetryable(error);
+  }
 
-  public BaseServiceException(int code, String message, boolean retryable) {
+  public BaseServiceException(int code, String message, String reason, boolean idempotent) {
```
aozarov: Nit, this could just call `this(code, message, reason, idempotent, null);`

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mziccard: We use `null` to mean "any" in the `retriablyErrors()` `Set`. We could use `UNKNOWN_CODE` if you prefer.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -46,9 +135,57 @@ public int code() {
...
+  protected static String reason(GoogleJsonError error) {
+    if (error.getErrors() != null  && !error.getErrors().isEmpty()) {
+      return error.getErrors().get(0).getReason();
+    }
+    return null;
+  }
+
+  protected static Error error(GoogleJsonError error) {
+    int code = error.getCode();
+    String reason = null;
```
aozarov: `String reason = reason(error)`?

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -16,26 +16,115 @@
...
+      return isRetryable(((GoogleJsonResponseException) exception).getDetails());
+    }
+    return exception instanceof SocketTimeoutException;
+  }
+
+  protected boolean isRetryable(Error error) {
+    for (Error retryableError : retryableErrors()) {
+      if ((retryableError.code() == null || retryableError.code().equals(error.code()))
+          && (retryableError.reason() == null || retryableError.reason().equals(error.reason()))) {
+          return true;
```
aozarov: wrong indent. (needs 2 backspaces).

mziccard:
Fixed comments in place

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -46,9 +146,52 @@ public int code() {
...
+
+  protected static BaseServiceException translateAndThrow(
+      RetryHelper.RetryHelperException ex) {
+    if (ex.getCause() instanceof BaseServiceException) {
+      throw (BaseServiceException) ex.getCause();
+    }
+    if (ex instanceof RetryHelper.RetryInterruptedException) {
+      RetryHelper.RetryInterruptedException.propagate();
+    }
+    return null;
```
aozarov: Similar to guava's [Thrwoables](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Throwables.html) previously we would both
throw and return so one could invoke the method as `throw translateAndThrow` if desired.

Now, when this method can return `null` I see no reason for the return value.
Also, maybe we should consider rename it to `propagateIfPossible` or `translateAndPropagateIfPossible`?

Another option could get a "default" exception to throw if could not propagate (which seems to match current usage pattern).

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/DatastoreException.java
```diff
@@ -16,141 +16,70 @@
...
    */
-  static DatastoreException translateAndThrow(DatastoreRpcException exception) {
-    String message = exception.getMessage();
-    DatastoreError error = REASON_TO_ERROR.get(exception.reason());
-    if (error == null) {
-      error = MoreObjects.firstNonNull(
-          HTTP_TO_ERROR.get(exception.httpStatus()), DatastoreError.UNKNOWN);
-    }
-    throw error.translate(exception, message);
+  public static DatastoreException translateAndThrow(RetryHelperException ex) {
```
aozarov: Here and for the other services. Any reason for not keeping it package scope?

gcloud-java-datastore/src/main/java/com/google/gcloud/spi/DefaultDatastoreRpc.java
```diff
@@ -106,90 +87,82 @@ private static boolean includesScheme(String url) {
...
-      reasonStr = "Unknown";
-      if (exception.getCause() instanceof SocketTimeoutException) {
-        retryable = true;
-        reasonStr = "Request timeout";
+      if (exception.getCause() instanceof IOException) {
+        return new DatastoreException((IOException) exception.getCause());
       }
-      return new DatastoreRpcException(reasonStr, exception.getCode(), retryable, message, exception);
     }
+    return new DatastoreException(code, message, reason);
```
aozarov: Any reason for reverting #546 (issue #478)?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerException.java
```diff
@@ -29,11 +32,32 @@
...
 public class ResourceManagerException extends BaseServiceException {
 
-  private static final long serialVersionUID = 6841689911565501705L;
-  private static final int UNKNOWN_CODE = -1;
+  // see https://cloud.google.com/resource-manager/v1/errors/core_errors
+  private static final Set<Error> RETRYABLE_ERRORS = ImmutableSet.of(
+      new Error(503, null),
+      new Error(500, null),
+      new Error(429, null),
+      new Error(403, "concurrentLimitExceeded"),
```
aozarov: Nit, fine as is but could be optimized if reason was a `Set<String>`.
Also, Error could have a package scope `isRetryable` for doing the internal check of code and reason matching.

aozarov:
Looks great. Few more comments.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerException.java
mziccard: > Nit, fine as is but could be optimized if reason was a Set<String>.

I prefer to leave it as is as so that an `Error` object represents a single possible error that can occur.

> Also, Error could have a package scope isRetryable for doing the internal check of code and reason matching.

Done in cf2d7f4

gcloud-java-datastore/src/main/java/com/google/gcloud/spi/DefaultDatastoreRpc.java
mziccard: Fixed this in 40fd11c

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/DatastoreException.java
mziccard: Fixed in 86b91a1

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mziccard: I renamed the method to `translateAndPropagateIfPossible ` in 86b91a1

mziccard:
I fixed the comments, after review I will have to rebase and fix conflicts so that we can merge

aozarov:
LGTM. Feel free to merge once you rebase and fix conflict (will gladly take another look if you think necessary).

PR #558, opened 2016-01-16
Gcloud dns

mderka:
This is the initial project setup including code for the DnsRecord class.


mderka:
Closing. Wrong branch, wrong pom. Will do again.

PR #563, opened 2016-01-21
Add code to initialize BigQueryError in BigQueryException

mziccard:


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryException.java
```diff
@@ -53,7 +55,16 @@ public BigQueryException(int code, String message, BigQueryError error) {
 
   public BigQueryException(IOException exception) {
     super(exception, true);
-    this.error = null;
+    BigQueryError bigqueryError = null;
```
aozarov: Should we add debugInfo and location to  `BaseServiceException`. then we can just use the other getters to populate BigQueryError...

PR #565, opened 2016-01-21
Add equals and hashCode to BaseTableInfo subclasses

### LINKED ISSUE #564: BaseTableInfo subclasses are missing equals and hashCode methods
`hashCode` and `equals` methods must be overridden in `TableInfo`, `ViewInfo` and `ExternalTableInfo`.

mziccard:
This fixes #564

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ViewInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ViewInfo.java
@@ -25,6 +25,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Google BigQuery View Table information. BigQuery's views are logical views, not materialized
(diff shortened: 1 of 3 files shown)


ajkannan:
LGTM

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ViewInfo.java
```diff
@@ -144,6 +145,16 @@ ToStringHelper toStringHelper() {
   }
 
   @Override
+  public boolean equals(Object obj) {
+    return obj instanceof ViewInfo && Objects.equals(toPb(), ((ViewInfo) obj).toPb());
```
aozarov: Nit, you may want to consider renaming the base class, BaseTableInfo, equals and hashCode to
`baseEquals` and `baseHashCode`. Also I think `baseEquals` can work for all if you write it this way:
```java
  return obj != null && getClass().isInstance(obj) && Objects.equals(toPb(), ((BaseTableInfo) obj).toPb());
```

mziccard: I see no real benefit from renaming `equals` and `hashCode`, do you? There's no good answer to this I guess but I think that defining `equals` and `hashCode` also for abstract classes is a good practice in general. Also this PR is consistent with `JobInfo` hierarchy, so if we change here we must change also there.

aozarov:
> but I think that defining equals and hashCode also for abstract classes is a good practice in general

I would actually say the opposite. if you define it in the base class and override it only in some of the sub-classes then you can have a *symmetric* issues (B equals C but not C equals B) and can even get weird *transitive* issues. This is not an issue if you make the base class methods final.

In your case you do override all the sub-classes equals & hashCode methods, so right now we don't have an issue (other than the base-class equals is never used) but more likely to get broken in the future
if someone is not aware of it.

PR #572, opened 2016-01-22
Make Project a subclass of ProjectInfo

ajkannan:
This is an alternate proof of concept for #530.  This PR introduces the following changes:
- Project is now a subclass of ProjectInfo.
  - `ResourceManager` methods accept ProjectInfo objects but always return a Project
- A new abstract builder class is added to `ProjectInfo`.  Both `ProjectInfo.BuilderImpl` and `Project.Builder` extend this abstract Builder.  
  - The methods of `Project.Builder` delegate to `ProjectInfo.BuilderImpl` to avoid setter code duplication.
  - The abstract class is for code maintenance.  If there is a new field for projects, adding it to the abstract class will ensure that we add it to both `ProjectInfo.BuilderImpl` and `Project.Builder`. 

A user can create a `ProjectInfo` (without supplying a service object) when creating or updating a project.  Since Resource Manager methods will return `Project` objects instead of `ProjectInfo` objects, the user doesn't have to supply the service object any additional times for the typical use case.

### DIFF
--- a/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
+++ b/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
@@ -18,25 +18,110 @@ package com.google.gcloud.resourcemanager;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.util.Map;
+import java.util.Objects;
+
 /**
  * A Google Cloud Resource Manager project object.
  *
  * <p>A Project is a high-level Google Cloud Platform entity. It is a container for ACLs, APIs,
  * AppEngine Apps, VMs, and other Google Cloud Platform resources. This class' member variables are
- * immutable.  Methods that change or update the underlying Project information return a new Project
- * instance.
+ * immutable. Methods that change or update the underlying Project information return a new Project
+ * instance. {@code Project} adds a layer of service-related functionality over {@link ProjectInfo}.
  */
-public class Project {
+public class Project extends ProjectInfo {
 
-  private final ResourceManager resourceManager;
-  private final ProjectInfo info;
+  private static final long serialVersionUID = 6767630161335155133L;
 
-  /**
-   * Constructs a Project object that contains the ProjectInfo given.
-   */
-  public Project(ResourceManager resourceManager, ProjectInfo projectInfo) {
+  private final ResourceManagerOptions options;
+  private transient ResourceManager resourceManager;
+
+  public static class Builder extends ProjectInfo.Builder {
+    private final ResourceManager resourceManager;
+    private ProjectInfo.BuilderImpl infoBuilder;
+
+    Builder(ResourceManager resourceManager) {
+      this.resourceManager = resourceManager;
+      this.infoBuilder = new ProjectInfo.BuilderImpl();
+    }
+
+    Builder(Project project) {
+      this.resourceManager = project.resourceManager;
+      this.infoBuilder = new ProjectInfo.BuilderImpl(project);
+    }
+
+    @Override
+    public Builder name(String name) {
+      infoBuilder.name(name);
+      return this;
+    }
+
+    @Override
+    public Builder projectId(String projectId) {
+      infoBuilder.projectId(projectId);
+      return this;
+    }
+
+    @Override
+    public Builder addLabel(String key, String value) {
+      infoBuilder.addLabel(key, value);
+      return this;
+    }
+
+    @Override
+    public Builder removeLabel(String key) {
+      infoBuilder.removeLabel(key);
+      return this;
+    }
+
+    @Override
+    public Builder clearLabels() {
+      infoBuilder.clearLabels();
+      return this;
+    }
+
+    @Override
+    public Builder labels(Map<String, String> labels) {
+      infoBuilder.labels(labels);
+      return this;
+    }
+
+    @Override
+    Builder projectNumber(Long projectNumber) {
+      infoBuilder.projectNumber(projectNumber);
+      return this;
+    }
+
+    @Override
+    Builder state(State state) {
+      infoBuilder.state(state);
+      return this;
+    }
+
+    @Override
+    Builder createTimeMillis(Long createTimeMillis) {
+      infoBuilder.createTimeMillis(createTimeMillis);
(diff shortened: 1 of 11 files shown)


ajkannan:
/cc @mziccard @mderka @aozarov 

Thoughts?

aozarov:
I like it! :+1:

mderka:
I need a bit more time to go through this. I follow up on Monday.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
```diff
@@ -26,17 +31,97 @@
...
+      return this;
+    }
+
+    @Override
+    public Project build() {
+      return new Project(resourceManager, infoBuilder);
+    }
+  }
+
+  Project(ResourceManager resourceManager, ProjectInfo.BuilderImpl infoBuilder) {
```
mziccard: Nit: this could take a `Builder` rather than the two parameters

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
```diff
@@ -130,6 +207,38 @@ public void undelete() {
...
+    return Objects.hash(super.hashCode(), options);
+  }
+
+  private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
+    in.defaultReadObject();
+    this.resourceManager = options.service();
+  }
+
+  static Project fromPb(
+      ResourceManager resourceManager,
```
mziccard: Nit: I think this could fit in the previous line

gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/ProjectTest.java
```diff
@@ -48,13 +47,19 @@
...
 
+  private ResourceManager serviceMockReturnsOptions = createStrictMock(ResourceManager.class);
+  private ResourceManagerOptions mockOptions = createMock(ResourceManagerOptions.class);
   private ResourceManager resourceManager;
+  private Project expectedProject;
   private Project project;
 
   @Before
   public void setUp() throws Exception {
+    expect(serviceMockReturnsOptions.options()).andReturn(mockOptions).anyTimes();
```
mziccard: I think we should avoid using `anyTimes()` as this makes a lot of different behaviors pass the test. If possible I would prefer a fine grain definition of `expect`, i.e. in each test method rather than in a common `@Before` method. If you think this is not possible we could use [times(min, max)](http://easymock.org/api/org/easymock/IExpectationSetters.html#times-int-int-)

gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/ProjectTest.java
```diff
@@ -62,22 +67,53 @@ public void tearDown() throws Exception {
...
+        .projectNumber(PROJECT_NUMBER)
+        .createTimeMillis(CREATE_TIME_MILLIS)
+        .state(STATE)
+        .build();
+    assertEquals(PROJECT_ID, builtProject.projectId());
+    assertEquals(NAME, builtProject.name());
+    assertEquals(LABELS, builtProject.labels());
+    assertEquals(PROJECT_NUMBER, builtProject.projectNumber());
+    assertEquals(CREATE_TIME_MILLIS, builtProject.createTimeMillis());
+    assertEquals(STATE, builtProject.state());
```
mziccard: Should we also check that `resourceManager` has been set correctly?

mziccard:
I added a few nit comments. Great job @ajkannan I think this makes the API more usable without being verbose and keeps the code clean and maintainable. :+1:  

Btw, I can implement this in either bigquery or storage or both, just let me know!

mziccard:
Looking a bit more into this. In bigquery we have a more complex situation where Info classes are part of a hierarchy. 
For instance consider the `TableInfo` hierarchy where the `BaseTableInfo` class is extended by `TableInfo`, `ViewInfo` and `ExternalTableInfo`. Do you guys think this solution would still fit to the bigquery case?

ajkannan:
@mziccard I'll take Storage and Resource Manager, and if you can take BigQuery, that'd be great.

I'll remove the "don't merge" label on this PR after I add changes to docs, fix the tests, and address the comments above.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
ajkannan: That's true, but having two parameters does make some other package-scope code slightly simpler.  For example, in [`Project.fromPb`](https://github.com/GoogleCloudPlatform/gcloud-java/pull/572/files#diff-f71225b79e2dfd12febf71cecffed53aR238), we avoid having to explicitly set all the fields of a `Project.Builder` using the `ProjectInfo`'s fields.  For that reason, I'm inclined to keep this function signature.

ajkannan:
Incorporated comments thus far and also updated README, javadoc, and tests.  Ready for another pass.

ajkannan:
Fixed more documentation issues and added a commit that adds a `replace` method with no arguments.  Ready for another pass.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
mziccard: Makes sense!

mziccard:
@ajkannan empty replace method looks good to me!

ajkannan:
Added a commit to remove the original replace method.

aozarov:
LGTM, please fix the doc issues we discussed before merge.

ajkannan:
Fixed docs and re-tested sample in the docs.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ProjectInfo.java
```diff
@@ -115,7 +115,66 @@ static ResourceId fromPb(
     }
   }
 
-  public static class Builder {
+  public static abstract class Builder {
```
mziccard: I just noticed this: is there a reason for this to be an abstract class rather than an interface?

PR #575, opened 2016-01-22
ManagedZoneInfo class added

mderka:


gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+import org.joda.time.format.ISODateTimeFormat;
+
+import java.io.Serializable;
+import java.math.BigInteger;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * This class is a container for the managed zone metainformation. Managed zone is a resource that
```
aozarov: Personally I prefer the public [doc](https://cloud.google.com/dns/api/v1/managedZones) but reversed.

 A ManagedZone is a resource that represents a DNS zone hosted by the Cloud DNS service.
A zone is a subtree of the DNS namespace under one administrative responsibility.

I would not use metainformation as one word and I am not sure that is correct (managed zone) is meta information about a DNS zone.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+
+  /**
+   * Returns a builder for {@code ManagedZoneInfo} with an assigned {@code name} and {@code id}.
+   */
+  public static Builder builder(String name, Long id) {
+    return new Builder(name, id);
+  }
+
+  /**
+   * Returns an empty builder for {@code ManagedZoneInfo}. We use it internally in {@code toPb()}.
```
aozarov: Maybe move the comment "We use it internally in {@code toPb()}." to the constructor (and you may call it directly).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+     */
+    public Builder name(String name) {
+      this.name = checkNotNull(name);
+      return this;
+    }
+
+    /**
+     * Sets an id for the managed zone which is assigned to the managed zone by the server.
+     */
+    public Builder id(long id) {
```
aozarov: package scope.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+      this.id = id;
+      return this;
+    }
+
+    /**
+     * Sets the time when this managed zone was created.
+     */
+    Builder creationTimeMillis(long creationTimeMillis) {
+      checkArgument(creationTimeMillis >= 0, "The timestamp cannot be negative.");
```
aozarov: Because this is package scope (and basically comes from the proto, I would not have the check).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+    /**
+     * Sets a mandatory DNS name of this managed zone, for instance "example.com.".
+     */
+    public Builder dnsName(String dnsName) {
+      this.dnsName = checkNotNull(dnsName);
+      return this;
+    }
+
+    /**
+     * Sets a mandatory description for this managed zone. The value is a string of at most 1024
```
aozarov: We typically just document the enforcement but do not elaborate what part actually enforce it.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+    public Builder description(String description) {
+      this.description = checkNotNull(description);
+      return this;
+    }
+
+    /**
+     * Optionally specifies the NameServerSet for this managed zone. A NameServerSet is a set of DNS
+     * name servers that all host the same ManagedZones.
+     */
+    public Builder nameServerSet(String nameServerSet) {
```
aozarov: We need to understand what is the delimiter and if we can we should get the value as a Set<String> (or List<String> if there is a meaning for order) and apply the delimiter ourselves.

Also, we should keep the " Most users will leave this field unset." comment (though probably say "Most users will not need to specify this value" or something like that).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+     */
+    Builder nameServers(List<String> nameServers) {
+      this.nameServers.addAll(checkNotNull(nameServers));
+      return this;
+    }
+
+    /**
+     * Removes all the nameservers from the list.
+     */
+    Builder clearNameServers() {
```
aozarov: package scope and therefore you don't have to add it unless used.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+      this.nameServers.clear();
+      return this;
+    }
+
+    /**
+     * Builds the instance of ManagedZoneInfo based on the information set here.
```
aozarov: Nit, {@code ManagedZoneInfo}

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+
+  /**
+   * Returns the DNS name of this managed zone, for instance "example.com.".
+   */
+  public String dnsName() {
+    return dnsName;
+  }
+
+  /**
+   * Returns the description of this managed zone. This is at most 1024 long mandatory string
```
aozarov: Restriction is not necessary on getter.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+    return new Builder(this);
+  }
+
+  com.google.api.services.dns.model.ManagedZone toPb() {
+    com.google.api.services.dns.model.ManagedZone pb =
+        new com.google.api.services.dns.model.ManagedZone();
+    pb.setDescription(this.description());
+    pb.setDnsName(this.dnsName());
+    if (this.id() != null) {
+      pb.setId(BigInteger.valueOf(this.id()));
```
aozarov: If their ID can be more than Long.MAX_VALUE than I don't think you can use `long`.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+    }
+    pb.setName(this.name());
+    pb.setNameServers(this.nameServers());
+    pb.setNameServerSet(this.nameServerSet());
+    if (this.creationTimeMillis() != null) {
+      pb.setCreationTime(ISODateTimeFormat.dateTime()
```
aozarov: Lets make sure we have a test that cover the string parsing.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
```diff
@@ -0,0 +1,198 @@
...
+  private static final String NS3 = "name server 3";
+  private static List<String> nameServers = new LinkedList<>();
+  private static ManagedZoneInfo info;
+
+  @BeforeClass
+  public static void setUp() {
+    nameServers.add(NS1);
+    nameServers.add(NS2);
+    nameServers.add(NS3);
+    assertEquals(3, nameServers.size());
```
aozarov: No need.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
```diff
@@ -0,0 +1,198 @@
...
+    info = ManagedZoneInfo.builder(NAME, ID)
+        .creationTimeMillis(CREATION_TIME_MILLIS)
+        .dnsName(DNS_NAME)
+        .description(DESCRIPTION)
+        .nameServerSet(NAME_SERVER_SET)
+        .nameServers(nameServers)
+        .build();
+    System.out.println(info);
```
aozarov: remove.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,334 @@
...
+      this.nameServerSet = checkNotNull(nameServerSet);
+      return this;
+    }
+
+    /**
+     * Sets a list of servers that hold the information about the managed zone. This information is
+     * provided by Google Cloud DNS and is read only.
+     */
+    Builder nameServers(List<String> nameServers) {
+      this.nameServers.addAll(checkNotNull(nameServers));
```
aozarov: did you wish to replace it or `addAll`?

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
```diff
@@ -0,0 +1,198 @@
...
+  private static final String NAME_SERVER_SET = "some set";
+  private static final String NS1 = "name server 1";
+  private static final String NS2 = "name server 2";
+  private static final String NS3 = "name server 3";
+  private static List<String> nameServers = new LinkedList<>();
+  private static ManagedZoneInfo info;
+
+  @BeforeClass
+  public static void setUp() {
+    nameServers.add(NS1);
```
aozarov: Are you sure this should be static? Looks like you are modifying it in some of the tests...

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Replaced.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Done and called directly.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Done.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Check gone.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Ok, removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Acknowledged. This is a pending question. Both gcloud-node and gcloud-ruby handle it as a string without parsing. I will keep asking and dig through some code.

Documentation fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Both done.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Switched to BigInteger.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
mderka: Removed.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
mderka: Removed (used for debugging and forgotten).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Replace. Good catch.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
mderka: I believe I was modifying local variables only, but just to make this cleaner, I turned into a non-static attribute and used @Before method.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Added.

mderka:
I finished processing the comments. The PR is ready for another review pass.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
aozarov: If you were not modifying it in the tests than that is my mistake. I would than switch it  back to static and final. Also, you don't need the `@BeforeClass` setup in this case as both nameServers could be assigned in the declaration. For `nameServers` use `ImmutableList.of(NS1, NS2, NS3)`.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
```diff
@@ -0,0 +1,199 @@
...
+    assertEquals(NAME, info.name());
+    assertEquals(ID, info.id());
+    assertEquals(CREATION_TIME_MILLIS, info.creationTimeMillis());
+    assertEquals(NAME_SERVER_SET, info.nameServerSet());
+    assertEquals(DESCRIPTION, info.description());
+    assertEquals(DNS_NAME, info.dnsName());
+  }
+
+  @Test
+  public void testValidCreationTime() {
```
aozarov: I don't think this test is needed.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
```diff
@@ -0,0 +1,199 @@
...
+    ManagedZoneInfo.builder(NAME).creationTimeMillis(-1);
+    ManagedZoneInfo.builder(NAME).creationTimeMillis(0);
+    ManagedZoneInfo.builder(NAME).creationTimeMillis(Long.MAX_VALUE);
+  }
+
+  @Test
+  public void testEqualsAndNotEquals() {
+    ManagedZoneInfo clone = info.toBuilder().build();
+    assertEquals(clone, info);
```
aozarov: Nit, here and below (I also saw that you did the same in DnsRecordTest):
assert's first argument should be the *expected* value.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
```diff
@@ -0,0 +1,323 @@
...
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(name, id, creationTimeMillis, dnsName,
+        description, nameServerSet, nameServers);
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
```
aozarov: creationTimeMillis ?

aozarov:
Looks good! Just few more nits

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
mderka: Removed

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
mderka: Fixed both places

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ManagedZoneInfo.java
mderka: Added

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ManagedZoneInfoTest.java
mderka: Ok, changed back to static and final. I also got rid of the before class.

mderka:
I finished the processing the feedback. This is ready for another pass.

PR #576, opened 2016-01-23
merge from upstream, fix merge confilict and fix build on non java 8

aozarov:


garrettjonesgoogle:
For the code that we have added in this branch (gcloud-java-gax and gcloud-java-pubsub), it looks like you have only touched the pom.xml files. The diff for these files isn't available, but I'm not too concerned. I'm fine with letting the rest through without reviewing it.

PR #582, opened 2016-01-26
Use pagination in DatastoreExample

ajkannan:
Pagination is a common use case, so using it in the example is good as an example and for our own testing purposes.  This PR also fixes a docs bug and makes the README formatting better.


ajkannan:
Increased query limit size.

PR #586, opened 2016-01-26
The part of managed zone that relates DnsRecords and ManagedZoneInfo

mderka:
This is the first part which will be later extended with what relates to changes. Furthermore, we do not provide the user with options for service calls yet. This is to be implemented shortly in the subsequent phase too.


aozarov:
I think title should be renamed to describe that this PR is adding "functional" managed zone and the Dns main service. However, I suggest we split this PR to first, provide a complete Dns service interface including the options and only after sending another PR for the functional managed zone.

PR #587, opened 2016-01-26
Implemented ChangeRequest.

mderka:


gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,340 @@
...
+    DONE("done");
+
+    private final String status;
+
+    Status(String status) {
+      this.status = status;
+    }
+
+    static Status translate(String status) {
+      if ("pending".equals(status)) {
```
aozarov: I think you should either assume enum name is always equals to string value lowercased or use a similar pattern
to [that](https://github.com/GoogleCloudPlatform/gcloud-java/blob/e3dba5e9ffcb2bf812af8155f9c846aa1d274940/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/ValueType.java).

mderka: Ok, I made the assumption.

mderka:
Ready for another pass.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+
+import java.io.Serializable;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A class representing a change. A change is an atomic update to a collection of {@link DnsRecord}s
```
ajkannan: Nit: I think the first sentence sounds a bit strange because "change" is a general term.  Maybe we could either say "A class representing an atomic update to a collection..." or use "record change" instead of "change"?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+     * Creates a {@code ChangeRequest} instance populated by the values associated with this
+     * builder.
+     */
+    public ChangeRequest build() {
+      return new ChangeRequest(this);
+    }
+  }
+
+  private ChangeRequest(Builder builder) {
+    this.additions = builder.additions;
```
aozarov: copy the lists (`ImmutableList.copyOf()`)

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+
+  /**
+   * Creates a builder populated with values of this {@code ChangeRequest}.
+   */
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  /**
+   * Returns the list of {@link DnsRecord}s to be added to the zone upon executing this {@code
```
aozarov: Nit, did we say we use "submit"?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+    if (id() != null) {
+      pb.setId(id());
+    }
+    // set timestamp
+    if (startTimeMillis() != null) {
+      pb.setStartTime(ISODateTimeFormat.dateTime().withZoneUTC().print(startTimeMillis()));
+    }
+    // set status
+    if (status() != null) {
+      pb.setStatus(status().status);
```
aozarov: lets remove `status` private field. and make this one calls `toLowerCase` to be consistent with `fromPb`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+    }
+    // set a list of additions
+    if (additions() != null) {
+      LinkedList<com.google.api.services.dns.model.ResourceRecordSet> additionsPb =
```
aozarov: You can initialize an array list with an expected size, however what we should probably do, is not create any new list
but rather use [`Lists.tranform`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Lists.html#transform(java.util.List,%20com.google.common.base.Function)). see [this](https://github.com/GoogleCloudPlatform/gcloud-java/blob/3f2799def9e05353087d2f5816b93eafef8bce38/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobStatus.java#L116) as an example.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Fixed. Chose the first option (...an atomic update...).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: That was for batch, but can be done here too. Changed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Done.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Fixed using Lists.transform. I also adjusted fromPb.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+     * Sets a collection of {@link DnsRecord}s which are to be added to the zone upon executing this
+     * {@code ChangeRequest}.
+     */
+    public Builder additions(List<DnsRecord> additions) {
+      this.additions = Lists.newLinkedList(checkNotNull(additions));
+      return this;
+    }
+
+    /**
+     * Adds a {@link DnsRecord} which to be <strong>added</strong> to the zone upon executing this
```
ajkannan: "which to be" --> "to be" here and below.

Also if we use <strong> for the action verb here and below, do we want the same in `additions()` (one method up)?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+     * Clears the collection of {@link DnsRecord}s which are to be deleted from the zone upon
+     * executing this {@code ChangeRequest}.
+     */
+    public Builder clearDeletions() {
+      this.deletions.clear();
+      return this;
+    }
+
+    /**
+     * Removes a single {@link DnsRecord} from the collection of of records to be
```
ajkannan: double "of" on this line

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+     * Sets a collection of {@link DnsRecord}s which are to be deleted from the zone upon executing
+     * this {@code ChangeRequest}.
+     */
+    public Builder deletions(List<DnsRecord> deletions) {
+      this.deletions = Lists.newLinkedList(checkNotNull(deletions));
+      return this;
+    }
+
+    /**
+     * Associates a server-assigned id to this {@code ChangeRequest}.
```
ajkannan: In general we don't really care about adding javadoc for package-protected code.  It can stay if you think it adds value, but just wanted to point that out so you don't spend too much time in the future writing docs for non-visible code

mderka: Acknowledged. I wrote this one specifically for myself so to remind me that this is a server-assigned value which should be read only.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+
+import org.joda.time.DateTime;
+import org.joda.time.format.ISODateTimeFormat;
+
+import java.io.Serializable;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
```
ajkannan: Checkstyle complains that this import is in the wrong order, I think static imports go inbetween "package..." and before non-static imports.

mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    private Builder() {
+    }
+
+    /**
+     * Sets a collection of {@link DnsRecord}s which are to be added to the zone upon executing this
+     * {@code ChangeRequest}.
+     */
+    public Builder additions(List<DnsRecord> additions) {
```
ajkannan: Since a `ChangeRequest` is atomic, could we make the parameter more general (here and for `deletions`)?  For example, it could take an `Iterable` instead of a `List`.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+      for (com.google.api.services.dns.model.ResourceRecordSet addition : pb.getAdditions()) {
+        b.add(DnsRecord.fromPb(addition));
+      }
+    }
+    return b.build();
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    return (o instanceof ChangeRequest) && this.toPb().equals(((ChangeRequest) o).toPb());
```
ajkannan: `this` in `&& this.toPb().equals`is unnecessary

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ChangeRequestTest.java
```diff
@@ -0,0 +1,219 @@
...
+    assertFalse(clone.additions().contains(RECORD2));
+    assertFalse(clone.additions().contains(RECORD1));
+    assertTrue(clone.additions().isEmpty());
+    assertTrue(clone.deletions().contains(RECORD3));
+  }
+
+  @Test
+  public void testRemoveDeletion() {
+    ChangeRequest clone = CHANGE.toBuilder().removeDeletion(RECORD3).build();
+    assertFalse(clone.deletions().contains(RECORD3));
```
ajkannan: Nit: this line and the one below are slightly redundant; we could probably just choose one of them.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
```diff
@@ -0,0 +1,331 @@
...
+    // set timestamp
+    if (startTimeMillis() != null) {
+      pb.setStartTime(ISODateTimeFormat.dateTime().withZoneUTC().print(startTimeMillis()));
+    }
+    // set status
+    if (status() != null) {
+      pb.setStatus(status().status);
+    }
+    // set a list of additions
+    if (additions() != null) {
```
ajkannan: `additions()` and `deletions()` can never be null, so this check (and the one for `deletions()`) can be removed.

ajkannan:
A couple minor comments from me, looks good overall!

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Removed.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ChangeRequestTest.java
mderka: Ok, kept the isEmpty test.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ChangeRequest.java
mderka: Yes, switched to iterable.

mderka:
I pushed fixes for all the comments. Ready for another pass.

aozarov:
LGTM

PR #598, opened 2016-01-29
Pull changes from master

ajkannan:
The purpose of this PR is to reduce the delta between the update-datastore branch and master to facilitate easier merging into master when v1beta3 is released.  The last two commits include merge conflict resolution.


ajkannan:
Noticed one issue: the merge didn't take the master's up-to-date ServiceOptions.Builder.projectId(String projectId) method, instead taking the `update-datastore` version.  Will fix in next commit.

ajkannan:
In addition to fixing merge conflicts, the following changes were applied:
* Change the retry codes from HTTP codes to gRPC ordinal codes.
* Remove unnecessary "protected" modifier from two `DateTime` functions.
* Set `QueryResults.cursorAfter` to be the batch's `endCursor` when we run out of results.
* Fix update-datastore's `ServiceOptions` to accept null project ID.

PR #606, opened 2016-02-02
Added Dns methods, Zone and ZoneTest. Fixes #596.

### LINKED ISSUE #596: Add methods to DnsService interface
Complete implementation when the options PR is approved.

mderka:
Fixes #596.

### DIFF
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
@@ -524,16 +524,6 @@ public interface Dns extends Service<DnsOptions> {
   ProjectInfo getProjectInfo(ProjectOption... fields);
 
   /**
-   * Returns the current project id.
-   */
-  String getProjectId();
-
-  /**
-   * Returns the current project number.
-   */
-  BigInteger getProjectNumber();
-
-  /**
    * Submits a change request for the specified zone. The returned object contains the following
    * read-only fields supplied by the server: id, start time and status. time, id, and list of name
    * servers. The fields to be returned can be selected by {@link ChangeRequestOption}s.
@@ -566,7 +556,7 @@ public interface Dns extends Service<DnsOptions> {
    * @throws DnsException upon failure or if the zone cannot be found
    * @see <a href="https://cloud.google.com/dns/api/v1/changes/get">Cloud DNS Chages: get</a>
    */
-  ChangeRequest getChangeRequest(String changeRequestId, BigInteger zoneId,
+  ChangeRequest getChangeRequest(BigInteger zoneId, String changeRequestId,
       ChangeRequestOption... options);
 
   /**
@@ -578,7 +568,7 @@ public interface Dns extends Service<DnsOptions> {
    * @throws DnsException upon failure or if the zone cannot be found
    * @see <a href="https://cloud.google.com/dns/api/v1/changes/get">Cloud DNS Chages: get</a>
    */
-  ChangeRequest getChangeRequest(String changeRequestId, String zoneName,
+  ChangeRequest getChangeRequest(String zoneName, String changeRequestId,
       ChangeRequestOption... options);
 
   /**
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
@@ -130,8 +130,7 @@ public class Zone implements Serializable {
    * Lists all {@link DnsRecord}s associated with this zone. First searches for zone by ID and if
    * not found then by name.
    *
-   * @param options optional restriction on listing and on what fields of {@link DnsRecord} should
-   * be returned by the service
+   * @param options optional restriction on listing and fields of {@link DnsRecord}s returned
    * @return a page of DNS records
    * @throws DnsException upon failure or if the zone is not found
    * @throws NullPointerException if both zone ID and name are not initialized
@@ -191,11 +190,11 @@ public class Zone implements Serializable {
     checkNotNull(changeRequestId);
     ChangeRequest updated = null;
     if (zoneInfo.id() != null) {
-      updated = dns.getChangeRequest(changeRequestId, zoneInfo.id(), options);
+      updated = dns.getChangeRequest(zoneInfo.id(), changeRequestId, options);
     }
     if (updated == null && zoneInfo.name() != null) {
       // zone was not found by id or id is not set at all
-      updated = dns.getChangeRequest(changeRequestId, zoneInfo.name(), options);
+      updated = dns.getChangeRequest(zoneInfo.name(), changeRequestId, options);
     }
     return updated;
   }
@@ -205,8 +204,7 @@ public class Zone implements Serializable {
    * then by name. Returns a page of {@link ChangeRequest}s or {@code null} if the zone is not
    * found.
    *
-   * @param options optional restriction on listing and on what fields of {@link ChangeRequest}s
-   * should be returned by the service
+   * @param options optional restriction on listing and fields to be returned
    * @return a page of change requests
    * @throws DnsException upon failure or if the zone is not found
    * @throws NullPointerException if both zone ID and name are not initialized
@@ -236,7 +234,7 @@ public class Zone implements Serializable {
    * Returns the {@link ZoneInfo} object containing information about this zone.
    */
   public ZoneInfo info() {
-    return this.zoneInfo;
+    return zoneInfo;
   }
 
   /**
--- a/gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
+++ b/gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
@@ -101,16 +101,15 @@ public class ZoneTest {
     Zone retrieved = Zone.get(dns, ZONE_ID);
     assertSame(dns, retrieved.dns());
     assertEquals(ZONE_INFO, retrieved.info());
-    BigInteger id = null;
     try {
-      Zone.get(dns, id);
-      fail("Cannot get null zone.");
+      Zone.get(dns, (BigInteger) null);
+      fail("Cannot get by null id.");
     } catch (NullPointerException e) {
       // expected
     }
     try {
-      Zone.get(null, id);
-      fail("Cannot get null zone.");
+      Zone.get(null, new BigInteger("12"));
+      fail("Cannot get anything from null service.");
     } catch (NullPointerException e) {
       // expected
     }
@@ -126,16 +125,15 @@ public class ZoneTest {
     Zone retrieved = Zone.get(dns, ZONE_NAME);
     assertSame(dns, retrieved.dns());
     assertEquals(ZONE_INFO, retrieved.info());
-    String name = null;
     try {
-      Zone.get(dns, name);
-      fail("Cannot get null zone.");
+      Zone.get(dns, (String) null);
+      fail("Cannot get by null name.");
     } catch (NullPointerException e) {
       // expected
     }
     try {
-      Zone.get(null, name);
-      fail("Cannot get null zone.");
+      Zone.get(null, "Not null");
+      fail("Cannot get anything from null service.");
     } catch (NullPointerException e) {
       // expected
     }
@@ -195,6 +193,7 @@ public class ZoneTest {
 
   @Test
   public void listDnsRecordsByIdAndFound() {
+    @SuppressWarnings("unchecked")
     Page<DnsRecord> pageMock = createStrictMock(Page.class);
     replay(pageMock);
     expect(dns.listDnsRecords(ZONE_ID)).andReturn(pageMock);
@@ -210,6 +209,7 @@ public class ZoneTest {
 
   @Test
   public void listDnsRecordsByIdAndNotFoundAndNameSetAndFound() {
+    @SuppressWarnings("unchecked")
     Page<DnsRecord> pageMock = createStrictMock(Page.class);
     replay(pageMock);
     expect(dns.listDnsRecords(ZONE_ID)).andReturn(null);
@@ -251,6 +251,7 @@ public class ZoneTest {
 
   @Test
   public void listDnsRecordsByNameAndFound() {
+    @SuppressWarnings("unchecked")
     Page<DnsRecord> pageMock = createStrictMock(Page.class);
     replay(pageMock);
     expect(dns.listDnsRecords(ZONE_NAME)).andReturn(pageMock);
@@ -486,9 +487,9 @@ public class ZoneTest {
 
   @Test
   public void getChangeByIdAndFound() {
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID)).andReturn(CHANGE_REQUEST_AFTER);
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id())).andReturn(CHANGE_REQUEST_AFTER);
     // again for options
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(CHANGE_REQUEST_AFTER);
     replay(dns);
     ChangeRequest result = zone.getChangeRequest(CHANGE_REQUEST.id());
@@ -503,13 +504,13 @@ public class ZoneTest {
 
   @Test
   public void getChangeByIdAndNotFoundAndNameSetAndFound() {
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID)).andReturn(null);
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME))
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id())).andReturn(null);
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id()))
         .andReturn(CHANGE_REQUEST_AFTER);
     // again for options
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(null);
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(CHANGE_REQUEST_AFTER);
     replay(dns);
     ChangeRequest result = zone.getChangeRequest(CHANGE_REQUEST.id());
@@ -523,12 +524,12 @@ public class ZoneTest {
 
   @Test
   public void getChangeIdAndNotFoundAndNameSetAndNotFound() {
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID)).andReturn(null);
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME)).andReturn(null);
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id())).andReturn(null);
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id())).andReturn(null);
     // again with options
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(null);
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(null);
     replay(dns);
     ChangeRequest result = zone.getChangeRequest(CHANGE_REQUEST.id());
@@ -540,8 +541,8 @@ public class ZoneTest {
 
   @Test
   public void getChangeRequestByIdAndNotFoundAndNameNotSet() {
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID)).andReturn(null);
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_ID, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id())).andReturn(null);
+    expect(dns.getChangeRequest(ZONE_ID, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(null); // for options
     replay(dns);
     ChangeRequest result = zoneNoName.getChangeRequest(CHANGE_REQUEST.id());
@@ -554,10 +555,10 @@ public class ZoneTest {
   @Test
   public void getChangeByNameAndFound() {
     // ID is not set
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME))
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id()))
         .andReturn(CHANGE_REQUEST_AFTER);
     // again for options
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(CHANGE_REQUEST_AFTER);
     replay(dns);
     ChangeRequest result = zoneNoId.getChangeRequest(CHANGE_REQUEST.id());
@@ -570,9 +571,9 @@ public class ZoneTest {
   @Test
   public void getChangeByNameAndNotFound() {
     // ID is not set
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME)).andReturn(null);
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id())).andReturn(null);
     // again for options
-    expect(dns.getChangeRequest(CHANGE_REQUEST.id(), ZONE_NAME, CHANGE_REQUEST_FIELD_OPTIONS))
+    expect(dns.getChangeRequest(ZONE_NAME, CHANGE_REQUEST.id(), CHANGE_REQUEST_FIELD_OPTIONS))
         .andReturn(null);
     replay(dns);
     ChangeRequest result = zoneNoId.getChangeRequest(CHANGE_REQUEST.id());
@@ -666,6 +667,7 @@ public class ZoneTest {
 
   @Test
   public void listChangeRequestsByIdAndFound() {
+    @SuppressWarnings("unchecked")
     Page<ChangeRequest> pageMock = createStrictMock(Page.class);
     replay(pageMock);
     expect(dns.listChangeRequests(ZONE_ID)).andReturn(pageMock);
@@ -681,6 +683,7 @@ public class ZoneTest {
 
   @Test
   public void listChangeRequestsByIdAndNotFoundAndNameSetAndFound() {
+    @SuppressWarnings("unchecked")
     Page<ChangeRequest> pageMock = createStrictMock(Page.class);
     replay(pageMock);
     expect(dns.listChangeRequests(ZONE_ID)).andReturn(null);
@@ -724,6 +727,7 @@ public class ZoneTest {
 
   @Test
   public void listChangeRequestsByNameAndFound() {
+    @SuppressWarnings("unchecked")
     Page<ChangeRequest> pageMock = createStrictMock(Page.class);
     replay(pageMock);
     expect(dns.listChangeRequests(ZONE_NAME)).andReturn(pageMock);


mziccard:
A general comment: does it make sense to merge this PR where `Zone` and `ZoneInfo` are separate objects and then add a new one to merge them? I would prefer this to be in one single PR if there are no strong arguments against.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * object will include the following read-only fields supplied by the server: creation time, id,
+   * and list of name servers.
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/managedZones/create">Cloud DNS Managed Zones:
+   * create</a>
+   */
+  ZoneInfo create(ZoneInfo zoneInfo);
+
+  /**
+   * Retrieves the zone by the specified zone name. Returns {@code null} is the zone is not found.
```
mziccard: Nit: "Retrieves" -> "Returns"? "is" -> "if"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * The returned fields can be optionally restricted by specifying {@code ZoneFieldOptions}.
```
mziccard: Isn't it `ZoneOption`? Maybe use `@link` instead.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
     }
   }
 
-  // TODO(mderka) Add methods. Created issue #596.
+  /**
+   * Creates a new zone.
+   *
+   * @return ZoneInfo object representing the new zone's metadata. In addition to the name, dns name
+   * and description (supplied by the user within the {@code zoneInfo} parameter, the returned
```
mziccard: Missing ")"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * The returned fields can be optionally restricted by specifying {@code ZoneFieldOptions}.
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/managedZones/get">Cloud DNS Managed Zones:
+   * get</a>
+   */
+  ZoneInfo getZone(String zoneName, ZoneOption... options);
+
+  /**
+   * Retrieves the zone by the specified zone name. Returns {@code null} is the zone is not found.
```
mziccard: "zone name" -> "zone id", "is" -> "if"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * The returned fields can be optionally restricted by specifying {@code ZoneFieldOptions}.
```
mziccard: Isn't it `ZoneOption`? Maybe use `@link` instead.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/managedZones/get">Cloud DNS Managed Zones:
+   * get</a>
+   */
+  ZoneInfo getZone(BigInteger zoneId, ZoneOption... options);
```
mziccard: Just some thoughts:)

I am still unsure if we should expose this method or not. I know that the APIs expose this but do we really want to have more than one way to identify a resource? I am asking because across the services different ways of identifying resources are used (ids, names, URLs, ...) but in my mind we should try to provide a coherent an uniform identification mechanism.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+
+  /**
+   * Lists the zoned inside the project.
```
mziccard: "zoned" -> "zones"

mderka:
Regarding the general comment, we had discussed it here before doing this. You are correct that it is in part redundant and we will be taking a step back by refactoring later, but it is in fact my preferred way. By taking this detour, I am getting a much better grasp of the library architecture, internals and style, which helps me a lot (I am still a newbie here). Plus, I am already almost done. Unless you feel really strongly about it, I still prefer to merge this PR and then refactor.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * <p>This method returns zone in an unspecified order. New zones do not necessarily appear at the
```
mziccard: Nit: "zones" -> "zone"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * end of the list. Use {@link ZoneListOption} to restrict the listing to a domain name, set page
+   * size, and set page tokens.
```
mziccard: Nit: "tokens" -> "token"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @return {@code Page<Zone>}, a page of zones
```
mziccard: Is `{@code Page<Zone>}` needed?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/managedZones/list">Cloud DNS Managed Zones:
+   * list</a>
+   */
+  Page<Zone> listZones(ZoneListOption... options);
+
+  /**
+   * Deletes an existing zone identified by name. Returns true if the zone was successfully deleted
```
mziccard: Nit: true -> `{@code true}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * and false otherwise.
```
mziccard: I would say: and `{@code false}` if it was not found (this comment also applies to `@return`)

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @return {@code true} if zone was found and deleted and false otherwise
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/managedZones/delete">Cloud DNS Managed Zones:
+   * delete</a>
+   */
+  boolean delete(String zoneName); // delete does not admit any options
+
+  /**
+   * Deletes an existing zone identified by id. Returns true if the zone was successfully deleted
```
mziccard: Nit: true -> `{@code true}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * and false otherwise.
```
mziccard: I would say: and `{@code false}` if it was not found (this comment also applies to `@return`)

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: I agree that it would be ideal. However, I also think that this impuls needs to come from the service. I think that we inherently have to expose this now. Imagine a user who already has an application developed (in some other language, or in using the API directly) and that application internally works with zone IDs in a fundamental way (say persists them in the database etc). If we do not expose this, we are preventing the user from adopting gcloud-java  (or at least making it quite difficult for them).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * @see <a href="https://cloud.google.com/dns/api/v1/managedZones/delete">Cloud DNS Managed Zones:
+   * delete</a>
+   */
+  boolean delete(BigInteger zoneId); // delete does not admit any options
+
+  /**
+   * Lists the DNS records in the zone identified by name.
+   *
+   * <p>The fields to be returned, page size and page tokens can be specified using {@code
+   * DnsRecordOptions}. Returns null if the zone cannot be found.
```
mziccard: "DnsRecordOptions" -> "DnsRecordOption", null -> `{@code null}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * @see <a href="https://cloud.google.com/dns/api/v1/resourceRecordSets/list">Cloud DNS
+   * ResourceRecordSets: list</a>
+   */
+  Page<DnsRecord> listDnsRecords(String zoneName, DnsRecordListOption... options);
+
+  /**
+   * Lists the DNS records in the zone identified by ID.
+   *
+   * <p>The fields to be returned, page size and page tokens can be specified using {@code
+   * DnsRecordOptions}. Returns null if the zone cannot be found.
```
mziccard: "DnsRecordOptions" -> "DnsRecordOption", null -> `{@code null}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Yes, fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/resourceRecordSets/list">Cloud DNS
+   * ResourceRecordSets: list</a>
+   */
+  Page<DnsRecord> listDnsRecords(BigInteger zoneId, DnsRecordListOption... options);
+
+  /**
+   * Retrieves the metadata about the current project. The returned fields can be optionally
+   * restricted by specifying {@code ProjectOptions}.
```
mziccard: "ProjectOptions" -> "ProjectGetOption"

Also is there a reason for which zone options are called `ZoneOption` while project options are `ProjectGetOption`?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: This seems correct.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: I believe so. Am I wrong?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   */
+  String getProjectId();
+
+  /**
+   * Returns the current project number.
+   */
+  BigInteger getProjectNumber();
+
+  /**
+   * Submits a change requests for applying to the zone identified by ID to the service. The
```
mziccard: It might be me here but the meaning of "Submits a change requests for applying to the zone identified by ID to the service." is not clear to me.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * returned object contains the following read-only fields supplied by the server: id, start time
+   * and status. time, id, and list of name servers. The returned fields can be modified by {@code
```
mziccard: I would rephrase this as "The fields to be returned can selected by"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * ChangeRequestFieldOptions}. Returns null if the zone is not found.
```
mziccard: "ChangeRequestFieldOptions" -> "ChangeRequestOption", null -> `{@code null}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @return ChangeRequest object representing the new change request or null if zone is not found
```
mziccard: Is ChangeRequest needed? If yes I would use `{@code }` or `{@link}`. "null" -> `{@code null}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: No, unifying the name to ProjectOption.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/create">Cloud DNS Changes: create</a>
+   */
+  ChangeRequest applyChangeRequest(ChangeRequest changeRequest, BigInteger zoneId,
+                                   ChangeRequestOption... options);
```
mziccard: Too much indent (also elsewhere). I think it should be:
```java
ChangeRequest applyChangeRequest(ChangeRequest changeRequest, BigInteger zoneId,
    ChangeRequestOption... options);
```

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mziccard: I think it is redundant, return type is already showed in javadoc as part of method signature so no need to add it as `{@code ...}` in `@return`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Linked and fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * Submits a change requests for applying to the zone identified by name to the service. The
+   * returned object contains the following read-only fields supplied by the server: id, start time
+   * and status. time, id, and list of name servers. The returned fields can be modified by {@code
+   * ChangeRequestFieldOptions}. Returns null if the zone is not found.
+   *
+   * @return ChangeRequest object representing the new change request or null if zone is not found
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/create">Cloud DNS Changes: create</a>
+   */
+  ChangeRequest applyChangeRequest(ChangeRequest changeRequest, String zoneName,
```
mziccard: Same comments here for the above `applyChangeRequest`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+                                   ChangeRequestOption... options);
+
+  /**
+   * Retrieves updated information about a change request previously submitted for a zone identified
+   * by ID. Returns null if the zone or request cannot be found.
```
mziccard: "null" -> `{@code null}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * <p>The fields to be returned using {@code ChangeRequestFieldOptions}.
```
mziccard: "The fields to be returned can be specified using `{@code ChangeRequestOption}`"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/get">Cloud DNS Chages: get</a>
+   */
+  ChangeRequest getChangeRequest(ChangeRequest changeRequest, BigInteger zoneId,
+                                 ChangeRequestOption... options);
```
mziccard: Too much indent (also elsewhere). I think it should be:
```java
ChangeRequest getChangeRequest(ChangeRequest changeRequest, BigInteger zoneId,
    ChangeRequestOption... options);
```

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mziccard: A personal preference: I would let `zoneName` be the first parameter of this method.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * Submits a change requests for applying to the zone identified by ID to the service. The
+   * returned object contains the following read-only fields supplied by the server: id, start time
+   * and status. time, id, and list of name servers. The returned fields can be modified by {@code
+   * ChangeRequestFieldOptions}. Returns null if the zone is not found.
+   *
+   * @return ChangeRequest object representing the new change request or null if zone is not found
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/create">Cloud DNS Changes: create</a>
+   */
+  ChangeRequest applyChangeRequest(ChangeRequest changeRequest, BigInteger zoneId,
```
mziccard: A personal preference: I would let `zoneId` be the first parameter of this method.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+  /**
+   * Retrieves updated information about a change request previously submitted for a zone identified
+   * by ID. Returns null if the zone or request cannot be found.
+   *
+   * <p>The fields to be returned using {@code ChangeRequestFieldOptions}.
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/get">Cloud DNS Chages: get</a>
+   */
+  ChangeRequest getChangeRequest(ChangeRequest changeRequest, BigInteger zoneId,
```
mziccard: Why the `ChangeRequest` parameter?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+  /**
+   * Retrieves updated information about a change request previously submitted for a zone identified
+   * by name. Returns null if the zone or request cannot be found.
+   *
+   * <p>The fields to be returned using {@code ChangeRequestFieldOptions}.
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/get">Cloud DNS Chages: get</a>
+   */
+  ChangeRequest getChangeRequest(ChangeRequest changeRequest, String zoneName,
```
mziccard: Same comments here for the above `getChangeRequest`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+                                 ChangeRequestOption... options);
+
+  /**
+   * Lists the change requests for the zone identified by ID that were submitted to the service.
+   *
+   * <p>The sorting key for changes, fields to be returned, page and page tokens can be specified
```
mziccard: page size?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * using {@code ChangeRequestListOptions}. Note that the only sorting key currently supported is
```
mziccard: "ChangeRequestListOption" -> "ChangeRequestListOption"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * the timestamp of submitting the change request to the service.
+   *
+   * @return {@code Page<ChangeRequest>}, a page of change requests
```
mziccard: I think repeating `{@code Page<ChangeRequest>}` here is redundant

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   */
+  ChangeRequest getChangeRequest(ChangeRequest changeRequest, String zoneName,
+                                 ChangeRequestOption... options);
+
+  /**
+   * Lists the change requests for the zone identified by ID that were submitted to the service.
+   *
+   * <p>The sorting key for changes, fields to be returned, page and page tokens can be specified
+   * using {@code ChangeRequestListOptions}. Note that the only sorting key currently supported is
+   * the timestamp of submitting the change request to the service.
```
mziccard: Just to be sure I understand. [Here](https://cloud.google.com/dns/api/v1/changes/list#sortBy) I read that:

> Sorting criterion. The only supported value is change sequence. 

Also, how can the sorting key/criterion be specified? I can not find an option in `ChangeRequestListOption`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,191 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @return {@code Page<ChangeRequest>}, a page of change requests
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/list">Cloud DNS Chages: list</a>
+   */
+  Page<ChangeRequest> listChangeRequests(String zoneName, ChangeRequestListOption... options);
```
mziccard: Same comments here for the above `listChangeRequests `

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mziccard: Yes I meant: "zone" -> "zones"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed. Agreed and swapped.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Because we need to know which change request should be retrieved. Alternatively, it can be just String changeRequestId.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mziccard: I see your point but I do not agree. Your arguments makes sense in this context but is dangerous in general. We cannot design a new and more usable library that also addresses all previous usages. The question is: providing both ways is a good design choice? Is it maintainable in the long term? Does it create confusion? Are users going to benefit from this?

Also regarding the service it would be nice to know why are both id and name supported.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Removed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: No, we removed the sorting criterion option as it is not really an option as of now. I fixed the doc.

mderka: I fixed the doc.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

mziccard:
> Regarding the general comment, we had discussed it here before doing this. You are correct that it is in part redundant and we will be taking a step back by refactoring later, but it is in fact my preferred way. By taking this detour, I am getting a much better grasp of the library architecture, internals and style, which helps me a lot (I am still a newbie here). Plus, I am already almost done. Unless you feel really strongly about it, I still prefer to merge this PR and then refactor.

I see your point and I understand that this is the preferred way to get an understanding of the service. However means two commits in the project's history where the second one is basically a revert of the previous one. It's no big deal but it's not the best way :)

Also please do your best at using meaningful commit messages. Avoid messages like "Fix comments". I know it's boring but this makes reading history and writing a changelog much easier!

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Ok. Removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Right, makes sense. Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mziccard: Is there a benefit in passing the whole `ChangeRequest` object? otherwise I would prefer passing just name/id

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+  /**
+   * Constructs a {@code Zone} object that contains the given {@code zoneInfo}.
+   */
+  public Zone(Dns dns, ZoneInfo zoneInfo) {
+    this.zoneInfo = checkNotNull(zoneInfo);
+    this.dns = checkNotNull(dns);
+  }
+
+  /**
+   * Constructs a {@code Zone} object that contains meta information received from the Google Cloud
```
mziccard: In noticed that also above we refer to the content of  `XXXInfo` objects as: metadata, information or meta. I think that only "information" should be clear enough.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+   * DNS service for the provided zoneName.
```
mziccard: zoneName -> `{@code zoneName}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+   *
+   * @param zoneName Name of the zone to be searched for
```
mziccard: It think there's no need to start with a capital letter (also elsewhere)

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+   * @param options Optional restriction on what fields should be returned by the service
+   * @return Zone object containing metadata or null if not not found
```
mziccard: "null" -> `{@code null}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+   * @throws DnsException upon failure
+   */
+  public static Zone get(Dns dnsService, String zoneName,
+                         Dns.ZoneOption... options) {
```
mziccard: Does't this fit on the same line?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+  /**
+   * Constructs a {@code Zone} object that contains meta information received from the Google Cloud
+   * DNS service for the provided zoneName.
+   *
+   * @param zoneId ID of the zone to be searched for
+   * @param options Optional restriction on what fields should be returned by the service
+   * @return Zone object containing metadata or null if not not found
+   * @throws DnsException upon failure
+   */
+  public static Zone get(Dns dnsService, BigInteger zoneId,
```
mziccard: Same comments from above `get` method

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+                         Dns.ZoneOption... options) {
+    checkNotNull(zoneName);
+    checkNotNull(dnsService);
+    ZoneInfo zoneInfo = dnsService.getZone(zoneName, options);
+    return zoneInfo == null ? null : new Zone(dnsService, zoneInfo);
+  }
+
+  /**
+   * Constructs a {@code Zone} object that contains meta information received from the Google Cloud
+   * DNS service for the provided zoneName.
```
mziccard: zoneName -> `{@code zoneId}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Yes. Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+      zone = Zone.get(dns, zoneInfo.name(), options);
+    }
+    return zone;
+  }
+
+  /**
+   * Deletes the zone. The method first attempts to delete the zone by ID. If the zone is not found
+   * or id is not set, it attempts to delete by name.
+   *
+   * @return true is zone was found and deleted and false otherwise
```
mziccard: true -> `{@code true}`

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mziccard: In other services I think we throw exception if the parent entity (in this case zone) we are trying to list is not found. @aozarov @ajkannan can you confirm that?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+   * server-assigned ID or null if the zone was not found.
+   *
+   * @param options Optional restriction on what fields of {@link ChangeRequest} should be returned
+   * by the service
+   * @return ChangeRequest with server-assigned ID or null if the zone was not found.
+   * @throws DnsException upon failure
+   * @throws NullPointerException if both zone ID and name are not initialized
+   */
+  public ChangeRequest applyChangeRequest(ChangeRequest changeRequest,
+                                          Dns.ChangeRequestOption... options) {
```
mziccard: Too much indent

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+   * @param options Optional restriction on what fields of {@link ChangeRequest} should be returned
+   * by the service
+   * @return ChangeRequest with updated information of null if the change request or zone was not
+   * found.
+   * @throws DnsException upon failure
+   * @throws NullPointerException if both zone ID and name are not initialized
+   * @throws NullPointerException if the change request does not have initialized id
+   */
+  public ChangeRequest getChangeRequest(ChangeRequest changeRequest,
+                                        Dns.ChangeRequestOption... options) {
```
mziccard: Too much indent

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+      changeRequests = dns.listChangeRequests(zoneInfo.name(), options);
+    }
+    return changeRequests;
+  }
+
+  /**
+   * Check that at least one of name and ID are initialized and throw and exception if not.
+   */
+  private void checkNameOrIdNotNull() {
+    if (zoneInfo != null && zoneInfo.id() == null && zoneInfo.name() == null) {
```
mziccard: Nit: also [checkArgument](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Preconditions.html#checkArgument(boolean,%20java.lang.Object)) could be used

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,256 @@
...
+    return this.zoneInfo;
+  }
+
+  /**
+   * Returns the {@link Dns} service object associated with this managed zone.
+   */
+  public Dns dns() {
+    return this.dns;
+  }
+
```
mziccard: Remove newline

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Ok, will change it to ID.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
aozarov: My vote would be to remove the `zoneId` based operations and to go only with `zoneName` based operations. 

My reasoning:
1)  Adding it later, if needed, would be backward compatible change. 
2)  Both id and name are part of the zone information.
3) Even, if we decide to add it later, we should consider only adding it for `get` operation.
4) There would even be less of a need for such overloading when we have the functional Zone.

Other opinions: @mziccard, @ajkannan, @mderka, @jgeewax ?

mderka: This will be a bigger refactoring task and we should move it outside this PR. I opened issue #610.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Having consulted with @ajkannan, I removed the parts that can be inferred. Changed to "Submits a change request for the specified zone".

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Seems correct. I will change this.

ajkannan: I agree with @mziccard, this is the case for Storage.  Datastore and ResourceManager don't have any list-by-parent-entity I believe, so they aren't relevant.

mderka: Ok, changed here and everywhere where applicable.

mderka:
Fixed and ready for another pass.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+    return deleted;
+  }
+
+  /**
+   * Lists all {@link DnsRecord}s associated with this zone. First searches for zone by ID and if
+   * not found then by name.
+   *
+   * @param options optional restriction on listing and on what fields of {@link DnsRecord} should
+   * be returned by the service
```
aozarov: Nit, indent (4 spaces). Here and also line 209

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
```diff
@@ -0,0 +1,750 @@
...
+  public void deleteByNameAndNotFound() {
+    expect(dns.delete(ZONE_NAME)).andReturn(true);
+    replay(dns);
+    boolean result = zoneNoId.delete();
+    assertTrue(result);
+  }
+
+  @Test
+  public void listDnsRecordsByIdAndFound() {
+    Page<DnsRecord> pageMock = createStrictMock(Page.class);
```
aozarov: For here and bellow:
http://stackoverflow.com/questions/56954/easymock-how-do-i-create-a-mock-of-a-genericized-class-without-a-warning

Either of the following options is fine with me:
1) add `@SuppressWarnings("unchecked")` where needed
2) Use one the suggetions mentioned in SO
3) Add Page a simple "PageImpl" under https://github.com/GoogleCloudPlatform/gcloud-java/tree/master/gcloud-java-core/src/main/java/com/google/gcloud/testing, that gets one or more list of values.

I think Option (3) would be useful for you and others but I am fine if you choose any of the other options.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,188 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   * <p>Returns {@link ZoneInfo} object representing the new zone's information. In addition to the
+   * name, dns name and description (supplied by the user within the {@code zoneInfo} parameter),
+   * the returned object will include the following read-only fields supplied by the server:
+   * creation time, id, and list of name servers.
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/managedZones/create">Cloud DNS Managed Zones:
+   * create</a>
+   */
+  ZoneInfo create(ZoneInfo zoneInfo);
```
aozarov: I think the following methods are missing a test:
create, listZone, getProjectInfo, getProjectId, getProjectNumber (I am not sure we actually need the last 2 methods).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,188 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+   *
+   * @throws DnsException upon failure
+   * @see <a href="https://cloud.google.com/dns/api/v1/projects/get">Cloud DNS Projects: get</a>
+   */
+  ProjectInfo getProjectInfo(ProjectOption... fields);
+
+  /**
+   * Returns the current project id.
+   */
+  String getProjectId();
```
aozarov: I don't think we need `getProjectNumber` and `getProjectId` in this level.

However, at some point we should consider having them as part of the `Service` api (and probably implemented by the `BaseService`).

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
```diff
@@ -0,0 +1,750 @@
...
+  public void testGetByName() {
+    expect(dns.getZone(ZONE_NAME)).andReturn(ZONE_INFO);
+    expect(dns.getZone(ZONE_ID, ZONE_FIELD_OPTIONS)).andReturn(ZONE_INFO); // for options
+    replay(dns);
+    Zone retrieved = Zone.get(dns, ZONE_NAME);
+    assertSame(dns, retrieved.dns());
+    assertEquals(ZONE_INFO, retrieved.info());
+    String name = null;
+    try {
+      Zone.get(dns, name);
```
aozarov: use `null` directly instead of variable and fix the second check to have a non null name.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
```diff
@@ -0,0 +1,750 @@
...
+
+  @Test
+  public void testGetById() {
+    expect(dns.getZone(ZONE_ID)).andReturn(ZONE_INFO);
+    expect(dns.getZone(ZONE_ID, ZONE_FIELD_OPTIONS)).andReturn(ZONE_INFO); // for options
+    replay(dns);
+    Zone retrieved = Zone.get(dns, ZONE_ID);
+    assertSame(dns, retrieved.dns());
+    assertEquals(ZONE_INFO, retrieved.info());
+    BigInteger id = null;
```
aozarov: use `null` directly instead of variable and fix the second check to have a non null id.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -420,5 +420,188 @@ public static ChangeRequestListOption sortOrder(SortingOrder order) {
...
+  /**
+   * Retrieves updated information about a change request previously submitted for a zone identified
+   * by ID. Returns {@code null} if the request cannot be found and throws an exception if the zone
+   * does not exist. The fields to be returned using can be specified using {@link
+   * ChangeRequestOption}s.
+   *
+   * @throws DnsException upon failure or if the zone cannot be found
+   * @see <a href="https://cloud.google.com/dns/api/v1/changes/get">Cloud DNS Chages: get</a>
+   */
+  ChangeRequest getChangeRequest(String changeRequestId, String zoneName,
```
aozarov: To me me the order generic to specific (zoneName, changeRequestId) is more natural and more consistent with apply.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
mderka: As of now, the method is overloaded for the second parameter being String or BigInteger. The variable is used instead of a cast on null. However, if it is the preferred way, I am switching it into `Zone.get(dns, (String) null)`;

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Ok, refactored.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Two methods dropped. Created issue #612.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: The second two methods are dropped. The first two cannot have a test---this is just a Dns interface without an implementation. All tested methods are part of Zone, not of Dns. These two methods can be called on the service only and they cannot be called on a Zone object. They will be tested within in DnsImpl.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
mderka: Chose (1) to fix this PR now and created issue #613. I should get to it later this week.

mderka:
Edits finished, ready for another check.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: In fact, reworded and shortened.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,246 @@
...
+      changeRequests = dns.listChangeRequests(zoneInfo.name(), options);
+    }
+    return changeRequests;
+  }
+
+  /**
+   * Check that at least one of name and ID are initialized and throw and exception if not.
+   */
+  private void checkNameOrIdNotNull() {
+    checkArgument(zoneInfo != null && (zoneInfo.id() != null || zoneInfo.name() != null),
```
aozarov: See my [comment](https://github.com/GoogleCloudPlatform/gcloud-java/issues/610#issuecomment-178940712) in issue #610. We should check that we always have get name (builder can force it in compile time), and validate that
it is not null when creating the ZoneInfo. In that case, this check is not needed.

Also,  as discussed verbally, at that point we should change Zone methods to be based on name only.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -0,0 +1,246 @@
...
+        "Both zoneInfo.id and zoneInfo.name are null. This is should never happen.");
+  }
+
+  /**
+   * Returns the {@link ZoneInfo} object containing information about this zone.
+   */
+  public ZoneInfo info() {
+    return this.zoneInfo;
```
aozarov: redundant "this" here and bellow

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Yes. I will include this into refactoring done withing #610. This method will then be obsolete and I will remove it.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Fixed.

aozarov:
Feel free to merge after fixing the Nit and the null checks in the test

PR #609, opened 2016-02-02
Make bigquery functional classes extend info classes

mziccard:
This PR performs the following changes
- Make `Dataset` extend `DatasetInfo`
- Make `Table` extend `TableInfo`
- Make `Job` extend `JobInfo`
- Update READMEs and snippets
- Update `BigQueryExample`
- Update and add more tests

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Acl.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Acl.java
@@ -325,6 +325,8 @@ public final class Acl implements Serializable {
    */
   public static final class View extends Entity {
 
+    private static final long serialVersionUID = -6851072781269419383L;
+
     private final TableId id;
 
     /**
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQueryImpl.java
@@ -621,6 +621,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
         .results(transformTableData(rowsPb));
   }
 
+  @Override
   public TableDataWriteChannel writer(WriteChannelConfiguration writeChannelConfiguration) {
     return new TableDataWriteChannel(options(),
         writeChannelConfiguration.setProjectId(options().projectId()));
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/CopyJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/CopyJobConfiguration.java
@@ -204,6 +204,7 @@ public final class CopyJobConfiguration extends JobConfiguration {
     return builder.build();
   }
 
+  @Override
   com.google.api.services.bigquery.model.JobConfiguration toPb() {
     JobConfigurationTableCopy configurationPb = new JobConfigurationTableCopy();
     configurationPb.setDestinationTable(destinationTable.toPb());
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExtractJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ExtractJobConfiguration.java
@@ -225,6 +225,7 @@ public final class ExtractJobConfiguration extends JobConfiguration {
     return toBuilder().sourceTable(sourceTable().setProjectId(projectId)).build();
   }
 
+  @Override
   com.google.api.services.bigquery.model.JobConfiguration toPb() {
     JobConfigurationExtract extractConfigurationPb = new JobConfigurationExtract();
     extractConfigurationPb.setDestinationUris(destinationUris);
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadJobConfiguration.java
@@ -282,6 +282,7 @@ public final class LoadJobConfiguration extends JobConfiguration implements Load
     return toBuilder().destinationTable(destinationTable().setProjectId(projectId)).build();
   }
 
+  @Override
   com.google.api.services.bigquery.model.JobConfiguration toPb() {
     JobConfigurationLoad loadConfigurationPb = new JobConfigurationLoad();
     loadConfigurationPb.setDestinationTable(destinationTable.toPb());
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryJobConfiguration.java
@@ -472,6 +472,7 @@ public final class QueryJobConfiguration extends JobConfiguration {
     return builder.build();
   }
 
+  @Override
   com.google.api.services.bigquery.model.JobConfiguration toPb() {
     com.google.api.services.bigquery.model.JobConfiguration configurationPb =
         new com.google.api.services.bigquery.model.JobConfiguration();
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableDataWriteChannel.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/TableDataWriteChannel.java
@@ -53,6 +53,7 @@ class TableDataWriteChannel extends BaseWriteChannel<BigQueryOptions, WriteChann
     }
   }
 
+  @Override
   protected StateImpl.Builder stateBuilder() {
     return StateImpl.builder(options(), entity(), uploadId());
   }
--- a/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/WriteChannelConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/WriteChannelConfiguration.java
@@ -155,6 +155,7 @@ public class WriteChannelConfiguration implements LoadConfiguration, Serializabl
       return this;
     }
 
+    @Override
     public WriteChannelConfiguration build() {
       return new WriteChannelConfiguration(this);
     }


ajkannan:
Since the functional objects should now be serializable, we should add them to `SerializationTest`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -235,4 +233,42 @@ public Table create(String table, TableDefinition definition, BigQuery.TableOpti
   public BigQuery bigquery() {
     return bigquery;
   }
+
+  public static Builder builder(BigQuery bigquery, DatasetId datasetId) {
```
ajkannan: We can probably package scope or remove this because we don't expect users to ever have to build a functional object from scratch. They'll just be handed one if they use the BigQuery service object at any point.  I believe the parallel builder method for Resource Manager's Project is package scoped, and I didn't include it in the Storage PR that's out right now.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -235,4 +233,42 @@ public Table create(String table, TableDefinition definition, BigQuery.TableOpti
...
+    return new Builder(bigquery).datasetId(datasetId);
+  }
+
+  /**
+   * Returns a builder for a {@code DatasetInfo} object given it's user-defined id.
+   */
+  public static Builder builder(BigQuery bigquery, String datasetId) {
```
ajkannan: Package scope or remove (if we agree on the comment above)

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/DatasetInfo.java
```diff
@@ -380,65 +446,27 @@ Dataset toPb() {
   }
 
   /**
-   * Returns a builder for the DatasetInfo object given it's user-defined id.
+   * Returns a builder for a {@code DatasetInfo} object given it's identity.
```
ajkannan: Nit: I feel that "id" sounds better than "identity" (applies here and also in places unmodified by this PR).

README.md
```diff
@@ -128,18 +128,20 @@ must [supply credentials](#authentication) and a project ID if running this snip
...
+import com.google.gcloud.bigquery.Table;
 import com.google.gcloud.bigquery.TableId;
 import com.google.gcloud.bigquery.TableInfo;
 
 BigQuery bigquery = BigQueryOptions.defaultInstance().service();
 TableId tableId = TableId.of("dataset", "table");
-TableInfo info = bigquery.getTable(tableId);
-if (info == null) {
+Table table = bigquery.getTable(tableId);
+if (table == null) {
```
aozarov: I wonder if, as an example, the 'else' part is needed (vs, create if needed and upload data...)

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/BigQuery.java
```diff
@@ -535,54 +534,50 @@ public static QueryResultsOption maxWaitTime(long maxWaitTime) {
...
    */
-  TableInfo getTable(TableId tableId, TableOption... options)
-      throws BigQueryException;
+  Table getTable(TableId tableId, TableOption... options) throws BigQueryException;
 
   /**
    * Lists the tables in the dataset. This method returns partial information on each table
-   * ({@link TableInfo#tableId()}, {@link TableInfo#friendlyName()}, {@link TableInfo#id()} and
-   * type, which is part of {@link TableInfo#definition()}). To get complete information use either
+   * ({@link TableInfo#tableId()}, {@link Table#friendlyName()}, {@link Table#id()} and type, which
```
aozarov: Any reason for not also chaning `TableInfo#tableId` (here and below).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Job.java
```diff
@@ -53,15 +123,7 @@ public Job(BigQuery bigquery, JobInfo info) {
    * @throws BigQueryException upon failure
    */
   public static Job get(BigQuery bigquery, String job, BigQuery.JobOption... options) {
```
ajkannan: Do you think we can remove static methods like this one from functional classes?  Before this PR it made sense, because the service object didn't return funcitonal objects.  Now that the service object's `get` method returns a functional object, these methods are obsolete.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -35,89 +30,108 @@
...
-  private final BigQuery bigquery;
-  private final DatasetInfo info;
+  private static final long serialVersionUID = -4272921483363065593L;
 
-  private static class TablePageFetcher implements PageImpl.NextPageFetcher<Table> {
+  private final BigQueryOptions options;
+  private transient BigQuery bigquery;
 
-    private static final long serialVersionUID = 6906197848579250598L;
+  static final class Builder extends DatasetInfo.Builder {
```
ajkannan: Are there access issues when making this class package scope but including it in the function signature for public methods like `toBuilder()`?

Even if not, do you think there's a benefit to making the builders for the functional classes public?  As it stands, the javadoc won't have a `Dataset.Builder` page, just a `DatasetInfo.Builder` page.  As a result, the `Dataset.toBuilder()` function links to the `DatasetInfo.Builder`, which has a `build()` method that returns a `DatasetInfo` instead of a `Dataset`.  A user could interpret this to mean that if they call `someDataset.toBuilder().build()`, they will get a `DatasetInfo` and not necessarily a `Dataset`.

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
```diff
@@ -49,219 +61,291 @@
...
   @After
   public void tearDown() throws Exception {
-    verify(bigquery);
+    verify(bigquery, serviceMockReturnsOptions);
   }
 
   @Test
-  public void testInfo() throws Exception {
-    assertEquals(DATASET_INFO, dataset.info());
+  public void testBuilder() {
```
ajkannan: Do you think it'd be useful to also explicitly test the builder constructor that takes Pb objects as parameters?

ajkannan:
@mziccard took a first pass, looks good!  Just some minor comments from me.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
aozarov: I am +1 (this is also in-line with my earlier preference not to have a public constructor to the functional object that gets the info object, but I think now the case is stronger).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/DatasetInfo.java
```diff
@@ -380,65 +446,27 @@ Dataset toPb() {
...
    */
-  public static Builder builder(String projectId, String datasetId) {
-    return new Builder().datasetId(DatasetId.of(projectId, datasetId));
+  public static Builder builder(String datasetId) {
+    return builder(DatasetId.of(datasetId));
   }
 
   /**
-   * Returns a builder for the DatasetInfo object given it's identity.
+   * Returns a builder for the DatasetInfo object given it's project and user-defined id.
```
aozarov: "given it's project and dataset ids" ?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
mziccard: Yes sure, in fact it's package scope in `Table` and `Job`.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/DatasetInfo.java
mziccard: I do not agree. Across the library I have been careful to refer to `XXXId` classes/objects as "identity" not to make confusion with the `id()` of info classes and the String itself provided by the user to identify the entity (i.e. a job has an user-provided id, it's not called name in BigQuery).

README.md
mziccard: This is consistent with the storage example. I don't see this as a problem until it's understandable.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
mziccard: Yes we must make this public.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Job.java
mziccard: Yes I think we can remove this. We should remove such methods also in gcloud-java-resourcemanager

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/DatasetTest.java
mziccard: There's no such constructor in functional objects. We should rather test `fromPb`.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
mziccard: We can remove `builder` methods in functional classes. We should remove them also in `resourcemanager`

mziccard:
Comments addressed

ajkannan:
Looks good, one comment from me about removing the functional class builder() methods entirely.  I also think we should manually test the snippets in the READMEs/package-info to make sure that there aren't any access issues, missing imports, or typos.

I'll take care of comments that referenced Resource Manager (ie removing the package scope builder() methods from Resource Manager's Project class and making sure any Project.staticMethods() are removed as appropriate).

README.md
aozarov: Yes, it is just an example and it is readable.
However, I do think that the propose change is better for the following reasons:
1) slightly shorter
2) save some code indentation
3) more logical flow (and therefore example). create a table if does not exists and populate with data rather than either create a table if not exists or put data if exists.

Yes, if we decide to change it here we should also change it in other places were we apply a similar pattern.
We do not have to do it in this PR or at all if you disagree with my reasons.

ajkannan: +1 to removing the "else" either in this PR or in a subsequent one.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/JobInfo.java
```diff
@@ -78,16 +88,44 @@ public JobInfo apply(Job pb) {
...
-  private final String etag;
-  private final String id;
-  private final JobId jobId;
-  private final String selfLink;
-  private final JobStatus status;
-  private final JobStatistics statistics;
-  private final String userEmail;
-  private final JobConfiguration configuration;
+  /**
+   * Base class for a {@code JobInfo} builder.
```
aozarov: Nit. Though technically correct, I wonder if that is, "Base class", a good javadoc for the user which is unaware
of the internal implementation for this class builder.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -235,4 +219,39 @@ public Table create(String table, TableDefinition definition, BigQuery.TableOpti
   public BigQuery bigquery() {
     return bigquery;
   }
+
+  static Builder builder(BigQuery bigquery, DatasetId datasetId) {
+    return new Builder(bigquery).datasetId(datasetId);
```
aozarov: Nit (fine as is), but at this point we can change the constructor to get also `datasetId` and replace DatasetInfo.Builder empty constructor to one that accepts datasetId.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Table.java
```diff
@@ -193,15 +228,15 @@ Job copy(String destinationDataset, String destinationTable, BigQuery.JobOption.
 
   /**
    * Starts a BigQuery Job to copy the current table to the provided destination table. Returns the
-   * started {@link Job} object.
+   * started {@link Job} object. ddd
```
aozarov: ddd?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
```diff
@@ -35,111 +30,108 @@
...
 
-    private final BigQueryOptions options;
-    private final Page<TableInfo> infoPage;
+    private final BigQuery bigquery;
+    private final DatasetInfo.BuilderImpl infoBuilder;
 
-    TablePageFetcher(BigQueryOptions options, Page<TableInfo> infoPage) {
-      this.options = options;
-      this.infoPage = infoPage;
+    private Builder(BigQuery bigquery) {
```
ajkannan: If we remove the builder() methods completely, we can also remove this constructor.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
aozarov: My comment above is about the Builder's constructors and not about the static builder methods.
As for the static builder methods of functional objects, I do think they should not be public, however
I don't mind if we also keep them as package-scope as long as they are used by our non-test code (which does
not seem to be the case here, and therefore should be removed).

aozarov:
Looks great! I really like this change.
Few nits and suggestions. 
Also, unrelated:

Missing some `@Override` annotation
WriteChannelConfiguration
TableDataWriteChannel
QueryJobConfiguration
LoadJobConfiguration
ExtractJobConfiguration
CopyJobConfiguration
BigQueryImpl

Missing servialVersionUID
Acl.View

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
mziccard: How do we create functional objects in tests if we remove this constructor?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Table.java
mziccard: It seemed something important to remind the user of. If you don't like it I'll remove it :)

mziccard:
Comments fixed and opened #618

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Dataset.java
ajkannan: In resource manager and storage, I've been using the package-private constructor (e.g. `new Blob(options, new BlobInfo.BuilderImpl(BLOB_INFO))`, and to test the builder itself I did something [a little round about](https://github.com/ajkannan/gcloud-java/commit/75dbbf8263003e5df370c74bd2a96942b36d35ff#diff-81eda95fc1298d0b4b24c8561147a3d5R86).  If it makes testing way more convenient, we can keep it though.

Issue #615, opened 2016-02-03
Should we move Tuple<X, Y> from Rpc interfaces to common level?


mderka:
Whenever we need to retrieve data in pages, the Rpc interface contains a nested class `Tupe<X,Y>` which is afaik used to associate a page token with page iterable data. It is repeated code which I found in `BigQueryRpc`, `ProjectManagerRpc`, and I am now creating another copy in `DnsRpc`. We should consider moving it to the core.

mderka:
Perhaps could be done together with #613.

ajkannan:
It was mentioned in the discussion of #315, but we decided against adding it because we didn't want to expose a Tuple class.

mziccard:
I am still for not exposing it.

mderka:
Ok, got it. Then we can close this.

Issue #610, opened 2016-02-02
Should we overload methods for manipulating managed zone for both id and name?


mderka:
The service API allows for identifying managed zones by both unsigned long id and String name. Our original intention was to overload methods for manipulating managed zones to accept both BigInteger id or String name. We are currently leaning towards not doing this, treating id as read-only value and identifying zones by name only for the purposes of manipulation (get, delete, apply change, list changes). 

Also, there is a suggestion for providing ID as String instead of BigInteger which seems to be consistent with gcloud-python and other services.

Opening this base on discussion in PR #606.

aozarov:
+1 for not overloading. My reasons are mentioned [here](https://github.com/GoogleCloudPlatform/gcloud-java/pull/606#discussion_r51618080).

Also, +1 for changing the ID from `BigInteger` to `String` (which is more consistent with other services
that has both internal/read-only id and external id, such as project in resource-manager and bucket in storage).

aozarov:
Also, as part of this change (making dns id read-only) we should mandate that when ZoneInfo is constructed name would never be null. I think this makes sense and would work well with the Zone changes.

mderka:
While doing this, add options for zone create method.

PR #616, opened 2016-02-03
Completes DnsRpc interface by adding methods and doc. Implements DefaultDnsRpc.

### LINKED ISSUE #594: Add methods to DnsServiceRpc

mderka:
Closes #594.

### DIFF
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
@@ -1,6 +1,8 @@
 package com.google.gcloud.spi;
 
+import static com.google.gcloud.spi.DnsRpc.ListResult.of;
 import static com.google.gcloud.spi.DnsRpc.Option.DNS_NAME;
+import static com.google.gcloud.spi.DnsRpc.Option.NAME;
 import static com.google.gcloud.spi.DnsRpc.Option.DNS_TYPE;
 import static com.google.gcloud.spi.DnsRpc.Option.FIELDS;
 import static com.google.gcloud.spi.DnsRpc.Option.PAGE_SIZE;
@@ -30,7 +32,7 @@ import java.util.Map;
  */
 public class DefaultDnsRpc implements DnsRpc {
 
-  private static final String SORTING_KEY = "changeSequence";
+  private static final String SORT_BY = "changeSequence";
   private final Dns dns;
   private final DnsOptions options;
 
@@ -77,17 +79,16 @@ public class DefaultDnsRpc implements DnsRpc {
   }
 
   @Override
-  public Tuple<String, Iterable<ManagedZone>> listZones(Map<Option, ?> options)
-      throws DnsException {
+  public ListResult<ManagedZone> listZones(Map<Option, ?> options) throws DnsException {
     // fields, page token, page size
     try {
       ManagedZonesListResponse zoneList = dns.managedZones().list(this.options.projectId())
           .setFields(FIELDS.getString(options))
           .setMaxResults(PAGE_SIZE.getInt(options))
+          .setDnsName(DNS_NAME.getString(options))
           .setPageToken(PAGE_TOKEN.getString(options))
           .execute();
-      return Tuple.<String, Iterable<ManagedZone>>of(zoneList.getNextPageToken(),
-          zoneList.getManagedZones());
+      return of(zoneList.getNextPageToken(), zoneList.getManagedZones());
     } catch (IOException ex) {
       throw translate(ex);
     }
@@ -108,8 +109,8 @@ public class DefaultDnsRpc implements DnsRpc {
   }
 
   @Override
-  public Tuple<String, Iterable<ResourceRecordSet>> listDnsRecords(String zoneName,
-      Map<Option, ?> options) throws DnsException {
+  public ListResult<ResourceRecordSet> listDnsRecords(String zoneName, Map<Option, ?> options)
+      throws DnsException {
     // options are fields, page token, dns name, type
     try {
       ResourceRecordSetsListResponse response = dns.resourceRecordSets()
@@ -117,11 +118,10 @@ public class DefaultDnsRpc implements DnsRpc {
           .setFields(FIELDS.getString(options))
           .setPageToken(PAGE_TOKEN.getString(options))
           .setMaxResults(PAGE_SIZE.getInt(options))
-          .setName(DNS_NAME.getString(options))
+          .setName(NAME.getString(options))
           .setType(DNS_TYPE.getString(options))
           .execute();
-      return Tuple.<String, Iterable<ResourceRecordSet>>of(response.getNextPageToken(),
-          response.getRrsets());
+      return of(response.getNextPageToken(), response.getRrsets());
     } catch (IOException ex) {
       throw translate(ex);
     }
@@ -159,8 +159,7 @@ public class DefaultDnsRpc implements DnsRpc {
     } catch (IOException ex) {
       DnsException serviceException = translate(ex);
       if (serviceException.code() == HTTP_NOT_FOUND) {
-        // message format "The 'parameters.changeId' resource named '4' does not exist."
-        if (serviceException.getMessage().contains("changeId")) {
+        if (serviceException.location().equals("entity.parameters.changeId")) {
           // the change id was not found, but the zone exists
           return null;
         }
@@ -171,7 +170,7 @@ public class DefaultDnsRpc implements DnsRpc {
   }
 
   @Override
-  public Tuple<String, Iterable<Change>> listChangeRequests(String zoneName, Map<Option, ?> options)
+  public ListResult<Change> listChangeRequests(String zoneName, Map<Option, ?> options)
       throws DnsException {
     // options are fields, page token, page size, sort order
     try {
@@ -181,10 +180,10 @@ public class DefaultDnsRpc implements DnsRpc {
           .setPageToken(PAGE_TOKEN.getString(options));
       if (SORTING_ORDER.getString(options) != null) {
         // todo check and change if more sorting options are implemented, issue #604
-        request = request.setSortBy(SORTING_KEY).setSortOrder(SORTING_ORDER.getString(options));
+        request = request.setSortBy(SORT_BY).setSortOrder(SORTING_ORDER.getString(options));
       }
       ChangesListResponse response = request.execute();
-      return Tuple.<String, Iterable<Change>>of(response.getNextPageToken(), response.getChanges());
+      return ListResult.of(response.getNextPageToken(), response.getChanges());
     } catch (IOException ex) {
       throw translate(ex);
     }
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
@@ -20,6 +20,7 @@ import com.google.api.services.dns.model.Change;
 import com.google.api.services.dns.model.ManagedZone;
 import com.google.api.services.dns.model.Project;
 import com.google.api.services.dns.model.ResourceRecordSet;
+import com.google.common.collect.ImmutableList;
 import com.google.gcloud.dns.DnsException;
 
 import java.util.Map;
@@ -28,9 +29,10 @@ public interface DnsRpc {
 
   enum Option {
     FIELDS("fields"),
-    PAGE_SIZE("maxSize"),
+    PAGE_SIZE("maxResults"),
     PAGE_TOKEN("pageToken"),
     DNS_NAME("dnsName"),
+    NAME("name"),
     DNS_TYPE("type"),
     SORTING_ORDER("sortOrder");
 
@@ -58,26 +60,26 @@ public interface DnsRpc {
     }
   }
 
-  class Tuple<X, Y> {
+  class ListResult<T> {
 
-    private final X x;
-    private final Y y;
+    private final Iterable<T> results;
+    private final String pageToken;
 
-    private Tuple(X x, Y y) {
-      this.x = x;
-      this.y = y;
+    public ListResult(String pageToken, Iterable<T> results) {
+      this.results = ImmutableList.copyOf(results);
+      this.pageToken = pageToken;
     }
 
-    public static <X, Y> Tuple<X, Y> of(X x, Y y) {
-      return new Tuple<>(x, y);
+    public static <T> ListResult<T> of(String pageToken, Iterable<T> list) {
+      return new ListResult<>(pageToken, list);
     }
 
-    public X x() {
-      return x;
+    public Iterable<T> results() {
+      return results;
     }
 
-    public Y y() {
-      return y;
+    public String pageToken() {
+      return pageToken;
     }
   }
 
@@ -106,7 +108,7 @@ public interface DnsRpc {
    * @param options a map of options for the service call
    * @throws DnsException upon failure
    */
-  Tuple<String, Iterable<ManagedZone>> listZones(Map<Option, ?> options) throws DnsException;
+  ListResult<ManagedZone> listZones(Map<Option, ?> options) throws DnsException;
 
   /**
    * Deletes the zone identified by the name.
@@ -123,12 +125,12 @@ public interface DnsRpc {
    * @param options a map of options for the service call
    * @throws DnsException upon failure or if zone was not found
    */
-  Tuple<String, Iterable<ResourceRecordSet>> listDnsRecords(String zoneName,
-      Map<Option, ?> options) throws DnsException;
+  ListResult<ResourceRecordSet> listDnsRecords(String zoneName, Map<Option, ?> options)
+      throws DnsException;
 
   /**
    * Returns information about the current project.
-   * 
+   *
    * @param options a map of options for the service call
    * @return up-to-date project information
    * @throws DnsException upon failure or if the project is not found
@@ -136,7 +138,7 @@ public interface DnsRpc {
   Project getProject(Map<Option, ?> options) throws DnsException;
 
   /**
-   * Applies change request to a zone. 
+   * Applies change request to a zone.
    *
    * @param zoneName the name of a zone to which the {@code Change} should be applied
    * @param changeRequest change to be applied
@@ -144,7 +146,7 @@ public interface DnsRpc {
    * @return updated change object with server-assigned ID
    * @throws DnsException upon failure or if zone was not found
    */
-  Change applyChangeRequest(String zoneName, Change changeRequest, Map<Option, ?> options) 
+  Change applyChangeRequest(String zoneName, Change changeRequest, Map<Option, ?> options)
       throws DnsException;
 
   /**
@@ -156,7 +158,7 @@ public interface DnsRpc {
    * @return up-to-date change object or {@code null} if change was not found
    * @throws DnsException upon failure or if zone was not found
    */
-  Change getChangeRequest(String zoneName, String changeRequestId, Map<Option, ?> options) 
+  Change getChangeRequest(String zoneName, String changeRequestId, Map<Option, ?> options)
       throws DnsException;
 
   /**
@@ -166,6 +168,6 @@ public interface DnsRpc {
    * @param options a map of options for the service call
    * @throws DnsException upon failure or if zone was not found
    */
-  Tuple<String, Iterable<Change>> listChangeRequests(String zoneName, Map<Option, ?> options) 
+  ListResult<Change> listChangeRequests(String zoneName, Map<Option, ?> options)
       throws DnsException;
 }
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
@@ -230,7 +230,7 @@ public interface Dns extends Service<DnsOptions> {
      * Restricts the list to only DNS records with this fully qualified domain name.
      */
     public static DnsRecordListOption dnsName(String dnsName) {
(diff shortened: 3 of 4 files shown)


mderka:
DefaultDnsRpc is to be tested within integration tests in #617.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,178 @@
...
+      throw translate(ex);
+    }
+  }
+
+  @Override
+  public ManagedZone getZone(String zoneName, Map<Option, ?> options) throws DnsException {
+    // just fields option
+    try {
+      return dns.managedZones().get(this.options.projectId(), zoneName)
+          .setFields(FIELDS.getString(options)).execute();
```
mziccard: I would put `.execute()` on a new line, to have builder-like wrapping

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,178 @@
...
+    } catch (IOException ex) {
+      throw translate(ex);
```
mziccard: Should't we handle the not found case and return `null?
Something like:
```java
} catch (IOException ex) {
  DnsException serviceException = translate(ex);
  if (serviceException.code() == HTTP_NOT_FOUND) {
    return null;
  }
  throw serviceException;
}
```

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+
+  @Override
+  public Project getProject(Map<Option, ?> options) throws DnsException {
+    try {
+      return dns.projects().get(this.options.projectId())
+          .setFields(FIELDS.getString(options)).execute();
+    } catch (IOException ex) {
+      throw translate(ex);
```
mziccard: Should we return `null` for not found or throw exception in this case?
I would return `null` in other get calls (for consistency with other modules) but here a NOT FOUND project seems to me a more noteworthy error.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @Override
+  public Change getChangeRequest(String zoneName, String changeRequestId, Map<Option, ?> options)
+      throws DnsException {
+    try {
+      return dns.changes().get(this.options.projectId(), zoneName, changeRequestId)
+          .setFields(FIELDS.getString(options))
+          .execute();
+    } catch (IOException ex) {
+      throw translate(ex);
```
mziccard: Should't we handle the not found case and return `null?
Something like:
```java
} catch (IOException ex) {
  DnsException serviceException = translate(ex);
  if (serviceException.code() == HTTP_NOT_FOUND) {
    return null;
  }
  throw serviceException;
}
```

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,178 @@
...
+      throws DnsException {
+    // options are fields, page token, page size, sort order
+    try {
+      Dns.Changes.List request = dns.changes().list(this.options.projectId(), zoneName)
+          .setFields(FIELDS.getString(options))
+          .setMaxResults(PAGE_SIZE.getInt(options))
+          .setPageToken(PAGE_TOKEN.getString(options));
+      if (SORTING_ORDER.getString(options) != null) {
+        // this needs to be checked and changed if more sorting options are implemented, issue #604
+        String key = "changeSequence";
```
mziccard: maybe make this a `static final` constant?

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,178 @@
...
+  public Tuple<String, Iterable<Change>> listChangeRequests(String zoneName, Map<Option, ?> options)
+      throws DnsException {
+    // options are fields, page token, page size, sort order
+    try {
+      Dns.Changes.List request = dns.changes().list(this.options.projectId(), zoneName)
+          .setFields(FIELDS.getString(options))
+          .setMaxResults(PAGE_SIZE.getInt(options))
+          .setPageToken(PAGE_TOKEN.getString(options));
+      if (SORTING_ORDER.getString(options) != null) {
+        // this needs to be checked and changed if more sorting options are implemented, issue #604
```
mziccard: what about adding a `todo`?

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+    public Y y() {
+      return y;
+    }
+  }
+
+  /**
+   * Creates a new zone.
+   *
+   * @param zone a zone to be created
+   * @return Updated {@link ManagedZone} object
```
mziccard: Do `@link` to external libraries actually work?

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   * @throws DnsException upon failure
+   */
+  ManagedZone create(ManagedZone zone) throws DnsException;
+
+  /**
+   * Retrieves and returns an existing zone.
+   *
+   * @param zoneName name of the zone to be returned
+   * @param options a map of options for the service call
+   * @return a zone or {@code null} if not found
```
mziccard: This is not the case, see my comment on implementation

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   * @throws DnsException upon failure
+   */
+  boolean deleteZone(String zoneName) throws DnsException;
+
+  /**
+   * Lists DNS records for a given zone.
+   *
+   * @param zoneName name of the zone to be listed
+   * @param options a map of options for the service call
+   * @throws DnsException upon failure or if zone not found
```
mziccard: "zone was not found"

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+      Map<Option, ?> options) throws DnsException;
+
+  /**
+   * Returns information about the current project.
+   * 
+   * @param options a map of options for the service call
+   * @return up-to-date project information
+   * @throws DnsException upon failure
+   */
+  Project getProject(Map<Option, ?> options) throws DnsException;
```
mziccard: What happens if not found? See my comment above on implementation

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+
+  /**
+   * Applies change request to a zone. 
+   *
+   * @param zoneName the name of a zone to which the {@link Change} should be applied
```
mziccard: See my comment above on `{@link}` for external libraries

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   * @param changeRequest change to be applied
+   * @param options a map of options for the service call
+   * @return updated change object with server-assigned ID
+   * @throws DnsException upon failure or if zone not found
```
mziccard: "zone was not found"

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   */
+  Change applyChangeRequest(String zoneName, Change changeRequest, Map<Option, ?> options) 
+      throws DnsException;
+
+  /**
+   * Returns an existing change request.
+   *
+   * @param zoneName the name of a zone to which the {@link Change} was be applied
```
mziccard: See my comment above on `{@link}` for external libraries

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   * @param changeRequestId the unique id assigned to the change by the server
+   * @param options a map of options for the service call
+   * @return up-to-date change object
+   * @throws DnsException upon failure or if zone not found
```
mziccard: "zone was not found"

If we do (see my comment on implementation) we should document that we return null if the change was not found

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   */
+  Change getChangeRequest(String zoneName, String changeRequestId, Map<Option, ?> options) 
+      throws DnsException;
+
+  /**
+   * List an existing change requests for a zone.
```
mziccard: remove "an"

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   *
+   * @param zoneName the name of a zone to which the {@link Change}s were be applied
```
mziccard: See my comment above on `{@link}` for external libraries

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
...
+   * @param options a map of options for the service call
+   * @throws DnsException upon failure or if zone not found
```
mziccard: "zone was not found"

mziccard:
Not related to this PR. I think that the method [Dns.getProjectInfo()](https://github.com/GoogleCloudPlatform/gcloud-java/blob/dns-alpha/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java#L524) should be called `getProject`.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: Ok.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: My bad, forgotten. Fixing everywhere.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: added

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: Removed.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: removed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: fixed and updated doc to include null

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: removed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: removed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: fixed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: I think this should never happen in the first place. However, this is kind of the scenario when the parent container does not exist, is it not? Furthermore, it feels consistent with other parts of the library. Say that you want to fetch a bucket from non-existent project using Storage, you will get an exception as well, right?

The service call itself gives us 400 error.

mderka:
Fixed all except getProject()---see above. 

Regarding `Dns.getProjectInfo()`, I would agree if `Project` was functional, but it really is just a container for project quota. There are no methods that you can call on it which would contact the service. Would that not be a bit confusing?

mderka:
Ok. Please, check this despite the coverall drop. See the top comment regarding tests.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: (I originally typed this below) TL;DR: I would do an exception.

I think this should never happen in the first place. However, this is kind of the scenario when the parent container does not exist, is it not? Furthermore, it feels consistent with other parts of the library. Say that you want to fetch a bucket from non-existent project using Storage, you will get an exception as well, right?

The service call itself gives us 400 error.

For now, I at least specified it in the doc.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -16,6 +16,12 @@
 
 package com.google.gcloud.spi;
 
+import com.google.api.services.dns.model.Change;
+import com.google.api.services.dns.model.ManagedZone;
+import com.google.api.services.dns.model.Project;
+import com.google.api.services.dns.model.ResourceRecordSet;
+import com.google.gcloud.dns.DnsException;
+
```
aozarov: I think the following RPC string value need to change:
"maxSize" -> "maxResults"
"dnsName" -> "name"

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
```diff
@@ -52,5 +58,114 @@ Integer getInt(Map<Option, ?> options) {
     }
   }
 
-  //TODO(mderka) add supported operations. Created issue #594.
+  class Tuple<X, Y> {
```
aozarov: Per your issue #615, I think in this case API can look better if you have a specific 
`ListResult<T>` class that holds Iterable<T> for values and String cursor as it seems the Tuple
is used only for listing and in a consistent way.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,192 @@
...
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * A default implementation of the DnsRpc interface.
+ */
+public class DefaultDnsRpc implements DnsRpc {
+
+  private static final String SORTING_KEY = "changeSequence";
```
aozarov: Nit, at this layer I think SORT_BY would be a better name.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,192 @@
...
+    HttpRequestInitializer initializer = options.httpRequestInitializer();
+    this.dns = new Dns.Builder(transport, new JacksonFactory(), initializer)
+        .setRootUrl(options.host())
+        .setApplicationName(options.applicationName())
+        .build();
+    this.options = options;
+  }
+
+  @Override
+  public ManagedZone create(ManagedZone zone) throws DnsException {
```
aozarov: I now wonder why we didn't provide options also for create (in the API and RPC layer)...

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
```diff
@@ -0,0 +1,192 @@
...
+      throws DnsException {
+    try {
+      return dns.changes().get(this.options.projectId(), zoneName, changeRequestId)
+          .setFields(FIELDS.getString(options))
+          .execute();
+    } catch (IOException ex) {
+      DnsException serviceException = translate(ex);
+      if (serviceException.code() == HTTP_NOT_FOUND) {
+        // message format "The 'parameters.changeId' resource named '4' does not exist."
+        if (serviceException.getMessage().contains("changeId")) {
```
aozarov: Re error-codes, can we add that to the list. There should be a better way to detect that (e.g. different reason).

mziccard:
> Regarding Dns.getProjectInfo(), I would agree if Project was functional, but it really is just a container for project quota. There are no methods that you can call on it which would contact the service. Would that not be a bit confusing?

Before our most recent changes that make functional objects extend infos we used to return `XXXInfo` objects while the methods were called simply `getXXX`. I think we should stick to this convention, adding the `Info` suffix does not make the method clearer to me.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: Changed

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: Reason is the same. The responses are identical up to the message and "location" attribute of "debugInfo". I switched it to location.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: Agreed. Refactored.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DnsRpc.java
mderka: Yes for maxResults. Not really for dnsName. The service allows two options: dnsName (e.g. for listing zones) and name (e.g. for listing dns records), so we need both options. I made a mistake of tossing both into dnsName. I am adding the option name option.

gcloud-java-dns/src/main/java/com/google/gcloud/spi/DefaultDnsRpc.java
mderka: We did not have a special reason for it, probably just forgot. I will add when modifying for #610.

mderka:
Re: Dns.getProjectInfo(), consistency sounds like a good enough reason. Since this touches layers above this rpc and some code that I have been working on in the mean time, I will handle it within a separate PR. Opening #619.

mderka:
Ready for another check.

PR #620, opened 2016-02-04
Avoids exposing zone id

### LINKED ISSUE #610: Should we overload methods for manipulating managed zone for both id and name?

mderka:
Removes all the methods that identify zones by id. Also switches zone id attribute from BigInteger to String. Makes id read only attribute settable by package scope builder method only, and forces name to be set.

Adds options for zone create method.

Addressed all comments of #610. Fixes #610.

### DIFF
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
@@ -66,8 +66,7 @@ public class Zone implements Serializable {
   }
 
   /**
-   * Retrieves the latest information about the zone. The method retrieves the zone by name which
-   * must always be initialized.
+   * Retrieves the latest information about the zone. The method retrieves the zone by name.
    *
    * @param options optional restriction on what fields should be fetched
    * @return zone object containing updated information or {@code null} if not not found
@@ -100,7 +99,7 @@ public class Zone implements Serializable {
 
   /**
    * Submits {@link ChangeRequest} to the service for it to applied to this zone. The method
-   * searches for zone by name which must always be initialized.
+   * searches for zone by name.
    *
    * @param options optional restriction on what fields of {@link ChangeRequest} should be returned
    * @return ChangeRequest with server-assigned ID
@@ -114,8 +113,8 @@ public class Zone implements Serializable {
 
   /**
    * Retrieves an updated information about a change request previously submitted to be applied to
-   * this zone. Returns a {@link ChangeRequest} or {@code null} if the change request was not
-   * found. Throws {@link DnsException} if the zone is not found.
+   * this zone. Returns a {@link ChangeRequest} or {@code null} if the change request was not found.
+   * Throws {@link DnsException} if the zone is not found.
    *
    * @param options optional restriction on what fields of {@link ChangeRequest} should be returned
    * @return updated ChangeRequest
@@ -129,8 +128,8 @@ public class Zone implements Serializable {
   }
 
   /**
-   * Retrieves all change requests for this zone. The method searches for zone by name which must
-   * always be initialized. Returns a page of {@link ChangeRequest}s.
+   * Retrieves all change requests for this zone. The method searches for zone by name. Returns a
+   * page of {@link ChangeRequest}s.
    *
    * @param options optional restriction on listing and fields to be returned
    * @return a page of change requests


gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -68,166 +66,81 @@ public static Zone get(Dns dnsService, String zoneName, Dns.ZoneOption... option
...
-      zone = Zone.get(dns, zoneInfo.name(), options);
-    }
-    return zone;
+    return Zone.get(dns, zoneInfo.name(), options);
   }
 
   /**
-   * Deletes the zone. The method first attempts to delete the zone by ID. If the zone is not found
-   * or id is not set, it attempts to delete by name.
+   * Deletes the zone. The method first deletes the zone by name which must always be initialized.
```
aozarov: No need for "first". Also, name should always be available (checked by`ZoneInfo`).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -68,166 +66,81 @@ public static Zone get(Dns dnsService, String zoneName, Dns.ZoneOption... option
...
-    }
-    return deleted;
+    return dns.delete(zoneInfo.name());
   }
 
   /**
-   * Lists all {@link DnsRecord}s associated with this zone. First searches for zone by ID and if
-   * not found then by name.
+   * Lists all {@link DnsRecord}s associated with this zone. The method searches for zone by name
+   * which must always be initialized.
```
aozarov: Same here. I think the comment "which must always be initialized" is confusing.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -68,166 +66,81 @@ public static Zone get(Dns dnsService, String zoneName, Dns.ZoneOption... option
...
-    }
-    return updated;
+    return dns.applyChangeRequest(zoneInfo.name(), changeRequest, options);
   }
 
   /**
    * Retrieves an updated information about a change request previously submitted to be applied to
-   * this zone. First searches for the zone by ID and if not found then by name. Returns a {@link
-   * ChangeRequest} if found and {@code null} is the zone or the change request was not found.
+   * this zone. The method searches for zone by name which must always be initialized. Returns a
```
aozarov: I don't think we need to mention " The method searches for zone by name which must always be initialized."

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -68,166 +66,81 @@ public static Zone get(Dns dnsService, String zoneName, Dns.ZoneOption... option
...
+   * {@link ChangeRequest} if and {@code null} if the change request was not found. Throws {@link
```
aozarov: "if and"?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -68,166 +66,81 @@ public static Zone get(Dns dnsService, String zoneName, Dns.ZoneOption... option
...
+   * DnsException} if the zone is not found.
    *
    * @param options optional restriction on what fields of {@link ChangeRequest} should be returned
    * @return updated ChangeRequest
    * @throws DnsException upon failure or if the zone is not found
-   * @throws NullPointerException if both zone ID and name are not initialized
    * @throws NullPointerException if the change request does not have initialized id
```
aozarov: Did you mean "if {@code changeRequestId} is null"?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -68,166 +66,81 @@ public static Zone get(Dns dnsService, String zoneName, Dns.ZoneOption... option
...
-    return updated;
+    return dns.getChangeRequest(zoneInfo.name(), changeRequestId, options);
   }
 
   /**
-   * Retrieves all change requests for this zone. First searches for the zone by ID and if not found
-   * then by name. Returns a page of {@link ChangeRequest}s or {@code null} if the zone is not
-   * found.
+   * Retrieves all change requests for this zone. The method searches for zone by name which must
+   * always be initialized. Returns a page of {@link ChangeRequest}s. Throws a {@link DnsException}
```
aozarov: It is necessary to mention that DnsException is thrown if zone not found twice?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
```diff
@@ -255,7 +232,7 @@ public Builder toBuilder() {
     pb.setDescription(this.description());
     pb.setDnsName(this.dnsName());
     if (this.id() != null) {
-      pb.setId(this.id());
+      pb.setId(new BigInteger(this.id()));
```
aozarov: Fine as is. An alternative would be to keep it internally as BigInteger, however return it to the user as string (id.toString()).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
```diff
@@ -66,19 +66,10 @@
     private Builder() {
```
aozarov: I don't think we need it.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
```diff
@@ -137,311 +108,177 @@ public void testGetByName() {
...
-    expect(dns.getZone(ZONE_ID)).andReturn(null);
-    expect(dns.getZone(ZONE_ID, ZONE_FIELD_OPTIONS)).andReturn(null); // for options
+    expect(dns.listDnsRecords(ZONE_NAME, DNS_RECORD_OPTIONS)).andThrow(EXCEPTION);
+    expect(dns.listDnsRecords(ZONE_NAME, DNS_RECORD_OPTIONS)).andThrow(EXCEPTION);
     replay(dns);
-    Zone result = zoneNoName.reload();
-    assertNull(result);
-    zoneNoName.reload(ZONE_FIELD_OPTIONS); // for options
+    try {
+      zoneNoId.listDnsRecords();
```
aozarov: zoneNoId should not be needed once we validate that on Zone construction.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Right. Got rid of the initialization statement.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: removed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: removed

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: should be "or",

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Correct. Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
mderka: Probably not. Removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
mderka: I will leave it as is. I do not see any particular benefit of one solution over another.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
mderka: Correct if we assume that the pb object always has initialized name. This assumption should be safe to make with the exception of specifying giving the user the option to specify fields of zone to be returned. I rectified the issue by enforcing the field selectors of zone to always return zone name instead of zone id which it was set to up to now. Removing the builder.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
mderka: Removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Zone.java
```diff
@@ -68,166 +66,78 @@ public static Zone get(Dns dnsService, String zoneName, Dns.ZoneOption... option
...
-    return page;
+    return dns.listDnsRecords(zoneInfo.name(), options);
   }
 
   /**
-   * Submits {@link ChangeRequest} to the service for it to applied to this zone. First searches for
-   * the zone by ID and if not found then by name. Returns a {@link ChangeRequest} with
-   * server-assigned ID or {@code null} if the zone was not found.
+   * Submits {@link ChangeRequest} to the service for it to applied to this zone. The method
+   * searches for zone by name which must always be initialized.
```
aozarov: d/which must always be initialized/g

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
aozarov: As said. Fine as is. Just FYI, the slight advantage that I see with my suggestion is that because
the Builder.id method is package scope a code outside of this class may supply a string which
is could not be converted to BigInteger.

aozarov:
Looks good. @mderka feel free to merge after fixing the javadoc.

mderka:
Fixed. Merging.

PR #632, opened 2016-02-09
Fix NPE in bigquery XXXInfo.equals

mziccard:
This PR fixes a possible NPE in the equals method of `XXXInfo` objects.
This also:
- Updates javadoc of `TableDefinition` and subclasses to remove the `type` naming
- Fixes a style error in `Table`


ajkannan:
LGTM

PR #636, opened 2016-02-10
Docs, tests, and lint cleanup

### LINKED ISSUE #580: Move integration tests to different packages
While keeping unit tests in the same package as code is useful (so that we can test package scope code), this is less important for integration tests.  Moving integration tests to a separate package would provide some benefits, such as allowing us to catch issues related to member variable access.

ajkannan:
This PR makes the following changes:
- Fixes indentation issues in all modules in code blocks in javadoc
- Fixes lint errors
- Fixes #580 (moves integration tests to separate package)
- Avoids reading extra output during local gcd shutdown.

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
@@ -351,8 +351,10 @@ public final class Blob extends BlobInfo {
    * </p>
    *
    * <p>Example usage of replacing blob's metadata:
-   * <pre> {@code blob.toBuilder().metadata(null).build().update();}
-   * {@code blob.toBuilder().metadata(newMetadata).build().update();}
+   * <pre> {@code
+   * blob.toBuilder().metadata(null).build().update();
+   * blob.toBuilder().metadata(newMetadata).build().update();
+   * }
    * </pre>
    *
    * @param options update options
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
@@ -1300,8 +1300,10 @@ public interface Storage extends Service<StorageOptions> {
    * can be done by setting the provided {@code blobInfo}'s metadata to {@code null}.
    *
    * <p>Example usage of replacing blob's metadata:
-   * <pre> {@code service.update(BlobInfo.builder("bucket", "name").metadata(null).build());}
-   * {@code service.update(BlobInfo.builder("bucket", "name").metadata(newMetadata).build());}
+   * <pre> {@code
+   * service.update(BlobInfo.builder("bucket", "name").metadata(null).build());
+   * service.update(BlobInfo.builder("bucket", "name").metadata(newMetadata).build());
+   * }
    * </pre>
    *
    * @return the updated blob
@@ -1315,8 +1317,10 @@ public interface Storage extends Service<StorageOptions> {
    * can be done by setting the provided {@code blobInfo}'s metadata to {@code null}.
    *
    * <p>Example usage of replacing blob's metadata:
-   * <pre> {@code service.update(BlobInfo.builder("bucket", "name").metadata(null).build());}
-   * {@code service.update(BlobInfo.builder("bucket", "name").metadata(newMetadata).build());}
+   * <pre> {@code
+   * service.update(BlobInfo.builder("bucket", "name").metadata(null).build());
+   * service.update(BlobInfo.builder("bucket", "name").metadata(newMetadata).build());
+   * }
    * </pre>
    *
    * @return the updated blob


aozarov:
@mziccard That is great that you added codacy to gcloud-java and that we can see it in PR!!!

Any reason why the detail goes to https://www.codacy.com/app/mziccard/gcloud-java/pullRequest?prid=183606 (note the  mziccard part of the URL)?

Also, is there a way to add a link-able image with some scoring based on codacy in our main README.md (like the one we have for coveralls)?

mziccard:
@aozarov It goes to mziccard/... because I activated codacy for gcloud-java from my account. To have the URL point to GoogleCloudPlatform/... codacy should be activated by the owner/maintainer of the GoogleCloudPlatform account (I guess).

aozarov:
@mziccard I see. We can stay with that for now. How did you make it be shown in the PR list? I also just created one for gcloud-java just to check the URL issue. does it mean that any one that created an codeacy project that is associated with gcloud-java will appear in this list?

BTW, did you tune the checkstyle/pmd settings for our needs? Is there a way for you to share those?

mziccard:
@aozarov Actually, I have no idea. It seems a bit strange to me that all people that have access to the repository can create a different codacy project. 

To make the PR status be shown I enabled that in codacy's "Integrations" settings. So if you do not enable it we should just receive one status notification. If you enable status notifications I have no clue what will happen.

I haven't tuned checkstyle/pmd settings. This is why I asked yesterday if you were able to access those settings. I meant if you have rights to access settings for "my" gcloud-java's codacy project (as occurs in travis for instance where all of us change settings as we have write access to the github's repo).

aozarov:
Ah, no I don't have access to your settings or, more importantly, to "Code Patterns" Configuration.
However, I can see that they have an import/export option so maybe we can save the settings in gcloud-java. In that case feel free to do a first pass and then save it in gcloud-java and we can review it and modify if needed. Also, do you mind to add the badge?

gcloud-java-bigquery/src/test/java/com/google/gcloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -169,10 +197,9 @@ public static void afterClass() throws ExecutionException, InterruptedException
     if (bigquery != null) {
       RemoteBigQueryHelper.forceDelete(bigquery, DATASET);
     }
-    if (storage != null && !RemoteGcsHelper.forceDelete(storage, BUCKET, 10, TimeUnit.SECONDS)) {
-      if (LOG.isLoggable(Level.WARNING)) {
-        LOG.log(Level.WARNING, "Deletion of bucket {0} timed out, bucket is not empty", BUCKET);
-      }
+    if (storage != null && !RemoteGcsHelper.forceDelete(storage, BUCKET, 10, TimeUnit.SECONDS)
```
aozarov: I prefer:
```java
if (storage != null) {
  boolean wasDeleted = RemoteGcsHelper.forceDelete(storage, BUCKET, 10, TimeUnit.SECONDS);
  if (!wasDeleted && LOG.isLoggable(Level.WARNING)) {
    LOG.log(Level.WARNING, "Deletion of bucket {0} timed out, bucket is not empty", BUCKET);
  }  
}
```

ajkannan:
@aozarov made the style fix mentioned above.

aozarov:
@mziccard any reason the codacy reference disappeared from the list of checks?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -351,8 +351,8 @@ public Blob reload(BlobSourceOption... options) {
    * </p>
    *
    * <p>Example usage of replacing blob's metadata:
-   * <pre>    {@code blob.toBuilder().metadata(null).build().update();}
-   *    {@code blob.toBuilder().metadata(newMetadata).build().update();}
+   * <pre> {@code blob.toBuilder().metadata(null).build().update();}
```
aozarov: could this be part of one "{@code"?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1300,8 +1300,8 @@ private static void checkContentType(BlobInfo blobInfo) throws IllegalArgumentEx
    * can be done by setting the provided {@code blobInfo}'s metadata to {@code null}.
    *
    * <p>Example usage of replacing blob's metadata:
-   * <pre>    {@code service.update(BlobInfo.builder("bucket", "name").metadata(null).build());}
-   *    {@code service.update(BlobInfo.builder("bucket", "name").metadata(newMetadata).build());}
+   * <pre> {@code service.update(BlobInfo.builder("bucket", "name").metadata(null).build());}
```
aozarov: here and bellow. can we unify the "{@code" parts?

aozarov:
Looks good. One nit for unifying consecutive "{@code" sections.

ajkannan:
Unified code sections

mziccard:
@aozarov As you can see codacy checks did not disappear. They take some time to be run and hence the result takes some time to be displayed in the PR. I think the problem is that they are not triggered immediately after a commit. There's an option in "Settings" to set "the time to review a PR (hours) before receiving a notification", I set it to 0 which should make the review happen as fast as possible:

screen shot 2016-02-12 at 8 39 32 am

aozarov:
Ah, BTW did you enabled the Post-Commit Hook and the 2 options for GitHub integration in the codacy settings?

mziccard:
Yes. I enabled only the status option as comments are not necessary and messy

Issue #626, opened 2016-02-05
Adding checkstyle and other code analysis tools to our PR.


aozarov:
I think that it would be nice if we have more metrics on PR (including threshold that would flag the PR as red if exceeded). Checkstyle is one of these metrics.

I was looking at the checkstyle github [page](https://github.com/checkstyle/checkstyle), just to see
if they checkstyle their own code for PR and I think they do, but I didn't invest more time on it.

I also saw that they integrate with some other github friendly sites (and free to open source projects) such as:

https://www.versioneye.com/ - for dependency out-of-date notifications
https://www.codacy.com/ - for code analysis

mziccard:
I run codacy on our `gcloud-java`, you should be able to see the report [here](https://www.codacy.com/app/mziccard/gcloud-java/dashboard) so we can decide whether we find it useful.

ajkannan:
My first impression is that the UI is nice and some of the suggestions are helpful, but I do think there's a lot of irrelevant errors to sift through in order to get the useful tidbits.

aozarov:
I really like the UI and I think it would be great if we can get the PR summary (https://www.codacy.com/app/mziccard/gcloud-java/pullRequests) as part of the set of PR sets.
I have the feeling that some rules can be disabled (there is a notion of active/non-active) and customize (though I hope the latter does not mean a modified docker).
Also, some of the errors (such as "Enforce the use of explicit scoping"), can be resolved by 
code annotation (@VisibleForTesting in this [case](http://sourceforge.net/p/pmd/bugs/1410/)).

mziccard:
> I really like the UI and I think it would be great if we can get the PR summary (https://www.codacy.com/app/mziccard/gcloud-java/pullRequests) as part of the set of PR tests.

I should have just enabled this feature for PRs against the master branch. Should we enable the tool to run on all branches? (I think we should).

Also, out of curiosity, are you able to modify codacy settings?

aozarov:
> I should have just enabled this feature for PRs against the master branch. Should we enable the tool to run on all branches? (I think we should).

I also think we should.

> Also, out of curiosity, are you able to modify codacy settings?
Didn't try. but looks like it is doable - https://docs.codacy.com/v1.5/docs/code-patterns#2-configure-your-project-patterns no?

ajkannan:
Closing since both versioneye and codacy PRs are merged.

ajkannan:
[...]

PR #641, opened 2016-02-11
Add codacy badge and configuration

mziccard:
As per title this PR adds codacy badge to project's READMEs and also adda codacy configuration to the root of the project (named `codacy-conf.json`).

`codacy-conf.json` is not exactly human readable but can be imported and exported in codacy where you can edit it and export again (and possibly push to this repository).

I already did a pass disabling checks that we don't need and enabling the ones that I believe we should enforce. We could also consider disabling all checks related to CSS and Javascript.


aozarov:
BTW, did you understand how to include coverage data with codacy?
We don't really need it as Coveralls should be enough, but if trivial it would be nice.

aozarov:
This PR relates to #626

mziccard:
I did't look into it as we already have coveralls. Is it worth having two services checking for coverage? Or do you want to assess whether codacy's report is more accurate that Coveralls?

Also as you've seen by yourself codacy can be slow in providing the report to a PR, while coveralls shows the report almost in real time.

aozarov:
Right, it is not important. I was only curious as when you go to the codacy main page for the project you see it mentioned as "not configured". Also, I was curious to see if their report is more helpful in showing the missing coverage"...

mziccard:
I've been in touch with codacy's support today and they told me that the latency we are experiencing in seeing codacy's status after our PRs should be fixed soon. As soon as it becomes more responsive we can give it a try!

aozarov:
Sounds Good!

PR #644, opened 2016-02-12
Revert #625 and Fixes #624

### LINKED ISSUE #624: ITBigQueryTest failing due to missing query plan for query job statistics
The BigQuery service response for query jobs is missing the query plan field under job statistics.  `testQueryJob()` and `testQuery()` fail because they contain asserts that the query plan for queries is not null.

For now we will disable the tests that are failing.  This issue should be kept open until the tests are re-enabled.

aozarov:
Revert #625 and Fixes #624


mziccard:
Looks good to me

PR #645, opened 2016-02-12
Update dependencies versions and add versioneye badge

mziccard:
This follows from #626:
- Update dependencies versions whenever possible
- Add versioneye badge to READMEs

With respect to version changes I have two questions before we merge in:
- Is there a reason why we were using such an old version of `org.json.json` (20090211)? For instance GAE compatibility issues I am not aware of?
- `google-api-services-datastore-protobuf` at version v1beta2-rev1-4.0.0 lacks `EntityResult.getCursor()` method causing the error:

> [ERROR] symbol:   method getCursor()
> [ERROR] location: class com.google.api.services.datastore.DatastoreV1.EntityResult
> [ERROR] gcloud-java/gcloud-java-datastore/src/test/java/com/google/gcloud/datastore/DatastoreTest.java:[537,70] cannot find symbol

Should we update to the latest version or stick to the current one waiting for v1beta3?


aozarov:
> Is there a reason why we were using such an old version of org.json.json (20090211)? For instance GAE compatibility issues I am not aware of?

I don't think so, more of a copy/paste from an older pom.

> Should we update to the latest version or stick to the current one waiting for v1beta3?
#646 Feel free to close it and make it part of this PR if you prefer.

aozarov:
This is great!

BTW, what triggers your versioneye project to perform the check?

mziccard:
A separate PR is OK to update datastore! 

Versioneye checks are periodic. I configured it to check daily (finer granularity possible)

aozarov:
> Versioneye checks are periodic. I configured it to check daily (finer granularity possible)

Sounds good. So we should be all green if we run the check again after we merge #646 ?

PR #646, opened 2016-02-12
update google-api-services-datastore-protobuf dependency

aozarov:
upgrade google-api-services-datastore-protobuf to v1beta2-rev1-4.0.0 and fix test


mziccard:
looks good to me

PR #648, opened 2016-02-16
Datastore fixes

### LINKED ISSUE #639: Expose consistency setting in LocalGcdHelper
The default is 0.9, but it's often useful to set it to 1.0 for test utility methods (i.e. clearing all entities using a non-ancestor query in between unit tests).

This can be set in [LocalGcdHelper.startGcd(..)](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/testing/LocalGcdHelper.java#L488) using the following option:

--consistency=...
The fraction of job application attempts that will succeed, with 0.0 resulting in no attempts succeeding, and 1.0 resulting in all attempts succeeding. Defaults to 0.9. Note that setting this to 1.0 may mask incorrect assumptions about the consistency of non-ancestor queries; non-ancestor queries are eventually consistent.

### LINKED ISSUE #640: Provide methods to set list values of the same type
For example, we could expose a method:
`set(String value1, String value2, String others...)` to create a `ListValue` of String Values instead of having to write:
`entityBuilder.set("property", StringValue.of("val1"), StringValue.of("val2"), ...)`

### LINKED ISSUE #642: Provide getter for a key's parent
The `BaseKey` `parent()` method's return type will be `BaseKey`, and the `IncompleteKey` and `Key` classes can override this method to return a parent of type `Key`.

ajkannan:
This PR:
- Fixes #639 (exposes consistency in LocalGcdHelper)
- Fixes #640 (provides set methods for lists containing only one type)
- Fixes #642 (provide getter for a key's parent)
- Cleans up the getList return value

### DIFF
--- a/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseEntity.java
+++ b/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseEntity.java
@@ -370,7 +370,7 @@ public abstract class BaseEntity<K extends IncompleteKey> extends Serializable<D
      * @param others other values in the list
      */
     public B set(String name, Value<?> first, Value<?> second, Value<?>... others) {
-      properties.put(name, of(first, second, others));
+      properties.put(name, ListValue.builder().addValue(first).addValue(second, others).build());
       return self();
     }
 
--- a/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/ListValue.java
+++ b/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/ListValue.java
@@ -70,17 +70,16 @@ public final class ListValue extends Value<List<? extends Value<?>>> {
       super(ValueType.LIST);
     }
 
-    public Builder addValue(Value<?> value) {
+    private void addValueHelper(Value<?> value) {
       // see datastore_v1.proto definition for list_value
       Preconditions.checkArgument(value.type() != ValueType.LIST, "Cannot contain another list");
       listBuilder.add(value);
-      return this;
     }
 
     public Builder addValue(Value<?> first, Value<?>... other) {
-      addValue(first);
+      addValueHelper(first);
       for (Value<?> value : other) {
-        addValue(value);
+        addValueHelper(value);
       }
       return this;
     }
@@ -125,10 +124,6 @@ public final class ListValue extends Value<List<? extends Value<?>>> {
     this(new Builder().addValue(first, other));
   }
 
-  ListValue(Value<?> first, Value<?> second, Value<?>... other) {
-    this(new Builder().addValue(first).addValue(second, other));
-  }
-
   private ListValue(Builder builder) {
     super(builder);
   }
@@ -144,10 +139,6 @@ public final class ListValue extends Value<List<? extends Value<?>>> {
 
   public static ListValue of(Value<?> first, Value<?>... other) {
     return new ListValue(first, other);
-  }
-
-  static ListValue of(Value<?> first, Value<?> second, Value<?>... other) {
-    return new ListValue(first, second, other);
   }
 
   public static Builder builder() {


gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/IncompleteKey.java
```diff
@@ -84,6 +84,25 @@ static IncompleteKey fromPb(DatastoreV1.Key keyPb) {
     return new IncompleteKey(projectId, namespace, path);
   }
 
+  /**
+   * Returns the key's parent.
+   */
+  @Override
+  public Key parent() {
+    List<PathElement> ancestors = ancestors();
+    if (!ancestors.isEmpty()) {
```
aozarov: Nit (style), I would check
```java
if(ancestors.isEmpty()) {
   return null;`
}
```
and avoid the indent of the rest of the code.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/IncompleteKey.java
```diff
@@ -84,6 +84,25 @@ static IncompleteKey fromPb(DatastoreV1.Key keyPb) {
...
+      PathElement parent = ancestors.get(ancestors.size() - 1);
+      Key.Builder keyBuilder;
+      if (parent.hasName()) {
+        keyBuilder = Key.builder(projectId(), parent.kind(), parent.name());
+      } else {
+        keyBuilder = Key.builder(projectId(), parent.kind(), parent.id());
+      }
+      return keyBuilder.ancestors(ancestors.subList(0, ancestors.size() - 1)).build();
```
aozarov: what about namespace?

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/ListValue.java
```diff
@@ -138,8 +139,8 @@ public static ListValue of(List<? extends Value<?>> values) {
     return new ListValue(values);
   }
 
-  public static ListValue of(Value<?> first, Value<?>... other) {
-    return new ListValue(first, other);
+  public static ListValue of(Value<?> first, Value<?> second, Value<?>... other) {
```
aozarov: Was this change necessary? What about a single value list?

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseEntity.java
```diff
@@ -138,43 +139,120 @@ public B set(String name, String value) {
...
+    }
+
     public B set(String name, List<? extends Value<?>> values) {
       properties.put(name, of(values));
       return self();
     }
 
-    public B set(String name, Value<?> value, Value<?>... other) {
-      properties.put(name, of(value, other));
+    public B set(String name, Value<?> first, Value<?> second, Value<?>... other) {
```
aozarov: I don't think this change is necessary.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseEntity.java
```diff
@@ -183,6 +261,17 @@ public B set(String name, Blob value) {
       return self();
     }
 
+    public B set(String name, Blob first, Blob second, Blob... others) {
```
aozarov: Lets add javadoc (for all cases) to make it clear that this version is setting a list of values.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/testing/LocalGcdHelper.java
```diff
@@ -590,10 +600,13 @@ public static void main(String... args) throws IOException, InterruptedException
     String action = parsedArgs.get("action");
     int port =
         (parsedArgs.get("port") == null) ? DEFAULT_PORT : Integer.parseInt(parsedArgs.get("port"));
+    double consistency =
```
aozarov: this could be moved to be part of the "START" case.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseEntity.java
ajkannan: Because of [this overloaded set](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseEntity.java#L131) method, I believe it's impossible to set a list with one value at the moment.

ajkannan:
Incorporated the comments, ready for another pass.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/BaseEntity.java
aozarov: Right, luckily we have the set that accepts a list which will support the singleton case, no?

ajkannan:
Removed unnecessary changes to ListValue and made addValue() unambiguous.

PR #652, opened 2016-02-18
Create IAM Policy classes

ajkannan:


gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+import com.google.common.collect.ImmutableList;
+
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * An Identity and Access Management (IAM) policy. It is used to specify access control policies for
```
aozarov: Is it strange that we use the plural of the object as part of the description of that object?

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+ * Cloud Platform resources. A Policy consists of a list of ACLs (also known as bindings in Cloud
+ * IAM documentation). An ACL binds a list of identities to a role, where the identities can be user
+ * accounts, Google groups, Google domains, and service accounts. A role is a named list of
+ * permissions defined by IAM.
+ *
+ * @see <a href="https://cloud.google.com/iam/reference/rest/v1/Policy">Policy</a>
+ */
+public class IamPolicy implements Serializable {
+
+  static final long serialVersionUID = 1114489978726897720L;
```
aozarov: private

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(id, type);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (!(obj instanceof Identity)) {
```
aozarov: make Identity final

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  /**
+   * An ACL binds a list of identities to a role, where the identities can be user accounts, Google
+   * groups, Google domains, and service accounts. A role is a named list of permissions defined by
+   * IAM.
+   *
+   * @see <a href="https://cloud.google.com/iam/reference/rest/v1/Policy#Binding">Binding</a>
+   */
+  public static class Acl implements Serializable {
```
aozarov: As Acl does not have its own data did you consider using Policy like a Map of role -> List<Identity>?

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+     *
+     * <p>Etags are used for optimistic concurrency control as a way to help prevent simultaneous
+     * updates of a policy from overwriting each other. It is strongly suggested that systems make
+     * use of the etag in the read-modify-write cycle to perform policy updates in order to avoid
+     * race conditions. An etag is returned in the response to getIamPolicy, and systems are
+     * expected to put that etag in the request to setIamPolicy to ensure that their change will be
+     * applied to the same version of the policy.  If no etag is provided in the call to
+     * setIamPolicy, then the existing policy is overwritten blindly.
+     */
+    public Builder etag(String etag) {
```
aozarov: I think this the setter should be package scoped.

aozarov: Actually that would be an issue when this is in a shared/common place.

One option, just thinking out loud, is to make this class abstract and the builder protected.
That mean each service that supports IAM will provide its own subclass (with its own to/from pb)
and can even using generics for the role type (so each service can have its own roles enum if it has such). In this case etag can be protected.

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+      this.etag = etag;
+      return this;
+    }
+
+    /**
+     * Sets the version of the policy. The default version is 0.
+     */
+    public Builder version(int version) {
```
aozarov: is that something that comes from the service or the user specify? What is the meaning of the version (can you access/get by version)?

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(acls, etag, version);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof IamPolicy)) {
```
aozarov: This may change based on the proposal above

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,440 @@
...
+       */
+      GROUP,
+
+      /**
+       * Represents all the users of a Google Apps domain name.
+       */
+      DOMAIN
+    }
+
+    Identity(Type type, String id) {
```
aozarov: can this be private?

gcloud-java-core/src/test/java/com/google/gcloud/SerializationTest.java
```diff
@@ -0,0 +1,49 @@
...
+    for (Serializable obj : objects) {
+      Object copy = serializeAndDeserialize(obj);
+      assertEquals(obj, obj);
+      assertEquals(obj, copy);
+      assertNotSame(obj, copy);
+      assertEquals(copy, copy);
+    }
+  }
+
+  @SuppressWarnings("unchecked")
```
aozarov: I wonder if its about time to have a base class for Serialization tests that has a protected method to return the list of objects to test. If so, can be done in a separate PR.

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
ajkannan: The user can specify the version.  There's no get by version, nor do any docs suggest that it is an upcoming feature.  It seems to be a field purely for users to keep track of policy updates for themselves.  If the user doesn't specify the version when setting the policy, the service defaults the version to 0 in resource manager since setting a policy is a "replace" rather than an "update".

gcloud-java-core/src/test/java/com/google/gcloud/SerializationTest.java
ajkannan: I agree, created issue #657.

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
ajkannan: Edit: version is actually a means of determining whether alpha roles can be used.  Version 0 is default and means non-legacy roles (i.e. "pubsub.subscriber") cannot be used.  Version 1 means that alpha roles can be used.  New roles are only available via IAM alpha opt-in.  In light of this, I'll make version package scope.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Base class for Identity and Access Management (IAM) policies. IAM policies are used to specify
+ * access settings for Cloud Platform resources. A Policy consists of a list of bindings. An binding
```
aozarov: "A binding" ?

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+ * assigns a list of identities to a role, where the identities can be user accounts, Google groups,
+ * Google domains, and service accounts. A role is a named list of permissions defined by IAM.
+ *
+ * @see <a href="https://cloud.google.com/iam/reference/rest/v1/Policy">Policy</a>
+ */
+public abstract class BaseIamPolicy<R> implements Serializable {
```
aozarov: My naming preference would be to keep this one `IamPolicy` and call the specific subclasses `Policy`.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+
+  private static final long serialVersionUID = 1114489978726897720L;
+
+  private final Map<R, List<Identity>> bindings;
+  private final String etag;
+  private final int version;
+
+  public static final class Identity implements Serializable {
```
aozarov: Let's move it to its own file.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+    private static final long serialVersionUID = 30811617560110848L;
+
+    private final Type type;
+    private final String id;
+
+    /**
+     * The types of IAM identities.
+     */
+    public enum Type {
+      /**
```
aozarov: Nit, empty line before.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+     */
+    public static Identity allAuthenticatedUsers() {
+      return new Identity(Type.ALL_AUTHENTICATED_USERS, null);
+    }
+
+    /**
+     * Returns a new user identity.
+     *
+     * @param email An email address that represents a specific Google account. For example,
+     *     alice@gmail.com or joe@example.com.
```
aozarov: Nit, fine as is though it would be nice to make the examples using a different font style. e.g. `<I>`

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
aozarov: javadoc for param type

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+      this.bindings.clear();
+      this.bindings.putAll(bindings);
+      return self();
+    }
+
+    /**
+     * Adds one or more bindings to the policy.
+     */
+    public B addBinding(R role, List<Identity> identities) {
+      bindings.put(role, ImmutableList.copyOf(identities));
```
aozarov: what if role already exists?

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+      return self();
+    }
+
+    /**
+     * Removes the specified ACL.
+     */
+    public B removeBinding(R role) {
```
aozarov: should we have an option to remove an identity (or more) from a role binding?

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+    @SuppressWarnings("unchecked")
+    private B self() {
+      return (B) this;
+    }
+
+    public abstract BaseIamPolicy<R> build();
+  }
+
+  protected BaseIamPolicy(BaseBuilder<R, ? extends BaseBuilder<R, ?>> builder) {
+    this.bindings = builder.bindings;
```
aozarov: `ImmutableMap.copyOf`?

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  /**
+   * The version of the policy. The default version is 0.
+   */
+  public int version() {
+    return version;
+  }
+
+  public int baseHashCode() {
```
aozarov: protected

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+    return Objects.hash(bindings, etag, version);
+  }
+
+  public boolean baseEquals(Object obj) {
```
aozarov: protected

aozarov: Alternative to protected is to make rename them to hashCode and equals, make them public and final and change equals to check for class equality rather than instanceof.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+    if (!(obj instanceof BaseIamPolicy)) {
+      return false;
+    }
+    @SuppressWarnings("rawtypes")
+    BaseIamPolicy other = (BaseIamPolicy) obj;
+    return Objects.equals(bindings, other.bindings())
```
aozarov: I imagine that we would want to say that 2 policies are equal if the have the same role associated with the same identities but not in the same order, right?

gcloud-java-core/src/test/java/com/google/gcloud/BaseIamPolicyTest.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final Identity ALL_USERS = Identity.allUsers();
+  private static final Identity ALL_AUTH_USERS = Identity.allAuthenticatedUsers();
+  private static final Identity USER = Identity.user("abc@gmail.com");
+  private static final Identity SERVICE_ACCOUNT =
+      Identity.serviceAccount("service-account@gmail.com");
+  private static final Identity GROUP = Identity.group("group@gmail.com");
+  private static final Identity DOMAIN = Identity.domain("google.com");
+
+  @Test
+  public void testIdentityOf() {
```
aozarov: Move Identity testing to its own file (if we move Identity), this class also may need to rename
and add tests that check the Policy builder and Policy getters (and equals/hashCode or baseEquals/hashCode) using an anonymous subclass in the test.

gcloud-java-core/src/test/java/com/google/gcloud/BaseIamPolicyTest.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
aozarov: 2016

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Base class for Identity and Access Management (IAM) policies. IAM policies are used to specify
+ * access settings for Cloud Platform resources. A Policy consists of a list of bindings. An binding
+ * assigns a list of identities to a role, where the identities can be user accounts, Google groups,
+ * Google domains, and service accounts. A role is a named list of permissions defined by IAM.
+ *
+ * @see <a href="https://cloud.google.com/iam/reference/rest/v1/Policy">Policy</a>
+ */
+public class IamPolicy extends BaseIamPolicy<String> implements Serializable {
```
aozarov: The reason I prefer this one to be Policy is that all services may have a Policy where some of the are based on IamPolicy...

aozarov: redundant "implements Serializable"

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
```diff
@@ -0,0 +1,162 @@
...
+
+  private static final long serialVersionUID = -5573557282693961850L;
+
+  /**
+   * Builder for an IAM Policy.
+   */
+  protected static class Builder extends BaseBuilder<String, Builder> {
```
aozarov: public

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+      }
+      Identity other = (Identity) obj;
+      return Objects.equals(id, other.id()) && Objects.equals(type, other.type());
+    }
+  }
+
+  /**
+   * Builder for an IAM Policy.
+   */
+  protected abstract static class BaseBuilder<R, B extends BaseBuilder<R, B>> {
```
aozarov: Also to consider, making the class public and adding a protected constructor

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
```diff
@@ -0,0 +1,162 @@
...
+
+    Builder() {}
```
aozarov: private (here and below)?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
```diff
@@ -0,0 +1,162 @@
...
+
+    Builder(Map<String, List<Identity>> bindings, String etag, int version) {
+      bindings(bindings).etag(etag).version(version);
```
aozarov: let's make the parent Builder setters final.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
```diff
@@ -0,0 +1,162 @@
...
+        return "allUsers";
+      case ALL_AUTHENTICATED_USERS:
+        return "allAuthenticatedUsers";
+      case USER:
+        return "user:" + identity.id();
+      case SERVICE_ACCOUNT:
+        return "serviceAccount:" + identity.id();
+      case GROUP:
+        return "group:" + identity.id();
+      default:
```
aozarov: we should add a case for DOMAIN and make default fail if reached.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
```diff
@@ -0,0 +1,162 @@
...
+        return "domain:" + identity.id();
+    }
+  }
+
+  static Identity identityFromPb(String identityStr) {
```
aozarov: Also consider having this method in Identity. Maybe called fromEntity, valueOf, ...

I think using the enum `Type` would be nicer. Also, we should consider that a ':' may
be part of the value/id (is it documented otherwise)?

```java
int idx = identityStr.indexOf(':');
Type type = Type.valueOf(idx < 0 ? identityStr : identityStr.substring(0, idx));
switch (type) {
  case ALL_USERS:
    return Identity.allUsers();
  ....
  case USER:
    return Identity.user(identityStr.substring(idx + 1));
  ...
}
```

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+
+    private final Map<R, List<Identity>> bindings = new HashMap<>();
```
aozarov: see other related comments. Maybe this should be `Map<R, Set<Identity>>` and use `LinedHashSet` for maintaining the order.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+    private String etag;
+    private int version;
+
+    /**
+     * Replaces the builder's list of bindings with the given list of bindings.
+     */
+    public B bindings(Map<R, List<Identity>> bindings) {
+      this.bindings.clear();
+      this.bindings.putAll(bindings);
```
aozarov: This may have a mutable `Identity` list.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+      return self();
+    }
+
+    /**
+     * Adds one or more bindings to the policy.
```
ajkannan: role -> identity+ 

I'll go through the javadoc and standardize this.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
ajkannan: I made this constructor private, but I kept the one below package scope for testing (otherwise I can't construct a policy that has an etag and version)

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
```diff
@@ -0,0 +1,162 @@
...
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public Builder toBuilder() {
+    return new Builder(bindings(), etag(), version());
+  }
+
+  static String identityToPb(Identity identity) {
```
ajkannan: Any particular reason for calling it entity?  I made the change and named the method "strValue", because I thought that was slightly more fitting.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/IamPolicy.java
ajkannan: I believe that I can't use valueOf because the string values are camel case while the enum has all caps with underscores.

gcloud-java-core/src/main/java/com/google/gcloud/BaseIamPolicy.java
ajkannan: Fixed the classes to use sets.  I don't really see any particular reason for maintaining order, so I used a HashSet instead of a LinkedHashSet.

ajkannan:
Ready for another pass.

gcloud-java-core/src/main/java/com/google/gcloud/IamPolicy.java
```diff
@@ -0,0 +1,228 @@
...
+   */
+  public String etag() {
+    return etag;
+  }
+
+  /**
+   * Sets the version of the policy. The default version is 0, meaning only the "owner", "editor",
+   * and "viewer" roles are permitted. If the version is 1, you may also use other roles.
+   */
+  public Integer version() {
```
aozarov: For the users, this is read-only. Is there any meaning for `null` (and can it even happen)?

ajkannan: If a service specifies field options for a policy, this field can be null.  The handling of this return value depends on what we decide in #664.

gcloud-java-core/src/main/java/com/google/gcloud/Identity.java
```diff
@@ -0,0 +1,208 @@
...
+package com.google.gcloud;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * An identity in an {@link IamPolicy}.
+ */
```
aozarov: add a `@see` to [here](https://cloud.google.com/iam/docs/overview#concepts_related_to_identity)?

aozarov: Also, we can say a bit more about what is identity in the javadoc.

gcloud-java-core/src/main/java/com/google/gcloud/Identity.java
```diff
@@ -0,0 +1,208 @@
...
+      case USER:
+        return "user:" + id;
+      case SERVICE_ACCOUNT:
+        return "serviceAccount:" + id;
+      case GROUP:
+        return "group:" + id;
+      case DOMAIN:
+        return "domain:" + id;
+      default:
+        throw new IllegalArgumentException("Unexpected identity type: " + type);
```
aozarov: Nit, IllegalStateException

gcloud-java-core/src/test/java/com/google/gcloud/IamPolicyTest.java
```diff
@@ -0,0 +1,173 @@
...
+      public PolicyImpl build() {
+        return new PolicyImpl(this);
+      }
+    }
+
+    PolicyImpl(Builder builder) {
+      super(builder);
+    }
+
+    Builder toBuilder() {
```
aozarov: should we define it as an abstract method in `IamPolicy`?

gcloud-java-core/src/test/java/com/google/gcloud/IamPolicyTest.java
```diff
@@ -0,0 +1,173 @@
...
+    assertEquals(1, FULL_POLICY.version().intValue());
+    Map<String, Set<Identity>> editorBinding =
+        ImmutableMap.<String, Set<Identity>>builder().put("editor", BINDINGS.get("editor")).build();
+    PolicyImpl policy = FULL_POLICY.toBuilder().bindings(editorBinding).build();
+    assertEquals(editorBinding, policy.bindings());
+    assertEquals("etag", policy.etag());
+    assertEquals(1, policy.version().intValue());
+    policy = SIMPLE_POLICY.toBuilder().removeBinding("editor").build();
+    assertEquals(ImmutableMap.of("viewer", BINDINGS.get("viewer")), policy.bindings());
+    assertEquals(null, policy.etag());
```
aozarov: assertNull (here and below).

gcloud-java-core/src/test/java/com/google/gcloud/IamPolicyTest.java
```diff
@@ -0,0 +1,173 @@
...
+      SIMPLE_POLICY.toBuilder().addBinding("editor", ImmutableSet.of(USER));
+      fail("Should have failed due to duplicate role.");
+    } catch (IllegalArgumentException e) {
+      assertEquals("The policy already contains a binding with the role editor", e.getMessage());
+    }
+  }
+
+  @Test
+  public void testEqualsHashCode() {
+    assertNotEquals(FULL_POLICY, null);
```
aozarov: assertNotNull

ajkannan:
Ready for another pass.

PR #659, opened 2016-02-22
Add gcloud-java-compute module and immutable resources

mziccard:
As per title this PR adds the compute module.
This also adds classes for immutable resources (and their identities):
- DiskType
- MachineType
- License
- Region
- Zone

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
@@ -16,9 +16,11 @@
 
 package com.google.gcloud.compute;
 
-import com.google.api.client.util.DateTime;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
+
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import java.io.Serializable;
 import java.util.Objects;
@@ -31,6 +33,7 @@ import java.util.Objects;
 public final class DeprecationStatus<T extends ResourceId> implements Serializable {
 
   private static final long serialVersionUID = -2695077634793679794L;
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
 
   private final Long deleted;
   private final Long deprecated;
@@ -134,13 +137,13 @@ public final class DeprecationStatus<T extends ResourceId> implements Serializab
     com.google.api.services.compute.model.DeprecationStatus deprecationStatusPb =
         new com.google.api.services.compute.model.DeprecationStatus();
     if (deleted != null) {
-      deprecationStatusPb.setDeleted(new DateTime(deleted).toStringRfc3339());
+      deprecationStatusPb.setDeleted(TIMESTAMP_FORMATTER.print(deleted));
     }
     if (deprecated != null) {
-      deprecationStatusPb.setDeprecated(new DateTime(deprecated).toStringRfc3339());
+      deprecationStatusPb.setDeprecated(TIMESTAMP_FORMATTER.print(deprecated));
     }
     if (obsolete != null) {
-      deprecationStatusPb.setObsolete(new DateTime(obsolete).toStringRfc3339());
+      deprecationStatusPb.setObsolete(TIMESTAMP_FORMATTER.print(obsolete));
     }
     if (replacement != null) {
       deprecationStatusPb.setReplacement(replacement.selfLink());
@@ -156,11 +159,11 @@ public final class DeprecationStatus<T extends ResourceId> implements Serializab
       Function<String, T> fromUrl) {
     return new DeprecationStatus<>(
         deprecationStatusPb.getDeleted() != null
-            ? DateTime.parseRfc3339(deprecationStatusPb.getDeleted()).getValue() : null,
+            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getDeleted()) : null,
         deprecationStatusPb.getDeprecated() != null
-            ? DateTime.parseRfc3339(deprecationStatusPb.getDeprecated()).getValue() : null,
+            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getDeprecated()) : null,
         deprecationStatusPb.getObsolete() != null
-            ? DateTime.parseRfc3339(deprecationStatusPb.getObsolete()).getValue() : null,
+            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getObsolete()) : null,
         deprecationStatusPb.getReplacement() != null
             ? fromUrl.apply(deprecationStatusPb.getReplacement()) : null,
         deprecationStatusPb.getState() != null
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
@@ -16,9 +16,11 @@
 
 package com.google.gcloud.compute;
 
-import com.google.api.client.util.DateTime;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
+
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import java.io.Serializable;
 import java.math.BigInteger;
@@ -48,6 +50,7 @@ public final class DiskType implements Serializable {
       };
 
   private static final long serialVersionUID = -944042261695072026L;
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
 
   private final String id;
   private final DiskTypeId diskTypeId;
@@ -199,7 +202,7 @@ public final class DiskType implements Serializable {
       diskTypePb.setId(new BigInteger(id));
     }
     if (creationTimestamp != null) {
-      diskTypePb.setCreationTimestamp(new DateTime(creationTimestamp).toStringRfc3339());
+      diskTypePb.setCreationTimestamp(TIMESTAMP_FORMATTER.print(creationTimestamp));
     }
     diskTypePb.setDescription(description);
     diskTypePb.setValidDiskSize(validDiskSize);
@@ -222,8 +225,7 @@ public final class DiskType implements Serializable {
       builder.id(diskTypePb.getId().toString());
     }
     if (diskTypePb.getCreationTimestamp() != null) {
-      builder.creationTimestamp(
-          DateTime.parseRfc3339(diskTypePb.getCreationTimestamp()).getValue());
+      builder.creationTimestamp(TIMESTAMP_FORMATTER.parseMillis(diskTypePb.getCreationTimestamp()));
     }
     builder.diskTypeId(DiskTypeId.fromUrl(diskTypePb.getSelfLink()));
     builder.description(diskTypePb.getDescription());
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
@@ -16,11 +16,13 @@
 
 package com.google.gcloud.compute;
 
-import com.google.api.client.util.DateTime;
 import com.google.api.services.compute.model.MachineType.ScratchDisks;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.Lists;
+
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import java.io.Serializable;
 import java.math.BigInteger;
@@ -52,6 +54,7 @@ public final class MachineType implements Serializable {
           return type.toPb();
         }
       };
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
 
   private static final long serialVersionUID = -4210962597502860450L;
 
@@ -255,7 +258,7 @@ public final class MachineType implements Serializable {
       machineTypePb.setId(new BigInteger(id));
     }
     if (creationTimestamp != null) {
-      machineTypePb.setCreationTimestamp(new DateTime(creationTimestamp).toStringRfc3339());
+      machineTypePb.setCreationTimestamp(TIMESTAMP_FORMATTER.print(creationTimestamp));
     }
     machineTypePb.setName(machineTypeId.machineType());
     machineTypePb.setDescription(description);
@@ -292,7 +295,7 @@ public final class MachineType implements Serializable {
     }
     if (machineTypePb.getCreationTimestamp() != null) {
       builder.creationTimestamp(
-          DateTime.parseRfc3339(machineTypePb.getCreationTimestamp()).getValue());
+          TIMESTAMP_FORMATTER.parseMillis(machineTypePb.getCreationTimestamp()));
     }
     builder.description(machineTypePb.getDescription());
     builder.cpus(machineTypePb.getGuestCpus());
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
@@ -16,11 +16,13 @@
 
 package com.google.gcloud.compute;
 
-import com.google.api.client.util.DateTime;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import java.io.Serializable;
 import java.math.BigInteger;
@@ -50,6 +52,7 @@ public final class Region implements Serializable {
       };
 
   private static final long serialVersionUID = -3578710133393645135L;
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
 
   private final RegionId regionId;
   private final String id;
@@ -319,7 +322,7 @@ public final class Region implements Serializable {
       regionPb.setId(new BigInteger(id));
     }
     if (creationTimestamp != null) {
-      regionPb.setCreationTimestamp(new DateTime(creationTimestamp).toStringRfc3339());
+      regionPb.setCreationTimestamp(TIMESTAMP_FORMATTER.print(creationTimestamp));
     }
     regionPb.setName(regionId.region());
     regionPb.setDescription(description);
@@ -350,7 +353,7 @@ public final class Region implements Serializable {
       builder.id(regionPb.getId().toString());
     }
     if (regionPb.getCreationTimestamp() != null) {
-      builder.creationTimestamp(DateTime.parseRfc3339(regionPb.getCreationTimestamp()).getValue());
+      builder.creationTimestamp(TIMESTAMP_FORMATTER.parseMillis(regionPb.getCreationTimestamp()));
     }
     builder.description(regionPb.getDescription());
     if (regionPb.getStatus() != null) {
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
@@ -16,11 +16,13 @@
 
 package com.google.gcloud.compute;
 
-import com.google.api.client.util.DateTime;
 import com.google.api.services.compute.model.Zone.MaintenanceWindows;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.Lists;
+
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import java.io.Serializable;
 import java.math.BigInteger;
@@ -50,6 +52,7 @@ public final class Zone implements Serializable {
       };
 
   private static final long serialVersionUID = 6113636504417213010L;
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
 
   private final ZoneId zoneId;
   private final String id;
@@ -162,16 +165,16 @@ public final class Zone implements Serializable {
       return new MaintenanceWindows()
           .setName(name)
           .setDescription(description)
-          .setBeginTime(beginTime != null ? new DateTime(beginTime).toStringRfc3339() : null)
-          .setEndTime(endTime != null ? new DateTime(endTime).toStringRfc3339() : null);
+          .setBeginTime(beginTime != null ? TIMESTAMP_FORMATTER.print(beginTime) : null)
+          .setEndTime(endTime != null ? TIMESTAMP_FORMATTER.print(endTime) : null);
     }
 
     static MaintenanceWindow fromPb(MaintenanceWindows windowPb) {
       return new MaintenanceWindow(windowPb.getName(), windowPb.getDescription(),
           windowPb.getBeginTime() != null
-              ? DateTime.parseRfc3339(windowPb.getBeginTime()).getValue() : null,
+              ? TIMESTAMP_FORMATTER.parseMillis(windowPb.getBeginTime()) : null,
           windowPb.getEndTime() != null
-              ? DateTime.parseRfc3339(windowPb.getEndTime()).getValue() : null);
+              ? TIMESTAMP_FORMATTER.parseMillis(windowPb.getEndTime()) : null);
     }
   }
 
@@ -338,7 +341,7 @@ public final class Zone implements Serializable {
       zonePb.setId(new BigInteger(id));
     }
     if (creationTimestamp != null) {
-      zonePb.setCreationTimestamp(new DateTime(creationTimestamp).toStringRfc3339());
+      zonePb.setCreationTimestamp(TIMESTAMP_FORMATTER.print(creationTimestamp));
     }
     zonePb.setName(zoneId.zone());
     zonePb.setDescription(description);
@@ -370,7 +373,7 @@ public final class Zone implements Serializable {
       builder.id(zonePb.getId().toString());
     }
     if (zonePb.getCreationTimestamp() != null) {
-      builder.creationTimestamp(DateTime.parseRfc3339(zonePb.getCreationTimestamp()).getValue());
+      builder.creationTimestamp(TIMESTAMP_FORMATTER.parseMillis(zonePb.getCreationTimestamp()));
     }
     builder.description(zonePb.getDescription());
     if (zonePb.getStatus() != null) {


gcloud-java-compute/README.md
```diff
@@ -0,0 +1,100 @@
...
+Example Application
+-------------------
+<!-- TODO(mziccard): add example application -->
+
+Authentication
+--------------
+
+See the [Authentication](https://github.com/GoogleCloudPlatform/gcloud-java#authentication) section in the base directory's README.
+
+About Google Cloud BigQuery
```
ajkannan: s/BigQuery/Compute/

gcloud-java-compute/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>gcloud-java-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-compute</artifactId>
+      <version>v1-rev93-1.21.0</version>
```
ajkannan: From the public docs, it looks like there's a v1 and a beta version out.  Just so that I'm on the same page, why did we choose to go with v1 over beta?

Also it looks like a new revision was released a couple days ago: [v1-rev97-1.21.0](http://search.maven.org/#artifactdetails%7Ccom.google.apis%7Cgoogle-api-services-compute%7Cv1-rev97-1.21.0%7Cjar)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeException.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.io.IOException;
+import java.util.Set;
+
+/**
+ * Compute Engine service exception.
+ */
+public class ComputeException extends BaseServiceException {
+
+  private static final Set<Error> RETRYABLE_ERRORS = ImmutableSet.of(new Error(500, null));
```
ajkannan: Is there a webpage that details which errors should be retried?  If so, it'd be good to include as a comment here.

gcloud-java-compute/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+Java Versions
+-------------
+
+Java 7 or above is required for using this client.
+
+Testing
+-------
+
+This library has tools to help make tests for code using Cloud BigQuery.
```
ajkannan: s/BigQuery/Compute/

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeException.java
mziccard: I couldn't find it. Still trying to figure out which are the possible error codes and which of them are retryable.

gcloud-java-compute/pom.xml
mziccard: We discussed with the compute team and decided to stick to v1 for now.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,227 @@
...
+import com.google.common.base.Function;
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.math.BigInteger;
+import java.util.Objects;
+
+/**
+ * Google Compute Engine Disk type. A disk type represents the type of disk to use, such as a
+ * {@code pd-ssd} or {@code pd-standard}.
```
ajkannan: Could add link to https://cloud.google.com/compute/docs/reference/latest/diskTypes

gcloud-java-compute/src/main/java/com/google/gcloud/compute/License.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * A Google Compute Engine License. A License represents a software license. Licenses are used to
+ * track software usage in images, persistent disks, snapshots, and virtual machine instances.
+ */
```
ajkannan: Could add link to https://cloud.google.com/compute/docs/reference/latest/licenses (and same for other classes that correspond to model objects)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,227 @@
...
+      new Function<DiskType, com.google.api.services.compute.model.DiskType>() {
+        @Override
+        public com.google.api.services.compute.model.DiskType apply(DiskType diskType) {
+          return diskType.toPb();
+        }
+      };
+
+  private static final long serialVersionUID = -944042261695072026L;
+
+  private final Long id;
```
ajkannan: Do we care about the deprecation status? https://cloud.google.com/compute/docs/reference/latest/diskTypes#deprecated

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,227 @@
...
+    private DiskTypeId diskTypeId;
+    private String creationTimestamp;
+    private String description;
+    private String validDiskSize;
+    private String selfLink;
+    private Long defaultDiskSizeGb;
+
+    private Builder() {}
+
+    Builder id(Long id) {
```
ajkannan: Could this be private (since we have the static package scope "builder" method below)?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,227 @@
...
+    return diskTypePb;
+  }
+
+  static Builder builder() {
+    return new Builder();
+  }
+
+  static DiskType fromPb(com.google.api.services.compute.model.DiskType diskTypePb) {
+    Builder builder = builder();
+    if (diskTypePb.getId() != null ) {
```
ajkannan: Nit: extraneous space before ')'

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,227 @@
...
+      this.selfLink = selfLink;
+      return this;
+    }
+
+    Builder defaultDiskSizeGb(Long defaultDiskSizeGb) {
+      this.defaultDiskSizeGb = defaultDiskSizeGb;
+      return this;
+    }
+
+    /**
```
ajkannan: Nit: since this builder is package scope, this javadoc is unnecessary.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,227 @@
...
+    this.selfLink = builder.selfLink;
+    this.defaultDiskSizeGb = builder.defaultDiskSizeGb;
+  }
+
+  /**
+   * Returns the creation timestamp in RFC3339 text format.
+   *
+   * @see <a href="https://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
+   */
+  public String creationTimestamp() {
```
ajkannan: Do you think it'd be more helpful to users if we converted this into something like epoch milliseconds?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskTypeId.java
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Identity for a Google Compute Engine disk type.
+ */
+public final class DiskTypeId extends ZoneResourceId {
+
+  private static final long serialVersionUID = 7337881474103686219L;
+
+  private final String diskType;
+
+  DiskTypeId(String project, String zone, String diskType) {
```
ajkannan: private?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskTypeId.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof DiskTypeId && baseEquals((DiskTypeId) obj);
+  }
+
+  @Override
+  DiskTypeId setProjectId(String projectId) {
+    if (project() != null) {
+      return this;
```
ajkannan: I assume this means the DiskTypeId's project ID can't be changed.  Should this throw an exception instead of returning itself?

mziccard: As we did in bigquery we allow the user to provide an identity for a project-owned resource without specifying the project name (e.g. `DiskTypeId.of("some-type")`). Before sending such a resource to the service we must take care of setting the project id to the default project (the one in options). The contract of `setProjectId()` is to return a new identity where the project is set (so that it can be sent to the service), so I think it should not throw.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: I don't have a preference. I kept the string format as that is used everywhere across the compute service but I don't mind changing.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskTypeId.java
ajkannan: Ah, makes sense (I missed that this is a package scope method)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
```diff
@@ -0,0 +1,298 @@
...
+    private String description;
+    private String selfLink;
+    private Integer guestCpus;
+    private Integer memoryMb;
+    private List<Integer> scratchDisks;
+    private Integer maximumPersistentDisks;
+    private Long maximumPersistentDisksSizeGb;
+
+    Builder() {}
+
```
ajkannan: Nit: extra line

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
```diff
@@ -0,0 +1,298 @@
...
+        @Override
+        public com.google.api.services.compute.model.MachineType apply(MachineType type) {
+          return type.toPb();
+        }
+      };
+
+  private static final long serialVersionUID = -4210962597502860450L;
+
+  private final MachineTypeId machineTypeId;
+  private final Long id;
```
ajkannan: Do you think we should use BigInteger for IDs instead of Long (here and elsewhere)?  The protos seem to use BigInteger, and Long could lead to overflow issues.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/LicenseId.java
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public static LicenseId of(String project, String license) {
+    return new LicenseId(project, license);
+  }
+
+  static LicenseId fromUrl(String url) {
+    int projectsIndex = url.indexOf("/projects/");
+    int licensesIndex = url.indexOf("/global/licenses/");
+    String project = url.substring(projectsIndex + 10, licensesIndex);
+    String license = url.substring(licensesIndex + 17, url.length());
```
ajkannan: Is this the same thing as the [name](https://cloud.google.com/compute/docs/reference/latest/machineTypes#name) field of the model object, or is "name" something different?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
```diff
@@ -0,0 +1,319 @@
...
+          .setLimit(limit)
+          .setUsage(usage);
+    }
+
+    static Quota fromPb(com.google.api.services.compute.model.Quota quotaPb) {
+      return new Quota(quotaPb.getMetric(), quotaPb.getLimit(), quotaPb.getUsage());
+    }
+  }
+
+  private final RegionId regionId;
```
ajkannan: Nit: here and elsewhere, can we move these variables to the beginning of the class?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/License.java
```diff
@@ -0,0 +1,98 @@
...
+    return MoreObjects.toStringHelper(this)
+        .add("licenseId", licenseId)
+        .add("chargesUseFee", chargesUseFee)
+        .add("selfLink", selfLink)
+        .toString();
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(licenseId, chargesUseFee, selfLink);
```
ajkannan: To be consistent with the other classes' (eg DiskType, MachineType) hashcode methods, this could just be licenseId, right?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
```diff
@@ -0,0 +1,298 @@
...
+  private final String creationTimestamp;
+  private final String description;
+  private final String selfLink;
+  private final Integer guestCpus;
+  private final Integer memoryMb;
+  private final List<Integer> scratchDisks;
+  private final Integer maximumPersistentDisks;
+  private final Long maximumPersistentDisksSizeGb;
+
+  public static final class Builder {
```
ajkannan: This can be a package scope builder, right?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
```diff
@@ -0,0 +1,319 @@
...
+    private RegionId regionId;
+    private Long id;
+    private String creationTimestamp;
+    private String description;
+    private String selfLink;
+    private Status status;
+    private List<ZoneId> zones;
+    private List<Quota> quotas;
+
+    Builder() {}
```
ajkannan: Here (and in other builders), I think we can make this constructor private.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
```diff
@@ -0,0 +1,319 @@
...
+  private final RegionId regionId;
+  private final Long id;
+  private final String creationTimestamp;
+  private final String description;
+  private final String selfLink;
+  private final Status status;
+  private final List<ZoneId> zones;
+  private final List<Quota> quotas;
+
+  public static final class Builder {
```
ajkannan: package scope instead of public?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  Region(Builder builder) {
+    this.regionId = builder.regionId;
+    this.id = builder.id;
+    this.creationTimestamp = builder.creationTimestamp;
+    this.description = builder.description;
+    this.selfLink = builder.selfLink;
+    this.status = builder.status;
+    this.zones = builder.zones;
```
ajkannan: Should this be an immutable copy of the list?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
```diff
@@ -0,0 +1,319 @@
...
+      return this;
+    }
+
+    Builder status(Status status) {
+      this.status = status;
+      return this;
+    }
+
+    Builder zones(List<ZoneId> zones) {
+      this.zones = zones;
```
ajkannan: Here and in quotas, should we make a copy of the list?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
```diff
@@ -0,0 +1,380 @@
...
+  private final ZoneId zoneId;
+  private final Long id;
+  private final String creationTimestamp;
+  private final String description;
+  private final String selfLink;
+  private final Status status;
+  private final List<MaintenanceWindow> maintenanceWindows;
+  private final RegionId region;
+
+  public static final class Builder {
```
ajkannan: package scope

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
```diff
@@ -0,0 +1,380 @@
...
+      return Objects.hash(name, description, beginTime, endTime);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (!(obj instanceof MaintenanceWindow)) {
+        return false;
+      }
+      MaintenanceWindow other = (MaintenanceWindow) obj;
+      return Objects.equals(name, other.name)
```
ajkannan: Nit: could we use toPb() here as well?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
```diff
@@ -0,0 +1,380 @@
...
+  }
+
+  static Builder builder() {
+    return new Builder();
+  }
+
+  static Zone fromPb(com.google.api.services.compute.model.Zone zonePb) {
+    Builder builder = builder();
+    builder.zoneId(ZoneId.fromUrl(zonePb.getSelfLink()));
+    if (zonePb.getId() != null) {
```
ajkannan: If/when you implement a "get" with options to specify fields, will ID a field that is always returned, regardless of whether the user specifies it or not?  If it will always be returned, I think `zonePb.getId()` should never be null, and we could remove this check.  (Also applies to other `fromPb`s where we check for null ID.)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
mziccard: The fields are documented to be long values in the API docs. We used this conversion also elsewhere. I would avoid BigInteger if possible

gcloud-java-compute/src/main/java/com/google/gcloud/compute/LicenseId.java
mziccard: It's the same

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
mziccard: Yes I plan to have fields selector that make Id null, as we did also elsewhere

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
ajkannan: Because it says "unsigned" in the [docs](https://cloud.google.com/compute/docs/reference/latest/machineTypes#id), I wonder if there's a chance for overflow?  It seems slightly weird that the api client library would use BigInteger if it could have used a long instead.  In the dns-alpha branch, I think there's a similar situation where I believe we chose to go with BigInteger ([docs](https://cloud.google.com/dns/api/v1/projects#number) and our [code](https://github.com/GoogleCloudPlatform/gcloud-java/blob/dns-alpha/gcloud-java-dns/src/main/java/com/google/gcloud/dns/ProjectInfo.java#L38))

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskTypeTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class DiskTypeTest {
+
+  private static final Long ID = 42L;
+  private static final String CREATION_TIMESTAMP = "2016-01-20T04:39:00.210-08:00";
+  private static final String DESCRIPTION = "description";
+  private static final String VALID_DISK_SIZE = "10GB-10TB";
+  private static final Long DEFAULT_DISK_SIZE_GB = 10L;
+  private static final DiskTypeId DISK_TYPE_ID = DiskTypeId.of("project", "zone", "diskType");
+  private static final DiskType DISK_TYPE = DiskType.builder()
+    .id(ID)
```
ajkannan: Nit: should be indented 4 spaces

gcloud-java-compute/src/test/java/com/google/gcloud/compute/SerializationTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.List;
+
+import org.junit.Test;
```
ajkannan: The linter complains that this should go above the "java...." imports (b/c it's a third party import).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
ajkannan: Do you have a preference @aozarov @mderka?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskTypeIdTest.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Test
+  public void testToAndFromUrl() {
+    DiskTypeId diskTypeId = DiskTypeId.of(PROJECT, ZONE, DISK_TYPE);
+    compareDiskTypeId(diskTypeId, DiskTypeId.fromUrl(diskTypeId.toUrl()));
+  }
+
+  @Test
+  public void testSetProjectId() {
```
ajkannan: Nit: could increase coverage by adding a line `assertSame(diskTypeId, diskTypeId.setProjectId("something")) `

gcloud-java-compute/src/test/java/com/google/gcloud/compute/MachineTypeTest.java
```diff
@@ -0,0 +1,90 @@
...
+    MachineType machineType = MachineType.builder()
+        .id(ID)
+        .machineTypeId(MACHINE_TYPE_ID)
+        .selfLink(MACHINE_TYPE_ID.toUrl())
+        .build();
+    compareMachineTypes(machineType, MachineType.fromPb(machineType.toPb()));
+  }
+
+  private void compareMachineTypes(MachineType expected, MachineType value) {
+    assertEquals(expected.id(), value.id());
```
ajkannan: Duplicate of below.  I think you meant to remove the "id()" here to check the "equals" method.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskTypeTest.java
```diff
@@ -0,0 +1,73 @@
...
+    DiskType diskType = DiskType.builder()
+        .id(ID)
+        .diskTypeId(DISK_TYPE_ID)
+        .selfLink(DISK_TYPE_ID.toUrl())
+        .build();
+    compareDiskTypes(diskType, DiskType.fromPb(diskType.toPb()));
+  }
+
+  private void compareDiskTypes(DiskType expected, DiskType value) {
+    assertEquals(expected.id(), value.id());
```
ajkannan: Nit: we usually add an assertEquals(expected, value) in the compare function.

ajkannan:
Finished first pass; looks good, mostly minor comments.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
mziccard: You are totally right. I don't know why I was convinced it was just `long`. Thanks for spotting this!

mziccard:
@ajkannan Thanks for the thorough pass!

I addressed all comments in 6d1f2b9 (add deprecation status and address all visibility/docs issues) and deef55c (make ID a BigInteger and put fields at the top of classes).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeException.java
ajkannan: Do you mind adding a todo comment here or creating an issue for updating this set once we have that info?

ajkannan:
It looks like Region and Zone also have deprecation status as well.  Do you think we should add a deprecation status to those two classes as well?

gcloud-java-compute/README.md
```diff
@@ -0,0 +1,100 @@
...
+Java idiomatic client for [Google Cloud Compute] (https://cloud.google.com/compute).
+
+[![Build Status](https://travis-ci.org/GoogleCloudPlatform/gcloud-java.svg?branch=master)](https://travis-ci.org/GoogleCloudPlatform/gcloud-java)
+[![Coverage Status](https://coveralls.io/repos/GoogleCloudPlatform/gcloud-java/badge.svg?branch=master)](https://coveralls.io/r/GoogleCloudPlatform/gcloud-java?branch=master)
+<!-- TODO(mziccard): add in the maven shield once the artifact is pushed to maven -->
+[![Codacy Badge](https://api.codacy.com/project/badge/grade/9da006ad7c3a4fe1abd142e77c003917)](https://www.codacy.com/app/mziccard/gcloud-java)
+[![Dependency Status](https://www.versioneye.com/user/projects/56bd8ee72a29ed002d2b0969/badge.svg?style=flat)](https://www.versioneye.com/user/projects/56bd8ee72a29ed002d2b0969)
+
+-  [Homepage] (https://googlecloudplatform.github.io/gcloud-java/)
+<!-- TODO(mziccard): add link to API documentatin -->
```
ajkannan: Nit: this comment shows up in the README.  Also documentation is missing an "o".

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DeprecationStatusTest.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.gcloud.compute.DeprecationStatus.Status;
+
+import org.junit.Test;
+
+public class DeprecationStatusTest {
+
+  private static final String DELETED = "2016-01-20T04:39:00.210-08:00";
+  private static final String DEPRECATED = "2016-01-20T04:37:00.210-08:00";
+  private static final String OBSOLETE = "2016-01-20T04:38:00.210-08:00";
+  private static final DiskTypeId DISK_TYPE_ID = DiskTypeId.of("project", "zone","diskType");
```
ajkannan: Lint error: missing space after comma here and in line 32.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskTypeId.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Returns a disk type identity given the zone identity and the disk type name.
+   */
+  public static DiskTypeId of(ZoneId zoneId, String diskType) {
+    return new DiskTypeId(zoneId.project(), zoneId.zone(), diskType);
+  }
+
+  /**
+   * Returns a disk type identity given the zone disk and disk type.
```
ajkannan: Here and below, "zone disk" --> "zone" or "zone (as a string)" (unless "zone disk" a term used in compute that I'm unfamiliar with).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/LicenseId.java
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Returns the name of the license. The name must be 1-63 characters long, and comply with
+   * RFC1035. Specifically, the name must be 1-63 characters long and match the regular expression
+   * {@code [a-z]([-a-z0-9]*[a-z0-9])?} which means the first character must be a lowercase letter,
+   * and all following characters must be a dash, lowercase letter, or digit, except the last
+   * character, which cannot be a dash.
+   *
+   * @see <a href="https://www.ietf.org/rfc/rfc1035.txt">RFC1035</a>
+   */
+  public String license() {
```
ajkannan: If every Resource ID subclass has a field similar to this one, do you think it'd be useful to add a field "name" and name getter to ResourceId?  I don't feel strongly either way, I'm just suggesting an alternative in case somebody has a preference.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
```diff
@@ -0,0 +1,326 @@
...
+
+  /**
+   * Returns the machine type's identity.
+   */
+  public MachineTypeId machineTypeId() {
+    return machineTypeId;
+  }
+
+  /**
+   * Returns an unique identifier for the machin type; defined by the service.
```
ajkannan: s/machin/machine/

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RegionId.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(baseHashCode(), region);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof RegionId && baseEquals((RegionId) obj);
```
ajkannan: Here and in other ID classes' equals methods, shouldn't there also be a check that the name (here it's `region`) is equal?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskTypeTest.java
```diff
@@ -0,0 +1,85 @@
...
+    assertEquals(CREATION_TIMESTAMP, DISK_TYPE.creationTimestamp());
+    assertEquals(DESCRIPTION, DISK_TYPE.description());
+    assertEquals(VALID_DISK_SIZE, DISK_TYPE.validDiskSize());
+    assertEquals(DISK_TYPE_ID.toUrl(), DISK_TYPE.selfLink());
+    assertEquals(DEFAULT_DISK_SIZE_GB, DISK_TYPE.defaultDiskSizeGb());
+    assertEquals(DEPRECATION_STATUS, DISK_TYPE.deprecationStatus());
+  }
+
+  @Test
+  public void testToPbAndFromPb() {
```
ajkannan: Up to you, but we could test that a Disk Type (or any other model object) is properly handled by to/fromPb when all optional fields are left unspecified (it would mostly catch potential NPEs).

ajkannan:
I'm finished with my review of this PR.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RegionId.java
mziccard: `ResourceId`'s `baseEquals` compares `toUrl` that is redefined in subclasses so comparison should be already taking name into account.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/LicenseId.java
mziccard: This is similar to what we have in bigquery's identities. I don't mind changing either.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeException.java
mziccard: Opened #662

mziccard:
@ajkannan Comments are fixed in ea5f070 and 0940f31.

In  0465926 I add a static method to identity classes to help parse url into identity objects. It is needed also elsewhere (not in this PR) to distinguish between global and regional resources

gcloud-java-compute/pom.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0"?>
```
aozarov: This file is using tabs for indentation

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RegionResourceId.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(baseHashCode(), region);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj != null
+        && obj.getClass().equals(RegionResourceId.class)
```
aozarov: How could this ever be true?

aozarov: Same for `ZoneResourceId`

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskTypeId.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+  };
+  static final Function<DiskTypeId, String> TO_URL_FUNCTION = new Function<DiskTypeId, String>() {
+    @Override
+    public String apply(DiskTypeId diskTypeId) {
+      return diskTypeId.toUrl();
+    }
+  };
+
+  static final String REGEX = ZoneResourceId.REGEX + "diskTypes/[^/]+";
```
aozarov: seems like this can be private (unless you were planning to use it in tests later?).
Also, you can compile it to a Pattern to make the matching slightly faster.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
```diff
@@ -0,0 +1,76 @@
...
+  private String project;
+
+  ResourceId(String project) {
+    this.project = project;
+  }
+
+  /**
+   * Base URL for a resource URL.
+   */
+  private static final String BASE_URL = "https://www.googleapis.com/compute/v1/projects/";
```
aozarov: move up

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Base class for Google Compute Engine resource identities.
+ */
+public abstract class ResourceId implements Serializable {
+
+  static final String REGEX =
+      "(https?://(www|content).googleapis.com/compute/v1/)?projects/[^/]+/";
+  private static final long serialVersionUID = -8028734746870421573L;
+
+  private String project;
```
aozarov: final ?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -0,0 +1,164 @@
...
+    if (status() != null) {
+      deprecationStatusPb.setState(status.name());
+    }
+    return deprecationStatusPb;
+  }
+
+  static <T extends ResourceId> DeprecationStatus<T> fromPb(
+      com.google.api.services.compute.model.DeprecationStatus deprecationStatusPb,
+      Function<String, T> fromUrl) {
+    return new DeprecationStatus<T>(
```
aozarov: `<>`?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskTypeIdTest.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Test
+  public void testToAndFromUrl() {
+    DiskTypeId diskTypeId = DiskTypeId.of(PROJECT, ZONE, DISK_TYPE);
+    assertSame(diskTypeId, diskTypeId.setProjectId(PROJECT));
+    compareDiskTypeId(diskTypeId, DiskTypeId.fromUrl(diskTypeId.toUrl()));
+    thrown.expect(IllegalArgumentException.class);
+    thrown.expectMessage("notMatchingUrl is not a valid disk type URL");
+    diskTypeId = DiskTypeId.fromUrl("notMatchingUrl");
```
aozarov: Nit, redundant assignment.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
```diff
@@ -0,0 +1,402 @@
...
+     */
+    MaintenanceWindow(String name, String description, String beginTime, String endTime) {
+      this.name = name;
+      this.description = description;
+      this.beginTime = beginTime;
+      this.endTime = endTime;
+    }
+
+    /**
+     * Returns the disk of the maintanance window.
```
aozarov: maintenance

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Objects;
+
+/**
+ * The deprecation status associated to a Google Compute Engine resource.
+ */
+public final class DeprecationStatus<T extends ResourceId> implements Serializable {
+
+  private static final long serialVersionUID = -2695077634793679794L;
+
+  private final String deleted;
```
aozarov: 3 values are actually timestamps. Should really present them as strings?

mziccard: Timestamps are returned as RFC strings across all the service.  If we decide for one format we need to stick to it. 

I don't have a preference, I just thought to remain close to the service representation as (milli)seconds since epoch (what we use in Storage if I recall correctly) doesn't sound much more friendly than RF  string to me. I don't mind changing so if you guys have any preference I can do it. 

Also please @aozarov let me know whenever you are done reviewing so I a can address the changes alltogether

aozarov: I think it is much simpler for most people to convert from millisecond epoch rather than from a string (of which you need to know the format, and much easier to get wrong). However, consider it as a nit, as I agree that neither is great and will be changed when we move to Java 8.
BTW, I would argue that consistency with the rest of *our* API is more important in that case.

Sorry for the intermediate review. Will let you when I am done (hopefully by EOD PST).

mziccard: > BTW, I would argue that consistency with the rest of our API is more important in that case.

Agree! 

> Sorry for the intermediate review. Will let you know when I am done (hopefully by EOD PST).

It's fine with intermediate reviews, I just wanted to be sure to avoid pushing intermediate commits

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Base class for Google Compute Engine resource identities.
+ */
+public abstract class ResourceId implements Serializable {
+
+  static final String REGEX =
+      "(https?://(www|content).googleapis.com/compute/v1/)?projects/[^/]+/";
```
aozarov: shouldn't the base url and the regex be effected by the host in the service options?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -0,0 +1,164 @@
...
+
+import com.google.common.base.Function;
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * The deprecation status associated to a Google Compute Engine resource.
+ */
```
aozarov: `@param <T> ...`

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ZoneResourceId.java
```diff
@@ -0,0 +1,80 @@
...
+    return super.toStringHelper().add("zone", zone);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(baseHashCode(), zone);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
aozarov: My preference would be:

1) not to have equals/hashCode for this abstract class but rather override baseEquals/baseHashCode to consider the zone (and call super).
2) Change ResourceId baseEquals to consider just its member variable (projectId) and not the overriden toUrl
3) Change all leaf subclass (which already check instanceof their type) to also check the equality of their specific value.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,247 @@
...
+      new Function<DiskType, com.google.api.services.compute.model.DiskType>() {
+        @Override
+        public com.google.api.services.compute.model.DiskType apply(DiskType diskType) {
+          return diskType.toPb();
+        }
+      };
+
+  private static final long serialVersionUID = -944042261695072026L;
+
+  private final BigInteger id;
```
aozarov: Here in in other places with "id", do you think there is a value in returning it as a numeric value rather than a String?
I dns (e.g. `ZoneInfo`) we decided to go with String as it feels nicer to something as opaque as id.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
aozarov: +1 for long.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
mziccard: In some APIs the resource url is returned as `https://www.googleapis.com/...` (usually `selfLink` has this format) while for other resource's fields (e.g. attached instances in disk resources) the url is returned as `https:// content.googleapis.com/...`.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: I can use String, this will require adding a `null` check but it's ok! BTW I don't expect users to be dealing with this id very much.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
aozarov: Also, looks like input from the user can be partial URL. is it guarantied that input from the service will always be full URL?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,247 @@
...
+  private final DiskTypeId diskTypeId;
+  private final String creationTimestamp;
+  private final String description;
+  private final String validDiskSize;
+  private final String selfLink;
```
aozarov: Is that going to be difference than diskTypeId.toUrl ?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
mziccard: Yes to parse the identity I use `selfLink` is always guaranteed to be a full URL

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: Actually it's going to be the same url. We can remove it to avoid redundancy. Do you think we should keep the getter?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
aozarov: Can we just ignore the URL version part? A user can specify a host (e.g. when used against a beta version of the service or en emulator when/if that comes) and I am not sure if that may effect this URLs....

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
aozarov: +1 for String. Yes, I agree I also not expect users to use it and rather them treat it as a black box than something with meaning (e.g. numeric value).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
aozarov: Did you consider renaming ResourceId.toUrl to selfLink or link? In that case I would feel more comfrotable in not having this getter.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
mziccard: If you are concerned about this (and I think it's a reasonable concern) we could avoid the base url part as a whole. This will make the matching less strict but still sound enough.

aozarov: Yes, I would vote for that.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: I liked `toUrl` a little better for naming consistency with `toPb`/`fromPb`.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Returns a textual description of the disk type.
+   */
+  public String description() {
+    return description;
+  }
+
+  /**
+   * Returns an optional textual description of the valid disk size, such as "10GB-10TB".
```
aozarov: This is nasty. Why can't they have 2 fields for min and max? Is it something that we can parse?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
aozarov: I like toUrl too (though I prefer toUri), but toPb/fromPb are package scope and this is public and I assume for a reason?

Because, this url is really the selfLink and because we used selfLink in other places
to describe the "Resource URI" I would vote to change it (if we want to eliminate the
redundancy) but I would not mind a  global search and replace of selfLink to toUri.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,247 @@
...
+  com.google.api.services.compute.model.DiskType toPb() {
+    com.google.api.services.compute.model.DiskType diskTypePb =
+        new com.google.api.services.compute.model.DiskType();
+    diskTypePb.setId(id);
+    diskTypePb.setCreationTimestamp(creationTimestamp);
+    diskTypePb.setDescription(description);
+    diskTypePb.setValidDiskSize(validDiskSize);
+    diskTypePb.setSelfLink(selfLink);
+    diskTypePb.setDefaultDiskSizeGb(defaultDiskSizeGb);
+    diskTypePb.setZone(diskTypeId.zoneId().toUrl());
```
aozarov: So that redundancy (selfLink and zone fields) is also part of the underlying API...
Did you check with the service owners if that was intentional and if that may ever change?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: Yes we can parse it even though there's no real documentation of which units if measure can actually be returned (mb, gb, tb). Do you think it's safe enough for us to parse it and assume that possible units are only the tree above?

aozarov: I think it is worth checking that with the service owners.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: Actually I did not. We could be building the identity out of zone+name, but selfLink has all the info we needed so I decided to use that one. This allows for more filtering. Also the `selfLink` field is always present even in the other modules.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskTypeId.java
```diff
@@ -0,0 +1,136 @@
...
+  private final String diskType;
+
+  private DiskTypeId(String project, String zone, String diskType) {
+    super(project, zone);
+    this.diskType = checkNotNull(diskType);
+  }
+
+  /**
+   * Returns the name of the disk type resource. The name must be 1-63 characters long, and comply
+   * with RFC1035. Specifically, the name must be 1-63 characters long and match the regular
```
aozarov: Do you think the constraint settings would be more useful in the setter, `of` (or in addition, or reference)?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskTypeId.java
```diff
@@ -0,0 +1,136 @@
...
+   */
+  static boolean matchesUrl(String url) {
+    return url.matches(REGEX);
+  }
+
+  static DiskTypeId fromUrl(String url) {
+    if (!matchesUrl(url)) {
+      throw new IllegalArgumentException(url + " is not a valid disk type URL");
+    }
+    int projectsIndex = url.indexOf("/projects/");
```
aozarov: Fine as is, though I feel a regexp would have been more readable (and definitely more efficient).

mziccard: Yes I plan to follow your suggestion and use a Pattern

gcloud-java-compute/src/main/java/com/google/gcloud/compute/LicenseId.java
```diff
@@ -0,0 +1,127 @@
...
+  private final String license;
+
+  private LicenseId(String project, String license) {
+    super(project);
+    this.license = checkNotNull(license);
+  }
+
+  /**
+   * Returns the name of the license. The name must be 1-63 characters long, and comply with
+   * RFC1035. Specifically, the name must be 1-63 characters long and match the regular expression
```
aozarov: same comment as before regarding the javadoc location of the constraint.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
```diff
@@ -0,0 +1,326 @@
...
+
+  /**
+   * Returns a service-defined URL for the machine type.
+   */
+  public String selfLink() {
+    return selfLink;
+  }
+
+  /**
+   * Returns the number of virtual CPUs that are available to the instance.
```
aozarov: I wonder why they qualify this as "guest" but don't do the same for, e.g., memory.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/MachineType.java
```diff
@@ -0,0 +1,326 @@
...
+   * Returns the amount of physical memory available to the instance, defined in MB.
+   */
+  public Integer memoryMb() {
+    return memoryMb;
+  }
+
+  /**
+   * Returns the size of all extended scratch disks assigned to the instance, defined in GB.
+   */
+  public List<Integer> scratchDisks() {
```
aozarov: should we rename it to scratchDiskGb (to be consistent with maximumPersistentDisksSizeGb) ?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Region.java
```diff
@@ -0,0 +1,371 @@
...
+    /**
+     * Returns a region quota object.
+     */
+    Quota(String metric, double limit, double usage) {
+      this.metric = metric;
+      this.limit = limit;
+      this.usage = usage;
+    }
+
+    public String metric() {
```
aozarov: javadoc here and below.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Zone.java
aozarov: s/disk/name/

aozarov:
That is indeed a big API... Nice work! Done with the review. Didn't pay much attention to the tests but I assume @ajkannan did (and also looks good in coveralls).

gcloud-java-compute/pom.xml
mziccard: Are you sure about this? It seems to me it's using spaces (also I just copy pasted it from an existing pom of ours).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: I opened #683 and I'll take care of checking with the service owners.

mziccard:
I fixed all comments in 9d48528. I opened #683 and I suggest we keep the current String type until we close it.

gcloud-java-compute/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+      <artifactId>gcloud-java-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-compute</artifactId>
+      <version>v1-rev97-1.21.0</version>
+      <scope>compile</scope>
+      <exclusions>
+	<exclusion>
```
aozarov:
```
\t<exclusion>
\t  <groupId>com.google.guava</groupId>
\t  <artifactId>guava-jdk5</artifactId>
\t</exclusion>
```

gcloud-java-compute/pom.xml
aozarov: Yes, see bellow.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
```diff
@@ -0,0 +1,238 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gcloud.compute;
+
+import com.google.api.client.util.DateTime;
```
aozarov: Nit, I think we should avoid depending `com.google.api.client.util` as much as possible.
The only one exception is `Data` (which is part of the API).
In this case we should use Joda (until Java8).
I see that we already use it in other [places](https://github.com/GoogleCloudPlatform/gcloud-java/search?utf8=%E2%9C%93&q=%22com.google.api.client.util%22&type=Code), so this should be done in a separate cleanup PR (even more so for Lists which should be replaced with guava).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Base class for Google Compute Engine resource identities.
+ */
+public abstract class ResourceId implements Serializable {
+
+  static final String REGEX = ".*?projects/([^/]+)/";
+  private static final String BASE_URL = "https://www.googleapis.com/compute/v1/projects/";
```
aozarov: Fine as is for now but I think we can get that from the service options (without the "project/" suffix).
Maybe we should create an issue for it?

aozarov:
Looks good. I think there are 2 more issues that needs to be handled but can be handled in other PRs.
@mziccard feel free to merge if you also think that they should be done after or not at all.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ResourceId.java
mziccard: I opened #693

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskType.java
mziccard: I removed DateTime in favor of Joda's formatter in 9873ef6

PR #663, opened 2016-02-24
Udpate dependencies: source plugin, bigquery, storage and fluido skin

mziccard:
This PR updates dependencies to the latest version.
We should also update `maven-site-plugin` to version 3.5 but the skin we are using (`maven-fluido-skin`) is not compatible with this latest version yet. I checked and [no maven skin](http://maven.apache.org/skins/) works with 3.5 at the moment.


ajkannan:
LGTM

PR #668, opened 2016-02-25
Add pageToken to list fields option for storage and resourcemanager

### LINKED ISSUE #665: List operations do not return nextPageToken which prevents proper paging
When implementing integration tests for DNS, I noticed that if one restricts the fields to be retrieved for a list operation, `iterateAll()` does not work properly. The following test would always fail:

```java
dnsRecordIterator = DNS.listDnsRecords(zone.name(),
    Dns.DnsRecordListOption.fields(Dns.DnsRecordField.TYPE),
    Dns.DnsRecordListOption.pageSize(1)).iterateAll(); // to force the iterator fetch next page
  int counter = 0;
  while (dnsRecordIterator.hasNext()) {
    counter++;
  }
assertEquals(2, counter);
```

This is because the field selector silently prevents the service from returning the `nextPageToken`:

`builder.append("rrsets(").append(DnsRecordField.selector(fields)).append(')');`

instead of

`builder.append("nextPageToken,rrsets(").append(DnsRecordField.selector(fields)).append(')');`

I recall seeing the same handling in resource manage. Having discussed this with @ajkannan, this is likely a bug. The same bug may be present in other parts of the library when a combination of listing and field restriction is possible.

mziccard:
This fixes #665 for `gcloud-java-storage` and `gcloud-java-resourcemanager`. bigquery already takes care of adding the `nextPageToken` fields to the selector.

@mderka please forgive me for stealing this from you but I needed a distraction from compute :)

### DIFF
--- a/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
+++ b/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
@@ -400,17 +400,15 @@ public class LocalResourceManagerHelper {
     String[] listFields = (String[]) options.get("listFields");
     StringBuilder responseBody = new StringBuilder();
     responseBody.append('{');
-    boolean commaNeeded = false;
     // If fields parameter is set but no project field is selected we must return no projects.
     if (!(projectFields != null && projectFields.length == 0)) {
       responseBody.append("\"projects\": [");
       Joiner.on(",").appendTo(responseBody, projectsSerialized);
       responseBody.append(']');
-      commaNeeded = true;
     }
     if (nextPageToken != null && (listFields == null
         || ImmutableSet.copyOf(listFields).contains("nextPageToken"))) {
-      if (commaNeeded) {
+      if (responseBody.length() > 1) {
         responseBody.append(',');
       }
       responseBody.append("\"nextPageToken\": \"");


gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -236,10 +236,18 @@ private static String projectIdFromUri(String path) throws IOException {
...
           case "fields":
-            // List fields are in the form "projects(field1, field2, ...)"
-            options.put(
-                "fields",
-                argEntry[1].substring("projects(".length(), argEntry[1].length() - 1).split(","));
+            // List fields are in the form "projects(field1, field2, ...),nextPageToken"
+            String option = argEntry[1];
+            int projectStart = option.indexOf("projects(");
+            int projectEnd = option.indexOf(')');
+            if (projectStart != -1 && projectEnd != -1
```
ajkannan: I think checking for a -1 index is unnecessary here because we use "projects(" and ")" whenever we set the list fields option, right?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -236,10 +236,18 @@ private static String projectIdFromUri(String path) throws IOException {
         String[] argEntry = arg.split("=");
         switch (argEntry[0]) {
           case "fields":
-            // List fields are in the form "projects(field1, field2, ...)"
-            options.put(
-                "fields",
-                argEntry[1].substring("projects(".length(), argEntry[1].length() - 1).split(","));
+            // List fields are in the form "projects(field1, field2, ...),nextPageToken"
+            String option = argEntry[1];
+            int projectStart = option.indexOf("projects(");
```
aozarov: Fine as is, though maybe at this point you want to consider a pattern:
```java
  private final static Pattern PROJECTS_PATTERN =
      Pattern.compile(".*\\bprojects\\((.*)\\).*");

  Matcher m = PROJECTS_PATTERN.matcher(argEntry[1]);
  String option = m.matches() ? m.group(1) : argEntry[1];
  options.put("listFields", option.split(",");
```

ajkannan:
LGTM pending resolving those two comments

mderka:
LGTM too.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
mziccard: Yes, the idea was to make it more generic in case we ever decide to add some strange field selector that does not specify project fields

ajkannan:
@mziccard feel free to merge (I would do it, but I'm unsure on whether you want to use Arie's suggestion)

mziccard:
@ajkannan @aozarov I fixed both comments in  5fea7ce if you want to have a look!

ajkannan:
After discussing with @aozarov, we actually do want to handle the case where no project fields are specified (i.e., the options are "" or "nextPageToken").  While we never get this case from our own code, he suggests keeping the mock services general when possible for any valid requests, not just requests made via gcloud-java libraries.

(Sorry about pointing it out and then suggesting changing it back).

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -380,19 +390,30 @@ Response list(Map<String, Object> options) {
...
+    if (!(projectFields != null && projectFields.length == 0)) {
+      responseBody.append("\"projects\": [");
+      Joiner.on(",").appendTo(responseBody, projectsSerialized);
+      responseBody.append(']');
+      commaNeeded = true;
+    }
+    if (nextPageToken != null && (listFields == null
+        || ImmutableSet.copyOf(listFields).contains("nextPageToken"))) {
+      if (commaNeeded) {
+        responseBody.append(",");
```
ajkannan: Nit: codacy suggests using single quotes instead of double quotes here.

ajkannan:
Looks great!  One codacy bug that I think is worth fixing, and then feel free to merge.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -380,19 +390,30 @@ Response list(Map<String, Object> options) {
...
+    // If fields parameter is set but no project field is selected we must return no projects.
+    if (!(projectFields != null && projectFields.length == 0)) {
+      responseBody.append("\"projects\": [");
+      Joiner.on(",").appendTo(responseBody, projectsSerialized);
+      responseBody.append(']');
+      commaNeeded = true;
+    }
+    if (nextPageToken != null && (listFields == null
+        || ImmutableSet.copyOf(listFields).contains("nextPageToken"))) {
+      if (commaNeeded) {
```
aozarov: Nit and fine as is but I would have used `responseBody.length() > 1` instead of the extra `commaNeeded` variable.

aozarov:
LGTM

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
mziccard: You are right, fixed in a439628

PR #670, opened 2016-02-25
Fixing doc problems, improving formatting

### LINKED ISSUE #385: In ServiceApi classes, fix wording of documentation about returning errors vs throwing exceptions.
Currently, the documentation as copied from the proto includes wording like this:

* "Returns NOT_FOUND if the topic does not exist.
* "If the subscription already exists, returns ALREADY_EXISTS."
* "If the corresponding topic doesn't exist, returns NOT_FOUND."
* "The server may return UNAVAILABLE if there are too many concurrent pull requests pending for the given subscription."

These returns are referring to the service, not the api wrapper, which is confusing. The api wrapper actually throws exceptions in these circumstances. The wording of the documentation on the proto needs to be made more idiom-independent. Additionally, if the markup in the proto could indicate the error cases in a structured way, it would make it a lot easier to generate `@throws` documentation for the errors. That may be a longer-stretch task, though. Before that point, at least a general `@throws` clause needs to be added. Additionally, the documentation needs to make clear how to get the exact error that occurred (specifically, getting the status code).

### LINKED ISSUE #386: In ServiceApi classes, make clear how to cancel requests
Specifically, the documentation of SubscriberApi.pull says "The client may cancel the request if it does not wish to wait any longer for the response." This is currently only possible if the api caller uses the ApiCallable, and calls cancel on the ClientCall. The other method forms still make reference to canceling though, even though it's not possible. Possibly the interface needs to be updated somehow, or also the documentation needs to be adjusted.

### LINKED ISSUE #387: SubscriptionApi.createSubscription: avoid reference to deleted topics
The documentation for the "topic" parameter of the flattened form of createSubscription says "The value of this field will be `_deleted-topic_` if the topic has been deleted." This is not relevant to the createSubscription operation and should be avoided somehow.

garrettjonesgoogle:
fixes #387 - Moving references of delete topics to get/list calls
fixes #386 - Removing the problematic wording

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/gcloud/pubsub/spi/SubscriberApi.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/gcloud/pubsub/spi/SubscriberApi.java
@@ -675,7 +675,7 @@ public class SubscriberApi implements AutoCloseable {
    * @param subscription The name of the subscription.
    * @param ackIds List of acknowledgment IDs.
    * @param ackDeadlineSeconds The new ack deadline with respect to the time this request was sent to
-   * the Pub/Sub system. Must be >= 0. For example, if the value is 10, the new
+   * the Pub/Sub system. Must be &gt;= 0. For example, if the value is 10, the new
    * ack deadline will expire 10 seconds after the `ModifyAckDeadline` call
    * was made. Specifying zero may immediately make the message available for
    * another pull request.


aozarov:
Changes are fine, but I don't think this PR completely resolve #385.

I think "generate NOT_FOUND", though maybe better than "return NOT_FOUND", is still not clear.
Also, what is NOT_FOUND (which looks like a constant) is not clear and this part from the issue
description "Additionally, the documentation needs to make clear how to get the exact error that occurred (specifically, getting the status code)." is not dealt with.

garrettjonesgoogle:
Ok, I removed the "fixes 385" line from the PR description, and we'll take it up in a separate PR.

PR #688, opened 2016-02-29
Add a versions option to BlobListOption

### LINKED ISSUE #687: Add a flag to enable/disable versioned Blob listing.
see "versions" in [here](https://cloud.google.com/storage/docs/json_api/v1/objects/list).

Also, it looks like the SPI already [sets](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java#L158) this option if provided.

Related StackOverflow [issue](http://stackoverflow.com/questions/35690777/get-archived-versions-from-google-cloud-storage).

aozarov:
Fixes #687.


aozarov:
I added an integration test for listing with versioning enabled.

By doing so I realized 2 things:
1) blob delete using bucket name and blob name does not work when versioning enabled as it seems
to require a generation id (does not infer latest). Same applies to blob get without generation number (return not found).
2) listing a bucket with generation enabled returns nothing if versioning is not enabled in the listing operation.

Is that expected? @Capstan @BrandonY

gcloud-java-storage/src/test/java/com/google/gcloud/storage/it/ITStorageTest.java
```diff
@@ -28,23 +28,14 @@
...
-import com.google.gcloud.storage.BatchResponse;
-import com.google.gcloud.storage.Blob;
-import com.google.gcloud.storage.BlobId;
-import com.google.gcloud.storage.BlobInfo;
-import com.google.gcloud.storage.Bucket;
-import com.google.gcloud.storage.BucketInfo;
-import com.google.gcloud.storage.CopyWriter;
-import com.google.gcloud.storage.HttpMethod;
-import com.google.gcloud.storage.Storage;
+import com.google.gcloud.storage.*;
```
mziccard: We should not use on-demand imports right?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -701,6 +701,16 @@ public static BlobListOption recursive(boolean recursive) {
     }
 
     /**
+     * If set to {@code true}, lists all versions of a blob.
+     * The default is {@code false}.
```
mziccard: Can this fit on the above line?

gcloud-java-storage/src/test/java/com/google/gcloud/storage/it/ITStorageTest.java
aozarov: Right, auto import "organize" with wrong formatting rules on my laptop.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
aozarov: Yes.

BrandonY:
Deleting a blob without specifying a generation should always (and fully consistently) delete the LIVE version of said object, if any. If there is only an archived version, the delete will fail, as it cannot find a live version.

Listing the objects in a bucket with "versions" set to false ignores the existence of archived objects. If there are no live objects, it will return no results.

aozarov:
Thanks @BrandonY your description makes sense.

For the record, what confused me was that deleting a Blob (with also specifying its generation id) in a versioned bucket didn't actually delete it but rather removed its "live" status and therefore for completely deleting it (which is required if we want to delete its bucket) we need to call delete on it again.

aozarov:
Correction. Actually a delete with generation id is not only removing the "live" status but will completely
remove the blob. In our case, the delete happen after a listing with selective fields which didn't include
a generation id and therefore only function as a "live" status change.

PR #694, opened 2016-03-01
Replace com.google.api.client.util.Lists with Guava's Lists

mziccard:


ajkannan:
Do you think it's worth adding to this CL related changes, i.e. changing com.google.api.client.util.DateTime [in BucketInfo](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-storage/src/main/java/com/google/gcloud/storage/BucketInfo.java#L25) to joda?

Two additional cleanup changes could include:

* BucketInfo uses the wrong checkNotNull static import
* BlobInfo uses api client DateTime

mziccard:
`BucketInfo` and `BlobInfo` use `DateTime` because the proto objects timestamps are actually `DateTime` fields.

mziccard:
Fixed checkNotNull inclusion in 6e87fa9

ajkannan:
Looks like there are merge conflicts, but other than that, looks good to me.

Issue #456, opened 2015-12-15
Add a common interface for resource field enums


mziccard:
We should consider creating an interface for all field enums (e.g. `DatasetField` in gcloud-java-bigquery and`BlobField` in gcloud-java-storage).
```java
interface ResourceField {
  String selector();
}
```
Selector's creation code (at least part of it) should be moved to a common static method.

aozarov:
For background see [this](https://github.com/GoogleCloudPlatform/gcloud-java/pull/454#discussion_r47676271).

mziccard:
Fixed with #807

aozarov:
@mziccard I just wonder if we should keep this one open until you also apply the fix to compute?

mziccard:
Keeping this open until it's fixed in compute

mziccard:
This was fixed for compute in #970

PR #696, opened 2016-03-01
Add service methods for immutable Compute resources and tests

mziccard:
This PR adds spi and service methods for manipulating the following resources:
- DiskType
- MachineType
- License
- Zone
- Region

In c18730c integration tests are also added.

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
@@ -16,12 +16,17 @@
 
 package com.google.gcloud.compute;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.base.Joiner;
+import com.google.common.base.MoreObjects;
 import com.google.common.collect.Sets;
 import com.google.gcloud.Page;
 import com.google.gcloud.Service;
 import com.google.gcloud.spi.ComputeRpc;
 
+import java.io.Serializable;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -45,7 +50,8 @@ public interface Compute extends Service<ComputeOptions> {
     NAME("name"),
     SELF_LINK("selfLink"),
     VALID_DISK_SIZE("validDiskSize"),
-    ZONE("zone");
+    ZONE("zone"),
+    DEPRECATED("deprecated");
 
     private final String selector;
 
@@ -76,7 +82,7 @@ public interface Compute extends Service<ComputeOptions> {
   enum MachineTypeField {
     CREATION_TIMESTAMP("creationTimestamp"),
     DESCRIPTION("description"),
-    GUEST_CPUS("cpus"),
+    GUEST_CPUS("guestCpus"),
     ID("id"),
     IMAGE_SPACE_GB("imageSpaceGb"),
     MAXIMUM_PERSISTENT_DISKS("maximumPersistentDisks"),
@@ -84,9 +90,9 @@ public interface Compute extends Service<ComputeOptions> {
     MEMORY_MB("memoryMb"),
     NAME("name"),
     SCRATCH_DISKS("scratchDisks"),
-    DISK_GB("diskGb"),
     SELF_LINK("selfLink"),
-    ZONE("zone");
+    ZONE("zone"),
+    DEPRECATED("deprecated");
 
     private final String selector;
 
@@ -122,7 +128,8 @@ public interface Compute extends Service<ComputeOptions> {
     QUOTAS("quotas"),
     SELF_LINK("selfLink"),
     STATUS("status"),
-    ZONES("zones");
+    ZONES("zones"),
+    DEPRECATED("deprecated");
 
     private final String selector;
 
@@ -158,7 +165,8 @@ public interface Compute extends Service<ComputeOptions> {
     NAME("name"),
     REGION("region"),
     SELF_LINK("selfLink"),
-    STATUS("status");
+    STATUS("status"),
+    DEPRECATED("deprecated");
 
     private final String selector;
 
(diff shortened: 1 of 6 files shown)


mziccard:
I think the coverage drop is the usual one we see when we add the spi layer that is only tested through ITs (which do not concur to compute the coverage).

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
```diff
@@ -43,4 +64,214 @@ public DefaultComputeRpc(ComputeOptions options) {
...
+          .aggregatedList(this.options.projectId())
+          .setFilter(FILTER.getString(options))
+          .setMaxResults(MAX_RESULTS.getLong(options))
+          .setPageToken(PAGE_TOKEN.getString(options))
+          .setFields(FIELDS.getString(options))
+          .execute();
+      ImmutableList.Builder<DiskType> builder = ImmutableList.builder();
+      Map<String, DiskTypesScopedList> scopedList = aggregatedList.getItems();
+      if (scopedList != null) {
+        for (String key : scopedList.keySet()) {
```
aozarov: entrySet?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+    VALID_DISK_SIZE("validDiskSize"),
+    ZONE("zone");
+
+    private final String selector;
+
+    DiskTypeField(String selector) {
+      this.selector = selector;
+    }
+
+    public String selector() {
```
aozarov: I am not sure if we discussed that already.
If we don't mind to make this method `public` than we could make all the XXXField enums implement a "Selector" interface and have one static `String selector(Selector required, Selector... other)` method.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+  /**
+   * Fields of a Compute Engine DiskType resource.
+   *
+   * @see <a href="https://cloud.google.com/compute/docs/reference/latest/diskTypes#resource">Disk
+   *     Type Resource</a>
+   */
+  enum DiskTypeField {
+    CREATION_TIMESTAMP("creationTimestamp"),
+    DEFAULT_DISK_SIZE_GB("defaultDiskSizeGb"),
+    DESCRIPTION("description"),
```
aozarov: did you intentionally left out "deprecated"?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
mziccard: Yes we discussed it and opened #456 but it seems like we haven't had time to address that issue.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+  /**
+   * Fields of a Compute Engine MachineType resource.
+   *
+   * @see <a href="https://cloud.google.com/compute/docs/reference/latest/machineTypes#resource">
+   *     Machine Type Resource</a>
+   */
+  enum MachineTypeField {
+    CREATION_TIMESTAMP("creationTimestamp"),
+    DESCRIPTION("description"),
+    GUEST_CPUS("cpus"),
```
aozarov: "cpus" -> "guestCpus"?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+    ID("id"),
+    IMAGE_SPACE_GB("imageSpaceGb"),
+    MAXIMUM_PERSISTENT_DISKS("maximumPersistentDisks"),
+    MAXIMUM_PERSISTENT_DISKS_SIZE_GB("maximumPersistentDisksSizeGb"),
+    MEMORY_MB("memoryMb"),
+    NAME("name"),
+    SCRATCH_DISKS("scratchDisks"),
+    DISK_GB("diskGb"),
```
aozarov: isn't that part of "scratchDisks"?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+     */
+    public static DiskTypeOption fields(DiskTypeField... fields) {
+      return new DiskTypeOption(ComputeRpc.Option.FIELDS, DiskTypeField.selector(fields));
+    }
+  }
+
+  /**
+   * Class for specifying disk type list options.
+   */
+  class DiskTypeListOption extends Option {
```
aozarov: filter is coming later?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+
+    /**
+     * Returns an option to specify the region's fields to be returned by the RPC call. If this
+     * option is not provided all region's fields are returned. {@code RegionListOption.fields} can
+     * be used to specify only the fields of interest. {@link Region#regionId()} is always
+     * returned, even if not specified.
+     */
+    public static RegionListOption fields(RegionField... fields) {
+      StringBuilder builder = new StringBuilder();
+      builder.append("items(").append(RegionField.selector(fields)).append("),nextPageToken");
```
aozarov: With the "Selectable" interface we could also make fields -> list string a common method...

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+      return new LicenseOption(ComputeRpc.Option.FIELDS, LicenseField.selector(fields));
+    }
+  }
+
+  /**
+   * Returns the requested disk type or {@code null} if not found.
+   *
+   * @throws ComputeException upon failure
+   */
+  DiskType getDiskType(DiskTypeId diskTypeId, DiskTypeOption... options) throws ComputeException;
```
aozarov: No need for "throws ComputeException". Having it in the javadoc is enough.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +30,548 @@
...
+  /**
+   * Returns the requested disk type or {@code null} if not found.
+   *
+   * @throws ComputeException upon failure
+   */
+  DiskType getDiskType(String zone, String diskType, DiskTypeOption... options)
+      throws ComputeException;
+
+  /**
+   * Lists the disk types in the provided zone available to the current project.
```
aozarov: is the "to the current project" necessary or inferred?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
```diff
@@ -16,15 +16,469 @@
...
+      ComputeRpc.Tuple<String, Iterable<com.google.api.services.compute.model.DiskType>> result =
+          runWithRetries(new Callable<ComputeRpc.Tuple<String,
+              Iterable<com.google.api.services.compute.model.DiskType>>>() {
+            @Override
+            public ComputeRpc.Tuple<String,
+                Iterable<com.google.api.services.compute.model.DiskType>> call() {
+              return serviceOptions.rpc().listDiskTypes(optionsMap);
+            }
+          }, serviceOptions.retryParams(), EXCEPTION_HANDLER);
+      String cursor = result.x();
```
aozarov: Any reason for including it inside of the retry handling try~catch?

gcloud-java-compute/src/main/java/com/google/gcloud/spi/ComputeRpc.java
```diff
@@ -16,5 +16,155 @@
...
+      return y;
+    }
+  }
+
+  /**
+   * Returns the requested disk type or {@code null} if not found.
+   *
+   * @throws ComputeException upon failure
+   */
+  DiskType getDiskType(String zone, String diskType, Map<Option, ?> options)
```
aozarov: same comment about throws.

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
```diff
@@ -43,4 +64,214 @@ public DefaultComputeRpc(ComputeOptions options) {
...
+  @Override
+  public MachineType getMachineType(String zone, String machineType, Map<Option, ?> options)
+      throws ComputeException {
+    try {
+      return compute.machineTypes()
+          .get(this.options.projectId(), zone, machineType)
+          .setFields(FIELDS.getString(options))
+          .execute();
+    } catch (IOException ex) {
+      ComputeException serviceException = translate(ex);
```
aozarov: if you wanted you can make this re-occuring code a generic method that returns T (always null) or throws ComputeException.

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
```diff
@@ -43,4 +64,214 @@ public DefaultComputeRpc(ComputeOptions options) {
...
+    }
+  }
+
+  @Override
+  public Tuple<String, Iterable<DiskType>> listDiskTypes(String zone, Map<Option, ?> options)
+      throws ComputeException {
+    try {
+      DiskTypeList diskTypesList = compute.diskTypes()
+          .list(this.options.projectId(), zone)
+          .setFilter(FILTER.getString(options))
```
aozarov: Not used yet, right?

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
```diff
@@ -43,4 +64,214 @@ public DefaultComputeRpc(ComputeOptions options) {
   private static ComputeException translate(IOException exception) {
     return new ComputeException(exception);
   }
+
+  @Override
+  public DiskType getDiskType(String zone, String diskType, Map<Option, ?> options)
+      throws ComputeException {
```
aozarov: No need for the explicit "throws ComputeException" (I see that in Storage we do that in the spi RPC interface and implementation but I think it is a leftover as originally it was IOException). In any case, for consistency I am fine with leaving it as is (here and in the spi interface), but it should be removed from the API interface (that has a javadoc to cover it).

aozarov:
Nice work! Done with first iteration. Any reason we don't have codacy analysis here?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
mziccard: No particular reason but this method has the same structure we used almost everywhere else for listing methods.

mziccard:
@aozarov I addressed the comments in 28ed6e8

- Add support for filters: `filter` method to options + filter classes
- Removed unnecessary fields + added missing ones
- Removed redundant "throws" from method signatures
- Add static method in `DefaultComputeRpc` to handle the NOT_FOUND case

I also had to add dedicated option classes for aggregated lists as `fields` option is not completely supported (see #711).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -24,4 +35,858 @@
...
+   */
+  abstract class ListFilter implements Serializable {
+
+    private static final long serialVersionUID = -238638392811165127L;
+
+    private final String field;
+    private final ComparisonOperator operator;
+    private final Object value;
+
+    enum ComparisonOperator {
```
aozarov: public

aozarov: Oh, looks like you meant it to be package scoped. Fine in that case.

aozarov: The Beta API also supports filtering on multiple expressions. Is it something that you plan to add later?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
aozarov: Ok, Fine as is, but I don't think we need to consider that as a recommended way or a pattern we should follow (nor should we go and change other places).

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
```diff
@@ -43,4 +64,201 @@ public DefaultComputeRpc(ComputeOptions options) {
...
+
+  @Override
+  public Tuple<String, Iterable<MachineType>> listMachineTypes(Map<Option, ?> options) {
+    try {
+      MachineTypeAggregatedList aggregatedList = compute.machineTypes()
+          .aggregatedList(this.options.projectId())
+          .setFilter(FILTER.getString(options))
+          .setMaxResults(MAX_RESULTS.getLong(options))
+          .setPageToken(PAGE_TOKEN.getString(options))
+          // todo(mziccard): uncomment or remove once #711 is closed
```
aozarov: unnecessary comment?

aozarov:
Looks good. Few minor comments. Feel free to merge after they are resolved.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
mziccard: Yes we decided to stick to v1, I plan to support multiple filters as soon as the Beta API becomes v2. This should be quite simple.

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
mziccard: I forgot to comment the following line.

Issue #692, opened 2016-03-01
Support the delimiter parameter and prefixes result field for bucket listing


dimo414:
See [`delimiter`](https://cloud.google.com/storage/docs/json_api/v1/objects/list) - this is necessary to be able to list "directories".

In addition to `BlobListOption.delimiter(String)` it'd be nice if there were an argument-less method that uses `/` as the delimiter, since that's what ~99% of usages will want. I'd suggest `directories()` as that method name, so users can write `bucket.list("...", BlobListOption.directories())`.

dimo414:
Actually, I just realized this is what [`BlobListOption.recursive()`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/1583e2c8b8763a003b5d0d16d27cf53a97410603/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java#L699) (which is, imo, not a very clear name) is supposed to do. However that method treats the `delimiter` argument as a boolean; instead it should be passing in `/`. As is, I don't think it does anything meaningful. Given that the method is broken, I'd vote to rename it `directories()` which I think more clearly conveys its intent.

mziccard:
Hi @dimo414, thanks again for your feedback.
You can set the delimiter through Storage's global options:
```java
StorageOptions options = StorageOptions.builder().pathDelimiter("/").build();
```
By default "/" is used.
When you use `BlobListOption.recursive(true)` we take care of setting the `delimiter` option to the `pathDelimiter` you specified when you created the `StorageOptions`.

I agree that the option's name is counter-intuitive. In fact when set to `BlobListOption.recursive(true)` it does NOT list the content of subdirectories. We should probably make it such that `BlobListOption.recursive(false)` sets the `delimiter` option, or change the name. /cc @aozarov

dimo414:
Playing around with the API Explorer on the list doc page, I see my issue. I am trying to list "directories" which I now see are returned by the API in the `prefixes` field, not as part of the `items` field. I don't see a way to get the `prefixes` (unless, of course, I'm missing something):

Without a delimiter:
```
GET https://www.googleapis.com/storage/v1/b/my-bucket/o?prefix=dir%2F&fields=items%2Fname%2Ckind%2Cprefixes&key={YOUR_API_KEY}
{
 "kind": "storage#objects",
 "items": [
  {
   "name": "dir/1/foo"
  },
  {
   "name": "dir/2/bar"
  }
 ]
}
```

[...]

PR #697, opened 2016-03-01
Fix BlobListOption.recursive, add tests and documentation

mziccard:
This fixes the `BlobListOption.recursive` bug discussed in #692

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
@@ -695,6 +695,10 @@ public interface Storage extends Service<StorageOptions> {
 
     /**
      * Returns an option to specify whether blob listing should include subdirectories or not.
+     * {@link StorageOptions#pathDelimiter()} is used as a path delimiter. If set to {@code true}
+     * also blobs in subdirectories are listed. If set to {@code false} and used in combination
+     * with {@link #prefix(String)} only blobs in a directory can be listed. If not set also blobs
+     * in subdirectories are listed.
      */
     public static BlobListOption recursive(boolean recursive) {
       return new BlobListOption(StorageRpc.Option.DELIMITER, recursive);
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java
@@ -668,7 +668,7 @@ final class StorageImpl extends BaseService<StorageOptions> implements Storage {
       checkArgument(prev == null, "Duplicate option %s", option);
     }
     Boolean value = (Boolean) temp.remove(DELIMITER);
-    if (Boolean.TRUE.equals(value)) {
+    if (Boolean.FALSE.equals(value)) {
       temp.put(DELIMITER, options().pathDelimiter());
     }
     if (useAsSource) {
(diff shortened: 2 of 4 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -695,6 +695,10 @@ public static BlobListOption prefix(String prefix) {
 
     /**
      * Returns an option to specify whether blob listing should include subdirectories or not.
+     * {@link StorageOptions#pathDelimiter()} is used as a path delimiter. If set to {@code true}
+     * also blobs in subdirectories are listed. If set to {@code false} and used in combination
```
aozarov: I am not sure I understand the "and used in combination" part. delimiter (recursive false) can be used without prefix and should return all the blobs in the "root" folder (blobs name without any delimiter).

aozarov:
Looks good. Just one comment regarding the javadoc.
Also, do you plan to take care of the other item in the associated issue, #692 regarding the prefixes?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
dimo414: Since we're not really talking about "directories" it might be good to be more precise here and avoid using "directories" and "subdirectories" without more context. For example I noticed different behavior depending on whether my prefix ended with a trailing `/` or not, which I didn't see documented anywhere.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java
```diff
@@ -668,7 +668,7 @@ public T apply(BatchResponse.Result<T> result) {
       checkArgument(prev == null, "Duplicate option %s", option);
     }
     Boolean value = (Boolean) temp.remove(DELIMITER);
-    if (Boolean.TRUE.equals(value)) {
+    if (Boolean.FALSE.equals(value)) {
```
aozarov: Do you think we should also set it if value is `null`? Basically, what should be the default if recursive option was not specified? Whatever we pick, I think it would be worth to document
the default. Also, if we pick "non-recursive" when not specifying than we can consider removing the boolean param from the option.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
mziccard: > I am not sure I understand the "and used in combination" part.

I wanted to say that if used in combination `prefix` it enables to list the content of a specific "directory". I agree it's unclear as is.

> Since we're not really talking about "directories" it might be good to be more precise here and avoid using "directories" and "subdirectories" without more context.

I agree but this is not easy to be explained without using the concept of directory and subdirectory. In fact that is how it's explained in the [official docs](https://cloud.google.com/storage/docs/json_api/v1/objects/list).

gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java
mziccard: I think the default behavior should be recursive.

mziccard:
@aozarov The discussion in #692 is still in progress, I suggest we keep this on hold until we reach an agreement there.

mziccard:
This can be closed, replaced by #726

PR #700, opened 2016-03-01
Added retryable errors.

mderka:


gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsException.java
```diff
@@ -16,17 +16,26 @@
...
+import java.util.Set;
 
 /**
  * DNS service exception.
  */
 public class DnsException extends BaseServiceException {
 
+  // see: https://cloud.google.com/dns/troubleshooting
+  private static final Set<Error> RETRYABLE_ERRORS = ImmutableSet.of(
+      new Error(404, null),
```
aozarov: Per discussion, we probably don't want to include 404 and 409.
Also, it would be good to involve the DNS people in this PR.

aozarov:
Looks good. Please get some feedback from the DNS team before merge.

mderka:
Can we get some input here? Are there any retryable errors that you know of and we are forgetting?

/cc @benvitale @blowmage @stephenplusplus

blowmage:
The only guidance we've gotten for retries is a [500 or 503 response](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/5bc3364ec4b962b90f184c6fa96a047d0c1a14e0/lib/gcloud/backoff.rb#L62), and an error message reason for [https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/5bc3364ec4b962b90f184c6fa96a047d0c1a14e0/lib/gcloud/backoff.rb#L63](exceeding rate limits).

mderka:
Thanks @blowmage, I added the exceeding limits.

stephenplusplus:
Here are the codes that we use: https://github.com/GoogleCloudPlatform/gcloud-node/blob/058d7623ea3d54c0efc7db4757510fd53eabc492/lib/common/util.js#L271

mderka:
@stephenplusplus 429 is the error code for rateLimitExceeded and userRateLimitExceeded, right? Is there any reason why you are double checking first via code and then via error message?

stephenplusplus:
This code has been around for a while, so I'm a little rusty on why we went with the specific codes and ways of checking. But, the code check is when a response has the HTTP status code of 429, 500, 502, or 503. The other check isn't actually checking the HTTP status message, but a message that is contained in JSON response body `errors` object. I'm not sure if a situation exists where the status code could be 2xx, but the response body would return errors that occurred, but this check would handle it.

benvitale:
+ Jacob, David for additional comments

Retrying on 500 seems valuable; as this error could indicate a problem
between Apiary and our backend server.
I can't think of other cases. We generally have retry-loops around our
Spanner transactions, so if our server says things are errors, they are
unlikely to be temporary errors that could succeed if retried.

PR #704, opened 2016-03-02
Handle eventally consistent blob lists in storage ITs

mziccard:
This PR tries to make tests that involve listing blobs less fragile.
It adds a timeout to such tests and makes them wait for the list to be of the expected size. If the expected size is never reached the test fails due to the timeout.
After the size is reached the tests check for the proper blobs to be in the list.

We could consider removing timeout+size-check but that would make the tests pass also when an empty list is returned.

### DIFF
--- a/gcloud-java-storage/src/test/java/com/google/gcloud/storage/it/ITStorageTest.java
+++ b/gcloud-java-storage/src/test/java/com/google/gcloud/storage/it/ITStorageTest.java
@@ -101,13 +101,12 @@ public class ITStorageTest {
 
   @Test(timeout = 5000)
   public void testListBuckets() throws InterruptedException {
-    Iterator<Bucket> bucketIterator =
-        storage.list(Storage.BucketListOption.prefix(BUCKET),
-        Storage.BucketListOption.fields()).values().iterator();
+    Iterator<Bucket> bucketIterator = storage.list(Storage.BucketListOption.prefix(BUCKET),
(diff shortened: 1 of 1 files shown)


gcloud-java-storage/src/test/java/com/google/gcloud/storage/it/ITStorageTest.java
```diff
@@ -307,10 +308,18 @@ public void testListBlobsSelectedFields() {
     Page<Blob> page = storage.list(BUCKET,
         Storage.BlobListOption.prefix("test-list-blobs-selected-fields-blob"),
         Storage.BlobListOption.fields(BlobField.METADATA));
-    int index = 0;
+    // Listing blobs is eventually consistent, we loop until the list is of the expected size. The
+    // test fails if timeout is reached.
+    while (Iterators.size(page.values().iterator()) != 2) {
```
aozarov: `iterateAll()`?

aozarov:
Looks good. Feel free to merge with or without my suggestion.

gcloud-java-storage/src/test/java/com/google/gcloud/storage/it/ITStorageTest.java
mziccard: I used `values().iterator()` simply because I expected list of 2 elements. But you are right we better use `iterateAll` that covers all cases.

PR #705, opened 2016-03-02
Add BlobTargetOption and BlobWriteOption classes to Bucket

### LINKED ISSUE #702: Operations on Buckets should use Storage.*Option types
Some of the static classes in `Storage`, such as `BlobTargetOption` which is used by `Storage.create()`, are also being erroneously used in `Bucket`. This is problematic because `Bucket` keys of of path `Strings`, not `BlobInfo` instances, and therefore isn't able to extract information like the blob's generation, causing a line like this:

    bucket.create(file, data, null, BlobTargetOption.generationMatch());

to fail:

     java.lang.IllegalArgumentException: Option ifGenerationMatch is missing a value
         at com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
         at com.google.gcloud.storage.StorageImpl.addToOptionMap(StorageImpl.java:651)
         at com.google.gcloud.storage.StorageImpl.addToOptionMap(StorageImpl.java:643)
         at com.google.gcloud.storage.StorageImpl.optionMap(StorageImpl.java:681)
         at com.google.gcloud.storage.StorageImpl.optionMap(StorageImpl.java:660)
         at com.google.gcloud.storage.StorageImpl.optionMap(StorageImpl.java:695)
         at com.google.gcloud.storage.StorageImpl.optionMap(StorageImpl.java:703)
         at com.google.gcloud.storage.StorageImpl.create(StorageImpl.java:140)
         at com.google.gcloud.storage.StorageImpl.create(StorageImpl.java:129)
         at com.google.gcloud.storage.Bucket.create(Bucket.java:360)

And there is no way in `Storage.BlobTargetOption` to actually specify a generation.

I don't think `Bucket` should have any dependency on the `Storage.*Option` types. There's clearly a good amount of overlap, but I don't think that's a good enough reason for them to share these types. Even where they behave the same it's confusing for the caller.

They could extend from a shared parent if we wanted, but that should be an implementation detail.

I can put together a patch to remove the dependencies on `Storage.*Option` from `Bucket` if that's desired.

mziccard:
This fixes #702

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Bucket.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Bucket.java
@@ -207,8 +207,8 @@ public final class Bucket extends BucketInfo {
 
     /**
      * Returns an option for blob's metageneration match. If this option is used the request will
-     * fail if metageneration does not match the provided value. Either this option or
-     * {@link #metagenerationNotMatch(long)} can be provided at the same time.
+     * fail if metageneration does not match the provided value. This option can not be provided
+     * together with {@link #metagenerationNotMatch(long)}.
      */
     public static BlobTargetOption metagenerationMatch(long metageneration) {
       return new BlobTargetOption(StorageRpc.Option.IF_METAGENERATION_MATCH, metageneration);
@@ -216,8 +216,8 @@ public final class Bucket extends BucketInfo {
 
     /**
      * Returns an option for blob's metageneration mismatch. If this option is used the request will
-     * fail if metageneration matches the provided value. Either this option or
-     * {@link #metagenerationMatch(long)} can be provided at the same time.
+     * fail if metageneration matches the provided value. This option can not be provided together
+     * with {@link #metagenerationMatch(long)}.
      */
     public static BlobTargetOption metagenerationNotMatch(long metageneration) {
       return new BlobTargetOption(StorageRpc.Option.IF_METAGENERATION_NOT_MATCH, metageneration);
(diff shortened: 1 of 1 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/storage/Bucket.java
```diff
@@ -125,6 +129,287 @@ public static BucketSourceOption metagenerationNotMatch() {
...
+     * will fail if blob's generation matches the provided value. This option can not be provided
+     * together with {@link #generationMatch(long)} or {@link #doesNotExist()}.
+     */
+    public static BlobTargetOption generationNotMatch(long generation) {
+      return new BlobTargetOption(StorageRpc.Option.IF_GENERATION_NOT_MATCH, generation);
+    }
+
+    /**
+     * Returns an option for blob's metageneration match. If this option is used the request will
+     * fail if metageneration does not match the provided value. Either this option or
```
aozarov: "Either this ..." Why did you choose a different phrasing than what you did before "This option can not be ..."?
I would keep phrasing the same and I think the "This option can not .." is more clear.

aozarov:
Looks good. This PR does not cover *all* the comments mentioned in #702.
I think #702 was suggesting to not use any of the Storage.* options. Personally, I am fine either way.

mziccard:
It does not cover all comments because I see no reason for not using existing options when they apply

aozarov:
> It does not cover all comments because I see no reason for not using existing options when they apply.

Fine with me. Would be nice to state that in the issue when merging this PR (and auto-closing the issue).

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Bucket.java
mziccard: Fixed in 3600338

PR #706, opened 2016-03-03
Run google-java-format on GCS NIO

jart:
I also made the following additional changes:
- Don't use one-line javadocs, per request of @aozarov
- Use public domain John Keats poetry published in 1819
- Minor cleanup to a few JavaDocs

### DIFF
--- a/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStorageObjectImmutableException.java
+++ b/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStorageObjectImmutableException.java
@@ -1,7 +1,10 @@
 package com.google.gcloud.storage.contrib.nio;
 
-/** Exception thrown to indicate we don't support a mutation of a cloud storage object. */
+/**
+ * Exception thrown to indicate we don't support a mutation of a cloud storage object.
+ */
 public final class CloudStorageObjectImmutableException extends UnsupportedOperationException {
+
   CloudStorageObjectImmutableException() {
     super("Cloud Storage objects are immutable.");
   }
--- a/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStoragePath.java
+++ b/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStoragePath.java
@@ -45,19 +45,20 @@ public final class CloudStoragePath implements Path {
     this.path = path;
   }
 
-  static CloudStoragePath getPath(
-      CloudStorageFileSystem fileSystem, String path, String... more) {
+  static CloudStoragePath getPath(CloudStorageFileSystem fileSystem, String path, String... more) {
     return new CloudStoragePath(
         fileSystem, UnixPath.getPath(fileSystem.config().permitEmptyPathComponents(), path, more));
   }
 
-  /** Returns the Cloud Storage bucket name being served by this file system.
+  /**
+   * Returns the Cloud Storage bucket name being served by this file system.
    */
   public String bucket() {
     return fileSystem.bucket();
   }
 
-  /** Returns path converted to a {@link BlobId} so I/O can be performed.
+  /**
+   * Returns path converted to a {@link BlobId} so I/O can be performed.
    */
   BlobId getBlobId() {
     return BlobId.of(bucket(), toRealPath().path.toString());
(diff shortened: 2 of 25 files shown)


jart:
Ping @jean-philippe-martin

jean-philippe-martin:
LGTM, and no dig against Keats but why remove Hamlet?
Surely it's old enough to be public domain?

jart:
Hamlet was in two files. So I removed the duplication :)

gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/UnixPath.java
```diff
@@ -57,7 +57,9 @@ private UnixPath(boolean permitEmptyComponents, String path) {
     this.permitEmptyComponents = permitEmptyComponents;
   }
 
-  /** Returns new UnixPath of {@code first}. */
+  /**
+   * Returns new UnixPath of {@code first}.
```
ajkannan: Nit: could use {@code UnixPath} here

gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStorageObjectImmutableException.java
```diff
@@ -1,7 +1,10 @@
 package com.google.gcloud.storage.contrib.nio;
 
-/** Exception thrown to indicate we don't support a mutation of a cloud storage object. */
+/**
+ * Exception thrown to indicate we don't support a mutation of a cloud storage object.
```
ajkannan: Super nit: we can capitalize "Cloud Storage" (it's lower case in a couple other places, so maybe a search-replace might be helpful)

ajkannan:
It seems like there are a couple missed one/two-liner javadocs in CloudStorageConfiguration (for example [here](https://github.com/GoogleCloudPlatform/gcloud-java/blob/gcs-nio/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStorageConfiguration.java#L34) and [here](https://github.com/GoogleCloudPlatform/gcloud-java/blob/gcs-nio/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStorageConfiguration.java#L53)).  There are also a few scattered in CloudStorageFileAttributesTest and CloudStoragePathTest.

gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStoragePath.java
```diff
@@ -273,9 +277,9 @@ public int compareTo(Path other) {
   public boolean equals(@Nullable Object other) {
     return this == other
         || other instanceof CloudStoragePath
-        && Objects.equals(bucket(), ((CloudStoragePath) other).bucket())
-        && Objects.equals(toRealPathInternal(false),
-            ((CloudStoragePath) other).toRealPathInternal(false));
+            && Objects.equals(bucket(), ((CloudStoragePath) other).bucket())
```
aozarov: I mentioned it in the prior PR but I don't remember an answer (could be my fault).
FindBugs complains that setting `other` as `@Nullable` is wrong as it is not compatible with the method it overrides.
Clearly, it is a harmless message but I would still like to raise (I am actually not sure what is the recommendation for it).

jart: Sorry I forgot about that one. Thanks for bringing it up. I've fixed this in #715. I also remembered what you said about tests throwing exceptions, and fixed that too :)

gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/UnixPath.java
jart: Changed to "path" so it doesn't need code font in #715.

gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/gcloud/storage/contrib/nio/CloudStorageObjectImmutableException.java
jart: Good catch. I grepped and made sure I got all instances of this in #715.

PR #707, opened 2016-03-03
Bringing DNS branch up to speed with master

mderka:


jgeewax:
@ajkannan @mderka : I updated the title of the PR, but added a ? since it isn't super clear at a glance what the purpose is (if I had to guess I'd say we're effectively bringing our DNS branch up to speed with master?)

Either way, any chance you could change the title to be more descriptive to outsiders ?

If I'm right in my guess, maybe we can say "Bringing DNS branch up to speed with master"?

ajkannan:
Correct and fixed.

PR #712, opened 2016-03-04
Remove redundant throws from services method signatures

mziccard:
This PR removes redundant `throws` from service and rpc methods. This also adds basic javadoc to RPC methods.


aozarov:
Thank you for doing that!

PR #715, opened 2016-03-04
Correct additional style issues and remove warnings

jart:
This change is a follow-up to post-merge comments in #706 made by @ajkannan and @aozarov. The following changes have been made:
- Don't use @Nullable on equals() methods
- Don't `throws Exception` in test methods unless needed
- Use incomplete sentences in JavaDoc summary fragment
- Capitalize Cloud Storage
- Get rid of some one-line JavaDocs
- Use @code formatting when appropriate
- Remove Paths.get() usages which is deprecated in Java 8
- Remove @SuppressWarnings("null") which no one uses

CC: @jean-philippe-martin

### DIFF
--- a/gcloud-java-contrib/gcloud-java-nio/src/test/java/com/google/gcloud/storage/contrib/nio/CloudStorageFileAttributesTest.java
+++ b/gcloud-java-contrib/gcloud-java-nio/src/test/java/com/google/gcloud/storage/contrib/nio/CloudStorageFileAttributesTest.java
@@ -19,6 +19,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
+import java.io.IOException;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -35,7 +36,6 @@ public class CloudStorageFileAttributesTest {
   private Path path;
   private Path dir;
 
-  /** empty test storage and make sure we use it instead of the real GCS. Create a few paths. **/
   @Before
   public void before() {
     CloudStorageFileSystemProvider.setGCloudOptions(LocalGcsHelper.options());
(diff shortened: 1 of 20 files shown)


gcloud-java-contrib/gcloud-java-nio/src/test/java/com/google/gcloud/storage/contrib/nio/CloudStorageFileAttributesTest.java
```diff
@@ -35,7 +36,6 @@
   private Path path;
   private Path dir;
 
-  /** empty test storage and make sure we use it instead of the real GCS. Create a few paths. **/
   @Before
   public void before() {
```
jean-philippe-martin: Why remove this comment?

jean-philippe-martin:
Looks good to me, but let's wait to hear from the others.

gcloud-java-contrib/gcloud-java-nio/src/test/java/com/google/gcloud/storage/contrib/nio/CloudStorageFileAttributesTest.java
jart: Seemed easily inferred from the code itself, and not meriting a javadoc. I can put it back if you like.

jean-philippe-martin: No, leave it out, I was just curious. I'm glad the tools are not getting in the way of removing unnecessary comments.

ajkannan:
Question about your bullet point "Remove Paths.get() usages which is deprecated in Java 8".  I couldn't find online a reference that it's being deprecated; did I misunderstand that bullet?  I see that you changed the `Paths.get` that uses the string as an argument.  If the string argument is deprecated, then I see one more example (line 633 of CloudStorageFileSystemProviderTest) that uses a string.  All other changes look good to me.

jart:
Hmm it's possible that the deprecation was added to the JavaDoc only internally at Google. But it makes sense. It seems more readable for us to be doing it the way I refactored it.

PR #718, opened 2016-03-05
Add get, replace, and test for IAM

### LINKED ISSUE #439: Figure out the work needed for IAM Policy Manager service
https://cloud.google.com/iam/docs/managing-policies

Note: the maven artifact for this service is bundled with the Resource Manager artifact ([com.google.api.services.cloudresourcemanager](https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1beta1/java/latest/)).

ajkannan:
Fixes #439.

There are two outstanding issues to resolve regarding IAM:
- The string values corresponding to the permissions for `list` and `create` are correctly formed.  However, when I tested manually, I found that they cause a 400 error citing bad request.  I will ask about this.
- Version seems to be a completely useless field, because people will unset it completely by creating and setting a fresh policy (not using the read-write-modify pattern).  Should we remove version entirely from the `IamPolicy` class in core?

### DIFF
--- a/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Policy.java
+++ b/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Policy.java
@@ -17,18 +17,19 @@
 package com.google.gcloud.resourcemanager;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.CaseFormat;
 import com.google.common.base.Function;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.gcloud.IamPolicy;
 import com.google.gcloud.Identity;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -48,40 +49,101 @@ public class Policy extends IamPolicy<Policy.Role> {
   /**
    * Represents legacy roles in an IAM Policy.
    */
-  public enum Role {
+  public static class Role implements Serializable {
 
     /**
-     * Permissions for read-only actions that preserve state.
+     * The recognized roles in a Project's IAM policy.
      */
-    VIEWER("roles/viewer"),
+    public enum Type {
 
-    /**
-     * All viewer permissions and permissions for actions that modify state.
-     */
-    EDITOR("roles/editor"),
+      /**
+       * Permissions for read-only actions that preserve state.
+       */
+      VIEWER,
 
-    /**
-     * All editor permissions and permissions for the following actions:
-     * <ul>
-     * <li>Manage access control for a resource.
-     * <li>Set up billing (for a project).
-     * </ul>
-     */
-    OWNER("roles/owner");
+      /**
+       * All viewer permissions and permissions for actions that modify state.
+       */
+      EDITOR,
 
-    private String strValue;
-
-    private Role(String strValue) {
-      this.strValue = strValue;
+      /**
+       * All editor permissions and permissions for the following actions:
+       * <ul>
+       * <li>Manage access control for a resource.
+       * <li>Set up billing (for a project).
+       * </ul>
+       */
+      OWNER
     }
 
-    String strValue() {
-      return strValue;
+    private static final long serialVersionUID = 2421978909244287488L;
+
+    private final String value;
+    private final Type type;
+
+    private Role(String value, Type type) {
+      this.value = value;
+      this.type = type;
+    }
+
+    String value() {
+      return value;
+    }
+
+    /**
+     * Returns the type of role (editor, owner, or viewer).  Returns {@code null} if the role type
+     * is unrecognized.
+     */
+    public Type type() {
+      return type;
+    }
+
+    /**
+     * Returns a {@code Role} of type {@link Type#VIEWER VIEWER}.
+     */
+    public static Role viewer() {
+      return new Role("roles/viewer", Type.VIEWER);
+    }
+
+    /**
+     * Returns a {@code Role} of type {@link Type#EDITOR EDITOR}.
+     */
+    public static Role editor() {
+      return new Role("roles/editor", Type.EDITOR);
+    }
+
+    /**
+     * Returns a {@code Role} of type {@link Type#OWNER OWNER}.
+     */
+    public static Role owner() {
+      return new Role("roles/owner", Type.OWNER);
+    }
+
+    static Role rawRole(String roleStr) {
+      return new Role(roleStr, null);
     }
 
     static Role fromStr(String roleStr) {
-      return Role.valueOf(CaseFormat.LOWER_CAMEL.to(
-          CaseFormat.UPPER_UNDERSCORE, roleStr.substring("roles/".length())));
+      try {
+        Type type = Type.valueOf(roleStr.split("/")[1].toUpperCase());
+        return new Role(roleStr, type);
+      } catch (Exception ex) {
+        return new Role(roleStr, null);
+      }
+    }
+
+    @Override
+    public final int hashCode() {
+      return Objects.hash(value, type);
+    }
+
+    @Override
+    public final boolean equals(Object obj) {
+      if (!(obj instanceof Role)) {
+        return false;
+      }
+      Role other = (Role) obj;
+      return Objects.equals(value, other.value()) && Objects.equals(type, other.type());
     }
   }
 
@@ -124,7 +186,7 @@ public class Policy extends IamPolicy<Policy.Role> {
     for (Map.Entry<Role, Set<Identity>> binding : bindings().entrySet()) {
       com.google.api.services.cloudresourcemanager.model.Binding bindingPb =
           new com.google.api.services.cloudresourcemanager.model.Binding();
-      bindingPb.setRole(binding.getKey().strValue());
+      bindingPb.setRole(binding.getKey().value());
       bindingPb.setMembers(
           Lists.transform(
               new ArrayList<>(binding.getValue()),
--- a/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
+++ b/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
@@ -157,10 +157,10 @@ public class Project extends ProjectInfo {
    * completes, the project is not retrievable by the {@link ResourceManager#get} and
    * {@link ResourceManager#list} methods. The caller must have modify permissions for this project.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/delete">Cloud
-   * Resource Manager delete</a>
    * @throws ResourceManagerException upon failure
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/delete">Cloud
+   *     Resource Manager delete</a>
    */
   public void delete() {
     resourceManager.delete(projectId());
@@ -174,10 +174,10 @@ public class Project extends ProjectInfo {
    * state of {@link ProjectInfo.State#DELETE_IN_PROGRESS}, the project cannot be restored. The
    * caller must have modify permissions for this project.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/undelete">Cloud
-   * Resource Manager undelete</a>
    * @throws ResourceManagerException upon failure (including when the project can't be restored)
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/undelete">Cloud
+   *     Resource Manager undelete</a>
    */
   public void undelete() {
     resourceManager.undelete(projectId());
@@ -188,11 +188,11 @@ public class Project extends ProjectInfo {
    *
    * <p>The caller must have modify permissions for this project.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/update">Cloud
-   * Resource Manager update</a>
    * @return the Project representing the new project metadata
    * @throws ResourceManagerException upon failure
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/update">Cloud
+   *     Resource Manager update</a>
    */
   public Project replace() {
     return resourceManager.replace(this);
--- a/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
+++ b/gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
@@ -18,10 +18,12 @@ package com.google.gcloud.resourcemanager;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.Sets;
+import com.google.gcloud.IamPolicy;
 import com.google.gcloud.Page;
 import com.google.gcloud.Service;
 import com.google.gcloud.resourcemanager.spi.ResourceManagerRpc;
 
+import java.util.List;
 import java.util.Set;
 
 /**
@@ -168,19 +170,51 @@ public interface ResourceManager extends Service<ResourceManagerOptions> {
   }
 
   /**
+   * The permissions associated with a Google Cloud project. These values can be used when calling
+   * {@link #testPermissions}.
+   *
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/docs/access-control-proj#project-level_roles">
+   *     Project-level roles</a>
+   */
+  enum Permission {
+    DELETE("delete"),
+    GET("get"),
+    GET_POLICY("getIamPolicy"),
+    REPLACE("update"),
+    REPLACE_POLICY("setIamPolicy"),
+    UNDELETE("undelete");
+
+    private static final String PREFIX = "resourcemanager.projects.";
+
+    private final String value;
+
+    Permission(String suffix) {
+      this.value = PREFIX + suffix;
+    }
+
+    /**
+     * Returns the string representation of the permission.
+     */
+    public String value() {
+      return value;
+    }
+  }
+
+  /**
    * Creates a new project.
    *
    * <p>Initially, the project resource is owned by its creator exclusively. The creator can later
    * grant permission to others to read or update the project. Several APIs are activated
    * automatically for the project, including Google Cloud Storage.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/create">Cloud
-   * Resource Manager create</a>
    * @return Project object representing the new project's metadata. The returned object will
    *     include the following read-only fields supplied by the server: project number, lifecycle
    *     state, and creation time.
    * @throws ResourceManagerException upon failure
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/create">Cloud
+   *     Resource Manager create</a>
    */
   Project create(ProjectInfo project);
 
@@ -201,10 +235,10 @@ public interface ResourceManager extends Service<ResourceManagerOptions> {
    * completes, the project is not retrievable by the {@link ResourceManager#get} and
    * {@link ResourceManager#list} methods. The caller must have modify permissions for this project.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/delete">Cloud
-   * Resource Manager delete</a>
    * @throws ResourceManagerException upon failure
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/delete">Cloud
+   *     Resource Manager delete</a>
    */
   void delete(String projectId);
 
@@ -214,10 +248,9 @@ public interface ResourceManager extends Service<ResourceManagerOptions> {
    * <p>Returns {@code null} if the project is not found or if the user doesn't have read
    * permissions for the project.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/get">Cloud
-   * Resource Manager get</a>
    * @throws ResourceManagerException upon failure
+   * @see <a href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/get">
+   *     Cloud Resource Manager get</a>
    */
   Project get(String projectId, ProjectGetOption... options);
 
@@ -228,11 +261,11 @@ public interface ResourceManager extends Service<ResourceManagerOptions> {
    * at the end of the list. Use {@link ProjectListOption} to filter this list, set page size, and
    * set page tokens.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/list">Cloud
-   * Resource Manager list</a>
    * @return {@code Page<Project>}, a page of projects
    * @throws ResourceManagerException upon failure
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/list">Cloud
+   *     Resource Manager list</a>
    */
   Page<Project> list(ProjectListOption... options);
 
@@ -241,11 +274,11 @@ public interface ResourceManager extends Service<ResourceManagerOptions> {
    *
    * <p>The caller must have modify permissions for this project.
    *
-   * @see <a
-   * href="https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/update">Cloud
-   * Resource Manager update</a>
    * @return the Project representing the new project metadata
    * @throws ResourceManagerException upon failure
+   * @see <a href=
+   *     "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/update">Cloud
+   *     Resource Manager update</a>
    */
   Project replace(ProjectInfo newProject);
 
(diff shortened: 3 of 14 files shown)


aozarov:
build seems to be broken for Java 8 due to javadoc issues.

ajkannan:
Thanks for the heads up, fixed the javadoc issue.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +291,72 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   * Resource Manager getIamPolicy</a>
+   * @throws ResourceManagerException upon failure
+   */
+  Policy getPolicy(String projectId);
+
+  /**
+   * Sets the IAM access control policy for the specified project. Replaces any existing policy. The
+   * following constraints apply:
+   * <ul>
+   * <li>Projects currently support only <I>user:{emailid}</I> and <I>serviceAccount:{emailid}</I>
```
mziccard: Replace `<I>` with `<i>` and `</I>` with `</i>`

Also here and elsewhere: does closing `<li>` make javadoc fail? If not we should add `</li>`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +291,72 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   * <li>Calling this method requires enabling the App Engine Admin API.
+   * </ul>
+   * Note: Removing service accounts from policies or changing their roles can render services
+   * completely inoperable. It is important to understand how the service account is being used
+   * before removing or updating its roles.
+   *
+   * It is recommended that you use the read-modify-write pattern. This pattern entails reading the
+   * project's current policy, updating it locally, and then sending the modified policy for
+   * writing. Cloud IAM solves the problem of conflicting processes simultaneously attempting to
+   * modify a policy by using the etag property. This property is used to verify whether the
```
mziccard: Nit: link to etag?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +291,72 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   *
+   * @see <a href=
+   * "https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/setIamPolicy">
+   * Resource Manager setIamPolicy</a>
+   * @throws ResourceManagerException upon failure
+   */
+  Policy replacePolicy(String projectId, Policy newPolicy);
+
+  /**
+   * Returns the permissions that a caller has on the specified project. You typically don't call
```
mziccard: It might be worth clarifying how the input list of permissions relates to the output list of booleans. Also do you think a vararg overload of this method could be useful?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerImpl.java
```diff
@@ -181,6 +183,61 @@ public Void call() {
...
+              new Callable<com.google.api.services.cloudresourcemanager.model.Policy>() {
+                @Override
+                public com.google.api.services.cloudresourcemanager.model.Policy call() {
+                  return resourceManagerRpc.getPolicy(projectId);
+                }
+              },
+              options().retryParams(),
+              EXCEPTION_HANDLER);
+      return answer == null ? null : Policy.fromPb(answer);
+    } catch (RetryHelperException e) {
```
mziccard: Nit: avoid single letter variable names.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerImpl.java
```diff
@@ -181,6 +183,61 @@ public Void call() {
...
+  public Policy replacePolicy(final String projectId, final Policy newPolicy) {
+    try {
+      return Policy.fromPb(runWithRetries(
+          new Callable<com.google.api.services.cloudresourcemanager.model.Policy>() {
+            @Override
+            public com.google.api.services.cloudresourcemanager.model.Policy call() {
+              return resourceManagerRpc.replacePolicy(projectId, newPolicy.toPb());
+            }
+          }, options().retryParams(), EXCEPTION_HANDLER));
+    } catch (RetryHelperException e) {
```
mziccard: Nit: avoid single letter variable names.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerImpl.java
```diff
@@ -181,6 +183,61 @@ public Void call() {
...
+  public List<Boolean> testPermissions(final String projectId, final List<Permission> permissions) {
+    try {
+      return runWithRetries(
+          new Callable<List<Boolean>>() {
+            @Override
+            public List<Boolean> call() {
+              return resourceManagerRpc.testPermissions(projectId,
+                  Lists.transform(permissions, new Function<Permission, String>() {
+                      @Override
+                      public String apply(Permission p) {
```
mziccard: Nit: avoid single letter variable names.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -45,7 +53,25 @@
...
  *
  * <p>The mock runs in a separate thread, listening for HTTP requests on the local machine at an
- * ephemeral port.
+ * ephemeral port. While this mock attempts to simulate the Cloud Resource Manager, there are some
+ * divergences in behavior. The following is a non-exhaustive list of some of those behavioral
+ * differences:
+ *
+ * <ul>
+ * <li>This mock assumes you have adequate permissions for any action. Related to this,
+ *     <I>testIamPermissions</I> always indicates that the caller has all permissions listed in the
```
mziccard: Here and in the following: what about replacing `<I>` and `</I>` with `<i>` and `</i>` and adding missing `</li>`?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -168,7 +169,34 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+    OWN("own"),
+    REPLACE("update"),
+    REPLACE_POLICY("setIamPolicy"),
+    SET_BILLING("setBillingAccount"),
+    UNDELETE("undelete");
+
+    private final String strValue;
+
+    Permission(String suffix) {
+      this.strValue = "resourcemanager.projects." + suffix;
```
mziccard: Nit: the prefix could be made a `static final` field

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/spi/DefaultResourceManagerRpc.java
```diff
@@ -107,5 +116,50 @@ public Project replace(Project project) {
...
+
+  @Override
+  public List<Boolean> testPermissions(String projectId, List<String> permissions)
+      throws ResourceManagerException {
+    try {
+      TestIamPermissionsResponse response = resourceManager.projects()
+          .testIamPermissions(
+              projectId, new TestIamPermissionsRequest().setPermissions(permissions))
+          .execute();
+      Set<String> permissionsOwned = ImmutableSet.copyOf(response.getPermissions());
```
mziccard: `response.getPermissions()` can never be null?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/spi/ResourceManagerRpc.java
```diff
@@ -121,5 +123,27 @@ public Y y() {
...
 
+  /**
+   * Returns the IAM policy associated with a project.
+   *
+   * @throws ResourceManagerException upon failure
+   */
+  Policy getPolicy(String projectId);
+
+  /**
+   * Replaces the IAM policy associated with the given project ID.
```
mziccard: Nit: in `getPolicy` we say "associated with a project" while here ewe say "with the given project ID". I think we should either refer to the project or the project ID. My preference is for removing ID.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -168,7 +169,34 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+    }
+
+    String strValue() {
+      return strValue;
+    }
+  }
+
+  /**
+   * Create a new project.
```
mziccard: Nit s/Create/Creates

mziccard:
> The string values corresponding to the permissions for list and create are correctly formed. However, when I tested manually, I found that they cause a 400 error citing bad request. I will ask about this.

Should be open an issue for this?

> Version seems to be a completely useless field, because people will unset it completely by creating and setting a fresh policy (not using the read-write-modify pattern). Should we remove version entirely from the IamPolicy class in core?

Could there be other possible uses of `version` in user land? Did you try to ask service owners for this?

I just completed the first pass and added a bunch of minor comments, great work!

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
ajkannan: Will replace the capital "I"s with lowercase ones.  For the second point, I do wonder if it makes the javadoc (very slightly) more messy for people ready the source itself.  When I look up javadoc online, I usually don't see people adding the `</li>`, e.g. [in Guava](https://github.com/google/guava/search?utf8=%E2%9C%93&q=%3Cli%3E)

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/spi/DefaultResourceManagerRpc.java
ajkannan: Nice catch, according to the [javadoc](https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1beta1/java/latest/), yes.  Will fix.

ajkannan:
Thanks for the first pass @mziccard!  I incorporated all comments except the `</li>` tags, which I don't feel super strongly about either way, but I think it's slightly more readable and common as is.

I created issue #722 for the `create` and `list` permissions problem.  For the version issue, I checked with the IAM service owners, and they say that version is primarily used to differentiate between people with access to alpha features and people who don't have that access.  It's a temporary field that's useful during this "bridge" period when there are legacy roles (i.e. owner, editor, viewer), and more specialized roles that not everybody has access to at the moment.  I would consider keeping it, but it's behavior is somewhat erratic when interfacing with the resource manager service.

mziccard:
@ajkannan Added a few more nits, looks good to me!

ajkannan:
Added fixes for the latest round of comments.  I think we should keep this PR open until I clear up #722.

mziccard:
@ajkannan Sounds good!

ajkannan:
Removed create and list in the latest comment (as per reasoning in #722).

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -168,6 +170,33 @@ public static ProjectListOption fields(ProjectField... fields) {
   }
 
   /**
+   * The permissions associated with a Google Cloud project. These values can be used when calling
+   * {@link #testPermissions}.
+   */
+  public enum Permission {
```
aozarov: redundant public

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -168,6 +170,33 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+    DELETE("delete"),
```
aozarov: lets verify that this is the list the service wants to expose.
Would be good if the service updates the doc with this list (like other services that uses IAM) and for us to `@see` that page.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -168,6 +170,33 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+
+    private static final String PREFIX = "resourcemanager.projects.";
+
+    private final String strValue;
+
+    Permission(String suffix) {
+      this.strValue = PREFIX + suffix;
+    }
+
+    String strValue() {
```
aozarov: i would rather call it `value`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +292,90 @@ public static ProjectListOption fields(ProjectField... fields) {
    * @throws ResourceManagerException upon failure
    */
   void undelete(String projectId);
+
+  /**
+   * Returns the IAM access control policy for the specified project. Returns null if the resource
```
aozarov: nit {@code null}

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +292,90 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   * does not exist or if you do not have adequate permission to view the project.
+   *
+   * @see <a
```
aozarov: wrong [order](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#tag) of tags.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +292,90 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   * <li>There must be at least one owner who has accepted the Terms of Service (ToS) agreement in
+   *     the policy. An attempt to set a policy that removes the last ToS-accepted owner from the
+   *     policy will fail.
+   * <li>Calling this method requires enabling the App Engine Admin API.
+   * </ul>
+   * Note: Removing service accounts from policies or changing their roles can render services
+   * completely inoperable. It is important to understand how the service account is being used
+   * before removing or updating its roles.
+   *
+   * It is recommended that you use the read-modify-write pattern. This pattern entails reading the
```
aozarov: `<p>`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +292,90 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   * project's current policy, updating it locally, and then sending the modified policy for
+   * writing. Cloud IAM solves the problem of conflicting processes simultaneously attempting to
+   * modify a policy by using the {@link IamPolicy#etag etag} property. This property is used to
+   * verify whether the policy has changed since the last request. When you make a request to Cloud
+   * IAM with an etag value, Cloud IAM compares the etag value in the request with the existing etag
+   * value associated with the policy. It writes the policy only if the etag values match. If an
+   * etag is not provided, the policy is overwritten blindly.
+   *
+   * An example of using the read-write-modify pattern is as follows:
```
aozarov: `<p>`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -186,6 +212,30 @@ public void handle(HttpExchange exchange) {
...
+          return testPermissions(projectIdFromUri(path),
+              jsonFactory.fromString(requestBody, TestIamPermissionsRequest.class)
+                  .getPermissions());
+        default:
+          return Error.BAD_REQUEST.response(
+              "The server could not understand the following request URI: POST " + path);
+      }
+    }
+  }
+
```
aozarov: Unrelated comment. can you replace `Integer.valueOf` with `Integer.parseInt` (line 312)?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -539,6 +594,60 @@ synchronized Response undelete(String projectId) {
...
+    } catch (IOException e) {
+      return Error.INTERNAL_ERROR.response(
+          "Error when serializing the IAM policy for " + projectId);
+    }
+  }
+
+  synchronized Response replacePolicy(String projectId, Policy policy) {
+    Policy originalPolicy = policies.get(projectId);
+    if (originalPolicy == null) {
+        return Error.PERMISSION_DENIED.response("Error when replacing the policy for "
```
aozarov: wrong indent.

gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/LocalResourceManagerHelperTest.java
```diff
@@ -92,6 +102,13 @@ public void testCreate() {
     assertNull(returnedProject.getParent());
     assertNotNull(returnedProject.getProjectNumber());
     assertNotNull(returnedProject.getCreateTime());
+    com.google.api.services.cloudresourcemanager.model.Policy policy =
+        rpc.getPolicy(PARTIAL_PROJECT.getProjectId());
+    assertEquals(Collections.EMPTY_LIST, policy.getBindings());
```
aozarov: Prefer using `Collections.emptyList()`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +292,90 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   * completely inoperable. It is important to understand how the service account is being used
+   * before removing or updating its roles.
+   *
+   * It is recommended that you use the read-modify-write pattern. This pattern entails reading the
+   * project's current policy, updating it locally, and then sending the modified policy for
+   * writing. Cloud IAM solves the problem of conflicting processes simultaneously attempting to
+   * modify a policy by using the {@link IamPolicy#etag etag} property. This property is used to
+   * verify whether the policy has changed since the last request. When you make a request to Cloud
+   * IAM with an etag value, Cloud IAM compares the etag value in the request with the existing etag
+   * value associated with the policy. It writes the policy only if the etag values match. If an
```
aozarov: We should probably say what happens when etags are not matched (exception of type X?).

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
```diff
@@ -263,4 +292,90 @@ public static ProjectListOption fields(ProjectField... fields) {
...
+   * etag is not provided, the policy is overwritten blindly.
+   *
+   * An example of using the read-write-modify pattern is as follows:
+   * <pre> {@code
+   * Policy currentPolicy = resourceManager.getPolicy("my-project-id");
+   * Policy modifiedPolicy =
+   *     current.toBuilder().removeIdentity(Role.VIEWER, Identity.user("user@gmail.com"));
+   * Policy newPolicy = resourceManager.setPolicy("my-project-id", modified);
```
aozarov: `replacePolicy`?

gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/ResourceManagerImplTest.java
```diff
@@ -321,6 +330,60 @@ public void testUndelete() {
...
+    assertEquals(POLICY.bindings(), newPolicy.bindings());
+    assertNotNull(newPolicy.etag());
+    assertNotEquals(originalEtag, newPolicy.etag());
+  }
+
+  @Test
+  public void testTestPermissions() {
+    List<Permission> permissions = ImmutableList.of(Permission.GET);
+    try {
+      RESOURCE_MANAGER.testPermissions("nonexistent-project", permissions);
```
mderka: Missing fail.

gcloud-java-resourcemanager/src/test/java/com/google/gcloud/resourcemanager/ResourceManagerImplTest.java
```diff
@@ -321,6 +330,60 @@ public void testUndelete() {
...
+    } catch (ResourceManagerException e) {
+      assertEquals(403, e.code());
+      assertEquals("Project nonexistent-project not found.", e.getMessage());
+    }
+    RESOURCE_MANAGER.create(PARTIAL_PROJECT);
+    assertEquals(ImmutableList.of(true),
+        RESOURCE_MANAGER.testPermissions(PARTIAL_PROJECT.projectId(), permissions));
+    assertEquals(ImmutableList.of(true, true),
+        RESOURCE_MANAGER.testPermissions(
```
mderka: Is this an ideal wrapping? I would move line 382 to 381 and break this assert over two lines only.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
aozarov: @mderka will s/to view the project/to get the policy/ make it more clear?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManagerImpl.java
```diff
@@ -176,11 +179,74 @@ public Void call() {
...
+            }
+          }, options().retryParams(), EXCEPTION_HANDLER);
+    } catch (RetryHelperException ex) {
+      throw ResourceManagerException.translateAndThrow(ex);
+    }
+  }
+
+  @Override
+  public List<Boolean> testPermissions(String projectId, Permission first, Permission... others) {
+    List<Permission> permissions = Lists.newArrayListWithExpectedSize(others.length + 1);
```
aozarov: [`Lists.asList(first, others)`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Lists.html#asList(E,%20E[])).

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -61,6 +87,10 @@
   private static final Pattern LIST_FIELDS_PATTERN =
       Pattern.compile("(.*?)projects\\((.*?)\\)(.*?)");
   private static final String[] NO_FIELDS = {};
+  private static final String PERMISSION_PREFIX = "resourcemanager.projects.";
+  private static final Set<String> PERMISSION_SUFFIXES = ImmutableSet.of("create", "delete", "get",
```
aozarov: do you want to leave "create" and "list" here?

Also, you can use a static initializer to dynamically populate this list from the enum values.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
aozarov: Also, the relation of project to policy are 1:1 and there would always be a Policy for a project (is tha  t right @ajkannan ?) This is why this method only gets project id rather than (projectid, name).
I am fine with keeping it as is, however I understand your point, @mderka.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -539,6 +594,60 @@ synchronized Response undelete(String projectId) {
...
+    policies.put(projectId, policy);
+    try {
+      return new Response(HTTP_OK, jsonFactory.toString(policy));
+    } catch (IOException e) {
+      return Error.INTERNAL_ERROR.response(
+          "Error when serializing the policy for project " + projectId);
+    }
+  }
+
+  Response testPermissions(String projectId, List<String> permissions) {
```
aozarov: `synchronized`

aozarov:
Nice job. Done with first iteration.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
mderka: No, I think that this is good enough. It was more of a conceptual question. In other places, we throw exception if a container does not exist (e.g., trying to get a DNS record from a non-existent zone). If we consider a project to be a container for the policy, we should consider the same approach. If we consider policy to be an independent entity which may be associated to and deassociated from a project, then this is fine.

(If you discussed this before, feel free to ignore this comment.)

ajkannan:
Pushed commit [4117d61](https://github.com/GoogleCloudPlatform/gcloud-java/commit/4117d61d8e284c6afca1a663383a243f1f0ba121) to incorporate all comments thus far + a rebase to take care of merge conflicts.  In addition, the following changes were made as discussed verbally:
* Rename `id` in the `Identity` class to `value`
* Make Role a class instead of an enum to handle unsupported roles.  I haven't added `roles/resourcemanager.browser` because it is in beta.
* Remove the `own` and `setBillingAccount` permissions (in addition the previously removed `create` and `list` permissions)

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: Heads up: to use a static initializer in this package I made "Permission.value()" public.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/ResourceManager.java
ajkannan: Yup, there's a 1:1 correspondence between policies and projects.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Policy.java
```diff
@@ -48,17 +49,33 @@
   /**
    * Represents legacy roles in an IAM Policy.
    */
-  public enum Role {
+  public static class Role implements Serializable {
+
+    private static final long serialVersionUID = 2421978909244287488L;
+
+    private String value;
```
aozarov: final

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -540,6 +597,59 @@ synchronized Response undelete(String projectId) {
...
+      if (!PERMISSIONS.contains(p)) {
+        return Error.INVALID_ARGUMENT.response("Invalid permission: " + p);
+      }
+    }
+    try {
+      return new Response(HTTP_OK,
+          jsonFactory.toString(new TestIamPermissionsResponse().setPermissions(permissions)));
+    } catch (IOException e) {
+      return Error.INTERNAL_ERROR.response(
+          "Error when serializing permissions " + Arrays.asList(permissions));
```
aozarov: `ImmutableList.of(permissions)` or `Collections.singletonList(permissions)`

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
```diff
@@ -157,10 +157,10 @@ public Project reload() {
    * completes, the project is not retrievable by the {@link ResourceManager#get} and
```
aozarov: Unrelated to this PR. The Builder projectId, removeLabel, clearLabels (and parent, which is not that important) seems not to be covered by unit-test

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
aozarov: sounds good.
Nit,
```java
  private static final Set<String> PERMISSIONS;

  static {
    ImmutableSet.Builder<String> builder = ImmutableSet.builder();
    for (Permission permission : Permission.values()) {
      builder.add(permission.value());
    }
    PERMISSIONS = builder.build();
```

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Policy.java
```diff
@@ -48,17 +49,33 @@
   /**
    * Represents legacy roles in an IAM Policy.
    */
-  public enum Role {
+  public static class Role implements Serializable {
```
aozarov: Per discussion lets add a nested enum type.

ajkannan:
Incorporated comments, but instead of the ImmutableList.of suggestion, I just passed back the original list `permissions`.  I think the Arrays.asList was a vestige of when I passed in an array rather than a list.

aozarov:
Looks good. Feel free to merge after applying the comments that we discussed.

PR #719, opened 2016-03-06
Add NoAuthCredentials

### LINKED ISSUE #710: Auth issue when using revoked Cloud SDK creds with local test helpers.
When running code that uses local service emulators, users that have Google Cloud SDK installed but have out-of-date or revoked credentials (and don't specify other credentials in any other way) will see error messages when making API calls that send requests to the local emulator.  This is because we detect the Google Cloud SDK json credentials file exists, and use those to set credentials in `ServiceOptions`.  As a result, requests attempt to use these credentials (even though they aren't necessary) and fail.  

Should we:
1) Document that you should use `gcloud auth login` when using local test helpers?
or 
2) Figure out how to parse the credentials to figure out that they're valid or revoked?
I notice that under ".config/gcloud/legacy_credentials", the folder corresponding to my gcloud email account is removed when I revoke credentials.  However, I'm unsure that that's a valid way to check for revoked credentials.  The other files related to credentials in that directory don't seem to indicate whether they are valid or revoked.

ajkannan:
Fixes #710.

We should add `.authCredentials(AuthCredentials.noAuth())` to the `LocalDnsHelper.options()` method after this PR is merged in.

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
@@ -258,9 +258,9 @@ public abstract class AuthCredentials implements Restorable<AuthCredentials> {
   }
 
   /**
-   * Represents that requests sent to the server should not be authenticated. This is typically
-   * useful when using the local service emulators, such as {@code LocalGcdHelper} and
-   * {@code LocalResourceManagerHelper}.
+   * A placeholder for credentials to signify that requests sent to the server should not be
+   * authenticated. This is typically useful when using the local service emulators, such as
+   * {@code LocalGcdHelper} and {@code LocalResourceManagerHelper}.
    */
   public static class NoAuthCredentials extends AuthCredentials {
 


gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
```diff
@@ -282,6 +309,13 @@ public static ServiceAccountAuthCredentials createFor(String account, PrivateKey
   }
 
   /**
+   * Creates a placeholder denoting that no credentials should be used.
```
mziccard: Nit: maybe add this is meant to be used for testing against local emulators?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -550,17 +551,21 @@ private LocalResourceManagerHelper() {
   }
 
   /**
-   * Creates a LocalResourceManagerHelper object that listens to requests on the local machine.
+   * Creates a {@code LocalResourceManagerHelper} object that listens to requests on the local
```
mziccard: Nit: links to self in javadoc are not useful, we could use `{@code LocalResourceManagerHelper}` instead.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
```diff
@@ -550,17 +551,21 @@ private LocalResourceManagerHelper() {
...
+   * machine.
    */
   public static LocalResourceManagerHelper create() {
     return new LocalResourceManagerHelper();
   }
 
   /**
-   * Returns a ResourceManagerOptions instance that sets the host to use the mock server.
+   * Returns a {@link ResourceManagerOptions} instance that sets the host to use the mock server.
```
mziccard: Nit: links to self in javadoc are not useful, we could use `{@code LocalResourceManagerHelper}` instead.

gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
```diff
@@ -243,6 +243,33 @@ public GoogleCredentials credentials() {
     }
   }
 
+  public static class NoAuthCredentials extends AuthCredentials {
```
mziccard: Nit and maybe for a separate PR: we could add some javadoc to our `XXXCredentials` classes

mziccard:
Added a few nits, otherwise looks good to me.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: This one is an `@code` already

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
ajkannan: This one is to `ResourceManagerOptions` rather than `LocalResourceManagerHelper`, so keeping the link.

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/testing/LocalResourceManagerHelper.java
mziccard: Sure:)

ajkannan:
Added a commit with more javadoc.  It also specifies that NoAuthCredentials is helpful when using local service emulators.

PR #726, opened 2016-03-08
Remove BlobListOption.recursive option and fix delimiter handling

### LINKED ISSUE #692: Support the delimiter parameter and prefixes result field for bucket listing

mziccard:
This PR fixes #692 and does the following changes:
- Remove `BlobListOption.recursive` option
- Add `BlobListOption.currentDirectory` option that sets the '/' delimiter
- Add `isDirectory` method to `BlobInfo` objects
- Change `StorageRpc.list(bucket)` method to add prefixes to the list of storage objects

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
@@ -14,6 +14,7 @@
 
 package com.google.gcloud.spi;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.gcloud.spi.StorageRpc.Option.DELIMITER;
 import static com.google.gcloud.spi.StorageRpc.Option.FIELDS;
 import static com.google.gcloud.spi.StorageRpc.Option.IF_GENERATION_MATCH;
@@ -58,7 +59,6 @@ import com.google.api.services.storage.model.ComposeRequest.SourceObjects.Object
 import com.google.api.services.storage.model.Objects;
 import com.google.api.services.storage.model.StorageObject;
 import com.google.common.base.Function;
-import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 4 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/spi/DefaultStorageRpc.java
```diff
@@ -163,8 +166,20 @@ public StorageObject create(StorageObject storageObject, final InputStream conte
           .setPageToken(PAGE_TOKEN.getString(options))
           .setFields(FIELDS.getString(options))
           .execute();
-      return Tuple.<String, Iterable<StorageObject>>of(
-          objects.getNextPageToken(), objects.getItems());
+      Iterable<StorageObject> storageObjects = Iterables.concat(
+          objects.getItems() != null ? objects.getItems() : ImmutableList.<StorageObject>of(),
```
aozarov: [`MoreObjects.firstNonNull`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/MoreObjects.html#firstNonNull(T,%20T)) could make it slightly nicer.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/BlobInfo.java
```diff
@@ -589,6 +601,18 @@ public Long updateTime() {
   }
 
   /**
+   * Returns {@code true} if the current blob represents a directory. This can only happen if the
+   * blob is returned by {@link Storage#list(String, Storage.BlobListOption...)} when the
+   * {@link Storage.BlobListOption#currentDirectory()} option is used. If {@code true} only
```
aozarov: I would use `{@code true}` only when it is in a code context (e.g. boolean param, return type,...) but not when "true" is part of the sentence (and could be replaced with e.g. "this is the case").

gcloud-java-storage/src/main/java/com/google/gcloud/storage/BlobInfo.java
```diff
@@ -761,6 +785,9 @@ public Acl apply(ObjectAccessControl objectAccessControl) {
         }
       }));
     }
+    if (storageObject.get("isDirectory") != null) {
```
aozarov: If you don't care about the value you can use `containsKey`, no?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -694,10 +694,17 @@ public static BlobListOption prefix(String prefix) {
     }
 
     /**
-     * Returns an option to specify whether blob listing should include subdirectories or not.
+     * If specified, results are returned in a directory-like mode. Blobs whose names, aside from
```
aozarov: should we replace "aside from" with "after"?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageOptions.java
```diff
@@ -137,7 +110,7 @@ public boolean equals(Object obj) {
       return false;
     }
     StorageOptions other = (StorageOptions) obj;
-    return baseEquals(other) && Objects.equals(pathDelimiter, other.pathDelimiter);
+    return baseEquals(other);
```
aozarov: Alternative, `return obj instanceof StorageOptions && baseEquals((StorageOptions) obj)`

aozarov:
Very nice! LGTM. Feel free to merge after taking care of the comments which you agree with.

Issue #664, opened 2016-02-24
Return an empty collection instead of null.


aozarov:
Some of the the resource metadata properties are collections/maps.

I think currently we are not consistent and in some cases we return `null` if there is no value (e.g.
`BlobInfo` and `BucketInfo`) and in some cases we return an empty collection (e.g. `ProjectInfo`).

My preference would be to never return `null` for a collection as I think the value of being able to
distinct the case of "unassigned" and "assigned and empty" is very small and does not justify
the cost for checking for `null` _and_ empty.

In any case, I think we should be consistent.

/cc @mziccard @ajkannan @mderka

ajkannan:
+1 to never returning null.  Even when users say "get" or "list" and leave out metatdata properties using field options, I think it's reasonable to have empty collections instead of null collections.

mderka:
Agreed to never returning null. 

However, with fields, specification, there is a difference between fields being empty and fields being null. Null means that I did not filter fields. Empty means that I specifically asked for no fields. I believe that services allow this. A reason for asking for no fields is for example to simulate an exist() function. I can want to shoot a request with empty fields just to see if I get 200 response meaning found, or null/404 for not founds.

mziccard:
While I agree that in general that we should be consistent and favor empty lists over `null` I would be careful when it comes to fields selection.
For instance in compute a disk has a list of attached instances. An empty list might mean that the disk is not attached to any instance. If the `attachedInstances` field is not selected through the `DiskOption.fields()` option returning an empty list might be counter-intuitive.

However, one could also argue that now we return `null` for both the "non instances attached" and "field not selected" cases.

mderka:
I would actually prefer `null` to empty list in the scenario of attached disks described by @mziccard as the empty list has special and significant meaning.

aozarov:
@mderka Yes, I was regarding return values and therefore I think this issue should not affect the way we
currently handle field selection (using vararg). Did I get your comment wrong?

@mziccard You are raising an interesting case of an incomplete resource (due to field selection).
Which is similar (though not identical) to when constructing a source[Info] with partial fields.

However, as you mentioned, returning `null` still masks what I consider as a "user-error".
I agree with @mderka that `null` is more intuitive than an empty collection in that case but
it is still a user-error and can lead to undesired result. 

We could, with more work, identify the case of accessing a field that is not populated and throw
IllegalStateException (which we should clearly document). 

Thoughts?

[...]

PR #728, opened 2016-03-08
Add support for Compute's operations

mziccard:
This PR adds support to `gcloud-java-compute` for operations:
- Add `RegionOperationId`, `ZoneOperationId`, `GlobalOperationId` classes
- Add functional methods for operations to `Compute`'s service and rpc classes
- Add `Operation` class (with functional methods)

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
@@ -1092,7 +1092,8 @@ public interface Compute extends Service<ComputeOptions> {
   Page<Operation> listZoneOperations(String zone, OperationListOption... options);
 
   /**
-   * Deletes the requested operation.
+   * Deletes the requested operation. Delete is only possible for operations that have completed
+   * their execution. Any attempt to delete a running operation will fail.
    *
    * @return {@code true} if operation was deleted, {@code false} if it was not found
    * @throws ComputeException upon failure
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
@@ -542,8 +542,10 @@ final class ComputeImpl extends BaseService<ComputeOptions> implements Compute {
                   ZoneOperationId zoneOperationId = (ZoneOperationId) operationId;
                   return computeRpc.getZoneOperation(zoneOperationId.zone(),
                       zoneOperationId.operation(), optionsMap);
-                default:
+                case GLOBAL:
                   return computeRpc.getGlobalOperation(operationId.operation(), optionsMap);
+                default:
+                  throw new IllegalArgumentException("Unexpected operation identity type");
               }
             }
           }, options().retryParams(), EXCEPTION_HANDLER);
@@ -670,8 +672,10 @@ final class ComputeImpl extends BaseService<ComputeOptions> implements Compute {
               ZoneOperationId zoneOperationId = (ZoneOperationId) operation;
               return computeRpc.deleteZoneOperation(zoneOperationId.zone(),
                   zoneOperationId.operation());
-            default:
+            case GLOBAL:
               return computeRpc.deleteGlobalOperation(operation.operation());
+            default:
+              throw new IllegalArgumentException("Unexpected operation identity type");
           }
         }
       }, options().retryParams(), EXCEPTION_HANDLER);
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
@@ -692,7 +692,8 @@ public class Operation implements Serializable {
   }
 
   /**
-   * Deletes this operation.
+   * Deletes this operation. Delete is only possible for operations that have completed their
+   * execution. Any attempt to delete a running operation will fail.
    *
    * @return {@code true} if operation was deleted, {@code false} if it was not found
    * @throws ComputeException upon failure


gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -437,6 +492,68 @@ public static ZoneFilter notEquals(ZoneField field, String value) {
...
+  class OperationFilter extends ListFilter {
+
+    private static final long serialVersionUID = -3202249202748346427L;
+
+    OperationFilter(OperationField field, ComparisonOperator operator, Object value) {
+      super(field.selector(), operator, value);
+    }
+
+    /**
+     * Returns an equality filter for the given field and string value. For string fields,
```
ajkannan: I think the "For string fields, ..." is slightly confusing because we use both "field" and "string" in the previous sentence to specify two different inputs.  Can we change the wording to "The string value is interpreted as..."?  (Also applies for method just below this one)

ajkannan: Ah, actually, it makes sense, I was just misinterpreting this.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -437,6 +492,68 @@ public static ZoneFilter notEquals(ZoneField field, String value) {
...
+
+    /**
+     * Returns an equality filter for the given field and long value.
+     */
+    public static OperationFilter equals(OperationField field, long value) {
+      return new OperationFilter(checkNotNull(field), ComparisonOperator.EQ, value);
+    }
+
+    /**
+     * Returns an inequality filter for the given field and long value.
```
ajkannan: Inequality is probably a valid word here, but it makes me think ">" or "<" rather than "!=".  Would `Returns a "not equals" filter` be too awkward in terms of wording?  (I'm blanking on a better term.)  Applies here and for other notEquals methods.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -437,6 +492,68 @@ public static ZoneFilter notEquals(ZoneField field, String value) {
...
+     * match the entire field.
+     *
+     * @see <a href="https://github.com/google/re2">RE2</a>
+     */
+    public static OperationFilter equals(OperationField field, String value) {
+      return new OperationFilter(checkNotNull(field), ComparisonOperator.EQ, checkNotNull(value));
+    }
+
+    /**
+     * Returns an equality filter for the given field and string value. For string fields,
```
ajkannan: Copy paste error on "equality" (since it's a `notEquals` method).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -793,6 +910,73 @@ public static LicenseOption fields(LicenseField... fields) {
...
+
+    private static final long serialVersionUID = -4572636917684779912L;
+
+    private OperationOption(ComputeRpc.Option option, Object value) {
+      super(option, value);
+    }
+
+    /**
+     * Returns an option to specify the operation's fields to be returned by the RPC call. If this
+     * option is not provided all operation's fields are returned. {@code OperationOption.fields}
```
ajkannan: Nit, add comma between "not provided" and "all".

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -793,6 +910,73 @@ public static LicenseOption fields(LicenseField... fields) {
...
+
+    /**
+     * Returns an option to specify a filter to the operations being listed.
+     */
+    public static OperationListOption filter(OperationFilter filter) {
+      return new OperationListOption(ComputeRpc.Option.FILTER, filter.toPb());
+    }
+
+    /**
+     * Returns an option to specify the maximum number of operations to be returned.
```
ajkannan: Should we add "per page" at the end of this sentence?  Also, because the [docs](https://cloud.google.com/compute/docs/reference/latest/globalOperations/list#maxResults) suggest that it's max results per page, should we change the name of this option to "pageSize"?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -220,6 +220,61 @@ static String selector(LicenseField... fields) {
...
+    ID("id"),
+    INSERT_TIME("insertTime"),
+    NAME("name"),
+    OPERATION_TYPE("operationType"),
+    PROGRESS("progress"),
+    SELF_LINK("selfLink"),
+    START_TIME("startTime"),
+    STATUS("status"),
+    STATUS_MESSAGE("statusMessage"),
+    REGION("region"),
```
ajkannan: Nit: "region" goes before "self_link" (I only point it out because you it looks like you were alphabetizing this ginormous list of fields)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -220,6 +220,61 @@ static String selector(LicenseField... fields) {
...
+   * @see <a
+   *     href="https://cloud.google.com/compute/docs/reference/latest/globalOperations#resource">
+   *     GlobalOperation Resource</a>
+   * @see <a
+   *     href="https://cloud.google.com/compute/docs/reference/latest/regionOperations#resource">
+   *     RegionOperation Resource</a>
+   * @see <a href="https://cloud.google.com/compute/docs/reference/latest/zoneOperations#resource">
+   *     ZoneOperation Resource</a>
+   */
+  enum OperationField {
```
ajkannan: Is there a reason for leaving out "zone" from this enum?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -793,6 +910,73 @@ public static LicenseOption fields(LicenseField... fields) {
...
+    /**
+     * Returns an option to specify the page token from which to start listing operations.
+     */
+    public static OperationListOption startPageToken(String pageToken) {
+      return new OperationListOption(ComputeRpc.Option.PAGE_TOKEN, pageToken);
+    }
+
+    /**
+     * Returns an option to specify the operation's fields to be returned by the RPC call. If this
+     * option is not provided all operation's fields are returned.
```
ajkannan: Nit: ", " between "provided" and "all"

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
```diff
@@ -153,6 +172,48 @@
...
+
+    @Override
+    public Page<Operation> nextPage() {
+      return listRegionOperations(region, serviceOptions, requestOptions);
+    }
+  }
+
+  private static class ZoneOperationPageFetcher implements NextPageFetcher<Operation> {
+
+    private static final long serialVersionUID = 4111705358926164078L;
```
ajkannan: This should be regenerated (looks like a copy of above).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/GlobalOperationId.java
```diff
@@ -0,0 +1,107 @@
...
+   */
+  public static GlobalOperationId of(String project, String operation) {
+    return new GlobalOperationId(project, operation);
+  }
+
+  /**
+   * Returns {@code true} if the provided string matches the expected format of a global operation
+   * URL. Returns {@code false} otherwise.
+   */
+  static boolean matchesUrl(String url) {
```
ajkannan: Should this be public?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+import java.io.ObjectInputStream;
+import java.io.Serializable;
+import java.math.BigInteger;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Google Compute Engine operations. Operation identity can be obtained via {@link #operationId()}.
+ * For global operations {@link #operationId()} returns an {@link GlobalOperationId}, for region
```
ajkannan: I think this list would sound slightly better as:
`{@link #operationId()} returns {@link GlobalOperationId}s for global operations, {@link RegionOperationId}s for region operations, and {@ZoneOperationId}s for zone operations.`

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+  private final Long startTime;
+  private final Long endTime;
+  private final List<OperationError> errors;
+  private final List<OperationWarning> warnings;
+  private final Integer httpErrorStatusCode;
+  private final String httpErrorMessage;
+  private final String description;
+
+  /**
+   * Types of operations.
```
ajkannan: "Types of operation statuses"?  Or just "Operation statuses"? ("Types of operations" sounds closer to global vs zone vs region)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+        new Function<OperationError,
+            com.google.api.services.compute.model.Operation.Error.Errors>() {
+          @Override
+          public com.google.api.services.compute.model.Operation.Error.Errors apply(
+              OperationError operation) {
+            return operation.toPb();
+          }
+        };
+
+    private static final long serialVersionUID = -1155314394806515873L;
```
ajkannan: Nit: we usually move this and instance variables up to the top of the class.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+    private final String message;
+
+    OperationError(String code, String location, String message) {
+      this.code = code;
+      this.location = location;
+      this.message = message;
+    }
+
+    /**
+     * Returns an error type identifier for this error.
```
ajkannan: Do you know if there's a docs url that has a mapping from operation error codes to error type?  If there is, it'd be good to include it in the javadoc as an `@see`

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+        new Function<OperationWarning,
+            com.google.api.services.compute.model.Operation.Warnings>() {
+          @Override
+          public com.google.api.services.compute.model.Operation.Warnings apply(
+              OperationWarning operation) {
+            return operation.toPb();
+          }
+        };
+
+    private static final long serialVersionUID = 4917326627380228928L;
```
ajkannan: Nit: move to top of class

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+        FROM_PB_FUNCTION =
+        new Function<com.google.api.services.compute.model.Operation.Warnings,
+            OperationWarning>() {
+          @Override
+          public OperationWarning apply(
+              com.google.api.services.compute.model.Operation.Warnings pb) {
+            return OperationWarning.fromPb(pb);
+          }
+        };
+    static final
```
ajkannan: Nit: add space between methods

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+
+    /**
+     * Returns a human-readable error message.
+     */
+    public String message() {
+      return message;
+    }
+
+    /**
+     * Returns metadata about this warning.
```
ajkannan: Do you know if there's any docs/consistency on what kind of metadata is provided?  If so, it'd be good to add to the javadoc.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+    private String user;
+    private Integer progress;
+    private Long insertTime;
+    private Long startTime;
+    private Long endTime;
+    private List<OperationError> errors;
+    private List<OperationWarning> warnings;
+    private Integer httpErrorStatusCode;
+    private String httpErrorMessage;
+
```
ajkannan: Nit: extraneous line

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+      return this;
+    }
+
+    Builder description(String description) {
+      this.description = description;
+      return this;
+    }
+
+    /**
+     * Creates an object.
```
ajkannan: s/object/operation/

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+  /**
+   * Returns the creation timestamp in milliseconds since epoch.
+   */
+  public Long creationTimestamp() {
+    return creationTimestamp;
+  }
+
+  /**
+   * Returns the operation's identity. This method returns an {@link GlobalOperationId} for global
+   * operations, returns a {@link RegionOperationId} for region operations and returns a
```
ajkannan: Nit: can remove "returns" from the 2nd and 3rd items of this list.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+   *
+   * @see <a href="https://www.ietf.org/rfc/rfc1035.txt">RFC1035</a>
+   */
+  @SuppressWarnings("unchecked")
+  public <T extends OperationId> T operationId() {
+    return (T) operationId;
+  }
+
+  /**
+   * Reserved for future use.
```
ajkannan: The public docs don't say much about this at all.  Do you know what they mean when they say "reserved for future use?"  Should we make this package scope until it's purpose is revealed?

ajkannan: Ah, it is package scope, nevermind that comment.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/OperationId.java
```diff
@@ -0,0 +1,38 @@
...
+   */
+  String project();
+
+  /**
+   * Returns the name of the operation resource.
+   */
+  String operation();
+
+  /**
+   * Returns a fully qualified URL to the entity.
```
ajkannan: Nit: I think it would be slightly clearer to say "operation" rather than "entity"

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+   * @throws ComputeException upon failure
+   */
+  public boolean exists() throws ComputeException {
+    return reload(Compute.OperationOption.fields()) != null;
+  }
+
+  /**
+   * Checks if this operation has completed its execution, either failing or succeeding. If the
+   * operation does not exist this method returns {@code false}. To correctly wait for operation's
+   * completion check that the operation exists first, using {@link #exists()}:
```
ajkannan: Nit: add comma between completion and check and remove comma before using

ajkannan:
Hey @mziccard, just a heads up that I haven't finished my pass yet (I'll start back up tomorrow).  I'll probably be focusing more on javadocs/tests and @aozarov will be looking at the rest.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
mziccard: This option has the same behavior across all services (at least also for storage and bigquery): the maximum number is per page. I think `maxResults` is ok, but if you feel strongly about this better open an issue.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/GlobalOperationId.java
mziccard: It will be in the future

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
mziccard: They're preceded by static final variables so I think they should stay where they are. We do this also elsewhere where `TO_PB_FUNCTION` and `FROM_PB_FUNCTION` precede instance fields.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
mziccard: Couldn't find any

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
mziccard: See my comment above

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
mziccard: Couldn't find any

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
ajkannan: Interesting, if the behavior is the same (max results per page rather than max results overall), I feel like the cloud docs are slightly inconsistent. (While those docs say "per page", storage's and bigquery's list docs don't say "per page".)  I guess all those list functions return a page token to get more results?

I used pageSize in Resource Manager rather than maxResults (in part because the REST API called it pageSize rather than maxResults).  I prefer page size over max results, but at the very least we should standardize terminology in our library, so I'll create an issue.

mziccard: Yes, when you set `maxResults(X)` you get a page of `X` items plus the next page token.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationTest.java
```diff
@@ -0,0 +1,489 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.gcloud.compute.Operation.OperationError;
+import com.google.gcloud.compute.Operation.Status;
+import com.google.gcloud.compute.Operation.OperationWarning;
```
ajkannan: lint error: out of order

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationTest.java
```diff
@@ -0,0 +1,489 @@
...
+
+import org.junit.After;
+import org.junit.Test;
+
+import java.math.BigInteger;
```
ajkannan: lint error: unused

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationIdTest.java
```diff
@@ -0,0 +1,143 @@
...
+    assertSame(regionOperationId, regionOperationId.setProjectId(PROJECT));
+    compareRegionOperationId(regionOperationId,
+        RegionOperationId.of(REGION, NAME).setProjectId(PROJECT));
+  }
+
+  @Test
+  public void testMatchesUrl() {
+    assertTrue(GlobalOperationId.matchesUrl(GlobalOperationId.of(PROJECT, NAME).selfLink()));
+    assertFalse(GlobalOperationId.matchesUrl("notMatchingUrl"));
+    assertTrue(RegionOperationId.matchesUrl(RegionOperationId.of(PROJECT, REGION, NAME).selfLink()));
```
ajkannan: lint error: >100 characters

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+
+  /**
+   * Returns the time that this operation was completed. In milliseconds since epoch.
+   */
+  public Long endTime() {
+    return endTime;
+  }
+
+  /**
+   * Returns the errors encountered while processing this operation, if any. Returns {@code null} if
```
ajkannan: Related to #664, should this ever return null?  Would it be better if it returned an empty list?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/GlobalOperationId.java
ajkannan: Just for my own clarification, does this becoming public depend on something changing in the underlying rest api?  Or are we waiting on making it public in the future.  (Do you think this merits an issue to track it?)

mziccard: This is never going to be public. This is going to be used inside the package, namely in `ResourceId.fromUrl` that I will implement as soon as all identity classes are available.

gcloud-java-compute/src/main/java/com/google/gcloud/spi/ComputeRpc.java
```diff
@@ -161,4 +162,70 @@ public Y y() {
...
+
+  /**
+   * Returns the requested zone operation or {@code null} if not found.
+   *
+   * @throws ComputeException upon failure
+   */
+  Operation getZoneOperation(String zone, String operation, Map<Option, ?> options);
+
+  /**
+   * Lists the zone operations for the current project and zone.
```
ajkannan: Nit: would be slightly clearer if we replace "zone" with "the provided zone".

gcloud-java-compute/src/test/java/com/google/gcloud/compute/it/ITComputeTest.java
```diff
@@ -447,4 +446,190 @@ public void testListZonesWithFilter() {
     assertEquals(ZONE, zoneIterator.next().zoneId().zone());
     assertFalse(zoneIterator.hasNext());
   }
+
+  @Test
+  public void testListGlobalOperations() {
+    Page<Operation> operationPage = compute.listGlobalOperations();
+    Iterator<Operation> operationIterator = operationPage.iterateAll();
+    while(operationIterator.hasNext()) {
```
ajkannan: lint error: replace "while(" with "while (" here and below.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeImplTest.java
```diff
@@ -117,6 +119,37 @@
   private static final LicenseId LICENSE_ID = LicenseId.of("project", "license");
   private static final Boolean CHARGES_USE_FEE = true;
   private static final License LICENSE = new License(LICENSE_ID, CHARGES_USE_FEE);
+  private static final Operation.OperationError OPERATION_ERROR1 =
```
ajkannan: Nit: could we import Operation.Operation[Error|Warning] to avoid having to specify "Operation."?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RegionOperationId.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(baseHashCode(), operation);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof RegionOperationId && baseEquals((RegionOperationId) obj);
```
ajkannan: Does this need to check that the "operation" fields are equal (like in ZoneOperationId and GlobalOperationId)?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RegionOperationId.java
```diff
@@ -0,0 +1,112 @@
...
+      return this;
+    }
+    return RegionOperationId.of(projectId, region(), operation);
+  }
+
+  /**
+   * Returns a region operation identity given the region identity and the operation name.
+   */
+  public static RegionOperationId of(RegionId regionId, String operation) {
+    return new RegionOperationId(regionId.project(), regionId.region(), operation);
```
ajkannan: This "of" method isn't tested in OperationIdTest.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ZoneOperationId.java
```diff
@@ -0,0 +1,114 @@
...
+      return this;
+    }
+    return ZoneOperationId.of(projectId, zone(), operation);
+  }
+
+  /**
+   * Returns a zone operation identity given the zone identity and the operation name.
+   */
+  public static ZoneOperationId of(ZoneId zoneId, String operation) {
+    return new ZoneOperationId(zoneId.project(), zoneId.zone(), operation);
```
ajkannan: Nit: this "of" method isn't tested in OperationIdTest.testOf().

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationIdTest.java
```diff
@@ -0,0 +1,143 @@
...
+    assertEquals(REGION, regionOperationId.region());
+    assertEquals(NAME, regionOperationId.operation());
+  }
+
+  @Test
+  public void testToAndFromUrl() {
+    GlobalOperationId operationId = GlobalOperationId.of(PROJECT, NAME);
+    compareOperationId(operationId, GlobalOperationId.fromUrl(operationId.selfLink()));
+    thrown.expect(IllegalArgumentException.class);
+    thrown.expectMessage("notMatchingUrl is not a valid global operation URL");
```
ajkannan: I'm fairly sure that once an exception is hit, this method doesn't execute any other code below.  Perhaps we should split this method up into three, one per type of identity?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
ajkannan: If we decide to change this, we should also change warnings and OperationWarning's metadata.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/it/ITComputeTest.java
```diff
@@ -447,4 +446,190 @@ public void testListZonesWithFilter() {
...
+  @Test
+  public void testListZoneOperations() {
+    Page<Operation> operationPage = compute.listZoneOperations(ZONE);
+    Iterator<Operation> operationIterator = operationPage.iterateAll();
+    while(operationIterator.hasNext()) {
+      Operation operation = operationIterator.next();
+      assertNotNull(operation.id());
+      assertNotNull(operation.operationId());
+      assertEquals(ZONE, operation.<ZoneOperationId>operationId().zone());
+      // assertNotNull(operation.creationTimestamp());
```
ajkannan: Nit: could also add todo here, just for consistency

gcloud-java-compute/src/test/java/com/google/gcloud/compute/it/ITComputeTest.java
```diff
@@ -447,4 +446,190 @@ public void testListZonesWithFilter() {
     assertEquals(ZONE, zoneIterator.next().zoneId().zone());
     assertFalse(zoneIterator.hasNext());
   }
+
+  @Test
+  public void testListGlobalOperations() {
```
ajkannan: Is the reason there aren't integration tests for getting/deleting operations because it's unreliable to create and then get/delete an operation?  Or do you think we can test this with a timeout in a fashion that won't be too flaky?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -793,6 +910,73 @@ public static LicenseOption fields(LicenseField... fields) {
   }
 
   /**
+   * Class for specifying operation get options.
+   */
+  class OperationOption extends Option {
```
ajkannan: Just curious: this naming is consistent within Compute and also with BigQuery, but in storage we include "get" (eg BlobGetOption).  Is the rationale for this difference that we want shorter class names (so remove "get" where possible), but because there are many types of Blob Options (BlobTargetOption, BlobWriteOption, etc), it's clearer to leave "get" in the class name for storage?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,800 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(code, message, metadata);
+    }
+  }
+
+  /**
+   * Builder for Compute Engine operations.
```
ajkannan: Nit: since this builder is package scope, we can take out this javadoc.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
ajkannan: Also if this builder is supposed to be package scope, we can probably remove this javadoc and make build() package scope, right?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeImplTest.java
```diff
@@ -216,11 +249,33 @@
   private static final Compute.LicenseOption LICENSE_OPTION_FIELDS =
       Compute.LicenseOption.fields(Compute.LicenseField.CHARGES_USE_FEE);
 
+  // Operation options
+  private static final Compute.OperationOption OPERATION_OPTION_FIELDS =
```
ajkannan: Nit: maybe it won't make a difference on this line, but in general if we import Compute.*Option and Operation., we may be able to avoid some line wrapping in this file.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeImplTest.java
```diff
@@ -695,20 +832,355 @@ public void testGetLicenseFromId() {
...
+        .andReturn(regionOperation.toPb());
+    EasyMock.replay(computeRpcMock);
+    compute = options.service();
+    Operation operation = compute.get(REGION_OPERATION_ID, OPERATION_OPTION_FIELDS);
+    String selector = (String) capturedOptions.getValue().get(OPERATION_OPTION_FIELDS.rpcOption());
+    assertTrue(selector.contains("selfLink"));
+    assertTrue(selector.contains("id"));
+    assertTrue(selector.contains("description"));
+    assertEquals(23, selector.length());
+    assertEquals(regionOperation, operation);
```
ajkannan: Seems like a slightly weird thing to assert, considering the actual behavior from the service will be to return a RegionOperation with only those three fields populated.  (Also applies for the analogous test for Global and Zone)

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationTest.java
```diff
@@ -0,0 +1,489 @@
...
+
+  @After
+  public void tearDown() throws Exception {
+    verify(serviceMockReturnsOptions);
+  }
+
+  @Test
+  public void testBuilder() {
+    initializeExpectedOperation(6);
+    assertEquals(CREATION_TIMESTAMP, globalOperation.creationTimestamp());
```
ajkannan: I wonder if it's worth factoring any assertEquals common to all three operations into another function to reduce the size of this function.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationTest.java
```diff
@@ -0,0 +1,489 @@
...
+  @Test
+  public void testIsDone_NotExists() throws Exception {
+    initializeExpectedOperation(3);
+    Compute.OperationOption[] expectedOptions =
+        {Compute.OperationOption.fields(Compute.OperationField.STATUS)};
+    expect(compute.options()).andReturn(mockOptions);
+    expect(compute.get(GLOBAL_OPERATION_ID, expectedOptions)).andReturn(globalOperation);
+    replay(compute);
+    initializeOperation();
+    assertTrue(operation.isDone());
```
ajkannan: Should this be false?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationTest.java
```diff
@@ -0,0 +1,489 @@
...
+  @Test
+  public void testReloadWithOptions() throws Exception {
+    initializeExpectedOperation(5);
+    expect(compute.options()).andReturn(mockOptions);
+    expect(compute.get(GLOBAL_OPERATION_ID, Compute.OperationOption.fields()))
+        .andReturn(globalOperation);
+    replay(compute);
+    initializeOperation();
+    Operation updatedOperation = operation.reload(Compute.OperationOption.fields());
+    compareOperation(globalOperation, updatedOperation);
```
ajkannan: Similar to above, slightly weird to assert this considering the service would return a globalOperation with most fields set to null.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/it/ITComputeTest.java
ajkannan: Also, has gcloud-devel been populated with some operations just so these listing functions run at least one iteration of the loops?

ajkannan:
Done with my first pass at comments and tests.  Nice job chipping away at this huge API!

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
mziccard: Did we decide how to proceed with respect to #664? Should `errors` return an empty list even if the user is not selecting the `OperationField.ERROR` field? This returns `null` (as we do in most of the rest of the library) to avoid taking actions for #664 in PRs that are not directly related.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeImplTest.java
```diff
@@ -695,20 +832,355 @@ public void testGetLicenseFromId() {
...
+            new Function<Operation, com.google.api.services.compute.model.Operation>() {
+              @Override
+              public com.google.api.services.compute.model.Operation apply(Operation operation) {
+                return operation.toPb();
+              }
+            }));
+    EasyMock.expect(computeRpcMock.listGlobalOperations(EMPTY_RPC_OPTIONS)).andReturn(result);
+    EasyMock.replay(computeRpcMock);
+    Page<Operation> page = compute.listGlobalOperations();
+    assertEquals(cursor, page.nextPageCursor());
```
mziccard: This is unit tests for the service, testing next page is out of scope here and won't help to understand if cursor field and other options have been set correctly. This is tested in integration tests, where `iterateAll` is used.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ZoneOperationId.java
mziccard: It's tested [here](https://github.com/mziccard/gcloud-java/blob/compute/gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationIdTest.java#L59).

gcloud-java-compute/src/test/java/com/google/gcloud/compute/it/ITComputeTest.java
mziccard: Getting/deleting operation tests will arrive as soon as we have other Compute resources whose creation/deletion starts an operation (we will have to poll that operation to know when the resource is available). `gcloud-devel` already has plenty of operations.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
mziccard: Almost all Compute APIs (but gets) are async and return an operation. Field selection for operations can be provided in all such APIs, calling the option `OperationOption` sounded better to me given that it must be passed to methods that have nothing to do with `getOperation`.

And you are right in storage we have so many blob-related options that it's worth clarifying.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeImplTest.java
mziccard: The purpose of this is to assert that the object passed by the spi layer is correctly wrapped and passed on. It might be weird but this does the job. Creating a partial operation object adds nothing to the test and given that the file is already 1.2k lines long (and we tested only 1/10 of Compute resources) I would tend to avoid creating other objects if not strictly needed.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationTest.java
mziccard: See my comment above

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RegionOperationId.java
mziccard: Nice catch!

gcloud-java-compute/src/test/java/com/google/gcloud/compute/OperationTest.java
mziccard: Yes, in fact the method should return `null`, not `globalOperation`. Copy-paste error, nice catch!

mziccard:
Comments addressed in 912ec8b

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
ajkannan: Feel free to weigh in on #740, but I think so far the consensus is to use "pageSize" and standardize javadoc.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
ajkannan: Looking at #664 and talking to @aozarov, I think the consensus is that we can't return an empty collection when the user leaves out that field.  So fine as is, no change needed.

ajkannan:
Did a quick second pass (on comments and tests), and looks good.  Three outstanding things:

* #740 seems to be leaning towards using pageSize rather than maxResults.  If you agree, then that still needs to be changed
* One more grammar nit
* On paging, the integration test doesn't specify a page size, we may not actually be testing paging.  In addition, we can't really assert anything about the operations we're getting in the integration test, so I think we could conceptually call nextPage() in the ComputeImplTest list functions and just assert that the next request is populated with the page cursor we want.  Does that sound reasonable, or do you think that's still out of the scope of ComputeImplTest?

mziccard:
Addressed all comments in ef98269

gcloud-java-compute/src/test/java/com/google/gcloud/compute/it/ITComputeTest.java
```diff
@@ -29,8 +29,11 @@
...
 import com.google.gcloud.compute.LicenseId;
 import com.google.gcloud.compute.MachineType;
+import com.google.gcloud.compute.Operation;
 import com.google.gcloud.compute.Region;
+import com.google.gcloud.compute.RegionOperationId;
 import com.google.gcloud.compute.Zone;
+import com.google.gcloud.compute.ZoneOperationId;
 import com.google.gcloud.compute.testing.RemoteComputeHelper;
 
 import org.junit.BeforeClass;
```
aozarov: Unrelated. "throws InterruptedException" is not needed for `beforeClass`

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
```diff
@@ -248,11 +250,130 @@ public License getLicense(String project, String license, Map<Option, ?> options
...
+  }
+
   /**
    * This method returns {@code null} if the error code of {@code exception} was 404, re-throws the
    * exception otherwise.
    *
-   * @throws ComputeException if the error code of {@code exception} was not 404.
+   * @throws ComputeException if the error code of {@code exception} was not 404
    */
   private static <T> T nullForNotFound(IOException exception) {
```
aozarov: Unnecessary cast for `null`

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -237,7 +293,7 @@ static String selector(LicenseField... fields) {
       EQ,
 
       /**
-       * Defines an inequality filter.
+       * Defines a not-equals filter.
```
aozarov: Nit, with this change I think chaining the EQ javadoc to  "Defines an equals filter" would sound more consistent but fine as is.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -391,7 +447,7 @@ public static RegionFilter equals(RegionField field, String value) {
     }
 
     /**
-     * Returns an equality filter for the given field and string value. For string fields,
+     * Returns a not-equals filter for the given field and string value. For string fields,
      * {@code value} is interpreted as a regular expression using RE2 syntax. {@code value} must
```
aozarov: Fine as is. linking RE2 to https://github.com/google/re2/wiki/Syntax would be nice.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -437,6 +493,68 @@ public static ZoneFilter notEquals(ZoneField field, String value) {
...
+    OperationFilter(OperationField field, ComparisonOperator operator, Object value) {
+      super(field.selector(), operator, value);
+    }
+
+    /**
+     * Returns an equality filter for the given field and string value. For string fields,
+     * {@code value} is interpreted as a regular expression using RE2 syntax. {@code value} must
+     * match the entire field.
+     *
+     * @see <a href="https://github.com/google/re2">RE2</a>
```
aozarov: I think https://github.com/google/re2/wiki/Syntax is slightly more useful than https://github.com/google/re2 in this context.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -437,6 +493,68 @@ public static ZoneFilter notEquals(ZoneField field, String value) {
...
+     * Returns a not-equals filter for the given field and long value.
+     */
+    public static OperationFilter notEquals(OperationField field, long value) {
+      return new OperationFilter(checkNotNull(field), ComparisonOperator.NE, value);
+    }
+
+    /**
+     * Returns an equality filter for the given field and integer value.
+     */
+    public static OperationFilter equals(OperationField field, int value) {
```
aozarov: What is the value of providing an `int` settings when we have `long` and doc suggests another accepted type of `boolean` "only atomic field types are supported (string, number, boolean)", I assume you don't have it here
as Operation does not have boolean field but you are considering it for places that do have boolean fields (if any)?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -479,8 +597,8 @@ public static DiskTypeListOption filter(DiskTypeFilter filter) {
     /**
      * Returns an option to specify the maximum number of disk types to be returned.
```
aozarov: We should probably say "per page".

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
mziccard: Yes I plan to have a boolean `equals` for places where boolean fields exists. I added the integer `equals` just because an operation has some integer (not long) fields but you are right that it's not strictly necessary, if you prefer I can remove this.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -793,6 +911,73 @@ public static LicenseOption fields(LicenseField... fields) {
...
+
+    /**
+     * Returns an option to specify the operation's fields to be returned by the RPC call. If this
+     * option is not provided, all operation's fields are returned.
+     * {@code OperationListOption.fields} can be used to specify only the fields of interest.
+     * {@link Operation#operationId()} is always returned, even if not specified.
+     */
+    public static OperationListOption fields(OperationField... fields) {
+      StringBuilder builder = new StringBuilder();
+      builder.append("items(").append(OperationField.selector(fields)).append("),nextPageToken");
```
aozarov: cross referencings #456 to suggests that the shared selector method that would be used in this case could get the StingBuilder as the first argument and append to it.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -889,4 +1074,40 @@ public static LicenseOption fields(LicenseField... fields) {
...
+
+  /**
+   * Lists the global operations.
+   *
+   * @throws ComputeException upon failure
+   */
+  Page<Operation> listGlobalOperations(OperationListOption... options);
+
+  /**
+   * Lists the operations in the provided region.
```
aozarov: We should update Javadoc to make it clear that it is not a containment based listing but rather type based listing.
Hence, this is a filter for all the regional operations for the specific region rather than all the operations for that region (including all its zones).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
```diff
@@ -464,6 +525,159 @@ public License getLicense(LicenseId license, LicenseOption... options) {
...
 
+  @Override
+  public Operation get(final OperationId operationId, OperationOption... options) {
+    final Map<ComputeRpc.Option, ?> optionsMap = optionMap(options);
+    try {
+      com.google.api.services.compute.model.Operation answer =
+          runWithRetries(new Callable<com.google.api.services.compute.model.Operation>() {
+            @Override
+            public com.google.api.services.compute.model.Operation call() {
+              if (operationId instanceof RegionOperationId) {
```
aozarov: replace to switch when adding `type` to operationId. Also, consider making `OperationId` a base class instead of an interface with a package scope constructor to avoid getting "unrecognized" instances.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
aozarov: I think we should create an issue for it and discuss that with the service owners. A code without reference is not that helpful (as compared to message).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,793 @@
...
+
+    /**
+     * Returns an error type identifier for this error.
+     */
+    public String code() {
+      return code;
+    }
+
+    /**
+     * Returns the field in the request which caused the error. Might be {@code null}.
```
aozarov: Nit, Any reason for using "Might be {@code null}" instead of "an optional" like you do bellow?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
aozarov: Maybe at least we can provide the same sample that is provided by public doc (e.g.
"A key that provides more detail on the warning being returned. For example, for warnings where there are no results in a list request for a particular zone, this key might be scope and the key value might be the zone name. Other examples might be a key indicating a deprecated resource, and a suggested replacement, or a warning about invalid network settings (for example, if an instance attempts to perform IP forwarding but is not enabled for IP forwarding)").

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,793 @@
...
+import java.util.Objects;
+
+/**
+ * Google Compute Engine operations. Operation identity can be obtained via {@link #operationId()}.
+ * {@link #operationId()} returns {@link GlobalOperationId} for global operations,
+ * {@link RegionOperationId} for region operations, and {@link ZoneOperationId} for zone operations.
+ * To get an {@code Operation} object with the most recent information, use
+ * {@link #reload(Compute.OperationOption...)}.
+ */
+public final class Operation implements Serializable {
```
aozarov: Per the issue we had with `Storage`. if our API returns a type X we should provide a way for creating an instance of that type (when applicable) or not have it as final (so it will be easier to mock it). We should still make the constructor package scope (and make some of the methods final - such as hashCode/equals).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,793 @@
...
+    return user;
+  }
+
+  /**
+   * Returns an optional progress indicator that ranges from 0 to 100. There is no requirement that
+   * this be linear or support any granularity of operations. This should not be used to guess when
+   * the operation will be complete. This number should monotonically increase as the operation
+   * progresses.
+   */
+  public Integer progress() {
```
aozarov: I wonder why they have status when they have this (which seems to be more detailed than status).

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,793 @@
...
+  }
+
+  /**
+   * Checks if this operation exists.
+   *
+   * @return {@code true} if this operation exists, {@code false} otherwise
+   * @throws ComputeException upon failure
+   */
+  public boolean exists() throws ComputeException {
+    return reload(Compute.OperationOption.fields()) != null;
```
aozarov: Does delete operation cancels that operation?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
```diff
@@ -0,0 +1,793 @@
...
+
+  /**
+   * Returns the time that this operation was requested. In milliseconds since epoch.
+   */
+  public Long insertTime() {
+    return insertTime;
+  }
+
+  /**
+   * Returns the time that this operation was started by the service. In milliseconds since epoch.
```
aozarov: should we make it explicit stating that it would be null if not started and bellow it would be null if not ended?

aozarov:
Nice work. Done with first iteration.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
mziccard: No, if you try to delete a PENDING operation you get error 400 "the resource is not ready"

mziccard:
Comments fixed in c41ad14 and 484071c

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
aozarov: because there is no enforcement (one could use the `int` version for a `long` property and vise versa) I think having only the `long` version is preferred.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
```diff
@@ -464,6 +525,161 @@ public License getLicense(LicenseId license, LicenseOption... options) {
...
+              switch (operationId.type()) {
+                case REGION:
+                  RegionOperationId regionOperationId = (RegionOperationId) operationId;
+                  return computeRpc.getRegionOperation(regionOperationId.region(),
+                      regionOperationId.operation(), optionsMap);
+                case ZONE:
+                  ZoneOperationId zoneOperationId = (ZoneOperationId) operationId;
+                  return computeRpc.getZoneOperation(zoneOperationId.zone(),
+                      zoneOperationId.operation(), optionsMap);
+                default:
```
aozarov: Fine as is though being stricter by adding a case for `GLOBAL` and throwing IllegalArgumentException in default would make it easier to find bugs if/when more types are added to the enum and this method was not changed.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Operation.java
aozarov: Ah, good to know. Do you think it is worth mentioning in the `delete(OperationId)` javadoc?

aozarov:
Looks great! Feel free to merge without or without my latest nits.

PR #736, opened 2016-03-10
ZoneInfo.of() instead of builder

### LINKED ISSUE #698: What parameters should Zone/ZoneInfo builder require
There are 3 things mandatory for Zone: name, dnsName and description. In my opinion, it either makes sense to allow for creating zone with name being mandatory and the rest to be part of the builder, or require all three in `ZoneInfo.builder(String name, String dnsName, String description)`. I am inclined to the first option. Should we ask for name only, or for all three?

/cc @ajkannan @aozarov @mziccard

mderka:
Fixes #698.

### DIFF
--- a/gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneInfoTest.java
+++ b/gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneInfoTest.java
@@ -49,15 +49,15 @@ public class ZoneInfoTest {
       .build();
 
   @Test
-  public void testDefaultBuilders() {
-    ZoneInfo zone = ZoneInfo.of(NAME, DNS_NAME, DESCRIPTION);
-    assertTrue(zone.nameServers().isEmpty());
-    assertEquals(NAME, zone.name());
-    assertNull(zone.id());
-    assertNull(zone.creationTimeMillis());
-    assertNull(zone.nameServerSet());
-    assertEquals(DESCRIPTION, zone.description());
-    assertEquals(DNS_NAME, zone.dnsName());
+  public void testOf() {
+    ZoneInfo partial = ZoneInfo.of(NAME, DNS_NAME, DESCRIPTION);
+    assertTrue(partial.nameServers().isEmpty());
+    assertEquals(NAME, partial.name());
+    assertNull(partial.id());
+    assertNull(partial.creationTimeMillis());
+    assertNull(partial.nameServerSet());
+    assertEquals(DESCRIPTION, partial.description());
+    assertEquals(DNS_NAME, partial.dnsName());
   }
 
   @Test


gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
```diff
@@ -185,10 +185,10 @@ public ZoneInfo build() {
   }
 
   /**
-   * Returns a builder for {@code ZoneInfo} with an assigned {@code name}.
+   * Returns a ZoneInfo object with assigned {@code name}, {@code dnsName} and {@code description}.
    */
-  public static Builder builder(String name) {
-    return new BuilderImpl(name);
+  public static ZoneInfo of(String name, String dnsName, String description) {
```
mziccard: `description` is optional right? What if I don't want to provide one?
If you confirm description is optional (also what about `dnsName`) you could consider overloading `of` with:
```java
public static ZoneInfo of(String name, String dnsName) {
  return new BuilderImpl(name).dnsName(dnsName).build();
}
```

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneInfoTest.java
```diff
@@ -41,25 +41,23 @@
...
-      .dnsName(DNS_NAME)
-      .description(DESCRIPTION)
       .nameServerSet(NAME_SERVER_SET)
       .nameServers(NAME_SERVERS)
       .build();
 
   @Test
   public void testDefaultBuilders() {
-    ZoneInfo zone = ZoneInfo.builder(NAME).build();
+    ZoneInfo zone = ZoneInfo.of(NAME, DNS_NAME, DESCRIPTION);
```
mziccard: This method should probably be renamed to `testOf`, also why don't you use `INFO` for this test? Is it necessary to create a new `ZoneInfo` object?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
mziccard: I checked and they're all mandatory. Weird that `description` is a mandatory field. Sorry, forget this comment.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneInfoTest.java
mderka: I renamed the test.

I am using a different object for checking that the read-only fields `id`, `creationTimeMillis` and `nameServerSet` are not initialized. `INFO` has them initialized to some values. I renamed `zone` to `partial` so that this is more obvious.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
mderka: Acknowledged. They are mandatory.

mderka:
Done with revision.

mziccard:
Looks good to me, merging!

Issue #733, opened 2016-03-09
Datastore Javadoc includes example programs and more...


lesv:
It's nice to have around, but a clean API for Javadoc is also very useful.

Ideally, you would have just the API JavaDoc as the Reference Doc, and then if you want another doc w/ examples.

aozarov:
@lesv I am not sure I follow you.

Are you saying that you would like to take examples out of the javadoc (if so, can you point to some places, so we can discuses in more specifics). Also, what "and more..." is referring to?

lesv:
I'm talking about: http://googlecloudplatform.github.io/gcloud-java/0.1.5/apidocs/?com/google/gcloud/bigquery/package-summary.html

It would be nice if it were missing the examples and testing classes, so all-classes just had classes the developers can call by including the gcloud-java artifact.

aozarov:
Ah, I see. the packages with the ".test" suffix are not test but rather Test helpers (help the API users to test their code when using `gcloud-java`). I agree that taking out examples (and maybe even spi) from the main javadoc site would be a good thing.

aozarov:
@lesv see last image in #738. Changes will be publicly visible in the next release.
Is there anything else that is not covered or can we close the issue?

lesv:
LGTM

Issue #422, opened 2015-11-26
Should we repackage the spi layer?


ajkannan:
Right now all services' spi layers are bundled into one package (`com.google.gcloud.spi`).  While this provides the convenience of separating the spi javadoc from the api javadoc, the one package may not make sense as we add more and more services.

mziccard:
@ajkannan Could you please add more details on this? Are you suggesting to add one `spi` package per-service?

ajkannan:
Yes; we could make one spi package per service, and make sure that they get packaged separately from the api layer services

mziccard:
Reopening this. Should be closed as soon as this is also applied to compute.

mziccard:
Closed with #742 and #901

PR #738, opened 2016-03-10
Use groups to separate packages in javadoc's index

mziccard:
This follows from #733 and tries to make javadoc's index more clear by dividing packages into groups. Visual result is as follows:

screen shot 2016-03-10 at 4 38 18 pm

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -391,7 +391,7 @@
                 <outputDirectory>${project.build.directory}/javadoc</outputDirectory>
                 <groups>
                   <group>
-                    <title>Main packages</title>
+                    <title>API packages</title>
                     <packages>com.google.gcloud*</packages>
                   </group>
                   <group>
@@ -399,12 +399,12 @@
                     <packages>com.google.gcloud.bigquery.testing:com.google.gcloud.datastore.testing:com.google.gcloud.resourcemanager.testing:com.google.gcloud.storage.testing</packages>
                   </group>
                   <group>
-                    <title>SPI packages</title>
-                    <packages>com.google.gcloud.spi</packages>
-                  </group>
-                  <group>
                     <title>Example packages</title>
                     <packages>com.google.gcloud.examples*</packages>
+                  </group>
+                  <group>
+                    <title>SPI packages</title>
+                    <packages>com.google.gcloud.spi:com.google.gcloud.bigquery.spi:com.google.gcloud.datastore.spi:com.google.gcloud.resourcemanager.spi:com.google.gcloud.storage.spi</packages>
                   </group>
                 </groups>
               </configuration>


mziccard:
We could also consider grouping the testing helpers packages in their own group. Thoughts?

ajkannan:
Looks great!  I'm in favor of grouping the test helpers into their own group.  As we add more services, it'll help keep the "Main packages" section shorter.

gcloud-java-core/src/main/java/com/google/gcloud/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Core classes for the {@code gcloud-java} library.
+ */
+package com.google.gcloud;
```
ajkannan: Nit: add newline

mziccard:
Added a group for testing helpers packages. Unfortunately `*` wildcard is only allowed past a package name so I had to enumerate them. Visual result is:

screen shot 2016-03-10 at 6 56 19 pm

aozarov:
@mziccard *.test is not going to work?

mziccard:
> <packages>*.testing</packages> is not going to work?

No. I tried several combination and none worked. Quoting from the [docs](http://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#groups):

> The packages pattern can be any package name, or can be the start of any package name followed by an asterisk (*) meaning "match any characters".


> This brings me back to #422. Any thoughts?

It might be time to address this and create several `spi` packages as `com.google.gcloud.<service>.spi`. We could do it before this gets merged.

pom.xml
```diff
@@ -389,6 +389,24 @@
...
+                    <title>Main packages</title>
+                    <packages>com.google.gcloud*</packages>
+                  </group>
+                  <group>
+                    <title>Test helpers packages</title>
+                    <packages>com.google.gcloud.bigquery.testing:com.google.gcloud.datastore.testing:com.google.gcloud.resourcemanager.testing:com.google.gcloud.storage.testing</packages>
+                  </group>
+                  <group>
+                    <title>SPI packages</title>
+                    <packages>com.google.gcloud.spi</packages>
```
ludoch: Looks really great...
Maybe a different ordering (i.e API first, then SPI then Test, then
Examples)?
Also since we mention "SPI packages", should we rename the "Main packages"
as "API packages"?

ajkannan:
+1 for the `com.google.gcloud.<service>.spi` repackaging.

pom.xml
aozarov: +1 for s/Main packages/API packages/

My preference would be to order based on usefulness to users and therefore I would put testing before SPI (and maybe even consider examples before SPI). But not a strong opinion.

aozarov:
+1 for fixing #422 now.

> No. I tried several combination and none worked. Quoting from the docs:

Interesting the maven example itself uses * as a prefix.
see `<packages>*.util.*</packages>` [here](http://maven.apache.org/plugins/maven-javadoc-plugin/examples/group-configuration.html).

mziccard:
Rebased, renamed "Main packages" to "API packages", reordered groups to leave SPI at the end and updated spi package names to service-specific ones.

ludoch:
SGTM

mziccard:
PS: this is the visual effect

screen shot 2016-03-11 at 6 08 29 pm

ajkannan:
LGTM

PR #741, opened 2016-03-10
Fix merge mishap

jean-philippe-martin:
Sorry! I don't know how that slipped by!

### DIFF
--- a/gcloud-java-contrib/gcloud-java-nio/src/main/resources/META-INF/services/java.nio.file.spi.FileSystemProvider
+++ b/gcloud-java-contrib/gcloud-java-nio/src/main/resources/META-INF/services/java.nio.file.spi.FileSystemProvider
@@ -1,1 +1,0 @@
-com.google.gcloud.storage.contrib.nio.CloudStorageFileSystemProvider


aozarov:
Looks like there are more merge issues? See Travis logs.

jean-philippe-martin:
How can I reproduce the problem? The log [suggests](https://travis-ci.org/GoogleCloudPlatform/gcloud-java/jobs/115132729#L139) that `mvn clean` triggers the problem, but for me it doesn't.

I tried `mvn clean` and `mvn clean compile test`, both worked fine.

jean-philippe-martin:
OK this is super weird. I just tried `mvn compile` and it returned success, including GCloud Java NIO. However, when I rebuild this same project from the IDE I get an error about BlobInfo.Builder.size not being public (apparently [true](https://github.com/GoogleCloudPlatform/gcloud-java/blob/gcs-nio/gcloud-java-storage/src/main/java/com/google/gcloud/storage/BlobInfo.java#L154)). `mvn clean compile` also worked, similarly reporting success in compiling this code. 

Clearly there is something wrong in my mental model about what's going on here. Would you have some wisdom to share?

jean-philippe-martin:
The core problem with this merge appears to be that the Storage.get interface was changed to return a Blob instead of a BlobInfo, and there is no way to create a mock Blob to return. So I had to change its interface a little bit to allow us to still test against it.

Please have a look; you might prefer doing things differently, such as adding a static "builder" method to the Blob class.

aozarov:
when we do so, we should make `equals` and `hashCode` final.

jean-philippe-martin:
Merging with master's now-not-final Blob class allowed us to mock the Blob, and then make things work. Two things of note in the process, though:
 * I had to change Blob.equals to short-circuit the pointer-equality case (probably a limitation of the mock system)
 * I had to put the java.nio.file.spi.FileSystemProvider file back; without it, the gs provider wasn't installed (as if AutoService mysteriously stopped working - I tried building multiple times)

jean-philippe-martin:
Ready to merge as soon as I get the OK. Googlebot's warning seems spurious.

ajkannan:
I don't think the java.nio.file.spi.FileSystemProvider file is necessary.  After removing that file and building via command line with Maven, the Stat example seems to work.  Why do you think it's necessary/what kinds of issues were you seeing?

jean-philippe-martin:
The issue I had was that tests would fail in the IDE (intelliJ in my case) even after building multiple times. If adding a redundant file allows us to run tests in the IDE then it's worthwhile, even though of course ideally we shouldn't need it.

aozarov:
I verified that with my version of intelliJ (Community edition 2016.1) my tests pass fine without the meta file *after* I build the project (project is configured as a maven project).

Also, this should not be an issue for most people that depends on the gcs-nio jar directly.

However, if you still think we should keep the file then I see no reason for adding the `auto-service` dependency and using its annotation.

jean-philippe-martin:
I had version 15.0.4. I updated to 2016.1 and this seems to have fixed the problem. I was able to build and test fine without the spi file, so I'm removing it.

Issue #739, opened 2016-03-10
Possible Gcloud Java bug in signurl?


OverclockedTim:
So I'm using gcloud storage in a managed VM app and generally loving it, but I've hit something that I really think it is a bug so I wanted to throw it out there and see what people think.

I'm doing a few things with Storage and most of them are working perfectly  Reads work perfectly, writes work perfectly, creates work perfectly.  However, when I attempt to sign a URL like so:

Storage storage = StorageOptions.defaultInstance().service();

URL url = storage.signUrl(
        BlobInfo.builder(BUCKET_NAME,
                OBJECT_NAME).build(),
        30,
        TimeUnit.MINUTES,
        Storage.SignUrlOption.httpMethod(HttpMethod.GET));

I end up getting a "Signing key not provided" exception: 

java.lang.IllegalArgumentException: Signing key was not provided and could not be derived
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
	at com.google.gcloud.storage.StorageImpl.signUrl(StorageImpl.java:531)
        ...

This occurs on a deployed Managed VM server. It also occurs locally when gcloud is authorized either to myself or to a service account.  In other words, it appears that storage.signUrl is not picking up signing authorization from the standard locations.  I have not yet tried to manually add a signing option from a generated key, but that seems like it should not required.  Additionally, the signUrl example from the gcloud-java source in maven has no additional signingKey in the SignUrlOptions which leads me to believe that it is intended for signUrl to pick up its authorization from the environment.

mziccard:
Hi @OverclockedTim thanks for the report and the kind words:)

What `signUrl` does is looking in the credentials you provided to `StorageOptions` for a private key. If you are authenticated with default credentials you provided no private key (because you don't need to) hence the complaint `Signing key not provided`.

To use `signUrl` when authenticated with default credentials you can pass a private key using the `SignUrlOption.serviceAccount` option. Your code would become:

```java
// Create ServiceAccountAuthCredentials from a stream
ServiceAccountAuthCredentials credentials =
    AuthCredentials.createForJson(new FileInputStream("/path/to/key.json"));
// or from a String (account) and PrivateKey
ServiceAccountAuthCredentials credentials = AuthCredentials.createFor(ACCOUNT, PRIVATE_KEY);
// Create the signed URL
URL url = storage.signUrl(BlobInfo.builder(BUCKET_NAME, OBJECT_NAME).build(), 30,
[...]

PR #743, opened 2016-03-11
Add more detailed javadoc to Blob and Storage signUrl

mziccard:
This follows from #739 and tries to add a better javadoc to `Storage.signUrl` and `Blob.signUrl`

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
@@ -456,27 +456,27 @@ public final class Blob extends BlobInfo {
   /**
    * Generates a signed URL for this blob. If you want to allow access for a fixed amount of time to
    * this blob, you can use this method to generate a URL that is only valid within a certain time
-   * period. This is particularly useful if you don't want publicly accessible blobs, but don't want
-   * to require users to explicitly log in. Signing a URL requires a service account
-   * and its associated private key. If a {@link AuthCredentials.ServiceAccountAuthCredentials} was
-   * passed to {@link StorageOptions.Builder#authCredentials(AuthCredentials)} or the default
-   * credentials are being used and the environment variable {@code GOOGLE_APPLICATION_CREDENTIALS}
-   * is set, then {@code signUrl} will use that service account and associated key to sign the URL.
-   * If the credentials passed to {@link StorageOptions} do not expose a private key (this is the
-   * case for App Engine credentials, Compute Engine credentials and Google Cloud SDK credentials)
-   * then {@code signUrl} will throw an {@link IllegalArgumentException} unless a service account
-   * with associated key is passed using the {@code SignUrlOption.serviceAccount()} option. The
-   * service account and private key passed with {@code SignUrlOption.serviceAccount()} have
-   * priority over any credentials set with
-   * {@link StorageOptions.Builder#authCredentials(AuthCredentials)}.
+   * period. This is particularly useful if you don't want publicly accessible blobs, but also don't
+   * want to require users to explicitly log in. Signing a URL requires a service account and its
+   * associated private key. If a {@link AuthCredentials.ServiceAccountAuthCredentials} was passed
+   * to {@link StorageOptions.Builder#authCredentials(AuthCredentials)} or the default credentials
+   * are being used and the environment variable {@code GOOGLE_APPLICATION_CREDENTIALS} is set, then
+   * {@code signUrl} will use that service account and associated key to sign the URL. If the
+   * credentials passed to {@link StorageOptions} do not expose a private key (this is the case for
+   * App Engine credentials, Compute Engine credentials and Google Cloud SDK credentials) then
+   * {@code signUrl} will throw an {@link IllegalArgumentException} unless a service account with
+   * associated key is passed using the {@code SignUrlOption.serviceAccount()} option. The service
+   * account and private key passed with {@code SignUrlOption.serviceAccount()} have priority over
+   * any credentials set with {@link StorageOptions.Builder#authCredentials(AuthCredentials)}.
    *
-   * <p>Example usage of creating a signed URL that is valid for 2 weeks:
+   * <p>Example usage of creating a signed URL that is valid for 2 weeks, using the default
+   *     credentials for signing the URL:
    * <pre> {@code
    * blob.signUrl(14, TimeUnit.DAYS);
    * }</pre>
    *
    * <p>Example usage of creating a signed URL passing the {@code SignUrlOption.serviceAccount()}
-   *     option:
+   *     option, that will be used for signing the URL:
    * <pre> {@code
    * blob.signUrl(14, TimeUnit.DAYS, SignUrlOption.serviceAccount(
    *     AuthCredentials.createForJson(new FileInputStream("/path/to/key.json"))));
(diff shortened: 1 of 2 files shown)


gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -456,13 +457,40 @@ public WriteChannel writer(BlobWriteOption... options) {
    * Generates a signed URL for this blob. If you want to allow access to for a fixed amount of time
    * for this blob, you can use this method to generate a URL that is only valid within a certain
    * time period. This is particularly useful if you don't want publicly accessible blobs, but don't
-   * want to require users to explicitly log in.
+   * want to require users to explicitly log in. Signing a URL requires a service account
+   * and its associated key. If a {@link AuthCredentials.ServiceAccountAuthCredentials} was passed
+   * to {@link StorageOptions.Builder#authCredentials(AuthCredentials)} or the default credentials
+   * are being used and the environment variable {@code GOOGLE_APPLICATION_CREDENTIALS} is set, then
+   * {@code signUrl} will use that service account and associated key to sign the URL. If this
```
ajkannan: I think we can clarify the precedence of credentials specification a bit more here.  If the user supplies a SignUrlOption with service credentials, we [use those](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java#L540) instead of the service options' credentials.  If it isn't specified, then we turn to the other options.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1476,23 +1477,43 @@ private static void checkContentType(BlobInfo blobInfo) throws IllegalArgumentEx
...
-   * This is particularly useful if you don't want publicly
-   * accessible blobs, but don't want to require users to explicitly log in.
+   * Generates a signed URL for a blob. If you have a blob that you want to allow access to for a
+   * fixed amount of time, you can use this method to generate a URL that is only valid within a
+   * certain time period. This is particularly useful if you don't want publicly accessible blobs,
+   * but don't want to require users to explicitly log in. Signing a URL requires a service account
+   * and its associated key. If a {@link ServiceAccountAuthCredentials} was passed to
+   * {@link StorageOptions.Builder#authCredentials(AuthCredentials)} or the default credentials are
+   * being used and the environment variable {@code GOOGLE_APPLICATION_CREDENTIALS} is set, then
+   * {@code signUrl} will use that service account and associated key to sign the URL. If this
```
ajkannan: Same comment above applies here.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -456,13 +457,40 @@ public WriteChannel writer(BlobWriteOption... options) {
...
+   * is not the case, a service account with associated key can be passed to {@code signUrl} using
```
ajkannan: Nit: I think it would sound slightly better to replace "a service account with associated key" with "service account credentials"

ajkannan:
Thanks for taking this issue!  Just a few nits from me.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -456,13 +457,40 @@ public WriteChannel writer(BlobWriteOption... options) {
    * Generates a signed URL for this blob. If you want to allow access to for a fixed amount of time
```
aozarov: "access to for a" -> "access for a"?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -456,13 +457,40 @@ public WriteChannel writer(BlobWriteOption... options) {
...
    * for this blob, you can use this method to generate a URL that is only valid within a certain
```
aozarov: "for this blob" -> "to this blob"?

aozarov:
Great. +1 for @ajkannan point about making the order clear.  @OverclockedTim is that clear enough?

OverclockedTim:
This is definitely looking great, very helpful.  Would you guys consider adding a comment similar to what @mziccard mentioned in #739: "Signing a URL involves signing with a private key. If the credentials used to authenticate do not expose one (this is the case for App Engine credentials, Compute Engine credentials and Google Cloud SDK credentials) there is nothing we can do to help."?  That was the most helpful line for me personally, since I did not previously realize that those credential types didn't contain the necessary private keys.

aozarov:
Yes, that is a good idea. Though I think we should direct that statement in the fallback option when `serviceAccount` is not set.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
mziccard: I know but I am trying to make clear the point here is having a "key". While "service account credentials" is used in official docs it's also a vague term, a user might assume that since he is logged in gcloud he already has the "credentials" he needs. I would stick to the verbose "a service account with associated key" if you don't feel too strong about this.

mziccard:
I rephrased javadoc to address @ajkannan and @OverclockedTim comments.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -453,16 +454,48 @@ public WriteChannel writer(BlobWriteOption... options) {
   }
 
   /**
-   * Generates a signed URL for this blob. If you want to allow access to for a fixed amount of time
-   * for this blob, you can use this method to generate a URL that is only valid within a certain
-   * time period. This is particularly useful if you don't want publicly accessible blobs, but don't
-   * want to require users to explicitly log in.
+   * Generates a signed URL for this blob. If you want to allow access for a fixed amount of time to
+   * this blob, you can use this method to generate a URL that is only valid within a certain time
+   * period. This is particularly useful if you don't want publicly accessible blobs, but don't want
```
aozarov: Nit, I think "but also don't want" would sound slightly better in this 2 negative clauses.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -453,16 +454,48 @@ public WriteChannel writer(BlobWriteOption... options) {
...
+   * is set, then {@code signUrl} will use that service account and associated key to sign the URL.
+   * If the credentials passed to {@link StorageOptions} do not expose a private key (this is the
+   * case for App Engine credentials, Compute Engine credentials and Google Cloud SDK credentials)
+   * then {@code signUrl} will throw an {@link IllegalArgumentException} unless a service account
+   * with associated key is passed using the {@code SignUrlOption.serviceAccount()} option. The
+   * service account and private key passed with {@code SignUrlOption.serviceAccount()} have
+   * priority over any credentials set with
+   * {@link StorageOptions.Builder#authCredentials(AuthCredentials)}.
+   *
+   * <p>Example usage of creating a signed URL that is valid for 2 weeks:
```
aozarov: should we say "and using the default credentials for signing the URL" or is that stating the obvious?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -453,16 +454,48 @@ public WriteChannel writer(BlobWriteOption... options) {
...
+   * <pre> {@code
+   * blob.signUrl(14, TimeUnit.DAYS);
+   * }</pre>
+   *
+   * <p>Example usage of creating a signed URL passing the {@code SignUrlOption.serviceAccount()}
```
aozarov: should we say "that will be used for signing the URL"?

aozarov:
Looks good. Feel free to merge with or without my suggestions.

mziccard:
Comments fixed. Merging!

PR #745, opened 2016-03-14
Rename maxResults to pageSize

### LINKED ISSUE #740: Standardize "maxResults"/"pageSize" option terminology
We use "pageSize" in Resource Manager and "maxResults" in Storage and BigQuery to mean "max results per page".  We should standardize our terminology.  I slightly prefer the term "pageSize" because a user could want some total number of results that's greater than what the service allows as the pageSize.  As a result, the user might set a "maxResults" field, get less results than they asked for, and perhaps assume (without checking the page token) that they got all the results that existed, even if there is another page of results.

We should standardize the javadoc to include that it's "the maximum number of results returned per page."

mziccard:
This fixes #740

### DIFF
--- a/gcloud-java-bigquery/README.md
+++ b/gcloud-java-bigquery/README.md
@@ -185,7 +185,7 @@ Then add the following code to run the query and wait for the result:
 QueryRequest queryRequest =
     QueryRequest.builder("SELECT * FROM my_dataset_id.my_table_id")
         .maxWaitTime(60000L)
-        .maxResults(1000L)
+        .pageSize(1000L)
         .build();
 // Request query to be executed and wait for results
 QueryResponse queryResponse = bigquery.query(queryRequest);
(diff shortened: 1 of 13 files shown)


mderka:
Looks good. Feel free to merge.

A follow up question though. I ran grep on all .java files that we have in the library. We are still using `MAX_RESULTS` enum in our Rpc.Option classes. While this is an constant for "maxResults" selector, would it be better to rename this one as well, i.e., to have `PAGE_SIZE("maxResults")`? If we do this, constructs such as

```
return new BucketListOption(StorageRpc.Option.MAX_RESULTS, pageSize);
```

will become

```
return new BucketListOption(StorageRpc.Option.PAGE_SIZE, pageSize);
```

I do not have a clear opinion, however, the latter seems a nicer to me. The name switch from "page size" and to "max results" has to happen somewhere and I would probably try to push it as close to the call itself as possible. 

/cc @ajkannan @aozarov

mziccard:
The name `MAX_RESULTS` remained the `StorageRpc.Option` enum because of the actual name of the parameter that is expected by the JSON API. `StorageRpc.Option` is not exposed to the user so I think we should stick to this naming as it's a better remainder of the actual param name.

mderka:
Ok, fine with me.

PR #746, opened 2016-03-14
LocalDnsHelper should add the default change upon creating a zone.

### LINKED ISSUE #672: LocalDnsHelper does not create a change for new records
The service does. Add a change.

mderka:
This now matches the behaviour of the service. Fixes #672.


mziccard:
Looks good to me

PR #760, opened 2016-03-17
IAM docs and functional methods for policies

ajkannan:


gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
```diff
@@ -198,6 +201,71 @@ public Project replace() {
     return resourceManager.replace(this);
   }
 
+  /**
+   * Returns the IAM access control policy for the specified project. Returns {@code null} if the
```
aozarov: for this project?

gcloud-java-resourcemanager/src/main/java/com/google/gcloud/resourcemanager/Project.java
```diff
@@ -198,6 +201,71 @@ public Project replace() {
...
+    return resourceManager.getPolicy(projectId());
+  }
+
+  /**
+   * Sets the IAM access control policy for the specified project. Replaces any existing policy.
+   * It is recommended that you use the read-modify-write pattern. See code samples and important
+   * details of replacing policies in the documentation for {@link ResourceManager#replacePolicy}.
+   *
+   * @throws ResourceManagerException upon failure
+   * @see ResourceManager#replacePolicy
```
aozarov: Any reason for this reference (I mean, is necessary and if so why not in others).

aozarov: and `@return` is missing from some.

gcloud-java-examples/src/main/java/com/google/gcloud/examples/resourcemanager/snippets/ModifyPolicy.java
```diff
@@ -0,0 +1,64 @@
...
+    Project project = resourceManager.get(projectId);
+
+    // Get the project's policy
+    Policy policy = project.getPolicy();
+
+    // Add a viewer
+    Policy.Builder modifiedPolicy = policy.toBuilder();
+    Identity newViewer = Identity.user("<insert user's email address here>");
+    if (policy.bindings().containsKey(Role.viewer())) {
+      modifiedPolicy.addIdentity(Role.viewer(), newViewer);
```
aozarov: This example really makes me rethink how we deal with the `Policy.Builder`.

We discussed that before however I don't remember the resolution.
I wonder if having `addBinding(Role, Identity, Identity...)` and `removeBinding(Role)`
and `removeBinding(Role, identity, identity...)` which will remove or create the role entry automatically
is not a nicer API. Do you see issues with it? Also, when adding the same identity twice to the same role
I think we should silently ignore it (Maybe LinkedHashSet)?

aozarov:
Few javadoc comments and a suggest to revisit `IamPolicy.Builder`

gcloud-java-examples/src/main/java/com/google/gcloud/examples/resourcemanager/snippets/ModifyPolicy.java
ajkannan: The above is implemented in [c05e738](https://github.com/GoogleCloudPlatform/gcloud-java/pull/760/commits/c05e738b1f17048560b781f4445ca1ccff7f0759).  I didn't use a LinkedHashSet b/c we already use a HashSet to represent identities, and we don't really care about order.

ajkannan:
Rebased to deal with merge conflicts and added [c05e738](https://github.com/GoogleCloudPlatform/gcloud-java/pull/760/commits/c05e738b1f17048560b781f4445ca1ccff7f0759) to address comments.

aozarov:
Looks good. Feel free to merge after applying the minor changes we discussed including:
1. rename removeBinding to removeRole
2. remove role if all identities where removed
3. embed verify applicable parts
4. replace checkArgument with checkNotNull when we check for nulls
5. simplify addIdentity javadoc

ajkannan:
Added a commit to take care of comments from @aozarov.

PR #761, opened 2016-03-17
Update and fix gcloud-java-logging

shinfan:


aozarov:
Can we have a PR *before* this one to pull updates from master (and then scrap this PR and create a new based on that update)?

If you prefer not to do that then let's at least update `.travis.yml` as the one you have here does not build
your PR on Travis.

PR #762, opened 2016-03-18
Remove default contentType from Storage.compose and Bucket.create

### LINKED ISSUE #288: Remove setContentType from DefaultStorageRpc.compose
This depends on b/20681287. Compose API needs a `contentType` to be always provided. As soon as the bug gets fixed we should remove `setContentType("application/octet-stream")`.

mziccard:
This fixes #288

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/CopyWriter.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/CopyWriter.java
@@ -33,8 +33,8 @@ import java.util.concurrent.Callable;
 
 /**
  * Google Storage blob copy writer. A {@code CopyWriter} object allows to copy both blob's data and
- * information. To override source blob's information call {@link Storage#copy(Storage.CopyRequest)}
- * with a {@code CopyRequest} object where the copy target is set via
+ * information. To override source blob's information supply a {@code BlobInfo} to the
+ * {@code CopyRequest} using either
  * {@link Storage.CopyRequest.Builder#target(BlobInfo, Storage.BlobTargetOption...)} or
  * {@link Storage.CopyRequest.Builder#target(BlobInfo, Iterable)}.
  *
@@ -176,7 +176,7 @@ public class CopyWriter implements Restorable<CopyWriter> {
       private final BlobId source;
       private final Map<StorageRpc.Option, ?> sourceOptions;
       private final boolean overrideInfo;
-      private BlobInfo target;
+      private final BlobInfo target;
       private final Map<StorageRpc.Option, ?> targetOptions;
       private BlobInfo result;
       private long blobSize;
--- a/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
+++ b/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
@@ -1386,9 +1386,10 @@ public interface Storage extends Service<StorageOptions> {
 
   /**
    * Sends a copy request. This method copies both blob's data and information. To override source
-   * blob's information set the copy target via
-   * {@link CopyRequest.Builder#target(BlobInfo, BlobTargetOption...)} or
-   * {@link CopyRequest.Builder#target(BlobInfo, Iterable)}.
+   * blob's information supply a {@code BlobInfo} to the
+   * {@code CopyRequest} using either
+   * {@link Storage.CopyRequest.Builder#target(BlobInfo, Storage.BlobTargetOption...)} or
+   * {@link Storage.CopyRequest.Builder#target(BlobInfo, Iterable)}.
    *
    * <p>This method returns a {@link CopyWriter} object for the provided {@code CopyRequest}. If
    * source and destination objects share the same location and storage class the source blob is


gcloud-java-storage/src/main/java/com/google/gcloud/storage/Bucket.java
```diff
@@ -654,16 +651,51 @@ public Blob create(String blob, byte[] content, String contentType, BlobTargetOp
...
+   * computed and used for validating transferred data.
+   *
+   * @param blob a blob name
+   * @param content the blob content
+   * @param options options for blob creation
+   * @return a complete blob information
+   * @throws StorageException upon failure
+   */
+  public Blob create(String blob, byte[] content, BlobTargetOption... options) {
+    BlobInfo blobInfo = BlobInfo.builder(BlobId.of(name(), blob)).build();
```
aozarov: Any reason for not delegating to `create(blob, content, null, options)`? Same for bellow.

aozarov: Oh, I see. Yes there is a conceptual difference between unset and explicit set to null.
However, because this is create and not an update is there a practical difference?

aozarov:
Looks great. Thanks for doing that. Should we also add something to the integration test?
Also, I suspect [this](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java#L1039) enforcment is no longer needed?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Bucket.java
mziccard: Yes in practice there should be no difference as this is a create operation. I slightly prefer this way as it does not mess with `Data.nullOf`.

aozarov: I am fine with keeping as is. Was just wondering if there was more to it.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -973,7 +972,8 @@ public static Builder builder() {
       private final Set<BlobSourceOption> sourceOptions = new LinkedHashSet<>();
       private final Set<BlobTargetOption> targetOptions = new LinkedHashSet<>();
       private BlobId source;
-      private BlobInfo target;
+      private BlobId targetId;
```
aozarov: Considering that this is private and we control how it is set I think it is fine, however my preference
would have been not to have duplicate `BlobId` (which conceptually can get out of sync) and instead have a boolean to indicate if body should be used to replace metadata

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1021,39 +1021,37 @@ public Builder sourceOptions(Iterable<BlobSourceOption> options) {
...
         return this;
       }
 
       /**
        * Sets the copy target and target options. {@code target} parameter is used to override
-       * source blob information (e.g. {@code contentType}, {@code contentLanguage}). {@code
-       * target.contentType} is a required field.
+       * source blob information (e.g. {@code contentType}, {@code contentLanguage}). Target blob
+       * information is set exactly to {@code target}, no information is inherited from the source
+       * blob. If not set, target blob information is inherited from the source blob.
```
aozarov: I am not sure I understand the last statement, "f not set, target blob information is inherited from the source blob.", in this context.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1021,39 +1021,37 @@ public Builder sourceOptions(Iterable<BlobSourceOption> options) {
...
        *
        * @return the builder
-       * @throws IllegalArgumentException if {@code target.contentType} is {@code null}
        */
-      public Builder target(BlobInfo target, BlobTargetOption... options)
-          throws IllegalArgumentException {
-        checkContentType(target);
-        this.target = target;
+      public Builder target(BlobInfo targetInfo, BlobTargetOption... options) {
+        this.targetId = targetInfo.blobId();
```
aozarov: We missed that before but an explicit `checkNotNull` for target would be nice.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1074,8 +1072,6 @@ public Builder megabytesCopiedPerChunk(Long megabytesCopiedPerChunk) {
        * Creates a {@code CopyRequest} object.
        */
       public CopyRequest build() {
-        checkNotNull(source);
```
aozarov: I think we could have left that here (though only checking source and targetId if you ignore my previous suggestion).

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1103,10 +1100,20 @@ public BlobId source() {
...
-      return target;
+    public BlobId targetId() {
+      return targetId;
+    }
+
+    /**
+     * Returns the {@link BlobInfo} for the target blob. If set, this value is used to replace
+     * source blob information (e.g. {@code contentType}, {@code contentLanguage}). Target blob
+     * information is set exactly to this value, no information is inherited from the source blob.
+     * If not set, target blob information is inherited from the source blob.
```
aozarov: shall we be explicit and say 'If null target blob information is inhe...."?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1127,34 +1134,27 @@ public Long megabytesCopiedPerChunk() {
...
      * @param sourceBucket name of the bucket containing the source blob
      * @param sourceBlob name of the source blob
      * @param target a {@code BlobInfo} object for the target blob
      * @return a copy request
-     * @throws IllegalArgumentException if {@code target.contentType} is {@code null}
      */
-    public static CopyRequest of(String sourceBucket, String sourceBlob, BlobInfo target)
-        throws IllegalArgumentException {
-      checkContentType(target);
+    public static CopyRequest of(String sourceBucket, String sourceBlob, BlobInfo target) {
```
aozarov: Fine with you prior change of calling target targetId or targetInfo though my preference would have been to leave it target in all places as before.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/CopyWriter.java
```diff
@@ -120,8 +120,12 @@ public RewriteResponse call() {
         serviceOptions,
         BlobId.fromPb(rewriteResponse.rewriteRequest.source),
         rewriteResponse.rewriteRequest.sourceOptions,
-        BlobInfo.fromPb(rewriteResponse.rewriteRequest.target),
+        BlobId.of(rewriteResponse.rewriteRequest.targetBucket,
+            rewriteResponse.rewriteRequest.targetName),
         rewriteResponse.rewriteRequest.targetOptions)
+        .targetInfo(rewriteResponse.rewriteRequest.targetObject != null
+            ? BlobInfo.fromPb(rewriteResponse.rewriteRequest.targetObject) : null)
+        .result(rewriteResponse.result != null ? BlobInfo.fromPb(rewriteResponse.result) : null)
```
aozarov: Wow, did we miss result from capture? Nice catch. did you update the test for it?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -1021,39 +1021,37 @@ public Builder sourceOptions(Iterable<BlobSourceOption> options) {
        *
        * @return the builder
        */
-      public Builder target(BlobId target) {
-        this.target = BlobInfo.builder(target).build();
+      public Builder target(BlobId targetId) {
```
aozarov: Do you think we need to make it clear in Storage.copy command or CopyWriter class that copy copies the content AND information though latter could be overridden?

aozarov:
Nice, few more comments/suggestions.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/CopyWriter.java
mziccard: I added a dedicated test

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
mziccard: Those checks are already made in the constructor, that's why I removed them.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/CopyWriter.java
```diff
@@ -171,9 +175,9 @@ public RewriteResponse call() {
       private final StorageOptions serviceOptions;
       private final BlobId source;
       private final Map<StorageRpc.Option, ?> sourceOptions;
-      private final BlobId targetId;
+      private final boolean overrideInfo;
+      private BlobInfo target;
```
aozarov: can this be `null`?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/CopyWriter.java
```diff
@@ -32,7 +32,13 @@
 import java.util.concurrent.Callable;
 
 /**
- * Google Storage blob copy writer. This class holds the result of a copy request. If source and
+ * Google Storage blob copy writer. A {@code CopyWriter} object allows to copy both blob's data and
+ * information. To override source blob's information call {@link Storage#copy(Storage.CopyRequest)}
```
aozarov: Fine as is. I would just have said something like "To override source blob's information supply a BlobInfo as the target" or something like but if you think what you have is more clear feel free to keep it.

aozarov:
LGTM. Feel free to merge after considering the 2 comments above.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/CopyWriter.java
mziccard: Yep I missed it.

PR #765, opened 2016-03-18
Added batch concept supporting list zones calls only.

mderka:
This commit contains a concept of batches. It currently supports a single operation (list zones) only, but it is functional and contains the entire stack including tests. If this gets approved, I will proceed by implementing the remaining methods.

### DIFF
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
@@ -533,4 +533,19 @@ public interface Dns extends Service<DnsOptions> {
    * @see <a href="https://cloud.google.com/dns/api/v1/changes/list">Cloud DNS Chages: list</a>
    */
   Page<ChangeRequest> listChangeRequests(String zoneName, ChangeRequestListOption... options);
+
+  /**
+   * Submits a {@code batch} for processing to the Google Cloud DNS. The operations in the batch
+   * will be called using a single HTTP request. For each successfully executed operation, its
+   * {@link DnsBatch.Callback#success(Object, DnsBatch.Request)} callback method will be invoked.
+   * For each operation which returned an error, its {@link DnsBatch.Callback#error(DnsException,
+   * DnsBatch.Request)} callback method will be invoked.
+   */
+  void submitBatch(DnsBatch batch);
+
+  /**
+   * Initiates a new empty batch ready to be populated with service calls, which will use this
+   * {@code Dns} instance when submitted for processing to Google Cloud DNS.
+   */
+  DnsBatch batch();
 }
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
@@ -1,0 +1,231 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gcloud.dns;
+
+import com.google.gcloud.Page;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * A batch of operations to be submitted to Google Cloud DNS using a single HTTP request.
+ */
+public class DnsBatch {
+
+  private Map<Request, Callback> requests = new LinkedHashMap<>();
+  private Dns dns;
+
+  /**
+   * An interface for the callback which will be invoked when the operation has been executed. The
+   * parameter {@code <T>} represents the type of the result of the operation and thus depends on
+   * the {@link DnsBatch.Request} that this call back belongs to and it should be as follows:
+   *
+   * <ul>
+   *   <li>{@link Zone} for creating and getting a zone</li>
+   *   <li>{@link Boolean} for deleting a zone</li>
+   *   <li>{@link ChangeRequest} for creating and getting a change request</li>
+   *   <li>{@link ProjectInfo} for getting a project</li>
+   *   <li>{@code Page<Zone>} for listing zones</li>
+   *   <li>{@code Page<DnsRecord>} for listing {@link DnsRecord}s inside a zone</li>
+   *   <li>{@code Page<ChangeRequest>} for listing {@link ChangeRequest}s for a zone</li>
+   * </ul>
+   */
+  public interface Callback<T> {
+    /**
+     * A method which will be called if the {@link DnsBatch.Request} succeeds. See the {@link
+     * Callback} documentation for details on type {@code T}.
+     *
+     * @param output the result of the operation
+     * @param request the request which succeeded
+     */
+    void success(T output, DnsBatch.Request request);
+
+    /**
+     * A method which will be called if the {@link DnsBatch.Request} fails.
+     *
+     * @param ex the error
+     * @param request the request which failed
+     */
+    void error(DnsException ex, DnsBatch.Request request);
+  }
+
+  /**
+   * An operation to be submitted to Google Cloud DNS within this batch. Only an subset of the class
+   * attributes appropriate for the represented operation is initialized. Refer to the class method
+   * and attribute documentation for the specific fields.
+   */
+  public static class Request {
+
+    private final String zoneName;
+    private final String changeId;
+    private final ChangeRequest changeRequest;
+    private final ZoneInfo zoneInfo;
+    private final Operation operation;
+    private final AbstractOption[] options;
+
+    private Request(RequestBuilder builder) {
+      this.zoneName = builder.zoneName;
+      this.changeId = builder.changeId;
+      this.changeRequest = builder.changeRequest;
+      this.zoneInfo = builder.zoneInfo;
+      this.operation = builder.operation;
+      this.options = builder.options;
+    }
+
+    private static RequestBuilder builder(Operation operation, AbstractOption... options) {
+      return new RequestBuilder(operation, options);
+    }
+
+    /**
+     * Returns the name of the zone to which the operation is applied. This field is initialized for
+     * zone create, get and delete operation, and listing DNS records and changes within a zone.
+     * Returns {@code null} in other cases.
+     */
+    public String zoneName() {
+      return zoneName;
+    }
+
+    /**
+     * Returns the id of the change request which is being retrieved. Getting a change request is
+     * the only operation when this attribute is initialized. The method returns {@code null} in the
+     * remaining cases.
+     */
+    public String changeId() {
+      return changeId;
+    }
+
+    /**
+     * Returns the change request which is being created. Creating a change request is the only
+     * operation when this attribute is initialized. The method returns {@code null} in the
+     * remaining cases.
+     */
+    public ChangeRequest changeRequest() {
+      return changeRequest;
+    }
+
+    /**
+     * Returns the zone which is being created. Creating a zone is the only operation when this
+     * attribute is initialized. The method returns {@code null} in the remaining cases.
+     */
+    public ZoneInfo zoneInfo() {
+      return zoneInfo;
+    }
+
+    /**
+     * Returns the type of the operation represented by this {@link DnsBatch.Request}. This field is
+     * always initialized.
+     */
+    public Operation operation() {
+      return operation;
+    }
+
+    /**
+     * Returns options provided to the operation. Returns an empty array if no options were
+     * provided.
+     */
+    public AbstractOption[] options() {
+      return options == null ? new AbstractOption[0] : options;
+    }
+  }
+
+  static class RequestBuilder {
+    private final AbstractOption[] options;
+    private String zoneName;
+    private String changeId;
+    private ChangeRequest changeRequest;
+    private ZoneInfo zoneInfo;
+    private final Operation operation;
+
+    RequestBuilder(Operation operation, AbstractOption... options) {
+      this.operation = operation;
+      this.options = options;
+    }
+
+    RequestBuilder zoneName(String zoneName) {
+      this.zoneName = zoneName;
+      return this;
+    }
+
+    RequestBuilder changeId(String changeId) {
+      this.changeId = changeId;
+      return this;
+    }
+
+    RequestBuilder changeRequest(ChangeRequest changeRequest) {
+      this.changeRequest = changeRequest;
+      return this;
+    }
+
+    RequestBuilder zoneInfo(ZoneInfo zoneInfo) {
+      this.zoneInfo = zoneInfo;
+      return this;
+    }
+
+    Request build() {
+      return new Request(this);
+    }
+  }
+
+  /**
+   * Represents the type of the batch operation.
+   */
+  public enum Operation {
+    CREATE_ZONE,
+    DELETE_ZONE,
+    GET_ZONE,
+    LIST_ZONES,
+    APPLY_CHANGE_REQUEST,
+    GET_CHANGE_REQUEST,
+    LIST_CHANGES_REQUESTS,
+    LIST_DNS_RECORDS
+  }
+
+  DnsBatch(Dns dns) {
+    this.dns = dns;
+  }
+
+  public Dns service() {
+    return dns;
+  }
+
+  Map<Request, Callback> requests() {
+    return requests;
+  }
+
+  /**
+   * Adds a {@code DnsBatch.Request} represeting the list zones operation to this batch. The request
+   * will not have initialized any fields except for the operation type and options (if provided).
+   * The {@code callback} will receive a page of {@link Zone}s upon success of the request. The
+   * {@code options} can be used to restrict the fields returned or provide page size limits in the
+   * same way as for {@link Dns#listZones(Dns.ZoneListOption...)}.
+   */
+  public DnsBatch listZones(Callback<Page<Zone>> callback, Dns.ZoneListOption... options) {
+    Request request = Request.builder(Operation.LIST_ZONES, options).build();
+    requests.put(request, callback);
+    return this;
+  }
+
+  // todo(mderka) add the rest of the operations
+
+  /**
+   * Submits this batch for processing using a single HTTP request. This will invoke all callbacks
+   * for the invidual {@link DnsBatch.Request}s contained in this batch.
+   */
+  public void submit() {
+    dns.submitBatch(this);
+  }
+}
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
@@ -21,8 +21,13 @@ import static com.google.gcloud.RetryHelper.RetryHelperException;
 import static com.google.gcloud.RetryHelper.runWithRetries;
 import static com.google.gcloud.dns.ChangeRequest.fromPb;
 
+import com.google.api.client.googleapis.batch.BatchRequest;
+import com.google.api.client.googleapis.batch.json.JsonBatchCallback;
+import com.google.api.client.googleapis.json.GoogleJsonError;
+import com.google.api.client.http.HttpHeaders;
 import com.google.api.services.dns.model.Change;
 import com.google.api.services.dns.model.ManagedZone;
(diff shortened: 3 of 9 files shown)


gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
```diff
@@ -18,6 +18,8 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.api.services.dns.model.ManagedZone;
+import com.google.common.base.Function;
```
mziccard: Are the two above imports ever used?

You use `ManagedZone` but it's fully qualified. This is issue is also present in `DnsImpl` where you import `com.google.api.services.dns.model.ManagedZone` and also use it as fully qualified in your code (e.g. in `listZones`).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -533,4 +533,19 @@ ChangeRequest getChangeRequest(String zoneName, String changeRequestId,
    * @see <a href="https://cloud.google.com/dns/api/v1/changes/list">Cloud DNS Chages: list</a>
    */
   Page<ChangeRequest> listChangeRequests(String zoneName, ChangeRequestListOption... options);
+
+  /**
+   * Submits a {@code batch} for processing to the Google Cloud DNS. The operations in the batch
```
mziccard: Nit: I wouldn't use `{@code}` here, or explicit "the {@code batch} parameter"

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
```diff
@@ -533,4 +533,19 @@ ChangeRequest getChangeRequest(String zoneName, String changeRequestId,
...
+   * will be called using a single HTTP request. For each successfully executed operation, its
```
mziccard: Nit: s/called/carried out

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
```diff
@@ -0,0 +1,231 @@
...
+ */
+public class DnsBatch {
+
+  private Map<Request, Callback> requests = new LinkedHashMap<>();
+  private Dns dns;
+
+  /**
+   * An interface for the callback which will be invoked when the operation has been executed. The
+   * parameter {@code <T>} represents the type of the result of the operation and thus depends on
+   * the {@link DnsBatch.Request} that this call back belongs to and it should be as follows:
```
mziccard: s/call back/callback

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
```diff
@@ -0,0 +1,231 @@
...
+   *   <li>{@link ProjectInfo} for getting a project</li>
+   *   <li>{@code Page<Zone>} for listing zones</li>
+   *   <li>{@code Page<DnsRecord>} for listing {@link DnsRecord}s inside a zone</li>
+   *   <li>{@code Page<ChangeRequest>} for listing {@link ChangeRequest}s for a zone</li>
+   * </ul>
+   */
+  public interface Callback<T> {
+    /**
+     * A method which will be called if the {@link DnsBatch.Request} succeeds. See the {@link
+     * Callback} documentation for details on type {@code T}.
```
mziccard: Does this link to self work?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @param output the result of the operation
+     * @param request the request which succeeded
```
mziccard: Nit: the successful request?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
```diff
@@ -0,0 +1,231 @@
...
+     * A method which will be called if the {@link DnsBatch.Request} fails.
+     *
+     * @param ex the error
+     * @param request the request which failed
+     */
+    void error(DnsException ex, DnsBatch.Request request);
+  }
+
+  /**
+   * An operation to be submitted to Google Cloud DNS within this batch. Only an subset of the class
```
mziccard: s/an/a

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
```diff
@@ -0,0 +1,231 @@
...
+   * attributes appropriate for the represented operation is initialized. Refer to the class method
+   * and attribute documentation for the specific fields.
+   */
+  public static class Request {
```
mziccard: Did you consider having a hierarchy of requests? i.e. one for each operation each containing only its specific parameters? I think that would be cleaner and more in line with what we did so far. /cc @aozarov @ajkannan 

Also should this class be `Serializable`?

This class is also used as key in a map so we better add `hashCode` and `equals` methods.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * Returns options provided to the operation. Returns an empty array if no options were
+     * provided.
+     */
+    public AbstractOption[] options() {
+      return options == null ? new AbstractOption[0] : options;
+    }
+  }
+
+  static class RequestBuilder {
```
mziccard: Is there a reason for this to be outside of `Request` class?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
```diff
@@ -0,0 +1,231 @@
...
+
+    Request build() {
+      return new Request(this);
+    }
+  }
+
+  /**
+   * Represents the type of the batch operation.
+   */
+  public enum Operation {
```
mziccard: Nit: do you think it's worth adding dedicated javadoc for each enum or are the names already self-explanatory?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
```diff
@@ -309,11 +315,86 @@ public ChangeRequest getChangeRequest(final String zoneName, final String change
...
+  public DnsBatch batch() {
+    return new DnsBatch(this);
+  }
+
+  private JsonBatchCallback listZonesCallback(final DnsBatch.Callback callback,
+      final DnsBatch.Request request) {
+    return new JsonBatchCallback<ManagedZonesListResponse>() {
+      @Override
+      public void onFailure(GoogleJsonError error, HttpHeaders httpHeaders) {
+        if (callback != null) {
```
mziccard: Can this really happen? Also applies to `onSuccess`.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
```diff
@@ -309,11 +315,86 @@ public ChangeRequest getChangeRequest(final String zoneName, final String change
...
+      }
+
+      @Override
+      public void onSuccess(ManagedZonesListResponse zoneList, HttpHeaders httpHeaders) {
+        if (callback != null) {
+          DnsRpc.ListResult<ManagedZone> listResult =
+              DnsRpc.ListResult.of(zoneList.getNextPageToken(), zoneList.getManagedZones());
+          String cursor = listResult.pageToken();
+          Iterable<Zone> zones = listResult.results() == null ? ImmutableList.<Zone>of()
+              : Iterables.transform(listResult.results(), new Function<ManagedZone, Zone>() {
```
mziccard: This function is used also elsewhere (if I am not mistaken you define id as `pbToZoneFunction` in `listZones`). We should probably use a static factory method to create it.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
```diff
@@ -309,11 +315,86 @@ public ChangeRequest getChangeRequest(final String zoneName, final String change
...
+      }
+    };
+  }
+
   private Map<DnsRpc.Option, ?> optionMap(AbstractOption... options) {
     Map<DnsRpc.Option, Object> temp = Maps.newEnumMap(DnsRpc.Option.class);
-    for (AbstractOption option : options) {
-      Object prev = temp.put(option.rpcOption(), option.value());
-      checkArgument(prev == null, "Duplicate option %s", option);
+    if (options != null) {
```
mziccard: Can this ever happen? If you call it using `optionMap(request.options())` the getter already takes care of not returning `null`.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
```diff
@@ -54,6 +62,26 @@ public DefaultDnsRpc(DnsOptions options) {
   }
 
   @Override
+  public BatchRequest enqueueListZones(BatchRequest batch, final DnsBatch.Request request,
```
mziccard: This seems to me like a utility conversion methods that should reside in `DnsImpl` rather than a public method for `DnsRpc`. I see however that [queue](https://developers.google.com/api-client-library/java/google-api-java-client/reference/1.19.1/com/google/api/client/googleapis/services/json/AbstractGoogleJsonClientRequest#queue(com.google.api.client.googleapis.batch.BatchRequest,%20com.google.api.client.googleapis.batch.json.JsonBatchCallback)) can throw `IOException`. However I think that this `IOException` should not contain any DNS-related error as no network comunication should happen at this stage, right?

I am wondering wether we should expose this method or rather hide it in `DnsImpl`.

Also, is the `request` parameter used?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DnsRpc.java
```diff
@@ -171,4 +175,8 @@ Change getChangeRequest(String zoneName, String changeRequestId, Map<Option, ?>
    */
   ListResult<Change> listChangeRequests(String zoneName, Map<Option, ?> options)
       throws DnsException;
+
+  BatchRequest enqueueListZones(BatchRequest batch, DnsBatch.Request request,
```
mziccard: As commented above I am wondering whether we should expose this method or rather hide it in `DnsImpl`.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DnsRpc.java
```diff
@@ -171,4 +175,8 @@ Change getChangeRequest(String zoneName, String changeRequestId, Map<Option, ?>
...
+      JsonBatchCallback callback, Map<Option, ?> options) throws IOException;
```
mziccard: `IOException` is caught in the method right?

gcloud-java-dns/src/test/java/com/google/gcloud/dns/it/ITDnsTest.java
```diff
@@ -23,10 +23,12 @@
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import com.google.api.client.util.Lists;
```
mziccard: Use `com.google.common.collect.Lists` instead.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/it/ITDnsTest.java
```diff
@@ -952,4 +956,62 @@ public void testListDnsRecords() {
...
+  public void testListZoneBatch() {
+    DNS.create(ZONE1);
+    DNS.create(ZONE_EMPTY_DESCRIPTION);
+    // the mock is used to check that the callbacks were not skipped
+    final RuntimeException callbackCalled = EasyMock.createStrictMock(RuntimeException.class);
+    EasyMock.expect(callbackCalled.initCause(null)).andReturn(callbackCalled);
+    EasyMock.expect(callbackCalled.initCause(null)).andReturn(callbackCalled);
+    EasyMock.expect(callbackCalled.initCause(null)).andReturn(callbackCalled);
+    EasyMock.replay(callbackCalled);
+    DnsBatch.Callback<Page<Zone>> callback = new DnsBatch.Callback<Page<Zone>>() {
```
mziccard: Why not mocking the callback and use `expect` and possibly [capture](http://easymock.org/api/org/easymock/EasyMock.html#capture-org.easymock.Capture-) to check that the calls are made correctly? You would no longer need the `callbackCalled` workaround.

gcloud-java-dns/src/test/java/com/google/gcloud/dns/it/ITDnsTest.java
```diff
@@ -952,4 +956,62 @@ public void testListDnsRecords() {
...
+      @Override
+      public void success(Page<Zone> output, DnsBatch.Request request) {
+        List<Zone> zones = Lists.newArrayList(DNS.listZones().iterateAll());
```
mziccard: What about [Sets.newHashSet](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Sets.html#newHashSet(java.util.Iterator))? Also applies to the next callback.

mziccard:
I did a first pass. There are a couple of comments that are worth discussing.

I know that we already discussed this but seeing it done makes me think whether implementing batch with callbacks could give users the wrong idea that something is happening asynchronously. It feels slightly confusing to me.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/ZoneInfo.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Removed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/Dns.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mderka: No, useless.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mderka: fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mderka: I did consider it. Having an AbstractRequest and inherit from it 8 times seemed kind of overkill for the benefit, especially from the maintenance point of view, so I chose this solution. I would like to hear other opinions. @aozarov @ajkannan 

Made serializable for now.

With `hashCode` and `equals`, I did not override them on purpose. This should allow an operation to be called several times within a batch (I am not sure why it would be beneficial, but I definitely do not think that it should be prevented). If this has some caveat that I do not see, I will use a list of `Request,Callback` pairs instead to support duplicate calls. Let me know what you prefer.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mderka: None other than not having to type Dns.Request.Builder. Refactored.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mderka: I think this is self explanatory. It would look like:

```
/** create zone operation */
CREATE_ZONE
```

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
mderka: I think that a `null` callback should be supported. The user may want to just create bunch of zones without caring about the result.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
mderka: Done.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsImpl.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
mderka: This is a tricky one. I considered many options and this seems the best one to me. Below is the list. Long story short, something has to be exposed in `DnsRpc` interface and it should make sense, but there is no clear cut here.

1. The cleanest would be to have `DnsRpc` take care of submitting the batch completely so the exposed method would be `submitBatch(DnsBatch toSubmit)`. `DnsRpc` would then take care of everything but this includes paging. A page implementation requires page fetcher which is in the `DnsImpl` class in a different package. So, this option is out.
2. We can do everything from the DnsImpl class. However, this is completely bypassing the Rpc facade as the DnsImpl makes calls on its own. It is not the worst solution, but as a user, I would expect that if I provide my own implementation for `DnsRpc`, all the calls would be handled using that one.
3. My chosen solution was to use `DnsImpl` to create a collection of requests inside the `BatchRequest`. The callbacks are created in `Dns` which has the paging helpers, and the quing is handled via `Rpc` method, which is exposed. The `BatchRequest` object is created inside `Rpc` as it requires `com.google.api.services.dns.Dns;` which already contain the `HttpTransport`. Doing anything more in `DnsImpl` would require defining the `Rpc` functionality in `DnsImpl` and that is again bypassing the `DnsRpc` layer and duplicating the code.

I will welcome more suggestions, but for now, I am not touching this.

/cc @ajkannan @aozarov

mderka:
I think that we need to make a joint decision on what should be done with the `DnsRpc` and `DnsImpl`. In the mean time, I will fix the clear things and commit them tomorrow.

As far as the callbacks go, would you do instead? I think that futures also make the hint to asynchronous processing. I agree that we should specifically say that no asynchronous processing is happening in the docs, but I still like the callbacks. (Check the design doc for our initial discussion.)

Do you maybe want to schedule a meeting about it?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsBatch.java
mziccard: > I did consider it. Having an AbstractRequest and inherit from it 8 times seemed kind of overkill for the benefit, especially from the maintenance point of view, so I chose this solution. I would like to hear other opinions

I don't feel strongly about this as users do not create request objects directly so at least the confusion at creation time is avoided. However, even with the nice javadoc you added it might be confusing to have many `null` fields in `Request`. Let's wait for more opinion.

> Made serializable for now.

Here is my point on this being Serializable. I thing there's a benefit for this class but I also see a benefit in `DnsBatch` being [Restorable](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-core/src/main/java/com/google/gcloud/Restorable.java#L39). However this should not come at the price of making `Callback` serializable (which is too restrictive for users). Thoughts? /cc @aozarov @ajkannan 

> With hashCode and equals, I did not override them on purpose.

This is a good point. And I guess this means that operation in a batch are executed in order, right? If we keep this `Serializable` we must have proper `hashCode` and `equals` methods. Maybe wait for the discussion on `Serializable`/`Restorable` to settle but I would consider a list of pairs (we are not using the "Hash" part of LinkedHashMap now, right?). Also using two separate lists would do no harm.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
mziccard: Your reasoning is sound. My point is that I would like to see all the `DnsBatch` -> `BatchRequest` conversion happen in the same place but as you outlined this is probably not attainable. Fine as is for me, thanks for taking the time to explain your reasoning.

mderka:
We decided to go for a different concept. Closing.

PR #774, opened 2016-03-22
Rename startPageToken to pageToken in Storage and BigQuery

### LINKED ISSUE #759: Both pageToken and startPageToken are used as list options
As noticed by @mderka we are using `pageToken` in `resourcemanager` while `startPageToken` in `storage` and `bigquery`. We should probably pick a name and stick to it.

My feeling is that `pageToken` is already clear enough so we can go with it and remove all `startPageToken`. What do you guys think?

mziccard:
Fixes #759


ajkannan:
LGTM

PR #776, opened 2016-03-22
Adding bundling support for PublisherApi.publish

garrettjonesgoogle:


gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
```diff
@@ -223,7 +232,7 @@ protected PublisherSettings(MethodBuilders methods) {
   }
 
   /**
-   * Returns the ApiCallableBuilder for the API method createTopic.
+   * Returns the builder for the API method createTopic.
```
aozarov: Did I mention that I think you should consider renaming `ApiCallableBuilder` to `Builder`? As a nested class I think
its context is clear and such naming convention is very common ([e.g.](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableList.java#L612)).

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/SubscriberApi.java
```diff
@@ -66,9 +65,26 @@
  */
 @javax.annotation.Generated("by GAPIC")
 public class SubscriberApi implements AutoCloseable {
+  // ========
```
aozarov: Per earlier discussion (I don't remember the status). Are these comments really necessary for the code-generation part? (This is not a common practice).

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
garrettjonesgoogle: I don't think you mentioned that yet. I'll go ahead and make that change in GAX. Would you like to see that in this PR, or is a following PR ok?

aozarov: following PR sounds good to me.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
```diff
@@ -337,4 +346,66 @@ public Object extractNextToken(ListTopicSubscriptionsResponse payload) {
...
+
+        @Override
+        public void splitResponse(
+            PublishResponse bundleResponse,
+            Collection<? extends RequestIssuer<PublishRequest, PublishResponse>> bundle) {
+          int bundleMessageIndex = 0;
+          for (RequestIssuer<PublishRequest, PublishResponse> responder : bundle) {
+            List<String> subresponseElements = new ArrayList<>();
+            int subresponseCount = responder.getRequest().getMessagesCount();
+            for (int i = 0; i < subresponseCount; i++) {
```
aozarov: Any reason for not using for each loop? Is response.getRequest().getMessageCount() can be less
than the size of bundleResponse.getMessageIdsList()?

Also (nit pick), you can Initialize subresponseElements to the expected size if you already know it.

aozarov:
Looks good. Feel free to merge unless you want to apply any of the suggested changes in this PR.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
garrettjonesgoogle: Any time more than 1 PublishRequest is bundled, response.getRequest().getMessageCount() will be less than bundleResponse.getMessageIdsList(). So if PublishRequest A has M messages and PublishRequest B has N messages, then response.getRequest().getMessageCount() will return M and then N.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/SubscriberApi.java
garrettjonesgoogle: I don't recall an objection to them before. They make it less likely for there to be change conflicts that go across multiple sections.

aozarov: Less likely by the code-generator? Again, feel free to look at other common java libraries (guava, and JDK included).
This is by far not a common practice.

garrettjonesgoogle: Less likely by the 3-way merge process.

They are really just a holdover from the coding style of the first prototype; I'll see what the team thinks and possibly remove them in the next version (which will have other style fixes).

PR #777, opened 2016-03-22
The second concept of the DNS batch

mderka:
I added a concept based on our discussion this morning. This is to solicit some feedback.

So far, I created the `BatchResult` interface in a separate package in DNS, but we said that we would make it part of the core. I will make this happen for the actual PR. I also did not implement all the test, just the integration one to demonstrate the functionality:

``` java
DnsBatch batch = DNS.batch();
BatchResult<Page<Zone>, DnsException> batchResult = batch.listZones();
batch.submit();
assertTrue(batchResult.submitted());
Iterator<Zone> iteratorBatch = batchResult.get().iterateAll();
```

Lastly, the `BatchResult` interface does not have the `notify()` method as I am not entirely clear on its meaning. I will add it after some clarifications.

Please, let me know if you like this better and we want to move forward with this concept.

/cc @ajkannan @aozarov @mziccard

### DIFF
--- a/gcloud-java-dns/src/main/java/com/google/gcloud/dns/BatchResultImpl.java
+++ b/gcloud-java-dns/src/main/java/com/google/gcloud/dns/BatchResultImpl.java
@@ -1,0 +1,44 @@
+package com.google.gcloud.dns;
+
+import com.google.gcloud.dns.batch.BatchResult;
+
+/**
+ * This class holds a single result of a batch call to the Cloud DNS.
+ */
+public class BatchResultImpl<T> implements BatchResult<T, DnsException> {
+
+  private T result;
+  private boolean submitted = false;
+  private DnsException error;
+
+  BatchResultImpl() {
+  }
+
+  @Override
+  public boolean submitted() {
+    return submitted;
+  }
+
+  @Override
+  public T get() throws DnsException {
+    if(!submitted()) {
+      throw new IllegalStateException("Batch has not been submitted yet");
+    }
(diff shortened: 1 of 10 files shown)


gcloud-java-dns/src/main/java/com/google/gcloud/dns/BatchResultImpl.java
```diff
@@ -0,0 +1,44 @@
+package com.google.gcloud.dns;
+
+import com.google.gcloud.dns.batch.BatchResult;
+
+/**
+ * This class holds a single result of a batch call to the Cloud DNS.
+ */
+public class BatchResultImpl<T> implements BatchResult<T, DnsException> {
```
aozarov: As you said, this should we should have the interface or base class in core, however it may make
sense to subclass it in DNS (to avoid the need to specify the generic type by the user).

gcloud-java-dns/src/main/java/com/google/gcloud/dns/BatchResultImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    if(!submitted()) {
+      throw new IllegalStateException("Batch has not been submitted yet");
+    }
+    if(error != null) {
+      throw error;
+    }
+    return result;
+  }
+
+  void result(T result) {
```
aozarov: those will need to be protected, so can be used by subclasses (see my comment above) in the other service packages.

Issue #732, opened 2016-03-09
Image's deprecationStatus can hold wrongly formatted timestamps


mziccard:
When a Compute Image is [deprecated](https://cloud.google.com/compute/docs/reference/latest/images/deprecate) a [DeprecationStatus](https://cloud.google.com/compute/docs/reference/latest/images/deprecate#request_properties_JSON) must be provided in the request body. Part of deprecation status are 3 timestamps (deprecated, obsolete, deleted) that are documented to be:

> An optional RFC3339 timestamp on or after which the deprecation state of this resource will be changed to DEPRECATED/OBSOLETE/DELETED.

However, the service does no validation and any string can be submitted for such fields. In fact older images (e.g. `backports-debian-7-wheezy-v20131127`) have non-RFC3339 timestamp values:

```JSON
"deprecated": {
  "state": "DEPRECATED",
  "replacement": "https://content.googleapis.com/compute/v1/projects/debian-cloud/global/images/backports-debian-7-wheezy-v20131127",
  "deprecated": "1970-01-01",
  "obsolete": "1970-01-02",
  "deleted": "1970-01-03"
}
```

mziccard:
Checks have been now added to ensure that timestamps are valid RFC3339 timestamps upon deprecation. However, older deprecated images might still have wrongly-formatted timestamps.

To consider this closed we should at least not expose timestamp setters that take a string. However we still need to store timestamps as strings to support older images that might have wrongly-formatted timestamps.

PR #794, opened 2016-03-26
Add ImageInfo, ImageId and test classes

mziccard:
This PR adds `ImageId`, `ImageInfo` and related test classes.
Images can be deprecated using [deprecate](https://cloud.google.com/compute/docs/reference/latest/images/deprecate) by passing a `DeprecationStatus` object.
This PR also refactors `DeprecationStatus` to add a public builder.

Following issue #732 this PR also makes `DeprecationStatus` more tolerant in parsing timestamps (also partial timestamps are supported now).
However as per #732 no checks are made on `deleted`, `deprecated` and `obsolete` timestamps provided to [deprecate](https://cloud.google.com/compute/docs/reference/latest/images/deprecate). If a user misused such fields, creating a `DeprecationStatus` might fail. Should we make `deleted`, `deprecated` and `obsolete` of `String` type? Or keep them of `Long` type and set them to `null` if the string value is not parsable?

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
@@ -16,6 +16,8 @@
 
 package com.google.gcloud.compute;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 
@@ -28,16 +30,17 @@ import java.util.Objects;
 /**
  * The deprecation status associated to a Google Compute Engine resource.
  *
- * @param <T> The Google Compute Engine resource to which the deprecation status refers to.
+ * @param <T> The Google Compute Engine resource identity to which the deprecation status refers
  */
 public final class DeprecationStatus<T extends ResourceId> implements Serializable {
 
   private static final long serialVersionUID = -2695077634793679794L;
   private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
+  private static final DateTimeFormatter TIMESTAMP_PARSER = ISODateTimeFormat.dateTimeParser();
 
-  private final Long deleted;
-  private final Long deprecated;
-  private final Long obsolete;
+  private final String deleted;
+  private final String deprecated;
+  private final String obsolete;
   private final T replacement;
   private final Status status;
 
@@ -64,36 +67,201 @@ public final class DeprecationStatus<T extends ResourceId> implements Serializab
     DELETED
   }
 
-  DeprecationStatus(Long deleted, Long deprecated, Long obsolete, T replacement, Status status) {
-    this.deleted = deleted;
-    this.deprecated = deprecated;
-    this.obsolete = obsolete;
-    this.replacement = replacement;
-    this.status = status;
+  /**
+   * A builder for {@code DeprecationStatus} objects.
+   *
+   * @param <T> The Google Compute Engine resource identity to which the deprecation status refers
+   */
+  public static final class Builder<T extends ResourceId> {
+
+    private String deleted;
+    private String deprecated;
+    private String obsolete;
+    private T replacement;
+    private Status status;
+
+    Builder() {}
+
+    Builder(DeprecationStatus<T> deprecationStatus) {
+      this.deleted = deprecationStatus.deleted;
+      this.deprecated = deprecationStatus.deprecated;
+      this.obsolete = deprecationStatus.obsolete;
+      this.replacement = deprecationStatus.replacement;
+      this.status = deprecationStatus.status;
+    }
+
+    /**
+     * Sets the timestamp on or after which the deprecation state of this resource will be changed
+     * to {@link Status#DELETED}. Timestamp should be in RFC3339 format.
+     *
+     * @see <a href="http://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
+     */
+    // todo(mziccard): remove this method if #732 is closed
+    public Builder<T> deleted(String deleted) {
+      this.deleted = deleted;
+      return this;
+    }
+
+    /**
+     * Sets the timestamp on or after which the deprecation state of this resource will be changed
+     * to {@link Status#DEPRECATED}. Timestamp should be in RFC3339 format.
+     *
+     * @see <a href="http://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
+     */
+    // todo(mziccard): remove this method if #732 is closed
+    public Builder<T> deprecated(String deprecated) {
+      this.deprecated = deprecated;
+      return this;
+    }
+
+    /**
+     * Sets the timestamp on or after which the deprecation state of this resource will be changed
+     * to {@link Status#OBSOLETE}. Timestamp should be in RFC3339 format.
+     *
+     * @see <a href="http://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
+     */
+    // todo(mziccard): remove this method if #732 is closed
+    public Builder<T> obsolete(String obsolete) {
+      this.obsolete = obsolete;
+      return this;
+    }
+
+    /**
+     * Sets the timestamp on or after which the deprecation state of this resource will be changed
+     * to {@link Status#DELETED}. In milliseconds since epoch.
+     */
+    public Builder<T> deleted(long deleted) {
+      this.deleted = TIMESTAMP_FORMATTER.print(deleted);
+      return this;
+    }
+
+    /**
+     * Sets the timestamp on or after which the deprecation state of this resource will be changed
+     * to {@link Status#DEPRECATED}. In milliseconds since epoch.
+     */
+    public Builder<T> deprecated(long deprecated) {
+      this.deprecated = TIMESTAMP_FORMATTER.print(deprecated);
+      return this;
+    }
+
+    /**
+     * Sets the timestamp on or after which the deprecation state of this resource will be changed
+     * to {@link Status#OBSOLETE}. In milliseconds since epoch.
+     */
+    public Builder<T> obsolete(long obsolete) {
+      this.obsolete = TIMESTAMP_FORMATTER.print(obsolete);
+      return this;
+    }
+
+    /**
+     * Sets the identity of the suggested replacement for a deprecated resource. The suggested
+     * replacement resource must be the same kind of resource as the deprecated resource.
+     */
+    public Builder<T> replacement(T replacement) {
+      this.replacement = replacement;
+      return this;
+    }
+
+    /**
+     * Sets the status of the deprecated resource.
+     */
+    public Builder<T> status(Status status) {
+      this.status = checkNotNull(status);
+      return this;
+    }
+
+    /**
+     * Creates a {@code DeprecationStatus} object.
+     */
+    public DeprecationStatus<T> build() {
+      return new DeprecationStatus<T>(this);
+    }
+  }
+
+  DeprecationStatus(Builder<T> builder) {
+    this.deleted = builder.deleted;
+    this.deprecated = builder.deprecated;
+    this.obsolete = builder.obsolete;
+    this.replacement = builder.replacement;
+    this.status = checkNotNull(builder.status);
   }
 
   /**
    * Returns the timestamp on or after which the deprecation state of this resource will be changed
-   * to {@link Status#DELETED}. In milliseconds since epoch.
+   * to {@link Status#DELETED}. Returns {@code null} if not set. This value should be in RFC3339
+   * format.
+   *
+   * @see <a href="http://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
    */
-  public Long deleted() {
+  // todo(mziccard): remove this method if #732 is closed
+  public String deleted() {
     return deleted;
   }
 
   /**
    * Returns the timestamp on or after which the deprecation state of this resource will be changed
-   * to {@link Status#DEPRECATED}. In milliseconds since epoch.
+   * to {@link Status#DEPRECATED}. Returns {@code null} if not set. This value should be in RFC3339
+   * format.
+   *
+   * @see <a href="http://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
    */
-  public Long deprecated() {
+  // todo(mziccard): remove this method if #732 is closed
+  public String deprecated() {
     return deprecated;
   }
 
   /**
    * Returns the timestamp on or after which the deprecation state of this resource will be changed
-   * to {@link Status#OBSOLETE}. In milliseconds since epoch.
+   * to {@link Status#OBSOLETE}. Returns {@code null} if not set. This value should be in RFC3339
+   * format.
+   *
+   * @see <a href="http://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
    */
-  public Long obsolete() {
+  // todo(mziccard): remove this method if #732 is closed
+  public String obsolete() {
     return obsolete;
+  }
+
+  /**
+   * Returns the timestamp (in milliseconds since epoch) on or after which the deprecation state of
+   * this resource will be changed to {@link Status#DELETED}. Returns {@code null} if not set.
+   *
+   * @throws IllegalStateException if {@link #deleted()} is not a valid date, time or datetime
+   */
+  public Long deletedMillis() {
+    try {
+      return deleted != null ? TIMESTAMP_PARSER.parseMillis(deleted) : null;
+    } catch (IllegalArgumentException ex) {
+      throw new IllegalStateException(ex.getMessage(), ex);
+    }
+  }
+
+  /**
+   * Returns the timestamp (in milliseconds since epoch) on or after which the deprecation state of
+   * this resource will be changed to {@link Status#DEPRECATED}. Returns {@code null} if not set.
+   *
+   * @throws IllegalStateException if {@link #deprecated()} is not a valid date, time or datetime
+   */
+  public Long deprecatedMillis() {
+    try {
+      return deprecated != null ? TIMESTAMP_PARSER.parseMillis(deprecated) : null;
+    } catch (IllegalArgumentException ex) {
+      throw new IllegalStateException(ex.getMessage(), ex);
+    }
+  }
+
+  /**
+   * Returns the timestamp (in milliseconds since epoch) on or after which the deprecation state of
+   * this resource will be changed to {@link Status#OBSOLETE}. Returns {@code null} if not set.
+   *
+   * @throws IllegalStateException if {@link #obsolete()} is not a valid date, time or datetime
+   */
+  public Long obsoleteMillis() {
+    try {
+      return obsolete != null ? TIMESTAMP_PARSER.parseMillis(obsolete) : null;
+    } catch (IllegalArgumentException ex) {
+      throw new IllegalStateException(ex.getMessage(), ex);
+    }
   }
 
   /**
@@ -109,6 +277,13 @@ public final class DeprecationStatus<T extends ResourceId> implements Serializab
    */
   public Status status() {
     return status;
+  }
+
+  /**
+   * Returns a builder for the {@code DeprecationStatus} object.
+   */
+  public Builder<T> toBuilder() {
+    return new Builder<>(this);
   }
 
   @Override
@@ -136,37 +311,49 @@ public final class DeprecationStatus<T extends ResourceId> implements Serializab
   com.google.api.services.compute.model.DeprecationStatus toPb() {
     com.google.api.services.compute.model.DeprecationStatus deprecationStatusPb =
         new com.google.api.services.compute.model.DeprecationStatus();
-    if (deleted != null) {
-      deprecationStatusPb.setDeleted(TIMESTAMP_FORMATTER.print(deleted));
-    }
-    if (deprecated != null) {
-      deprecationStatusPb.setDeprecated(TIMESTAMP_FORMATTER.print(deprecated));
-    }
-    if (obsolete != null) {
-      deprecationStatusPb.setObsolete(TIMESTAMP_FORMATTER.print(obsolete));
-    }
-    if (replacement != null) {
-      deprecationStatusPb.setReplacement(replacement.selfLink());
-    }
-    if (status() != null) {
-      deprecationStatusPb.setState(status.name());
-    }
+    deprecationStatusPb.setDeleted(deleted);
+    deprecationStatusPb.setDeprecated(deprecated);
+    deprecationStatusPb.setObsolete(obsolete);
+    deprecationStatusPb.setReplacement(replacement.selfLink());
+    deprecationStatusPb.setState(status.name());
     return deprecationStatusPb;
+  }
+
+  /**
+   * Returns the builder for a {@code DeprecationStatus} object given the status.
+   */
+  public static <T extends ResourceId> Builder<T> builder(Status status) {
+    return new Builder<T>().status(status);
+  }
+
+  /**
+   * Returns the builder for a {@code DeprecationStatus} object given the status and replacement's
+   * identity.
+   */
+  public static <T extends ResourceId> Builder<T> builder(Status status, T replacement) {
+    return new Builder<T>().status(status).replacement(replacement);
+  }
+
+  /**
+   * Returns a {@code DeprecationStatus} object given the status and replacement's identity.
+   */
+  public static <T extends ResourceId> DeprecationStatus<T> of(Status status, T replacement) {
+    return builder(status, replacement).build();
   }
 
   static <T extends ResourceId> DeprecationStatus<T> fromPb(
       com.google.api.services.compute.model.DeprecationStatus deprecationStatusPb,
       Function<String, T> fromUrl) {
-    return new DeprecationStatus<>(
-        deprecationStatusPb.getDeleted() != null
-            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getDeleted()) : null,
-        deprecationStatusPb.getDeprecated() != null
-            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getDeprecated()) : null,
-        deprecationStatusPb.getObsolete() != null
-            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getObsolete()) : null,
-        deprecationStatusPb.getReplacement() != null
-            ? fromUrl.apply(deprecationStatusPb.getReplacement()) : null,
-        deprecationStatusPb.getState() != null
-            ? Status.valueOf(deprecationStatusPb.getState()) : null);
+    Builder<T> builder = new Builder<>();
+    builder.deleted(deprecationStatusPb.getDeleted());
+    builder.deprecated(deprecationStatusPb.getDeprecated());
+    builder.obsolete(deprecationStatusPb.getObsolete());
+    if (deprecationStatusPb.getReplacement() != null) {
+      builder.replacement(fromUrl.apply(deprecationStatusPb.getReplacement()));
+    }
+    if (deprecationStatusPb.getState() != null) {
+      builder.status(Status.valueOf(deprecationStatusPb.getState()));
+    }
+    return builder.build();
   }
 }
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gcloud.compute;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.api.services.compute.model.Image;
+import com.google.common.base.MoreObjects;
+
+import java.util.Objects;
+
+/**
+ * A Google Compute Engine disk image configuration. This class can be used to create images from an
+ * existing Google Compute Engine disk.
+ */
+public class DiskImageConfiguration extends ImageConfiguration {
+
+  private static final long serialVersionUID = 2716403667042981170L;
+
+  private final DiskId sourceDisk;
+  private final String sourceDiskId;
+
+  /**
+   * A builder for {@code DiskImageConfiguration} objects.
+   */
+  public static final class Builder
+      extends ImageConfiguration.Builder<DiskImageConfiguration, Builder> {
+
+    private DiskId sourceDisk;
+    private String sourceDiskId;
+
+    private Builder() {
+      super(Type.DISK);
+    }
+
+    private Builder(DiskImageConfiguration imageConfiguration) {
+      super(Type.DISK, imageConfiguration);
+      this.sourceDisk = imageConfiguration.sourceDisk;
+      this.sourceDiskId = imageConfiguration.sourceDiskId;
+    }
+
+    private Builder(Image imagePb) {
+      super(Type.DISK, imagePb);
+      this.sourceDisk = DiskId.fromUrl(imagePb.getSourceDisk());
+      this.sourceDiskId = imagePb.getSourceDiskId();
+    }
+
+    /**
+     * Sets the identity of the source disk used to create the image.
+     */
+    public Builder sourceDisk(DiskId sourceDisk) {
+      this.sourceDisk = checkNotNull(sourceDisk);
+      return this;
+    }
+
+    Builder sourceDiskId(String sourceDiskId) {
+      this.sourceDiskId = sourceDiskId;
+      return this;
+    }
+
+    /**
+     * Creates a {@code DiskImageConfiguration} object.
+     */
+    @Override
+    public DiskImageConfiguration build() {
+      return new DiskImageConfiguration(this);
+    }
+  }
+
+  private DiskImageConfiguration(Builder builder) {
+    super(builder);
+    this.sourceDisk = checkNotNull(builder.sourceDisk);
+    this.sourceDiskId = builder.sourceDiskId;
+  }
+
+  /**
+   * Returns the identity of the source disk used to create this image.
+   */
+  public DiskId sourceDisk() {
+    return sourceDisk;
+  }
+
+  /**
+   * Returns the id of the disk used to create this image. This value may be used to determine
+   * whether the image was taken from the current or a previous instance of a given disk name.
+   */
+  public String sourceDiskId() {
+    return sourceDiskId;
+  }
+
+  @Override
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  @Override
+  MoreObjects.ToStringHelper toStringHelper() {
+    return super.toStringHelper()
+        .add("sourceDisk", sourceDisk)
+        .add("sourceDiskId", sourceDiskId);
+  }
+
+  @Override
+  public final int hashCode() {
+    return Objects.hash(baseHashCode(), sourceDisk, sourceDiskId);
+  }
+
+  @Override
+  public final boolean equals(Object obj) {
+    return obj instanceof DiskImageConfiguration && baseEquals((DiskImageConfiguration) obj);
+  }
+
+  @Override
+  DiskImageConfiguration setProjectId(String projectId) {
+    if (sourceDisk.project() != null) {
+      return this;
+    }
+    return toBuilder().sourceDisk(sourceDisk.setProjectId(projectId)).build();
+  }
+
+  @Override
+  Image toPb() {
+    Image imagePb = super.toPb();
+    imagePb.setSourceDisk(sourceDisk.selfLink());
+    imagePb.setSourceDiskId(sourceDiskId);
+    return imagePb;
+  }
+
+  /**
+   * Creates a builder for a {@code DiskImageConfiguration} given the source disk identity.
+   */
+  public static Builder builder(DiskId sourceDisk) {
+    return new Builder().sourceDisk(sourceDisk);
+  }
+
+  /**
+   * Creates a {@code DiskImageConfiguration} object given the source disk identity.
+   */
+  public static DiskImageConfiguration of(DiskId sourceId) {
+    return builder(sourceId).build();
+  }
+
+  @SuppressWarnings("unchecked")
+  static DiskImageConfiguration fromPb(Image imagePb) {
+    return new Builder(imagePb).build();
+  }
+}
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gcloud.compute;
+
+import com.google.api.services.compute.model.Image;
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Base class for Google Compute Engine image configuration. Use {@link DiskImageConfiguration} to
+ * create an image from an existing Google Compute Engine disk. Use
+ * {@link StorageImageConfiguration} to create an image from a file stored in Google Cloud Storage.
+ */
+public abstract class ImageConfiguration implements Serializable {
+
+  private static final long serialVersionUID = -9154332316597745316L;
+
+  private final Type type;
+  private final SourceType sourceType;
+  private final Long archiveSizeBytes;
+
+  /**
+   * Type of a Google Compute Engine image.
+   */
+  public enum Type {
+    /**
+     * A Google Compute Engine image created from a Google Compute Engine disk.
+     */
+    DISK,
+
+    /**
+     * A Google Compute Engine image created from a file saved in Google Cloud Storage.
+     */
+    STORAGE
+  }
+
+  /**
+   * Image source type. The only admissible value is {@code RAW}.
+   */
+  public enum SourceType {
+    RAW
+  }
+
+  /**
+   * Base builder for image configurations.
+   *
+   * @param <T> the image configuration class
+   * @param <B> the image configuration builder
+   */
+  public abstract static class Builder<T extends ImageConfiguration, B extends Builder<T, B>> {
+
+    private Type type;
+    private SourceType sourceType;
+    private Long archiveSizeBytes;
+
+    Builder(Type type) {
+      this.type = type;
+    }
+
+    Builder(Type type, ImageConfiguration imageConfiguration) {
+      this.type = type;
+      this.sourceType = imageConfiguration.sourceType;
+      this.archiveSizeBytes = imageConfiguration.archiveSizeBytes;
+    }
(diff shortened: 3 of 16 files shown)


gcloud-java-compute/src/test/java/com/google/gcloud/compute/RawImageConfigurationTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+import com.google.gcloud.compute.ImageConfiguration.SourceType;
+
+import org.junit.Test;
+
+public class RawImageConfigurationTest {
+
+  private static final String SOURCE = "source";
+  private static final SourceType SOURCE_TYPE = SourceType.RAW;
+  private static RawImageConfiguration CONFIGURATION = RawImageConfiguration.builder(SOURCE)
```
mziccard: Codacy Issue found: [The field name indicates a constant but its modifiers do not](https://www.codacy.com/app/mziccard/gcloud-java/file/2314353647/issues/source?bid=3140968&fileBranchId=3200309#l32)

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskImageConfigurationTest.java
```diff
@@ -0,0 +1,102 @@
...
+import com.google.gcloud.compute.ImageConfiguration.SourceType;
+
+import org.junit.Test;
+
+public class DiskImageConfigurationTest {
+
+  private static final DiskId SOURCE_DISK =  DiskId.of("project", "zone", "disk");
+  private static final String SOURCE_DISK_ID = "diskId";
+  private static final SourceType SOURCE_TYPE = SourceType.RAW;
+  private static DiskImageConfiguration CONFIGURATION = DiskImageConfiguration.builder(SOURCE_DISK)
```
mziccard: Codacy Issue found: [The field name indicates a constant but its modifiers do not](https://www.codacy.com/app/mziccard/gcloud-java/file/2314353634/issues/source?bid=3140968&fileBranchId=3200309#l32)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
```diff
@@ -0,0 +1,416 @@
...
+    private Long creationTimestamp;
+    private ImageId imageId;
+    private String description;
+    private ImageConfiguration configuration;
+    private Status status;
+    private Long diskSizeGb;
+    private List<LicenseId> licenses;
+    private DeprecationStatus<ImageId> deprecationStatus;
+
+    BuilderImpl() {}
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2314353601/issues/source?bid=3140968&fileBranchId=3200309#l149)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -64,12 +67,77 @@
...
+   *     to
+   */
+  public static final class Builder<T extends ResourceId> {
+
+    private Long deleted;
+    private Long deprecated;
+    private Long obsolete;
+    private T replacement;
+    private Status status;
+
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2314353590/issues/source?bid=3140968&fileBranchId=3200309#l84)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+   * Base builder for image configurations.
+   *
+   * @param <T> the image configuration class
+   * @param <B> the image configuration builder
+   */
+  public abstract static class Builder<T extends ImageConfiguration, B extends Builder<T, B>> {
+
+    private SourceType sourceType;
+
+    Builder() {}
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2314353599/issues/source?bid=3140968&fileBranchId=3200309#l53)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -64,12 +67,77 @@
...
+    /**
+     * Sets the identity of the suggested replacement for a deprecated resource. The suggested
+     * replacement resource must be the same kind of resource as the deprecated resource.
+     */
+    public Builder<T> replacement(T replacement) {
+      this.replacement = replacement;
+      return this;
+    }
+
+    public Builder<T> status(Status status) {
```
ajkannan: Is this meant to be package scope?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -28,12 +30,13 @@
 /**
  * The deprecation status associated to a Google Compute Engine resource.
  *
- * @param <T> The Google Compute Engine resource to which the deprecation status refers to.
+ * @param <T> The Google Compute Engine resource identity to which the deprecation status refers to
```
ajkannan: Nit: can remove the "to" at the end of the line b/c you have "to which" already.  Also applies to the Builder down below.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
ajkannan: Actually, I see it's supposed to be public.  We could add javadoc to this method.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+     */
+    public abstract T build();
+  }
+
+  ImageConfiguration(Builder builder) {
+    this.sourceType = builder.sourceType;
+  }
+
+  /**
+   * Returns the source type of the disk. The default and only value is {@link SourceType#RAW}
```
ajkannan: Nit: could add "." at the end of this comment.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
```diff
@@ -0,0 +1,416 @@
...
+  private static final long serialVersionUID = -1061916352807358977L;
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
+
+  private final String id;
+  private final ImageId imageId;
+  private final Long creationTimestamp;
+  private final String description;
+  private final ImageConfiguration configuration;
+  private final Status status;
+  private final Long diskSizeGb;
```
ajkannan: Is [archiveSizeBytes](https://cloud.google.com/compute/docs/reference/latest/images#archiveSizeBytes) something we care about, or was it deliberately left out?

ajkannan: Ah, I see it's placed in RawImageConfiguration -- ignore the previous comment.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+      if (imagePb.getRawDisk().getContainerType() != null) {
+        this.containerType = ContainerType.valueOf(imagePb.getRawDisk().getContainerType());
+      }
+      this.sha1Checksum = imagePb.getRawDisk().getSha1Checksum();
+      this.source = imagePb.getRawDisk().getSource();
+      this.archiveSizeBytes = imagePb.getArchiveSizeBytes();
+    }
+
+    /**
+     * Sets the format used to encode and transmit the block device. Supported value is only
```
ajkannan: Nit: I would reword the second sentence as "The only supported value is {@code TAR}"

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+     * {@code TAR}. This is just a container and transmission format, not a runtime format.
+     */
+    public Builder containerType(ContainerType containerType) {
```
ajkannan: The cloud docs leave me a bit uncertain as to whether the user has to set this, is encouraged to set this, or if this field is optional.  Do you happen to know whether the user has to set this from experimentation?

mziccard: In fact there's no need to set this now. I left it public as it is settable and setting this value does not affect the result of the operation. And my guess is that this more containers types will be added in the future (ZIP for instance?)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /**
+   * A builder for {@code RawImageConfiguration} objects.
+   */
+  public static final class Builder
+      extends ImageConfiguration.Builder<RawImageConfiguration, Builder> {
+
+    private ContainerType containerType;
+    private String sha1Checksum;
```
ajkannan: Nit: I feel like we could shorten this to sha1 without any loss of clarity (just my opinion)

gcloud-java-compute/src/test/java/com/google/gcloud/compute/SerializationTest.java
```diff
@@ -105,12 +105,8 @@
       .maintenanceWindows(WINDOWS)
       .region(REGION_ID)
       .build();
-  private static final Long DELETED = 1453293540000L;
-  private static final Long DEPRECATED = 1453293420000L;
-  private static final Long OBSOLETE = 1453293480000L;
   private static final DeprecationStatus<MachineTypeId> DEPRECATION_STATUS =
-      new DeprecationStatus<>(DELETED, DEPRECATED, OBSOLETE, MACHINE_TYPE_ID,
-          DeprecationStatus.Status.DELETED);
+      DeprecationStatus.of(DeprecationStatus.Status.DELETED, MACHINE_TYPE_ID);
```
ajkannan: New classes in this PR that implement Serializable (e.g. `ImageInfo`) can be added to this file.

ajkannan:
Question about the deprecation status: the dates a user inputs in those fields don't cause any behavioral changes on the server, do they?  If a user passes in valid dates in the proper format, does the service move the deprecation status automatically for the user from `DEPRECATED` to `OBSOLETE`, for example?  Alternatively, does the user manually have to send a request to update the deprecation status?  I imagine it's the latter because the service would encounter errors if the service doesn't input-validate these fields.

Would the following proposal work?
* Provide getters for the 3 timestamp fields as strings (and represent them internally as strings) so that the user can see the values.  Also this way the user won't unset the fields if they were set to non-parseable (but potentially informative) fields .
* Provide (or document) a method to convert the strings to longs.  We could also provide getters for each as longs, but this might be a lot of methods.
* Only provide setters for those 3 fields as longs.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskImageConfigurationTest.java
```diff
@@ -0,0 +1,102 @@
...
+      .sourceDiskId(SOURCE_DISK_ID)
+      .sourceType(SOURCE_TYPE)
+      .build();
+
+  @Test
+  public void testToBuilder() {
+    compareDiskImageConfiguration(CONFIGURATION, CONFIGURATION.toBuilder().build());
+    DiskId newDisk = DiskId.of("newProjet", "newZone", "newDisk");
```
ajkannan: Nit/doesn't really matter, but s/newProjet/newProject/

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -154,19 +222,47 @@ public boolean equals(Object obj) {
     return deprecationStatusPb;
   }
 
+  /**
+   * Returns the builder for a {@code DeprecationStatus} object given the status.
+   */
+  public static <T extends ResourceId> Builder<T> builder(Status status) {
```
ajkannan: Do you think it'd also be valuable to have a `toBuilder()`?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/RawImageConfigurationTest.java
ajkannan: I think this is worth fixing (can add "final")

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskImageConfigurationTest.java
ajkannan: Worth fixing, I think you can add "final" here

gcloud-java-compute/src/test/java/com/google/gcloud/compute/RawImageConfigurationTest.java
```diff
@@ -0,0 +1,86 @@
...
+  @Test
+  public void testSetProjectId() {
+    assertSame(CONFIGURATION, CONFIGURATION.setProjectId("project"));
+  }
+
+  private void compareRawImageConfiguration(RawImageConfiguration expected,
+      RawImageConfiguration value) {
+    assertEquals(expected, value);
+    assertEquals(expected.source(), value.source());
+    assertEquals(expected.sourceType(), value.sourceType());
```
ajkannan: Could also assert things specific to a `RawImageConfiguration`, like containerType, sha1Checksum, and archiveSizeBytes.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/RawImageConfigurationTest.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.gcloud.compute.ImageConfiguration.SourceType;
+
+import org.junit.Test;
+
+public class RawImageConfigurationTest {
+
+  private static final String SOURCE = "source";
+  private static final SourceType SOURCE_TYPE = SourceType.RAW;
+  private static RawImageConfiguration CONFIGURATION = RawImageConfiguration.builder(SOURCE)
+      .sourceType(SOURCE_TYPE)
```
ajkannan: Could set 3 fields specific to raw image config class; would make the builder test below stronger.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/RawImageConfigurationTest.java
```diff
@@ -0,0 +1,86 @@
...
+    compareRawImageConfiguration(CONFIGURATION,
+        ImageConfiguration.<RawImageConfiguration>fromPb(CONFIGURATION.toPb()));
+    RawImageConfiguration configuration = RawImageConfiguration.of(SOURCE);
+    compareRawImageConfiguration(configuration, RawImageConfiguration.fromPb(configuration.toPb()));
+  }
+
+  @Test
+  public void testOf() {
+    RawImageConfiguration configuration = RawImageConfiguration.of(SOURCE);
+    assertNull(configuration.sourceType());
```
ajkannan: Can assert other raw image fields are null

mziccard:
> The dates a user inputs in those fields don't cause any behavioral changes on the server, do they?

Dates have no impact on the service. I think that what you suggest makes sense. I would also add a getter for each timestamp that returns the long timestamp:

```java
public String deprecated();
public Long deprecatedTimestamp();
```

I agree that setters should only take long. I also think that getters for `Long` values should be permissive and parse even partial timestamps.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+        this.sourceType = SourceType.valueOf(imagePb.getSourceType());
+      }
+    }
+
+    @SuppressWarnings("unchecked")
+    B self() {
+      return (B) this;
+    }
+
+    B sourceType(SourceType sourceType) {
```
ajkannan: Is this something the user will need to set (should it be public or is it meant to be package scope?)

aozarov:
>> I agree that setters should only take long. I also think that getters for Long values should be permissive and parse even partial timestamps.

If I understand the issue correctly the timestamp in this case is merely a label (open text supplied by the user) with a recommended convention for the value.

If so, honestly I am not sure why the service API even bothered with proving a specific setting for it (as oppose to providing a way to specify any set of attributes) and in this case would not bother in trying
to parse it. I think the sometimes work sometimes does not is not ideal.

ajkannan:
@mziccard Is the input validation an issue for deprecation statuses for other types of compute resources, or just for images?

@aozarov's point about having methods that sometimes work and sometimes don't makes me rethink my proposal and instead side with your string proposal.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+      this.sha1Checksum = sha1Checksum;
+      return this;
+    }
+
+    /**
+     * Sets the full Google Cloud Storage URL where the disk image is stored (e.g.
+     * {@code gs://bucket/file}).
+     */
+    public Builder source(String source) {
+      this.source = source;
```
ajkannan: Should we checkNotNull the source?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  @Override
+  public final int hashCode() {
+    return Objects.hash(baseHashCode(), source, containerType, sha1Checksum, archiveSizeBytes);
+  }
+
+  @Override
+  public final boolean equals(Object obj) {
+    return obj instanceof RawImageConfiguration && baseEquals((RawImageConfiguration) obj);
```
ajkannan: Should the fields specific to this class also be checked for equality?

ajkannan: (Using this subclass' fields in hashCode() but not in equals() would break the contract that two equal objects' hashcodes are equal)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  @Override
+  public final int hashCode() {
+    return Objects.hash(baseHashCode(), sourceDisk, sourceDiskId);
+  }
+
+  @Override
+  public final boolean equals(Object obj) {
+    return obj instanceof DiskImageConfiguration && baseEquals((DiskImageConfiguration) obj);
```
ajkannan: Check equality of sourceDisk/sourceDiskId?

ajkannan:
Done with my pass.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
ajkannan: Actually, this is fine, I forgot that the baseEquals method calls toPb()

mziccard:
> If I understand the issue correctly the timestamp in this case is merely a label (open text supplied by the user) with a recommended convention for the value.

Yes but I don't think this was intended but rather a missing check on the service's side. If you look at the docs for [image.deprecate](https://cloud.google.com/compute/docs/reference/latest/images/deprecate#deprecated) it seems to me that the RFC format is not intended to be just a suggestion. `DeprecationStatus` is also used for disk types, machine types, regions and zones and there all timestamps have the expected format.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -64,12 +67,87 @@
...
+   * @param <T> The Google Compute Engine resource identity to which the deprecation status refers
+   */
+  public static final class Builder<T extends ResourceId> {
+
+    private Long deleted;
+    private Long deprecated;
+    private Long obsolete;
+    private T replacement;
+    private Status status;
+
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2320385269/issues/source?bid=3140968&fileBranchId=3200309#l83)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2320385278/issues/source?bid=3140968&fileBranchId=3200309#l53)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2320385280/issues/source?bid=3140968&fileBranchId=3200309#l149)

mziccard:
I fixed all comments but the ones on `DeprecationStatus` timestamps as discussion is still open

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -179,6 +189,13 @@ public Status status() {
     return status;
   }
 
+  /**
+   * Returns a builder for the {@code DeprecationStatus} object.
+   */
+  public Builder<T> toBuilder() {
```
ajkannan: Could add test for toBuilder in `DeprecationStatusTest`

mziccard:
I add String setters/getters to `DeprecationStatus` and updated the tests. Let me know how you like it!

aozarov:
Looks like compute tests fail on Travis.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -64,39 +67,192 @@
...
+  // todo(mziccard): remove this method if #732 is closed
+  public String obsolete() {
     return obsolete;
   }
 
   /**
+   * Returns the timestamp (in milliseconds since epoch) on or after which the deprecation state of
+   * this resource will be changed to {@link Status#DELETED}. Returns {@code null} if not set.
+   *
+   * @throws IllegalArgumentException if {@link #deleted()} is not a valid date, time or datetime
```
aozarov: `IllegalStateException` ?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -136,37 +299,55 @@ public boolean equals(Object obj) {
...
-            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getDeprecated()) : null,
-        deprecationStatusPb.getObsolete() != null
-            ? TIMESTAMP_FORMATTER.parseMillis(deprecationStatusPb.getObsolete()) : null,
-        deprecationStatusPb.getReplacement() != null
-            ? fromUrl.apply(deprecationStatusPb.getReplacement()) : null,
-        deprecationStatusPb.getState() != null
-            ? Status.valueOf(deprecationStatusPb.getState()) : null);
+    Builder<T> builder = new Builder<T>();
+    if (deprecationStatusPb.getDeleted() != null) {
+      builder.deleted(TIMESTAMP_PARSER.parseMillis(deprecationStatusPb.getDeleted()));
```
aozarov: Now that we have it, any reason for not using the string version.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  @Override
+  DiskImageConfiguration setProjectId(String projectId) {
+    Builder builder = toBuilder().sourceDisk(sourceDisk.setProjectId(projectId));
```
aozarov: Not worth checking if sourceDisk has already projectId set?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
mziccard: I am throwing what [parseMillis](http://joda-time.sourceforge.net/apidocs/org/joda/time/format/DateTimeFormatter.html#parseMillis(java.lang.String)) throws. Do you think we should catch and rethrow `IllegalStateException`?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
mziccard: Oh sure I missed it.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
aozarov: Yes, I realized that. I think it is strange to throw an `IllegalArgumentException` for a method that does not accept an argument and in such cases `IllegalStateException` is commonly used. It is a nit, but catch and re-throw would be my preference but will not object if you think otherwise.

mziccard: I think you are right, I'll fix this.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+   */
+  public DiskId sourceDisk() {
+    return sourceDisk;
+  }
+
+  /**
+   * Returns the id of the disk used to create this image. This value may be used to determine
+   * whether the image was taken from the current or a previous instance of a given disk name.
+   */
+  public String sourceDiskId() {
```
aozarov: This is the first case that managed to confuse me with DiskId (as type) and diskId as string.

mziccard: We will have the very same problem for other configurations coming in next PRs (disk configuration). Any suggestion on naming is appreciated.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private static final long serialVersionUID = -9154332316597745316L;
+
+  private final SourceType sourceType;
+
+  /**
+   * Image source type. The only admissible value is {@code RAW}.
+   */
+  public enum SourceType {
+    RAW
```
aozarov: I couldn't find one in the compute site (but didn't look too carefully). Is there anything that we can say for the RAW type?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Base class for Google Compute Engine image configuration. Use {@link DiskImageConfiguration} to
+ * create an image from an existing Google Compute Engine disk. Use {@link RawImageConfiguration} to
+ * create an image from a file stored in Google Cloud Storage.
+ */
+public abstract class ImageConfiguration implements Serializable {
```
aozarov: Do you think we need to have an ImageType enum (so instanceof may not be the only option)?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private Builder(Image imagePb) {
+      super(imagePb);
+      if (imagePb.getRawDisk().getContainerType() != null) {
+        this.containerType = ContainerType.valueOf(imagePb.getRawDisk().getContainerType());
+      }
+      this.sha1 = imagePb.getRawDisk().getSha1Checksum();
+      this.source = imagePb.getRawDisk().getSource();
+      this.archiveSizeBytes = imagePb.getArchiveSizeBytes();
```
aozarov: Any reason this was not part of the RawDisk in the proto?

mziccard: I checked and in fact this field is set also for image created from an existing disk (i.e. with DiskImageConfiguration). Weird that it's documented as: 

> Size of the image tar.gz archive stored in Google Cloud Storage (in bytes).

That made me think that it only applied when the disk was created from a source in Storage.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
mziccard: I think this would help but my fear is that it might be confusing with `sourceType`. Also how should we call the tokens? `RAW` and `DISK`? `RAW` is confusing as it's also the name of the only value for `sourceType`. Maybe something like `DISK` and `STORAGE`? Then should we also rename the `RawImageConfiguration` to `StorageImageConfiguration`?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
mziccard: I couldn't find any information either, also the name is quite confusing to me.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/RawImageConfiguration.java
aozarov: I think Ajay has a point. we should probably make the `baseEquals` check `getClass()` equality just in case that the toPb() equality (based on map equality) is true (if all values are the same). Because the subclasses may have their own specific *required* fields that many not be an issue now but I see no reason why we should make it work regardless.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
aozarov: So far (and the reason I was confused) I thought we used something like `DiskId sourceDiskId()` and `String sourceId()`. no?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
aozarov: +1 for your renaming suggestions (though a little detached from the underlying values they are more more clear).

aozarov:
Looks good. I think there few comments we need to address before we can merge.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2323737960/issues/source?bid=3140968&fileBranchId=3200309#l149)

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DeprecationStatusTest.java
```diff
@@ -17,40 +17,138 @@
...
+    DeprecationStatus<MachineTypeId> deprecationStatus =
+        DeprecationStatus.builder(STATUS, MACHINE_TYPE_ID)
+            .deprecated("deprecated")
+            .obsolete("obsolete")
+            .deleted("delete")
+            .build();
+    assertEquals("deprecated", deprecationStatus.deprecated());
+    try {
+      deprecationStatus.deprecatedMillis();
+      fail("Expected IllegalArgumentException");
```
mziccard: Codacy Issue found: [In JUnit4, use the @Test(expected) annotation to denote tests that should throw exceptions](https://www.codacy.com/app/mziccard/gcloud-java/file/2323737991/issues/source?bid=3140968&fileBranchId=3200309#l103)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageConfiguration.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2323737958/issues/source?bid=3140968&fileBranchId=3200309#l53)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -64,39 +67,192 @@
...
+   * @param <T> The Google Compute Engine resource identity to which the deprecation status refers
+   */
+  public static final class Builder<T extends ResourceId> {
+
+    private String deleted;
+    private String deprecated;
+    private String obsolete;
+    private T replacement;
+    private Status status;
+
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2323737949/issues/source?bid=3140968&fileBranchId=3200309#l83)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
mziccard: I think the convention I used so far (here and in other services) is to remove the `Id` suffix from the name of fields with `XxxId` type. So I think in this case the code follows that convention.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DeprecationStatusTest.java
```diff
@@ -17,40 +17,138 @@
...
+    try {
+      deprecationStatus.obsoleteMillis();
+      fail("Expected IllegalArgumentException");
+    } catch (IllegalStateException ex) {
+      // never reached
+    }
+    assertEquals("delete", deprecationStatus.deleted());
+    try {
+      deprecationStatus.deletedMillis();
+      fail("Expected IllegalArgumentException");
```
mziccard: Codacy Issue found: [In JUnit4, use the @Test(expected) annotation to denote tests that should throw exceptions](https://www.codacy.com/app/mziccard/gcloud-java/file/2325228367/issues/source?bid=3140968&fileBranchId=3200309#l117)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
```diff
@@ -64,39 +67,204 @@
...
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2325228325/issues/source?bid=3140968&fileBranchId=3200309#l83)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
```diff
@@ -0,0 +1,416 @@
...
+    private Long creationTimestamp;
+    private ImageId imageId;
+    private String description;
+    private ImageConfiguration configuration;
+    private Status status;
+    private Long diskSizeGb;
+    private List<LicenseId> licenses;
+    private DeprecationStatus<ImageId> deprecationStatus;
+
+    BuilderImpl() {}
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2325228336/issues/source?bid=3140968&fileBranchId=3200309#l149)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2325228325/issues/source?bid=3140968&fileBranchId=3200309#l83)

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DeprecationStatusTest.java
mziccard: Codacy Issue found: [In JUnit4, use the @Test(expected) annotation to denote tests that should throw exceptions](https://www.codacy.com/app/mziccard/gcloud-java/file/2325228367/issues/source?bid=3140968&fileBranchId=3200309#l117)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2325228336/issues/source?bid=3140968&fileBranchId=3200309#l149)

mziccard:
Comments should be fixed now (but the one concerning `sourceDisk` and `sourceDiskId`):
- `DeprecationStatus` timestamp getters now throw `IllegalStateException`
- Move `archiveSizeBytes` to `ImageConfiguration`
- Add `Type` enum to `ImageConfiguration` and type field
- Rename `RawImageConfiguration` to `ComputeImageConfiguration`
- Update tests
- Add `getClass()` equality check to `ImageConfiguration.baseEquals`

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskImageConfiguration.java
aozarov: I mean `BlobId blobId() vs `String id()` in Blob, ` JobId jobId()` and `String id()` in BigQuery Job, `TableId tableId()` and `String id()` in BigQuery TableInfo,... Are we talking about two different things or at some point changed style?

mziccard: Those are two different conventions we used, in my opinion. In `BlobInfo` we have `id` that is part of the original proto and we added `blobId` (with `Id` suffix) to indicate the identity object. This is also due to the fact that having `blob()` getter in `BlobInfo` and `Blob` looked ugly. For all classes where no name clashes were possible I think we used simply `blob`, `targetBlob` or `sourceBlob`.

If you think that `DiskId sourceDiskId()` and `String sourceId()` are easier to understand I am happy to change. Personally I don't think they are but I am not against.

ajkannan: I see, I think @mziccard uses `[class]Id()` and `id()` in [class].java, but anywhere there's a field that points to the identity of another class, he uses `[class]` instead of `[classId]`, probably for the sake of brevity.  In this case, what @mziccard is consistent with other usage.

This again makes me wonder whether we can come up with an alternative to better differentiate the two fields.  Perhaps instead of calling the string id `id`, maybe we could say something bit more verbose to differentiate it, like "serviceGeneratedId" (maybe not the best name, but I couldn't think of a better one after a couple minutes of thought).  The string id used less often in our code, and our users probably shouldn't be depending on it all that much, so more verbosity in my mind is okay.

aozarov: Oh, I am not sure why we deviated. I think because we have confusing terms complete/strict consistency with method name to expected return type is important. And therefore keeping `xxxId()` to always refer/return `XxxId`  and `id` to always refer and return the string generated id.

I think we should keep what you have as is *but* like @ajkannan I think we should find a way to distinct the 2 by property name. I think `serviceGeneratedId` is too long and would vote for `genratedId` (javadoc should mention that
this is a service generated id for the resource).

mziccard: `generatedId` is ok with me, also `opaqueId` might make sense though opaque is maybe a more obscure word.

I'll open an issue for this.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
aozarov: Thanks. We should also update the javadoc.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
```diff
@@ -393,7 +393,7 @@ Image toPb() {
   /**
    * Returns a builder for an {@code ImageInfo} object given the image identity and an image
    * configuration. Use {@link DiskImageConfiguration} to create an image from an existing disk. Use
-   * {@link RawImageConfiguration} to create an image from a raw image stored in Google Cloud
+   * {@link StorageImageConfiguration} to create an image from a raw image stored in Google Cloud
```
aozarov: Fine as is but I do wonder if we should really say "from a raw image". I assume the "raw" part depends on the `sourceType` which indeed `RAW` is its only option but was not sure if that needs to slip into here.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/StorageImageConfiguration.java
```diff
@@ -27,14 +27,13 @@
  * A Google Compute Engine raw image configuration. This class can be used to create images from a
```
aozarov: Related to the comment above. Is raw refers to the sourceType and if so is that guarantied to be fixed?
If the answers are yes and no then when we add a new enum value to the `SourceType` we would also need
to cleanup the javadoc...

aozarov:
Looks good! Few javadoc related fixes (IlegalArgumentException) and question/suggestions.
Feel free to squash and merge after you fix them.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2330195970/issues/source?bid=3140968&fileBranchId=3200309#l83)

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DeprecationStatusTest.java
mziccard: Codacy Issue found: [In JUnit4, use the @Test(expected) annotation to denote tests that should throw exceptions](https://www.codacy.com/app/mziccard/gcloud-java/file/2330196012/issues/source?bid=3140968&fileBranchId=3200309#l117)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
```diff
@@ -0,0 +1,416 @@
...
+    private Long creationTimestamp;
+    private ImageId imageId;
+    private String description;
+    private ImageConfiguration configuration;
+    private Status status;
+    private Long diskSizeGb;
+    private List<LicenseId> licenses;
+    private DeprecationStatus<ImageId> deprecationStatus;
+
+    BuilderImpl() {}
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2330195981/issues/source?bid=3140968&fileBranchId=3200309#l149)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
mziccard: I agree this might be confusing. I am updating the javadoc here and elsewhere to mention the Storage file, rather than the "raw" concept.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DeprecationStatusTest.java
mziccard: Codacy Issue found: [In JUnit4, use the @Test(expected) annotation to denote tests that should throw exceptions](https://www.codacy.com/app/mziccard/gcloud-java/file/2330220789/issues/source?bid=3140968&fileBranchId=3200309#l117)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DeprecationStatus.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2330220747/issues/source?bid=3140968&fileBranchId=3200309#l83)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageInfo.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2330220758/issues/source?bid=3140968&fileBranchId=3200309#l149)

PR #796, opened 2016-03-29
Update with latest surface fixes based Java export review.

shinfan:
Refactorings:
- Remove innerClass and move util methods to top level.
- Rename ApiCallableBuilder to just Builder
- Rename XPath into XName
- Rename XMethod into XCallSettings in XSettings classe

Improvements:
- Let the user decide whether to auto-close provided channel
- Make all generated methods final
- Use Duration instead of long to avoid time unit confusion
- Support bundling limits configuration

Others:
- Update unit tests
- Update pom and dependency

### DIFF
--- a/gcloud-java-pubsub/pom.xml
+++ b/gcloud-java-pubsub/pom.xml
@@ -10,7 +10,7 @@
   <parent>
     <groupId>com.google.gcloud</groupId>
     <artifactId>gcloud-java-pom</artifactId>
-    <version>0.1.4-SNAPSHOT</version>
+    <version>0.1.5</version>
   </parent>
   <properties>
     <site.installationModule>gcloud-java-pubsub</site.installationModule>


aozarov:
Hey Shin per earlier comment do you mind to update this branch from master or at least update travis config so I can tell that this build is not broken?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -84,97 +80,80 @@
...
-     * <!-- end manual edit -->
-     */
-    public static final String formatProjectPath(String project) {
-      return PROJECT_PATH_TEMPLATE.instantiate("project", project);
-    }
+  /**
+   * A PathTemplate representing the fully-qualified path to represent
+   * a project resource.
+   *
+   * <!-- manual edit -->
```
aozarov: Do you see any reason for providing this '` for a 'private' field?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -84,97 +80,80 @@
...
-      return PROJECT_PATH_TEMPLATE.parse(projectPath).get("project");
-    }
+  /**
+   * Formats a string containing the fully-qualified path to represent
+   * a project resource.
+   *
+   * <!-- manual edit -->
+   * <!-- end manual edit -->
+   */
+  public static final String formatProjectName(String project) {
```
aozarov: Do you see any reason for providing this '` for a 'final' method?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -84,97 +80,80 @@
...
+    return PROJECT_PATH_TEMPLATE.instantiate("project", project);
```
aozarov: did you consider `format`? [e.g.](https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html#format(java.lang.String,%20java.lang.Object...)).

garrettjonesgoogle: The path template string that we get as input, which is language independent and relates to the definition of the API in the proto, supports a different set of things than what MessageFormat supports: https://github.com/googleapis/gax-java/blob/master/src/main/java/com/google/api/gax/protobuf/PathTemplate.java

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
garrettjonesgoogle: The manual edit section is for adding documentation that is relevant to this Java entry point that the code generator can't add in the general case. This is independent of whether the method is final or not.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
garrettjonesgoogle: Well, in fact, we probably don't need a comment on this at all.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
aozarov: No doubt, but don't you think `format` would be more appropriate (and consistent with the parse as well as your own javadoc "Formats a string containing the...."?

PR #800, opened 2016-03-29
Update with latest surface fixes based Java export review.

shinfan:
Refactorings:
- Remove section headers from comments
- Remove innerClass and move util methods to top level.
- Rename ApiCallableBuilder to just Builder
- Rename XPath into XName
- Rename XMethod into XCallSettings in XSettings classe

Improvements:
- Let the user decide whether to auto-close provided channel
- Make all generated methods final
- Use Duration instead of long to avoid time unit confusion
- Support bundling limits configuration

Others:
- Update unit tests
- Update pom and dependency

### DIFF
--- a/gcloud-java-pubsub/pom.xml
+++ b/gcloud-java-pubsub/pom.xml
@@ -26,7 +26,7 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-pubsub-v1</artifactId>
-      <version>0.0.1</version>
+      <version>0.0.0</version>
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>


aozarov:
@shinfan gcloud-java is supporting Java 7 and above. It looks like you gax and/or the grpc-pubsub-v1 were built using Java 8 (and using the 1.8 target) which breaks this build.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -84,97 +80,80 @@
...
-     * <!-- end manual edit -->
-     */
-    public static final String formatProjectPath(String project) {
-      return PROJECT_PATH_TEMPLATE.instantiate("project", project);
-    }
+  /**
+   * A PathTemplate representing the fully-qualified path to represent
+   * a project resource.
+   *
+   * <!-- manual edit -->
```
aozarov: Did you consider my comments from #796

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
```diff
@@ -123,17 +122,17 @@
...
                 BackoffParams.newBuilder()
-                    .setInitialDelayMillis(300L)
-                    .setDelayMultiplier(1.3)
-                    .setMaxDelayMillis(3000L)
+                    .setInitialDelay(Duration.millis(2000L))
+                    .setDelayMultiplier(1.5)
+                    .setMaxDelay(Duration.millis(30000L))
                     .build())
-            .setTotalTimeout(30000L)
+            .setTotalTimeout(Duration.millis(45000L))
```
aozarov: Are you sure you want to expose Joda in your API surface?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
```diff
@@ -123,17 +122,17 @@
...
                     .setDelayMultiplier(1.2)
-                    .setMaxDelayMillis(1000L)
+                    .setMaxDelay(Duration.millis(1000L))
                     .build())
             .setTimeoutBackoff(
                 BackoffParams.newBuilder()
-                    .setInitialDelayMillis(300L)
-                    .setDelayMultiplier(1.3)
-                    .setMaxDelayMillis(3000L)
+                    .setInitialDelay(Duration.millis(2000L))
```
aozarov: Honestly, I am not sure how "back-off" and "max-delay" relates to timeout. Do you mean
initial time before timeout happens is 200L after that further tries are multiplied by 1.5 but will
not exceed 30000L? If so, though I understand that settings may feel similar to retry backoff
the method names do not.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
```diff
@@ -142,45 +141,54 @@
...
+    private final ApiCallable.Builder<DeleteTopicRequest, Empty> deleteTopicMethod;
     private final ImmutableList<? extends ApiCallSettings> allMethods;
 
     public MethodBuilders() {
-      createTopicMethod = new ApiCallableBuilder<>(PublisherGrpc.METHOD_CREATE_TOPIC);
+      createTopicMethod = new Builder<>(PublisherGrpc.METHOD_CREATE_TOPIC);
       createTopicMethod.setRetryableCodes(RETRYABLE_CODE_DEFINITIONS.get("idempotent"));
       createTopicMethod.setRetryParams(RETRY_PARAM_DEFINITIONS.get("default"));
 
+      BundlingSettings publishBundlingSettings =
```
aozarov: I question the value of the bundling for createTopic method. If any, I would say getTopic is more likely to benefit from it.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/SubscriberApi.java
```diff
@@ -573,7 +550,7 @@ public void modifyAckDeadline(ModifyAckDeadlineRequest request) {
    * @param ackIds The acknowledgment ID for the messages being acknowledged that was returned
    * by the Pub/Sub system in the `Pull` response. Must not be empty.
    */
-  public void acknowledge(String subscription, List<String> ackIds) {
+  public final void acknowledge(String subscription, List<String> ackIds) {
```
aozarov: Did you consider `Iterable<String>`?

gcloud-java-pubsub/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -66,10 +66,6 @@
  */
 @javax.annotation.Generated("by GAPIC")
 public class PublisherApi implements AutoCloseable {
```
aozarov: Any updates on considering using a binary file for baseline?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/SubscriberApi.java
shinfan: Can you specify the reason why we should choose Iterable over List here? Currently we always use List for paramenters across the files so it would be good to keep the consistency. And I do feel List sometimes is more flexible to use than Iterable.

gcloud-java-pubsub/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
shinfan: Garrett maybe a better person than me to answer this question. @garrettjonesgoogle

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
garrettjonesgoogle: But createTopic is not bundled... are you misreading the code?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/SubscriberApi.java
aozarov: see "Effective Java" rule 40. If order does not matter and if implementation does not really need or use List specific methods why would you restrict the possible input for your method?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
shinfan: Yes, we have maximum and initial value for timeout. The initial timeout is 200L in first trial and then be multiplied by 1.5 in the second trial and so on. We use the |BackoffParams| for such settings for both retry and timeout.
I understand where the confusion comes from and I agree that we may need to reconsider the method names to fit both cases.  https://github.com/GoogleCloudPlatform/gcloud-java/issues/814

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
aozarov: Yes, it does look like I misread the code.

shinfan: Oh okay nvm then :8ball:

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/SubscriberApi.java
shinfan: I see, but I am not sure if we can make every such parameter Iterable type. The List type is auto- determined by looking at "repeated" field in the proto files during code generation. So it is hard to make some cases "Iterable" and the others "List". And we do not know if the order matters or not.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
shinfan: I am not sure if we should relate "editable" with method modifiers. Ideally users could edit everything in the file and their manual change could be preserved through three-way merge.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
shinfan: Using Duration instead of long can avoid potential issues of time units (seconds/milliseconds)

garrettjonesgoogle: FYI an internal Java expert affirmed this usage and said he would have recommended it (over using a 'long') if we weren't using it already. What would be the problem with exposing it?

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherApi.java
shinfan: Issue created: 
https://github.com/GoogleCloudPlatform/gcloud-java/issues/815

aozarov: Ok, I also [replied](https://github.com/GoogleCloudPlatform/gcloud-java/pull/796#discussion_r57816065) for the other issue I mentioned.

gcloud-java-pubsub/baseline/src/main/java/com/google/gcloud/pubsub/spi/v1/PublisherSettings.java
aozarov: 1. In general I prefer to avoid making client of API X to be familiar with API Y (that X is using). Internal usage is totally fine.

2. Java 8 has supports for Duration (which will remove the need to use Joda). I didn't check if their semantics is identical or very close.

For these 2 reasons we chose to use this signature instead (int time, TimeUnit unit). (`TimeUnit` is part of the Java SDK). Later we can add (or replace) with Java 8 specifics.

Fine as is, wanted to voice the reasoning and style we use.

PR #802, opened 2016-03-29
update pom.xml to include DNS in javadoc groups

aozarov:


ajkannan:
LGTM.  Tomorrow I will add a package-info for the testing package so that the description matches the others and also document that javadoc bundling needs to be handled when releasing a new module.

Issue #808, opened 2016-03-29
Investigate flaky DNS integration test


ajkannan:
[This build](https://travis-ci.org/GoogleCloudPlatform/gcloud-java/builds/119298316) saw the integration test pass in two modules and fail in the third with the following issue:

```
Running com.google.gcloud.dns.it.ITDnsTest
Tests run: 12, Failures: 1, Errors: 1, Skipped: 0, Time elapsed: 114.627 sec <<< FAILURE! - in com.google.gcloud.dns.it.ITDnsTest
testInvalidChangeRequest(com.google.gcloud.dns.it.ITDnsTest)  Time elapsed: 24.69 sec  <<< ERROR!
com.google.gcloud.dns.DnsException: The 'entity.change.deletions[0]' resource named 'subdomain.gcldjvit-65ccc5a3-5403-47f3-94ee.com. (A)' does not exist.
	at com.google.gcloud.dns.it.ITDnsTest.testInvalidChangeRequest(ITDnsTest.java:652)
Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 
404 Not Found
{
  "code" : 404,
  "errors" : [ {
    "domain" : "global",
    "message" : "The 'entity.change.deletions[0]' resource named 'subdomain.gcldjvit-65ccc5a3-5403-47f3-94ee.com. (A)' does not exist.",
    "reason" : "notFound"
  } ],
  "message" : "The 'entity.change.deletions[0]' resource named 'subdomain.gcldjvit-65ccc5a3-5403-47f3-94ee.com. (A)' does not exist."
}
	at com.google.gcloud.dns.it.ITDnsTest.testInvalidChangeRequest(ITDnsTest.java:652)
testListZones(com.google.gcloud.dns.it.ITDnsTest)  Time elapsed: 0.686 sec  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<1>
	at com.google.gcloud.dns.it.ITDnsTest.testListZones(ITDnsTest.java:379)
Results :
Failed tests: 
  ITDnsTest.testListZones:379 expected:<0> but was:<1>
Tests in error: 
  ITDnsTest.testInvalidChangeRequest:652 » Dns The 'entity.change.deletions[0]' ...
Tests run: 12, Failures: 1, Errors: 1, Skipped: 0
```

ajkannan:
[...]

PR #822, opened 2016-03-30
Check if record added in ITDnsTest

ajkannan:
Rationale mentioned in #808

### DIFF
--- a/gcloud-java-dns/src/test/java/com/google/gcloud/dns/it/ITDnsTest.java
+++ b/gcloud-java-dns/src/test/java/com/google/gcloud/dns/it/ITDnsTest.java
@@ -650,10 +650,11 @@ public class ITDnsTest {
         assertEquals(400, ex.code());
       }
     } finally {
-      assertTrue(recordAdded);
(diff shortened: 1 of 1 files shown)


gcloud-java-dns/src/test/java/com/google/gcloud/dns/it/ITDnsTest.java
```diff
@@ -648,6 +650,7 @@ public void testInvalidChangeRequest() {
         assertEquals(400, ex.code());
       }
     } finally {
+      assertTrue(recordAdded);
```
aozarov: I was thinking more of:
```java
if (recordAdded) {
  ChangeRequestInfo deletion = ChangeRequest.builder().delete(validA).build();
  ChangeRequest request = zone.applyChangeRequest(deletion);
  waitForChangeToComplete(zone.name(), request.id());
}
zone.delete();
```

to allow cleanup while, hopefully, propagating the original error.

ajkannan: fixed

PR #825, opened 2016-03-30
Update v1beta3 dependencies

ajkannan:


gcloud-java-datastore/pom.xml
```diff
@@ -34,12 +34,12 @@
     <dependency>
       <groupId>com.google.cloud.datastore</groupId>
       <artifactId>datastore-v1beta3-protos</artifactId>
-      <version>0.0.1-SNAPSHOT</version>
+      <version>1.0.0-beta-SNAPSHOT</version>
```
mziccard: This is consistent with what we did before but any particular reason to rely on the snapshot version? We should remember to change this once we merge into master.

ajkannan: Before, the artifact was only available from the sonatype snapshot repository and not the maven central repository.  I've modified it and rebased.

PR #826, opened 2016-03-31
Add functional methods for images and Image class

mziccard:
As for snapshots integration tests will come as soon as we have disks.

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
@@ -1832,9 +1832,14 @@ public interface Compute extends Service<ComputeOptions> {
   Image get(ImageId imageId, ImageOption... options);
 
   /**
-   * Lists images in the provided project that are available to the current user.
+   * Lists images in the provided project that are available to the current user. This method can be
+   * used to list publicly-available images by providing the respective image project. Examples of
+   * image projects are: {@code centos-cloud}, {@code coreos-cloud}, {@code debian-cloud},
+   * {@code opensuse-cloud}, {@code rhel-cloud}, {@code suse-cloud}, {@code ubuntu-os-cloud} and
+   * {@code windows-cloud}. Attempting to delete or deprecate a publicly-available image will fail.
    *
    * @throws ComputeException upon failure
+   * @see <a href="https://cloud.google.com/compute/docs/operating-systems/">Operating Systems</a>
    */
   Page<Image> listImages(String project, ImageListOption... options);
 
@@ -1850,7 +1855,7 @@ public interface Compute extends Service<ComputeOptions> {
    *
    * @return a global operation if the delete request was issued correctly, {@code null} if the
    *     image was not found
-   * @throws ComputeException upon failure
+   * @throws ComputeException upon failure or if {@code image} is a publicly-available image
    */
   Operation delete(ImageId image, OperationOption... options);
 
@@ -1859,7 +1864,7 @@ public interface Compute extends Service<ComputeOptions> {
    *
    * @return a global operation if the deprecation request was issued correctly, {@code null} if the
    *     image was not found
-   * @throws ComputeException upon failure
+   * @throws ComputeException upon failure or if {@code image} is a publicly-available image
    */
   Operation deprecate(ImageId image, DeprecationStatus<ImageId> deprecationStatus,
       OperationOption... options);
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
@@ -1124,14 +1124,15 @@ final class ComputeImpl extends BaseService<ComputeOptions> implements Compute {
   }
 
   @Override
-  public Operation delete(final ImageId image, OperationOption... options) {
+  public Operation delete(ImageId image, OperationOption... options) {
+    final ImageId completeId = image.setProjectId(options().projectId());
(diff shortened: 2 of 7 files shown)


gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -1699,4 +1815,52 @@ public static SnapshotListOption fields(SnapshotField... fields) {
...
+   * @throws ComputeException upon failure
+   */
+  Image get(ImageId imageId, ImageOption... options);
+
+  /**
+   * Lists images in the provided project that are available to the current user.
+   *
+   * @throws ComputeException upon failure
+   */
+  Page<Image> listImages(String project, ImageListOption... options);
```
aozarov: According to [this](https://cloud.google.com/compute/docs/images#public_images) private images
should be imported to the current project and public images can only be created and maintained by Google.

What I think is more useful is to do what "gcloud compute images list" is doing and return all the publicly available images. 

If that is not doable (provided by the API), we should at least point to a doc  that lists all the Google maintained projects that holds the public images (centos-cloud, debian-cloud, ...).

Also, do you think in this case we should return a functional Image where half of the operations will
fail and the other half are doubtedly useful?

gcloud-java-compute/src/main/java/com/google/gcloud/spi/DefaultComputeRpc.java
```diff
@@ -564,6 +567,72 @@ public Operation deleteSnapshot(String snapshot, Map<Option, ?> options) {
...
+    } catch (IOException ex) {
+      throw translate(ex);
+    }
+  }
+
+  @Override
+  public Operation deleteImage(String image, Map<Option, ?> options) {
+    try {
+      return compute.images()
+          .delete(this.options.projectId(), image)
```
aozarov: This could be a problem if we try to delete a public-image which the same name as the image in this project. No?

aozarov:
Looks good. Few questions/comments regarding public images.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
mziccard: > What I think is more useful is to do what "gcloud compute images list" is doing and return all the publicly available images.

There is no such API. The only listing API for images requires a `projectId` to be passed.

> Also, do you think in this case we should return a functional Image where half of the operations will fail and the other half are doubtedly useful?

There is no easy way to distinguish between private and public images as `list(project)` could be also used passing the current project or a project that the user owns. The only possible check that comes to my mind is throwing when `image.imageId().project() != options.projectId()`, this would allow to keep `delete` and `deprecate` methods as is. However, this prevents (once `delete` and `deprecate` are fixed) the user from deleting image that he actually has rights to delete, only because they are in a different project. Thoughts?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -870,6 +870,54 @@ public static SnapshotFilter notEquals(SnapshotField field, long value) {
...
+    }
+
+    /**
+     * Returns an equals filter for the given field and string value. For string fields,
+     * {@code value} is interpreted as a regular expression using RE2 syntax. {@code value} must
+     * match the entire field.
+     *
+     * @see <a href="https://github.com/google/re2/wiki/Syntax">RE2</a>
+     */
+    public static ImageFilter equals(ImageField field, String value) {
```
mziccard: Codacy Issue found: [The method name and parameter number are suspiciously close to equals(Object)](https://www.codacy.com/app/mziccard/gcloud-java/file/2341145921/issues/source?bid=3140968&fileBranchId=3207740#l890)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/Compute.java
```diff
@@ -870,6 +870,54 @@ public static SnapshotFilter notEquals(SnapshotField field, long value) {
...
+     * @see <a href="https://github.com/google/re2/wiki/Syntax">RE2</a>
+     */
+    public static ImageFilter notEquals(ImageField field, String value) {
+      return new ImageFilter(checkNotNull(field), ComparisonOperator.NE, checkNotNull(value));
+    }
+
+    /**
+     * Returns an equals filter for the given field and long value.
+     */
+    public static ImageFilter equals(ImageField field, long value) {
```
mziccard: Codacy Issue found: [The method name and parameter number are suspiciously close to equals(Object)](https://www.codacy.com/app/mziccard/gcloud-java/file/2341145921/issues/source?bid=3140968&fileBranchId=3207740#l908)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ComputeImpl.java
```diff
@@ -1026,6 +1047,120 @@ public Operation deleteSnapshot(final String snapshot, OperationOption... option
...
+  public Page<Image> listImages(String project, ImageListOption... options) {
+    return listImages(project, options(), optionMap(options));
+  }
+
+  @Override
+  public Page<Image> listImages(ImageListOption... options) {
+    return listImages(options().projectId(), options(), optionMap(options));
+  }
+
+  private static Page<Image> listImages(final String project, final ComputeOptions serviceOptions,
```
mziccard: Codacy Issue found: [Avoid unused private methods such as 'listImages(String,ComputeOptions,Map)'.](https://www.codacy.com/app/mziccard/gcloud-java/file/2341145924/issues/source?bid=3140968&fileBranchId=3207740#l1097)

mziccard:
@aozarov fixed both comments according to our discussion

Issue #827, opened 2016-03-31
Wrong default project ID is chosen when active configuration is not the default or service keys are used


tswast:
# Wrong gcloud command configuration is loaded

## Problem

My `~/.config/gcloud/configurations/config_default` contents:

```
[core]
account = ***REDACTED***@gmail.com
project = gcpnext2016-swast
[compute]
zone = us-central1-a
region = us-central1
```

My active configuration:

```
$ gcloud config list
Your active configuration is: [swast-test-flexible]
[app]
suppress_change_warning = true
[compute]
region = us-central1
zone = us-central1-b
[core]
account = swast@google.com
disable_usage_reporting = False
project = swast-test-flexible
[metrics]
command_name = gcloud.config.list
```

When I run the [Bookshelf](https://cloud.google.com/java/getting-started/tutorial-app) application, which uses Application Default Credentials locally with `mvn -Plocal clean jetty:run-exploded` and I connect to http://localhost:8080, I see data from my `gcpnext2016-swast` project, not the `swast-test-flexible` project like I expect.

## Steps to reproduce

1. Run the [Bookshelf](https://cloud.google.com/java/getting-started/tutorial-app) application, which uses Application Default Credentials locally with `mvn -Plocal clean jetty:run-exploded` and I connect to http://localhost:8080
1. Add a few books.
1. Run `gcloud init`.
1. Create a new configuration instead of modifying the default, and select a different project this time.
1. Run the Bookshelf application again.
1. Notice that you still see the same data from the project selected in the default configuration, not the new configuration that you just activated.

## Suspected root cause

gcloud-java uses default configuration rather than the currently selected one to figure out which project to send RPCs to. https://github.com/GoogleCloudPlatform/gcloud-java/blob/d03e5a7d80dc17d9012476557015e4f44be64687/gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java#L394

# Project ID is not extracted from service keys

## Problem
 
When I use a service key, with the Bookshelf app command

```
GOOGLE_APPLICATION_CREDENTIALS=$HOME/src/service-keys/swast-test-flexible.json mvn -Plocal clean jetty:run-exploded
```

I get an exception when my default gcloud command configuration has a different project ID than the service key.

```
Caused by: com.google.gcloud.datastore.DatastoreException: Unauthorized.
```

## Steps to reproduce

1. Create a JSON service key for a project (a different project than what the default gcloud command configuration is set to).
[...]

PR #831, opened 2016-03-31
Get projectId from gcloud active config

mziccard:
This addresses the first part of #827: "Wrong gcloud command configuration is loaded"

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java
@@ -27,6 +27,7 @@ import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.auth.http.HttpCredentialsAdapter;
 import com.google.common.collect.Iterables;
+import com.google.common.io.Files;
 import com.google.gcloud.spi.ServiceRpcFactory;
 
 import java.io.BufferedReader;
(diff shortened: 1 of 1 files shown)


gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java
```diff
@@ -381,6 +383,15 @@ protected String defaultProject() {
     return projectId != null ? projectId : googleCloudProjectId();
   }
 
+  private static String activeGoogleCloudConfig(File configDir) {
+    try {
+      return Files.readFirstLine(new File(configDir, "active_config"), Charset.defaultCharset());
```
aozarov: Is it possible for this file to be empty (and therefore return `null`)?

mziccard: When you have only `default` config this file contains `default`. This always points to the active config.

aozarov: Are you saying that we need no worry about the case of an empty file?
Clearly file could be modified/truncated by users (though maybe that is not much of a concern?)
Personally I would prefer to be slightly more resilient in that case (and fallback to config_default rather than config_null) however it is a Nit and feel free to merge if you think we should not care about it.

mziccard:
@aozarov Now the code fallbacks to `default` if the file is empty.

Issue #824, opened 2016-03-30
Better naming for resource identities and service ids


mziccard:
Our `XxxInfo` classes have two fields with a similar name
- `xxxId`: this fields is of type `XxxId` and holds a user-defined identity for the resource
- `id`: this fields holds a service-generated (opaque) id

We should provide a better naming to easily distinguish the two fields. Possible solutions are:
- renaming `id` to `generatedId` (this is the preferred way so far)
- renaming `id` to `opaqueId`

Note: this issue also applies to other classes (e.g. compute's `DiskImageConfiguration`).

ajkannan:
I'm +1 for `generatedId`

mziccard:
Just added the compute label, this must remain open unti fixed in compute.

aozarov:
I guess this can be closed now?

mziccard:
Fixed with #832 and #966

PR #832, opened 2016-04-01
Rename id to generatedId

mziccard:
This PR addresses #824:
- Rename `id` to `generatedId` for BigQuery resources
- Rename `id` to `generatedId` for Storage resources
- Rename `id` to `generatedId` for Dns resources 

For what concerns Dns I also renamed `id` in `ChangeRequestInfo` for consistency and because it is actually generated from the service. However, for this class there's no sources of confusion with other fields' names and `id` is much more used than in other resources/services so we could keep `id`. Do you have a preference?

### DIFF
--- a/gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
+++ b/gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
@@ -45,7 +45,7 @@ public class ZoneTest {
   private static final String ZONE_ID = "123";
   private static final ZoneInfo ZONE_INFO = Zone.of(ZONE_NAME, "example.com", "description")
       .toBuilder()
-      .id(ZONE_ID)
+      .generatedId(ZONE_ID)
       .creationTimeMillis(123478946464L)
       .build();
   private static final ZoneInfo NO_ID_INFO =
@@ -61,7 +61,7 @@ public class ZoneTest {
   private static final Dns.ChangeRequestListOption CHANGE_REQUEST_LIST_OPTIONS =
       Dns.ChangeRequestListOption.fields(Dns.ChangeRequestField.START_TIME);
   private static final ChangeRequestInfo CHANGE_REQUEST =
-      ChangeRequestInfo.builder().id("someid").build();
+      ChangeRequestInfo.builder().generatedId("someid").build();
   private static final ChangeRequestInfo CHANGE_REQUEST_NO_ID =
       ChangeRequestInfo.builder().build();
   private static final DnsException EXCEPTION = createStrictMock(DnsException.class);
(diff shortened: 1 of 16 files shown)


gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
```diff
@@ -482,7 +484,7 @@ public void testBuilder() {
     expect(dns.options()).andReturn(OPTIONS).times(8);
     replay(dns);
     assertNotEquals(zone, zone.toBuilder()
-        .id((new BigInteger(zone.id())).add(BigInteger.ONE).toString())
+        .generatedId((new BigInteger(zone.generatedId())).add(BigInteger.ONE).toString())
```
mziccard: Codacy Issue found: [Useless parentheses.](https://www.codacy.com/app/mziccard/gcloud-java/file/2340913774/issues/source?bid=3123103&fileBranchId=3212043#l487)

ajkannan:
Three issues that break the javadoc build on Travis jdk8:

```
BigQuery.java:475: error: reference not found:
({@link Dataset#datasetId()}, {@link Dataset#friendlyName()} and {@link Dataset#id()})

BigQuery.java:545: error: reference not found
({@link Table#tableId()}, {@link Table#friendlyName()}, {@link Table#id()} and type, which

BigQuery.java:556: error: reference not found:
({@link Table#tableId()}, {@link Table#friendlyName()}, {@link Table#id()} and type, which
```

gcloud-java-dns/src/test/java/com/google/gcloud/dns/ZoneTest.java
ajkannan: Unrelated to this PR, but seems like a good thing to fix.

mziccard:
@ajkannan yes I'm fixing it now, thanks

ajkannan:
Question: just grepping through the database, I see two instances of `id` left that I'm slightly unsure about.  Are they meant to be kept as `id` rather than `generatedId`?

* InsertAllRequest line 84
* QueryStage line 275 (which is actually a long, so perhaps not the related to this change)

mziccard:
- `InsertAllRequest`: `id` here is the optional id of a row and is user-provided
- `QueryStage`: query stage can not be created by the user and has no "identity" so I decided to keep id for brevity as it's already clear it's service-generated. If you prefer we can make it `generatedId` for consistency.

ajkannan:
LGTM

aozarov:
> QueryStage: query stage can not be created by the user and has no "identity" so I decided to keep id for brevity as it's already clear it's service-generated. If you prefer we can make it generatedId for consistency.

I would have called it `generatedId` so our code is consistent and without the need of understanding context. id -> user defined/supplied id or reference, generated-id read-only & service-genarated id.

ajkannan:
I can take care of this change

Issue #791, opened 2016-03-25
Decide whether to add another NIO example and split into two example jars


jean-philippe-martin:
Discussing #714, we considered adding another example that uses NIO in the "normal" way (via Maven), in addition to the unit tests. The downside of doing that is that we'd have to split the examples into two projects, since if we add a gcloud-java-nio dependency to the examples project then the Stat example will become pointless (its purpose is to show how to add GCS support to a project that does _not_ depend on gcloud-java-nio).

context: [#714 (comment)](https://github.com/GoogleCloudPlatform/gcloud-java/pull/714#discussion_r57388202)

jean-philippe-martin:
Fixed in PR #839.

Issue #778, opened 2016-03-22
Consider running all examples from a jar file


jean-philippe-martin:
context: https://github.com/GoogleCloudPlatform/gcloud-java/pull/714#discussion_r55616331

The Stat example uses a jar (that's its raison d'etre). @aozarov suggests we change the other examples to also use a jar.

jean-philippe-martin:
Settled in PR #839: all examples now use appassembler.

PR #839, opened 2016-04-01
Split off the "normal case" example from the "legacy case" one.

jean-philippe-martin:
This PR does two things:
- Move the current NIO example to a separate jar (this one still uses the "case 2" technique for adding GCS support to legacy jars).
- Create a very-similar NIO example in the examples jar, but this time using the "case 1" normal technique of adding gcloud-java-nio as a dependency.

This way we satisfy issue #778's goal of having all the examples being consistent per project. We also satisfy issue #791's goal of adding a new example and splitting into two example jars.

We'll probably want to subsequently add more NIO examples to the gcloud-java-examples project.

I put the new jar in gcloud-java-contrib, but we could put it somewhere else if appropriate, let me know.

I'm pulling this to start the discussion, but we can't merge it yet because (at least for me) it isn't working - the gcs filesystem provider doesn't load in the case 1 example. I don't yet know why.

_edit_: now I know and it works (see down the comment thread)


gcloud-java-contrib/README.md
```diff
@@ -7,6 +7,12 @@ Contents
 --------
 
  * [gcloud-java-nio](./gcloud-java-nio/): NIO Filesystem Provider for Google Cloud Storage.
+ * [gcloud-java-nio-example](./gcloud-java-nio-example/): How to add GCS NIO after the fact.
```
aozarov: Any reason for using example and not examples?

gcloud-java-examples/README.md
```diff
@@ -89,28 +89,22 @@ To run examples from your command line:
...
-     Compile the JAR with:
-    ```
-    mvn package -DskipTests -Dmaven.javadoc.skip=true  -Dmaven.source.skip=true
+    Before running the example, go to the [Google Developers Console][developers-console] to ensure that Google Cloud Storage API is enabled and that you have a bucket with a file in it.
+    Run the sample with:
     ```
-    Then run the sample with:
-    ```
-    java -cp gcloud-java-contrib/gcloud-java-nio/target/gcloud-java-nio-0.1.6-SNAPSHOT-shaded.jar:gcloud-java-examples/target/gcloud-java-examples-0.1.6-SNAPSHOT.jar com.google.gcloud.examples.nio.Stat --check
+    mvn exec:java -Dexec.mainClass="com.google.gcloud.examples.nio.Stat" -Dexec.args="--check"
```
aozarov: Wasn't this moved?

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/Query.java
```diff
@@ -30,7 +30,7 @@
 
 /**
  * A Google Cloud Datastore query.
- * For usage examples see {@link GqlQuery} and {@link StructuredQuery}.
+ * For usage example see {@link GqlQuery} and {@link StructuredQuery}.
```
aozarov: Is that better english?

gcloud-java-examples/src/main/java/com/google/gcloud/examples/bigquery/BigQueryExample.java
```diff
@@ -63,7 +63,7 @@
  * <li>login using gcloud SDK - {@code gcloud auth login}.</li>
  * <li>compile using maven - {@code mvn compile}</li>
  * <li>run using maven -
- * <pre>{@code mvn exec:java -Dexec.mainClass="com.google.gcloud.examples.bigquery.BigQueryExample"
+ * <pre>{@code mvn exec:java -Dexec.mainClass="com.google.gcloud.example.bigquery.BigQueryExample"
```
aozarov: here and in other places. I assume the com.google.gcloud.examples.rename was not intentional?

gcloud-java-contrib/README.md
jean-philippe-martin: Because there's only one example? I don't think the "use an external jar" needs more than one example.

gcloud-java-examples/README.md
jean-philippe-martin: that's the case 1. I did warn you it was *very* similar! 

The idea is that we'll have examples in both gcloud-java-examples and gcloud-java-nio-example. The former have gcloud-java-nio as a dependency, the latter add its jar in the classpath.

gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/Query.java
jean-philippe-martin: Wait, I didn't mean to change anything there! I think the IDE went a little crazy with the "helpful" renaming as I moved Stat. I'll revert that.

gcloud-java-examples/src/main/java/com/google/gcloud/examples/bigquery/BigQueryExample.java
jean-philippe-martin: indeed!

jean-philippe-martin:
OK, this should address all the issues. 

We still have the problem that `mvn exec:java -Dexec.mainClass="com.google.gcloud.examples.nio.Stat" -Dexec.args="--check"` does not work. I checked, and the jar file seems to have the correct SPI magic, so I'm not sure why we still get `java.nio.file.ProviderNotFoundException: Provider "gs" not found`.

As an experiment I have tried running the Stat example via an appassembler script, and this works just fine. This confirms that the jar is sound, and when adding it the way the end user would, the program works.

At this point it looks like `mvn exec:java` is the cause of our problems. We could either switch everything to appassembler

jean-philippe-martin:
Here's a proposal to switch the examples to using appassembler instead of mvn.exec. We'll have to talk about it first of course.

The benefits are:

 * Stat example now works (both use cases)
 * Easier to see what is available (list the bin folder)
 * Easier to see what running an example entails (read the script)
 * Simpler command line (no more -Dexec.args)

The first one is of course the biggest deal from my point of view. 

PTAL!

jean-philippe-martin:
I see Travis failed. Something about `RemoteGcsHelperTest`. This is surprising to me since I didn't change that class at all or (as far as I can tell) anything related to it.

gcloud-java-contrib/README.md
aozarov: OK, though I think a notion of an "examples" folder would not confuse people if there is only one example there however if we ever do add more examples calling the folder example is bit confusing.

aozarov: I think using "How to add GCS NIO to existing code" or "legacy code" is a bit more clear than "after the fact".

gcloud-java-contrib/gcloud-java-nio-example/README.md
```diff
@@ -0,0 +1,38 @@
+Example of adding the Google Cloud Storage NIO Provider to a legacy jar
+=======================================================================
+
+This project shows how to add GCS capabilities to a jar file for a Java 7 application
+that uses Java NIO but that you cannot recompile.
```
aozarov: > that you cannot recompile
Maybe, "without the need to recompile" ?

gcloud-java-contrib/gcloud-java-nio-example/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+3. Compile the JAR with:
+    ```
+    mvn package -DskipTests -Dmaven.javadoc.skip=true  -Dmaven.source.skip=true
+    ```
+
+4.    Run the sample with:
+
+    ```
+    java -cp gcloud-java-contrib/gcloud-java-nio/target/gcloud-java-nio-0.1.6-SNAPSHOT-shaded.jar:gcloud-java-contrib/gcloud-java-nio-example/target/gcloud-java-nio-example-0.1.6-SNAPSHOT.jar com.google.gcloud.nio.example.Stat --check
```
aozarov: You probably want to update [this](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/utilities/update_docs_version.sh) script, so the doc stays consistent with the current version.

/cc @ajkannan

gcloud-java-contrib/gcloud-java-nio-example/README.md
```diff
@@ -0,0 +1,38 @@
...
+    ```
+    Or, if you have a file in `gs://mybucket/myfile.txt`, you can run:
+    ```
+    java -cp gcloud-java-contrib/gcloud-java-nio/target/gcloud-java-nio-0.1.6-SNAPSHOT-shaded.jar:gcloud-java-contrib/gcloud-java-nio-example/target/gcloud-java-nio-example-0.1.6-SNAPSHOT.jar com.google.gcloud.nio.example.Stat gs://mybucket/myfile.txt
+    ```
+
+The sample doesn't have anything about GCS in it. It gets that ability from the nio jar that
```
aozarov: Would you like to shed some light on this magic and suggest that we are using the standard `ServiceLoader` facility here?

gcloud-java-examples/src/main/java/com/google/gcloud/examples/nio/Stat.java
```diff
@@ -20,6 +20,9 @@
  * be logged in (using e.g. the gcloud auth command).
  *
  * <p>See the README for a command line to run this example.
+ *
```
aozarov: It does not feel right this duplicate.

Instead of a duplicate example did you consider replacing this one with something else simple?

If not, maybe we can add `contrib-nio-example` as a dependency and reference the code from here?

gcloud-java-examples/pom.xml
```diff
@@ -37,6 +42,38 @@
           <skip>false</skip>
         </configuration>
       </plugin>
+        <!-- make a script so we can run stuff directly by calling
+             mvn package appassembler:assemble -DskipTests -Dmaven.javadoc.skip=true  -Dmaven.source.skip=true
+             from the gcloud-java-examples folder. The script is written to target/appassembler/bin/. -->
```
aozarov: /cc @ajkannan for including it in our `writing a new service` doc.

aozarov:
Looks good few more comments.

gcloud-java-contrib/README.md
jean-philippe-martin: OK, I put "legacy code" then (the distinction is whether you are recompiling or not - adding the gcloud-java-nio dependency to existing code would fall under "case 1" i.e. what gcloud-java-examples illustrates).

jean-philippe-martin: I've also renamed it "gcloud-java-nio-examples" since you think that's better.

gcloud-java-contrib/gcloud-java-nio-example/README.md
jean-philippe-martin: Sure, ok.

gcloud-java-contrib/gcloud-java-nio-example/README.md
jean-philippe-martin: Ah, so that's how you do it! I was wondering if it was all by hand. Updated.

gcloud-java-contrib/gcloud-java-nio-example/README.md
jean-philippe-martin: Sure! I added a bit of text to explain what's going on.

gcloud-java-examples/src/main/java/com/google/gcloud/examples/nio/Stat.java
jean-philippe-martin: I agree it's a bit odd, but the benefit would be that we can see that the very same code works using either mechanism.

If that bothers you, we can certainly replace it by something else - I'd prefer this to having the modules depend from each other. 

How about just something that lists the filesystem providers? We can call it "ListFilesystems". That still makes it super-easy to compare the output with and without the nio jar. It's a subset of what Stat can do but that should be enough to remove the confusion. What do you think?

gcloud-java-examples/pom.xml
ajkannan: I added this as a list item on the "writing a new service" doc issue; will include when I write the doc.

jean-philippe-martin:
Thanks @aozarov! This should address the comments.

I went ahead and changed one of the Stat examples to ListFilesystems instead. Let me know if you want it to be some other example instead!

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
```diff
@@ -0,0 +1,32 @@
+package com.google.gcloud.nio.examples;
```
aozarov: copyright

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
```diff
@@ -0,0 +1,32 @@
...
+    listFilesystems();
+  }
+
+  private static void listFilesystems() {
+    System.out.println("Installed filesystem providers:");
+    for (FileSystemProvider p : FileSystemProvider.installedProviders()) {
+      System.out.println("  " + p.getScheme());
+    }
+  }
+
```
aozarov: Nit, redundant line

utilities/update_docs_version.sh
```diff
@@ -17,6 +17,9 @@ if [ "${RELEASED_VERSION##*-}" != "SNAPSHOT" ]; then
         sed -ri "s/:[0-9]+\.[0-9]+\.[0-9]+'/:${RELEASED_VERSION}'/g" ${item}/README.md
         sed -ri "s/\"[0-9]+\.[0-9]+\.[0-9]+\"/\"${RELEASED_VERSION}\"/g" ${item}/README.md
     done
+
+    echo "Changing version to $((RELEASED_VERSION + 1))-SNAPSHOT in gcloud-java-nio-example README"
```
aozarov: @ajkannan does this look right to you?

ajkannan: Hm, wouldn't you want the gcloud-java-nio-example README to use the released version, rather than the "-SNAPSHOT" version?  In that case, the line above could become:

```
echo "Changing version to $RELEASED_VERSION in gcloud-java-nio-example README"
```
and you could add gcloud-java-nio-examples/README.md to the list "module folders" or use a command something like this:

```
sed -ri "s/gcloud-java-nio-examples-[0-9]+\.[0-9]+\.[0-9]/gcloud-java-nio-examples-${RELEASED_VERSION}/g" gcloud-java-contrib/gcloud-java-nio-examples/README.md
```

ajkannan: Actually, I see that a user might clone the repo and try this for him/herself, in which case the version will have a SNAPSHOT in it; I'll post commands for that.

ajkannan: I think you'll need to split the +1 b/c there are two "." in the version.  So something like this would work:

```
NEW_VERSION=${RELEASED_VERSION%.*}.$((${RELEASED_VERSION##*.}+1))-SNAPSHOT
echo "Changing version to $NEW_VERSION in gcloud-java-nio-example README"
sed -ri "s/gcloud-java-nio-examples-[0-9]+\.[0-9]+\.[0-9]+-SNAPSHOT/gcloud-java-nio-examples-$NEW_VERSION/g" gcloud-java-contrib/gcloud-java-nio-examples/README.md
```

mziccard:
> We still have the problem that mvn exec:java -Dexec.mainClass="com.google.gcloud.examples.nio.Stat" -Dexec.args="--check" does not work. I checked, and the jar file seems to have the correct SPI magic, so I'm not sure why we still get java.nio.file.ProviderNotFoundException: Provider "gs" not found.

This may come a little late but this issue can be overcome. The problem here is that `gcloud-java-nio` is not available as an artifact dependency and that `exec:java` does not pass it in the classpath. To fix this add

```xml
    <dependency>
      <groupId>com.google.gcloud</groupId>
      <artifactId>gcloud-java-nio</artifactId>
      <version>0.1.6-SNAPSHOT</version>
    </dependency>
```
to the example's `pom.xml` and execute the class with:
```bash
mvn exec:exec -Dexec.args="-classpath %classpath com.google.gcloud.nio.example.Stat --check"
```
This will pass all jars in the classpath and the file system provider will be available.

gcloud-java-contrib/gcloud-java-nio-examples/pom.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.gcloud</groupId>
```
mziccard: This is redundant right? Should be inherited from the parent project.

gcloud-java-contrib/gcloud-java-nio-examples/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+      <version>0.27</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>1.9.5</version>
+    </dependency>
+      <dependency>
+          <groupId>org.apache.maven.plugins</groupId>
```
mziccard: Using wrong indent here

gcloud-java-contrib/gcloud-java-nio-examples/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+          <artifactId>maven-assembly-plugin</artifactId>
+          <version>2.5.4</version>
+      </dependency>
+  </dependencies>
+  <build>
+      <plugins>
```
mziccard: Also here and in the following: using the wrong indent

jean-philippe-martin:
@mziccard in reply to the exec.java comment above (GitHub doesn't seem to have a better way to reply?)

Thank you for your suggestion, that works indeed! I just had to change the command to

```
mvn exec:exec -Dexec.args="-classpath %classpath com.google.gcloud.examples.nio.Stat --check"
```

I didn't need to change the pom because gcloud-java-nio was already listed as a dependency.

It is weird though to have this example use exec:exec and all the other ones use exec:java. I tried to make it work with the latter; my attempt is `mvn exec:java -Dexec.mainClass="com.google.gcloud.examples.nio.Stat" -Dexec.args="-classpath %classpath --check" ` but this failed because of a "malformed escape". It also doesn't work if I don't specify the classpath, even though one might reasonably expect maven to set the classpath correctly since it already knows about all the dependencies.

So now we have to decide whether we want to go with appassembler, or go back to mvn exec:*. Personally I lean towards appassembler because it's more consistent and for the reasons highlighted before (in short, the magic is better contained: for example I tried to use exec:exec before but failed because I didn't know it needed an explicit classpath argument with a magical string).

gcloud-java-contrib/gcloud-java-nio-examples/pom.xml
jean-philippe-martin: Fixed.

gcloud-java-contrib/gcloud-java-nio-examples/pom.xml
jean-philippe-martin: Fixed.

gcloud-java-contrib/gcloud-java-nio-examples/pom.xml
jean-philippe-martin: Removed here and in gcloud-java-nio. Also reformatted with a 2-space indent rule.

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
jean-philippe-martin: added

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
jean-philippe-martin: Removed the empty line (I'm guessing that's what you meant?)

utilities/update_docs_version.sh
jean-philippe-martin: Thank you very much, changed it to that.

mziccard:
@jean-philippe-martin I am fine with `appassembler` as long as we are consistent. This branch was detached from master before the `dns` branch was merged. We must remember to also switch `dns` examples to `appassembler` before merging.

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
```diff
@@ -28,5 +43,4 @@ private static void listFilesystems() {
       System.out.println("  " + p.getScheme());
     }
   }
-
-}
+}
```
mziccard: Nit: missing newline

jean-philippe-martin: Ah, that's what I get for removing all empty lines in sight when I wasn't sure which empty line you didn't like. OK, put that one back in.

jean-philippe-martin:
Agreed then, let's use `appassembler` and remember to switch the `dns` example to it before merging the gcs-nio branch.

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
```diff
@@ -43,4 +43,5 @@ private static void listFilesystems() {
       System.out.println("  " + p.getScheme());
     }
   }
+
```
aozarov: Why?

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
```diff
@@ -43,4 +43,5 @@ private static void listFilesystems() {
...
 }
```
aozarov: add new line here

aozarov:
Other than the last comment is there anything else we need to do before merge? @jean-philippe-martin @mziccard

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
jean-philippe-martin: Because @mziccard asked me to add a newline [just a moment ago](https://github.com/GoogleCloudPlatform/gcloud-java/pull/839#discussion_r58595935).

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
jean-philippe-martin: You want me to add a blank line after the final closing bracket? The same line that @mziccard asked me to remove [just a little bit ago](https://github.com/GoogleCloudPlatform/gcloud-java/pull/839#discussion_r58479804)?

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
aozarov: I am pretty sure he was refereing to a new line at the end of file. Am I wrong @mziccard ?

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
aozarov: Yes, but I think you misunderstood @mziccard.

jean-philippe-martin: OK, I put it back. I believe that some editors like to strip (or add) a final \n. I can't imagine it would ever matter, though.

aozarov: True, though some code review tools (such as this one) complain about it.

jean-philippe-martin:
Nothing left to do before the merge as far as I know.

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
mziccard: @aozarov is right. I meant to add a newline before the EOF and remove the one before the last `}`. To give some context here [why should text files end with a newline?](http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

mziccard:
No more comments from me but the very important matter about newlines :) Sorry for not being clear!

gcloud-java-contrib/gcloud-java-nio-examples/src/main/java/com/google/gcloud/nio/examples/ListFilesystems.java
jean-philippe-martin: Thanks for the context, I learned something. I pushed an update with the newline but it doesn't seem to show up on GitHub yet, not sure why.

utilities/update_docs_version.sh
```diff
@@ -17,7 +17,11 @@ if [ "${RELEASED_VERSION##*-}" != "SNAPSHOT" ]; then
         sed -ri "s/:[0-9]+\.[0-9]+\.[0-9]+'/:${RELEASED_VERSION}'/g" ${item}/README.md
         sed -ri "s/\"[0-9]+\.[0-9]+\.[0-9]+\"/\"${RELEASED_VERSION}\"/g" ${item}/README.md
     done
-    
+
+    NEW_VERSION=${RELEASED_VERSION%.*}.$((${RELEASED_VERSION##*.}+1))-SNAPSHOT
+    echo "Changing version to $NEW_VERSION in gcloud-java-nio-example README"
+    sed -ri "s/gcloud-java-nio-examples-[0-9]+\.[0-9]+\.[0-9]+-SNAPSHOT/gcloud-java-nio-examples-$NEW_VERSION/g" gcloud-java-contrib/gcloud-java-nio-examples/README.md
```
ajkannan: One thing I should add here is that there will be a small amount of time (during the release process) when the gcloud-java-nio-examples readme will be out of sync with the actual version of the project.  This script gets run while the project version is a non-snapshot.  Until the next PR goes through to update the version to x.x.x+1-SNAPSHOT, this readme will be out of date.  However, the amount of time that the readme will be out of sync should be small, since we typically push the version to x.x.x+1-SNAPSHOT as soon as the push to maven is complete (which usually takes under an hour).  An alternative approach to instruct the user to use a `find` command to get the proper version, and then use that environment variable to run the example. 

I think it's fine as is, but it's worth mentioning the alternative.

jean-philippe-martin: Good point. I agree that we're probably fine as is, no point making the instructions super-complicated just to account for that short time window.

PR #844, opened 2016-04-03
Add DiskInfo, disk configurations and test classes

mziccard:
This PR adds non-functional classes for disk and disk configurations. Unit tests are also added.

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
@@ -109,7 +109,7 @@ public abstract class DiskConfiguration implements Serializable {
     }
 
     /**
-     * Sets the identity of the disk type.
+     * Sets the identity of the disk type. If not set {@code pd-standard} will be used.
      */
     public B diskType(DiskTypeId diskType) {
       this.diskType = diskType;
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
@@ -147,7 +147,10 @@ public class DiskInfo implements Serializable {
     private Long lastAttachTimestamp;
     private Long lastDetachTimestamp;
 
-    BuilderImpl() {}
+    BuilderImpl(DiskId diskId, DiskConfiguration configuration) {
+      this.diskId = checkNotNull(diskId);
+      this.configuration = checkNotNull(configuration);
+    }
 
     BuilderImpl(DiskInfo diskInfo) {
       this.id = diskInfo.id;
@@ -258,10 +261,10 @@ public class DiskInfo implements Serializable {
 
   DiskInfo(BuilderImpl builder) {
     this.id = builder.id;
-    this.configuration = checkNotNull(builder.configuration);
+    this.configuration = builder.configuration;
     this.creationTimestamp = builder.creationTimestamp;
     this.creationStatus = builder.creationStatus;
-    this.diskId = checkNotNull(builder.diskId);
+    this.diskId = builder.diskId;
     this.description = builder.description;
     this.licenses = builder.licenses;
     this.attachedInstances = builder.attachedInstances;
@@ -383,7 +386,7 @@ public class DiskInfo implements Serializable {
    * {@link ImageDiskConfiguration} to create a disk from a disk image.
    */
   public static Builder builder(DiskId diskId, DiskConfiguration configuration) {
-    return new BuilderImpl().diskId(diskId).configuration(configuration);
+    return new BuilderImpl(diskId, configuration);
   }
 
   /**
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageDiskConfiguration.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageDiskConfiguration.java
@@ -44,8 +44,9 @@ public class ImageDiskConfiguration extends DiskConfiguration {
     private ImageId sourceImage;
     private String sourceImageId;
 
-    private Builder() {
+    private Builder(ImageId sourceImage) {
       super(Type.IMAGE);
+      this.sourceImage = checkNotNull(sourceImage);
     }
 
     private Builder(ImageDiskConfiguration configuration) {
@@ -58,6 +59,17 @@ public class ImageDiskConfiguration extends DiskConfiguration {
       super(Type.IMAGE, diskPb);
       this.sourceImage = ImageId.fromUrl(diskPb.getSourceImage());
       this.sourceImageId = diskPb.getSourceImageId();
+    }
+
+    /**
+     * Sets the size of the persistent disk, in GB. If not set the disk will have the size of the
+     * image. This value can be larger than the image's size. If the provided size is smaller than
+     * the image's size then disk creation will fail.
+     */
+    @Override
+    public Builder sizeGb(Long sizeGb) {
+      super.sizeGb(sizeGb);
+      return this;
     }
 
     /**
@@ -84,7 +96,7 @@ public class ImageDiskConfiguration extends DiskConfiguration {
 
   private ImageDiskConfiguration(Builder builder) {
     super(builder);
-    this.sourceImage = checkNotNull(builder.sourceImage);
+    this.sourceImage = builder.sourceImage;
     this.sourceImageId = builder.sourceImageId;
   }
 
@@ -145,7 +157,7 @@ public class ImageDiskConfiguration extends DiskConfiguration {
    * Returns a builder for an {@code ImageDiskConfiguration} object given the image identity.
    */
   public static Builder builder(ImageId imageId) {
-    return new Builder().sourceImage(imageId);
+    return new Builder(imageId);
   }
 
   /**
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/SnapshotDiskConfiguration.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/SnapshotDiskConfiguration.java
@@ -45,8 +45,9 @@ public class SnapshotDiskConfiguration extends DiskConfiguration {
     private SnapshotId sourceSnapshot;
     private String sourceSnapshotId;
 
-    private Builder() {
+    private Builder(SnapshotId sourceSnapshot) {
       super(Type.SNAPSHOT);
+      this.sourceSnapshot = checkNotNull(sourceSnapshot);
     }
 
     private Builder(SnapshotDiskConfiguration configuration) {
@@ -59,6 +60,21 @@ public class SnapshotDiskConfiguration extends DiskConfiguration {
       super(Type.SNAPSHOT, diskPb);
       this.sourceSnapshot = SnapshotId.fromUrl(diskPb.getSourceSnapshot());
       this.sourceSnapshotId = diskPb.getSourceSnapshotId();
+    }
+
+    /**
+     * Sets the size of the persistent disk, in GB. If not set the disk will have the size of the
+     * snapshot. This value can be larger than the snapshot's size. If the provided size is smaller
+     * than the snapshot's size then disk creation will fail.
+     *
+     * @see <a href=
+     *     "https://cloud.google.com/compute/docs/disks/persistent-disks#restoresnapshotlargersize">
+     *     Restoring a snapshot to a larger size</a>
+     */
+    @Override
+    public Builder sizeGb(Long sizeGb) {
+      super.sizeGb(sizeGb);
+      return this;
     }
 
     /**
@@ -85,7 +101,7 @@ public class SnapshotDiskConfiguration extends DiskConfiguration {
 
   private SnapshotDiskConfiguration(Builder builder) {
     super(builder);
-    this.sourceSnapshot = checkNotNull(builder.sourceSnapshot);
+    this.sourceSnapshot = builder.sourceSnapshot;
     this.sourceSnapshotId = builder.sourceSnapshotId;
   }
 
@@ -148,7 +164,7 @@ public class SnapshotDiskConfiguration extends DiskConfiguration {
    * Returns a builder for a {@code SnapshotDiskConfiguration} object given the snapshot identity.
    */
   public static Builder builder(SnapshotId sourceSnapshot) {
-    return new Builder().sourceSnapshot(sourceSnapshot);
+    return new Builder(sourceSnapshot);
   }
 
   /**
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
@@ -49,6 +49,15 @@ public class StandardDiskConfiguration extends DiskConfiguration {
     }
 
     /**
+     * Sets the size of the persistent disk, in GB. If not set, 500GB is used.
+     */
+    @Override
+    public Builder sizeGb(Long sizeGb) {
+      super.sizeGb(sizeGb);
+      return this;
+    }
+
+    /**
      * Creates a {@code StandardDiskConfiguration} object.
      */
     @Override


gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
```diff
@@ -0,0 +1,439 @@
...
+    private DiskConfiguration configuration;
+    private Long creationTimestamp;
+    private CreationStatus creationStatus;
+    private String description;
+    private List<LicenseId> licenses;
+    private List<InstanceId> attachedInstances;
+    private Long lastAttachTimestamp;
+    private Long lastDetachTimestamp;
+
+    BuilderImpl() {}
```
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2350430174/issues/source?bid=3140968&fileBranchId=3217543#l150)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+
+import com.google.api.services.compute.model.Disk;
+import com.google.common.base.MoreObjects;
+import com.google.common.base.MoreObjects.ToStringHelper;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Base class for Google Compute Engine disk configuration. A disk can be used as primary storage
```
ajkannan: Nit: I think this would sound slightly nicer if you make "configuration" plural.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+ * for your virtual machine instances. Use {@link StandardDiskConfiguration} to create a standard
+ * disk given disk type and size. Use {@link ImageDiskConfiguration} to create a disk from a Compute
```
ajkannan: Nit: it'd be slightly nicer to add "a" before "disk type".

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * Sets the size of the persistent disk, in GB.
+     */
+    public B sizeGb(Long sizeGb) {
+      this.sizeGb = sizeGb;
+      return self();
+    }
+
+    /**
+     * Sets the identity of the disk diskType.
```
ajkannan: s/disk diskType/disk type/?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
```diff
@@ -0,0 +1,439 @@
...
+        @Override
+        public Disk apply(DiskInfo diskType) {
+          return diskType.toPb();
+        }
+      };
+
+  private static final long serialVersionUID = -7173418340679279619L;
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER = ISODateTimeFormat.dateTime();
+
+  private final String id;
```
ajkannan: s/id/generatedId/g

mziccard: I'll apply all the `s/id/generatedId/` changes together. Next PRs will have the usual `id` naming. Once instance and network-related classes are merged I'll fix this issue everywhere in compute.

ajkannan: Ah, actually this is something you plan on fixing in the future, and not in this PR, correct?

ajkannan: Sounds good.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
```diff
@@ -0,0 +1,439 @@
...
+
+  /**
+   * Returns all applicable publicly visible licenses for the disk.
+   */
+  public List<LicenseId> licenses() {
+    return licenses;
+  }
+
+  /**
+   * Returns all instances' identities this disk is attached to.
```
ajkannan: Nit: I think this could sound slightly clearer as "Returns all the identities of the instances this disk is attached to"

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
```diff
@@ -0,0 +1,439 @@
...
+    return obj != null
+        && obj.getClass().equals(DiskInfo.class)
+        && Objects.equals(toPb(), ((DiskInfo) obj).toPb());
+  }
+
+  /**
+   * Returns a builder for a {@code DiskInfo} object given its identity and configuration. Use
+   * {@link StandardDiskConfiguration} to create a simple disk given its type and size. Use
+   * {@link SnapshotDiskConfiguration} to create a disk from a snapshot. Use
+   * {@link ImageDiskConfiguration} to create a disk form a disk image.
```
ajkannan: s/form/from/

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
```diff
@@ -0,0 +1,439 @@
...
+   */
+  public static Builder builder(DiskId diskId, DiskConfiguration configuration) {
+    return new BuilderImpl().diskId(diskId).configuration(configuration);
+  }
+
+  /**
+   * Returns a {@code DiskInfo} object given its identity and configuration. Use
+   * {@link StandardDiskConfiguration} to create a simple disk given its type and size. Use
+   * {@link SnapshotDiskConfiguration} to create a disk from a snapshot. Use
+   * {@link ImageDiskConfiguration} to create a disk form a disk image.
```
ajkannan: s/form/from/

gcloud-java-compute/src/main/java/com/google/gcloud/compute/SnapshotDiskConfiguration.java
```diff
@@ -0,0 +1,165 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.api.services.compute.model.Disk;
+import com.google.common.base.MoreObjects;
+
+import java.util.Objects;
+
+/**
+ * A Google Compute Engine disk configuration to create a disk from a Google Compute Engine
+ * snaoshot.
```
ajkannan: s/snaoshot/snapshot/

gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+
+package com.google.gcloud.compute;
+
+import com.google.api.services.compute.model.Disk;
+
+import java.util.Objects;
+
+/**
+ * A Google Compute Engine standard persistent disk configuration. Allows to create a disk given its
```
ajkannan: s/Allows to/Allows the user to/?

ajkannan:
On naming, I've momentarily confused myself thinking that `DiskImageConfiguration` was  `ImageDiskConfiguration` while reviewing.  I imagine that if I were coding using our library, I'd have to think a little extra to remember which one pertained to each situation just because the names are similar.  I know it's difficult to think of brief class names that differentiate the two, but I wanted to hear your opinion on whether it's something worth thinking about renaming?

mziccard:
@ajkannan I agree this is a little confusing however it also follows the convention `<Source><Destination>Configuration`. I couldn't think of a better naming because the two configuration are specular: an image used to create a disk and a disk used to create an image. Any suggestion is appreciated.

The javadoc (as well as java's static types + IDE autocompletion) should help users sorting this out.

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskInfoTest.java
```diff
@@ -0,0 +1,248 @@
...
+        .diskId(DiskId.of(DISK_ID.zone(), DISK_ID.disk()))
+        .configuration(imageDiskConfiguration)
+        .build();
+    compareDiskInfo(IMAGE_DISK_INFO, diskInfo.setProjectId("project"));
+  }
+
+  public void compareDiskInfo(DiskInfo expected, DiskInfo value) {
+    assertEquals(expected, value);
+    assertEquals(expected.configuration(), value.configuration());
+    assertEquals(expected.id(), value.id());
```
ajkannan: Is it valuable to `assertEquals` for `diskId` as well?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskInfoTest.java
```diff
@@ -0,0 +1,248 @@
...
+    assertEquals(expected.creationTimestamp(), value.creationTimestamp());
+    assertEquals(expected.creationStatus(), value.creationStatus());
+    assertEquals(expected.description(), value.description());
+    assertEquals(expected.licenses(), value.licenses());
+    assertEquals(expected.attachedInstances(), value.attachedInstances());
```
ajkannan: Is there a reason lastAttatchTimestamp and lastDetachTimestamp are omitted?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/DiskInfoTest.java
```diff
@@ -0,0 +1,248 @@
...
+    diskInfo = DiskInfo.of(DISK_ID, SNAPSHOT_DISK_CONFIGURATION);
+    assertEquals(diskInfo, diskInfo.toBuilder().build());
+    diskInfo = DiskInfo.of(DISK_ID, IMAGE_DISK_CONFIGURATION);
+    assertEquals(diskInfo, diskInfo.toBuilder().build());
+  }
+
+  @Test
+  public void testBuilder() {
+    assertEquals(ID, DISK_INFO.id());
+    assertEquals(DISK_CONFIGURATION, DISK_INFO.configuration());
```
ajkannan: Do you think it's worth checking diskId, lastAttachTimestamp, and lastDetachTimestamp here as well?

gcloud-java-compute/src/test/java/com/google/gcloud/compute/ImageDiskConfigurationTest.java
```diff
@@ -0,0 +1,112 @@
...
+      ImageDiskConfiguration.builder(IMAGE)
+          .sizeGb(SIZE)
+          .diskType(DISK_TYPE)
+          .sourceImageId(IMAGE_ID)
+          .build();
+
+  @Test
+  public void testToBuilder() {
+    compareImageDiskConfiguration(DISK_CONFIGURATION, DISK_CONFIGURATION.toBuilder().build());
+    ImageId newImageId = ImageId.of("newProjet", "newImage");
```
ajkannan: Nit: doesn't really matter for test correctness, but "newProjet" --> "newProject"

gcloud-java-compute/src/test/java/com/google/gcloud/compute/SnapshotDiskConfigurationTest.java
```diff
@@ -0,0 +1,112 @@
...
+      SnapshotDiskConfiguration.builder(SNAPSHOT)
+          .sizeGb(SIZE)
+          .diskType(DISK_TYPE)
+          .sourceSnapshotId(SNAPSHOT_ID)
+          .build();
+
+  @Test
+  public void testToBuilder() {
+    compareSnapshotDiskConfiguration(DISK_CONFIGURATION, DISK_CONFIGURATION.toBuilder().build());
+    SnapshotId newSnapshot = SnapshotId.of("newProjet", "newSnapshot");
```
ajkannan: Same as above, doesn't matter, but could change "newProjet" to "newProject"

gcloud-java-compute/src/test/java/com/google/gcloud/compute/StandardDiskConfigurationTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Test
+  public void testSetProjectId() {
+    StandardDiskConfiguration configuration = DISK_CONFIGURATION.toBuilder()
+        .diskType(DiskTypeId.of(DISK_TYPE.zone(), DISK_TYPE.diskType()))
+        .build();
+    compareDefaultDiskConfiguration(DISK_CONFIGURATION, configuration.setProjectId("project"));
+  }
+
+  private void compareDefaultDiskConfiguration(StandardDiskConfiguration expected,
```
ajkannan: Nit: could call this `compareStandardDiskConfiguration` (to align with the class name)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @Override
+  public final boolean equals(Object obj) {
+    return obj instanceof StandardDiskConfiguration && baseEquals((StandardDiskConfiguration) obj);
+  }
+
+  @Override
+  StandardDiskConfiguration setProjectId(String projectId) {
+    if (diskType() == null || diskType().project() != null) {
+      return this;
```
ajkannan: In the other disk configuration subclasses, you use the project ID given as a parameter, no matter what (probably because there are two IDs that need to be updated with a project ID, and it's cleaner and simpler to replace both rather than check for an existing project ID).  Do you think there's any issue that the behavior here is slightly different (we only set the project ID if one isn't set already)?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Creates a {@code StandardDiskConfiguration} object.
+     */
+    @Override
+    public StandardDiskConfiguration build() {
+      return new StandardDiskConfiguration(this);
+    }
+  }
+
+  private StandardDiskConfiguration(StandardDiskConfiguration.Builder builder) {
```
ajkannan: Nit: could change StandardDiskConfiguration.Builder to Builder

gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+   * A builder for {@code StandardDiskConfiguration} objects.
+   */
+  public static class Builder
+      extends DiskConfiguration.Builder<StandardDiskConfiguration, Builder> {
+
+    private Builder() {
+      super(Type.STANDARD);
+    }
+
+    private Builder(StandardDiskConfiguration diskInfo) {
```
ajkannan: Is `diskInfo` misnamed here?  Should it be something like `configuration`?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/SnapshotDiskConfiguration.java
```diff
@@ -0,0 +1,165 @@
...
+      extends DiskConfiguration.Builder<SnapshotDiskConfiguration, Builder> {
+
+    private SnapshotId sourceSnapshot;
+    private String sourceSnapshotId;
+
+    private Builder() {
+      super(Type.SNAPSHOT);
+    }
+
+    private Builder(SnapshotDiskConfiguration snapshotDiskInfo) {
```
ajkannan: I feel like the name containing "DiskInfo" is a bit misleading; maybe change to "snapshotDiskConfiguration" or just "configuration"?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/SnapshotDiskConfiguration.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  @Override
+  public final int hashCode() {
+    return Objects.hash(baseHashCode(), sourceSnapshot, sourceSnapshotId);
+  }
+
+  @Override
+  public final boolean equals(Object obj) {
+    return obj instanceof SnapshotDiskConfiguration && baseEquals((SnapshotDiskConfiguration) obj);
```
ajkannan: Since `baseEquals` checks if the runtime classes are the same, does that mean that the first half of this boolean statement is unnecessary?

ajkannan: Actually we'd need an instanceof anyways to cast, so ignore that comment.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/SnapshotDiskConfiguration.java
```diff
@@ -0,0 +1,165 @@
...
+  Disk toPb() {
+    return super.toPb()
+        .setSourceSnapshot(sourceSnapshot.selfLink())
+        .setSourceSnapshotId(sourceSnapshotId);
+  }
+
+  /**
+   * Returns a builder for a {@code SnapshotDiskConfiguration} object given the snapshot identity.
+   */
+  static Builder builder(SnapshotId sourceSnapshot) {
```
ajkannan: public?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    diskPb.setSizeGb(sizeGb);
+    if (diskType != null) {
+      diskPb.setType(diskType.selfLink());
+    }
+    return diskPb;
+  }
+
+  @SuppressWarnings("unchecked")
+  static <T extends DiskConfiguration> T fromPb(Disk diskPb) {
+    if (diskPb.getSourceImage() != null) {
```
ajkannan: Fields selection will come in a future PR (I assume); we plan to always choose the sourceImage, sourceSnapshot, and diskConfiguration fields, correct?  (That way we won't accidentally default to a `StandardDiskConfiguration`)

ajkannan:
Done with my pass.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2355537424/issues/source?bid=3140968&fileBranchId=3217543#l150)

mziccard:
Thanks for the pass @ajkannan all comments are fixed

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    private Type type;
+    private Long sizeGb;
+    private DiskTypeId diskType;
+
+    Builder(Type type) {
+      this.type = type;
+    }
+
+    Builder(Type type, DiskConfiguration diskConfiguration) {
+      this.type = type;
```
aozarov: Any reason for not taking it from `diskConfiguration.type()`?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+   * Base builder for disk configurations.
+   *
+   * @param <T> the disk configuration type
+   * @param <B> the disk configuration builder
+   */
+  public abstract static class Builder<T extends DiskConfiguration, B extends Builder<T, B>> {
+
+    private Type type;
+    private Long sizeGb;
+    private DiskTypeId diskType;
```
aozarov: Any reason for not calling it diskTypeId (and its setter/accesor)?

mziccard: Aren't setters/getters called `diskType` as well?

aozarov: They are. I just wonder why they are not called `diskTypeId`?

mziccard: I believe this is consistent with what we did so far. Quoting @ajkannan from our previous discussion on this

> I see, I think @mziccard uses [class]Id() and id() in [class].java, but anywhere there's a field that points to the identity of another class, he uses [class] instead of [class]Id, probably for the sake of brevity. In this case, @mziccard's code as it is is consistent with other usage.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
mziccard: Codacy Issue found: [Document empty constructor](https://www.codacy.com/app/mziccard/gcloud-java/file/2362402782/issues/source?bid=3140968&fileBranchId=3217543#l150)

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
mziccard: No. Fixed in 1fb53fa

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskInfo.java
```diff
@@ -0,0 +1,439 @@
...
+  }
+
+  /**
+   * Returns a builder for a {@code DiskInfo} object given its identity and configuration. Use
+   * {@link StandardDiskConfiguration} to create a simple disk given its type and size. Use
+   * {@link SnapshotDiskConfiguration} to create a disk from a snapshot. Use
+   * {@link ImageDiskConfiguration} to create a disk from a disk image.
+   */
+  public static Builder builder(DiskId diskId, DiskConfiguration configuration) {
+    return new BuilderImpl().diskId(diskId).configuration(configuration);
```
aozarov: Any reason not to move these arguments to the empty constructor?
Also, if we do so, the checkNotNull(configuration) in the DiskInfo constructor may not be necessary.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageDiskConfiguration.java
```diff
@@ -0,0 +1,162 @@
...
+ * A Google Compute Engine disk configuration to create a disk from a Google Compute Engine image.
+ *
+ * @see <a href="https://cloud.google.com/compute/docs/disks/">Block Storage</a>
+ */
+public class ImageDiskConfiguration extends DiskConfiguration {
+
+  private static final long serialVersionUID = 6469117882950722812L;
+
+  private final ImageId sourceImage;
+  private final String sourceImageId;
```
aozarov: is that a generated id one?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/SnapshotDiskConfiguration.java
```diff
@@ -0,0 +1,165 @@
...
+    return super.toPb()
+        .setSourceSnapshot(sourceSnapshot.selfLink())
+        .setSourceSnapshotId(sourceSnapshotId);
+  }
+
+  /**
+   * Returns a builder for a {@code SnapshotDiskConfiguration} object given the snapshot identity.
+   */
+  public static Builder builder(SnapshotId sourceSnapshot) {
+    return new Builder().sourceSnapshot(sourceSnapshot);
```
aozarov: similar comment like above. Any reason not to move it to the empty constructor.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+
+    Builder(DiskConfiguration diskConfiguration) {
+      this.type = diskConfiguration.type;
+      this.sizeGb = diskConfiguration.sizeGb;
+      this.diskType = diskConfiguration.diskType;
+    }
+
+    Builder(Type type, Disk diskPb) {
+      this.type = type;
+      this.sizeGb = diskPb.getSizeGb();
```
aozarov: How is the size relates to image or snapshot?
What if the value is smaller than the image or the snapshot?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+     * Creates a {@code StandardDiskConfiguration} object.
+     */
+    @Override
+    public StandardDiskConfiguration build() {
+      return new StandardDiskConfiguration(this);
+    }
+  }
+
+  private StandardDiskConfiguration(Builder builder) {
+    super(builder);
```
aozarov: What if type and size or type were not provided?

aozarov:
Looks good. Few comments/question. Feel free to merge after you consider them.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/ImageDiskConfiguration.java
mziccard: Yes this is a generated one, the name will change as soon as I fix #832 for the whole Compute.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/DiskConfiguration.java
mziccard: Snapshots and images have their size depending on the disk/raw source from which they have been taken. If this size is not specified the image/snapshot is used. If this is specified and is bigger than the size of the image/snapshot the the image/snapshot [will be resized](https://cloud.google.com/compute/docs/disks/persistent-disks#restoresnapshotlargersize).
If this value is set and is smaller than the image/snapshot size then disk creation will fail.

I'll update the docs to include this information.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/StandardDiskConfiguration.java
mziccard: If type is not provided `pd-standard` is used
If size is not provided 500GB is used
I will update the docs to add this information.

PR #845, opened 2016-04-04
Extract projectId from GOOGLE_APPLICATION_CREDENTIALS if set

mziccard:
This fixes the second part of #827

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,8 @@ Most `gcloud-java` libraries require a project ID.  There are multiple ways to s
 
 1. Project ID supplied when building the service options
 2. Project ID specified by the environment variable `GCLOUD_PROJECT`
-3. Project ID specified in the JSON credentials file pointed by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
-4. App Engine project ID
+3. App Engine project ID
+4. Project ID specified in the JSON credentials file pointed by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
 5. Google Cloud SDK project ID
 6. Compute Engine project ID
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,9 +83,10 @@ Most `gcloud-java` libraries require a project ID.  There are multiple ways to s
 
 1. Project ID supplied when building the service options
 2. Project ID specified by the environment variable `GCLOUD_PROJECT`
-3. App Engine project ID
-4. Google Cloud SDK project ID
-5. Compute Engine project ID
+3. Project ID specified in the JSON credentials file pointed by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
```
aozarov: Credentials order has options (3) and (4) switched.
Do you think we should make the order here consistent with credentials order?

gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java
```diff
@@ -479,6 +487,20 @@ protected static String getAppEngineProjectId() {
     }
   }
 
+  protected static String serviceAccountProjectId() {
+    String project = null;
+    String credentialsPath = System.getenv("GOOGLE_APPLICATION_CREDENTIALS");
+    if(credentialsPath != null) {
+      try (InputStream credentialsStream = new FileInputStream(credentialsPath)) {
```
aozarov: That is another reason why we would like the AE to pick up first, as this would likely to fail when run on AE.

mziccard: This should never be executed on AE as the `GOOGLE_APPLICATION_CREDENTIALS` is not set there, right? BTW you are right we better keep the same order that we have for credentials.

aozarov: One could set it like [so] (https://cloud.google.com/appengine/docs/java/config/appconfig#Java_appengine_web_xml_System_properties_and_environment_variables). However in this case, that would not be a good idea.

mziccard:
Updated the ordering to match credentials' one.

Issue #701, opened 2016-03-02
Is it possible to use ApplicationDefaultCredentials for SignUrlOption?


tmatsuo:
Currently [Storage.SignUrlOption](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java#L782) receives `com.google.gcloud.AuthCredentials.ServiceAccountAuthCredentials`, but how to use ApplicationDefaultAuthCredentials with that method? I think we should allow using ApplicationDefaultAuthCredentials if possible.

tmatsuo:
I think there should be an interface for Credentials and use it as the argument, just my 2 cents

aozarov:
Indeed as of now you can only explicitly set `com.google.gcloud.AuthCredentials.ServiceAccountAuthCredentials`.

However if you don't specify it `signUrl` may still work based on your service auth config (if it is configured with `AuthCredentials.ServiceAccountAuthCredentials` or using `ApplicationDefaultAuthCredentials` AND the default auth config is associated with a [service account](https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java).

`ApplicationDefaultAuthCredentials` is not required to be associated with a service account (and provide private-key).

tmatsuo:
Ok, thanks! I didn't know that you can not use the GCE service account for signing URL. However, isn't it appropriate for that method to receive an interface? Especially if there is a possibility that GCE service account can be used for signing in the future (I'm not sure though).


Related:
http://googlecloudplatform.github.io/gcloud-python/stable/storage-blobs.html#gcloud.storage.blob.Blob.generate_signed_url
https://github.com/GoogleCloudPlatform/gcloud-python/issues/922

aozarov:
Yes, we could do something similar to `gcloud-python` and accept `AuthCredentials` instead of the specific `ServiceAccountAuthCredentials` but similar to the Python API this class does not expose
in its API a service-account or private key.

We could still get it and then reject any instances that are not known to us to support the signing.
I don't like that such enforcement is a run-time failure but maybe that is not *too* bad as a runtime
failure is already possible when not providing anything.
Also, I don't like that this way a user could not implement its own AuthCrendentials to work
with the signing as the signature does not specify it.

We could define a Signing interface and make some of our AuthCrendentials credentials implement it.
[...]

PR #854, opened 2016-04-05
Add ServiceAccountSigner interface, enable signing for AE credentials

mziccard:
This PR follows from #701:
- Add `SigningAuthCredentials` interface
- Make `AppEngineAuthCredentials` extend `SigningAuthCredentials` and enable signing on AE
- Make `AppEngineAuthCredentials` public as other AuthCredentials classes
- Change `createApplicationDefaults` to return `ServiceAccountAuthCredentials` if a service account is available
- Add constructor to `ApplicationDefaultAuthCredentials` that takes `GoogleCredentials`

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
@@ -25,8 +25,13 @@ import com.google.auth.oauth2.ServiceAccountCredentials;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Serializable;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
+import java.security.Signature;
+import java.security.SignatureException;
 import java.util.Collection;
 import java.util.Objects;
 
@@ -35,16 +40,26 @@ import java.util.Objects;
  */
 public abstract class AuthCredentials implements Restorable<AuthCredentials> {
 
-  private static class AppEngineAuthCredentials extends AuthCredentials {
+  /**
+   * Represents built-in credentials when running in Google App Engine.
+   */
+  public static class AppEngineAuthCredentials extends AuthCredentials
+      implements ServiceAccountSigner {
 
     private static final AuthCredentials INSTANCE = new AppEngineAuthCredentials();
     private static final AppEngineAuthCredentialsState STATE = new AppEngineAuthCredentialsState();
 
-    private static class AppEngineCredentials extends GoogleCredentials {
+    private AppEngineCredentials credentials;
+
+    private static class AppEngineCredentials extends GoogleCredentials
+        implements ServiceAccountSigner {
 
       private final Object appIdentityService;
+      private final String account;
       private final Method getAccessToken;
       private final Method getAccessTokenResult;
+      private final Method signForApp;
+      private final Method getSignature;
       private final Collection<String> scopes;
 
       AppEngineCredentials() {
@@ -59,6 +74,12 @@ public abstract class AuthCredentials implements Restorable<AuthCredentials> {
               "com.google.appengine.api.appidentity.AppIdentityService$GetAccessTokenResult");
           this.getAccessTokenResult = serviceClass.getMethod("getAccessToken", Iterable.class);
           this.getAccessToken = tokenResultClass.getMethod("getAccessToken");
+          this.account = (String) serviceClass.getMethod("getServiceAccountName")
+              .invoke(appIdentityService);
+          this.signForApp = serviceClass.getMethod("signForApp", byte[].class);
+          Class<?> signingResultClass = Class.forName(
+              "com.google.appengine.api.appidentity.AppIdentityService$SigningResult");
+          this.getSignature = signingResultClass.getMethod("getSignature");
           this.scopes = null;
         } catch (Exception e) {
           throw new RuntimeException("Could not create AppEngineCredentials.", e);
@@ -69,11 +90,14 @@ public abstract class AuthCredentials implements Restorable<AuthCredentials> {
         this.appIdentityService = unscoped.appIdentityService;
         this.getAccessToken = unscoped.getAccessToken;
         this.getAccessTokenResult = unscoped.getAccessTokenResult;
+        this.account = unscoped.account;
+        this.signForApp = unscoped.signForApp;
+        this.getSignature = unscoped.getSignature;
         this.scopes = scopes;
       }
 
       /**
-       * Refresh the access token by getting it from the App Identity service
+       * Refresh the access token by getting it from the App Identity service.
        */
       @Override
       public AccessToken refreshAccessToken() throws IOException {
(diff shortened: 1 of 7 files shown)


gcloud-java-core/src/main/java/com/google/gcloud/SigningAuthCredentials.java
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gcloud;
+
+/**
+ * Interface for authentication credentials that support signing.
```
aozarov: Do you really think we should mention here (Javadoc, name,..) that it is intended only for Credentials?

I am not convinced, and therefore would have called it with a name more generic (e.g. `Signer` or `ServiceAccountSigner`) and document it as such (after all one could provide an instance of it for the signUrl
params and without supporting all the other auth credentials requirements). 

Otherwise, any reason for not making it a specific subclass of `AuthCredentials`?

My preference would be the former.

gcloud-java-core/src/main/java/com/google/gcloud/SigningAuthCredentials.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface SigningAuthCredentials {
+
+  class SigningException extends RuntimeException {
+
+    SigningException(String message, Exception cause) {
```
aozarov: serialVersionUID (and add to serialization tests).

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -765,13 +768,13 @@ public static SignUrlOption withMd5() {
...
-     * Service account credentials which are used for signing the URL.
-     * If not provided an attempt will be made to get it from the environment.
+     * Signing credentials which should be used to sign the URL. If not provided an attempt will be
+     * made to get it from the environment.
      *
      * @see <a href="https://cloud.google.com/storage/docs/authentication#service_accounts">Service
      *     account</a>
      */
-    public static SignUrlOption serviceAccount(ServiceAccountAuthCredentials credentials) {
+    public static SignUrlOption serviceAccount(SigningAuthCredentials credentials) {
```
aozarov: Java doc may change per my earlier suggestion.

Also, I think method name should change.

gcloud-java-core/src/main/java/com/google/gcloud/SigningAuthCredentials.java
mziccard: For the purpose of signing URLs as well as policies (which we don't have know) we also need the `account()` value, do you think it makes sense to expose that in an interface called `Signer`?

gcloud-java-storage/src/main/java/com/google/gcloud/storage/StorageImpl.java
```diff
@@ -538,15 +533,13 @@ public URL signUrl(BlobInfo blobInfo, long duration, TimeUnit unit, SignUrlOptio
...
-        (ServiceAccountAuthCredentials) optionMap.get(SignUrlOption.Option.SERVICE_ACCOUNT_CRED);
-    ServiceAccountCredentials cred = authCred != null ? authCred.credentials() : null;
-    if (authCred == null) {
-      checkArgument(
-          this.options().authCredentials() != null
-          && this.options().authCredentials().credentials() instanceof ServiceAccountCredentials,
+    SigningAuthCredentials authCredentials =
+        (SigningAuthCredentials) optionMap.get(SignUrlOption.Option.SERVICE_ACCOUNT_CRED);
+    if (authCredentials == null) {
+      checkArgument(this.options().authCredentials() != null
```
aozarov: the `!=null` check is redundant.

gcloud-java-core/src/main/java/com/google/gcloud/SigningAuthCredentials.java
aozarov: Yes, though maybe `ServiceAccountSigner` would be more appropriate...?

mziccard: Yes this makes sense, I'll go with `ServiceAccountSigner`.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
```diff
@@ -23,12 +23,15 @@
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.gcloud.AuthCredentials;
+import com.google.gcloud.AuthCredentials.AppEngineAuthCredentials;
```
aozarov: Is that import necessary?

gcloud-java-core/src/main/java/com/google/gcloud/AuthCredentials.java
```diff
@@ -180,22 +234,45 @@ public boolean equals(Object obj) {
     ServiceAccountAuthCredentials(String account, PrivateKey privateKey) {
       this.account = checkNotNull(account);
       this.privateKey = checkNotNull(privateKey);
+      this.credentials = new ServiceAccountCredentials(null, account, privateKey, null, null);
```
aozarov: or `this(new ServiceAccountCredentials(null, account, privateKey, null, null));`
and adding `checkNotNull` to the other constructor `account` and `privateKey` assignments.

aozarov:
Nice work! I really like the way you handled the default credentials.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Storage.java
mziccard: The import is there to avoid fully qualifying the class in `signUrl` javadoc

mziccard:
Comments fixed in 08e23ff

aozarov:
Looks great. Few more suggestions.

gcloud-java-core/src/test/java/com/google/gcloud/SerializationTest.java
```diff
@@ -52,6 +53,8 @@ public Builder toBuilder() {
   private static final Identity IDENTITY = Identity.allAuthenticatedUsers();
   private static final PageImpl<String> PAGE =
       new PageImpl<>(null, "cursor", ImmutableList.of("string1", "string2"));
+  private static final SigningException SIGNING_EXCEPTION =
```
mziccard: Codacy Issue found: [Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.](https://www.codacy.com/app/mziccard/gcloud-java/file/2361106585/issues/source?bid=3123103&fileBranchId=3219607#l56)

mziccard: Codacy Issue found: [Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.](https://www.codacy.com/app/mziccard/gcloud-java/file/2362596895/issues/source?bid=3123103&fileBranchId=3219607#l56)

mziccard:
Latest comments fixed in 72ae3cb

gcloud-java-storage/src/main/java/com/google/gcloud/storage/Blob.java
```diff
@@ -470,20 +470,26 @@ public WriteChannel writer(BlobWriteOption... options) {
    * {@link StorageOptions} do not implement {@link ServiceAccountSigner} (this is the case for
    * Compute Engine credentials and Google Cloud SDK credentials) then {@code signUrl} will throw an
    * {@link IllegalArgumentException} unless an implementation of {@link ServiceAccountSigner} is
```
aozarov: IllegalArgumentException?

gcloud-java-core/src/test/java/com/google/gcloud/SerializationTest.java
aozarov: I think we can fix that.

aozarov:
Looks good. One ( x 2) Javadoc fix, and I think we can fix the codacy complaint as well.
Feel free to merge after.

tmatsuo:
:+1: Thanks!

Issue #819, opened 2016-03-29
Write docs on steps necessary for supporting a new service.


ajkannan:
This readme should contain items such as (but not limited to):

* Splitting code into distinct SPI and API layer
* Writing local test helpers if an emulator doesn't already exist (and discuss standards for a test helper/emulator)
* Documentation (package info, step-by-step guide, other docs that need to be updated)
* Adding example code to gcloud-java-examples
* Making sure any versioned jars/docs/etc are included in the update_pom_version.sh script
* Add the example to the list of examples in gcloud-java-examples/pom.xml in the assembler plugin
* Look at the SLA to determine the appropriate default retry settings (see #857)
* Be aware that library version in READMEs is auto-updated by utilities/update_docs_version.sh when releasing.
* Mention that code should be added to [gcloud-java-examples/test-apps](https://github.com/GoogleCloudPlatform/gcloud-java-examples/tree/master/test-apps) for each new service. 

This doc can be linked from the CONTRIBUTING docs page.

PR #858, opened 2016-04-05
Cleanup build scripts

### LINKED ISSUE #841: Update order of after_success tasks for deployment
We should push website and readme version updates after the artifacts have successfully been pushed.  Integration tests (part of the deployment step) sometimes fail because of flakiness, and it's wasted effort to push a new version of the website send a PR to update the readme versions when Travis will have to be rerun.

ajkannan:
The following changes were made:
- Do not deploy artifacts when SITE_VERSION is empty (due to maven plugin failure when extracting the version)
- Fixes #841 by reordering tasks in after_success.sh and also skipping ITs in deployment
- Makes the max depth for README/pom.xml version updates 2 to accommodate modules under `gcloud-java-contrib`
- Skips CI for `gh-pages` push


ajkannan:
I'll make a note in #819 to note that auto-version updating occurs, and that any new READMEs should be written to conform to avoid bugs/not being updated OR the script should be updated.

ajkannan:
Added message before exit and made `find` in update_docs_version.sh more specific.  There is an -o option, but that can only be used with "tests" on a filename, not options that specify how find generates a list of filenames (e.g. -maxdepth).  As a result, I changed the find statement to be a bit different than discussed.

RELEASING.md
```diff
@@ -10,6 +10,8 @@ This script takes an optional argument denoting the new version.  By default, if
 2. Create a PR to update the pom.xml version.  If releasing a new client library, this PR should also update javadoc grouping in the base directory's [pom.xml](./pom.xml).
 PRs that don't release new modules should look something like [#225](https://github.com/GoogleCloudPlatform/gcloud-java/pull/225).  PRs that do release a new module should also add the appropriate packages to the javadoc groups "SPI" and "Test helpers", as shown in [#802](https://github.com/GoogleCloudPlatform/gcloud-java/pull/802) for `gcloud-java-dns`. After this PR is merged into GoogleCloudPlatform/gcloud-java, Travis CI will push a new website to GoogleCloudPlatform/gh-pages, push a new artifact to the Maven Central Repository, and update versions in the README files.
 
+  Important: Do not merge in any non-release-related pull requests between the start of step 2 and the end of step 6.  Between these steps, the project version is a non-snapshot version, so any commits to the master branch will cause the artifacts to be redeployed and the website to be regenerated via Travis.
```
mziccard: Can artifacts in the Central Repository actually be redeployed? I thought that any attempt at deploying an existing version would fail.

ajkannan: You're right; I fixed the releasing doc.

RELEASING.md
```diff
@@ -8,9 +8,9 @@ Most of the release process is handled by the `after_success.sh` script, trigger
...
+3. Before moving on, verify that the artifacts have successfully been pushed to the Maven Central Repository.  Open Travis CI, click the ["Build History" tab](https://travis-ci.org/GoogleCloudPlatform/gcloud-java/builds), and open the second build's logs for Step 2's PR.  Be sure that you are not opening the "Pull Request" build logs.  When the build finishes, scroll to the end of the log and verify that the artifacts were successfully staged and deployed.  Search for `gcloud-java` on the [Sonatype website](https://oss.sonatype.org/#nexus-search;quick~gcloud-java) and check the latest version number.  In rare cases (when the Maven plugin that determines the version of the repository fails), the artifacts may not be deployed even if the version in the pom.xml files don't contain `SNAPSHOT`.  If the artifacts weren't deployed due to invalid version parsing or a flaky test, rerun the build.
```
mziccard: "even if the version in the pom.xml files don't" I believe don't should be doesn't

utilities/update_docs_version.sh
```diff
@@ -10,7 +10,7 @@ RELEASED_VERSION=$(mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate
 if [ "${RELEASED_VERSION##*-}" != "SNAPSHOT" ]; then
     echo "Changing version to $RELEASED_VERSION in README files"
     # Get list of directories for which README.md must be updated
-    module_folders=($(find . -maxdepth 1 -name 'gcloud-java*' -type d) .)
+    module_folders=($(find . -maxdepth 2 \( -regex "./gcloud-java-[a-z]*" -o -regex "./gcloud-java-contrib/gcloud-java-[a-z]*" \) -type d) . ./gcloud-java)
```
mziccard: I believe with this we miss `gcloud-java-contrib/gcloud-java-nio-examples`, am I wrong or is this intended?

ajkannan:
Fixed the regex, fixed the RELEASING doc, and check if the artifact was deployed before attempting to deploy.  The last item is so that if the deployment succeeds, but for some reason Travis times out or the website deployment errors, we will be able to rerun the script without failing due to the deployment failing (from redeployment).

utilities/after_success.sh
```diff
@@ -12,8 +12,20 @@ if [ "${TRAVIS_JDK_VERSION}" == "oraclejdk7" ]; then
...
+        echo "Used the maven-help-plugin to determine that the version is $SITE_VERSION"
+        if [ "$SITE_VERSION" == "" ]; then
+            echo "Could not determine the version, so we're exiting."
+            exit 1
+        fi
         if [ "${SITE_VERSION##*-}" != "SNAPSHOT" ]; then
-            # Deploy site if not a SNAPSHOT
+            # Deploy Maven artifacts (if they don't exist yet) and update artifact version in READMEs.
+            ARTIFACT_EXISTS=$(mvn wagon:exist -Dwagon.url=https://oss.sonatype.org/content/repositories/releases/com/google/gcloud/gcloud-java/$SITE_VERSION)
+            if [[ "$ARTIFACT_EXISTS" == *"does not exists."* -a "$ARTIFACT_EXISTS" != "" ]]; then
```
mziccard: Is this line actually working? If I remember correctly `-a` does not work with `[[` (I might be wrong). Something like the following should work:
```bash
if [[ "$ARTIFACT_EXISTS" == *"does not exists."* &&  "$ARTIFACT_EXISTS" != "" ]]; then
```
Also is the second check `"$ARTIFACT_EXISTS" != ""` ever used?

ajkannan:
Added a commit that changes the if statement to decide whether or not to deploy.

utilities/update_pom_version.sh
```diff
@@ -11,8 +11,7 @@
 # Get the previous maven project version.
 CURRENT_VERSION=$(mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version | grep -Ev '(^\[|\w+:)')
 # Get list of directories for which pom.xml must be updated
-module_folders=($(find . -maxdepth 1 -name 'gcloud-java*' -type d) .)
-
+module_folders=($(find . -maxdepth 2 \( -regex "./gcloud-java[-[a-z]+]+" -o -regex "./gcloud-java-contrib/gcloud-java[-[a-z]+]+" \) -type d) . ./gcloud-java)
```
aozarov: Does maxdepth needs to change now when you added the "or" expression?

utilities/after_success.sh
```diff
@@ -12,8 +12,22 @@ if [ "${TRAVIS_JDK_VERSION}" == "oraclejdk7" ]; then
...
         SITE_VERSION=$(mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version | grep -Ev '(^\[|\w+:)')
+        echo "Used the maven-help-plugin to determine that the version is $SITE_VERSION"
+        if [ "$SITE_VERSION" == "" ]; then
+            echo "Could not determine the version, so we're exiting."
+            exit 1
+        fi
         if [ "${SITE_VERSION##*-}" != "SNAPSHOT" ]; then
-            # Deploy site if not a SNAPSHOT
+            # Deploy Maven artifacts (if they don't exist yet) and update artifact version in READMEs.
+            URL=https://oss.sonatype.org/content/repositories/releases/com/google/gcloud/gcloud-java/$SITE_VERSION
```
aozarov: Looks like it works as is but did you consider to add a trailing '/' ?

utilities/after_success.sh
```diff
@@ -12,8 +12,22 @@ if [ "${TRAVIS_JDK_VERSION}" == "oraclejdk7" ]; then
...
+            if curl --output /dev/null --silent --head --fail "$URL"; then
+                echo "Not deploying artifacts because it seems like they already exist."
```
aozarov: I would also print the URL

aozarov:
Looks good to me few minor comment/questions. Feel free to merge after considering them.

utilities/update_pom_version.sh
ajkannan: -maxdepth controls how far find looks for potential matching files (rather than being a "test" condition on whether or not to include a file).  For that reason I believe it has to be kept as 2.

utilities/after_success.sh
ajkannan: Added.

utilities/after_success.sh
ajkannan: Done.

ajkannan:
Addressed @aozarov's comments, will wait on hearing from @mziccard before merging.

ajkannan:
Found a regex bug. I'm going to do some more in-depth testing before removing the don't merge label.

ajkannan:
Fixed bug and did a bit more testing.  Any comments/suggestions @mziccard?

utilities/update_docs_version.sh
```diff
@@ -10,7 +10,7 @@ RELEASED_VERSION=$(mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate
 if [ "${RELEASED_VERSION##*-}" != "SNAPSHOT" ]; then
     echo "Changing version to $RELEASED_VERSION in README files"
     # Get list of directories for which README.md must be updated
-    module_folders=($(find . -maxdepth 1 -name 'gcloud-java*' -type d) .)
+    module_folders=($(find . -maxdepth 2 \( -regex "./gcloud-java[-[a-z]+]+*" -o -regex "./gcloud-java-contrib/gcloud-java[-[a-z]+]+*" \) -type d) . ./gcloud-java)
```
mziccard: Not sure what's wrong but if I run

```bash
module_folders=($(find . -maxdepth 2 \( -regex "./gcloud-java[-[a-z]+]+*" -o -regex "./gcloud-java-contrib/gcloud-java[-[a-z]+]+*" \) -type d) . ./gcloud-java)
echo $module_folders
```
in the root of `gcloud-java` I get only `.` as output. Am I doing something wrong?

ajkannan: Try `echo ${module_folders[*]}` for the second line - does that give you all the results?

mziccard: I am running:
```bash
module_folders=($(find . -maxdepth 2 \( -regex "./gcloud-java[-[a-z]+]+*" -o -regex "./gcloud-java-contrib/gcloud-java[-[a-z]+]+*" \) -type d) . ./gcloud-java)
echo ${module_folders[*]}
```
And what I get is:
```
. ./gcloud-java
```
For completeness `find . -maxdepth 1 -type d` returns:
```
.
./.git
./.settings
./gcloud-java
./gcloud-java-bigquery
./gcloud-java-contrib
./gcloud-java-core
./gcloud-java-datastore
./gcloud-java-dns
./gcloud-java-examples
./gcloud-java-resourcemanager
./gcloud-java-storage
./src
./target
./utilities		
```

ajkannan: Hm, when I try those two same lines, I get
```
./gcloud-java-dns ./gcloud-java-contrib ./gcloud-java-contrib/gcloud-java-nio ./gcloud-java-storage ./gcloud-java-examples ./gcloud-java-datastore ./gcloud-java-bigquery ./gcloud-java-pubsub ./gcloud-java-resourcemanager ./gcloud-java-compute ./gcloud-java-core . ./gcloud-java
```
It seems like the find returning an empty result.  I've probably used something platform dependent (are you using mac by any chance?), but I imagine that's okay because our Travis builds are on Debian.

ajkannan: Actually, we do run utilities/update_pom_version.sh manually to trigger releases, so being able to run on both mac and linux would be nice.  At @aozarov's suggestion, I will change this line to:

```
module_folders=($(find . -maxdepth 2 -type d | sed -E -n "/^(\.|\.\/gcloud-java-contrib)\/gcloud-java(-[a-z]+)+$/p") . ./gcloud-java)
```

ajkannan:
@mziccard, the module listing command has been changed so that it works on mac and linux

mziccard:
Looks good to me. You'll have to rebase to fix conflicts. While at it consider squashing the commits.

ajkannan:
Rebased and squashed commits

PR #860, opened 2016-04-06
Add default retry params that align with SLA

### LINKED ISSUE #857: When using default retry settings we should get that default from the specific service options
As service may provide a specific exponential back-off requirements in its SLA.
E.g. for Datastore see [this](https://cloud.google.com/datastore/sla).

We should add a protected method in `ServiceOptions` to return the defaultRetryParams which 
will be used when user didn't explicitly provided one. Default implementation should return `RetryParams.defaultInstance()`.

We should override this method in services that provide their own specific retry settings such as Datastore and Storage.

ajkannan:
Fixes #857 

Some notes/questions:
- I've changed the retry settings for services that explicitly state the back-off requirement.  However, since they're all the same, do you think that's a general recommendation?  If so, do you think we should change RetryParams.defaultInstance() instead?
- I've also increased the total retry period to 80000, since 1 + 2 + 4 + 8 + 16 + 32 = 63 seconds, which is greater than the 50 default

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/gcloud/RetryParams.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/RetryParams.java
@@ -165,7 +165,7 @@ public final class RetryParams implements Serializable {
     /**
      * Sets totalRetryPeriodMillis. Note that App Engine Standard Environment front-end modules have
      * a 60 second deadline for HTTP requests. For that reason, you should set the total retry
-     * period to under 60 seconds if you are using the App Engine front-end module.
+     * period to under 60 seconds if you are using it on an App Engine front-end module.
      *
      * @param totalRetryPeriodMillis the totalRetryPeriodMillis to set
      * @return the Builder for chaining
--- a/gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java
+++ b/gcloud-java-core/src/main/java/com/google/gcloud/ServiceOptions.java
@@ -655,6 +655,11 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
 
   public abstract <B extends Builder<ServiceT, ServiceRpcT, OptionsT, B>> B toBuilder();
 
+  /**
+   * Some services may have different backoff requirements listed in their SLAs. Be sure to override
+   * this method in options subclasses when the service's backoff requirement differs from the
+   * default parameters listed in {@link RetryParams}.
+   */
   protected RetryParams defaultRetryParams() {
     return RetryParams.defaultInstance();
   }


mziccard:
> However, since they're all the same, do you think that's a general recommendation?

I can't be sure of this but given that all params look the same I would tend to avoid replicating them in all services. Also this new retry params should work also for other services.

ajkannan:
Took into account feedback; ready for another pass.

gcloud-java-core/src/main/java/com/google/gcloud/RetryParams.java
```diff
@@ -62,6 +66,9 @@
   private final double retryDelayBackoffFactor;
   private final long totalRetryPeriodMillis;
 
+  // Some services may have different backoff requirements listed in their SLAs. Be sure to override
```
aozarov: I would move that comment to ServiceOptions.defaultRetryParams (you can *also* keep it here if you want).

gcloud-java-core/src/main/java/com/google/gcloud/RetryParams.java
```diff
@@ -48,12 +48,16 @@
 
   private static final long serialVersionUID = -8492751576749007700L;
 
+  /**
+   * Note that App Engine Standard Environment front-end modules have a 60 second deadline for HTTP
+   * requests. For that reason, we set the default total retry period to less than 60 seconds.
+   */
+  public static final long DEFAULT_TOTAL_RETRY_PERIOD_MILLIS = 50_000L;
```
aozarov: I am fine with that though I thought your preference was to make it 80_000 *and* add to the APIs javadoc the suggestion to lower it if planned to be used on an AE front-end.

gcloud-java-core/src/main/java/com/google/gcloud/RetryParams.java
```diff
@@ -156,7 +163,9 @@ public Builder retryDelayBackoffFactor(double retryDelayBackoffFactor) {
     }
 
     /**
-     * Sets totalRetryPeriodMillis.
+     * Sets totalRetryPeriodMillis. Note that App Engine Standard Environment front-end modules have
+     * a 60 second deadline for HTTP requests. For that reason, you should set the total retry
+     * period to under 60 seconds if you are using the App Engine front-end module.
```
aozarov: maybe "using it on App Engine front-end module"?

aozarov:
LGTM. Feel free to merge after considering the Javadoc suggestions.

gcloud-java-core/src/main/java/com/google/gcloud/RetryParams.java
ajkannan: I thought about it some more and decided to stick with 50 seconds because users may not even think to read the RetryParams javadoc, since we set it by default.  So even though it's a particular use case that may result in only slightly more errors for users, I decided to leave it as is.

PR #864, opened 2016-04-06
Fix docs on which apis need to be enabled for Storage

### LINKED ISSUE #790: Update example documentation with correct APIs to enable (if necessary)
Check whether we need to tell people to enable both"Google Cloud Storage" and "Google Cloud Storage JSON API". If so, update the docs.

Context: [#714 (comment)](https://github.com/GoogleCloudPlatform/gcloud-java/pull/714#discussion_r57408959).

ajkannan:
Fixes #790

### DIFF
--- a/gcloud-java-examples/README.md
+++ b/gcloud-java-examples/README.md
@@ -82,7 +82,7 @@ To run examples from your command line:
 
   * Here's an example run of `StorageExample`.
 
-    Before running the example, go to the [Google Developers Console][developers-console] to ensure that "Google Cloud Storage" and "Google Cloud Storage API" are enabled and that you have a bucket.  Also ensure that you have a test file (`test.txt` is chosen here) to upload to Cloud Storage stored locally on your machine.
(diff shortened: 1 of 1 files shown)


gcloud-java-examples/README.md
```diff
@@ -82,7 +82,7 @@ To run examples from your command line:
 
   * Here's an example run of `StorageExample`.
 
-    Before running the example, go to the [Google Developers Console][developers-console] to ensure that Google Cloud Storage API is enabled and that you have a bucket.  Also ensure that you have a test file (`test.txt` is chosen here) to upload to Cloud Storage stored locally on your machine.
+    Before running the example, go to the [Google Developers Console][developers-console] to ensure that "Google Cloud Storage" and "Google Cloud Storage API" are enabled and that you have a bucket.  Also ensure that you have a test file (`test.txt` is chosen here) to upload to Cloud Storage stored locally on your machine.
```
mziccard: Do you think it's worth to use "Google Cloud Storage JSON API" to be consistent with how it's called in the console?

ajkannan: Thanks for catching that, fixed.

PR #869, opened 2016-04-07
Add getting-started-java repo to releasing docs

ajkannan:
I also updated the doc to mention the Bookshelf app needs to be updated.


aozarov:
LGTM

PR #871, opened 2016-04-07
Fix checkstyle issues in gcloud-java-core

mziccard:


gcloud-java-core/src/main/java/com/google/gcloud/spi/ServiceRpcFactory.java
```diff
@@ -19,10 +19,8 @@
 import com.google.gcloud.ServiceOptions;
 
 /**
- * A base interface for all service RPC factories.
- *
- * Implementation must provide a public no-arg constructor.
- * Loading of a factory implementation is done via {@link java.util.ServiceLoader}.
+ * A base interface for all service RPC factories. Implementation must provide a public no-arg
```
ajkannan: Nit unrelated to this PR: s/Implementation/Implementations/

ajkannan:
Related to core cleanup, I'm seeing testConstructor() and testChunkSize() in BaseWriterChannelTest throwing IOExceptions that may be unnecessary; do you think we can remove there here?

ajkannan:
Feel free to merge after considering those two comments.

gcloud-java-core/src/main/java/com/google/gcloud/spi/ServiceRpcFactory.java
ajkannan: Actually, fine as is, no need to change it.

PR #875, opened 2016-04-07
Remove unnecessary project ID normalization

ajkannan:
Normalization is no longer necessary because Datastore v1beta3 doesn't expose any prefixes to the project ID (e.g. "s~" or "e~").


gcloud-java-datastore/src/main/java/com/google/gcloud/datastore/DatastoreOptions.java
```diff
@@ -139,7 +100,7 @@ protected String defaultProject() {
         com.google.datastore.v1beta3.client.DatastoreHelper.PROJECT_ID_ENV_VAR,
         System.getenv(com.google.datastore.v1beta3.client.DatastoreHelper.PROJECT_ID_ENV_VAR));
     if (projectId == null) {
```
aozarov: I think this can be removed.

PR #876, opened 2016-04-08
Speed up builds by using an older version of maven-source-plugin

cgruber:
Speed up builds by using an older version of maven-source-plugin (2.4) that doesn't have speed issues on ubuntu (as does 3.0.0)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -237,7 +237,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-source-plugin</artifactId>
-        <version>3.0.0</version>
+        <version>2.4</version>
         <executions>
           <execution>
             <id>attach-sources</id>


aozarov:
@mziccard opinions?

aozarov:
@cgruber Thank you for doing that! Do you mind to add a comment explaining why we should not upgrade to the newer version. Typically we try to keep our dependencies up-to-date (and using [versioneyey](https://www.versioneye.com/user/projects/56bd8ee72a29ed002d2b0969) for it) and such a comment would help not replacing it by mistake. Also, if there is a reference to a known bug that we can periodically check if it was fixed that would be great.

cgruber:
maven-source-plugin-3.0.0 currently exhibits the zipping behavior on ubuntu that was seen in maven-jar-plugin prior to version 2.5 (https://issues.apache.org/jira/browse/MJAR-188).

In maven-jar-plugin, it's use of plexus-io was getting stuck in java7+'s implementation of getgrid, which it did on every file. This was fixed in plexus-io-1.0.5 (but report no longer available - auto-forwards to github but the issues haven't been migrated).

I'm can't be sure (without more investigation I don't have time for) that maven-source-plugin is encountering the same bug, but it's the same behavior, and it seems that it is newly regressed in 3.0.0.  Unless you need 3.0.0 functionality, prefer 2.4 until such time as this is addressed in a subsequent 3.x line, or unless you are willing to take the hit to obtain some feature in the 3.x line.

FWIW, you should not generally be building sources except on release (or snapshot deploy).  There's no need.  Regardless of this fix, I would put generation of javadoc and source jars into a profile only used for release, as it slows down the build for no particular benefit.

mziccard:
This looks good to me. We are not using any particular feature of the latest maven-source-plugin so using an earlier version should be fine. I'll verify if the issue still holds once the next version is released. We should probably file a bug if it hasn't been done yet.

aozarov:
Ok, @cgruber I was suggesting to put your comment "maven-source-plugin-3.0.0 currently exhibits the zipping behavior on ubuntu that was seen in maven-jar-plugin prior to version 2.5 (https://issues.apache.org/jira/browse/MJAR-188)" in the pom, so one would not upgrade it without being aware of why it is kept that way.

@mziccard filing a bug or pinging the existing bug (associated with jar plugin) and suggesting that it also affects the source plugin would be nice. Also, do you want to take @cgruber suggestion for creating a release profile and move the source+javadoc generation/jar to it (or create an issue for it)?

mziccard:
Opened issue #888 for creating a release profile

aozarov:
@cgruber if you are too busy and would rather have @mziccard copy your changes to #889 and add the comment (so we will not revert it by accident before the Jira issue is resolved) let us know.

cgruber:
I can do it. Sorry.  I have such a flood of github notifications, I didn't see this one.

mziccard:
I took the liberty to cherry-pick @cgruber commit, rebase and add a comment as per @aozarov request (see #961).

I tried to reproduce this problem (Java 7, Ubuntu 14.04 LTS) but failed. Could you please provide me with more details on a failing setup so that I can file a (better) issue? /cc @jean-philippe-martin

Issue #816, opened 2016-03-29
Create a retry helper that doesn't retry on socket timeouts


ajkannan:
For non-reentrant operations (e.g. creating resources and applying a change request), we should have a separate retry handler that won't retry on socket timeouts.  This is because a socket timeout error can occur after a request was successfully sent to the service.  This would be helpful in DNS (see issue #808), but may also be helpful in other services.

aozarov:
This was fixed by #883.

PR #883, opened 2016-04-08
Marked exceptions for create and deletes in DNS as non-idempotent.

### LINKED ISSUE #808: Investigate flaky DNS integration test

mderka:
Fixes #808, relates to #816 .

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/cloud/BaseServiceException.java
+++ b/gcloud-java-core/src/main/java/com/google/cloud/BaseServiceException.java
@@ -110,12 +110,11 @@ public class BaseServiceException extends RuntimeException {
     String reason = null;
     String location = null;
     String debugInfo = null;
-    Error error = null;
     Boolean retryable = null;
     if (exception instanceof GoogleJsonResponseException) {
       GoogleJsonError jsonError = ((GoogleJsonResponseException) exception).getDetails();
       if (jsonError != null) {
-        error = new Error(jsonError.getCode(), reason(jsonError));
+        Error error = new Error(jsonError.getCode(), reason(jsonError));
         code = error.code;
         reason = error.reason;
         retryable = isRetryable(idempotent, error);
--- a/gcloud-java-dns/src/main/java/com/google/cloud/dns/spi/DefaultDnsRpc.java
+++ b/gcloud-java-dns/src/main/java/com/google/cloud/dns/spi/DefaultDnsRpc.java
@@ -61,7 +61,8 @@ public class DefaultDnsRpc implements DnsRpc {
           .setFields(FIELDS.getString(options))
           .execute();
     } catch (IOException ex) {
-      throw translate(ex, false);
+      // todo this can cause misleading report of a failure, intended to be fixed within #924
+      throw translate(ex, true);
     }
   }
 


mderka:
@aozarov PTAL

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -147,15 +162,15 @@ public BaseServiceException(int code, String message, String reason, boolean ide
...
-    return error != null && error(error).isRetryable(retryableErrors());
+  protected boolean isRetryable(boolean idempotent, GoogleJsonError error) {
+    return error != null && error(error).isRetryable(idempotent, retryableErrors());
   }
 
-  protected boolean isRetryable(IOException exception) {
+  protected boolean isRetryable(boolean idempotent, IOException exception) {
     if (exception instanceof GoogleJsonResponseException) {
-      return isRetryable(((GoogleJsonResponseException) exception).getDetails());
+      return isRetryable(idempotent, (((GoogleJsonResponseException) exception).getDetails()));
```
mziccard: Codacy Issue found: [Useless parentheses.](https://www.codacy.com/app/mziccard/gcloud-java/file/2381910367/issues/source?bid=3123103&fileBranchId=3225961#l171)

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -62,16 +68,25 @@ public Integer code() {
...
     }
 
-    boolean isRetryable(Set<Error> retryableErrors) {
+    boolean isRetryable(boolean idempotent, Set<Error> retryableErrors) {
       for (Error retryableError : retryableErrors) {
-        if ((retryableError.code() == null || retryableError.code().equals(this.code()))
-            && (retryableError.reason() == null || retryableError.reason().equals(this.reason()))) {
+        boolean match = (retryableError.code() == null || retryableError.code().equals(this.code()))
+            && (retryableError.reason() == null || retryableError.reason().equals(this.reason()));
+        if (match && idempotent || match && retryableError.rejected()) {
```
aozarov: I would leave the condition as it was and just instead of `return true` have `return idempotent || retryableError.rejected()`

There is no reason to continue the loop once we have a match.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -123,7 +138,7 @@ public BaseServiceException(GoogleJsonError error, boolean idempotent) {
     this.code = error.getCode();
```
aozarov: I wonder if this one could call `this(error.getCode(), error.getMessage(), reason(error), idempotent)` instead?

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -147,15 +162,15 @@ public BaseServiceException(int code, String message, String reason, boolean ide
     return Collections.emptySet();
   }
 
-  protected boolean isRetryable(GoogleJsonError error) {
-    return error != null && error(error).isRetryable(retryableErrors());
+  protected boolean isRetryable(boolean idempotent, GoogleJsonError error) {
```
aozarov: Instead of this method, which I think we should probably embed in the `isRetryable(boolean, IOException)` I think we should have `protected boolean isRetryable(boolean idempotent, Error error)`. default implementation would just delegated to `error.isRetryable(idempotent, retryableErrors())`.

Having that I would change the first constructor and the last constructor to call it instead.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsException.java
```diff
@@ -31,16 +31,16 @@
...
-      new Error(500, null),
-      new Error(502, null),
-      new Error(503, null),
-      new Error(null, "userRateLimitExceeded"),
-      new Error(null, "rateLimitExceeded"));
+      new Error(429, null, true),
+      new Error(500, null, false),
+      new Error(502, null, false),
+      new Error(503, null, false),
+      new Error(null, "userRateLimitExceeded", true),
```
aozarov: can these reasons be part of the previous error codes? If so, this is why you made the loop in Error iterate over all values?

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
aozarov: We can fix that.

aozarov:
Looks good! Few comments/suggestions for further cleanup.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Fixed.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/DnsException.java
mderka: I am not sure. Both node and ruby are handling these explicitly without code. We cannot simulate these errors ourselves. We concluded before that this was a way to go.

No, it was not the reason. The match will be reported either way.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Done.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Seems viable. However, then we do not actually use the suggested `protected boolean isRetryable(boolean idempotent, Error error)` anywhere. Is there a reason for keeping it?

mderka:
All done. PTAL

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -48,10 +48,16 @@
 
     private final Integer code;
     private final String reason;
+    private final boolean rejected;
 
     public Error(Integer code, String reason) {
+      this(code, reason, false);
+    }
+
+    public Error(Integer code, String reason, boolean rejected) {
```
mziccard: Should we use this constructor also in other services when it applies? For instance also in storage we retry on `429—Too Many Requests`.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -62,17 +68,25 @@ public Integer code() {
     }
 
     /**
+     * Returns true if the error indicates that the API call was certainly not accepted by the
+     * server.
```
mziccard: Maybe an example could help users understand what `rejected()` means? We could mention that `rejected()` can be `true` for rate limit exceeded exceptions.

garrettjonesgoogle:
Hi,

I wanted to remark that create and delete are actually idempotent, and changing them to non-idempotent miscategorizes them. Create is idempotent because after the first or subsequent calls, the result is already "X exists"; Delete is idempotent because after the first or subsequent calls, the result is "X does not exist". The success or error returned to the client is a separate issue from the idempotency of the operation on the data. 

Here is an alternative that is slightly more complex, but would probably do what the client expects:

* On the first call, if the result is ALREADY_EXISTS, then fail with that.
* On the first call, if the result is a different but retryable error code, then do a retry.
* On the retry, if the result is ALREADY_EXISTS, then return success. (The assumption is that the first failure happened after successful creation on the server)

The only case that might be odd is if there is a failure to contact the service, but the entity was already created before, then the result will be a success even though per service semantics it should return ALREADY_EXISTS. I would imagine this should be pretty uncommon though, and should probably only happen when the entity was only very recently created.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
aozarov: We probably should for completeness, however setting up rejected only matters if the operation is not idempotent.
When idempotent is false we will not retry even if there is a match with a retryable exception unless we know for
sure that the server didn't process the exception (`rejected == true`).

In Storage I think all the exception are thrown as `idempotent == true`, so technically this will not change much there.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
aozarov: Right, rate limit exceeded errors should be indicated as rejected.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -138,7 +146,7 @@ public BaseServiceException(int code, String message, String reason, boolean ide
     this.code = code;
     this.reason = reason;
     this.idempotent = idempotent;
-    this.retryable = idempotent && new Error(code, reason).isRetryable(retryableErrors());
+    this.retryable = new Error(code, reason).isRetryable(idempotent, retryableErrors());
```
aozarov: I think this code could be cleaned a little. I sent @mderka a small patch to consider.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
```diff
@@ -187,8 +197,8 @@ public boolean idempotent() {
   }
 
   /**
-   * Returns the service location where the error causing the exception occurred. Returns
-   * {@code null} if not set.
+   * Returns the service location where the error causing the exception occurred. Returns {@code
+   * null} if not set.
```
aozarov: Nit, s/set/available/ ?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
```diff
@@ -61,7 +61,7 @@ public ManagedZone create(ManagedZone zone, Map<Option, ?> options) throws DnsEx
           .setFields(FIELDS.getString(options))
           .execute();
     } catch (IOException ex) {
-      throw translate(ex);
+      throw translate(ex, false);
```
aozarov: see @garrettjonesgoogle comment, which I partially agree with.

I think that means there is some more work to do, however for now I suggest to keep this as `true` and create a separate issue suggesting that some idempotent method may report a wrong exception when retried (such as in this case 'already exists'). 

The reason I think it is OK to retry it because the implication is just misreported error (e.g. 'already exists' instead of 500 or timeout). In either case the user will get an exception.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
```diff
@@ -103,7 +103,7 @@ public boolean deleteZone(String zoneName) throws DnsException {
       dns.managedZones().delete(this.options.projectId(), zoneName).execute();
       return true;
     } catch (IOException ex) {
-      DnsException serviceException = translate(ex);
+      DnsException serviceException = translate(ex, false);
```
aozarov: I think we can safely consider this as idempotent, no?

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
```diff
@@ -148,7 +148,7 @@ public Change applyChangeRequest(String zoneName, Change changeRequest, Map<Opti
           .setFields(FIELDS.getString(options))
           .execute();
     } catch (IOException ex) {
-      throw translate(ex);
+      throw translate(ex, false);
```
aozarov: This change looks reasonable to me.

aozarov:
@garrettjonesgoogle I agree conceptually with what you are suggesting (see my other comment). However for `create` I am not convinced that it is OK to hide from the user the fact that his "Create" was actually successful (it was not there before) or that it was already created (maybe with different values then provided).

I don't think that "The assumption is that the first failure happened after successful creation on the server" is always correct (it could have been created by a parallel call).

garrettjonesgoogle:
@aozarov Good point about the entity possibly being created with different values than the current request. One option there is when the situation is ambiguous, then do a get call to the server to get the existing entity, and determine if the values are the same, and report back appropriately. It's definitely higher overhead and more complex (you will have made a minimum of 3 RPC calls - 1 failed create, one create resulting in already_exists, and 1 get call), but it would be more correct.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
mderka: I do not think so. It would be the same case as create. If a zone exists, it will be deleted with success (200) return value. If the zone does not exist, the service returns 404. So, if the socket times out but the delete is processed, the retry will receive in a subsequent call 404 and return `false` instead of `true`.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Done.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Patch applied. It is nice.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Fixed.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
mderka: I am keeping this one.

gcloud-java-core/src/main/java/com/google/gcloud/BaseServiceException.java
mderka: Create #924.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
mderka: I am switching it back to true. It should be handled within #924.

gcloud-java-dns/src/main/java/com/google/gcloud/dns/spi/DefaultDnsRpc.java
mderka: I am switching it back to true. It should be handled within #924.

mderka:
I created issue #924 to continue with the discussion.

I removed the conflicts and incorporated the comments. Ready to proceed.

PR #889, opened 2016-04-11
Add release profile and use it to deploy artifacts

### LINKED ISSUE #888: Create release profile and move source and javadoc generation/jar to it
This follows the discussion in #876

mziccard:
This fixes #888

### DIFF
--- a/utilities/after_success.sh
+++ b/utilities/after_success.sh
@@ -34,9 +34,9 @@ if [ "${TRAVIS_JDK_VERSION}" == "oraclejdk7" ]; then
 
             cd ..
             utilities/update_docs_version.sh # Update version in READMEs
-            mvn clean deploy --settings ~/.m2/settings.xml -P sign-deploy
+            mvn clean deploy --settings ~/.m2/settings.xml -P sign-deploy,release
         else
-            mvn clean deploy -DskipTests=true -Dgpg.skip=true --settings ~/.m2/settings.xml
+            mvn clean deploy -DskipTests=true -Dgpg.skip=true --settings ~/.m2/settings.xml -P release
         fi
     else
         echo "Not deploying artifacts. This is only done with non-pull-request commits to master branch with Oracle Java 7 builds."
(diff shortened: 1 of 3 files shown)


utilities/verify.sh
```diff
@@ -10,7 +10,7 @@ if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
     chmod 700 $TRAVIS_BUILD_DIR/signing-tools
     tar xvf $TRAVIS_BUILD_DIR/signing-tools.tar -C $TRAVIS_BUILD_DIR/signing-tools
     # Run verify
-    mvn verify
+    mvn verify -P release
```
aozarov: Any reason for this change? I mean what happens if I don't specify "-P release"?
At the verify stage I think it is enough to generate the javadoc (so we can see any javadoc related errors) however
creating a jar for them is not that important (and no need for a source jar).

mziccard: Without the `-P release` javadoc jars are not generated. If javadoc jars are not generated javadoc building is not triggered. Also this allows us to detect whether there are some problems with the packaging plugins before attempting to deploy the artifacts.

utilities/after_success.sh
```diff
@@ -34,9 +34,9 @@ if [ "${TRAVIS_JDK_VERSION}" == "oraclejdk7" ]; then
 
             cd ..
             utilities/update_docs_version.sh # Update version in READMEs
-            mvn clean deploy --settings ~/.m2/settings.xml -P sign-deploy
+            mvn clean deploy --settings ~/.m2/settings.xml -P sign-deploy,release
```
aozarov: I completely forgot how we got to the sign-deploy profile (who defines it and is order matters?) I assume you tested it?

utilities/verify.sh
aozarov: Ok, I am not sure what kind of problem can be caused in the packaging stage (I assumed this is plain jarring and also was not sure that we care to do that on each platform) but I get your point. Fine as is.

aozarov:
Feel free to merge after validating that modified mvn commands work.

utilities/after_success.sh
mziccard: The `sign-deploy` profile is defined in [`settings.xml`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/travis/settings.xml#L31) and contains gpg signing configuration.

Yes I tested the command and gpg configuration is loaded correctly (deploy fails as I don't have any gpc key in the path specified in `settings.xml`). In fact, according to `mvn` docs:
```
-P,--activate-profiles <arg>           Comma-delimited list of profiles
                                       to activate
```

Issue #890, opened 2016-04-11
StackOverflowError mocking our functional objects


mziccard:
There seem to be a problem with `EasyMock` (as well as `Mockito`) when mocking classes that have a `final` `equals()` method that in its body calls other methods on the current object.
This results in a StackOverflowError as the following:
```
java.lang.StackOverflowError
	at java.util.ArrayList$Itr.<init>(ArrayList.java:820)
	at java.util.ArrayList$Itr.<init>(ArrayList.java:820)
	at java.util.ArrayList.iterator(ArrayList.java:814)
	at org.easymock.internal.UnorderedBehavior.addActual(UnorderedBehavior.java:50)
	at org.easymock.internal.MocksBehavior.addActual(MocksBehavior.java:87)
	at org.easymock.internal.ReplayState.invokeInner(ReplayState.java:58)
	at org.easymock.internal.ReplayState.invoke(ReplayState.java:46)
	at org.easymock.internal.MockInvocationHandler.invoke(MockInvocationHandler.java:40)
	at org.easymock.internal.ObjectMethodsFilter.invoke(ObjectMethodsFilter.java:94)
	at org.easymock.internal.ClassProxyFactory$MockMethodInterceptor.intercept(ClassProxyFactory.java:97)
	at com.google.gcloud.storage.Blob$$EnhancerByCGLIB$$aa59c155.toPb(<generated>)
	at com.google.gcloud.storage.Blob.equals(Blob.java:529)
	at org.easymock.internal.ExpectedInvocation.matches(ExpectedInvocation.java:85)
	at org.easymock.internal.UnorderedBehavior.addActual(UnorderedBehavior.java:57)
	at org.easymock.internal.MocksBehavior.addActual(MocksBehavior.java:87)
	at org.easymock.internal.ReplayState.invokeInner(ReplayState.java:58)
	at org.easymock.internal.ReplayState.invoke(ReplayState.java:46)
	at org.easymock.internal.MockInvocationHandler.invoke(MockInvocationHandler.java:40)
	at org.easymock.internal.ObjectMethodsFilter.invoke(ObjectMethodsFilter.java:94)
	at org.easymock.internal.ClassProxyFactory$MockMethodInterceptor.intercept(ClassProxyFactory.java:97)
	at com.google.gcloud.storage.Blob$$EnhancerByCGLIB$$aa59c155.toPb(<generated>)
	at com.google.gcloud.storage.Blob.equals(Blob.java:529)
```
[...]

PR #892, opened 2016-04-11
Enable mocking of functional classes and configurations

mziccard:
This PR removes `final` from functional classes definition and configurations. For those classes this also adds `final` to `equals` and `hashCode` methods if not present. `equals` and `hashCode` are updated to avoid issue #890.

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/CopyJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/CopyJobConfiguration.java
@@ -31,7 +31,7 @@ import java.util.Objects;
  * Google BigQuery copy job configuration. A copy job copies an existing table to another new or
  * existing table. Copy job configurations have {@link JobConfiguration.Type#COPY} type.
  */
-public class CopyJobConfiguration extends JobConfiguration {
+public final class CopyJobConfiguration extends JobConfiguration {
 
   private static final long serialVersionUID = 1140509641399762967L;
 
@@ -180,15 +180,14 @@ public class CopyJobConfiguration extends JobConfiguration {
   }
 
   @Override
-  public final boolean equals(Object obj) {
+  public boolean equals(Object obj) {
     return obj == this
-        || obj != null
-        && obj.getClass().equals(CopyJobConfiguration.class)
+        || obj instanceof CopyJobConfiguration
         && baseEquals((CopyJobConfiguration) obj);
   }
 
   @Override
-  public final int hashCode() {
+  public int hashCode() {
     return Objects.hash(baseHashCode(), sourceTables, destinationTable, createDisposition,
         writeDisposition);
   }
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/ExtractJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/ExtractJobConfiguration.java
@@ -30,7 +30,7 @@ import java.util.Objects;
  * Cloud Storage. The extract destination provided as URIs that point to objects in Google Cloud
  * Storage. Extract job configurations have {@link JobConfiguration.Type#EXTRACT} type.
  */
-public class ExtractJobConfiguration extends JobConfiguration {
+public final class ExtractJobConfiguration extends JobConfiguration {
 
   private static final long serialVersionUID = 4147749733166593761L;
 
@@ -210,15 +210,14 @@ public class ExtractJobConfiguration extends JobConfiguration {
   }
 
   @Override
-  public final boolean equals(Object obj) {
+  public boolean equals(Object obj) {
     return obj == this
-        || obj != null
-        && obj.getClass().equals(ExtractJobConfiguration.class)
+        || obj instanceof ExtractJobConfiguration
         && baseEquals((ExtractJobConfiguration) obj);
   }
 
   @Override
-  public final int hashCode() {
+  public int hashCode() {
     return Objects.hash(baseHashCode(), sourceTable, destinationUris, printHeader, fieldDelimiter,
         format, compression);
   }
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/LoadJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/LoadJobConfiguration.java
@@ -30,7 +30,7 @@ import java.util.Objects;
  * table. Data is provided as URIs that point to objects in Google Cloud Storage. Load job
  * configurations have {@link JobConfiguration.Type#LOAD} type.
  */
-public class LoadJobConfiguration extends JobConfiguration implements LoadConfiguration {
+public final class LoadJobConfiguration extends JobConfiguration implements LoadConfiguration {
 
   private static final long serialVersionUID = -2673554846792429829L;
 
@@ -268,15 +268,14 @@ public class LoadJobConfiguration extends JobConfiguration implements LoadConfig
   }
 
   @Override
-  public final boolean equals(Object obj) {
+  public boolean equals(Object obj) {
     return obj == this
-        || obj != null
-        && obj.getClass().equals(LoadJobConfiguration.class)
+        || obj instanceof LoadJobConfiguration
         && baseEquals((LoadJobConfiguration) obj);
   }
 
   @Override
-  public final int hashCode() {
+  public int hashCode() {
     return Objects.hash(baseHashCode(), sourceUris);
   }
 
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
@@ -35,9 +35,22 @@ import java.util.Objects;
  * Google BigQuery Query Job configuration. A Query Job runs a query against BigQuery data. Query
  * job configurations have {@link JobConfiguration.Type#QUERY} type.
  */
-public class QueryJobConfiguration extends JobConfiguration {
+public final class QueryJobConfiguration extends JobConfiguration {
 
   private static final long serialVersionUID = -1108948249081804890L;
+
+  private final String query;
+  private final TableId destinationTable;
+  private final Map<String, ExternalTableDefinition> tableDefinitions;
+  private final List<UserDefinedFunction> userDefinedFunctions;
+  private final CreateDisposition createDisposition;
+  private final WriteDisposition writeDisposition;
+  private final DatasetId defaultDataset;
+  private final Priority priority;
+  private final Boolean allowLargeResults;
+  private final Boolean useQueryCache;
+  private final Boolean flattenResults;
+  private final Boolean dryRun;
 
   /**
    * Priority levels for a query. If not specified the priority is assumed to be
@@ -58,19 +71,6 @@ public class QueryJobConfiguration extends JobConfiguration {
      */
     BATCH
   }
-
-  private final String query;
-  private final TableId destinationTable;
-  private final Map<String, ExternalTableDefinition> tableDefinitions;
-  private final List<UserDefinedFunction> userDefinedFunctions;
-  private final CreateDisposition createDisposition;
-  private final WriteDisposition writeDisposition;
-  private final DatasetId defaultDataset;
-  private final Priority priority;
-  private final Boolean allowLargeResults;
-  private final Boolean useQueryCache;
-  private final Boolean flattenResults;
-  private final Boolean dryRun;
 
   public static final class Builder
       extends JobConfiguration.Builder<QueryJobConfiguration, Builder> {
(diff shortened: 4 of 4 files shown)


gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/LoadJobConfiguration.java
```diff
@@ -268,12 +268,14 @@ ToStringHelper toStringHelper() {
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof LoadJobConfiguration && baseEquals((LoadJobConfiguration) obj);
+  public final boolean equals(Object obj) {
+    return this == obj
+        || obj instanceof LoadJobConfiguration
+        && baseEquals((LoadJobConfiguration) obj);
```
aozarov: for pure delegation like this one we could have changed the baseEquals to return `true` if `this == obj`

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
```diff
@@ -373,8 +373,9 @@ public String toString() {
 
   @Override
   public int hashCode() {
```
aozarov: Any reason this is and bellow is not final?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Table.java
```diff
@@ -322,14 +322,15 @@ public Builder toBuilder() {
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof Table
+  public final boolean equals(Object obj) {
+    return this == obj
+        || obj instanceof Table
```
aozarov: Fine as is (as nothing extends it) however wouldn't it be better if used 'obj != null && object.getClass().equals(Table.class)` consistently?

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/ViewDefinition.java
```diff
@@ -146,12 +146,14 @@ ToStringHelper toStringHelper() {
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof ViewDefinition && baseEquals((ViewDefinition) obj);
+  public final boolean equals(Object obj) {
+    return this == obj
```
aozarov: Same comment like above. I wonder if `baseEquals` could actually do:

```java
  final boolean baseEquals(TableDefinition tableDefinition) {
    return tableDefinition == this
        || tableDefinition.getClass().equals(getClass())
        && Objects.equals(toPb(), tableDefinition.toPb());
  }
```

aozarov:
cleanup looks fine, however I wonder if we can make it more consistent.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
mziccard: What policy should we follow with `final` and `equals`/`hashCode` methods?
Do we want all `equals()` and `hashCode()` to be final? Or just for functional classes and configurations? (where bad subclassing would be more painful). Once we decide this I can take care of making everything more consistent.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Table.java
mziccard: The method is `final` so even if extended that should not be a problem.
However I believe that in all "functional" classes we used `instanceof` while in all `Info` objects we used `obj.getClass().equals(XxxInfo.class)` as those classes are extended. I need to double check that this is consistent.

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/QueryStage.java
aozarov: I think there should be 2 options for testing (other than using the service as is via a local or remote helper):

1) Don't mock use as is (and class is final). Typically for data types that a user can fully configure for its tests.
2) This class can't (or shouldn't) be created by the user and therefore have to be mocked for tests.
  In this case, class should not be final, constructor should be typically package scope and hashCode/equals methods should be final.

As this class is not final (and I didn't take the time to see why - other than seeing if it has subclasses) I assumed
it is part of (2).

gcloud-java-bigquery/src/main/java/com/google/gcloud/bigquery/Table.java
aozarov: > The method is final so even if extended that should not be a problem.

This PR removes the `final`, no?
I think that because it does not have final it would be better off with the getClass() check.

mziccard: Not sure I understand here. This PR is adding final to the `equals` method and removes it from classes.

aozarov: Yes, you are right.

The difference is what would you *prefer* to happen when I create a subclass (not a mock, so with real values) and call `tableInstance.equals(subclassInstance)`.
If you are fine with True in this case than keep as is, otherwise I think you should change to getClass() equality. 

To me, the fact that the class is not final is a compromise that is done (and should be done) for testing (and only for testing) and therefore would do whatever is possible to minimize errors.
However, this is not a big deal and I am fine either way.

mziccard:
@aozarov I had to rebase to trigger CI. I added one more commit that does a better job in making `equals` and `hashCode` consistent across all modules. Please have a look.

gcloud-java-storage/src/main/java/com/google/gcloud/storage/BatchResponse.java
```diff
@@ -121,13 +121,16 @@ public String toString() {
...
   @Override
-  public int hashCode() {
+  public final int hashCode() {
     return Objects.hash(deleteResult, updateResult, getResult);
   }
 
   @Override
-  public boolean equals(Object obj) {
-    if (!(obj instanceof BatchResponse)) {
+  public final boolean equals(Object obj) {
```
mziccard: Codacy Issue found: [Unnecessary final modifier in final class](https://www.codacy.com/app/mziccard/gcloud-java/file/2392420400/issues/source?bid=3123103&fileBranchId=3228372#l129)

gcloud-java-storage/src/main/java/com/google/gcloud/storage/BatchResponse.java
```diff
@@ -121,13 +121,16 @@ public String toString() {
   }
 
   @Override
-  public int hashCode() {
+  public final int hashCode() {
```
mziccard: Codacy Issue found: [Unnecessary final modifier in final class](https://www.codacy.com/app/mziccard/gcloud-java/file/2392420400/issues/source?bid=3123103&fileBranchId=3228372#l124)

gcloud-java-dns/src/main/java/com/google/gcloud/dns/RecordSet.java
```diff
@@ -276,13 +276,15 @@ public Type type() {
   }
 
   @Override
-  public int hashCode() {
+  public final int hashCode() {
```
mziccard: Codacy Issue found: [Unnecessary final modifier in final class](https://www.codacy.com/app/mziccard/gcloud-java/file/2392420333/issues/source?bid=3123103&fileBranchId=3228372#l279)

gcloud-java-dns/src/main/java/com/google/gcloud/dns/RecordSet.java
```diff
@@ -276,13 +276,15 @@ public Type type() {
...
     return Objects.hash(name, rrdatas, ttl, type);
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof RecordSet && Objects.equals(this.toPb(), ((RecordSet) obj).toPb());
+  public final boolean equals(Object obj) {
```
mziccard: Codacy Issue found: [Unnecessary final modifier in final class](https://www.codacy.com/app/mziccard/gcloud-java/file/2392420333/issues/source?bid=3123103&fileBranchId=3228372#l284)

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/CopyJobConfiguration.java
```diff
@@ -31,7 +31,7 @@
  * Google BigQuery copy job configuration. A copy job copies an existing table to another new or
  * existing table. Copy job configurations have {@link JobConfiguration.Type#COPY} type.
  */
-public final class CopyJobConfiguration extends JobConfiguration {
+public class CopyJobConfiguration extends JobConfiguration {
```
ajkannan: I think this class should be an instance of "final class, non-final equals/hashcode" since this class has a builder with only public setters.

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
```diff
@@ -255,12 +255,15 @@ ToStringHelper toStringHelper() {
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof ExternalTableDefinition && baseEquals((ExternalTableDefinition) obj);
+  public final boolean equals(Object obj) {
```
ajkannan: Is this class is another instance of "final class, non-final equals/hashCode"?

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/ExtractJobConfiguration.java
```diff
@@ -30,7 +30,7 @@
  * Cloud Storage. The extract destination provided as URIs that point to objects in Google Cloud
  * Storage. Extract job configurations have {@link JobConfiguration.Type#EXTRACT} type.
  */
-public final class ExtractJobConfiguration extends JobConfiguration {
+public class ExtractJobConfiguration extends JobConfiguration {
```
ajkannan: Same comment as for CopyJobConfiguration

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/LoadJobConfiguration.java
```diff
@@ -30,7 +30,7 @@
  * table. Data is provided as URIs that point to objects in Google Cloud Storage. Load job
  * configurations have {@link JobConfiguration.Type#LOAD} type.
  */
-public final class LoadJobConfiguration extends JobConfiguration implements LoadConfiguration {
+public class LoadJobConfiguration extends JobConfiguration implements LoadConfiguration {
```
ajkannan: This might also be a case of "final class, non-final equals/hashCode"

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -35,7 +35,7 @@
  * Google BigQuery Query Job configuration. A Query Job runs a query against BigQuery data. Query
  * job configurations have {@link JobConfiguration.Type#QUERY} type.
  */
-public final class QueryJobConfiguration extends JobConfiguration {
+public class QueryJobConfiguration extends JobConfiguration {
```
ajkannan: final class & not-final hashCode/equals

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/StandardTableDefinition.java
```diff
@@ -252,12 +252,15 @@ ToStringHelper toStringHelper() {
   }
 
   @Override
-  public boolean equals(Object obj) {
-    return obj instanceof StandardTableDefinition && baseEquals((StandardTableDefinition) obj);
+  public final boolean equals(Object obj) {
```
ajkannan: The `TableDefinition` subclasses probably also fall under the "final class, non-final equals/hashcode" case

ajkannan: Actually I take that back, it looks like there are some package-scope values in the builder.

mziccard: I am unsure whether we should make some of `TableDefinition` subclasses final (where your comment applies) and some other not (as this `StandardTableDefinition`) or just leave them as is for consistency in the hierarchy. Thoughts? /cc @aozarov

aozarov: If we make it final we are basically saying that users should not mock this one and instead should create an instance with their own pre-defined state in their tests. 

In this particular case, I don't think users can do that. They can create an instance with some of the values but not all. Therefore I would keep it as is (and personally I am fine with keeping its siblings consistent even if some can actually be final).

ajkannan:
Finished my pass.  Once you've addressed the comments and fixed the merge conflicts, feel free to merge.

mziccard:
Had to rebase to fix conflicts and trigger travis. Comments fixed in 5adf431

ajkannan:
LGTM

Issue #877, opened 2016-04-08
Migrate from com.google.gcloud.* to com.google.cloud


aozarov:
The current plan is to:

1) rename our Java package structure from `com.google.gcloud.` to `com.google.cloud.`

2) rename maven group-id from `com.google.gcloud` to `com.google.cloud`.

For now we are going to keep the following:

1) repository (and repository name).

2) maven artifact-id (should match repository name and therefore should be kept for now as gcloud-java).

3) directory structure and maven modules.

/cc @jgeewax

aozarov:
Was applied on master and released.
We need to make sure that we also apply it on the compute-alpha, gcs-nio and compute-alpha branches.

/cc @mziccard @mderka @jart @jean-philippe-martin

jean-philippe-martin:
Applied to gcs-nio as part of #937.

jean-philippe-martin:
Done as part of PR #937 (commit shown above).

jart:
I'm sorry I was on vacation the past few weeks. I like the new package name. But I'm very surprised that such a radical change is happening. I've also noticed a lot of other naming changes prior to this.

I've found it challenging as a 20% contributor to keep up with the pace of changes. I imagine it would be even more difficult for an independent user who isn't in collaboration with the developers.

The reason I'm concerned is because I'm going to be promoting this project in the near future. I take an extremely conservative outlook on software. And for that reason, I think some people trust my judgement. So when I run around telling everyone that gcs-nio / gcloud-java is the greatest thing they need to start using immediately, I want to make sure I'm offering the best recommendation possible.

So what assurances can gcloud-java offer that its API will become stable? What processes are being used, and will continue to be used, for API review?

mziccard:
Hi @jart,
I believe your concerns are more than reasonable however I see this from a different perspective. `gcloud-java` is in version `0.2.0` and we document in the README it is a work in progress:

> Note: This client is a work-in-progress, and may occasionally make backwards-incompatible changes.

Until we kick out our first major release (i.e. `1.0.0`) it's reasonable to assume that the API will make non-backward compatible changes. This very same approach has bean adopted in other `gcloud-*` libraries. Instead of seeing this as a released API that keeps changing this should be seen as a work-in-progress API that allows users to shape it and use/familiarize with it while being developed.

[...]

PR #899, opened 2016-04-12
Rename com.google.gcloud to com.google.cloud

ajkannan:
Related to #877.  Does not close the issue because the READMEs are left unchanged by this PR.  I will update the READMEs to use com.google.cloud once we release an artifact, otherwise users may get confused.

### DIFF
--- a/utilities/after_success.sh
+++ b/utilities/after_success.sh
@@ -19,7 +19,7 @@ if [ "${TRAVIS_JDK_VERSION}" == "oraclejdk7" ]; then
         fi
         if [ "${SITE_VERSION##*-}" != "SNAPSHOT" ]; then
             # Deploy Maven artifacts (if they don't exist yet) and update artifact version in READMEs.
-            URL=https://oss.sonatype.org/content/repositories/releases/com/google/gcloud/gcloud-java/$SITE_VERSION/
+            URL=https://oss.sonatype.org/content/repositories/releases/com/google/cloud/gcloud-java/$SITE_VERSION/
             if curl --output /dev/null --silent --head --fail "$URL"; then
                 echo "Not deploying artifacts because it seems like they already exist."
                 echo "Existence was checked using the url $URL"


mziccard:
I see that in READMEs we still have links to old javadoc and files (try searching `com/google/gcloud` in the files). Is this intended? I believe these links should be updated.

Also as soon as we merge this we should release new artifacts so that README links to javadoc are not broken.

ajkannan:
Fixed README links.  And yup, as soon as this gets merged, I plan on releasing artifacts.

mziccard:
README links seem to be fixed now. Since we merged #858 we should update the [URL](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/utilities/after_success.sh#L22) we use to check whether the artifact is already deployed or not.

mziccard:
Also is there something we can do to make it more clear that we renamed our `group` and `package` to users that already have `gcloud-java` as a dependency?

I am thinking about adding some temporary text on top of http://googlecloudplatform.github.io/gcloud-java/ and our main `README.md`. /cc @jgeewax @aozarov

ajkannan:
Good catch on the URL update.  I rebased, fixed that link, and grepped for "com.google.gcloud" and "com/google/gcloud" and found no results.  

I think it's a good idea to add some temporary text on the website and our README, they're both pretty common entrypoints to our repo according to our [traffic stats](https://github.com/GoogleCloudPlatform/gcloud-java/graphs/traffic).  I'll also add a note at the top of our release notes.

mziccard:
This looks good to me. I couldn't find any more stale naming. Text on top of site and README could be added in the release commit, right? Do you think it's worth adding it on top of each README?
I was thinking about something like:

> "As of April 12th, 2015, `gcloud-java`'s group ID and base package were renamed to `com.google.cloud`. If you haven't already please update your dependency."

To be placed just before where we show our dependency code. Could this work?

ajkannan:
That text sounds good to me.  I'll add that in the release commit.

ajkannan:
Just a note: after we release with this packaging change, we'll have to update the external applications (in java-docs-samples and getting-started-java) and also the links in cloud.google.com to particular gcloud-java classes.

aozarov:
Looks good. I think we only need to note in the release-notes, main readme page and maybe in the site page. FYI, I did a test and deployed it to [snapshots](https://oss.sonatype.org/content/repositories/snapshots/com/google/cloud/).

aozarov:
Feel free to merge when you are ready.

PR #909, opened 2016-04-12
Add url to the pom for nio and nio-examples

jean-philippe-martin:
Mirroring the changes in #906 that applied to master (looking forward to gcs-nio merging with master!)

/cc @mziccard

### DIFF
--- a/gcloud-java-contrib/pom.xml
+++ b/gcloud-java-contrib/pom.xml
@@ -4,7 +4,6 @@
   <artifactId>gcloud-java-contrib</artifactId>
   <packaging>pom</packaging>
   <name>GCloud Java contributions</name>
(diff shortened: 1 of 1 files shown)


gcloud-java-contrib/pom.xml
```diff
@@ -4,6 +4,7 @@
   <artifactId>gcloud-java-contrib</artifactId>
   <packaging>pom</packaging>
   <name>GCloud Java contributions</name>
+  <url>https://github.com/GoogleCloudPlatform/gcloud-java/tree/master/gcloud-java-contrib</url>
```
aozarov: This change was already applied on master by #906

jean-philippe-martin: OK, let me remove it then so we have an easier time merging!

PR #917, opened 2016-04-13
Add AttachedDisk class, related configurations and tests

mziccard:
This PR adds support for Compute's instances [attached disks](https://cloud.google.com/compute/docs/reference/latest/instances#disks).

Attached disks are provided to an instance upon creation or via the [attachDisk](https://cloud.google.com/compute/docs/reference/latest/instances/attachDisk) API. Since `attachDisk` only supports attaching already existing persistent disks (no support for scratch disks or creation on the fly) our `attachDisk` method (coming in a following PR) will look like:

``` Java
Operation attachDisk(InstanceId instance, PersistentDiskConfiguration configuration,
    OperationOption... options);

Operation attachDisk(InstanceId instance, String deviceName,
    PersistentDiskConfiguration configuration, OperationOption... options);

Operation attachDisk(InstanceId instance, String deviceName,
    PersistentDiskConfiguration configuration, int deviceIndex,
    OperationOption... options);
```

Also the `InstanceInfo.of` method will look like:

``` Java
InstanceInfo of(InstanceId instance, MachineTypeId machineType,
    AttachedDisk attachedDisk, NetworkInterface networkinterface);
```

However an instance can be created only if an `AttachedDisk` with `PersistentDiskConfiguration` or `CreateDiskConfiguration` is passed, and this method does not enforce that constraint.

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/gcloud/compute/AttachedDisk.java
+++ b/gcloud-java-compute/src/main/java/com/google/gcloud/compute/AttachedDisk.java
@@ -46,8 +46,7 @@ public class AttachedDisk implements Serializable {
       };
   static final Function<AttachedDisk, com.google.api.services.compute.model.AttachedDisk>
       TO_PB_FUNCTION =
-      new Function<AttachedDisk,
-          com.google.api.services.compute.model.AttachedDisk>() {
+      new Function<AttachedDisk, com.google.api.services.compute.model.AttachedDisk>() {
         @Override
         public com.google.api.services.compute.model.AttachedDisk apply(
             AttachedDisk attachedDisk) {
@@ -233,7 +232,9 @@ public class AttachedDisk implements Serializable {
       private Boolean boot;
       private Boolean autoDelete;
 
-      private Builder() {}
+      private Builder(DiskId sourceDisk) {
+        this.sourceDisk = checkNotNull(sourceDisk);
+      }
 
       private Builder(PersistentDiskConfiguration configuration) {
         sourceDisk = configuration.sourceDisk;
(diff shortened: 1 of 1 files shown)


aozarov:
I am confused.

The associated link, [attached disks](https://cloud.google.com/compute/docs/reference/latest/instances/attachDisk) specify that both PERSISTENT AND SCRATCH are valid options as a disk type.

mziccard:
> I am confused.
The associated link, attached disks specify that both PERSISTENT AND SCRATCH are valid options as a disk type.

I guess that is just a copy-paste from the description of an attached disk resource. In reality if I try to pass either a scratch disk or create a persistent disk on the fly (`ScratchDiskConfiguration` or `CreateDiskConfiguration`) I get the following error:

```JSON
{
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "invalid",
    "message": "Invalid value for field 'resource.initializeParams': ''. Cannot specify initialize params during attach disk, the persistent disk must exist."
   }
  ],
  "code": 400,
  "message": "Invalid value for field 'resource.initializeParams': ''. Cannot specify initialize params during attach disk, the persistent disk must exist."
 }
}
```
Which made me conclude that only attaching an existing persistent disk is supported.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/AttachedDisk.java
```diff
@@ -261,8 +262,8 @@ public Builder mode(Mode mode) {
 
       /**
        * Sets whether to use the attached disk as a boot disk. If {@code true} the virtual machine
-       * will use the first partition of the disk for its root filesystem. If not specified, the
-       * disk is not used as a boot disk.
+       * instance will use the first partition of the disk for its root filesystem. If not
+       * specified, the isk is not used as a boot disk.
```
aozarov: missing 'd' before 'isk'?

aozarov:
As discuessed in #928 I like this approach better. Looks good to me but I didn't look carefully at the details.

PR #919, opened 2016-04-13
Integration test for batch. Fixes #874.

### LINKED ISSUE #874: Implement integration tests for DNS batch

mderka:
The test performs the same operations as the integration test without batch. Where possible, it groups calls into a single batch call. It adds one test checking that a combination of operations is possible as well.

### DIFF
--- a/gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
+++ b/gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
@@ -532,7 +532,7 @@ public class ITDnsTest {
       assertEquals(CHANGE_ADD_ZONE1.additions(), created.additions());
       assertNotNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("1", created.generatedId());
+      assertNotNull(created.generatedId());
       assertTrue(ImmutableList.of(ChangeRequest.Status.PENDING, ChangeRequest.Status.DONE)
           .contains(created.status()));
       assertEqChangesIgnoreStatus(created, DNS.getChangeRequest(ZONE1.name(), "1"));
@@ -545,7 +545,7 @@ public class ITDnsTest {
       assertTrue(created.additions().isEmpty());
       assertNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("3", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       waitForChangeToComplete(created);
       created = DNS.applyChangeRequest(ZONE1.name(), CHANGE_DELETE_ZONE1);
@@ -555,7 +555,7 @@ public class ITDnsTest {
       assertTrue(created.additions().isEmpty());
       assertNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("5", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNotNull(created.status());
       waitForChangeToComplete(created);
       created = DNS.applyChangeRequest(ZONE1.name(), CHANGE_DELETE_ZONE1);
@@ -565,7 +565,7 @@ public class ITDnsTest {
       assertTrue(created.additions().isEmpty());
       assertNotNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("7", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       waitForChangeToComplete(created);
       created = DNS.applyChangeRequest(ZONE1.name(), CHANGE_DELETE_ZONE1);
@@ -575,7 +575,7 @@ public class ITDnsTest {
       assertEquals(CHANGE_ADD_ZONE1.additions(), created.additions());
       assertNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("9", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       // finishes with delete otherwise we cannot delete the zone
       waitForChangeToComplete(created);
@@ -585,7 +585,7 @@ public class ITDnsTest {
       assertEquals(CHANGE_DELETE_ZONE1.deletions(), created.deletions());
       assertNull(created.startTimeMillis());
       assertTrue(created.additions().isEmpty());
-      assertEquals("10", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       waitForChangeToComplete(created);
     } finally {
@@ -731,7 +731,7 @@ public class ITDnsTest {
       change = changes.get(1);
       assertEquals(CHANGE_ADD_ZONE1.additions(), change.additions());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertNull(change.status());
       changes = ImmutableList.copyOf(DNS.listChangeRequests(ZONE1.name(),
@@ -740,7 +740,7 @@ public class ITDnsTest {
       change = changes.get(2);
       assertTrue(change.additions().isEmpty());
       assertNotNull(change.deletions());
-      assertEquals("2", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertNull(change.status());
       changes = ImmutableList.copyOf(DNS.listChangeRequests(ZONE1.name(),
@@ -749,7 +749,7 @@ public class ITDnsTest {
       change = changes.get(1);
       assertTrue(change.additions().isEmpty());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertNull(change.status());
       changes = ImmutableList.copyOf(DNS.listChangeRequests(ZONE1.name(),
@@ -758,7 +758,7 @@ public class ITDnsTest {
       change = changes.get(1);
       assertTrue(change.additions().isEmpty());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNotNull(change.startTimeMillis());
       assertNull(change.status());
       changes = ImmutableList.copyOf(DNS.listChangeRequests(ZONE1.name(),
@@ -767,7 +767,7 @@ public class ITDnsTest {
       change = changes.get(1);
       assertTrue(change.additions().isEmpty());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertEquals(ChangeRequest.Status.DONE, change.status());
     } finally {
@@ -1439,7 +1439,7 @@ public class ITDnsTest {
       assertEquals(CHANGE_ADD_ZONE1.additions(), created.additions());
       assertNotNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("1", created.generatedId());
+      assertNotNull(created.generatedId());
       assertTrue(ImmutableList.of(ChangeRequest.Status.PENDING, ChangeRequest.Status.DONE)
           .contains(created.status()));
       assertEqChangesIgnoreStatus(created, DNS.getChangeRequest(ZONE1.name(), "1"));
@@ -1455,7 +1455,7 @@ public class ITDnsTest {
       assertTrue(created.additions().isEmpty());
       assertNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("3", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       waitForChangeToComplete(created);
       created = DNS.applyChangeRequest(ZONE1.name(), CHANGE_DELETE_ZONE1);
@@ -1468,7 +1468,7 @@ public class ITDnsTest {
       assertTrue(created.additions().isEmpty());
       assertNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("5", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNotNull(created.status());
       waitForChangeToComplete(created);
       created = DNS.applyChangeRequest(ZONE1.name(), CHANGE_DELETE_ZONE1);
@@ -1481,7 +1481,7 @@ public class ITDnsTest {
       assertTrue(created.additions().isEmpty());
       assertNotNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("7", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       waitForChangeToComplete(created);
       created = DNS.applyChangeRequest(ZONE1.name(), CHANGE_DELETE_ZONE1);
@@ -1494,7 +1494,7 @@ public class ITDnsTest {
       assertEquals(CHANGE_ADD_ZONE1.additions(), created.additions());
       assertNull(created.startTimeMillis());
       assertTrue(created.deletions().isEmpty());
-      assertEquals("9", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       // finishes with delete otherwise we cannot delete the zone
       waitForChangeToComplete(created);
@@ -1507,7 +1507,7 @@ public class ITDnsTest {
       assertEquals(CHANGE_DELETE_ZONE1.deletions(), created.deletions());
       assertNull(created.startTimeMillis());
       assertTrue(created.additions().isEmpty());
-      assertEquals("10", created.generatedId());
+      assertNotNull(created.generatedId());
       assertNull(created.status());
       waitForChangeToComplete(created);
     } finally {
@@ -1671,31 +1671,31 @@ public class ITDnsTest {
       change = Iterables.get(resultAdditions.get().values(), 1);
       assertEquals(CHANGE_ADD_ZONE1.additions(), change.additions());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertNull(change.status());
       change = Iterables.get(resultDeletions.get().values(), 2);
       assertTrue(change.additions().isEmpty());
       assertNotNull(change.deletions());
-      assertEquals("2", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertNull(change.status());
       change = Iterables.get(resultId.get().values(), 1);
       assertTrue(change.additions().isEmpty());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertNull(change.status());
       change = Iterables.get(resultTime.get().values(), 1);
       assertTrue(change.additions().isEmpty());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNotNull(change.startTimeMillis());
       assertNull(change.status());
       change = Iterables.get(resultStatus.get().values(), 1);
       assertTrue(change.additions().isEmpty());
       assertTrue(change.deletions().isEmpty());
-      assertEquals("1", change.generatedId());
+      assertNotNull(change.generatedId());
       assertNull(change.startTimeMillis());
       assertEquals(ChangeRequest.Status.DONE, change.status());
     } finally {


gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      batch = DNS.batch();
+      result = batch.listZones();
+      batch.submit();
+      zones = filter(result.get().iterateAll());
+      assertEquals(created, zones.get(0));
+      assertEquals(1, zones.size());
+      created = DNS.create(ZONE_EMPTY_DESCRIPTION);
+      batch = DNS.batch();
+      result = batch.listZones();
+      DnsBatchResult<Page<Zone>> zeroSizeError =
```
ajkannan: For clarity's sake, I would move all these batch results to right before they are used.  For example, this could be moved to just before line 1023.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+          batch.listZones(Dns.ZoneListOption.dnsName(ZONE1.dnsName()),
+              Dns.ZoneListOption.fields(Dns.ZoneField.NAME_SERVER_SET));
+      DnsBatchResult<Page<Zone>> combinationResult =
+          batch.listZones(Dns.ZoneListOption.fields(Dns.ZoneField.ZONE_ID,
+              Dns.ZoneField.DESCRIPTION),
+              Dns.ZoneListOption.pageSize(1));
+      batch.submit();
+      zones = filter(result.get().iterateAll());
+      assertEquals(2, zones.size());
+      assertTrue(zones.contains(created));
```
ajkannan: Why not save both created zones and assert this for both?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      assertEquals(created, zones.get(0));
+      assertEquals(1, zones.size());
+      created = DNS.create(ZONE_EMPTY_DESCRIPTION);
+      batch = DNS.batch();
+      result = batch.listZones();
+      DnsBatchResult<Page<Zone>> zeroSizeError =
+          batch.listZones(Dns.ZoneListOption.pageSize(0));
+      DnsBatchResult<Page<Zone>> negativeSizeError =
+          batch.listZones(Dns.ZoneListOption.pageSize(-1));
+      DnsBatchResult<Page<Zone>> okSize = batch.listZones(Dns.ZoneListOption.pageSize(1000));
```
ajkannan: I would slightly prefer to see page size of 1, and then you can assert down below that you only get one result.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
ajkannan: Ah, ignore this, they're all part of the same batch.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      DNS.delete(ZONE_DNS_NO_PERIOD.name());
+    }
+  }
+
+  @Test
+  public void testCreateZoneWithOptionsBatch() {
+    try {
+      DnsBatch batch = DNS.batch();
+      DnsBatchResult<Zone> batchResult =
+          batch.createZone(ZONE1, Dns.ZoneOption.fields(Dns.ZoneField.CREATION_TIME));
```
ajkannan: Nit: you could import the options from Dns (e.g. Dns.ZoneOption) and remove "Dns." everywhere you use options.  Up to you.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      DNS.delete(ZONE1.name());
+    }
+  }
+
+  @Test
+  public void testGetProjectBatch() {
+    // fetches all fields
+    DnsBatch batch = DNS.batch();
+    DnsBatchResult<ProjectInfo> result = batch.getProject();
+    DnsBatchResult<ProjectInfo> resultQuota =
```
ajkannan: This this batch result (and all below this one) supposed to be used?

ajkannan: Ah, I see, they are, nvm.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      DNS.create(ZONE1, Dns.ZoneOption.fields(Dns.ZoneField.NAME));
+      DnsBatch batch = DNS.batch();
+      DnsBatchResult<ChangeRequest> result =
+          batch.applyChangeRequest(ZONE1.name(), CHANGE_ADD_ZONE1);
+      batch.submit();
+      ChangeRequest created = result.get();
+      assertEquals(CHANGE_ADD_ZONE1.additions(), created.additions());
+      assertNotNull(created.startTimeMillis());
+      assertTrue(created.deletions().isEmpty());
+      assertEquals("1", created.generatedId());
```
ajkannan: Do we know whether the service owners plan to keep this numbering scheme?  (Is it documented somewhere?)

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      assertTrue(ImmutableList.of(ChangeRequest.Status.PENDING, ChangeRequest.Status.DONE)
+          .contains(created.status()));
+      assertEqChangesIgnoreStatus(created, DNS.getChangeRequest(ZONE1.name(), "1"));
+      waitForChangeToComplete(created);
```
ajkannan: I wonder if this test would be sped up significantly if we added different zones each time and didn't have to wait.  Not sure if it's worth changing, but I wanted to bring it up to get your opinion.  How long does a change request usually take to complete?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      batch.submit();
+      assertEqChangesIgnoreStatus(created, result.get());
+      waitForChangeToComplete(zone.name(), created.generatedId());
+      zone.applyChangeRequest(CHANGE_DELETE_ZONE1);
+      created = zone.applyChangeRequest(CHANGE_ADD_ZONE1,
+          Dns.ChangeRequestOption.fields(Dns.ChangeRequestField.STATUS));
+      batch = DNS.batch();
+      result = batch.getChangeRequest(zone.name(), created.generatedId(),
+          Dns.ChangeRequestOption.fields(Dns.ChangeRequestField.STATUS));
+      batch.submit();
```
ajkannan: Don't you have to wait for the delete to finish?

ajkannan: If waiting does take a long time, you could alternate additions and deletions like you did above to cut the number of waits in half.  Otherwise, probably not worth it.

ajkannan: Actually, can't you just create one change request and call get a bunch of times rather than creating and deleting a bunch of them?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      result = batch.listChangeRequests(ZONE1.name());
+      batch.submit();
+      ImmutableList<ChangeRequest> changes = ImmutableList.copyOf(result.get().iterateAll());
+      assertEquals(1, changes.size()); // default change creating SOA and NS
+      // zone has changes
+      ChangeRequest change = DNS.applyChangeRequest(ZONE1.name(), CHANGE_ADD_ZONE1);
+      waitForChangeToComplete(ZONE1.name(), change.generatedId());
+      change = DNS.applyChangeRequest(ZONE1.name(), CHANGE_DELETE_ZONE1);
+      waitForChangeToComplete(ZONE1.name(), change.generatedId());
+      change = DNS.applyChangeRequest(ZONE1.name(), CHANGE_ADD_ZONE1);
```
ajkannan: You could probably cut the last add and delete to make 3 total changes and still test everything you want to, right?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+        assertTrue(defaultRecords.contains(recordSet.type()));
+      }
+      // field options
+      batch = DNS.batch();
+      DnsBatchResult<Page<RecordSet>> ttlResult = batch.listRecordSets(zone.name(),
+          Dns.RecordSetListOption.fields(Dns.RecordSetField.TTL));
+      DnsBatchResult<Page<RecordSet>> nameResult = batch.listRecordSets(zone.name(),
+          Dns.RecordSetListOption.fields(Dns.RecordSetField.NAME));
+      DnsBatchResult<Page<RecordSet>> recordsResult = batch.listRecordSets(zone.name(),
+          Dns.RecordSetListOption.fields(Dns.RecordSetField.DNS_RECORDS));
```
ajkannan: Should we rename this as per an earlier discussion to rename DnsRecord to RecordSet?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      DnsBatchResult<Page<RecordSet>> typeResult = batch.listRecordSets(zone.name(),
+          Dns.RecordSetListOption.fields(Dns.RecordSetField.TYPE),
+          Dns.RecordSetListOption.pageSize(1)); // also paging
+      DnsBatchResult<Page<RecordSet>> pageSizeResult = batch.listRecordSets(zone.name(),
```
ajkannan: Why do we need two list calls with the same fields?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+        assertTrue(recordSet.records().isEmpty());
+        assertNull(recordSet.ttl());
+        counter++;
+      }
+      assertEquals(2, counter);
+      // test page size
+      Page<RecordSet> recordSetPage = pageSizeResult.get();
+      assertEquals(1, ImmutableList.copyOf(recordSetPage.values().iterator()).size());
+      // test name filter
+      ChangeRequest change = DNS.applyChangeRequest(ZONE1.name(), CHANGE_ADD_ZONE1);
```
ajkannan: Could you have used one of the two default records instead of creating a new one?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      counter = 0;
+      while (recordSetIterator.hasNext()) {
+        RecordSet recordSet = recordSetIterator.next();
+        assertTrue(ImmutableList.of(A_RECORD_ZONE1.type(), AAAA_RECORD_ZONE1.type())
+            .contains(recordSet.type()));
+        counter++;
+      }
+      assertEquals(2, counter);
+      // test type filter
+      waitForChangeToComplete(ZONE1.name(), change.generatedId());
```
ajkannan: Did you already wait for this change to complete?

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +967,920 @@ public void testListDnsRecords() {
...
+      batch.submit();
+      recordSetIterator = result.get().iterateAll();
+      counter = 0;
+      while (recordSetIterator.hasNext()) {
+        RecordSet recordSet = recordSetIterator.next();
+        assertEquals(A_RECORD_ZONE1, recordSet);
+        counter++;
+      }
+      assertEquals(1, counter);
+      change = zone.applyChangeRequest(CHANGE_DELETE_ZONE1);
```
ajkannan: Could you have left this to clear()?

ajkannan:
Done with my pass.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Done.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Done.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: No, we do not. There was no indication of an intention to change it.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Change requests complete almost instantaneously. The delay is given by network communication. 

I would not do this. The zones need to be unqiue. Generating more unique zones would require more code to maintain, and it would have more overhead when deleting. Given how scarcely these tests run, I do not see trading additional code for potential (I seriously doubt it) performance increase as the way to go.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Correct. Done.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: No, this is correct. This indicates records inside a record set.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: We do not. Removed.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: The point is that the default records, which have the same name, are not present when the filter is used.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Correct. Removed.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Yes, removed.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Correct. Did only one change apply.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Done.

mderka:
Ready to proceed.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +969,920 @@ public void testListDnsRecords() {
...
       batch.submit();
-      assertTrue(batchResult.completed());
-      Iterator<Zone> iteratorBatch = batchResult.get().iterateAll();
-      Iterator<Zone> iteratorList = DNS.listZones().iterateAll();
-      assertEquals(Iterators.size(iteratorList), Iterators.size(iteratorBatch));
-      while (iteratorBatch.hasNext()) {
-        assertTrue(Iterators.contains(iteratorList, iteratorBatch.next()));
+      zones = filter(result.get().iterateAll());
+      assertEquals(firstZone, zones.get(0));
+      assertEquals(1, zones.size());
```
mziccard: Maybe assert size before getting the first element? To get a nice failure rather than an error

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +969,920 @@ public void testListDnsRecords() {
...
+      try {
+        negativeSizeError.get();
+        fail();
+      } catch (DnsException ex) {
+        // expected
+        assertEquals(400, ex.code());
+        assertFalse(ex.retryable());
+      }
+      // ok size
+      assertEquals(1, Iterators.size(okSize.get().values().iterator()));
```
mziccard: Nit: could have used [`Iterables.size(okSize.get().values())`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterables.html#size(java.lang.Iterable))

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +969,920 @@ public void testListDnsRecords() {
...
+        // expected
+        assertEquals(404, ex.code());
+        assertFalse(ex.retryable());
+      }
+      // zone exists but has no changes
+      DNS.create(ZONE1);
+      batch = DNS.batch();
+      result = batch.listChangeRequests(ZONE1.name());
+      batch.submit();
+      ImmutableList<ChangeRequest> changes = ImmutableList.copyOf(result.get().iterateAll());
```
mziccard: Is it necessary to copy the iterable into a list? You could use [Iterables.size](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterables.html#size(java.lang.Iterable))

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +969,920 @@ public void testListDnsRecords() {
...
+          Dns.ChangeRequestListOption.sortOrder(Dns.SortingOrder.ASCENDING),
+          Dns.ChangeRequestListOption.fields(ChangeRequestField.ID));
+      DnsBatchResult<Page<ChangeRequest>> resultTime = batch.listChangeRequests(ZONE1.name(),
+          Dns.ChangeRequestListOption.sortOrder(Dns.SortingOrder.ASCENDING),
+          Dns.ChangeRequestListOption.fields(ChangeRequestField.START_TIME));
+      DnsBatchResult<Page<ChangeRequest>> resultStatus = batch.listChangeRequests(ZONE1.name(),
+          Dns.ChangeRequestListOption.sortOrder(Dns.SortingOrder.ASCENDING),
+          Dns.ChangeRequestListOption.fields(ChangeRequestField.STATUS));
+      batch.submit();
+      changes = ImmutableList.copyOf(result.get().iterateAll());
```
mziccard: Also here you could use [Iterables.size](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterables.html#size(java.lang.Iterable))

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +969,920 @@ public void testListDnsRecords() {
...
+      ImmutableList<ChangeRequest> descending =
+          ImmutableList.copyOf(resultDescending.get().iterateAll());
+      int size = 3;
+      assertEquals(size, descending.size());
+      assertEquals(size, ascending.size());
+      for (int i = 0; i < size; i++) {
+        assertEquals(descending.get(i), ascending.get(size - i - 1));
+      }
+      // field options
+      changes = ImmutableList.copyOf(resultAdditions.get().iterateAll());
```
mziccard: I think that in `testListZonesBatch` you use the iterator to go through the results. I slightly prefer that to avoid copies. (also applies to the following)

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -967,25 +969,920 @@ public void testListDnsRecords() {
...
+      }
+      batch.submit();
+      assertNotNull(zoneResult.get().creationTimeMillis());
+      assertEquals(ZONE1.dnsName(), zoneResult.get().dnsName());
+      assertEquals(ZONE1.description(), zoneResult.get().description());
+      assertFalse(zoneResult.get().nameServers().isEmpty());
+      assertNull(zoneResult.get().nameServerSet()); // we did not set it
+      assertNotNull(zoneResult.get().generatedId());
+      assertNotNull(projectResult.get().quota());
+      assertEquals(2, Iterators.size(pageResult.get().iterateAll()));
```
mziccard: Nit: [`Iterables.size`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterables.html#size(java.lang.Iterable))

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Done.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Done.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Correct. Done.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Done.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Fixed.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
mderka: Done.

mderka:
PTAL

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
```diff
@@ -40,6 +40,7 @@
 import com.google.cloud.dns.Zone;
 import com.google.cloud.dns.ZoneInfo;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
```
mziccard: Is `Iterators` still needed?

mderka: No, removed.

mderka:
PTAL

mziccard:
Looks good to me. @ajkannan do you have any further comments? If not feel free to merge after squashing the commits.

mderka:
Fixes #874.

gcloud-java-dns/src/test/java/com/google/cloud/dns/it/ITDnsTest.java
ajkannan: Hm, I feel that we should only assertNotNull instead of checking for equality for generatedIds for two reasons:
1. The service owners could choose to change it (I don't know how likely that is)
2. We don't want to mislead users into thinking generated IDs will always follow this pattern.

mderka:
Fixed everywhere.

ajkannan:
LGTM

PR #921, opened 2016-04-14
Disallowed null error in BatchResult. Adjusted doc.

mderka:
A few small adjustments based on @aozarov's post-merge comments in #787.


mziccard:
Looks good to me

PR #925, opened 2016-04-14
Added batch support to the local DNS helper.

mderka:


gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -273,13 +281,18 @@ private String toJson(String message) throws IOException {
   private class RequestHandler implements HttpHandler {
 
     private Response pickHandler(HttpExchange exchange, CallRegex regex) {
-      URI relative = BASE_CONTEXT.relativize(exchange.getRequestURI());
+      URI relative = null;
+      try {
+        relative = BASE_CONTEXT.relativize(new URI(exchange.getRequestURI().getRawPath()));
```
ajkannan: Out of curiosity, what was the impetus for this change and the one on line 295?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -430,6 +520,21 @@ private static void writeResponse(HttpExchange exchange, Response response) {
     }
   }
 
+  private static void writeBatchResponse(HttpExchange exchange, ByteArrayOutputStream out,
+      String boundary) {
+    exchange.getResponseHeaders().set("Content-type", "multipart/mixed; boundary=" + boundary);
+    try {
+      exchange.getResponseHeaders().add("Connection", "close");
+      exchange.getResponseHeaders().add("Connection", "close");
```
ajkannan: Is the duplication intended?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Batch sends request addresses in full "http://localhost:port/path" form which does not match the base context and fails relativization.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -328,6 +352,71 @@ public void handle(HttpExchange exchange) throws IOException {
           requestMethod, exchange.getRequestURI())));
     }
 
+    private Response handleBatch(final HttpExchange exchange) throws IOException {
+      String contentType = exchange.getRequestHeaders().getFirst("Content-type");
+      if (contentType != null) {
+        int port = server.getAddress().getPort();
+        String responseBoundary = "____THIS_IS_HELPERS_BOUNDARY____";
```
ajkannan: Do you think it'd be preferable to define this and the two vars below as private static final vars?  That way they won't be recreated each time a new batch request comes in.

ajkannan:
LocalDnsHelperTest.java:2916:86: error: ',' is not followed by whitespace.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Yes. Done.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: No. Removed.

ajkannan:
On lines 1951-2 in the tests, why do we assertEquals(first, results.get(1)); and assertEquals(second, results.get(0));?

ajkannan:
Missing "batch.submit()" in testCombined()

ajkannan:
Done with my pass.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -112,6 +119,15 @@
   private static final ScheduledExecutorService EXECUTORS =
       Executors.newScheduledThreadPool(2, Executors.defaultThreadFactory());
   private static final String PROJECT_ID = "dummyprojectid";
+  private static final String responseBoundary = "____THIS_IS_HELPERS_BOUNDARY____";
```
aozarov: RESPONSE_BOUNDRY (same for bellow).

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -112,6 +119,15 @@
...
+  private static final String responseSeparator = new StringBuilder("--")
```
aozarov: Nit, because all string values are constants simple concatenation would be more efficient:
```java
private static final String responseSeparator = "--" + responseBoundary + "\r\n";
```
as the compiler will make it one string.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -138,7 +154,8 @@
     ZONE_GET("GET", CONTEXT + "/[^/]+/managedZones/[^/]+"),
     ZONE_LIST("GET", CONTEXT + "/[^/]+/managedZones"),
     PROJECT_GET("GET", CONTEXT + "/[^/]+"),
-    RECORD_LIST("GET", CONTEXT + "/[^/]+/managedZones/[^/]+/rrsets");
+    RECORD_LIST("GET", CONTEXT + "/[^/]+/managedZones/[^/]+/rrsets"),
+    BATCH("POST", "/batch");
```
aozarov: Was this URL path validated via integration test?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -307,6 +329,13 @@ private Response pickHandler(HttpExchange exchange, CallRegex regex) {
           } catch (IOException ex) {
             return Error.BAD_REQUEST.response(ex.getMessage());
           }
+        case BATCH:
+          try {
+            return handleBatch(exchange);
+          } catch (IOException ex) {
+            ex.printStackTrace();
```
aozarov: remove

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
aozarov: I assume that you are talking about the Batch request itself (POST /batch) and not the individual batch calls that this emulator is triggering. Right? If so, why not to check for if `getPath()` for `new URI(exchange.getRequestURI().getRawPath()))` (which you can save in a local variable) is equals to "/batch" and in that case assign relative to it without applying `relativize` on it?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -430,6 +520,20 @@ private static void writeResponse(HttpExchange exchange, Response response) {
     }
   }
 
+  private static void writeBatchResponse(HttpExchange exchange, ByteArrayOutputStream out) {
```
aozarov: To avoid codacy complain rename `out` to `output`.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -328,6 +361,62 @@ public void handle(HttpExchange exchange) throws IOException {
           requestMethod, exchange.getRequestURI())));
     }
 
+    private Response handleBatch(final HttpExchange exchange) throws IOException {
+      String contentType = exchange.getRequestHeaders().getFirst("Content-type");
+      if (contentType != null) {
```
aozarov: What if `contentType` is null?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -328,6 +361,62 @@ public void handle(HttpExchange exchange) throws IOException {
...
+              new InputStreamReader(new ByteArrayInputStream(section.toByteArray())));
+          String line;
+          String contentId = null;
+          while (!(line = reader.readLine()).isEmpty()) {
+            if (line.toLowerCase().startsWith("content-id")) {
+              contentId = line.split(":")[1].trim();
+            }
+          }
+          String requestLine = reader.readLine();
+          socketOutput.write((requestLine + "  \r\n").getBytes());
```
aozarov: is the space before "\r\n" is necessary here and bellow?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -328,6 +361,62 @@ public void handle(HttpExchange exchange) throws IOException {
...
+          int length;
+          out.write(responseSeparator.getBytes());
+          out.write("Content-Type: application/http \r\n".getBytes());
+          out.write(("Content-ID: " + contentId + " \r\n\r\n").getBytes());
+          try {
+            while ((length = in.read(bytes)) != -1) {
+              out.write(bytes, 0, length);
+            }
+          } catch (IOException ex) {
+            // this handles connection reset error
```
aozarov: This can mask other errors. Why do you need it?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
```diff
@@ -328,6 +361,62 @@ public void handle(HttpExchange exchange) throws IOException {
...
+            new MultipartStream(exchange.getRequestBody(), boundary.getBytes(), 1024, null);
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        byte[] bytes = new byte[1024];
+        multipartStream.skipPreamble();
+        while (multipartStream.readBoundary()) {
+          Socket socket = new Socket("localhost", port);
+          OutputStream socketOutput = socket.getOutputStream();
+          ByteArrayOutputStream section = new ByteArrayOutputStream();
+          multipartStream.readBodyData(section);
+          BufferedReader reader = new BufferedReader(
```
aozarov: Fine as is, though because this is basically treated as a String (converted from byte[] using local encoding - is that desired or should it be UTF-8 ?) you could have used many String specific tokenizing (using, Scanner, StringTokenizer, split,...).

e.g. 
```java
Scanner scanner = new Scanner(new String(section.toByteArray()));
while (scanner.hasNextLine()) {
  String line = scanner.nextLine();
  ...
}
```

aozarov:
Finished with my pass on the main code. Didn't complete a thorough review of the tests, though they seem fine (I think code could be more compact if you created 2 inner class for `RpcBatch.Callback<T> `one of which fails on success and other fails on exception and used them accordingly).

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: I am talking about the calls that are inside the multipart POST /batch body:

```
POST /batch

--boundary
GET http://localhost:2020/path/query

--boundary
POST http://localhhost:2020/path/query

--boundary--
```

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Done.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Done.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Done.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Can you be more specific in what you have in mind? We do have integration tests for batch which are using the same RPC components.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Done.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Added bad request response.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Yes.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: I did not find another way of detecting the connection reset. In that case, any attempted read from the socket throws an exception instead of -1. Yet, available() returns true. Any suggestions?

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
mderka: Yes. Fixed.

mderka:
I made the test code more compact by adding the nested implementations of callbacks. Ready to proceed.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
aozarov: Oh, I wonder if that is correct... considering the example [here](https://cloud.google.com/storage/docs/json_api/v1/how-tos/batch).

mderka: I created an issue #929.

gcloud-java-dns/src/main/java/com/google/cloud/dns/testing/LocalDnsHelper.java
aozarov: This does not seem right. Lets create an issue for it and deal with it later.

mderka: Created #930. I am right now wondering if response writer should writes `\r\n` after the json body.

PR #936, opened 2016-04-18
Add options(namespace) method to LocalDatastoreHelper

### LINKED ISSUE #910: Overload options method to take a default namespace in LocalDatastoreHelper
We should also document the default namespace in the options class and builder.

mziccard:
This fixes #910

### DIFF
--- a/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
+++ b/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
@@ -556,17 +556,20 @@ public class LocalDatastoreHelper {
     }
   }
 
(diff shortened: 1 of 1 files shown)


gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
```diff
@@ -570,6 +570,20 @@ public DatastoreOptions options() {
   }
 
   /**
+   * Returns a {@link DatastoreOptions} instance that sets the host to use the Datastore emulator on
+   * localhost. The default namespace is set to {@code namespace}.
+   */
+  public DatastoreOptions options(String namespace) {
+    return DatastoreOptions.builder()
```
aozarov: Maybe the  two `options()` method could delegate to an internal optionsBuilder method for the common settings?

PR #946, opened 2016-04-20
Chain cause exception to service exception in translateAndThrow

### LINKED ISSUE #945: BigQueryException discards exception context
If BigQueryException.translateAndThrow is passed an exception that cannot be handled by BaseServiceException it throws a new exception with the message from the passed exception. However this throws away context.

It would be better if the fallback chained the inner exception.

mziccard:
This fixes #945

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryException.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryException.java
@@ -31,7 +31,7 @@ import java.util.Set;
  * @see <a href="https://cloud.google.com/bigquery/troubleshooting-errors">Google Cloud
  *      BigQuery error codes</a>
  */
-public class BigQueryException extends BaseServiceException {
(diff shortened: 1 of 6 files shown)


aozarov:
LGTM

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryException.java
```diff
@@ -44,10 +44,15 @@
   private final BigQueryError error;
 
   public BigQueryException(int code, String message) {
```
aozarov: Any reason for leaving this one, and few others in Datastore public?

mziccard: They're all used in the SPI layer

aozarov: I see. Then in that case I change my mind and suggests to make all public and make the classes final.

PR #947, opened 2016-04-20
Add NetworkInterface, AccessConfig classes and related tests

mziccard:


gcloud-java-compute/src/main/java/com/google/gcloud/compute/NetworkInterface.java
```diff
@@ -0,0 +1,505 @@
...
+
+    private final String name;
+    private final String natIp;
+    private final Type type;
+
+    /**
+     * The type of network access configuration. The only supported value is {@code ONE_TO_ONE_NAT}.
+     */
+    public enum Type {
+      ONE_TO_ONE_NAT
```
aozarov: No public documentation on what this value means?

gcloud-java-compute/src/main/java/com/google/gcloud/compute/NetworkInterface.java
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+    /**
+     * Returns an {@code AccessConfig} object. IP address for this access configuration will be
+     * taken from a pool of ephemeral addresses.
+     *
+     * @see <a href="https://cloud.google.com/compute/docs/instances-and-network#ephemeraladdress">
+     *     Ephemeral external IP addresses</a>
+     */
+    public static AccessConfig of() {
```
aozarov: Just wondering, is that (empty AccessConfig) useful?

aozarov:
Looks good to me.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/NetworkInterface.java
mziccard: I could find any in the service docs but i think it refers to the networking concept of 1:1 NAT. Which (very roughly speaking) is assigning a public address to a machine so that it can be reached from the outside.

gcloud-java-compute/src/main/java/com/google/gcloud/compute/NetworkInterface.java
mziccard: Yes it is. If you pass that empty access config an ephemeral address is associated to the instance. Instead, if you pass no access configs at all no address is associated to the instance.

PR #948, opened 2016-04-20
Chain cause in ComputeException.translateAndThrow, add tests

mziccard:
This fixes #946 for the compute module

### DIFF
--- a/gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeExceptionTest.java
+++ b/gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeExceptionTest.java
@@ -23,6 +23,7 @@ import static org.easymock.EasyMock.verify;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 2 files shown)


gcloud-java-compute/src/test/java/com/google/gcloud/compute/ComputeExceptionTest.java
```diff
@@ -0,0 +1,96 @@
...
+    expect(exceptionMock.getCause()).andReturn(cause).times(2);
+    replay(exceptionMock);
+    try {
+      ComputeException.translateAndThrow(exceptionMock);
+    } catch (BaseServiceException ex) {
+      assertEquals(ComputeException.UNKNOWN_CODE, ex.code());
+      assertEquals("message", ex.getMessage());
+      assertFalse(ex.retryable());
+      assertTrue(ex.idempotent());
+      assertEquals(cause, ex.getCause());
```
aozarov: I think you could have used `assertSame here.

aozarov:
LGTM

mziccard:
@aozarov I just noticed that we do very have different access modifiers for exceptions' constructors. In `dns` they are private while in other services they're public. I suggest we make the package scoped for the sake of testing. What do you think?

aozarov:
Yes, I think consistency here is good and package scope would be a good choice.

PR #951, opened 2016-04-22
Update gcloud pubsub

shinfan:
- Compatible with GAX 0.0.11
- Introduce PageAccessor
- Always throws ApiException instead of UncheckedException
- Better documentation

### DIFF
--- a/gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
+++ b/gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
@@ -33,6 +33,7 @@
 
 package com.google.cloud.pubsub.spi.v1;
 
+import com.google.api.gax.core.PageAccessor;
 import com.google.api.gax.grpc.ApiCallable;
 import com.google.api.gax.protobuf.PathTemplate;
 import com.google.protobuf.Empty;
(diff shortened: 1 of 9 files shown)


gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -59,6 +60,60 @@
...
+ *
+ * <p>Many parameters require resource names to be formatted in a particular way. To assist
+ * with these names, this class includes a format method for each type of name, and additionally
+ * a parse method to extract the individual identifiers contained within names that are
+ * returned.
+ *
+ * <p>This class can be customized by passing in a custom instance of PublisherSettings to
+ * create(). For example:
+ *
+ * <!-- TODO(garrettjones) refactor code to make this simpler -->
```
aozarov: I assume you plan to have the ServiceApiSettings.Builder return the nested builders similar to what you have now in ApiCallSettings.Builder?

gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -59,6 +60,60 @@
...
  * messages to a topic.
  *
+ * <p>This class provides the ability to make remote calls to the backing service through method
+ * calls that map to API methods. Sample code to get started:
+ *
+ * <pre>
+ * <code>
+ * try (PublisherApi publisherApi = PublisherApi.defaultInstance()) {
+ *   // make calls here
+ * String name = "";
```
aozarov: this is inside a `<pre>` you should be able to ident (similar comment to other cases in this PR).

gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
garrettjonesgoogle: This is a particular issue with ConnectionSettings. It wasn't part of the settings/builder refactoring, but we have it on our list to make it consistent with the rest.

PR #954, opened 2016-04-22
Move Clock out of ServiceOptions, use it in RetryHelper

### LINKED ISSUE #950: Replace RetryHandler package-scoped StopWatch with Option's Clock
We should use the Clock from ServiceOptions and pass it along to the RetryHandler (instead of the StopWatch)

mziccard:
This fixes #950. This also moves `Clock` out of `ServiceOptions` as it makes more sense since it's shared with `RetryHelper`.


aozarov:
Looks great. Next we need to use that Clock in *all* our RetryHelper usages (and make `runWithRetries` with `Clock` public).

mziccard:
Before merging this we should make:
```java
runWithRetries(Callable<V> callable, RetryParams params, ExceptionHandler exceptionHandler, Clock clock)
```
public and use it across all services passing the clock contained in `ServiceOptions`.

mziccard:
Rebased to get compute's module. Added 497ccbd to pass `options.clock()` to `retryHelper.runWithRetries` everywhere we use it.

PR #960, opened 2016-04-26
Move Stat and clarify example command line

jean-philippe-martin:
- Stat is now in com.google.cloud.examples.nio (following the gcloud ->
  cloud move)
- Added version for appassembler to remove warning
- added 'mvn install' to instructions because it's necessary.

### DIFF
--- a/gcloud-java-examples/README.md
+++ b/gcloud-java-examples/README.md
@@ -9,8 +9,8 @@ Examples for gcloud-java (Java idiomatic client for [Google Cloud Platform][clou
 [![Codacy Badge](https://api.codacy.com/project/badge/grade/9da006ad7c3a4fe1abd142e77c003917)](https://www.codacy.com/app/mziccard/gcloud-java)
 [![Dependency Status](https://www.versioneye.com/user/projects/56bd8ee72a29ed002d2b0969/badge.svg?style=flat)](https://www.versioneye.com/user/projects/56bd8ee72a29ed002d2b0969)
 
--  [Homepage] (https://googlecloudplatform.github.io/gcloud-java/)
--  [Examples] (http://googlecloudplatform.github.io/gcloud-java/apidocs/index.html?com/google/cloud/examples/package-summary.html)
+-  [Homepage](https://googlecloudplatform.github.io/gcloud-java/)
+-  [Examples](http://googlecloudplatform.github.io/gcloud-java/apidocs/index.html?com/google/cloud/examples/package-summary.html)
 
 Quickstart
 ----------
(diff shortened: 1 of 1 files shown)


jean-philippe-martin:
I realized I couldn't run the examples anymore. Updated the instructions, they work now.

mziccard:
Looks good to me.

Not related to this PR: I just noticed that the example's README.md is broken (see [here](https://github.com/GoogleCloudPlatform/gcloud-java/blob/gcs-nio/gcloud-java-examples/README.md)). I believe the reason is a missing closing ``` at line 72.

Just let me know if you plan to fix it in this PR.

gcloud-java-examples/README.md
```diff
@@ -37,7 +37,8 @@ To run examples from your command line:
 
 2. Set your current project using `gcloud config set project PROJECT_ID`. This step is not necessary for `ResourceManagerExample`.
 
-3. Compile using Maven: `cd gcloud-java-examples` in command line from your base project directory and then `mvn package appassembler:assemble -DskipTests -Dmaven.javadoc.skip=true  -Dmaven.source.skip=true`.
+3. Compile using Maven: `mvn install -DskipTests -Dmaven.javadoc.skip=true -Dmaven.source.skip=true` in command line from your base project directory
```
mziccard: We recently added a [release profile](https://github.com/GoogleCloudPlatform/gcloud-java/pull/889), javadoc and source building have been disabled unless the release profile is used (`-P release`). So i believe you can remove the `-Dmaven.javadoc.skip=true -Dmaven.source.skip=true` options (also in the following command).

jean-philippe-martin: Ah, most excellent! This makes the commands much simpler. Fixed.

jean-philippe-martin:
Thanks for finding the broken markdown! Of course I fixed it!
I pushed the new version with the changes, will merge the PR once the tests are green since you gave the LGTM.

jean-philippe-martin:
Oh I noticed some links in the README were broken (space between [text] and (link), no idea how it got there!) Fixed those too.

jean-philippe-martin:
Actually it looks like this is broken in a few other places too, for example gcloud-java-storage/README.md also has a space after the [text] in the header and the bit about semantic versioning. That can't be on purpose, can it?

mziccard:
> Actually it looks like this is broken in a few other places too, for example gcloud-java-storage/README.md also has a space after the [text] in the header and the bit about semantic versioning. That can't be on purpose, can it?

Yes there's no reason for the space to be there, but the anchor works anyway right? Feel free to merge this PR, I will open an issue for those extra spaces.

jean-philippe-martin:
Merging.

You're right, it seems to work fine on GitHub's web site. It's broken in my IDE's preview, which is how I found it.

PR #975, opened 2016-05-02
NIO CountBytes example

jean-philippe-martin:
A CountBytes examples that reads through the provided file (using NIO), counting the bytes. 

It also reports elapsed time, allowing for a simple form of benchmarking.

Amusingly, it currently says that the file sizes are off by one: it sees one fewer bytes than Files.size() reports. This is also noticeably slower than gsutil (perhaps because it's single-threaded?).

Sample output:

```
$ target/appassembler/bin/CountBytes <redacted>/dbsnp_138.b37.1.1-65M.vcf
<redacted>/dbsnp_138.b37.1.1-65M.vcf: 237432747 bytes.
Reading the whole file...
Read all 237432746 bytes in 8s. (6 calls to chan.read)
Wait, this doesn't match! We saw 237432746 bytes, yet the file size is listed at 237432747 bytes.
```

### DIFF
--- a/gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
+++ b/gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
@@ -30,7 +30,10 @@ import java.util.concurrent.TimeUnit;
 /**
  * CountBytes will read through the whole file given as input.
  *
- * <p>It's meant for testing that NIO doesn't read too slowly.
+ * <p>This example shows how to read a file size using NIO.
+ * File.size returns the size of the file as saved in Storage metadata.
+ * This class also shows how to read all of the file's contents using NIO,
+ * and reports how long it took.
  *
  * <p>See the README for compilation instructions. Run this code with
  * {@code target/appassembler/bin/CountBytes <file>}


mziccard:
Let me start with general comments:
> Amusingly, it currently says that the file sizes are off by one: it sees one fewer bytes than Files.size() reports.

This is not true. This error comes from the way you loop and compute `total` size:
```java
long total = 0;
int readCalls = 0;
for (int read = 1; read > 0; ) {
  readCalls ++;
  read = chan.read(buf);
  buf.flip();
  total += read; // On the last call you read -1 and sum it to total size, thus the mismatch
}
```
In general the for loop seems a bit strange to me, I would have used something like (not tested):
```java
while (chan.read(buf) > 0) {
  readCalls++;
  total += buf.position();
  buf.flip();
}
readCalls++; // We must count the last call
```

> This is also noticeably slower than gsutil (perhaps because it's single-threaded?).

What does this mean? Is it slower at reading the size metadata or at downloading the file? Can you share some numbers?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
```diff
@@ -0,0 +1,82 @@
+package com.google.cloud.examples.nio;
```
mziccard: Missing license header

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
```diff
@@ -0,0 +1,82 @@
...
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * CountBytes will read through the whole file given as input.
+ *
+ * <p>It's meant for testing that NIO correctly saturates your network link.
```
mziccard: Can you clarify what you mean with this?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
```diff
@@ -0,0 +1,82 @@
...
+      long size = Files.size(path);
+      System.out.println(fname + ": " + size + " bytes.");
+      ByteBuffer buf = ByteBuffer.allocate(bufSize);
+      System.out.println("Reading the whole file...");
+      Stopwatch sw = Stopwatch.createStarted();
+      SeekableByteChannel chan = Files.newByteChannel(path);
+      long total = 0;
+      int readCalls = 0;
+      for (int read = 1; read > 0; ) {
+        readCalls ++;
```
mziccard: space before `++` is not needed

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
```diff
@@ -0,0 +1,82 @@
...
+        read = chan.read(buf);
+        buf.flip();
+        total += read;
+      }
+      long elapsed = sw.elapsed(TimeUnit.SECONDS);
+      System.out.println("Read all " + total + " bytes in " + elapsed + "s. (" + readCalls +" calls to chan.read)");
```
mziccard: Line is too long (> 100 characters)

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
```diff
@@ -0,0 +1,82 @@
...
+      if (total != size) {
+        System.out.println("Wait, this doesn't match! We saw " + total + " bytes, yet the file size is listed at " + size + " bytes.");
```
mziccard: Line is too long (> 100 characters). Also I believe we should remove this once we fix the error with file size computation.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
```diff
@@ -0,0 +1,82 @@
...
+      }
+    } catch (Exception ex) {
+      System.out.println(fname + ": " + ex.toString());
+    }
+  }
+
+  private static void help() {
+    String[] help =
+        {"The argument is a <path>",
+         "and we show the length of that file.",
```
mziccard: Why the last `,`?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
```diff
@@ -0,0 +1,82 @@
...
+        };
+    for (String s : help) {
+      System.out.println(s);
+    }
+  }
+
```
mziccard: Remove this empty line.

jean-philippe-martin:
>> Amusingly, it currently says that the file sizes are off by one: it sees one fewer bytes than Files.size() reports.
> This is not true. This error comes from the way you loop and compute total size:

Thank you! Indeed it's obvious now.

>>This is also noticeably slower than gsutil (perhaps because it's single-threaded?).
> What does this mean? Is it slower at reading the size metadata or at downloading the file? Can you share some numbers?

Downloading the file. Here's what I see:
````
$ time gsutil cp gs://<redacted>/dbsnp_138.b37.256m.vcf /tmp/
Copying gs://<redacted>/dbsnp_138.b37.256m.vcf...
Downloading file:///tmp/dbsnp_138.b37.256m.vcf:                  214.89 MiB/214.89 MiB    
Downloading file:///tmp/dbsnp_138.b37.256m.vcf:                  214.89 MiB/214.89 MiB    
Downloading file:///tmp/dbsnp_138.b37.256m.vcf:                  214.89 MiB/214.89 MiB    
Downloading file:///tmp/dbsnp_138.b37.256m.vcf:                  214.89 MiB/214.89 MiB    

real	0m11.984s
user	0m7.618s
sys	0m4.513s
$ target/appassembler/bin/CountBytes gs://<redacted>dbsnp_138.b37.256m.vcf
gs://<redacted>/dbsnp_138.b37.256m.vcf: 901305932 bytes.
Reading the whole file...
Read all 901305931 bytes in 42s. (19 calls to chan.read)
````

We can see that gsutil takes 1/4th of the time. The way it prints it suggests that it did 4 parallel downloads (each of 1/th the size).

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java (hunk 1 above)
Comment (jean-philippe-martin) [inline gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java]: Added, and in Stat.java too since it looked like it was missing there as well.

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java (hunk 2 above)
Comment (jean-philippe-martin) [inline gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java]: I really meant that we can use this to see roughly how fast things go and to make sure we're not reading unnaturally slowly. I changed the text to reflect this more modest reality :)

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java (hunk 3 above)
Comment (jean-philippe-martin) [inline gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java]: Thank goodness.

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java (hunk 5 above)
Comment (jean-philippe-martin) [inline gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java]: Shortened the line. I want to keep the check, though - it's nice if the tool can detect if NIO is broken in that particular way (or, more likely, that the sample itself is broken).

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java (hunk 6 above)
Comment (jean-philippe-martin) [inline gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java]: Because I write too much [Go](https://tour.golang.org/moretypes/8). I removed the extra comma though since you don't like it.

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java (hunk 4 above)
Comment (jean-philippe-martin) [inline gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java]: Shortened.

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java [hunk 8]
```diff
@@ -0,0 +1,99 @@
...
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * CountBytes will read through the whole file given as input.
+ *
+ * <p>It's meant for testing that NIO doesn't read too slowly.
```
mziccard: I still don't like this. I would prefer to describe what the application does rather than stating that it's meant to evaluate performance (which is very opinable), something like:

> This example shows how to read a file size using NIO. `File.size` returns the size of the file as saved in Storage metadata. This class also shows how to compute the file's size by reading all bytes through a channel.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
mziccard: Feel free to keep it.

Personally I believe that these checks belong to tests rather than examples. Examples should show code that a user could actually be interested in writing. Translated: in an example I would show how to handle errors that can really occur at runtime, rather than errors that only occur if the example/library is broken.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
jean-philippe-martin: OK. How about the one below?

> This example shows how to read a file size using NIO.
>  File.size returns the size of the file as saved in Storage metadata.
>  This class also shows how to read all of the file's contents using NIO,
>  and reports how long it took.

Counting bytes itself isn't very useful, but an example for how to read the contents of the file makes it sound more useful.

mziccard: This sounds good to me!

jean-philippe-martin: Excellent, pushed!

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/CountBytes.java
jean-philippe-martin: Yes thanks, that makes a lot of sense. We should probably do this in an integration test as well. I see where you're coming from, but it also would feel wrong to have the sample count the bytes and then do nothing with the result.

jean-philippe-martin:
Are we OK to merge?

mziccard:
@jean-philippe-martin just merged. As usual: thanks!

jean-philippe-martin:
You are welcome!

PR #977, opened 2016-05-03
Add javadoc and unit tests for SubscriptionInfo

mziccard:


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/SubscriptionInfo.java
```diff
@@ -108,31 +157,60 @@ public SubscriptionInfo build() {
...
+  public boolean equals(Object obj) {
+    if (this == obj) {
       return true;
     }
-    if (o == null || getClass() != o.getClass()) {
+    if (obj == null || !obj.getClass().equals(this.getClass())) {
       return false;
     }
-    return Objects.equals(toPb(), ((SubscriptionInfo) o).toPb());
+    SubscriptionInfo other = (SubscriptionInfo) obj;
```
aozarov: Just curious. The protobuf equality does not work?

mziccard: Yes it works but now the `toPb()` method takes also the project id as a string. It seemed ugly to me to pass an empty string just for the purpose of comparing the objects

Issue #972, opened 2016-04-29
Add support for EmbeddedEntity in Datastore


volgin:
Currently there is no support for embedded entities and no mentions of them in the documentation or Datastore example.

mziccard:
Actually this is supported.
You can set an entity as a property of another entity by using either [`set(String, Value<?>)`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java#L140) and passing an [`EntityValue`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/EntityValue.java) or by using the newly released [`set(String, FullEntity<?>)`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java#L361).

volgin:
It will be nice to add this information to the Datastore example and JavaDoc.

mziccard:
> It will be nice to add this information to the Datastore example and JavaDoc.

Javadoc should already be in place. I will take care of adding this to the example.

mziccard:
An example for datastore's embedded entities has been added in #980, see [here](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java#L215).

PR #980, opened 2016-05-04
Refactor Datastore example, add example on embedded entities

mziccard:
This PR refactors `DatastoreExample` class to have the same structure as other examples. This also adds an example for embedded entities, thus fixing #972

### DIFF
--- a/gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
+++ b/gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
@@ -66,7 +66,7 @@ public class DatastoreExample {
 
   private abstract static class DatastoreAction<T> {
 
-    abstract void run(Transaction tx, Key userKey, T request) throws Exception;
+    abstract void run(Transaction tx, Key userKey, T arg) throws Exception;
 
     abstract T parse(String... args) throws Exception;
 
@@ -81,7 +81,7 @@ public class DatastoreExample {
    */
   private static class DeleteAction extends DatastoreAction<Void> {
     @Override
-    public void run(Transaction tx, Key userKey, Void request) {
+    public void run(Transaction tx, Key userKey, Void arg) {
       Entity user = tx.get(userKey);
       if (user == null) {
         System.out.println("Nothing to delete, user does not exist.");
@@ -114,17 +114,18 @@ public class DatastoreExample {
    */
   private static class DisplayAction extends DatastoreAction<Void> {
     @Override
-    public void run(Transaction tx, Key userKey, Void request) {
+    public void run(Transaction tx, Key userKey, Void arg) {
       Entity user = tx.get(userKey);
       if (user == null) {
-        System.out.println("User '" + userKey.name() + "' does not exist.");
+        System.out.printf("User '%s' does not exist.%n", userKey.name());
         return;
       }
       if (user.contains("contact")) {
         FullEntity<IncompleteKey> contact = user.getEntity("contact");
         String email = contact.getString("email");
         String phone = contact.getString("phone");
-        System.out.printf("User '%s' email is '%s', phone is '%s'%n", userKey.name(), email, phone);
+        System.out.printf("User '%s' email is '%s', phone is '%s'.%n",
+            userKey.name(), email, phone);
       }
       System.out.printf("User '%s' has %d comment[s].%n", userKey.name(), user.getLong("count"));
       int limit = 200;
@@ -172,7 +173,7 @@ public class DatastoreExample {
       Entity user = tx.get(userKey);
       if (user == null) {
         System.out.println("Adding a new user.");
-        user = Entity.builder(userKey).set("count", 1L).build();
+        user = Entity.builder(userKey).set("count", 1).build();
         tx.add(user);
       } else {
         user = Entity.builder(user).set("count", user.getLong("count") + 1L).build();
@@ -184,7 +185,7 @@ public class DatastoreExample {
           .set("timestamp", DateTime.now())
           .build();
       tx.addWithDeferredIdAllocation(comment);
-      System.out.println("Adding a comment to user '" + userKey.name() + "'.");
+      System.out.printf("Adding a comment to user '%s'.%n", userKey.name());
     }
 
     @Override
@@ -245,7 +246,7 @@ public class DatastoreExample {
           .set("phone", contact.phone())
           .build();
       tx.update(Entity.builder(user).set("contact", contactEntity).build());
-      System.out.println("Setting contact for user '" + userKey.name() + "'.");
+      System.out.printf("Setting contact for user '%s'.%n", userKey.name());
     }
 
     @Override
@@ -281,7 +282,7 @@ public class DatastoreExample {
 
       String param = entry.getValue().params();
       if (param != null && !param.isEmpty()) {
-        actionAndParams.append(' ').append(param.replace("\n", "\n\t\t"));
+        actionAndParams.append(' ').append(param);
       }
     }
     System.out.printf("Usage: %s <project_id> <user> operation <args>*%s%n",
@@ -316,8 +317,8 @@ public class DatastoreExample {
     try {
       request = action.parse(args);
     } catch (IllegalArgumentException ex) {
-      System.out.println("Invalid input for action '" + actionName + "'. " + ex.getMessage());
-      System.out.println("Expected: " + action.params());
+      System.out.printf("Invalid input for action '%s'. %s%n", actionName, ex.getMessage());
+      System.out.printf("Expected: %s%n", action.params());
       return;
     } catch (Exception ex) {
       System.out.println("Failed to parse request.");
--- a/gcloud-java-examples/src/main/java/com/google/cloud/examples/bigquery/BigQueryExample.java
+++ b/gcloud-java-examples/src/main/java/com/google/cloud/examples/bigquery/BigQueryExample.java
@@ -108,7 +108,7 @@ public class BigQueryExample {
 
   private abstract static class BigQueryAction<T> {
 
-    abstract void run(BigQuery bigquery, T request) throws Exception;
+    abstract void run(BigQuery bigquery, T arg) throws Exception;
 
     abstract T parse(String... args) throws Exception;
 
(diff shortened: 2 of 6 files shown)


gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -56,15 +64,24 @@
...
-    void run(Transaction tx, Key userKey, String... args);
+  private abstract static class DatastoreAction<T> {
+
+    abstract void run(Transaction tx, Key userKey, T request) throws Exception;
 
-    String getRequiredParams();
+    abstract T parse(String... args) throws Exception;
+
+    protected String params() {
+      return "";
```
jgeewax: Is this right? Empty string? Not null? or abstract?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -86,19 +103,29 @@ public void run(Transaction tx, Key userKey, String... args) {
...
 
-  private static class DisplayAction implements DatastoreAction {
+  /**
+   * This class demonstrates how to get a user. The action also queries all comments associated
+   * with the user.
+   */
+  private static class DisplayAction extends DatastoreAction<Void> {
     @Override
-    public void run(Transaction tx, Key userKey, String... args) {
+    public void run(Transaction tx, Key userKey, Void request) {
```
jgeewax: Why do we call this `request`? Seems a bit misleading...

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -56,15 +64,24 @@
...
 
-  private static class DeleteAction implements DatastoreAction {
+  /**
+   * This class demonstrates how to delete a user. This action also queries the keys of all comments
+   * associated with the user and uses them to delete comments.
+   */
+  private static class DeleteAction extends DatastoreAction<Void> {
     @Override
-    public void run(Transaction tx, Key userKey, String... args) {
+    public void run(Transaction tx, Key userKey, Void request) {
```
jgeewax: Why `request` ?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -86,19 +103,29 @@ public void run(Transaction tx, Key userKey, String... args) {
...
+   * with the user.
+   */
+  private static class DisplayAction extends DatastoreAction<Void> {
     @Override
-    public void run(Transaction tx, Key userKey, String... args) {
+    public void run(Transaction tx, Key userKey, Void request) {
       Entity user = tx.get(userKey);
       if (user == null) {
-        System.out.println("No comments for '" + userKey.name() + "'.");
+        System.out.println("User '" + userKey.name() + "' does not exist.");
```
jgeewax: Might be a bit more readable (and easier to edit) with `String.format`?

```java
System.out.println(String.format("User '%s' does not exist.", userKey.name());
```
?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -86,19 +103,29 @@ public void run(Transaction tx, Key userKey, String... args) {
...
         return;
       }
+      if (user.contains("contact")) {
+        FullEntity<IncompleteKey> contact = user.getEntity("contact");
+        String email = contact.getString("email");
+        String phone = contact.getString("phone");
+        System.out.printf("User '%s' email is '%s', phone is '%s'%n", userKey.name(), email, phone);
```
jgeewax: Missing `.` before `%n` ?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -131,25 +158,37 @@ public void run(Transaction tx, Key userKey, String... args) {
...
     @Override
-    public void run(Transaction tx, Key userKey, String... args) {
+    public void run(Transaction tx, Key userKey, String content) {
       Entity user = tx.get(userKey);
       if (user == null) {
         System.out.println("Adding a new user.");
-        user = Entity.builder(userKey)
-            .set("count", 1L)
-            .build();
+        user = Entity.builder(userKey).set("count", 1L).build();
```
jgeewax: What happens if you do `1` instead of `1L` ?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -131,25 +158,37 @@ public void run(Transaction tx, Key userKey, String... args) {
...
         tx.add(user);
       } else {
         user = Entity.builder(user).set("count", user.getLong("count") + 1L).build();
         tx.update(user);
       }
+      IncompleteKey commentKey = IncompleteKey.builder(userKey, COMMENT_KIND).build();
```
jgeewax: Why do we pull `COMMENT_KIND` out here? Seems like most of this file is in-line strings floating around.

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -131,25 +158,37 @@ public void run(Transaction tx, Key userKey, String... args) {
...
+      FullEntity<IncompleteKey> comment = FullEntity.builder(commentKey)
+          .set("content", content)
+          .set("timestamp", DateTime.now())
+          .build();
+      tx.addWithDeferredIdAllocation(comment);
+      System.out.println("Adding a comment to user '" + userKey.name() + "'.");
```
jgeewax: Might be better with `printf`

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -159,28 +198,98 @@ public void run(Transaction tx, Key userKey, String... args) {
...
-      IncompleteKey commentKey = IncompleteKey.builder(userKey, COMMENT_KIND).build();
-      FullEntity<IncompleteKey> comment = FullEntity.builder(commentKey)
-          .set("content", content)
-          .set("timestamp", DateTime.now())
+      return content;
+    }
+
+    @Override
+    protected String params() {
+      return "<comment>";
```
jgeewax: If this takes multiple, wouldn't it make sense to return `String[]` and then have the thing that prints the help-text render that with `<>`'s and all that?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -159,28 +198,98 @@ public void run(Transaction tx, Key userKey, String... args) {
...
+        tx.add(user);
+      }
+      FullEntity<IncompleteKey> contactEntity = FullEntity.builder()
+          .set("email", contact.email())
+          .set("phone", contact.phone())
           .build();
-      tx.addWithDeferredIdAllocation(comment);
-      System.out.println("Adding a comment to user '" + userKey.name() + "'.");
+      tx.update(Entity.builder(user).set("contact", contactEntity).build());
+      System.out.println("Setting contact for user '" + userKey.name() + "'.");
```
jgeewax: `printf`?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
```diff
@@ -159,28 +198,98 @@ public void run(Transaction tx, Key userKey, String... args) {
...
 
-  public static void main(String... args) {
+  private static void printUsage() {
+    StringBuilder actionAndParams = new StringBuilder();
+    for (Map.Entry<String, DatastoreAction> entry : ACTIONS.entrySet()) {
+      actionAndParams.append("\n\t").append(entry.getKey());
+
+      String param = entry.getValue().params();
+      if (param != null && !param.isEmpty()) {
+        actionAndParams.append(' ').append(param.replace("\n", "\n\t\t"));
```
jgeewax: What's happening here? Maybe a comment would be helpful?

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
mziccard: This returns an empty string so that extending actions that take no arguments can avoid redefining.

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
mziccard: I agree, will change!

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
mziccard: Nothing, 1 is automatically promoted to long. `1L` was used to show that the method takes long parameter.

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
mziccard: All constants in this file (as `COMMENT_KIND`) are those values that users might want to change before running the example (might clash with existing-kinds). Other strings floating are names for entity properties, that won't clash.

gcloud-java-examples/src/main/java/com/google/cloud/examples/datastore/DatastoreExample.java
mziccard: This returns the command line help which can be more complex of a simple list of tokens to surround with `<>`. For instance we might also have info on the cardinality `?*+` or grouping `(<value1><value2>)`. This is used in other module's examples so I would keep it as is for consistency.

PR #983, opened 2016-05-04
Add ParallelCountBytes

jean-philippe-martin:
Made it and CountBytes compute an MD5, so I could check that they match
(they do).

### DIFF
--- a/gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
+++ b/gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
@@ -19,6 +19,7 @@ package com.google.cloud.examples.nio;
 import com.google.common.base.Stopwatch;
 import com.google.common.io.BaseEncoding;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.ByteBuffer;
@@ -49,8 +50,16 @@ public class ParallelCountBytes {
 
   /**
    * WorkUnit holds a buffer and the instructions for what to put in it.
+   *
+   * <p>Use it like this:
+   * <ol>
+   *   <li> call()
+   *   <li> the data is now in buf, you can access it directly
+   *   <li> if need more, call resetForIndex(...) and go back to the top.
+   *   <li> else, call close()
+   * </ol>
    */
-  private class WorkUnit implements Callable<WorkUnit> {
+  private static class WorkUnit implements Callable<WorkUnit>, Closeable {
     public final ByteBuffer buf;
     final SeekableByteChannel chan;
     final int blockSize;
@@ -70,7 +79,7 @@ public class ParallelCountBytes {
         return this;
       }
       chan.position(pos);
-      // read until buffer is full, or EOF
+      // read until buffer it is full, or EOF
       while (chan.read(buf) > 0) {};
       return this;
     }
@@ -80,16 +89,16 @@ public class ParallelCountBytes {
       buf.flip();
       return this;
     }
+
+    public void close() throws IOException {
+      chan.close();
+    }
   }
 
   /**
    * See the class documentation.
    */
-  public static void main(String[] args) throws IOException {
-    new ParallelCountBytes().start(args);
-  }
-
-  public void start(String[] args) throws IOException {
+  public static void main(String[] args) throws Exception {
     if (args.length == 0 || args[0].equals("--help")) {
       help();
       return;
@@ -100,16 +109,15 @@ public class ParallelCountBytes {
   }
 
   /**
-   * Print the length of the indicated file.
+   * Print the length and MD5 of the indicated file.
    *
    * <p>This uses the normal Java NIO Api, so it can take advantage of any installed
    * NIO Filesystem provider without any extra effort.
    */
-  private void countFile(String fname) throws IOException{
+  private static void countFile(String fname) throws Exception {
     // large buffers pay off
     final int bufSize = 50 * 1024 * 1024;
     Queue<Future<WorkUnit>> work = new ArrayDeque<>();
-    try {
       Path path = Paths.get(new URI(fname));
       long size = Files.size(path);
       System.out.println(fname + ": " + size + " bytes.");
@@ -125,14 +133,15 @@ public class ParallelCountBytes {
       for (blockIndex = 0; blockIndex < nThreads; blockIndex++) {
         work.add(exec.submit(new WorkUnit(Files.newByteChannel(path), bufSize, blockIndex)));
       }
-      while (true) {
+      while (!work.isEmpty()) {
         WorkUnit full = work.remove().get();
         md.update(full.buf.array(), 0, full.buf.position());
         total += full.buf.position();
         if (full.buf.hasRemaining()) {
-          break;
+          full.close();
+        } else {
+          work.add(exec.submit(full.resetForIndex(blockIndex++)));
         }
-        work.add(exec.submit(full.resetForIndex(blockIndex++)));
       }
       exec.shutdown();
 
@@ -141,12 +150,9 @@ public class ParallelCountBytes {
       String hex = String.valueOf(BaseEncoding.base16().encode(md.digest()));
       System.out.println("The MD5 is: 0x" + hex);
       if (total != size) {
-        System.out.println("Wait, this doesn't match! We saw " + total + " bytes, " +
-            "yet the file size is listed at " + size + " bytes.");
+        System.out.println("Wait, this doesn't match! We saw " + total + " bytes, "
+            + "yet the file size is listed at " + size + " bytes.");
       }
-    } catch (Exception ex) {
-      System.out.println(fname + ": " + ex.toString());
-    }
   }
 
   private static void help() {


jean-philippe-martin:
Inspired by the [performance difference](https://github.com/GoogleCloudPlatform/gcloud-java/pull/975#issuecomment-216610819) between the previous CL and gsutil I wrote a simple example of a multithreaded reader that keeps the bytes in order for processing.

This indeed goes faster than before: 12s vs the earlier 42s. So now we match gsutil's 11.98s.
````
$ target/appassembler/bin/ParallelCountBytes gs://<redacted>/dbsnp_138.b37.256m.vcf
gs://<redacted>/dbsnp_138.b37.256m.vcf: 901305932 bytes.
Reading the whole file using 4 threads...
Read all 901305932 bytes in 12s. 
The MD5 is: 0xDD9CE9E581559B57F9E08A8883276AE6
````

File: gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java [hunk 1]
```diff
@@ -0,0 +1,174 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.examples.nio;
+
+import com.google.common.base.Stopwatch;
+
+import javax.xml.bind.annotation.adapters.HexBinaryAdapter;
```
mziccard: I would prefer not to use this (it's very out of context here and I believe not available in Android).

Instead we could use Guava's [Hasher](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/hash/Hasher.html). The code becomes:
```java
Hasher hasher = Hashing.md5().newHasher();
for (int block = 0; block < nBlocks; block++) {
  BufWithLock bwl = bufs[block % bufs.length];
  synchronized (bwl.lock) {
    while (!bwl.full) {
      bwl.lock.wait();
    }
    hasher.putBytes(bwl.buf.array(), 0, bwl.buf.position());
    total += bwl.buf.position();
    bwl.full = false;
    bwl.lock.notifyAll();
  }
}
long elapsed = sw.elapsed(TimeUnit.SECONDS);
System.out.println("Read all " + total + " bytes in " + elapsed + "s. ");
System.out.println("The MD5 is: 0x" + hasher.hash().toString());
```

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.MessageDigest;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * ParallelCountBytes will read through the whole file given as input.
+ *
+ * <p>This example shows how to go through all the contents of a file,
+ * in order, using multithreaded NIO reads.It also reports how long it took.
```
mziccard: I would also mention we compute the MD5 (also in `CountBytes` example).

space needed in `reads.It`

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * <p>See the README for compilation instructions. Run this code with
+ * {@code target/appassembler/bin/ParallelCountBytes <file>}
+ */
+public class ParallelCountBytes {
+
+  private class BufWithLock {
+    public Object lock;
```
mziccard: Should be final

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+    public ByteBuffer buf;
```
mziccard: Should be final

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+    public boolean full;
+    public Thread t;
```
mziccard: Is saving the thread here really needed?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+      synchronized(output.lock) {
+        while (true) {
+          if (pos > chan.size()) {
+            break;
+          }
+          chan.position(pos);
+          // read until buffer is full, or EOF
+          while (chan.read(output.buf) > 0) {};
+          output.full = true;
+          output.lock.notifyAll();
```
mziccard: There's at most one thread waiting here right?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+   * NIO Filesystem provider without any extra effort.
+   */
+  private void countFile(String fname) throws IOException{
+    // large buffers pay off
+    final int bufSize = 50 * 1024 * 1024;
+    try {
+      Path path = Paths.get(new URI(fname));
+      long size = Files.size(path);
+      System.out.println(fname + ": " + size + " bytes.");
+      ByteBuffer buf = ByteBuffer.allocate(bufSize);
```
mziccard: Is this ever used?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+      int nBlocks = (int)Math.ceil( size / (double)bufSize);
```
mziccard: Space is needed between the cast indication and the casted value, no space is needed after open `(`:
```java
int nBlocks = (int) Math.ceil(size / (double) bufSize);
```

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+        final SeekableByteChannel chan = Files.newByteChannel(path);
+        final int finalNThreads = nThreads;
+        final int finalI = i;
+        bufs[i].t = new Thread(new Runnable() {
+          @Override
+          public void run() {
+            stridedRead(chan, bufSize, finalI, finalNThreads, bufs[finalI]);
+          }
+        });
+            bufs[i].t.start();
```
mziccard: Too much indentation

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -0,0 +1,174 @@
...
+      for (int block = 0; block < nBlocks; block++) {
+        BufWithLock bwl = bufs[block % bufs.length];
+        synchronized (bwl.lock) {
+          while (!bwl.full) {
+            bwl.lock.wait();
+          }
+          md.update(bwl.buf.array(), 0, bwl.buf.position());
+          total += bwl.buf.position();
+          bwl.full = false;
+          bwl.lock.notifyAll();
```
mziccard: There should be always at most one thread waiting there right?

mziccard:
This looks fine to me, however I believe this example cloud be easier if using a [fixed size ExecutorService](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newFixedThreadPool(int)).

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: Counter-proposal: keep Java's hashing code but use Guava's `BaseEncoding` to do the hex conversion. This way it works on Android but we don't have to change the way we hash the data.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: Thank you, fixed.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: Done, and for buf too.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: Not really, no. Sometimes it's handy to keep the Thread object around to be able to interrupt it, but we're not doing that here. I removed it.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: Yes; as a habit I always use notifyAll since it's hard to spot a bug where one is calling notify instead of notifyAll (the choice of which to use is global, it depends on the whole rest of the code). Would you like me to switch to notify here?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: I forgot to remove that one, thank you!

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: done.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: Indentation gone.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: yes; same rationale as above.

jean-philippe-martin:
OK, I think you'll find this version much cleaner. Its output still matches.

jean-philippe-martin:
Is this OK for merge?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -70,7 +79,7 @@ public WorkUnit call() throws IOException {
         return this;
       }
       chan.position(pos);
-      // read until buffer is full, or EOF
+      // read until buffer it is full, or EOF
```
mziccard: Why the extra `it`?

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
```diff
@@ -80,16 +89,16 @@ public WorkUnit resetForIndex(int blockIndex) {
       buf.flip();
       return this;
     }
+
+    public void close() throws IOException {
+      chan.close();
```
mziccard: Channel must be closed also in the `CountBytes` example.

jean-philippe-martin:
You know, this PR is now a perfect illustration of a problem with the current adversarial way that review is done. One person's job is to write the code and fix it, the other person's job is to make sure it doesn't go in with anything bad in it. This seems reasonable but the result is that even with perfectly reasonable and capable people we occasionally end up with a PR like this one that is delayed by 3 weeks just because of a typo and a disagreement about what belongs in this PR or another. This sort of delay could be very painful and make a PR hard to merge (luckily that doesn't seem to be the case here).

If we instead had a collaborative model where the coder could indicate that they trust the reviewer to just go ahead and fix typos directly themselves, together we could get things moving a lot quicker.

lesv:
Jean - apologies - I'm trying to help clear a bit of the backlog, but it's taking longer for me to be meaningful than it should.  It looks like this can still be merged.  Unfortunately, GH tools aren't the best for reviewer submitted changes to a PR. (unlike internal ones)

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
lesv: #1035 is issue for this.

lesv:
LGTM

mziccard:
@jean-philippe-martin Please don't read this latency in merging the PR as me not appreciating your work/effort (I appreciate both in fact). This PR adds an example to a non-master branch that is (sadly) stalling for other reasons (see #770). So, even though it might be frustrating from your viewpoint, there was no actual reason to speed the merge. Otherwise, I would have cherry-picked your comments in a different PR and added the fixes myself. That said, thanks Jean-Philippe :)

@jean-philippe-martin @lesv Regardless of the more general discussion, I would still love to see my comments (unnecessary "it" and closing channel in `CountBytes`) fixed. Now I can take care of them myself without cherry-picking :)

jean-philippe-martin:
@mziccard no offense taken, this is really about the process and not the people. And you're absolutely right, since this branch is off-master there isn't as much change and so not as much urgency to merge. I agree 100% with you that both your comments should be addressed. Since this was merged I'll do it in a different PR.

@lesv thanks, and no need to apologize! We're doing things the normal way, it's just that the normal way might perhaps be improvable. No harm done in this case anyways.

gcloud-java-examples/src/main/java/com/google/cloud/examples/nio/ParallelCountBytes.java
jean-philippe-martin: No reason, that was a mistake. #1036 removes it.

Issue #978, opened 2016-05-03
PubSub.listSubscriptions(topic) returns only names, not full subscriptions


mziccard:
We need to change the signature of:
- `Page<Subscription> listSubscriptions(String topic, ListOption... options)`
- `Future<AsyncPage<Subscription>> listSubscriptionsAsync(String topic, ListOption... options);`

to:
- `Page<String> listSubscriptions(String topic, ListOption... options)`
- `Future<AsyncPage<String>> listSubscriptionsAsync(String topic, ListOption... options);`

As listing subscriptions by topic returns only names (see [here](https://cloud.google.com/pubsub/reference/rpc/google.pubsub.v1#google.pubsub.v1.ListTopicSubscriptionsResponse)). /cc @aozarov

eschapira:
The topic owner/viewer may not have access to the subscription details; but they can see the name of the subscriptions and delete those subscriptions.

If we are using Subscription objects, as long as you don't need to fetch the details in order to instantiate it, meaning that you can instantiate a subscription object by just using the name, it should be fine to return the subscription object.

It is important to keep in mind that these subscriptions may belong to a different project.

mziccard:
This was taken care of in #984

PR #984, opened 2016-05-05
Add TopicId and SubscriptionId classes

mziccard:
This PR adds the `TopicId` class, thus allowing to create subscriptions for topics that reside in different projects.
This PR also adds the `SubscriptionId` class that is returned by `pubsub.list(topic)` and `pubsub.listAsync(topic)` (thus fixing #978).

/cc @aozarov

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/SubscriptionInfo.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/SubscriptionInfo.java
@@ -182,6 +182,8 @@ public class SubscriptionInfo implements Serializable {
   /**
    * Returns the identity of the topic this subscription refers to. If {@link TopicId#project()} is
    * {@code null} the topic is assumed to reside in the {@link PubSubOptions#projectId()} project.
+   * After a topic is deleted, existing subscriptions to that topic are not deleted, but their topic
+   * field is set to {@link TopicId#deletedTopic()}.
    */
   public TopicId topic() {
     return topic;
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/TopicId.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/TopicId.java
@@ -33,13 +33,21 @@ import java.util.Objects;
 public final class TopicId implements Serializable {
 
   private static final long serialVersionUID = -4913169763174877777L;
+  private static final String DELETED_TOPIC_NAME = "_deleted_topic_";
+  private static final TopicId DELETED_TOPIC = new TopicId(null, DELETED_TOPIC_NAME, true);
 
   private final String project;
   private final String topic;
+  private final boolean isDeleted;
 
-  private TopicId(String project, String topic) {
+  private TopicId(String project, String topic, boolean isDeleted) {
     this.project = project;
     this.topic = checkNotNull(topic);
+    this.isDeleted = isDeleted;
+  }
+
+  private TopicId(String project, String topic) {
+    this(project, topic, false);
   }
 
   /**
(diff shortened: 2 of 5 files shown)


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/SubscriptionId.java
```diff
@@ -0,0 +1,89 @@
...
+
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Identity for a Google PubSub subscription. {@code SubscriptionId} objects are returned by the
+ * {@link PubSub#listSubscriptions(String, PubSub.ListOption...)} and
+ * {@link PubSub#listSubscriptionsAsync(String, PubSub.ListOption...)} methods as the same topic
```
aozarov: Alternative:
as a topic may have subscriptions from different projects.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/SubscriptionInfo.java
```diff
@@ -158,9 +180,10 @@ public SubscriptionInfo build() {
   }
 
   /**
-   * Returns the name of the topic this subscription refers to.
+   * Returns the identity of the topic this subscription refers to. If {@link TopicId#project()} is
+   * {@code null} the topic is assumed to reside in the {@link PubSubOptions#projectId()} project.
```
aozarov: I assume this will never be `null` for `Subscription`?

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/SubscriptionInfo.java
```diff
@@ -303,6 +362,22 @@ public static SubscriptionInfo of(String topic, String name, String endpoint) {
...
+   * Creates a builder for {@code SubscriptionInfo} objects given the identity of the topic and the
+   * name of the subscription. If {@code topic.project()} is {@code null} the topic is assumed to
+   * reside in the {@link PubSubOptions#projectId()} project.
+   *
+   * @param topic the identity of the topic the subscription refers to
+   * @param name the name of the subscription. The name must start with a letter, and contain only
+   *     letters ({@code [A-Za-z]}), numbers ({@code [0-9]}), dashes ({@code -}), underscores
+   *     ({@code _}), periods ({@code .}), tildes ({@code ~}), plus ({@code +}) or percent signs
+   *     ({@code %}). It must be between 3 and 255 characters in length and cannot begin with the
+   *     string {@code goog}
```
aozarov: Nit, '.' (http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@param).

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/TopicId.java
```diff
@@ -0,0 +1,102 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.base.MoreObjects;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Identity for a Google PubSub topic. A {@code TopicId} object can be used to create subscriptions
+ * for topics, possibly residing on a different project.
```
aozarov: I think "that" or ", which" sounds nicer than just ",". 
Also, I think "belong to different" sounds better and in either case I think "project" needs to be plural.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/TopicId.java
```diff
@@ -0,0 +1,102 @@
...
+  private final String topic;
+
+  private TopicId(String project, String topic) {
+    this.project = project;
+    this.topic = checkNotNull(topic);
+  }
+
+  /**
+   * Returns the name of the project where the topic resides.
+   */
```
aozarov: should you mention the `null` case?

aozarov:
Not a thorough review by any means but this change looks good to me.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/SubscriptionInfo.java
mziccard: Exactly, for `Subscription` objects `TopicId.project()` is always set.

PR #1000, opened 2016-05-11
Add tests and javadoc for Message and ByteArray

mziccard:


gcloud-java-core/src/main/java/com/google/cloud/ByteArray.java
```diff
@@ -75,91 +74,91 @@ public final boolean equals(Object obj) {
...
     return new ByteArray(ByteString.copyFrom(bytes));
   }
 
   /**
-   * Copy the bytes using {@code UTF-8} decoding.
+   * Creates a {@code ByteArray} object given a string. The string is encoded in {@code UTF-8}. The
+   * bytes are copied.
    */
-  public static final ByteArray copyFrom(String string) {
+  public static ByteArray copyFrom(String string) {
```
jgeewax: why chopping the final ?

mziccard: Removed by mistake, will add them back.

PR #1001, opened 2016-05-11
Make PubSub and PubSubRpc extends AutoCloseable, fix grpc settings

mziccard:
This PR does the following
- Make `PubSub` and `PubSubRpc` extend `AutoCloseable`
- Implement `close()` in `PubSubImpl` and `DefaultPubSubRpc` 
- Create a common `ScheduledExecutorService` for both `PublisherApi` and `SubscriberApi` and add `executor()` getter to `PubSubRpc`
- Fix the way we create `PublisherSettings` and `SubscriberSettings` to support testing with local emulator
- Fix initialization of retry settings

/cc @aozarov @jgeewax @lesv

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
@@ -19,6 +19,8 @@ package com.google.cloud.pubsub.spi;
 import com.google.api.gax.core.RetrySettings;
 import com.google.api.gax.grpc.ApiCallSettings;
 import com.google.api.gax.grpc.ApiException;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.AuthCredentials;
 import com.google.cloud.RetryParams;
 import com.google.cloud.pubsub.PubSubException;
 import com.google.cloud.pubsub.PubSubOptions;
@@ -27,8 +29,10 @@ import com.google.cloud.pubsub.spi.v1.PublisherSettings;
 import com.google.cloud.pubsub.spi.v1.SubscriberApi;
 import com.google.cloud.pubsub.spi.v1.SubscriberSettings;
 import com.google.common.base.Function;
+import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.protobuf.Empty;
 import com.google.pubsub.v1.AcknowledgeRequest;
 import com.google.pubsub.v1.DeleteSubscriptionRequest;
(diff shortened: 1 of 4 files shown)


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
```diff
@@ -50,48 +54,78 @@
...
-      SubscriberSettings.Builder sBuilder =
-          SubscriberSettings.defaultBuilder()
-              .provideChannelWith(options.authCredentials().credentials())
-              .applyToAllApiMethods(apiCallSettings(options));
-      subscriberApi = SubscriberApi.create(sBuilder.build());
+      PublisherSettings.Builder pubBuilder =
+          PublisherSettings.defaultBuilder().provideExecutorWith(executor, false);
+      SubscriberSettings.Builder subBuilder =
+          SubscriberSettings.defaultBuilder().provideExecutorWith(executor, false);
+      // todo(mziccard): PublisherSettings should support null/absent credentials for testing
```
lesv: Most users will have `gcloud beta auth application-default login` of some kind set, and all of our testing environments should.  So this might not be necessary.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
```diff
@@ -117,7 +151,7 @@ public V apply(ApiException exception) {
 
   @Override
   public Future<Topic> create(Topic topic) {
-    // TODO: it would be nice if we can get the idempotent inforamtion from the ApiCallSettings
+    // TODO: it would be nice if we can get the idempotent information from the ApiCallSettings
```
lesv: Do you use Google Guava?

lesv:
Seems ok to merge, my comments were on your todo's.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
```diff
@@ -50,48 +54,78 @@
...
-import io.grpc.Status.Code;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 
 public class DefaultPubSubRpc implements PubSubRpc {
 
   private final PublisherApi publisherApi;
   private final SubscriberApi subscriberApi;
+  private final ScheduledExecutorService executor =
+      MoreExecutors.getExitingScheduledExecutorService(new ScheduledThreadPoolExecutor(8));
```
aozarov: Per earlier discussion that is not going to work for AE Standard. A comment?

Also, I don't think using `MoreExecutors.getExitingScheduledExecutorService` is a good idea unless that was `static`. Regardless the static part, a better option would be to make sure we are using a thread pool that shrink
to zero threads after a reasonable period of no use.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
aozarov: This is also used by unit-tests and I don't think we should require an auth logic for them. I think comment is valid and that VTKIT should support this option.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
```diff
@@ -50,48 +54,78 @@
...
+      pubBuilder.applyToAllApiMethods(apiCallSettings(options));
+      subBuilder.applyToAllApiMethods(apiCallSettings(options));
+      publisherApi = PublisherApi.create(pubBuilder.build());
+      subscriberApi = SubscriberApi.create(subBuilder.build());
     } catch (Exception ex) {
       throw new IOException(ex);
     }
   }
 
+  private static long translateTimeout(long timeout) {
```
aozarov: Worth thinking how to make this logic common for all services (which can override a different default value).

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
aozarov: Guava is used internally (see imports for this file) but not exposed in the API.
Are you asking because you would have expected it to return 'ListenableFuture'?

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/PubSubRpc.java
```diff
@@ -69,4 +70,6 @@
   Future<PullResponse> pull(PullRequest request);
 
   Future<Empty> modify(ModifyPushConfigRequest request);
+
+  ScheduledExecutorService executor();
```
aozarov: Probably fine for now, but we should think how we can avoid thread-pools proliferation when using multiple gcloud services.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
mziccard: Yes I agree, it's out of the scope of this PR but we should probably address this soon. Couldn't think of an effective way of doing that though.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/PubSubRpc.java
mziccard: Yes I believe we should allow users to set the executor via the service options. That way the same executor can be reused across services.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
mziccard: > A comment?

Will add one. Also I believe `getExitingScheduledExecutorService` won't work on AE as it requires permission to register system hooks.

> Also, I don't think using MoreExecutors.getExitingScheduledExecutorService is a good idea unless that was static

Can you provide some more details? Why should local vs static declaration matter in this case?

> Regardless the static part, a better option would be to make sure we are using a thread pool that shrink
to zero threads after a reasonable period of no use.

The VTIK layer uses non-shrinkable pools so I guessed that was the preferred choice. /cc @garrettjonesgoogle 
Also we should decide the `corePoolSize` and `keepAliveTime` parameters, which I am not so sure about. I know that cached thread pool use `keepAliveTime=60seconds`. I wouldn't reduce `corePoolSize` to 0 though.

aozarov: > Can you provide some more details? Why should local vs static declaration matter in this case?

Every time you call this it adds a new Shutdown hook.

> . I wouldn't reduce corePoolSize to 0 though.

Why?

mziccard: > Every time you call this it adds a new Shutdown hook.

I don't see the problem with that. If more executors are created more must be closed.

> Why?

For the cost of creating threads. The library uses threads, it does not seem a waste of resources to me to always have a minimum-size pool ready to be more responsive.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/PubSubRpc.java
mziccard: Opened #1005 for this

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
aozarov: > I don't see the problem with that. If more executors are created more must be closed.

Are you sure closing the executor will *also* remove it from the Shutdown hook?

> For the cost of creating threads. The library uses threads, it does not seem a waste of resources to me to always have a minimum-size pool ready to be more responsive.

If we have one thread-pool for all gcloud-java service that would make sense. Otherwise it is a waste.
If one does not call a service for N minutes. At some point the cost balance of creating a thread and caching vs just holding the resource for that long will shift.
for too long

mziccard: > Are you sure closing the executor will also remove it from the Shutdown hook?

I just checked and it does not remove the hook. But what is your concern here? The extra thread needed for the hook or the possible double shutdown? Double shutdown should not be an issue. The idea of using `getExitingScheduledExecutorService` was to ensure that the executor is shutdown even if the user forgets.

aozarov: Consider a long running application that uses various services and keeps creating and closing them. This implementation is basically piling up both memory and the need to iterate all of them upon shutdown.

mziccard:
This can be closed as was replaced by #1011

PR #1002, opened 2016-05-11
Use updated datastore client that permits IP addresses for localhost

ajkannan:
See GoogleCloudPlatform/google-cloud-datastore#117 for context for the update.


mziccard:
Thanks @ajkannan!

PR #1009, opened 2016-05-13
Regenerated pubsub surface

michaelbausor:
Improved code samples
Client side validation of resource names
Updated GAX

### DIFF
--- a/gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
+++ b/gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
@@ -288,7 +288,7 @@ public class PublisherApi implements AutoCloseable {
    * <!-- manual edit -->
    * <!-- end manual edit -->
    *
-   * @param formattedName The name of the topic. It must have the format
+   * @param name The name of the topic. It must have the format
    * `"projects/{project}/topics/{topic}"`. `{topic}` must start with a letter,
    * and contain only letters (`[A-Za-z]`), numbers (`[0-9]`), dashes (`-`),
    * underscores (`_`), periods (`.`), tildes (`~`), plus (`+`) or percent
@@ -296,9 +296,9 @@ public class PublisherApi implements AutoCloseable {
    * must not start with `"goog"`.
    * @throws com.google.api.gax.grpc.ApiException if the remote call fails
    */
-  public final Topic createTopic(String formattedName) {
-    TOPIC_PATH_TEMPLATE.validate(formattedName);
-    Topic request = Topic.newBuilder().setName(formattedName).build();
+  public final Topic createTopic(String name) {
+    TOPIC_PATH_TEMPLATE.validate(name);
+    Topic request = Topic.newBuilder().setName(name).build();
     return createTopic(request);
   }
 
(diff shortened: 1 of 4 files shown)


gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -277,27 +277,46 @@ public void close() throws IOException {
...
+   *   String formattedName = PublisherApi.formatTopicName("[PROJECT]", "[TOPIC]");
+   *   Topic response = publisherApi.createTopic(formattedName);
+   * }
+   * </code></pre>
+   *
    * <!-- manual edit -->
    * <!-- end manual edit -->
    *
-   * @param name The name of the topic. It must have the format
+   * @param formattedName The name of the topic. It must have the format
```
aozarov: "It must have the format" does not sound right to me (but I could be wrong). Maybe, "it must be in the xxx format" or "it must be formatted as xxx".

gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -277,27 +277,46 @@ public void close() throws IOException {
...
    * underscores (`_`), periods (`.`), tildes (`~`), plus (`+`) or percent
    * signs (`%`). It must be between 3 and 255 characters in length, and it
    * must not start with `"goog"`.
    * @throws com.google.api.gax.grpc.ApiException if the remote call fails
    */
-  public final Topic createTopic(String name) {
-    Topic request = Topic.newBuilder().setName(name).build();
-
+  public final Topic createTopic(String formattedName) {
+    TOPIC_PATH_TEMPLATE.validate(formattedName);
```
aozarov: If code generation already treat the `formattedName` in a special way and validate it couldn't it replace it with 2 params (e.g. in this case project and topic)?

Also, because the generated message is part of this API I think you should find some consistent naming. I think it is a bit strange to call this one "formattedName" but when I get the *same* value from the Topic it is called "name".  It would have been easier if Topic.name was called Topic.fullyQualfiedName or had separate fields for project and name but mixing term that refer to the same value may be confusing.

michaelbausor: Splitting the formattedName into two params would be possible, but I think the output would be odd in some other cases. For example when accepting two formatted parameters like a subscription and a topic the user would need to pass in the same project in two different parameters.

garrettjonesgoogle: First point: Yes, we would be capable of doing something like splitting into two params. I think we should be careful of that for a couple reasons though:
1. It takes us further away from the raw API surface. We want to track the API surface predictably.
2. In general use, people should be passing around the qualified resource names, and they should only have to be aware of the individual IDs when creating resources. I think it is better to be consistent and use the fully-qualified resource names everywhere.

Second point: I see your point about consistency in naming. Since we really don't have any control over Topic's fields, maybe we shouldn't try to use different naming approaches for our flattened parameters; that also kind of violates the principle I mentioned in the first point, which is that we should strive to adhere more closely to the raw API surface in this layer.

michaelbausor: Regarding the formattedName, the idea was to try to indicate that the parameter needs to be a formatted name rather than just "my-topic" or "my-subscription", which was something we saw come up in an informal usability study. You are right that fullyQualifiedName or similar would be nicer, as it is we need to trade off making our naming informative or consistent.
cc @garrettjonesgoogle

michaelbausor: Removed the formatted prefix from the parameters

gcloud-java-pubsub/baseline/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -289,15 +297,26 @@ public void close() throws IOException {
...
   }
 
   // AUTO-GENERATED DOCUMENTATION AND METHOD - see instructions at the top of the file for editing.
   /**
    * Creates the given topic with the given name.
    *
+   * Sample code:
+   * <pre><code>
+   * try (PublisherApi publisherApi = PublisherApi.createWithDefaults()) {
+   *   String formattedName = PublisherApi.formatTopicName("[PROJECT]", "[TOPIC]");
```
jgeewax: Weird question -- can createWithDefaults pick up the project ID from the environment ? @garrettjonesgoogle

garrettjonesgoogle: It currently does not attempt to do that.

PR #1016, opened 2016-05-19
Add whenDone method and CompletionCallback to Job and Operation

### LINKED ISSUE #963: Provide means to wait for an Operation/Job to complete
To wait for a `Compute` operation or a `BigQuery` job users can now do:
```java
while (!operation.isDone()) {
  Thread.sleap(1000);
}
// User operation
```
We could provide a better way to wait for completion, as, for instance:
```java
operation.whenDone(new OperationCallback() {
  public void onSuccess(Operation operation) {
    // operation succeeded
    // if operation == null => operation no longer exists
  }

  public void onFailure(List<OperationError> errors) {
    // handle operation failure
  }
});
```

mziccard:
This PR adds a friendly method to wait for `Job` and `Operation` to complete, in the current thread. Methods accept a callback that can be used to react to job/operation failure or success. This fixes #963.

/cc @jgeewax @lesv

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/cloud/WaitForOption.java
+++ b/gcloud-java-core/src/main/java/com/google/cloud/WaitForOption.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.base.MoreObjects;
@@ -185,24 +184,8 @@ public abstract class WaitForOption implements Serializable {
     return Objects.equals(option.optionType, option.optionType);
   }
 
-  @Override
-  public boolean equals(Object obj) {
-    if (obj == this) {
-      return true;
-    }
-    if (obj == null || !(obj.getClass().equals(WaitForOption.class))) {
-      return false;
-    }
-    return baseEquals((WaitForOption) obj);
-  }
-
   final int baseHashCode() {
     return Objects.hash(optionType);
-  }
-
-  @Override
-  public int hashCode() {
-    return baseHashCode();
   }
 
   @SuppressWarnings("unchecked")
@@ -215,7 +198,7 @@ public abstract class WaitForOption implements Serializable {
         foundOption = (T) option;
       }
     }
-    return firstNonNull(foundOption, defaultValue);
+    return foundOption != null ? foundOption : defaultValue;
   }
 
   /**


gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -144,14 +163,51 @@ public boolean isDone() {
...
+   *   void error(BigQueryError error, List<BigQueryError> executionErrors) {
+   *     // handle error
+   *   }
+   * });}</pre>
+   *
+   * @throws BigQueryException upon failure
+   * @throws InterruptedException if the current thread gets interrupted while waiting for the job
+   *     to complete
+   */
+  public void whenDone(CompletionCallback callback) throws InterruptedException {
```
aozarov: I am not too excited about this method even though it does reduce some boilerplate code.
I think it would have been much more useful if it was not blocking and we could return a "listenable future".

My issues with it are:

1) It is not clear that this method is blocking.Though Java doc says "Waits until..." I would be more explicit saying "this method will block until..." and also change the name of the method to something like `waitFor`.

2)  Existing the method without a callback *and* without indicating that a callback was not called is error prone (e.g. imagine callbacks were used to toggle some latch/semaphore). I think we should either make the method  return a boolean or add another explicit callback for not found.

3) There is actually another exit point in this case and that could be that the RPC for `isDone` or `reload` failed, which similar to (2) makes the callback usage more complex. A solution for that would be to never fail 
calling this method and to indicate any failures (in a distinct way) in the callback.

4) `500L` sampling rate may be too course or to grain depending on the job and the client needs. Maybe we could provide a var-arg way to tune it.

mziccard: I don't like this method either and I don't like the idea of writing `Thread.sleep` somewhere in our code. But if we see this as a way to avoid the user to write boilerplate code then it might make sense.

> I think it would have been much more useful if it was not blocking and we could return a "listenable future".

Yes but as long as we stick to Java 7 this is not doable. Also the whole BigQuery/Compute API is synchronous, such a change would make sense when providing all other methods in an async flavor. 

> 1) It is not clear that this method is blocking.Though Java doc says "Waits until..." I would be more explicit saying "this method will block until..." and also change the name of the method to something like waitFor.

Ok to me

> 2) Existing the method without a callback and without indicating that a callback was not called is error prone

Non-existing operations/jobs are a very unlikely case. I believe that having a dedicated method in the callback is not worth it. Maybe the return value is the less invasive way of handling this case, but it's hardly intuitive.

> 3) There is actually another exit point in this case and that could be that the RPC for isDone or reload failed, which similar to (2) makes the callback usage more complex. A solution for that would be to never fail 
calling this method and to indicate any failures (in a distinct way) in the callback.

I don't feel too strange that the method throws isDone or reload failed. The method is blocking and the callback is related to the operation status, not the method's error condition, so exceptions make sense to me.

> 4) 500L sampling rate may be too course or to grain depending on the job and the client needs. Maybe we could provide a var-arg way to tune it.

Vararg options to tune this sound good to me even though I don't expect much users to be interested in tuning this. BTW 500 if I recall correctly is the value adopted by node guys.

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -144,14 +163,51 @@ public boolean isDone() {
...
 
   /**
+   * Waits until this job completes its execution, either failing or succeeding. If the job does not
+   * exist, this method returns without executing any method of the provided callback. If the job
+   * completed successfully the {@link CompletionCallback#success(Job)} method is called. If the job
+   * completed with errors the {@link CompletionCallback#error(BigQueryError, List)} method is
+   * called.
+   * <pre> {@code
+   * job.whenDone(new CompletionCallback() {
+   *   void success(Job job) {
```
aozarov: you miss `public` and if you want `@Override`.

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
aozarov: This method is blocking. So what should people expect once it completed... They should expect that job is completed (failed or successful). So whatever they do in the callback they can do *after* the blocking call.

If you provide a callback I think it is much more expected that all would be handled in the callback. If that is not the case then I think the following application code would be simpler:

```java
try {
  Job completedJob = job.waitFor();
  if (completedJob == null) {
     // job was removed (if you care, otherwise just check for completedJob != null
  } else if (completedJob.status().error() != null) {
    // job failed,  do something AND also have a reference to the job as you may need it
    // e.g. to log job X failed.
  } else {
    // job sucessful
  }
} catch (ComputeException ex) {
...
}
```

As oppose to:
```java
try {
  job.whenDone(new CompletionCallback() {
    @Override
     public void success(Job job) {
        // if the code after the blocking `whenDone` needs an updated reference
        // I will need to play some trickery to get this reference visible after
     }

     @Override
      public void error(BigQueryError error, List<BigQueryError> executionErrors) {
        // what if I need a reference to the Job?
      }
  });
  
   // now I can also deal with the job and I will only get here once the job was completed
   // however I don't have a reference for the updated job nor do I know if actually the job
   // was removed and non of my callbacks were called. So, if I care about it I have to do:
  job = job.reload();
  if (job = null) {
   ...
  } else {
    ...
  }    
} catch (ComputeException ex) {
...
}
```

mziccard: I don't agree.

The `job == null` case is so unlikely that no user will care about it in reality. For it to happen you must be checking the state of a months old job at the time the service does the job history cleaning, or you have somebody else deleting an operation (which can be done only after it completed) before you manage to read its state.

For what concerns exceptions I believe they should stay out of the callback. When you go async you are forced to handle "all" in the callback but here we can avoid that, the callback handles changes in the job/operation state while service exceptions can be handled the usual way.

If you prefer `Job completedJob = job.waitFor();` we can go for it even though i think that this kind of `waitFor` methods is ugly in general.

Maybe @ajkannan and @jgeewax can give us more opinions.

ajkannan: My opinion is that it's better to leave out the callback.  As a user, my gut reaction to seeing this method would be that either the success or error method of the callback would be called.  Also, the callback doesn't reduce the amount of boilerplate code too much, and as Arie mentions, getting references to updated job is more cumbersome.

gcloud-java-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -69,9 +69,9 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
+import org.easymock.EasyMock;
```
mziccard: Codacy Issue found: [Avoid unused imports such as 'org.easymock.EasyMock'](https://www.codacy.com/app/mziccard/gcloud-java/file/2615266202/issues/source?bid=3123103&fileBranchId=3307743#l72)

mziccard:
In c3ce162 I removed `whenDone` as well as the callback class. Instead, I added two blocking methods (`waitFor` and `waitFor(int, TimeUnit)`) that return the completed job/operation (on `null` if not found).

You should like this better no :)

README.md
```diff
@@ -196,21 +194,18 @@ import com.google.cloud.compute.Compute;
 import com.google.cloud.compute.ComputeOptions;
 import com.google.cloud.compute.Disk;
 import com.google.cloud.compute.DiskId;
-import com.google.cloud.compute.Operation;
 import com.google.cloud.compute.Snapshot;
 
-Compute compute = ComputeOptions.defaultInstance().service();
+final Compute compute = ComputeOptions.defaultInstance().service();
```
aozarov: final (here and bellow) are no longer needed, right?

mziccard: You are right

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -144,14 +145,62 @@ public boolean isDone() {
...
+   *   // job failed, handle error
+   * } else {
+   *   // job completed successfully
+   * }}</pre>
+   *
+   * @throws BigQueryException upon failure
+   * @throws InterruptedException if the current thread gets interrupted while waiting for the job
+   *     to complete
+   */
+  public Job waitFor(int checkEvery, TimeUnit unit) throws InterruptedException {
```
aozarov: `@param`? Also, did you consider using var-args option for it (e.g. if at some point you will want to add also an option to timeout)?

aozarov:
Yes, I like it. One more suggestion and javadoc needs some cleanup.

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
mziccard: I considered it, I preferred overloading because:
- less duplicated code (the vararg option would be very similar both in `BigQuery` and `Compute` and I wasn't sure whether moving it to `gcloud-java-core` given the small use such a class would have)
- I thought about a timeout but it seemed not very useful in this case. And I couldn't think of other options

Vararg option sound good to me so if you have a preference I am happy to give it a try.

aozarov: I like he var-args option better as it uses the same method signature and can support more options (I actually do think that one may want to say "wait until done or t=x", but with var-args that could always be added later based on demand). However, if in doubt I would only leave the `waitFor()` and add the second option later once we get more feedback.

lesv:
ping

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
mziccard: The only thing left that does not convince me is the following. If we have a var-arg options parameter (`WaitForOption...`) and add support for a timeout option (`WaitForOption.timeout()`) the signature of the method becomes:
```java
waitFor(WaitForOption... options) throws InterruptedException, TimeoutException;
```
So users will have to always catch the `TimeoutException` even though that can only be thrown when `WaitForOption.timeout()` is passed to the method. Isn't this a bit annoying?

aozarov: I don't think this is an issue considering that `waitFor()` already throws a checked exception (`InterruptedException`).

gcloud-java-compute/src/test/java/com/google/cloud/compute/OperationTest.java
```diff
@@ -353,6 +365,109 @@ public void testIsDone_NotExists() throws Exception {
...
+    initializeExpectedOperation(4);
+    Compute.OperationOption[] expectedOptions =
+        {Compute.OperationOption.fields(Compute.OperationField.STATUS)};
+    TimeUnit timeUnit = createStrictMock(TimeUnit.class);
+    timeUnit.sleep(1);
+    EasyMock.expectLastCall();
+    Clock clock = createStrictMock(Clock.class);
+    EasyMock.expect(clock.millis()).andReturn(0L);
+    EasyMock.expect(clock.millis()).andReturn(1L);
+    EasyMock.expect(clock.millis()).andReturn(3L);
```
mziccard: Codacy Issue found: [Unnecessary use of fully qualified name 'EasyMock.expect' due to existing static import 'org.easymock.EasyMock.expect'](https://www.codacy.com/app/mziccard/gcloud-java/file/2653889166/issues/source?bid=3123103&fileBranchId=3307743#l455)

gcloud-java-compute/src/test/java/com/google/cloud/compute/OperationTest.java
```diff
@@ -353,6 +365,109 @@ public void testIsDone_NotExists() throws Exception {
...
+  @Test
+  public void testWaitForWithTimeout() throws InterruptedException, TimeoutException {
+    initializeExpectedOperation(4);
+    Compute.OperationOption[] expectedOptions =
+        {Compute.OperationOption.fields(Compute.OperationField.STATUS)};
+    TimeUnit timeUnit = createStrictMock(TimeUnit.class);
+    timeUnit.sleep(1);
+    EasyMock.expectLastCall();
+    Clock clock = createStrictMock(Clock.class);
+    EasyMock.expect(clock.millis()).andReturn(0L);
```
mziccard: Codacy Issue found: [Unnecessary use of fully qualified name 'EasyMock.expect' due to existing static import 'org.easymock.EasyMock.expect'](https://www.codacy.com/app/mziccard/gcloud-java/file/2653889166/issues/source?bid=3123103&fileBranchId=3307743#l453)

gcloud-java-compute/src/test/java/com/google/cloud/compute/OperationTest.java
```diff
@@ -353,6 +365,109 @@ public void testIsDone_NotExists() throws Exception {
...
+    EasyMock.expect(clock.millis()).andReturn(1L);
```
mziccard: Codacy Issue found: [Unnecessary use of fully qualified name 'EasyMock.expect' due to existing static import 'org.easymock.EasyMock.expect'](https://www.codacy.com/app/mziccard/gcloud-java/file/2653889166/issues/source?bid=3123103&fileBranchId=3307743#l454)

gcloud-java-bigquery/src/test/java/com/google/cloud/bigquery/JobTest.java
```diff
@@ -178,12 +189,112 @@ public void testIsDone_NotExists() throws Exception {
...
+  @Test
+  public void testWaitForWithTimeout() throws InterruptedException, TimeoutException {
+    initializeExpectedJob(2);
+    BigQuery.JobOption[] expectedOptions = {BigQuery.JobOption.fields(BigQuery.JobField.STATUS)};
+    TimeUnit timeUnit = createStrictMock(TimeUnit.class);
+    timeUnit.sleep(1);
+    EasyMock.expectLastCall();
+    Clock clock = createStrictMock(Clock.class);
+    EasyMock.expect(clock.millis()).andReturn(0L);
+    EasyMock.expect(clock.millis()).andReturn(1L);
```
mziccard: Codacy Issue found: [Unnecessary use of fully qualified name 'EasyMock.expect' due to existing static import 'org.easymock.EasyMock.expect'](https://www.codacy.com/app/mziccard/gcloud-java/file/2653889044/issues/source?bid=3123103&fileBranchId=3307743#l273)

gcloud-java-bigquery/src/test/java/com/google/cloud/bigquery/JobTest.java
```diff
@@ -178,12 +189,112 @@ public void testIsDone_NotExists() throws Exception {
...
+
+  @Test
+  public void testWaitForWithTimeout() throws InterruptedException, TimeoutException {
+    initializeExpectedJob(2);
+    BigQuery.JobOption[] expectedOptions = {BigQuery.JobOption.fields(BigQuery.JobField.STATUS)};
+    TimeUnit timeUnit = createStrictMock(TimeUnit.class);
+    timeUnit.sleep(1);
+    EasyMock.expectLastCall();
+    Clock clock = createStrictMock(Clock.class);
+    EasyMock.expect(clock.millis()).andReturn(0L);
```
mziccard: Codacy Issue found: [Unnecessary use of fully qualified name 'EasyMock.expect' due to existing static import 'org.easymock.EasyMock.expect'](https://www.codacy.com/app/mziccard/gcloud-java/file/2653889044/issues/source?bid=3123103&fileBranchId=3307743#l272)

gcloud-java-bigquery/src/test/java/com/google/cloud/bigquery/JobTest.java
```diff
@@ -178,12 +189,112 @@ public void testIsDone_NotExists() throws Exception {
...
+    EasyMock.expect(clock.millis()).andReturn(1L);
+    EasyMock.expect(clock.millis()).andReturn(3L);
```
mziccard: Codacy Issue found: [Unnecessary use of fully qualified name 'EasyMock.expect' due to existing static import 'org.easymock.EasyMock.expect'](https://www.codacy.com/app/mziccard/gcloud-java/file/2653889044/issues/source?bid=3123103&fileBranchId=3307743#l274)

mziccard:
In e5af988 I added vararg options for both the checking period and the timeout. How do you like it now?

I had to rebase to fix conflicts and trigger CI.

gcloud-java-compute/src/main/java/com/google/cloud/compute/Operation.java
```diff
@@ -652,21 +660,77 @@ public boolean exists() throws ComputeException {
...
+   * @throws ComputeException upon failure
+   * @throws InterruptedException if the current thread gets interrupted while waiting for the
+   *     operation to complete
+   * @throws TimeoutException if the timeout provided with
+   *     {@link WaitForOption#timeout(long, TimeUnit)} is exceeded. If no such option is provided
+   *     this exception is never thrown.
+   */
+  public Operation waitFor(WaitForOption... waitOptions)
+      throws InterruptedException, TimeoutException {
+    Map<WaitForOption.Option, ?> optionMap = WaitForOption.asMap(waitOptions);
```
aozarov: I like the change! and fine as is.

One suggestion though.

Because you made `WaitForOption` public (for making it common) you also exposed many elements that otherwise would be package scope.
I suggest to minimize those as much as possible. One way that may look less "internal" would be to eliminate WaitForOption.Option and its related methods and instead make `CheckingPeriod` and Timeout (a new inner class) sub-classes of `WaitForOption`. These 2 classes can have a public static `getOrDefault(WaitForOption... options)` which can use an internal/private enum to avoid "instanceof" and have specific methods to get their values.

mziccard: I agree with the suggested change. The rationale behind my choice was to have a similar implementation for all var-arg options, but I agree with your point of not exposing "internal" methods (even though thet should pretty harmless in this case). Will work on this soon, thanks Arie.

mziccard: Fixed according to your suggestion in 71cd868

aozarov: Looks good. Few minor comments.

aozarov:
I like the change now! I will leave it to @lesv to do a thorough review but looks good to me.

lesv:
LGTM

PR #1022, opened 2016-05-23
Implement modifyAckDeadline methods, add javadoc and tests

mziccard:
System tests against emulator will come as soon as also `pull` methods are implemented.

/cc @jgeewax @lesv


jgeewax:
LGTM :+1:

PR #1027, opened 2016-05-26
Implement ack and nack methods, add javadoc and tests

mziccard:
This PR implements `ack` and `nack` methods, adds javadoc and unit tests. System tests will come as soon as `pull` methods are implemented.

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
@@ -387,25 +387,107 @@ public interface PubSub extends AutoCloseable, Service<PubSubOptions> {
 
   MessageConsumer pullAsync(String subscription, MessageProcessor callback, PullOption... options);
 
+  /**
+   * Acknowledges the given messages for the provided subscription. Ack ids identify the messages to
+   * acknowledge, as returned in {@link ReceivedMessage#ackId()} by {@link #pull(String, int)} and
+   * {@link #pullAsync(String, int)}.
+   *
+   * @param subscription the subscription whose messages must be acknowledged
+   * @param ackId the ack id of the first message to acknowledge
+   * @param ackIds other ack ids of messages to acknowledge
+   * @throws PubSubException upon failure, or if the subscription was not found
+   */
   void ack(String subscription, String ackId, String... ackIds);
 
+  /**
+   * Sends a request to acknowledge the given messages for the provided subscription. Ack ids
+   * identify the messages to acknowledge, as returned in {@link ReceivedMessage#ackId()} by
+   * {@link #pull(String, int)} and {@link #pullAsync(String, int)}. The method returns a
+   * {@code Future} object that can be used to wait for the acknowledge operation to be completed.
+   *
+   * @param subscription the subscription whose messages must be acknowledged
+   * @param ackId the ack id of the first message to acknowledge
+   * @param ackIds other ack ids of messages to acknowledge
+   */
   Future<Void> ackAsync(String subscription, String ackId, String... ackIds);
 
+  /**
+   * Acknowledges the given messages for the provided subscription. Ack ids identify the messages to
+   * acknowledge, as returned in {@link ReceivedMessage#ackId()} by {@link #pull(String, int)} and
+   * {@link #pullAsync(String, int)}.
+   *
+   * @param subscription the subscription whose messages must be acknowledged
+   * @param ackIds the ack ids of messages to acknowledge
+   * @throws PubSubException upon failure, or if the subscription was not found
+   */
   void ack(String subscription, Iterable<String> ackIds);
 
+  /**
+   * Sends a request to acknowledge the given messages for the provided subscription. Ack ids
+   * identify the messages to acknowledge, as returned in {@link ReceivedMessage#ackId()} by
+   * {@link #pull(String, int)} and {@link #pullAsync(String, int)}. The method returns a
+   * {@code Future} object that can be used to wait for the acknowledge operation to be completed.
+   *
+   * @param subscription the subscription whose messages must be acknowledged
+   * @param ackIds the ack ids of messages to acknowledge
+   */
   Future<Void> ackAsync(String subscription, Iterable<String> ackIds);
 
+  /**
+   * "Nacks" the given messages for the provided subscription. Ack ids identify the messages to
+   * "nack", as returned in {@link ReceivedMessage#ackId()} by {@link #pull(String, int)} and
+   * {@link #pullAsync(String, int)}. This method corresponds to calling
+   * {@link #modifyAckDeadline(String, int, TimeUnit, String, String...)} with a deadline of 0.
+   *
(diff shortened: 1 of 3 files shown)


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
```diff
@@ -387,25 +387,107 @@ public static PullOption maxConcurrentCallbacks(int maxConcurrency) {
 
   MessageConsumer pullAsync(String subscription, MessageProcessor callback, PullOption... options);
 
+  /**
+   * Acknowledges the given messages for the provided subscription. Ack ids identify the messages to
```
lesv: Ack ids <== shouldn't this mention one of the param's?

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
```diff
@@ -387,25 +387,107 @@ public static PullOption maxConcurrentCallbacks(int maxConcurrency) {
...
+   *
+   * @param subscription the subscription whose messages must be acknowledged
+   * @param ackId the ack id of the first message to acknowledge
+   * @param ackIds other ack ids of messages to acknowledge
+   * @throws PubSubException upon failure, or if the subscription was not found
+   */
   void ack(String subscription, String ackId, String... ackIds);
 
+  /**
+   * Sends a request to acknowledge the given messages for the provided subscription. Ack ids
```
lesv: Ack ids <== same ?

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
lesv: I guess I'm ok w/ this.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
lesv: I guess I'm ok w/ this.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
```diff
@@ -387,25 +387,107 @@ public static PullOption maxConcurrentCallbacks(int maxConcurrency) {
 
   MessageConsumer pullAsync(String subscription, MessageProcessor callback, PullOption... options);
 
+  /**
+   * Acknowledges the given messages for the provided subscription. Ack ids identify the messages to
+   * acknowledge, as returned in {@link ReceivedMessage#ackId()} by {@link #pull(String, int)} and
+   * {@link #pullAsync(String, int)}.
+   *
+   * @param subscription the subscription whose messages must be acknowledged
```
lesv: I'm new to this (for this repo), so take these with a grain of salt, but looking at other javadoc, this seems overly wordy. @jgeewax @aozarov (please agree or disagree -- I see earlier examples that contradict my comments, many other published javadocs are as I describe.)

The initial "sentence" [should be a fragment](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment).
`Acknowledges given messages for provided subscription.` ...
```
@param subscription whose messages are being acknowledged
@param ackId first message to acknowledge
```

lesv:
Once I understand (or get feedback from JJ or Arie)  the javadoc story for the repo, I expect to give quite a bit of feedback.

The code and the testing look ok, but I want to spend a few minutes at an earlier time of day re-reviewing to make sure that I believe your tests actually cover things and are helpful.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
mziccard: I believe we are quite close to what Guava does. See [Lists.asList](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Lists.java#L312) or other methods.

aozarov:
I think so far we were careful to follow the correct Javadoc guidlines for structure (spaces, tag ordering) and rules (punctuation, case,..). As for the descriptions, I think the focus was on clarity which in some
cases made it less concise. You can find in both the JDK and Guava examples to match either style.
I don't have strong opinion about it, however if you think clarity is not effected or even improved by making it shorter than I think you should go for it.

BTW, Looking at the generated [Javadoc](http://googlecloudplatform.github.io/gcloud-java/0.2.2/apidocs/index.html) it looks like `@param` descriptions are not enabled.

mziccard:
> BTW, Looking at the generated Javadoc it looks like @param descriptions are not enabled.

Arie, what do you mean with this?

aozarov:
> Arie, what do you mean with this?

I was wrong. I sampled few places (just to see how things look like per earlier discussion) and unfortunately each of these did not specify `@param`...
E.g. [this](http://googlecloudplatform.github.io/gcloud-java/0.2.1/apidocs/com/google/cloud/dns/Dns.html#applyChangeRequest(java.lang.String,%20com.google.cloud.dns.ChangeRequestInfo,%20com.google.cloud.dns.Dns.ChangeRequestOption...)) and [this](http://googlecloudplatform.github.io/gcloud-java/0.2.1/apidocs/com/google/cloud/compute/Address.html#delete(com.google.cloud.compute.Compute.OperationOption...)) and [this](http://googlecloudplatform.github.io/gcloud-java/0.2.1/apidocs/com/google/cloud/compute/Compute.html#addAccessConfig(com.google.cloud.compute.InstanceId,%20java.lang.String,%20com.google.cloud.compute.NetworkInterface.AccessConfig,%20com.google.cloud.compute.Compute.OperationOption...)) and more...

lesv:
Our internal guidelines specify that it should be a fragment, not a sentence and that it's ok to not use `@param` if you mention the value in the paragraph.  It also specifies fewer words (ie. fragments) are better as long as it's clear.

So, I'm going to suggest a few re-wordings to be simpler - we won't go through and change existing stuff, but I think being concise is better for our users.

mziccard:
Not to be picky but let me clarify. Fragments are sequences of words that look like a sentence but are not. The difference from a sentence is that fragments are allowed to contain no **independent clauses**. An independent clause is a sequence of words that contains a subject and a verb and makes sense by itself (extrapolated from the context).

`Acknowledges the given messages for the provided subscription.` is already a fragment (it's missing the subject). @lesv suggestion (`Acknowledges given messages for provided subscription.`) simply removes the articles but does not affect its fragment nature.

For what concerns `@param` descriptions, according to [Oracle's guide](http://www.oracle.com/technetwork/articles/java/index-137868.html), they should start with a noun (hence the suggested `@param subscription whose messages are being acknowledged` is wrong), possibly preceded by an article.

I don't mind applying changes (even project-wide) if that improve docs readability I just doubt that removing some articles or conjunctions would help users.

lesv:
Yep - I agree - It's what I've been pondering.   I do dislike several mentions of the same thing for the future.  LGTM

mziccard:
Thanks for the pass @lesv

PR #1030, opened 2016-05-27
Add OAuth2AuthCredentials class and tests

### LINKED ISSUE #1028: We should add tests for AuthCredentials
Not all `AuthCredentials` subclasses are platform-dependent. In our unit tests we should cover those credentials whose creation is platform-independent.

### LINKED ISSUE #1029: Provide means to create AuthCredentials given an OAuth2 access token

mziccard:
This PR adds the `OAuth2AuthCredentials` class to fix #1029. Unit tests are added to fix #1028 

/cc @lesv

### DIFF
--- a/gcloud-java-core/src/test/java/com/google/cloud/AuthCredentialsTest.java
+++ b/gcloud-java-core/src/test/java/com/google/cloud/AuthCredentialsTest.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud;
+
+import static com.google.common.base.Charsets.UTF_8;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+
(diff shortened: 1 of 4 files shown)


gcloud-java-core/src/test/java/com/google/cloud/AuthCredentialsTest.java
```diff
@@ -0,0 +1,128 @@
...
+import java.security.Signature;
+import java.security.SignatureException;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Date;
+
+public class AuthCredentialsTest {
+
+  private static final String ACCESS_TOKEN = "accessToken";
+  private static final Date EXPIRATION_DATE = new Date();
```
lesv: Should this be a tiny bit in the future?

gcloud-java-core/src/test/java/com/google/cloud/SerializationTest.java
```diff
@@ -94,7 +95,8 @@ public Builder toBuilder() {
   protected Restorable<?>[] restorableObjects() {
     try {
       return new Restorable<?>[]{AuthCredentials.createForAppEngine(), AuthCredentials.noAuth(),
-          AuthCredentials.createForJson(new ByteArrayInputStream(JSON_KEY.getBytes()))};
+          AuthCredentials.createForJson(new ByteArrayInputStream(JSON_KEY.getBytes())),
+          AuthCredentials.createFor("accessToken", new Date())};
```
lesv: Should the date be slightly in the future?

lesv:
LGTM - You can ignore my ask about Date().  It would only be appropriate if somewhere you tested to see if a token / credential was still valid.  In looking through more of the code, I see that refresh is specifically disabled for some token kinds.  Somewhere there should be token expiration validation / refresh that happens, but not in any of your tests, nor the paths I traced through.

mziccard:
@lesv Thanks for the pass.

Refresh is handled by `google-auth-library-java`'s classes. It happens when [`getRequestMetadata(Uri)`](https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L70) is called on a credentials object with expired token.

lesv:
Do you prefer that I merge your code on LGTM, or will you do it?  Thanks.

mziccard:
I will take care of merging, thanks.

PR #1024, opened 2016-05-25
Add AckDeadlineRenewer and AckDeadlineRenewerImpl classes

mziccard:
This PR adds support for automatic ack deadline renewal. This adds:
- `AckDeadlineRenewer` interface for auto ack deadline renewers
- `AckDeadlineRenewerImpl` baseline implementation of an automatic ack deadline renewer
- `AckDeadlineRenewParams` option class to configure how ack deadlines should be renewed (e.g. backoff)

@aozarov any comment would be really valuable here.

PR #1031, opened 2016-05-27
Add AckDeadlineRenewer class for automatic ack deadline renewal

mziccard:
This is a simpler version of #1024 to provide support for automatic ack deadline renewals

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/AckDeadlineRenewer.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/AckDeadlineRenewer.java
@@ -41,7 +41,8 @@ import java.util.concurrent.TimeUnit;
  */
 class AckDeadlineRenewer implements AutoCloseable {
 
-  private static final int MIN_DEADLINE_MILLIS = 9_000;
+  private static final int MIN_DEADLINE_MILLIS = 10_000;
+  private static final int DEADLINE_SLACK_MILLIS = 1_000;
   private static final int RENEW_THRESHOLD_MILLIS = 3_000;
   private static final int NEXT_RENEWAL_THRESHOLD_MILLIS = 1_000;
 
@@ -213,8 +214,8 @@ class AckDeadlineRenewer implements AutoCloseable {
     }
     for (Map.Entry<String, List<String>> entry : Multimaps.asMap(messagesToRenewNext).entrySet()) {
       // We send all ack deadline renewals for a subscription
-      pubsub.modifyAckDeadlineAsync(entry.getKey(), MIN_DEADLINE_MILLIS,
-          TimeUnit.MILLISECONDS, entry.getValue());
+      pubsub.modifyAckDeadlineAsync(entry.getKey(), MIN_DEADLINE_MILLIS, TimeUnit.MILLISECONDS,
+          entry.getValue());
     }
     unsetAndScheduleNextRenewal();
   }
@@ -253,7 +254,7 @@ class AckDeadlineRenewer implements AutoCloseable {
    */
   void add(String subscription, String ackId) {
     synchronized (lock) {
-      long deadline = clock.millis() + MIN_DEADLINE_MILLIS;
+      long deadline = clock.millis() + MIN_DEADLINE_MILLIS - DEADLINE_SLACK_MILLIS;
       Message message = new Message(new MessageId(subscription, ackId), deadline);
       messageQueue.add(message);
       messageDeadlines.put(message.messageId(), deadline);
@@ -272,7 +273,7 @@ class AckDeadlineRenewer implements AutoCloseable {
    */
   void add(String subscription, Iterable<String> ackIds) {
     synchronized (lock) {
-      long deadline = clock.millis() + MIN_DEADLINE_MILLIS;
+      long deadline = clock.millis() + MIN_DEADLINE_MILLIS - DEADLINE_SLACK_MILLIS;
       for (String ackId : ackIds) {
         Message message = new Message(new MessageId(subscription, ackId), deadline);
         messageQueue.add(message);
--- a/gcloud-java-pubsub/src/test/java/com/google/cloud/pubsub/AckDeadlineRenewerTest.java
+++ b/gcloud-java-pubsub/src/test/java/com/google/cloud/pubsub/AckDeadlineRenewerTest.java
@@ -36,7 +36,7 @@ import java.util.concurrent.atomic.AtomicLong;
 
 public class AckDeadlineRenewerTest {
 
-  private static final int MIN_DEADLINE_MILLIS = 9_000;
+  private static final int MIN_DEADLINE_MILLIS = 10_000;
 
   private static final String SUBSCRIPTION1 = "subscription1";
   private static final String SUBSCRIPTION2 = "subscription2";


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
```diff
@@ -86,6 +86,11 @@ protected PubSubOptions(Builder builder) {
   }
 
   @Override
+  protected ExecutorFactory executorFactory() {
```
mziccard: Codacy Issue found: [Overriding method merely calls super](https://www.codacy.com/app/mziccard/gcloud-java/file/2647881484/issues/source?bid=3123109&fileBranchId=3323306#l89)

mziccard: Codacy Issue found: [Overriding method merely calls super](https://www.codacy.com/app/mziccard/gcloud-java/file/2648334987/issues/source?bid=3123109&fileBranchId=3323306#l89)

mziccard: Codacy Issue found: [Overriding method merely calls super](https://www.codacy.com/app/mziccard/gcloud-java/file/2648419172/issues/source?bid=3123109&fileBranchId=3323306#l89)

mziccard: Codacy Issue found: [Overriding method merely calls super](https://www.codacy.com/app/mziccard/gcloud-java/file/2648607896/issues/source?bid=3123109&fileBranchId=3323306#l89)

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/AckDeadlineRenewer.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Class for an automatic ack deadline renewer. An ack deadline renewer automatically renews the
+ * acknowledge deadline of messages added to it (via {@link #add(String, String)} or
+ * {@link #add(String, Iterable)}. The acknowledge deadlines of added messages are renewed until the
+ * messages are explicitly removed using {@link #remove(String, String)}.
+ */
+class AckDeadlineRenewer implements AutoCloseable {
```
aozarov: Overall implementation looks fine.

I think it can be improved (e.g. wake up as needed and not based on a constant internal, auto-remove items that are known to have expired, handle deadline renewal failures, minimize possible RPC by including more ackIds for a subscription if we are already going to make an RPC for that subscription,...) however, starting with simple and improving, if needed, later sounds like a right move.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/AckDeadlineRenewer.java
```diff
@@ -0,0 +1,268 @@
...
+
+  private static final int MIN_DEADLINE_MILLISECONDS = 10_000;
```
aozarov: maybe for defaults we should have a bit more slack in both the min and the threshold. I would consult the PubSub team for that.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/AckDeadlineRenewer.java
```diff
@@ -0,0 +1,268 @@
...
+   */
+  void remove(String subscription, String ackId) {
+    synchronized (lock) {
+      messageDeadlines.remove(new MessageId(subscription, ackId));
+    }
+  }
+
+  @Override
+  public void close() throws Exception {
+    renewerFuture.cancel(false);
```
aozarov: this could have clear messageQueue and/or messageDeadlines.

gcloud-java-pubsub/src/test/java/com/google/cloud/pubsub/AckDeadlineRenewerTest.java
```diff
@@ -0,0 +1,234 @@
...
+  private Clock clock;
+  private PubSub pubsub;
+  private ScheduledExecutorService executor;
+  private ExecutorFactory executorFactory;
+  private AckDeadlineRenewer ackDeadlineRenewer;
+  private ScheduledFuture scheduledFuture;
+
+  @Before
+  public void setUp() {
+    clock = EasyMock.createStrictMock(Clock.class);
```
aozarov: This test seems to me too fragile/internal. I am not sure why we could not use non mocked executor and use either Clock or enable setting the deadline (package scope, testing only), so only mock and verification would be PubSub.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/AckDeadlineRenewer.java
mziccard: Agree, I was also planning to change to `cancel(true)` and add a `closed` field to avoid doing the cleanup twice.

gcloud-java-pubsub/src/test/java/com/google/cloud/pubsub/AckDeadlineRenewerTest.java
mziccard: I agree that this test is very internal but I don't see this as necessarily bad. How would using actual executors help? If we mock clock and use actual executors we will end up with very similar tests plus the variability given by the fact that we are actually using threads. I am not sure it's a good idea in general to use executors for unit testing.

aozarov: The more you mock the more susceptible you are to implementation details changes and in some cases you may over-mock and ignore potential issues.

I think it is reasonable to have some tests with timeouts and  you can use the `PusSub` last expected call as away indicate completion  (e.g. look at the first answer [here](http://stackoverflow.com/questions/16651578/using-easymock-in-multithreading)).

Just something to consider...

mziccard: > Just something to consider...

I'll give this a try tomorrow and see if I can reduce mocking while still providing adequate coverage. My fear is that if with using actual executors we will end up testing the same behavior as before but also add the risk of tests being flaky. But again, let me give this a try :)

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
mziccard: Codacy Issue found: [Overriding method merely calls super](https://www.codacy.com/app/mziccard/gcloud-java/file/2670963122/issues/source?bid=3123109&fileBranchId=3323306#l89)

mziccard:
@aozarov In d4b8f6c I refactored the tests to follow your suggestion. I also addressed other comments. I removed the fixed interval and now renewer is only scheduled when actually needed. PTAL

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
mziccard: Codacy Issue found: [Overriding method merely calls super](https://www.codacy.com/app/mziccard/gcloud-java/file/2673950796/issues/source?bid=3123109&fileBranchId=3323306#l89)

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/AckDeadlineRenewer.java
```diff
@@ -0,0 +1,316 @@
...
+    while ((message = nextMessageToRenew(threshold)) != null) {
+      // If the expected deadline is null the message was removed and we must ignore it, otherwise
+      // we renew its ack deadline
+      if (message.expectedDeadline() != null) {
+        messagesToRenewNext.put(message.messageId().subscription(), message.messageId().ackId());
+      }
+    }
+    for (Map.Entry<String, List<String>> entry : Multimaps.asMap(messagesToRenewNext).entrySet()) {
+      // We send all ack deadline renewals for a subscription
+      pubsub.modifyAckDeadlineAsync(entry.getKey(), MIN_DEADLINE_MILLIS,
```
aozarov: I actually think that we should ask to lease a bit *longer* than the 10 seconds minimum however assume that we only got 10 seconds and use slack before we renew again.

aozarov:
Change looks good. Implementation looks good and api looks good (I provided one more minor suggestion). Maybe @lesv want to go over this code as well (I didn't pay attention to anything but API and impl).

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
mziccard: Codacy Issue found: [Overriding method merely calls super](https://www.codacy.com/app/mziccard/gcloud-java/file/2681421883/issues/source?bid=3123109&fileBranchId=3323306#l89)

mziccard:
I am merging this, if @lesv has some more comments I'd be happy to address them in separate PRs.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
lesv: So why have it?

lesv:
I'm happy with this change.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
mziccard: So that the method is visible in the package.

PR #1033, opened 2016-05-31
Extend StorageExample to show how to add ACLs to blobs and buckets

### LINKED ISSUE #1012: Storage example should show how to set ACL
ACLs are not even mentioned in the examples. Example needs to show both how to set a specific ACL and how to set a PredefinedAcl.

When I use this this code, the console Cloud Storage viewer does not show items as shared publicly:

	storage.create(coverBlob, coverImage,
		BlobTargetOption.predefinedAcl(PredefinedAcl.PUBLIC_READ));

So it's not clear what this code is supposed to do.

Am I right to assume that ACL should be set on BlobInfo like this?

        .acl(Arrays.asList(Acl.of(User.ofAllUsers(), Role.READER)))

If this is correct, is it enough, or should be used together with `BlobTargetOption.predefinedAcl(PredefinedAcl.PUBLIC_READ)`?

mziccard:
This PR adds examples to `StorageExample` that show how to add different types of ACLs to existing blobs/buckets. This fixes #1012.

### DIFF
--- a/gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
+++ b/gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
@@ -178,6 +178,12 @@ public class StorageExample {
    * @see <a href="https://cloud.google.com/storage/docs/json_api/v1/objects/get">Objects: get</a>
    */
   private static class InfoAction extends BlobsAction {
+
+    /**
+     * Gets information for the provided blobs, using the {@code storage} service. If
+     * {@code blobIds} contains only one blob identity and {@code blobIds[0].name()} is empty, this
+     * method gets information for the bucket identified by {@code blobIds[0].bucket()}.
+     */
     @Override
     public void run(Storage storage, BlobId... blobIds) {
       if (blobIds.length == 1) {
@@ -565,6 +571,12 @@ public class StorageExample {
 
   private abstract static class AclAction extends StorageAction<Tuple<BlobId, Acl>> {
 
+    /**
+     * Sets the ACL according to the provided {@code params}, using the {@code storage} service. If
+     * {@code params.x()} returns a complete blob identity, the {@code params.y()} ACL is added to
+     * the blob. If {@code params.x().name()} is empty, the {@code params.y()} ACL is added to the
+     * bucket identified by {@code params.x().bucket()}.
+     */
     @Override
     public void run(Storage storage, Tuple<BlobId, Acl> params) {
       BlobId blobId = params.x();


gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
```diff
@@ -512,6 +563,188 @@ public String params() {
     }
   }
 
+  private abstract static class AclAction extends StorageAction<Tuple<BlobId, Acl>> {
+
+    @Override
+    public void run(Storage storage, Tuple<BlobId, Acl> params) {
```
lesv: This should have some explanation that we have two paths, one for just the bucket, and one for an actual blob.  We are showing both in this example.

lesv:
Where do you expect this to be used?  

If it's our formal docs, you are missing `// [START sectionName]` and `// [END sectionName]` tags.

Putting them in threads doesn't seem super clear, but that probably started before your time.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
```diff
@@ -512,6 +563,188 @@ public String params() {
...
+   * This class demonstrates how to add an ACL to a blob or a bucket for a group of users
+   * (identified by the group's email).
+   *
+   * @see <a href="https://cloud.google.com/storage/docs/access-control/lists#permissions">Access
+   *     Control Lists (ACLs)</a>
+   */
+  private static class AddGroupAclAction extends AclAction {
+
+    @Override
+    Tuple<BlobId, Acl> parse(String... args) {
```
lesv: Lots of setup for what is actually 3 line example - with much noise.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
```diff
@@ -512,6 +563,188 @@ public String params() {
...
+          nextArg = 1;
+        } else if (args.length == 4) {
+          blob = BlobId.of(args[0], args[1]);
+          nextArg = 2;
+        } else {
+          throw new IllegalArgumentException("Too many arguments.");
+        }
+        String group = args[nextArg++];
+        Acl.Role role = Acl.Role.valueOf(args[nextArg]);
+        return Tuple.of(blob, Acl.of(new Acl.Group(group), role));
```
lesv: Actually, this doesn't seem like it does anything.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
mziccard: > Lots of setup for what is actually 3 line example - with much noise.

I agree with this. However this is the structure that we gave so far to this kind of examples.

If you look at the `gcloud-java-examples` module you will see that we have one big example for each module (called `<Service>Example.java`) and a bunch of snippets (that contain code we reference in our READMEs). I think the idea behind this big examples (this precedes me) was that the best way to make a meaningful example that spans over all services functionality was coding a kind of command line tool. Given that the example basically is a command line tool we buy all the "noisy" code for parsing command line arguments.

mziccard:
> Where do you expect this to be used?
If it's our formal docs, you are missing // [START sectionName] and // [END sectionName] tags.

These examples don't go in the formal docs (AFAIK `gcloud-java` should not/will not go in the formal docs until v1.0.0 is out). We just reference them in our READMEs (see [here](https://github.com/GoogleCloudPlatform/gcloud-java/tree/master/gcloud-java-examples) for example).

> Putting them in threads doesn't seem super clear, but that probably started before your time.

What do you mean with this?

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
mziccard: I am not sure I understand this.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
mziccard: Agree, will add a comment for that.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
lesv: It's returning a Tuple, but does the ACL actually get set in GCS?  It may be a side effect I don't understand, but I don't see the blob actually getting the new ACL.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
```diff
@@ -512,6 +563,188 @@ public String params() {
...
+          nextArg = 1;
+        } else if (args.length == 4) {
+          blob = BlobId.of(args[0], args[1]);
+          nextArg = 2;
+        } else {
+          throw new IllegalArgumentException("Too many arguments.");
+        }
+        String domain = args[nextArg++];
+        Acl.Role role = Acl.Role.valueOf(args[nextArg]);
+        return Tuple.of(blob, Acl.of(new Acl.Domain(domain), role));
```
lesv: same issue here as 623?

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
mziccard: The actual RPC is carried out in the base class `AclAction`, see the `run` method.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
lesv: We should at least have a line break between the setup and the meat of the example -- it is easier for users to "get it" that way.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
lesv: Cool -- then the comment at the top is incorrect.  this is just creating the data structure for later action.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
```diff
@@ -512,6 +563,188 @@ public String params() {
...
+    }
+
+    @Override
+    public String params() {
+      return "<bucket> <path>? <group> OWNER|READER|WRITER";
+    }
+  }
+
+  /**
+   * This class demonstrates how to add an ACL to a blob or a bucket for a domain.
```
lesv: create the data structure for ...

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
```diff
@@ -512,6 +563,188 @@ public String params() {
...
+    private static List<Acl> addAcl(List<Acl> acls, Acl newAcl) {
+      List<Acl> newAcls = new LinkedList<>(acls);
+      newAcls.add(newAcl);
+      return newAcls;
+    }
+  }
+
+  /**
+   * This class demonstrates how to add an ACL to a blob or a bucket for a group of users
+   * (identified by the group's email).
```
lesv: creates the data structure for ...

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
mziccard: Not sure I agree with this, the class itself parses and potentially executes the RPC, it's just that the RPC part (`run` method) is inherited from the superclass.

A similar javadoc is also used in other examples that use a hierarchy of actions (see [TableAction](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-examples/src/main/java/com/google/cloud/examples/bigquery/BigQueryExample.java#L395) and subclasses in `BigQueryExample` for instance).

If you dislike this, I suggest we open an issue to be dealt with in a separate PR.

lesv: SGTM - I just read the example in full, its good correct code, but it's terrible as a teaching / learning vehicle.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java
lesv: ignore

lesv:
I think you'll find that they will start being more in formal docs quickly.

lesv:
I think there are a couple of AI's you agreed to above, do them, and it's ok to merge.  I'll figure out what I want as next steps and write an issue for it.

PR #1034, opened 2016-05-31
Add createTime() to BlobInfo and Blob classes

### LINKED ISSUE #1032: Storage BlobInfo createdTime not present in library
Hello,

I have been using the library and checking the [JSON API exposes] [1], as a reference to use Google Cloud Storage, I haven't been able to find these fields:

  * timeCreated
  * any of the encryption attributes

However one can find them in the library this library is relying on: https://developers.google.com/resources/api-libraries/documentation/storage/v1/java/latest/com/google/api/services/storage/model/StorageObject.html

Would it be possible to at least expose the timeCreated field?

Thank you!

[1]: https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations

mziccard:
This PR adds the missing `createTime` field and fixes #1032

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/gcloud-java-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -296,6 +296,12 @@ public class Blob extends BlobInfo {
     }
 
     @Override
+    Builder createTime(Long createTime) {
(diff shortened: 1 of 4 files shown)


lesv:
PTAL at codacy and comment, but LGTM.

PR #1040, opened 2016-06-02
Add support for entities with incomplete key to Datastore.put

### LINKED ISSUE #638: Support put for parameters of type FullEntity<IncompleteKey> in v1beta3

mziccard:
Support is also added to `DatastoreBatchWriter`. This fixes #638 

/cc @lesv @aozarov @ajkannan


lesv:
please take codacy/pr's suggestion.
Isn't this incomplete keys, not incomplete entities?  Perhaps you should change the title of this PR.
Once that is done, I'm happy w/ your code.

lesv:
I'm going to have to read a bunch more code, but it appears that most operations are buffered, do we have guarantees of when writes actually happen?  Must they be within transactions?

lesv:
Thanks - LGTM after you remove the two final's suggested by `codacy/pr`.

aozarov:
Looks good!

PR #1041, opened 2016-06-03
Implement Iterator pull methods, add javadoc and tests

mziccard:
This PR implements `Iterator<ReceivedMessage> pull` and `Future<Iterator<ReceivedMessage>> pullAsync` method.
This PR also:
- Adds unit tests for `pull` methods
- Adds system tests for `pull` methods, `ack`/`nack` messages and `renewAckDeadline`
- Add shorter timeout for `AckDeadlineRenewer` tests
- Add `closed` field to both `PubSubImpl` and `DefaultPubSubRpc`

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
@@ -16,9 +16,9 @@
 
 package com.google.cloud.pubsub;
 
+import static com.google.api.client.util.Preconditions.checkArgument;
 import static com.google.cloud.pubsub.PubSub.ListOption.OptionType.PAGE_SIZE;
 import static com.google.cloud.pubsub.PubSub.ListOption.OptionType.PAGE_TOKEN;
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.util.concurrent.Futures.lazyTransform;
 
 import com.google.cloud.AsyncPage;
@@ -29,10 +29,12 @@ import com.google.cloud.PageImpl;
 import com.google.cloud.pubsub.spi.PubSubRpc;
 import com.google.cloud.pubsub.spi.v1.PublisherApi;
 import com.google.cloud.pubsub.spi.v1.SubscriberApi;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Uninterruptibles;
@@ -52,6 +54,8 @@ import com.google.pubsub.v1.ModifyAckDeadlineRequest;
 import com.google.pubsub.v1.ModifyPushConfigRequest;
 import com.google.pubsub.v1.PublishRequest;
 import com.google.pubsub.v1.PublishResponse;
+import com.google.pubsub.v1.PullRequest;
+import com.google.pubsub.v1.PullResponse;
 
 import java.util.Collections;
 import java.util.Iterator;
@@ -64,6 +68,8 @@ import java.util.concurrent.TimeUnit;
 class PubSubImpl extends BaseService<PubSubOptions> implements PubSub {
 
   private final PubSubRpc rpc;
+  private final AckDeadlineRenewer ackDeadlineRenewer;
+  private boolean closed;
 
   private static final Function<Empty, Void> EMPTY_TO_VOID_FUNCTION = new Function<Empty, Void>() {
     @Override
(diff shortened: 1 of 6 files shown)


gcloud-java-pubsub/src/test/java/com/google/cloud/pubsub/PubSubImplTest.java
```diff
@@ -1191,149 +1229,209 @@ public void testListTopicSubscriptionsAsyncWithOptions()
   }
 
   @Test
+  public void testPullMessages() {
+    pubsub = new PubSubImpl(options, renewerMock);
+    PullRequest request = PullRequest.newBuilder()
+        .setSubscription(SUBSCRIPTION_NAME_PB)
+        .setMaxMessages(42)
```
lesv: I'm presuming there is no significance  to this other than the obvious reference to HG2G.

lesv:
LGTM - codacy/pr doesn't seem happy, but I couldn't find anything I agreed with it on.

gcloud-java-pubsub/src/test/java/com/google/cloud/pubsub/PubSubImplTest.java
mziccard: > I'm presuming there is no significance to this other than the obvious reference to HG2G.

Yes, no significance

mziccard:
@lesv thanks for the pass!

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
```diff
@@ -16,9 +16,9 @@
 
 package com.google.cloud.pubsub;
 
+import static com.google.api.client.util.Preconditions.checkArgument;
```
aozarov: Wrong import. You can configure IntelliJ to avoid import suggestions from `com.google.api.client.util.*`

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
```diff
@@ -445,17 +466,35 @@ public SubscriptionId apply(String compositeSubscription) {
...
-    // this method also should use the VTKIT thread-pool to renew ack deadline for non consumed
-    // messages
-    return null;
+  public Future<Iterator<ReceivedMessage>> pullAsync(final String subscription, int maxMessages) {
+    PullRequest request = PullRequest.newBuilder().setReturnImmediately(true)
+        .setSubscription(SubscriberApi.formatSubscriptionName(options().projectId(), subscription))
+        .setMaxMessages(maxMessages)
+        .setReturnImmediately(true)
+        .build();
+    Future<PullResponse> response = rpc.pull(request);
```
aozarov: I think we have a problem here. lazyTransorm will only happen when `get` is called.
Hence, ack-deadline renew only starting from that point which may be too late.

gcloud-java-pubsub/src/test/java/com/google/cloud/pubsub/BaseSystemTest.java
```diff
@@ -22,6 +22,7 @@
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
+import com.google.api.client.util.Lists;
```
aozarov: wrong `Lists`

aozarov:
I think this PR has a few things that need to be fixed.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
mziccard: I could not find any other reasonable way of implementing this while keeping the async nature of the method. We could use another thread to wait for the future to complete (basically faking a listenable future) and add messages to the renewer as soon as they are pulled. The problem with this approach is that we cannot run such a thread on the same executor of the "RPC future" (as the pool's threads could potentially be all occupied by waiting tasks, thus starving the RPC tasks). Having an executor just for this feature seemed too much to me.

This seemed like a reasonable compromise to me.

aozarov: I don't think this is good enough. The whole notion of this async is get it in parallel and deal with it at some point. With this implementation that "some point" must be "almost immediately" otherwise all the messages might deadline might expire.

I think there are more ways to fix it. One way is to make the SPI layer expose the `ListenableFuture` as returned by the underlying VTKIT library or add our own version
of it (or use a Callback instead for this method).

mziccard: If in your opinion exposing `ListenableFuture` in the SPI layer is ok I would go with it. This would also make implementing `MessageConsumer` easier.

mziccard: Better than exposing `ListenableFuture` (until we migrate to Java8) we could expose something like:

```
Future<PullResponse> pull(String, int, PullCallback);
```
where `PullCallback` is:
```
interface PullCallback {
  void onSuccess(PullResponse response);
  void onFailure(Throwable error);
}
```
Thoughts?

aozarov: That sounds good to me, however I am not sure why you need to return a Future in this case.

mziccard: Yep, I agree we actually don't need to return a future in that case.

PR #1045, opened 2016-06-08
Update bigquery dependency and add support for BYTES datatype

### LINKED ISSUE #1039: Add support for BigQuery BYTES data type
BigQuery [BYTES](https://cloud.google.com/bigquery/data-types) data type was recently added (see [release notes](https://cloud.google.com/bigquery/release-notes#march_23_2016)). We should add support for it to both [`Field`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Field.java) and [`FieldValue`](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java) classes.

mziccard:
This PR does the following:
- update the `google-api-services-bigquery` dependency and applies the required adjustments
- add support and tests for the BYTES data type (fixes #1039)

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
@@ -73,7 +73,7 @@ public final class Field implements Serializable {
     private static final long serialVersionUID = 2841484762609576959L;
 
     public enum Value {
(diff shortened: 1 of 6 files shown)


lesv:
LGTM

PR #1048, opened 2016-06-09
Fix PubSub Iterator pullAsync: add callback to PubSubRpc.pull

mziccard:
This PR follows from the discussion in #1041 and fixes `Future<Iterator<ReceivedMessage>> pullAsync`.
`PubSubRpc.pull` changes signature to `PullFuture pull(String)`. `PullFuture` extends `Future<PullResponse>` and allows to register callbacks to the RPC's future so messages can be added to the ack deadline renewer as soon as pulled.

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
@@ -31,6 +31,7 @@ import com.google.cloud.pubsub.spi.v1.SubscriberApi;
 import com.google.cloud.pubsub.spi.v1.SubscriberSettings;
 import com.google.common.base.Function;
 import com.google.common.collect.Sets;
+import com.google.common.util.concurrent.ForwardingListenableFuture;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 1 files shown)


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
```diff
@@ -89,6 +93,56 @@ protected ExecutorFactory executorFactory() {
     }
   }
 
+  private static final class PullFutureImpl implements PullFuture {
```
aozarov: [SimpleForwardingFuture](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/ForwardingFuture.SimpleForwardingFuture.html)?

aozarov:
Looks good to me. I didn't look at the tests or the doc changes. I will let @lesv do that.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
mziccard: Better [SimpleForwardingListenableFuture](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/ForwardingListenableFuture.SimpleForwardingListenableFuture.html) so we can avoid casting `delegate()`. Fixed in 752fab0.

mziccard:
Comment fixed. @lesv ready if you want to have al look.

lesv:
Comments look fine 
Tests - look fine, but I wonder if we can't test more than we do.
LGTM

Issue #809, opened 2016-03-29
NIO: Close channels on FS close


jart:
CloudStorageFileSystem#close() should synchronously close all active channels associated with the file system instance, per NIO documentation.

garrettjonesgoogle:
@jean-philippe-martin this issue is more than a year old, can you comment on the need to address it?

danoscarmike:
This issue may no longer be relevant due to its age.  If you are still experiencing this issue and are reasonably certain that it is a bug, please feel free to re-open this it with enough information to help us reproduce it.

jean-philippe-martin:
close() is still a no-op, in violation of the NIO documentation. It hasn't been an issue but would be nice to have. I would argue we should keep this issue open.

PR #1060, opened 2016-06-16
File system close declares it throws IOException

jean-philippe-martin:
This is in case we decide to implement "close all owned channels" (issue #809)

cc:mziccard

### DIFF
--- a/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystem.java
+++ b/gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystem.java
@@ -145,10 +145,12 @@ public final class CloudStorageFileSystem extends FileSystem {
    * Does nothing.
    */
   @Override
-  public void close() {}
+  public void close() throws IOException {
+    // TODO(#809): Synchronously close all channels associated with this FileSystem instance.
+  }
 
   /**
-   * Returns {@code true}.
+   * Returns {@code true}, even if you previously called the {@link #close()} method.
    */
   @Override
   public boolean isOpen() {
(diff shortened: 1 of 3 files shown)


gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystem.java
```diff
@@ -145,10 +145,12 @@ public CloudStoragePath getPath(String first, String... more) {
    * Does nothing.
    */
   @Override
-  public void close() {}
+  public void close() throws IOException {
```
mziccard: Just wondering: how much would it take to actually close all channels rather that postponing this?

mziccard:
See my comment on implementing the method rather than putting a TODO. Otherwise LGTM.

gcloud-java-contrib/gcloud-java-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystem.java
jean-philippe-martin: It would actually be a significant refactor, because right now channels are created by the CloudStorageFileSystemProvider, completely bypassing the CloudStorageFileSystem. We'd probably want to think carefully before that refactor. And unless it's a requirement for merging into master, I'd rather first merge into master and then do it since every passing day means the codebases deviate a bit more.

mziccard: > We'd probably want to think carefully before that refactor. And unless it's a requirement for merging into master, I'd rather first merge into master

Fine with me

PR #1062, opened 2016-06-19
Add MonitoredResourceDescriptor class and tests

mziccard:


lesv:
LGTM

mziccard:
`MonitoredResourceDescriptor` and `LabelDescriptor` (as well as `MonitoredResource`) entities are also used in [Stackdriver monitoring](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.monitoredResourceDescriptors/list). We should move those classes to `gcloud-java-core`. This also means that conversion methods (`toPb` and `fromPb`) will have to become public.

lesv:
Are any of the other (API's) toPb() / fromPb() public elsewhere?

mziccard:
> Are any of the other (API's) toPb() / fromPb() public elsewhere?

No, they are all package scope. But to reuse these classes we might have no choice.

lesv:
Sigh - SGTM

mziccard:
@lesv I moved the classes to the core package, PTAL

lesv:
LGTM

PR #1065, opened 2016-06-20
Handle files over 2GB

jean-philippe-martin:
ParallelCountBytes fails for files over 2GB. It turns out that's because the position is stored internally as an int instead of a long in gcloud-java-storage. This CL fixes this. Once we merge into master, it'll solve the same problem for master.

cc: @jart

### DIFF
--- a/gcloud-java-storage/src/main/java/com/google/cloud/storage/spi/DefaultStorageRpc.java
+++ b/gcloud-java-storage/src/main/java/com/google/cloud/storage/spi/DefaultStorageRpc.java
@@ -31,6 +31,7 @@ import static com.google.cloud.storage.spi.StorageRpc.Option.PREDEFINED_DEFAULT_
 import static com.google.cloud.storage.spi.StorageRpc.Option.PREFIX;
 import static com.google.cloud.storage.spi.StorageRpc.Option.VERSIONS;
 import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.base.Preconditions.checkArgument;
 import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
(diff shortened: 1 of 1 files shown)


gcloud-java-storage/src/main/java/com/google/cloud/storage/spi/DefaultStorageRpc.java
```diff
@@ -467,10 +467,13 @@ public RpcBatch createBatch() {
           .setIfMetagenerationNotMatch(IF_METAGENERATION_NOT_MATCH.getLong(options))
           .setIfGenerationMatch(IF_GENERATION_MATCH.getLong(options))
           .setIfGenerationNotMatch(IF_GENERATION_NOT_MATCH.getLong(options));
+      if (position<0) {
```
mziccard: Use Guava's [`checkArgument`](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Preconditions.html#checkArgument(boolean, java.lang.Object)):
```java
checkArgument(position >= 0, "Position should be non-negative, is %d", position);
```

mziccard:
Nice catch, see one minor comment. I wonder whether we should merge this fix into master so that we don't have to wait for the bigger `gcs-nio` to be merged in.

gcloud-java-storage/src/main/java/com/google/cloud/storage/spi/DefaultStorageRpc.java
jean-philippe-martin: Imported and called checkArgument.

jean-philippe-martin:
Tests are green. Are we OK to merge this pull request?

jean-philippe-martin:
Thank you!

PR #1066, opened 2016-06-21
Add docs to PubSub spi layer

mziccard:


lesv:
LGTM - I'm happy to see this, but wonder who we expect to call these directly?

mziccard:
> I'm happy to see this, but wonder who we expect to call these directly?

Those are not supposed to be called directly by users. But some documentation is not going to hurt :)

lesv:
Sg - LGTM

PR #1069, opened 2016-06-21
Document limitations of FakeStorageRpc

jean-philippe-martin:


gcloud-java-storage/src/main/java/com/google/cloud/storage/testing/FakeStorageRpc.java
```diff
@@ -37,9 +37,16 @@
 
 /**
  * A bare-bones in-memory implementation of Storage, meant for testing.
- * See LocalGcsHelper.
  *
- * This class is NOT thread-safe.
+ * This class is NOT thread-safe. It's also (currently) limited in the following ways:
```
jart: Consider:

```
 * <p>This class is NOT thread-safe. It's also (currently) limited in the following ways:
 * <ul>
 * <li>Partially Supported
 *   <ul>
 *   <li>create
 *   <li>get
 *   <li>delete
 *   <li>directory listing
 *   </ul>
 * <li>Unsupported
 *   <ul>
 *   <li>generations
 *   <li>file attributes
 *   <li>patch
 *   <li>continueRewrite
 *   </ul>
 * </ul>
```

jean-philippe-martin: done.

jart:
Please squash to one commit.

jean-philippe-martin:
Squashed to a single commit.

jart:
LGTM

gcloud-java-storage/src/main/java/com/google/cloud/storage/testing/FakeStorageRpc.java
```diff
@@ -37,9 +37,24 @@
 
 /**
  * A bare-bones in-memory implementation of Storage, meant for testing.
- * See LocalGcsHelper.
  *
- * This class is NOT thread-safe.
+ * <p>This class is <i>not</i> thread-safe. It's also (currently) limited in the following ways:
+ * <ul>
+ * <li>Partially Supported
```
mziccard: What does partially supported mean? As a user I would like some more detail on that.

jean-philippe-martin: It means I implemented one of the two methods.

jean-philippe-martin: Sorry, it looks like we merged too fast. I'm making a new PR where we can
discuss additional detail we want in the description.
It's at https://github.com/GoogleCloudPlatform/gcloud-java/pull/1070

gcloud-java-storage/src/main/java/com/google/cloud/storage/testing/FakeStorageRpc.java
```diff
@@ -37,9 +37,24 @@
...
+ *   <ul>
+ *   <li>create
+ *   <li>get
+ *   <li>delete
+ *   <li>directory listing
+ *   </ul>
+ * <li>Unsupported
```
mziccard: It seems to me that also bucket operations (get, list, delete), batches and compose are not supported

Not related to this PR: I might be missing something but why do some methods throw UnsupportedOperationException (e.g. createBatch) while others return null (e.g. compose)?

gcloud-java-storage/src/main/java/com/google/cloud/storage/testing/FakeStorageRpc.java
mziccard: Sorry for the late comments but it's 10pm here and I shouldn't even be working :)

gcloud-java-storage/src/main/java/com/google/cloud/storage/testing/FakeStorageRpc.java
jean-philippe-martin: I *think* we return null in a few places because otherwise the tests wouldn't work. That is, we have to implement this method but a minimal implementation is enough.

jean-philippe-martin: As to the first comment, that's exactly what I meant by partially supported. I [pushed](https://github.com/GoogleCloudPlatform/gcloud-java/pull/1070) a change to split those apart in the description.

gcloud-java-storage/src/main/java/com/google/cloud/storage/testing/FakeStorageRpc.java
jean-philippe-martin: No worries! We can continue another day!

PR #1074, opened 2016-06-23
Add javadoc and tests for Subscription

mziccard:


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/Subscription.java
```diff
@@ -18,6 +18,7 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.cloud.GrpcServiceOptions;
```
mziccard: This import is needed for javadoc, we can ignore this comment

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/Subscription.java
```diff
@@ -103,62 +131,172 @@ public Builder toBuilder() {
...
+   * used to change a push subscription to a pull one (passing a {@code null} {@code pushConfig}
+   * parameter) or vice versa. This methods can also be used to change the endpoint URL and other
+   * attributes of a push subscription. Messages will accumulate for delivery regardless of changes
+   * to the push configuration. The method returns a {@code Future} object that can be used to wait
+   * for the replace operation to be completed.
+   *
+   * @param pushConfig the new push configuration. Use {@code null} to unset it
+   * @return a {@code Future} to wait for the replace operation to be completed.
+   */
   public Future<Void> replacePushConfigAsync(PushConfig pushConfig) {
```
lesv: This really is void?  No status of success?

mziccard: Yep it's void. If it returns successfully the change was applied otherwise you get an exception.

lesv:
LGTM

PR #1075, opened 2016-06-23
Add PubSub examples, update READMEs and package-info javadoc

mziccard:


lesv:
LGTM - comments:
1. I wish Github allowed :include in their markdown
2. We will probably need to reformat the snippets a bit when we publish or at least add doc tags, but that's a bit in the future.

Thanks -- looks good.

mziccard:
> 1. I wish Github allowed :include in their markdown

You're not alone :) https://github.com/github/markup/issues/172#issuecomment-33241601

PR #1078, opened 2016-06-24
Support setting library header in grpc services

mziccard:
This PR does the following:
- add `libraryName()` and `libraryVersion()` to `ServiceOptions`
- use `setClientLibHeader` in `DefaultPubSubRpc`

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/gcloud-java-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -442,7 +442,7 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
   }
 
   /**
-   * Returns {@code gcloud-java}, the library's name, as a string.
+   * Returns the library's name, {@code gcloud-java}, as a string.
    */
   public String libraryName() {
     return LIBRARY_NAME;


gcloud-java-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -439,6 +441,20 @@ public String applicationName() {
     return APPLICATION_NAME;
   }
 
+  /**
+   * Returns {@code gcloud-java}, the library's name, as a string.
```
aozarov: Fine as is though I like this one slightly better:
`Returns the library's name, {@code gcloud-java}, as a string.`

gcloud-java-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -215,6 +219,16 @@ public void testBaseEquals() {
   }
 
   @Test
+  public void testLibraryName() {
+    assertEquals(LIBRARY_NAME, OPTIONS.libraryName());
+  }
+
+  @Test
+  public void testApplicationName() {
```
aozarov: add a separate test to `libraryVersion`?

aozarov:
Looks good. 2 minor comments.

gcloud-java-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
mziccard: Library version is not returned when tests are run, I believe due to the fact that classes are loaded directly and not through the jar, so no manifest is there.

aozarov: Ah, you are right. It could probably be fixed by providing `META-INF/MANIFEST.MF` as a test resource (which would also make the `testApplicationName` a bit more relevant). Feel free to try or ignore or create a separate issue for it.

mziccard: Yes but that will cause the "fake" manifest to be put into the tests jar. Unless we exclude it via maven configuration. I'll give this some thoughts but I would leave it for a future PR.

aozarov: I am not sure where it would be placed in the tests.jar (under resources or at the top and if the latter would it be overridden with the manifest created by the jar) but I think it is doable. future PR sounds good to me.

aozarov:
Looks good!

PR #1079, opened 2016-06-24
Temporarily hide FakeStorageRpc and LocalStorageHelper

mziccard:
We should rollback this change as soon as `FakeStorageRpc` and `LocalStorageHelper` support a minimal set of functionality (we should implement bucket operations).

As soon as this is merged I will open an issue for moving `FakeStorageRpc` and `LocalStorageHelper` back.


jean-philippe-martin:
Sounds good to me!

PR #1080, opened 2016-06-24
Add integration tests to pubsub

mziccard:
This adds integration tests for PubSub and makes Travis more tolerant if tests are taking long.


lesv:
LGTM - Cool, Thanks.

PR #1083, opened 2016-06-27
Add PHP to language menu, other minor fixes

mziccard:
This PR adds `gcloud-php` to the landing page's menu and renames every use of `gcloud` to `gcloud-java` when required. Another PR will follow to actually update the landing page for the currently deployed site.


lesv:
LGTM

mziccard:
@lesv thanks

PR #1089, opened 2016-06-29
Fix Travis-related scripts

mziccard:
This PR does:
- fix the script that updates READMEs' versions to also add and commit nested READMEs
- Update github credentials used by travis (replaced "secure" should be the correct ones, they were added [here](https://github.com/GoogleCloudPlatform/gcloud-java/commit/e0113f30b6d767bff72257b8f6f821d4a5149caa))


mziccard:
Please @aozarov take a quick look

PR #1091, opened 2016-06-30
Add Metric class, implement metric methods and tests

mziccard:


lesv:
LGTM - you could also checkin your Integration tests at some point.

mziccard:
> you could also checkin your Integration tests at some point.

Yep, they will come soon

PR #1099, opened 2016-07-05
Add HttpRequest class and tests

mziccard:


lesv:
LGTM - Surprise - machine generated code might be difficult to read / maintain. (toPb / fromPb)

PR #1100, opened 2016-07-05
Add support for time-partitioned tables

C0urante:
Time partitioning can now be specified during table creation, and can be requested as a table option when fetching table information.

Currently, it is only represented as a boolean value in the client side, since the only possible partitioning scheme is by day. This is documented in TableInfo's Javadoc.

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
@@ -21,6 +21,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.client.util.Data;
 import com.google.api.services.bigquery.model.Table;
+import com.google.api.services.bigquery.model.TimePartitioning;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 
@@ -63,6 +64,7 @@ public class TableInfo implements Serializable {
   private final Long creationTime;
   private final Long expirationTime;
   private final Long lastModifiedTime;
+  private final TimePartitioning timePartitioning;
   private final TableDefinition definition;
 
   /**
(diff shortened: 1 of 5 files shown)


mziccard:
Hi @C0urante, thanks for your contribution. I am going through your changes right now.

gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
```diff
@@ -63,6 +64,7 @@ public Table apply(TableInfo tableInfo) {
   private final Long creationTime;
   private final Long expirationTime;
   private final Long lastModifiedTime;
+  private final TimePartitioning timePartitioning;
```
mziccard: Why storing `TimePartitioning` if we only expose this option as a boolean?

Also I believe we should no expose it as just a boolean as we miss the chance to configure the [`expirationMs` field](timePartitioning.expirationMs). Also I have the feeling this time partitioning option will be extended to finer granularities in the future.

My preferred approach here is to create our own `TimePartitioning` class (that follows our code style: with `toPb` and `fromPb` methods, as well as convenience `builder` and `of` methods) and add that as a field to `TableInfo`. A `Builder timePartitioning(TimePartitioning)` setter should be added to the builder and a `TimePartitioning timePartitioning()` getter to `TableInfo`.

I can take care of applying such changes if you prefer.

C0urante: Thanks for the review! I would definitely appreciate it if you took care of these changes yourself.

mziccard: Sure thing! I'll cherry pick your commit and open a new PR, I'll /cc you as soon as that is ready so you can have a look and let me know if it suits your needs, how does this sound?

C0urante: That sounds wonderful! Thank you very much.

PR #1102, opened 2016-07-06
Add log entry Payload class and tests

mziccard:
This PR adds a `Payload` class with 3 subclasses, one for each possible payload: UTF8 string, JSON, and protobuf object.

`JsonPayload` is also exposed as a `Map<String, Object>` (along with protobuf's `Struct`) to be friendlier to users and to be consistent with what we do in other parts of the library. `Structs` class contains utility methods to convert from `Struct` to `Map<String, Object>` and back.


lesv:
I'm happy w/ what is there, but I would expect that something that mentions testing JSON to actually have some raw JSON to either start or compare against.  The same could be said of the Struct testing as well.

If you disagree, feel free to merge.

PR #1103, opened 2016-07-07
Add automatic check for file header on mvn validate

### LINKED ISSUE #1071: Add compile-time check that source files start with license text
As [initially suggested](https://github.com/GoogleCloudPlatform/gcloud-java/pull/1067#discussion_r67925353) by @aozarov, this would make sure that code checked in follows that part of the guidelines.

mziccard:
This also adds license headers where they were missing or mis-formatted.

Fixes #1071

### DIFF
--- a/license-checks.xml
+++ b/license-checks.xml
@@ -3,9 +3,8 @@
         "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
         "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
 <module name="Checker">
-    <module name="Header">
+    <module name="RegexpHeader">
         <property name="fileExtensions" value="java"/>
-        <property name="ignoreLines" value="2"/>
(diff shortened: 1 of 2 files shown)


license-checks.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+<module name="Checker">
+    <module name="Header">
+        <property name="fileExtensions" value="java"/>
+        <property name="ignoreLines" value="2"/>
```
aozarov: Did you consider [this](http://checkstyle.sourceforge.net/config_header.html#RegexpHeader) as an alternative?

mziccard: That looks better in fact. Implemented in cae87d6, PTAL

PR #1107, opened 2016-07-12
Add logging.properties for travis build

mziccard:
This file is to be used by travis build scripts.

### DIFF
--- a/logging.properties
+++ b/logging.properties
@@ -1,0 +1,1 @@
+.level= INFO


mziccard:
This PR is required by #1108. These 2 PR in combination reduce the amount of logging produced by Travis build.

/cc @lesv

lesv:
LGTM

PR #1108, opened 2016-07-12
Reduce the log's size of Travis build

mziccard:
This depends on #1107 being merged first


lesv:
LGTM

mziccard:
I am glad to report that this change reduces Travis' log size from 2.7Mb to 1.1Mb

PR #1109, opened 2016-07-12
Exclude AutoValue class files from coverage computation

mziccard:
I just realized that the coveralls plugin is failing to find source code for `AutoValue_*.class` files. Which is reasonable as no source files exist for those classes. We need to exclude them from the coverage computation for coveralls to work properly.


lesv:
LGTM

PR #1111, opened 2016-07-14
Add RemoteLoggingHelper and integration tests

mziccard:


gcloud-java-logging/pom.xml
```diff
@@ -18,6 +18,11 @@
   </properties>
   <dependencies>
     <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative-boringssl-static</artifactId>
+      <version>1.1.33.Fork17</version>
```
lesv: Could be Fork18

lesv: Make that Fork19

lesv:
LGTM

gcloud-java-logging/pom.xml
mziccard: I am sticking to `fork17` as this is the suggested version in `grpc` docs see https://github.com/grpc/grpc-java/blob/master/SECURITY.md#openssl-statically-linked-netty-tcnative-boringssl-static

If you are ok with it I would stick to Fork17 until they update their docs (in case there are unforeseen compatibility issues).

lesv:
Totally cool w/ that Fork17

PR #1114, opened 2016-07-15
Add .NET to language menu

mziccard:
This follows from https://github.com/GoogleCloudPlatform/gcloud-common/issues/151. A separate PR will apply this fix to the latest deployed site.


lesv:
LGTM

PR #1117, opened 2016-07-18
Add JUL handlers for Logging, examples and tests

mziccard:


gcloud-java-logging/src/main/java/com/google/cloud/logging/AsyncLoggingHandler.java
```diff
@@ -0,0 +1,97 @@
...
+ *     (defaults to 1).
+ * <li>{@code com.google.cloud.logging.AsyncLoggingHandler.flushLevel} specifies the flush log
+ *     level. When a log with this level is published, logs are transmitted to the Stackdriver
+ *     Logging service (defaults to {@link LoggingLevel#ERROR}).
+ * </ul>
+ *
+ * <p>To add a {@code LoggingHandler} to an existing {@link Logger} and be sure to avoid infinite
+ * recursion when logging, use the {@link #addHandler(Logger, LoggingHandler)} method.
+ */
+public class AsyncLoggingHandler extends LoggingHandler {
```
lesv: We we using this w/ JUL?  Should we mention the required config line somewhere?

gcloud-java-logging/src/test/java/com/google/cloud/logging/LoggingLevelTest.java
```diff
@@ -0,0 +1,81 @@
...
+  @Test
+  public void testDebug() {
+    LoggingLevel debug = LoggingLevel.DEBUG;
+    assertEquals(Severity.DEBUG, debug.severity());
+    assertEquals("DEBUG", debug.getName());
+    assertTrue(debug.intValue() < Level.FINEST.intValue());
+    assertTrue(debug.intValue() > Level.ALL.intValue());
+  }
+
+  @Test
```
lesv: AFAICT (As far as I can tell) - All these test are figuring out that Enumeration works as advertised - it does test that your mapping is as you described in the comments at the start of this PR, but it really seems pointless. (End of rant -- this won't block approval)

gcloud-java-logging/src/main/java/com/google/cloud/logging/AsyncLoggingHandler.java
mziccard: You mean mentioning in [this paragraph](https://github.com/GoogleCloudPlatform/gcloud-java/pull/1117/files/99a6ae1354a36eb83f13e5802b3fcb73df7a08d0#diff-f653e035fa60c430e457a3ce1150169dR72) that the handler can also be registered via {@logging.properties}? Yes I agree we should mention it, I'll fix this.

gcloud-java-examples/src/main/java/com/google/cloud/examples/logging/snippets/AddLoggingHandler.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.cloud.logging.LoggingHandler;
+
+import java.util.logging.Logger;
+
+/**
+ * A snippet showing how to use {@link java.util.logging.Logger} to log entries to Stackdriver
+ * Logging. The snippet shows how to install a Stackdriver Logging handler using
+ * {@link com.google.cloud.logging.LoggingHandler#addHandler(Logger, LoggingHandler)}. Notice that
+ * this could also be done through the {@code logging.properties} file.
```
lesv: Care to show what lines get added to logging.properties ??  What about if I'd also like to log locally as well.  (two or more copies - I can do that w/ J.U.L.)?

gcloud-java-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -0,0 +1,406 @@
...
+ * <li>{@code com.google.cloud.logging.LoggingHandler.flushSize} specifies the maximum size of the
+ *     log buffer. Once reached, logs are transmitted to the Stackdriver Logging service (defaults
+ *     to 1).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.flushLevel} specifies the flush log level.
+ *     When a log with this level is published, logs are transmitted to the Stackdriver Logging
+ *     service (defaults to {@link LoggingLevel#ERROR}).
+ * </ul>
+ *
+ * <p>To add a {@code LoggingHandler} to an existing {@link Logger} and be sure to avoid infinite
+ * recursion when logging, use the {@link #addHandler(Logger, LoggingHandler)} method.
```
lesv: Care to repeat logging.properties info?

gcloud-java-logging/src/test/java/com/google/cloud/logging/LoggingLevelTest.java
mziccard: I disagree. This class checks that the additional logging levels that we define in the `LoggingLevel` class are correctly initialized. Initialization here is very important as it defines how those levels relate to the original JUL levels (i.e. whether they are more or less severe). This will affect "loggability" at both logger and handler levels, as well as flushing in some cases. I believe it's rather important that we check that we initialize these extra levels properly. 

Also they JUL levels as well as our "extra" levels are not enums but instance of the `Level` class and need a name to be explicitly provided via constructor (hence this test checks that the expected string was provided).

lesv: Yep - figured that out.  I figured it would look like a rant when I wrote it, and after I spent a bit more time, I realized why you were doing what you were doing.

gcloud-java-examples/src/main/java/com/google/cloud/examples/logging/snippets/AddLoggingHandler.java
mziccard: > Care to show what lines get added to logging.properties ??

Yes I guess we should say that the corresponding `logging.properties` line would be:

```
com.google.cloud.logging. AddLoggingHandler.handlers= com.google.cloud.logging.LoggingHandler
```

> What about if I'd also like to log locally as well. (two or more copies - I can do that w/ J.U.L.)?

To add different handlers standard JUL methods work just fine.

lesv: I was thinking about this either in the comments or README.md

mziccard:
@lesv I added 60a00b3 to address the comments

lesv:
LGTM

PR #1119, opened 2016-07-19
Add complete README for logging

mziccard:


gcloud-java-logging/README.md
```diff
@@ -15,32 +17,155 @@ Java idiomatic client for [Google Cloud Logging][cloud-logging].
...
+```java
+LoggingHandler.addHandler(LOGGER, new LoggingHandler());
+```
+After that, logs generated using `LOGGER` will be also directed to Stackdriver Logging.
+
+Notice that you can also register a `LoggingHandler` via the `logging.properties` configuration
+file. Adding, for instance, the following line:
+```
+com.google.cloud.examples.logging.snippets.AddLoggingHandler.handlers=com.google.cloud.logging.LoggingHandler
+```
```
lesv: Is there any reason we would choose one over the other?

gcloud-java-logging/README.md
```diff
@@ -66,7 +193,9 @@ Contributing
 
 Contributions to this library are always welcome and highly encouraged.
 
-See [CONTRIBUTING] for more information on how to get started.
+See `gcloud-java`'s [CONTRIBUTING] documentation and the `gcloud-*` [shared documentation](https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/contributing/readme.md#how-to-contribute-to-gcloud) for more information on how to get started.
```
lesv: `[CONTRIBUTING](../CONTRIBUTING.md)`

lesv:
Just the one Question, if no, then Ok to merge.  (LGTM)

gcloud-java-logging/README.md
mziccard: There could be depending on your application needs. But I would say that this does not depend on our handler but rather it's a design choice that comes with JUL.

lesv: LOL - it's reasonable to punt on this.  I agree.

PR #1122, opened 2016-07-20
Add support for Stackdriver Logging

mziccard:
This PR rebases commits form `logging-alpha` branch on top of `master` branch. All commits but the last 3 have already been reviewed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ This client supports the following Google Cloud Platform services:
 -  [Google Cloud Compute] (#google-cloud-compute-alpha) (Alpha)
 -  [Google Cloud Datastore] (#google-cloud-datastore)
 -  [Google Cloud DNS] (#google-cloud-dns-alpha) (Alpha)
+-  [Stackdriver Logging] (#stackdriver-logging-alpha) (Alpha - Not working on App Engine Standard)
 -  [Google Cloud Pub/Sub] (#google-cloud-pubsub-alpha) (Alpha - Not working on App Engine Standard)
 -  [Google Cloud Resource Manager] (#google-cloud-resource-manager-alpha) (Alpha)
 -  [Google Cloud Storage] (#google-cloud-storage)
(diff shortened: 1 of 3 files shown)


mziccard:
@lesv can you have a look at the following commits:
- 4c211cc: renames all instances of "Cloud Logging" to "Stackdriver Logging"
- 49be9ac: re-orders service list in examples README and add LoggingExample to appassembler config
- c8806b5: adds logging service to main README and re-order services in TESTING.md

Also please notice that Codacy comments are on auto-generated code that we should not fix by hand in this PR.

mziccard:
@lesv can you have a look at the following commits:

7168fff: renames all instances of "Cloud Logging" to "Stackdriver Logging"
a783200: re-orders service list in examples README and add LoggingExample to appassembler config
1842375: adds logging service to main README and re-order services in TESTING.md
Also please notice that Codacy comments are on auto-generated code that we should not fix by hand in this PR.

lesv:
Does the example app get tested anywhere?

lesv:
Codacy seems to have a bunch of issues at on the surface seem relevant.  Could you  take a look and comment why we shouldn't fix them?

Also, if you could respond to the CLA bot.  Curious why it's unhappy - it usually likes you.

It's probably too late for this, but it would seem like in an ideal world we'd do a jar (maven artifact) for writers, and a separate one for reading.  Since most apps will be doing writing.

I thought I commented on one of the branches, but I can't find it now -- it was ignorable -- question about Cloud Logging  vs. Stackdriver Logging.

LGTM once you address Codacy's comments.

mziccard:
Thanks for the pass!

> Does the example app get tested anywhere?

Examples are not tested. We have [this issue](https://github.com/GoogleCloudPlatform/gcloud-java/issues/623) open that asks for example testing.

> Codacy seems to have a bunch of issues at on the surface seem relevant. Could you take a look and comment why we shouldn't fix them?

I had a look at the list of issues [here](https://www.codacy.com/app/mziccard/gcloud-java/pullRequest?prid=288317). I confirm that we have the following kind of issues:
- Issues with classes in `com.google.cloud.logging.spi.v2.*`: these classes are autogenerated so we should/can not address the issues here
- `println` or `printf` used in examples classes: we can ignore this error, prints are used for the sake of example
- Examples classes contain only static methods: as above, we can ignore this
- Other errors: these are all other errors that we decided to ignore when merging single PRs to the `logging-alpha` branch

> Also, if you could respond to the CLA bot. Curious why it's unhappy - it usually likes you.

This PR contains commits non only for me but also from other people, this is the reason CLA bot is unhappy. Luckily all other authors are googlers and already signed the CLA so CLA is ok in reality.

> It's probably too late for this, but it would seem like in an ideal world we'd do a jar (maven artifact) for writers, and a separate one for reading. Since most apps will be doing writing.

We discussed about this also on Pub/Sub but we decided to stick to one single surface for the whole service. Reasons are mainly:
- With the amount of services we need to support doing too much fragmentation might just be confusing. A 1:1 correspondence of our modules to Cloud services is easier to understand.
- Consistency with other `gcloud-java` modules
- Consistency with other `gcloud-*` libraries
- It's not straightforward to decide whether some functionality should go in the "writer" or "reader" artifact. Where should the `Metric`-related functionality go for instance?

> I thought I commented on one of the branches, but I can't find it now -- it was ignorable -- question about Cloud Logging vs. Stackdriver Logging.

I found it. Following [this issue](https://github.com/GoogleCloudPlatform/gcloud-common/issues/147) we should always name it `Stackdriver Logging` for consistency with service docs.

PR #1127, opened 2016-07-22
Hide string setter for DeprecationStatus timestamps

### LINKED ISSUE #732: Image's deprecationStatus can hold wrongly formatted timestamps

mziccard:
This PR fixes #732 on our side.

Since wrongly-formattted timestamps might still be around as part of older deprecated objects we must still save timestamps as strings and expose a string getter. This PR however hides the string setter and repaces todos with explanatory comments.

### DIFF
--- a/gcloud-java-compute/src/main/java/com/google/cloud/compute/DeprecationStatus.java
+++ b/gcloud-java-compute/src/main/java/com/google/cloud/compute/DeprecationStatus.java
@@ -96,8 +96,8 @@ public final class DeprecationStatus<T extends ResourceId> implements Serializab
      *
      * @see <a href="http://www.ietf.org/rfc/rfc3339.txt">RFC3339</a>
      */
-    // todo(mziccard): remove this method if #732 is closed
-    public Builder<T> deleted(String deleted) {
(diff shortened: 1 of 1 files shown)


lesv:
LGTM

PR #1129, opened 2016-07-22
Fix DNS integration tests for long zone names

### LINKED ISSUE #1128: Dns zone names can now be up to 64 characters long
Our DNS tests:
- `testCreateZoneWithErrors`
- `testCreateZoneWithErrorsBatch`

started failing out of the blue.

Those tests verify that zone creation fails with names that are longer than 32 characters. They started to fail because Cloud Dns now supports longer names for zones (up to 64 characters). We should go ahead and update our integration tests.

mziccard:
This fixes #1128


mziccard:
@lesv would you mind having a look?

lesv:
LGTM

PR #1139, opened 2016-08-05
update storage use example in README to create BlobInfo with builder …

drmarshall:
…instead of passing a String blobname, which is not supported as per docs:

http://googlecloudplatform.github.io/gcloud-java/0.2.6/apidocs/com/google/cloud/storage/Storage.html#create(com.google.cloud.storage.BlobInfo,%20byte[],%20com.google.cloud.storage.Storage.BlobTargetOption...)

### DIFF
--- a/gcloud-java-storage/README.md
+++ b/gcloud-java-storage/README.md
@@ -101,9 +101,9 @@ String bucketName = "my_unique_bucket"; // Change this to something unique
 Bucket bucket = storage.create(BucketInfo.of(bucketName));
 
 // Upload a blob to the newly created bucket
-BlobId blobId = BlobId.of(bucketName, "my_blob_name");
-Blob blob = storage.create(
-    "my_blob_name", "a simple blob".getBytes(UTF_8), "text/plain");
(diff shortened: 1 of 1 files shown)


mziccard:
Hi @drmarshall, thanks for doing this!

The example in the README is taken from the [CreateAndListBucketsAndBlobs](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/snippets/CreateAndListBucketsAndBlobs.java#L51) snippet. I believe that to fix the README we should just replace:
```java
Blob blob = storage.create(
    "my_blob_name", "a simple blob".getBytes(UTF_8), "text/plain");
```
with
```java
Blob blob = bucket.create(
    "my_blob_name", "a simple blob".getBytes(UTF_8), "text/plain");
```
Also, could you make the CLA bot happy? :)

mziccard:
Closing this as it was fixed in #1202. Thanks again @drmarshall for reporting this.

drmarshall:
Thanks Marco. Will stick to issues until I can appease CLA bot.

PR #1142, opened 2016-08-08
Add useLegacySql support to QueryRequest and QueryJobConfiguration

### LINKED ISSUE #1137: BigQuery Cannot specify to use Standard SQL instead of Legacy SQL.
http://googlecloudplatform.github.io/gcloud-java/0.2.6/apidocs/com/google/cloud/bigquery/QueryRequest.Builder.html

I want to be able to set useLegacySql to false here: https://cloud.google.com/bigquery/docs/reference/v2/jobs/query#request-body

Also an issue in Python: https://github.com/GoogleCloudPlatform/gcloud-python/issues/2053

mziccard:
This fixes #1137

@tswast does this fit with your needs?

### DIFF
--- a/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
+++ b/gcloud-java-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
@@ -51,6 +51,7 @@ public final class QueryJobConfiguration extends JobConfiguration {
   private final Boolean useQueryCache;
   private final Boolean flattenResults;
   private final Boolean dryRun;
+  private final Boolean useLegacySql;
 
   /**
    * Priority levels for a query. If not specified the priority is assumed to be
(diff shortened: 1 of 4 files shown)


lesv:
LGTM - give @tswast a few minutes to comment.

tswast:
Cool. Would you mind adding an example with use legacy sql set to false in https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-examples/src/main/java/com/google/cloud/examples/bigquery/BigQueryExample.java#L628 ?

mziccard:
> Would you mind adding an example with use legacy sql set to false

I am not sure `useLegacySql` or any "experimental" feature should be in the examples (also considering that examples do not cover all query options). In this case I expect `useLegacySql` to either disappear or be defaulted to `false` as soon as Standard SQL reaches maturity. However, if you feel strongly about this, I would be glad to add the example.

tswast:
Okay to merge from me. Discussed offline. We can wait until after standard SQL syntax goes GA before documenting in the examples.

PR #1143, opened 2016-08-08
Add SPI layer for Google Translate

mziccard:


gcloud-java-translate/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>gcloud-java-translate</artifactId>
+  <packaging>jar</packaging>
+  <name>GCloud Java translate</name>
+  <url>https://github.com/GoogleCloudPlatform/gcloud-java/tree/master/gcloud-java-translate</url>
+  <description>
+    Java idiomatic client for Google Translate.
+  </description>
+  <parent>
```
lesv: Any chance you could put more whitespace in here to make it a bit more readable?  ideally between major nodes.  Ok to say no, as this isn't example code, but folks do look at our code as how to write well.

gcloud-java-translate/src/main/java/com/google/cloud/translate/TranslateOptions.java
```diff
@@ -0,0 +1,187 @@
...
+  private final String targetLanguage;
+
+  public static class DefaultTranslateFactory implements TranslateFactory {
+
+    private static final TranslateFactory INSTANCE = new DefaultTranslateFactory();
+
+    @Override
+    public Translate create(TranslateOptions options) {
+      return null;
+      //return new TranslateImpl(options);
```
lesv: Is this a TODO(), or should it be removed?

lesv:
Please consider the two comments, otherwise, LGTM

gcloud-java-translate/src/main/java/com/google/cloud/translate/TranslateOptions.java
mziccard: Yes this is actutally a todo, I'll add an explicit comment!

gcloud-java-translate/pom.xml
mziccard: > Any chance you could put more whitespace in here to make it a bit more readable? 

I see not problems with this. However, we adopted this compact format across all modules so I believe we should change it once and for all of them (i.e. not in this PR).

lesv: SGTM

PR #1144, opened 2016-08-09
Fix setting HTTP transport factory for Datastore

### LINKED ISSUE #1138: The httpTransportFactory is not used
If you set the DatastoreOptions' builder's httpTransportFactory it is not used.  The "create" method of the factory is never called.  This does not allow me to configure a proxy.  This is a defect, but is there a workaround to configure a proxy?

`	public static Datastore createService() {
		HttpTransportFactory httpTransportFactory = new HttpTransportWithProxyFactory();
		return DatastoreOptions.builder()
				.httpTransportFactory(httpTransportFactory).build().
			service();
	}
	
	public static class HttpTransportWithProxyFactory implements HttpTransportFactory {
		
		public HttpTransport create() {
			// PUT BREAKPOINT ON NEXT LINE AND IT IS NEVER CALLED
			Proxy proxy = new Proxy(Type.HTTP, new InetSocketAddress("myproxy.com", 8080));
			HttpTransport t = new NetHttpTransport.Builder().setProxy(proxy).build();
			return t;
		}
	}`

mziccard:
This fixes #1138

@successhawk this should fix your issue, any chance you can give it a try?

### DIFF
--- a/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/spi/DefaultDatastoreRpc.java
+++ b/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/spi/DefaultDatastoreRpc.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.datastore.spi;
 
+import com.google.api.client.http.HttpTransport;
 import com.google.cloud.datastore.DatastoreException;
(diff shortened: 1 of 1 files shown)


lesv:
LGTM

PR #1146, opened 2016-08-10
Add snippets to Storage's javadoc

mziccard:
This goes in the direction of adding snippets to each method's javadoc. Does the job for the `Storage` interface.

Snippets are all taken from and link back to the `StorageSnippets` class.


lesv:
Either you write tests for this stuff, or I have to next Q.

LGTM

mziccard:
> Either you write tests for this stuff, or I have to next Q.

I understand. I still haven't come up with an effective way of testing this stuff:

- Against the actual service:
  - PRO: gives us more confidence on the soundness of these snippets
  - CONS: not easy to do, snippets are designed to show meaningful usages of the service, but it's not necessarily easy to set up the service state needed to execute the or to roll them back once finished testing
- Against a service mock:
  - PRO: easy to do
  - CONS: does not prove anything. A snippet might issue calls as expected but still not have the expected outcome when run against the service

lesv:
Yep - we try to test against an actual service.   Totally understand the problem.  What might make sense is to:

1. encapsulate example code in `// [start blah]` & `// [end blah]` tags.
1. code to help w/ testing s.b. outside the tags.

mziccard:
> encapsulate example code in // [start blah] & // [end blah] tags.
code to help w/ testing s.b. outside the tags.

I am working in this direction right now.

mziccard:
@lesv I updated the snippets to make them testable, added a script that parses a class containing snippets and adds them to the corresponding methods. I also added integration tests for `StorageSnippets`. I hope this makes you happy :)

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -0,0 +1,687 @@
...
+  }
+
+  /**
+   * Example of creating a blob with no content.
+   */
+  // [SNIPPET TARGET create(BlobInfo, BlobTargetOption...)]
+  // [SNIPPET VARIABLE "my_unique_bucket"]
+  // [SNIPPET VARIABLE "my_blob_name"]
+  public Blob createBlob(String bucketName, String blobName) {
+    // [SNIPPET START]
```
lesv: `[START createBlob]`

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -0,0 +1,687 @@
...
+    BlobId blobId = BlobId.of(bucketName, blobName);
+    BlobInfo blobInfo = BlobInfo.builder(blobId).contentType("text/plain").build();
+    Blob blob = storage.create(blobInfo);
+    // [SNIPPET END]
```
lesv: `\\ [END createBlob]`

Must have a unique name.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -0,0 +1,687 @@
...
+  private final Storage storage;
+
+  public StorageSnippets(Storage storage) {
+    this.storage = storage;
+  }
+
+  /**
+   * Example of creating a bucket.
+   */
+  // [SNIPPET TARGET create(BucketInfo, BucketTargetOption...)]
```
lesv: This looks like a different tag system than what we use -- Is this for c.g.c?   Or somewhere else?

lesv: If you'd like to fix based on what I did below, great, or I can comment on the whole file.

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
mziccard: Is there a reason for this?

gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
mziccard: > This looks like a different tag system

I didn't know any "standardized" tag system existed. Any reference to it?
I will change `\\ [END createBlob]` and `\\ [START createBlob]` according to your suggestion.

The reason I added `// [SNIPPET TARGET ...`] which I will change to just `// [TARGET ...]` is that in Java we want to keep in our codebase the existing javadoc and enrich it with snippets, to do that we need to add this snippets on top of each method's javadoc (i.e. we don't just want to create a new "custom" body of docs). To achieve this I need to know on top of which method to put the snippet.

mziccard: Also, added tags should not represent a problem right? You can skip them as long as you don't need them.

gcloud-java-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -1213,6 +1213,12 @@ public static Builder builder() {
   /**
    * Creates a new bucket.
    *
+   * <p>Example of creating a bucket.
```
lesv: Are these examples that your script will replace with tested code?

lesv:
I need to spend a bit more time here, but it appears you are also doing substitution with your script.  Are you inserting into the JavaDocs?  If so, why do you have hardcoded examples still?

lesv:
And, YES - this is moving in the right direction for making me happy.

gcloud-java-storage/src/main/java/com/google/cloud/storage/Storage.java
mziccard: No, these are the snippets in `StorageSnippets` after they are inserted into the javadoc using the `add_snippets_to_file.py` script. The reason for this is that I want the javadoc that we check in into the repository to be complete, not missing any part. Hence, running the `add_snippets_to_file.py` script is a pre-commit action.

This can be changed, but I would prefer to have complete javadoc in the source code for now.

I believe this also answers your following question.

lesv:
Also, is this stuff running on Jenkins somewhere?  If so, send me some info to my email.

mziccard:
> Also, is this stuff running on Jenkins somewhere? If so, send me some info to my email.

What do you mean with this stuff? We use [Travis](https://travis-ci.org/GoogleCloudPlatform/gcloud-java) but as I said in a previous commit, the `add_snippets_to_file.py` is now meant to be a pre-commit action.

lesv:
@mziccard I sent you an internal email, if you don't use that often, could you reply with the correct email.

lesv:
@mziccard I'm still reading your python script, but what I don't see is how it lets you fix a snippet and update the code.  (ie. I see it written as a one time thing).  If you passed a working directory, and wrote a new file that we generated the JavaDocs from, but never checked in, it would make sense.  (or am I just missing something?)

mziccard:
> I don't see is how it lets you fix a snippet and update the code.

You are right, the script doesn't do that yet. As you said it's a one-time thing, if you want to update the snippets you need to remove them from the javadoc by hand. I know this is **very** suboptimal. I didn't mean this PR to be a "silver bullet" so there's ton of room for improvements. If you see this feature as a blocker I'll work on that immediately. Otherwise I encourage you to open an issue. I am fine either way.

lesv:
I hate to make work for anyone, but it seems like it makes more sense to run this just before release rather than as a pre-commit action.

`add_snippets_to_file.py <snippetPath> <classPath> <genSrcPath>`
followed by
`javadoc <genSrcPath> -d <targetDir>`

The first bug fix will require that you manually delete all the code from the source, and that will be a pain .  Datastore needs the Async API's soon, that will require redoing all this stuff.

mziccard:
> I hate to make work for anyone, but it seems like it makes more sense to run this just before release rather than as a pre-commit action.

I can change the script to automatically delete the snippets from the javadoc first (if any) and then write the new ones. That way the script can still be used as a pre-commit action (and we have complete javadoc + snippets in the repo) as well as post-release action (when/if we decide to use it as post-release action).

lesv:
Totally happy with that, I was thinking I was suggesting a simpler process for you.  (Not sure how you find the snippets in the code, especially if they have changed.

mziccard:
@lesv I updated the script to delete snippets first and the add new ones, PTAL

lesv:
LGTM - #1190 opened.

mziccard:
@lesv thanks for the precious help on this!

PR #1147, opened 2016-08-10
Set user agent for Logging client

mziccard:


mziccard:
/cc @lesv 

Codacy seems to be stuck on this (even though it reports failure).

mziccard:
Update: Codacy is now done and happy

gcloud-java-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java
```diff
@@ -97,13 +99,18 @@ private InternalLoggingOptions(LoggingOptions options) {
   public DefaultLoggingRpc(LoggingOptions options) throws IOException {
     executorFactory = new InternalLoggingOptions(options).executorFactory();
     executor = executorFactory.get();
+    String libraryName = options.libraryName();
```
lesv: Always gcloud-java from what I can see.  I guess that's a good thing.

lesv:
LGTM - I presume I'll be seeing this kind of CL for all rpc's.

mziccard:
> LGTM - I presume I'll be seeing this kind of CL for all rpc's.

User agent is all set for other rpc's. Just logging was left behind.

PR #1149, opened 2016-08-10
Change pubsub endpoint from pubsub-experimental to pubsub

mziccard:
This change will be also made at the Veneer Toolkit level at next code-generation round. While we wait for that we go ahead and override the endpoint.

For reference: `pubsub-experimental.googleapis.com` stopped working today with the following error:

> com.google.cloud.pubsub.PubSubException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Google Cloud Pub/Sub API (Experimental) has not been used in project google.com:gcloud-devel before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/pubsub-experimental.googleapis.com/overview?project=gcloud-devel then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.

See [this comment](https://github.com/GoogleCloudPlatform/gcloud-java/issues/381#issuecomment-239001059) and [this StackOverflow question](http://stackoverflow.com/questions/38860888/gloud-java-permission-denied-google-cloud-pub-sub-api-experimental).

### DIFF
--- a/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
+++ b/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
@@ -18,6 +18,7 @@ package com.google.cloud.pubsub.spi;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
 
+import com.google.api.gax.core.ConnectionSettings;
 import com.google.api.gax.core.RetrySettings;
 import com.google.api.gax.grpc.ApiCallSettings;
 import com.google.api.gax.grpc.ApiException;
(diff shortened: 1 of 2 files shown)


mziccard:
Please @lesv can you have a look at this?

mziccard:
@garrettjonesgoogle would you also mind having a quick look?

garrettjonesgoogle:
Why not just edit the line in PublisherSettings & SubscriberSettings with the default host?

mziccard:
> Why not just edit the line in PublisherSettings & SubscriberSettings with the default host? 

This way we also support setting other hosts using the `host()` setter in the option builder (even though it should not be needed). And I wanted to avoid conflicts with the autogen code.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
```diff
@@ -139,10 +140,20 @@ public DefaultPubSubRpc(PubSubOptions options) throws IOException {
...
-            credentials.createScoped(PublisherSettings.DEFAULT_SERVICE_SCOPES));
-        subBuilder.provideChannelWith(
-            credentials.createScoped(SubscriberSettings.DEFAULT_SERVICE_SCOPES));
+        ConnectionSettings pubConnectionSettings = ConnectionSettings.newBuilder()
+            .setServiceAddress(options.host())
+            .setPort(PublisherSettings.DEFAULT_SERVICE_PORT)
+            .provideCredentialsWith(
+                credentials.createScoped(PublisherSettings.DEFAULT_SERVICE_SCOPES))
+            .build();
+        ConnectionSettings subConnectionSettings = ConnectionSettings.newBuilder()
```
shinfan: It would make the code more readable to use a helper here. The initialization code for pub and sub are mostly identical.

garrettjonesgoogle:
>> Why not just edit the line in PublisherSettings & SubscriberSettings with the default host? 

>This way we also support setting other hosts using the `host()` setter in the option builder (even though it should not be needed). And I wanted to avoid conflicts with the autogen code.

Ok it makes sense to plumb through the host from the options. But, my preference would be for PubSubOptions.DEFAULT_HOST to be set to PublisherSettings.DEFAULT_SERVICE_ADDRESS. 

If you simply edit the DEFAULT_SERVICE_ADDRESS value e.g. at https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherSettings.java#L112 , then it should match what the generator will output in the next generation, and there won't be a conflict. 

That said, it's not a strong opinion, and we can edit the veneer code when we generate next time.

garrettjonesgoogle:
LGTM since this issue needs to be fixed urgently

teddziuba:
Confirmed this fixes #1150. I hand built `0.2.7-SNAPSHOT` and deployed it in my app. Looking forward to the next release.

trentonstrong:
Thanks for fixing this so quick! Appreciate it.

PR #1154, opened 2016-08-11
Fetch latest changes from master and bump translate version

mziccard:
This PR fetches latest changes from master (required to make integration tests work). This also updates translate version to `0.2.8-SNAPSHOT`.

### DIFF
--- a/gcloud-java-translate/pom.xml
+++ b/gcloud-java-translate/pom.xml
@@ -11,7 +11,7 @@
   <parent>
     <groupId>com.google.cloud</groupId>
     <artifactId>gcloud-java-pom</artifactId>
-    <version>0.2.7-SNAPSHOT</version>
+    <version>0.2.8-SNAPSHOT</version>
   </parent>
   <properties>
     <site.installationModule>gcloud-java-translate</site.installationModule>


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
```diff
@@ -31,7 +31,7 @@
   private static final long serialVersionUID = 5640180400046623305L;
   private static final String PUBSUB_SCOPE = "https://www.googleapis.com/auth/pubsub";
   private static final Set<String> SCOPES = ImmutableSet.of(PUBSUB_SCOPE);
-  private static final String DEFAULT_HOST = "https://pubsub.googleapis.com";
+  private static final String DEFAULT_HOST = "pubsub.googleapis.com";
```
lesv: Why did you drop the https://?

lesv:
One Q, otherwise LGTM

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/PubSubOptions.java
mziccard: I dropped this as part of #1149. Pub/Sub is over gRPC so no need for `https://`.

PR #1158, opened 2016-08-12
Add Translate.listSupportedLanguages method and tests

mziccard:


lesv:
LGTM

lesv:
Why Logging?  Isn't this Translation?

mziccard:
> Why Logging? Isn't this Translation?

Yep sorry

PR #1163, opened 2016-08-15
Refactor API call and connection settings for gRPC services

mziccard:
This also fixes some style errors.

### DIFF
--- a/gcloud-java-core/src/main/java/com/google/cloud/ByteArray.java
+++ b/gcloud-java-core/src/main/java/com/google/cloud/ByteArray.java
@@ -135,7 +135,7 @@ public class ByteArray implements Iterable<Byte>, Serializable {
   /**
    * Creates a {@code ByteArray} object given an array of bytes. The bytes are copied.
    */
-  public final static ByteArray copyFrom(byte[] bytes) {
+  public static final ByteArray copyFrom(byte[] bytes) {
     return new ByteArray(ByteString.copyFrom(bytes));
   }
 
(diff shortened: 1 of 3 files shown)


mziccard:
@garrettjonesgoogle @shinfan this should better fit with you previous comments

garrettjonesgoogle:
LGTM - I'd also like to see what @shinfan thinks.

mziccard:
@shinfan any comment on this?

gcloud-java-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java
```diff
@@ -94,10 +90,21 @@ private InternalLoggingOptions(LoggingOptions options) {
     protected ExecutorFactory<ScheduledExecutorService> executorFactory() {
       return super.executorFactory();
     }
+
+    @Override
+    protected ApiCallSettings.Builder apiCallSettings() {
```
shinfan: Is it necessary to override those methods? I think calling the super method is the default behavior.

mziccard: Methods are protected and here we are in a different package. We do this to avoid exposing these methods to the user.

shinfan: Ummm I don't understand. Isn't InternalLoggingOptions is a private class here? Why would any method under a private class exposed to the user?

mziccard: `InternalLoggingOptions` extends `LoggingOptions` so that `protected` methods inherited from `LoggingOptions` can be made accessible in a different package (as `DefaultLoggingRpc` is in `com.google.cloud.logging.spi` while `LoggingOptions` is in `com.google.cloud.logging`). The alternative to this would be much such methods in `LoggingOptions` `public`, which we don't want to do (yet).

shinfan: Okay I see now it makes sense.

shinfan:
LGTM

PR #1168, opened 2016-08-17
Use Futures.transform instead of lazyTransform when possible

### LINKED ISSUE #1148: request: return ListenableFutures in PubSub async methods rather than bare Futures
The com.google.cloud.pubsub.PubSub interface defines async operations which return java.util.concurrent.Future values. 

For anyone who uses the ListenableFuture concept from Guava (or CompletableFutures in Java 8), these [Futures have to be adapted into ListenableFutures by spawning a thread that blocks on the `Future.get()` return](https://github.com/google/guava/blob/8613c4dbde2aaeb871dce3ea0e52811039ee203d/guava/src/com/google/common/util/concurrent/JdkFutureAdapters.java#L40-L58).

It would be much more convenient if the Futures returned from PubSub were actually ListenableFuture instances (even if the interface did not define the return value to be ListenableFuture) - then it would not be necessary for users to block one thread per future they wish to adapt into a ListenableFuture.

For instance, internally `PubSubImpl.pullAsync(final String subscription, int maxMessages)` is implemented by using the listenable-like `com.google.cloud.pubsub.spi.PubSubRpc.PullFuture` and attaching callbacks to it, but the returned Future is an anonymous subclass of Future created by `Futures.lazyTransform(..)`.

Is it a conscious decision to avoid using ListenableFutures in the interfaces defined in `gcloud-java-pubsub` and other clients? 

It seems like the `com.google.cloud.pubsub.spi` layer has access to ListenableFuture instances via the generated grpc client code (like `subscriberApi.acknowledgeCallable().futureCall(request)`) but then these are discarded in the `PubSubImpl` layer.

mziccard:
This fixes #1148 

@mattnworb this should make you happy :) Any change you can give it a try?

### DIFF
--- a/gcloud-java-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/gcloud-java-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -24,7 +24,6 @@ import static com.google.cloud.logging.Logging.ListOption.OptionType.PAGE_TOKEN;
 import static com.google.cloud.logging.Logging.WriteOption.OptionType.LABELS;
 import static com.google.cloud.logging.Logging.WriteOption.OptionType.LOG_NAME;
 import static com.google.cloud.logging.Logging.WriteOption.OptionType.RESOURCE;
-import static com.google.common.util.concurrent.Futures.lazyTransform;
 
 import com.google.cloud.AsyncPage;
 import com.google.cloud.AsyncPageImpl;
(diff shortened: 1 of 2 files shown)


mattnworb:
Awesome, thanks so much @mziccard !

I built a 0.2.8-SNAPSHOT from your branch and ran it in one of our applications that calls `pubsub.publishAsync()`. After executing the code path that calls that method a bunch of times, I do not see any threads created via the ThreadFactory that JdkFutureAdapters uses (a thread named like `ListenableFutureAdapter-thread-%d`). When running with prior versions, I see a few threads like `ListenableFutureAdapter-thread-1`. 

thanks again!

lesv:
LGTM - I'm presuming some integration tests will run to verify this for logging.

Issue #1014, opened 2016-05-17
Update to the next datastore emulator script version


ajkannan:
This emulator uses the same URL structure as the Cloud Datastore.  Because `gcloud-java` depends on the [google-cloud-datastore](https://github.com/GoogleCloudPlatform/google-cloud-datastore/tree/master/java) client, we need to wait until the client updates to use a variable to decide when to append  "/datastore" for localhost URLs. (/cc @pcostell)

There are a couple related changes that need to be made:

* Update the version of the google cloud datastore client we depend on.
* The name of the script & folder will change from `gcd` to `cloud-datastore-emulator`.  We should make the necessary updates to accommodate this change.
* We should also change the logic that checks for a locally-installed emulator to check for major/minor version updates (but not build).  Right now, we do not check version (besides ensuring that the script is for v1beta3).
* Ensure the "gcd-emulator" version will be reported the same way as before by the gcloud SDK.

mziccard:
Fixed in #1169

PR #1169, opened 2016-08-18
Update datastore from v1beta3 to v1

### LINKED ISSUE #1165: Update gcloud-java/datastore to use the v1 API
The v1 API is now available.

The protos are identical (just a different package name). The only behavioral change is captured in the release notes:
https://cloud.google.com/datastore/release-notes

mziccard:
Changes are:
- Change package names
- Update LocalDatastoreHelper to use new emulator (cloud-datastore-emulator >= 1.2.0)
- Enable all tests in DatastoreTest

This fixes #1165 and #1014

### DIFF
--- a/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
+++ b/gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
@@ -66,16 +66,15 @@ import java.util.zip.ZipInputStream;
  */
 public class LocalDatastoreHelper {
   private static final Logger log = Logger.getLogger(LocalDatastoreHelper.class.getName());
-  private static final String GCD_VERSION = "v1beta3";
-  private static final String GCD_BUILD = "1.0.1";
+  private static final String GCD_VERSION = "1.2.0";
   private static final double DEFAULT_CONSISTENCY = 0.9;
-  private static final String GCD_BASENAME = "gcd-" + GCD_VERSION + "-" + GCD_BUILD;
+  private static final String GCD_BASENAME = "cloud-datastore-emulator-" + GCD_VERSION;
   private static final String GCD_FILENAME = GCD_BASENAME + ".zip";
-  private static final String MD5_CHECKSUM = "df876ba8f054d69acff30ec9540ec386";
+  private static final String MD5_CHECKSUM = "e68695ff005421ccb7144689d9633df1";
   private static final URL GCD_URL;
   private static final String GCLOUD = "gcloud";
   private static final Path INSTALLED_GCD_PATH;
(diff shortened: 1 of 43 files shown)


mziccard:
Putting this on hold until the emulator for `v1` works on Java7 again.

mziccard:
@lesv PTAL

gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
```diff
@@ -66,16 +66,15 @@
  */
 public class LocalDatastoreHelper {
   private static final Logger log = Logger.getLogger(LocalDatastoreHelper.class.getName());
-  private static final String GCD_VERSION = "v1beta3";
-  private static final String GCD_BUILD = "1.0.1";
+  private static final String GCD_VERSION = "1.2.0";
```
lesv: Why not 1.1?

gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
```diff
@@ -218,16 +224,10 @@ public static ProcessStreamReader start(InputStream inputStream) {
     private boolean collectionMode;
     private volatile boolean terminated;
 
-    ProcessErrorStreamReader(InputStream errorStream, String blockUntil) throws IOException {
+    ProcessErrorStreamReader(InputStream errorStream) {
```
lesv: Why did the API change?  (blockUntil)?

lesv:
LGTM - my questions appear to be trivial.  I would have expected to see a text version of the API change somewhere.  GitHub's tools are a bit inadequate to answer my only question, did you miss something? What I've seen looks good.  So, unless you think one of your answers would cause me to raise an object, it's ok to merge.

gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
mziccard: `1.1` does not work with version `v1` of the API.

gcloud-java-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
mziccard: This is protected stuff so I would't refer to it as API (it's not exposed to the users). Anyway, this changed because older versions of the emulator used to print a ready message to the error stream. Newer versions output the ready message to the standard stream.

lesv:
LGTM

PR #1173, opened 2016-08-18
Removing auto-generated baseline files

garrettjonesgoogle:
Since we have removed the messaging about hand-editing the
auto-generated files, keeping baseline files doesn't really make sense.


mziccard:
LGTM, thanks @garrettjonesgoogle

PR #1174, opened 2016-08-18
Initial auto-generated gcloud-java-vision

garrettjonesgoogle:
FYI @michaelbausor @shinfan

### DIFF
--- a/gcloud-java-core/pom.xml
+++ b/gcloud-java-core/pom.xml
@@ -106,7 +106,7 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax</artifactId>
-      <version>0.0.15</version>
+      <version>0.0.16</version>
     </dependency>
     <dependency>
       <groupId>com.google.api.grpc</groupId>
--- a/gcloud-java-vision/src/test/java/com/google/cloud/vision/spi/v1/ImageAnnotatorTest.java
+++ b/gcloud-java-vision/src/test/java/com/google/cloud/vision/spi/v1/ImageAnnotatorTest.java
@@ -24,23 +24,26 @@ import com.google.common.collect.Lists;
 import com.google.protobuf.GeneratedMessage;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
-import junit.framework.Assert;
 import org.junit.After;
 import org.junit.AfterClass;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 @javax.annotation.Generated("by GAPIC")
 public class ImageAnnotatorTest {
+  private static MockImageAnnotator mockImageAnnotator;
   private static MockServiceHelper serviceHelper;
   private ImageAnnotatorApi api;
 
   @BeforeClass
   public static void startStaticServer() {
-    MockImageAnnotator mockService = new MockImageAnnotator();
-    serviceHelper = new MockServiceHelper("in-process-1", mockService);
+    mockImageAnnotator = new MockImageAnnotator();
+    serviceHelper =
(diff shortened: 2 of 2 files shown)


gcloud-java-vision/src/test/java/com/google/cloud/vision/spi/v1/ImageAnnotatorTest.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.vision.spi.v1;
+
+import com.google.api.gax.core.PageAccessor;
```
garrettjonesgoogle: @shinfan FYI there are several unused imports in the auto-generated test code

shinfan: Right now all these import are statically determined so depending on the tests some of them may be unused. We could either dynamically determine imports during generation or add a post-processor to remove unused ones.

gcloud-java-core/pom.xml
```diff
@@ -106,7 +106,7 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax</artifactId>
-      <version>0.0.14</version>
+      <version>0.0.15</version>
```
mziccard: Build is failing as this version of gax does not exist. Did you release it?

garrettjonesgoogle: I had assumed it was released, but I was wrong. It is released now.

gcloud-java-vision/src/test/java/com/google/cloud/vision/spi/v1/ImageAnnotatorTest.java
mziccard: This should be done, there's a big amount of used imports in all generated code.

gcloud-java-vision/src/main/java/com/google/cloud/vision/spi/v1/ImageAnnotatorApi.java
```diff
@@ -0,0 +1,222 @@
...
+   *     .addAllRequests(requests)
+   *     .build();
+   *   BatchAnnotateImagesResponse response = imageAnnotatorApi.batchAnnotateImages(request);
+   * }
+   * </code></pre>
+   *
+   * @param request The request object containing all of the parameters for the API call.
+   * @throws com.google.api.gax.grpc.ApiException if the remote call fails
+   */
+  private final BatchAnnotateImagesResponse batchAnnotateImages(
```
mziccard: Is this meant to be private?

garrettjonesgoogle: Yes, our code generator automatically hides the request-based methods when the request has only one field, so the flattened version is more direct. Will this cause a problem? We can re-evaluate.

mziccard: No this is fine. Just wanted to be sure.

gcloud-java-vision/src/test/java/com/google/cloud/vision/spi/v1/ImageAnnotatorTest.java
garrettjonesgoogle: We will try to figure out a way to squash them all, but probably after this PR.

garrettjonesgoogle:
This has been rebased, PTAL

PR #1178, opened 2016-08-19
Initial auto-generated gcloud-java-monitoring

garrettjonesgoogle:


lesv:
@mziccard are you going to this, or would you rather I did it?  It would be helpful if you or @aozarov reviewed at least one of these.

mziccard:
@lesv This is on me, I'll handle these PRs tomorrow

garrettjonesgoogle:
FYI I didn't get a chance to update this particular PR - don't try merging it yet

garrettjonesgoogle:
It turns out that the Monitoring surface uses the "oneof" feature of proto, which Veneer Toolkit doesn't support yet:

https://github.com/googleapis/googleapis/blob/master/google/monitoring/v3/group_service.proto#L82

This means that the listGroups API method is broken. It will only work if you set only one value among the three parameters childrenOfGroup, ancestorsOfGroup, and descendantsOfGroup; if you set multiple, then the last one will win, without giving any indication of what happened. 

We will need to figure out how this surface should look before we can deliver this. Don't let this issue block the other APIs, though.

mziccard:
> We will need to figure out how this surface should look before we can deliver this. Don't let this issue block the other APIs, though.

Fine with me, I suggest we either close this or add the don't merge label until this issue is sorted out.

garrettjonesgoogle:
I will create a new PR after the rename and after the oneof issue is sorted out.

PR #1181, opened 2016-08-22
Add integration, serialization and option-related tests for Translate

mziccard:
- Change `TranslateOptions` to read API key from `GOOGLE_API_KEY` env variable
- Add `RemoteTranslateHelper` class and related javadoc
- Add `SerializationTest` class to test that objects can be serialized
- Add `TranslateTest` to test vararg options
- Add `ITTranslateTest` to run ntegration tests


lesv:
LGTM

Issue #1172, opened 2016-08-18
Update docs menu ordering to be consistent with other platforms


jskeet:
In all platforms other than Java, .NET is at the top. For instance, Python:

![python](https://cloud.githubusercontent.com/assets/17011/17777753/9605eb5e-6559-11e6-8d1f-4694389024d6.png)

Java has it at the bottom:

![java](https://cloud.githubusercontent.com/assets/17011/17777768/9e41c464-6559-11e6-95b4-1b2af27aba3a.png)

In order for this to feel like a single menu just changing the page, the order shouldn't change.

jgeewax:
And to avoid issues like this, we should try to move to the consolidated site code (gcloud-common/site)

mziccard:
This was fixed in #1184 and #1182.

> And to avoid issues like this, we should try to move to the consolidated site code (gcloud-common/site)

Is there a way I can start depending the app in `gcloud-common/site` without switching to the json doc format (yet)? Can I configure the manifest.json so that `Read the docs` still points to our javadoc?

PR #1182, opened 2016-08-22
Re-order landing page menu to put .NET first

mziccard:


mziccard:
This fixes #1172 for next site versions, while #1182 fixes it for the latest deployed version.

lesv:
LGTM -- at some point do you want to put this in [/docs/](https://github.com/blog/2228-simpler-github-pages-publishing) ?

mziccard:
> at some point do you want to put this in /docs/ ?

I believe that moving all content of `gh-pages` to /docs would increase too much the size of our master branch (given that we keep also older versions).

PR #1183, opened 2016-08-22
Remove unused directories and resources from gh-pages branch

mziccard:
This PR removes unused resources from the gh-pages branch.


lesv:
LGTM

PR #1184, opened 2016-08-22
Re-order landing page menu (put .NET first) in latest deployed site

### LINKED ISSUE #1172: Update docs menu ordering to be consistent with other platforms

mziccard:
This fixes #1172 for latest deployed site.


lesv:
LGTM

PR #1185, opened 2016-08-22
Initial auto-generated gcloud-java-speech

garrettjonesgoogle:


garrettjonesgoogle:
Apparently the v1 version of speech on googleapis is invalid. I will need to rescind this PR and generate a new one based off of v1beta1.

PR #1192, opened 2016-08-24
Update links to google-cloud-node and google-cloud-php landing pages

mziccard:


lesv:
LGTM - be prepared at some point they will rename us as well.

mziccard:
> be prepared at some point they will rename us as well.

I plan to do that by the end of the week :)

lesv:
So what they found on Node is that there will be broken links everywhere that don't get referrals.  We will also need to update docs on c.g.c & d.g.c once this happens.

PR #1194, opened 2016-08-24
Add support for Google Translate

mziccard:
Nothing new here, just merging the translate-alpha branch, all commits have already been reviewed.

/cc @lesv

### DIFF
--- a/gcloud-java-examples/pom.xml
+++ b/gcloud-java-examples/pom.xml
@@ -32,6 +32,12 @@
       <artifactId>maven-assembly-plugin</artifactId>
       <version>2.5.4</version>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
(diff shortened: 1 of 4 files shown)


lesv:
LGTM - once travis is happy.  Are there issues for the licenses that dependency-ci are seeing?

mziccard:
> Are there issues for the licenses that dependency-ci are seeing?

Dependency-ci is complaining  about `javax.inject` not being licenced, but that's not true so I guess it's just a bug on their side.

lesv:
Travis is using a very old version of Java8.

lesv:
I see a test for Listing country codes, they would like an actual example of it.

mziccard:
> I see a test for Listing country codes, they would like an actual example of it.

Yes there's a test (actually two) for listing supported languages. There are also snippets for that, see `listSupportedLanguages` and `listSupportedLanguagesWithTarget` in `DetectLanguageAndTranslate`.

lesv:
Cool - It's just a different name.

PR #1195, opened 2016-08-24
Fix flakiness of Logging integration tests

mziccard:
Logging integration tests started failing with the following errors:

> INVALID_ARGUMENT: Sinks with different output formats cannot share destination. There already is a project sink with V2 output format.

This PR fixes listing and update tests to make them more robust.


mziccard:
@lesv can you have a look please?

lesv:
LGTM

PR #1203, opened 2016-08-26
Move ITBucketSnippets and improve Storage snippets tests

mziccard:


gcloud-java-examples/src/main/java/com/google/cloud/examples/storage/snippets/BucketSnippets.java
```diff
@@ -110,15 +111,16 @@ public boolean delete() {
    * Example of listing the blobs in the bucket.
    */
   // [TARGET list(BlobListOption...)]
-  public Iterator<Blob> listBlobs() {
+  public Page<Blob> listBlobs() {
```
lesv: I obviously need to spend more time w/ our base layer - good change!

gcloud-java-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITBlobSnippets.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.cloud.example.storage.snippets;
+package com.google.cloud.examples.storage.snippets;
```
lesv: We've been trying to use com.example.blah (RFC 2606) for everything, so that folks can grab sample code directly.  Since these are snippets, it's probably Ok

lesv:
LGTM - I need tor read these closer.

gcloud-java-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITBlobSnippets.java
mziccard: > We've been trying to use com.example.blah (RFC 2606) for everything

Can you give me more details? Are you suggesting that the package should be named `com.example`? This sounds strange to me.

lesv: The repo uses this model, we should continue.  In general, for Java, we are following com.example.something model, so that if someone grabs our source file to start their own project, they are more likely to change it, than having a com.google on their own code.  What you are doing is fine.

PR #1205, opened 2016-08-26
Pub/Sub SPI: Adding IAM methods and auto-generated unit tests

garrettjonesgoogle:


gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -726,6 +741,230 @@ private final void deleteTopic(DeleteTopicRequest request) {
...
+   *
+   * @param resource REQUIRED: The resource for which policy is being specified.
+   * Resource is usually specified as a path, such as,
+   * projects/{project}/zones/{zone}/disks/{disk}.
+   * @param policy REQUIRED: The complete policy to be applied to the 'resource'. The size of
+   * the policy is limited to a few 10s of KB. An empty policy is in general a
+   * valid policy but certain services (like Projects) might reject them.
+   * @throws com.google.api.gax.grpc.ApiException if the remote call fails
+   */
+  public final Policy setIamPolicy(String resource, Policy policy) {
```
mziccard: Documentation here (and for other methods) is very vague IMHO. Also I believe that the parameter name `resource` should be called `topic`. Also the example for the `resource` parameter (`projects/{project}/zones/{zone}/disks/{disk}`) has nothing to do with Pub/Sub or topics and might be misleading.
I see that also [the official docs](https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions/setIamPolicy) are very vague, which makes me think that your hands are tied by auto-generation. It might be useful to raise this issue to the people writing the docs to make them more specific.

garrettjonesgoogle: You are correct, our hands are tied. Currently Pub/Sub is not declaring the IAM methods in the fully correct way. Once they do, the docs here can be improved by being made specific to topics and subscriptions. We have an open issue filed against them internally to change this.

gcloud-java-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherApi.java
```diff
@@ -16,6 +16,11 @@
 import com.google.api.gax.core.PageAccessor;
 import com.google.api.gax.grpc.ApiCallable;
 import com.google.api.gax.protobuf.PathTemplate;
+import com.google.iam.v1.GetIamPolicyRequest;
+import com.google.iam.v1.Policy;
+import com.google.iam.v1.SetIamPolicyRequest;
+import com.google.iam.v1.TestIamPermissionsRequest;
+import com.google.iam.v1.TestIamPermissionsResponse;
```
mziccard: All these seem common classes, used by all services. Why are they placed in the `grpc-google-pubsub-v1` package? It would be very nice if we could avoid this and maybe put these classes in `grpc-google-common-protos`. Reason is that we (`gcloud-java`) have a common `Policy` class in `gcloud-java-core` and I would prefer not to depend on `grpc-google-pubsub-v1` there.

garrettjonesgoogle: You're totally right. I will figure out a solution.

garrettjonesgoogle:
I will create a new PR after the rename is merged.

PR #1206, opened 2016-08-29
Enable appveyor for windows builds

mziccard:
This PR enables appveyor, so that our tests are run also on windows.


mziccard:
Putting this on hold until appveyor is enabled by the organization account.

lesv:
LGTM - looks like this helps immediately.

PR #1208, opened 2016-08-29
Enable travis osx build

mziccard:


lesv:
LGTM - BTW mvn 3.3.9 appears to no longer need JAVA_HOME defined, as long as your java install is on the classpath.

mziccard:
> BTW mvn 3.3.9 appears to no longer need JAVA_HOME defined, as long as your java install is on the classpath.

Yes but Travis's osx image is stuck to a previous version (3.2?), I subscribed to the issue [here](https://github.com/travis-ci/travis-ci/issues/2839) and will update as soon as it's fixed.

mziccard:
@jgeewax @omaray glad to announce that we have osx tests in place. Only java7 is supported at the time and the osx build job takes quite some time to start but it's still something :)

omaray:
Nice work Marco!!!

PR #1211, opened 2016-08-30
Initial auto-generated gcloud-java-speech

garrettjonesgoogle:


gcloud-java-speech/src/main/java/com/google/cloud/speech/spi/v1beta1/SpeechApi.java
```diff
@@ -0,0 +1,311 @@
...
+ * <code>
+ * SpeechSettings speechSettings = SpeechSettings.defaultBuilder()
+ *     .provideChannelWith(myCredentials)
+ *     .build();
+ * SpeechApi speechApi = SpeechApi.create(speechSettings);
+ * </code>
+ * </pre>
+ */
+@javax.annotation.Generated("by GAPIC")
+public class SpeechApi implements AutoCloseable {
```
mziccard: Just to be sure: why is there no `streamingRecognize` method?

garrettjonesgoogle: We specifically excluded it because we don't support it yet; the surface of streaming methods needs to be different. We are working on that separately.

mziccard: Do you suggest we merge this as is and add the streaming method later on? I am fine with this approach

garrettjonesgoogle: Yes, that was our plan.

PR #1217, opened 2016-08-31
Support listing/getting/creating/updating tables/datasets/jobs in other projects

### LINKED ISSUE #1123: BigQuery listTables(DatasetId) does not use project overrided in DatasetId
My Google Application Default Credentials are set to the swast-scratch project, but I wish to list the tables in the `publicdata` project's `samples` dataset. This is a public dataset.

```shell
$ bq ls publicdata:samples
      tableId       Type
 ----------------- -------
  github_nested     TABLE
  github_timeline   TABLE
  gsod              TABLE
  natality          TABLE
  shakespeare       TABLE
  trigrams          TABLE
  wikipedia         TABLE
```

Example code that displays this issue:

```
import com.google.cloud.bigquery.BigQuery;
import com.google.cloud.bigquery.BigQueryOptions;
import com.google.cloud.bigquery.DatasetId;
import com.google.cloud.bigquery.Table;
import java.io.PrintWriter;
import java.util.Iterator;

public class TableLister {
  public static void listTables(PrintWriter out) {
    BigQuery bigquery = BigQueryOptions.defaultInstance().service();
    Iterator<Table> tables = bigquery.listTables(DatasetId.of("publicdata", "samples")).iterateAll();
    while (tables.hasNext()) {
      out.println(tables.next());
    }
  }

    listTables(new PrintWriter(System.out));
  }
}
```

It fails with the error: `"message" : "Not found: Dataset swast-scratch:samples"`, so it appears that it is not using the project that I set in the DatasetId.

http://googlecloudplatform.github.io/gcloud-java/0.2.5/apidocs/com/google/cloud/bigquery/DatasetId.html#of-java.lang.String-java.lang.String-

mziccard:
This fixes #1123 

@tswast feel free to have a look and comment whether this meets your expectations.

/cc @lesv

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -48,9 +48,11 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
     private static final long serialVersionUID = -3057564042439021278L;
     private final Map<BigQueryRpc.Option, ?> requestOptions;
     private final BigQueryOptions serviceOptions;
+    private final String projectId;
 
-    DatasetPageFetcher(BigQueryOptions serviceOptions, String cursor,
+    DatasetPageFetcher(String projectId, BigQueryOptions serviceOptions, String cursor,
         Map<BigQueryRpc.Option, ?> optionMap) {
+      this.projectId = projectId;
       this.requestOptions =
           PageImpl.nextRequestOptions(BigQueryRpc.Option.PAGE_TOKEN, cursor, optionMap);
       this.serviceOptions = serviceOptions;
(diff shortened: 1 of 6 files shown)


lesv:
LGTM

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -265,13 +274,15 @@ public boolean delete(String datasetId, DatasetDeleteOption... options) {
   }
 
   @Override
-  public boolean delete(final DatasetId datasetId, DatasetDeleteOption... options) {
+  public boolean delete(DatasetId datasetId, DatasetDeleteOption... options) {
+    final DatasetId completeDatasetId = datasetId.setProjectId(options().projectId());
```
tswast: I could see this as possibly causing the same confusion if someone populates the projectId of the DatasetId. (It's certainly less likely that someone would try it, though.)

Maybe raise an error if the projectId is already populated and set to something that doesn't match options().projectId()?

mziccard: `setProjectId` does not override the value if it's already set, it just set if the user did not provide one (is `null`).

tswast: Ah, cool, thanks.

tswast:
LGTM

PR #1219, opened 2016-08-31
Pub/Sub SPI: Adding IAM methods and auto-generated unit tests

garrettjonesgoogle:
Note: grpc-google-pubsub-v1 depends on a new package grpc-google-iam-v1, which contains IAMPolicy and Policy.

### DIFF
--- a/google-cloud-core/pom.xml
+++ b/google-cloud-core/pom.xml
@@ -111,7 +111,7 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-common-protos</artifactId>
-      <version>0.0.7</version>
+      <version>0.0.9</version>
     </dependency>
   </dependencies>
 </project>
--- a/google-cloud-pubsub/pom.xml
+++ b/google-cloud-pubsub/pom.xml
@@ -30,7 +30,7 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-pubsub-v1</artifactId>
-      <version>0.0.7</version>
+      <version>0.0.9</version>
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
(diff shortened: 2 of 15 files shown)


mziccard:
A couple of questions on dependencies:

- `grpc-google-iam-v1` contains iam classes that will be used by all services, right?
- Why does `grpc-google-common-protos` depend on `protobuf-java 3.0.0-beta-3` instead of the latest `protobuf-java 3.0.0`?

garrettjonesgoogle:
* yes, grpc-google-iam-v1 will be shared by multiple services.
* I haven't updated the protobuf-java dependency in order to isolate changes. I will do a batch upgrade of grpc to 1.0.0 and protobuf to 3.0.0 in one go, without other changes at the same time.

mziccard:
@garrettjonesgoogle another general question: is there a real need for having a copy of `setIamPolicy`, `setIamPolicy` and `testIamPermissions` in both `PublisherApi` and `SubscriberApi`? It seems like a lot of duplicated code to me. Will other services use/require this approach as well? Did you consider putting the three methods in a dedicated `PubSubIamApi` class?

garrettjonesgoogle:
It's the general pattern for IAM. In services that follow the suggested pattern (see https://github.com/googleapis/googleapis/blob/master/google/genomics/v1/datasets.proto ), the methods are redeclared in each service declaration. This allows their documentation and http paths to be customized. Pub/Sub doesn't follow the suggested pattern in the proto declarations, although it shows up the correct way in the http view (see https://developers.google.com/apis-explorer/#p/pubsub/v1/ ).

mziccard:
Will setting the policy for a topic resource via the `SubscriberApi` (or for a subscription via the `PublisherApi`) fail? Or is this just for having specific docs and http endpoints?

garrettjonesgoogle:
The way it's set up currently, it will fail validation when the flattened methods are used, but the request-based methods would not fail. The http endpoints actually have zero impact on the gRPC call; we just use the http endpoint information to inform our codegen (we derive the validation from it).

mziccard:
I am closing this, rebased on top of master in #1229

PR #1220, opened 2016-08-31
Initial auto-generated google-cloud-monitoring

garrettjonesgoogle:
I removed the flattened method (listGroups) that was causing a problem because of the oneof feature in Proto. The surface is still complete since the request-based method variant and callable method variant are still present.


garrettjonesgoogle:
Any chance this could get merged soon? We would like to do some auto-generation updates across all of the APIs, and it would be less noisy for this to ride along with the rest.

PR #1221, opened 2016-09-01
Fix Metric.update and add snippets for Metric class

mziccard:
This PR does the following:
- Fix `Metric.update`: remove unnecessary MetricInfo parameter for Metric.update(Async)
- Update snippets script to support method parameters with the `final` modifier
- Add snippets to Metric's javadoc, `MetricSnippets` class and tests

### DIFF
--- a/google-cloud-examples/src/test/java/com/google/cloud/examples/logging/snippets/ITLoggingSnippets.java
+++ b/google-cloud-examples/src/test/java/com/google/cloud/examples/logging/snippets/ITLoggingSnippets.java
@@ -29,6 +29,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Sets;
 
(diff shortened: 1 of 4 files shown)


google-cloud-examples/src/test/java/com/google/cloud/examples/logging/snippets/ITLoggingSnippets.java
```diff
@@ -45,11 +46,11 @@
   private static final String DATASET = "dataset";
   private static final Set<String> DESCRIPTOR_TYPES = ImmutableSet.of("gce_instance", "gae_app",
```
lesv: There is probably nothing that can be done that is correct, but it would seem like this would be ideal at the start of the snippet rather than in the tests.

lesv:
LGTM - I don't think my comment is actionable in a way that makes sense.

PR #1222, opened 2016-09-01
Fix Sink.update/delete and add snippets for Sink class

mziccard:
This PR does the following:
- Remove unnecessary `SinkInfo` parameter for `Sink.update` and `Sink.updateAsync`
- Fix `deleteSink` to properly return `false` on `NOT_FOUND`
- Add snippets to `Sink`'s javadoc, `SinkSnippets` class and tests


lesv:
LGTM

PR #1224, opened 2016-09-02
Refactor IAM-related classes to be ready to support more services

mziccard:
This PR does the following:
- Remove resourcemanager's `Policy` class
- Rename core's `IamPolicy` to `Policy`
- Add `Role` class to hold string values for roles
- Add `Policy.Marshaller` and `Policy.DefaultMarshaller` classes to convert `Policy` to/from gRPC protos
- Add `PolicyMarshaller` to resourcemanager to convert `Policy` to/from resourcemanager's protos
- Update READMEs, javadoc and examples

@aozarov @ajkannan do you mind having a look? Also, should we add an `iam` package?

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/Role.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/Role.java
@@ -41,8 +41,8 @@ public final class Role implements Serializable {
   }
 
   /**
-   * Returns the string identifier for this role. For example, {@code roles/viewer},
-   * {@code roles/editor}, or {@code roles/owner}.
+   * Returns the string identifier for this role. For example, {@code "roles/viewer"},
+   * {@code "roles/editor"}, or {@code "roles/owner"}.
    */
   public String value() {
     return value;
@@ -81,9 +81,9 @@ public final class Role implements Serializable {
   /**
    * Returns a new role given its string value.
    *
-   * @param value the string value for the role, for example, {@code roles/viewer},
-   *     {@code roles/editor}, or {@code roles/owner}. If this value does not start with the role
-   *     prefix {@code roles/}, the prefix is prepended.
+   * @param value the string value for the role, for example, {@code "roles/viewer"},
+   *     {@code "roles/editor"}, or {@code "roles/owner"}. If this value does not start with the
+   *     role prefix {@code roles/}, the prefix is prepended.
    * @see <a href="https://cloud.google.com/iam/docs/viewing-grantable-roles">Viewing the Grantable
    *     Roles on Resources</a>
    */


google-cloud-core/src/main/java/com/google/cloud/Policy.java
```diff
@@ -37,69 +44,127 @@
  * a set of identities to a role, where the identities can be user accounts, Google groups, Google
```
aozarov: No longer a "Base class"

google-cloud-core/src/main/java/com/google/cloud/Policy.java
```diff
@@ -37,69 +44,127 @@
...
 
-  private final Map<R, Set<Identity>> bindings;
+  private final Map<Role, Set<Identity>> bindings;
   private final String etag;
-  private final Integer version;
+  private final int version;
+
+  public abstract static class Marshaller<T> {
+
+    protected static final Function<String, Identity> IDENTITY_VALUE_OF_FUNCTION =
```
aozarov: Nit, replace `protected` with package scope as this class is final.

aozarov: Ignore the above comment.

google-cloud-core/src/main/java/com/google/cloud/Policy.java
```diff
@@ -37,69 +44,127 @@
...
+    protected Policy fromPb(com.google.iam.v1.Policy policyPb) {
+      Map<Role, Set<Identity>> bindings = new HashMap<>();
+      for (com.google.iam.v1.Binding bindingPb : policyPb.getBindingsList()) {
+        bindings.put(Role.of(bindingPb.getRole()),
+            ImmutableSet.copyOf(
+                Lists.transform(bindingPb.getMembersList(), IDENTITY_VALUE_OF_FUNCTION)));
+      }
+      return builder()
+          .bindings(bindings)
+          .etag(policyPb.getEtag().size() == 0 ? null : policyPb.getEtag().toStringUtf8())
```
aozarov: Nit, https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/ByteString#isEmpty--

google-cloud-core/src/main/java/com/google/cloud/Policy.java
```diff
@@ -208,27 +275,42 @@ public String etag() {
    * Sets the version of the policy. The default version is 0, meaning only the "owner", "editor",
```
aozarov: Gets

google-cloud-resourcemanager/README.md
```diff
@@ -186,11 +186,7 @@ Policy policy = project.getPolicy();
 // Add a viewer
 Policy.Builder modifiedPolicy = policy.toBuilder();
 Identity newViewer = Identity.user("<insert user's email address here>");
-if (policy.bindings().containsKey(Role.viewer())) {
-  modifiedPolicy.addIdentity(Role.viewer(), newViewer);
-} else {
-  modifiedPolicy.addBinding(Role.viewer(), newViewer);
-}
+modifiedPolicy.addIdentity(Role.of("roles/viewer"), newViewer);
```
aozarov: Nit, for a snippet I would prefer the shorter/simpler version of `Role.of("viewer")`

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/ResourceManager.java
```diff
@@ -302,8 +302,9 @@ public static ProjectListOption fields(ProjectField... fields) {
    * <p>An example of using the read-write-modify pattern is as follows:
    * <pre> {@code
    * Policy currentPolicy = resourceManager.getPolicy("my-project-id");
-   * Policy modifiedPolicy =
-   *     current.toBuilder().removeIdentity(Role.VIEWER, Identity.user("user@gmail.com"));
+   * Policy modifiedPolicy = current.toBuilder()
+   *     .removeIdentity(Role.of("roles/viewer"), Identity.user("user@gmail.com"))
```
aozarov: Same here, prefer shorter version. Also, If I read it correctly and the notion of ["Primitive roles"](https://cloud.google.com/iam/docs/understanding-roles) apply to all, I highly suggest creating shortcut methods for them such as `Role.viewer()`, ...

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/PolicyMarshaller.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Set;
+
+final class PolicyMarshaller
+    extends Marshaller<com.google.api.services.cloudresourcemanager.model.Policy> {
+
+  static final PolicyMarshaller INSTANCE = new PolicyMarshaller();
+
+  private PolicyMarshaller() {}
+
+  private static class Builder extends com.google.cloud.Policy.Builder {
```
aozarov: I am not sure this need justify creating a subclass of `com.google.cloud.Policy.Builder` rather than setting the value directly on `Policy.builder()` (and in that case making that Builder final instead of some of its methods).

google-cloud-core/src/main/java/com/google/cloud/Role.java
```diff
@@ -0,0 +1,78 @@
...
+   */
+  public String value() {
+    return value;
+  }
+
+  /**
+   * Returns a new role given its string value.
+   *
+   * @param value the string value for the role, for example, {@code roles/viewer},
+   *     {@code roles/editor}, or {@code roles/owner}. If this value does not start with the role
```
aozarov: should the string value examples be quoted?

aozarov:
Looks good (only skimmed the tests though). Few comments/suggestions.

I am not convinced that we should put the Policy under iam sub-package unless we expect to have more kinds of policies in the future but fine either way.

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/ResourceManager.java
ajkannan: +1 for adding Role.viewer(), etc

google-cloud-core/src/main/java/com/google/cloud/Policy.java
ajkannan: I think this is supposed to stay "sets" since it's the builder's setter method, right?

Just FYI (no changes necessary), according to my (quick/limited) testing this morning, it seems like setting this (at least in cloud resource manager) has no effect.  I think they changed it recently to be read-only.  Unsure about IAM policies in general though.

ajkannan:
LGTM

google-cloud-core/src/main/java/com/google/cloud/Policy.java
aozarov: I meant at line 275 but folding and the inability to comment on parts that were not changed got me here.... :)

google-cloud-core/src/main/java/com/google/cloud/Role.java
mziccard: Yes I'll add quotes.

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/PolicyMarshaller.java
mziccard: To be able to remove this class we must make `Policy.Builder`'s `version()` and `etag` setters public (they are protected now). Can we leave with that?

aozarov: Ah, I see. I rather keep it as is then (and have theses methods protected).

Another option, though a bit awkward, is to consider resource manager (and api-client based api) as an exception to the gRPC case and having them translate from their generated API Policy class to `com.google.iam.v1.Policy` and then create `Policy` using one and only one DefaultMarshaller (though at that point should be called Default). 

Otherwise, fine as is.

google-cloud-core/src/main/java/com/google/cloud/Policy.java
mziccard: Fixed

google-cloud-core/src/main/java/com/google/cloud/Policy.java
mziccard: Fixed

google-cloud-core/src/main/java/com/google/cloud/Policy.java
mziccard: Fixed

google-cloud-resourcemanager/README.md
mziccard: Fixed

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/ResourceManager.java
mziccard: Fixed

mziccard:
Comments fixed, PTAL

> I am not convinced that we should put the Policy under iam sub-package unless we expect to have more kinds of policies in the future but fine either way.

I was considering to put: `Policy`, `Role` and `Identity` under the `iam` package but I am not convinced either.

google-cloud-core/src/main/java/com/google/cloud/Role.java
aozarov: I didn't see the change (which also could be applied in other places).

aozarov:
LGTM (one comment was not addressed, your call).

mziccard:
@aozarov Another thing we might want to discuss is wether to store `etag` as `byte[]` or `String`. gRPC stores it as a `ByteString` (immutable byte array), we currently store it as ` String`, do you have any preferences?

aozarov:
My preference would be that as long as we wrap (and not expose) the underlying protobuf we should return it as String (which is also what we did for the api-client based Services).

PR #1226, opened 2016-09-06
Add integration tests and remote test helper for Datastore

mziccard:


lesv:
LGTM - However, while in practice, this shouldn't happen, if you were to run this against the Datastore Emulator, you would find that datastore can be flakey.  (as consistency is by default 0.9)  I don't know if you'd want to put in code to address this, but I thought I'd mention it.  (Note - I have seen this in production as well, but it's been a while).

mziccard:
> you would find that datastore can be flakey

This is a good point. Queries can be eventually consistent in datastore, I'll add some come to handle such cases.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+    assertNull(list.get(0));
+    assertEquals(entity2, list.get(1));
+    assertEquals(ENTITY3, list.get(2));
+    assertEquals(3, list.size());
+
+    try {
+      transaction.commit();
+      fail("Expecting a failure");
+    } catch (DatastoreException ex) {
+      // expected to fail
```
lesv: Should you check the exception?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+    }
+
+    try {
+      transaction.rollback();
+      fail("Expecting a failure");
+    } catch (DatastoreException ex) {
+      // expected to fail
```
lesv: check the exception?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+    transaction.update(Entity.builder(ENTITY2).clear().set("from", "transaction").build());
+    try {
+      transaction.commit();
+      fail("Expecting a failure");
+    } catch (DatastoreException expected) {
+      assertEquals("ABORTED", expected.reason());
+    }
+  }
+
+  @Test
```
lesv: Should there be an overall time limit for this to complete?  (I seem to recall a very long period of time, but it doesn't appear to be published anywhere anymore).

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+
+    Batch.Response response = batch.submit();
+    entities = DATASTORE.fetch(KEY1, KEY2, KEY3, entity4.key(), entity5.key(), entity6.key());
+    assertEquals(entity1, entities.get(0));
+    assertEquals(entity2, entities.get(1));
+    assertEquals(ENTITY3, entities.get(2));
+    assertEquals(entity4, entities.get(3));
+    assertEquals(entity5, entities.get(4));
+    assertEquals(entity6, entities.get(5));
+    assertEquals(6, entities.size());
```
lesv: Should this not be the first assert? (ie before `assertEquals(entity1...`)

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+    assertEquals(ENTITY1, entities.get(0));
+    assertEquals(ENTITY2, entities.get(1));
+    assertEquals(ENTITY3, entities.get(2));
+    assertNull(entities.get(3));
+    assertNull(entities.get(4));
+    assertNull(entities.get(5));
+    assertEquals(6, entities.size());
+  }
+
+  @Test
```
lesv: Total time for test?  65 seconds seems good, but might still fail.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+    }
+    keyProjectionResult = DATASTORE.run(keyProjectionQuery);
+    assertTrue(keyProjectionResult.hasNext());
+    ProjectionEntity projectionEntity = keyProjectionResult.next();
+    assertEquals(KEY1, projectionEntity.key());
+    assertTrue(projectionEntity.names().isEmpty());
+    assertFalse(keyProjectionResult.hasNext());
+  }
+
+  @Test
```
lesv: Limit test time?

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+    results2 = DATASTORE.run(query1);
+    assertSame(Entity.class, results2.resultClass());
+    @SuppressWarnings("unchecked")
+    QueryResults<Entity> results3 = (QueryResults<Entity>) results2;
+    assertTrue(results3.hasNext());
+    assertEquals(ENTITY1, results3.next());
+    assertFalse(results3.hasNext());
+  }
+
+  @Test
```
lesv: Limt test time

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
mziccard: This is there just to ensure that there are no more entities than expected. In case an entity is missing we'll fail earlier with an exception.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+        .build();
+    transaction.update(entity2);
+    transaction.delete(KEY1);
+    transaction.commit();
+
+    List<Entity> list = DATASTORE.fetch(KEY1, KEY2, KEY3);
+    assertNull(list.get(0));
+    assertEquals(entity2, list.get(1));
+    assertEquals(ENTITY3, list.get(2));
+    assertEquals(3, list.size());
```
mziccard: This is there just to ensure that there are no more entities than expected. In case an entity is missing we'll fail earlier with an exception.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -0,0 +1,670 @@
...
+
+    try {
+      transaction.commit();
+      fail("Expecting a failure");
+    } catch (DatastoreException ex) {
+      // expected to fail
+    }
+
+    try {
+      transaction.rollback();
```
mziccard: Added

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
mziccard: Added

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
mziccard: I added an overall limit to each test of 100 seconds. This should be more than enough.

lesv:
LGTM

PR #1235, opened 2016-09-09
Retry ABORTED Datastore commit only when NON_TRANSACTIONAL

### LINKED ISSUE #1230: We should not retry ABORTED transactions
In datastore we have `ABORTED` among retryable exceptions. This causes `commit` requests failed due to contention to be retried. Retrying an `ABORTED` transaction always fails with `INVALID ARGUMENT` error.

I suggest we never retry `ABORTED`. \cc @aozarov @ajkannan

mziccard:
Fixes #1230

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
@@ -34,7 +34,9 @@ public final class DatastoreException extends BaseServiceException {
 
   // see https://cloud.google.com/datastore/docs/concepts/errors#Error_Codes"
   private static final Set<Error> RETRYABLE_ERRORS = ImmutableSet.of(
-      new Error(10, "ABORTED"), new Error(4, "DEADLINE_EXCEEDED"), new Error(14, "UNAVAILABLE"));
+      new Error(10, "ABORTED", true),
+      new Error(4, "DEADLINE_EXCEEDED", false),
+      new Error(14, "UNAVAILABLE", true));
   private static final long serialVersionUID = 2663750991205874435L;
 
   public DatastoreException(int code, String message, String reason) {


ajkannan:
LGTM

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
```diff
@@ -38,13 +38,18 @@
   private static final long serialVersionUID = 2663750991205874435L;
```
aozarov: It is unclear to me (@eddavisson may add some clarity) if DEADLINE_EXCEEDED or UNAVAILABLE should be treated as "rejected" suggesting the server did not process the request (the latter sounds more like it than the former).
And if not "rejected" are those still safe to retry the commit (If not, doc should be clear about it).

google-cloud-datastore/src/main/java/com/google/cloud/datastore/spi/DefaultDatastoreRpc.java
```diff
@@ -130,7 +134,7 @@ public CommitResponse commit(CommitRequest request) {
     try {
       return client.commit(request);
     } catch (com.google.datastore.v1.client.DatastoreException ex) {
-      throw translate(ex);
+      throw translate(ex, request.getMode() == CommitRequest.Mode.NON_TRANSACTIONAL);
```
aozarov: We could add a retry-logic to [`runInTransaction`](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreHelper.java#L92) as a feature later on.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreException.java
eddavisson: If you get DEADLINE_EXCEEDED, you don't know if the underlying commit was triggered or not (or whether it succeeded).

If you get UNAVAILABLE, I believe it is safe to retry the commit (we should only return that before we trigger the underlying commit). UNAVAILABLE can also be returned client-side if the HTTP request fails.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/spi/DefaultDatastoreRpc.java
aozarov: Create an issue?

aozarov:
LGTM

google-cloud-datastore/src/main/java/com/google/cloud/datastore/spi/DefaultDatastoreRpc.java
mziccard: Sure thing, I will create one myself

PR #1236, opened 2016-09-09
Add support for customer-supplied encryption keys

### LINKED ISSUE #1037: Support customer-supplied encryption keys in Storage
Support for customer-supplied encryption keys was added in storage and we should support it as well.

How to use such keys is documented [here](https://cloud.google.com/storage/docs/encryption#customer-supplied). Fields added to a blob are documented [here](https://cloud.google.com/storage/docs/json_api/v1/objects#customerEncryption).

mziccard:
This fixes #1037


mziccard:
/cc @lesv 

(I'll add some more tests to avoid this coverage drop)

lesv:
LGTM - Thanks for this!  Ping and I'll read the additional tests.

mziccard:
@lesv I rebased on top of master and added more coverage.

lesv:
LOL - coveralls says it decreased.

lesv:
LGTM

PR #1240, opened 2016-09-12
Remove Datastore's Serializable class and related changes

### LINKED ISSUE #1112: remove newlines from datastore Key.toString() values
Calling `.toString()` on a datastore `Key`, for instance when logging some value or operation in application code, returns a String containing newlines and tabs, which might be too verbose in places where each "log line" is expected to be a single line.

For instance:

```
partition_id {
  project_id: "some-project-id"
}
path {
  kind: "taskqueue"
  id: -1037827333
}
```

Since the toString implemented in BaseKey is converting the instance to protobuf and calling toString on the returned GeneratedMessage, could one of the protobuf methods for formatting without newlines be used instead for condensed output?

mziccard:
This fixes #1112

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java
@@ -32,6 +32,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.Maps;
 
+import java.io.Serializable;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -49,7 +50,7 @@ import java.util.Set;
  * @see <a href="https://cloud.google.com/datastore/docs/concepts/entities">Google Cloud Datastore
  *     Entities, Properties, and Keys</a>
  */
-public abstract class BaseEntity<K extends IncompleteKey> implements java.io.Serializable {
+public abstract class BaseEntity<K extends IncompleteKey> implements Serializable {
 
   private static final long serialVersionUID = -9070588108769487081L;
   private final ImmutableSortedMap<String, Value<?>> properties;
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQuery.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQuery.java
@@ -752,7 +752,7 @@ public abstract class StructuredQuery<V> extends Query<V> {
 
   @Override
   public String toString() {
-    return super.toStringHelper()
+    return toStringHelper()
         .add("kind", kind)
         .add("startCursor", startCursor)
         .add("endCursor", endCursor)
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreBatchWriter.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreBatchWriter.java
@@ -60,7 +60,7 @@ interface DatastoreBatchWriter extends DatastoreWriter {
   /**
    * {@inheritDoc}
    * This operation will also remove from this batch any prior writes for entities with the same
-   * keys
+   * keys.
    *
    * @throws DatastoreException if not active
    */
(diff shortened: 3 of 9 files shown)


mziccard:
This removes Datastore's `Serializable` class (as I mentioned this some time ago). Other than that this PR also:
- Implements `toString` on a per-class basis without relying on `toPb`
- Removes `to/fromPb` methods that are not needed
- Fixes some minor style issues

@aozarov your opinion would be valuable in this

google-cloud-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java
```diff
@@ -49,12 +49,10 @@
  * @see <a href="https://cloud.google.com/datastore/docs/concepts/entities">Google Cloud Datastore
  *     Entities, Properties, and Keys</a>
  */
-public abstract class BaseEntity<K extends IncompleteKey>
-    extends Serializable<com.google.datastore.v1.Entity> {
+public abstract class BaseEntity<K extends IncompleteKey> implements java.io.Serializable {
```
aozarov: import `java.io.Serializable`?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Cursor.java
```diff
@@ -22,17 +22,18 @@
 import com.google.common.base.MoreObjects.ToStringHelper;
 import com.google.common.io.BaseEncoding;
 import com.google.protobuf.ByteString;
-import com.google.protobuf.InvalidProtocolBufferException;
+
```
aozarov: FYI, The google style guide has changed/relaxed regarding [imports](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing).

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Cursor.java
```diff
@@ -22,17 +22,18 @@
...
+import java.io.Serializable;
 
 /**
  * A Google Cloud Datastore cursor.
  * The cursor can be used to as a starting point or an ending point for a {@link Query}
  */
-public final class Cursor extends Serializable<com.google.datastore.v1.Value> {
+public final class Cursor implements Serializable {
```
aozarov: A toString would be nice here.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DateTime.java
```diff
@@ -32,12 +31,11 @@
  * @see <a href="https://cloud.google.com/datastore/docs/concepts/entities">Google Cloud Datastore
  *     Entities, Properties, and Keys</a>
  */
-public final class DateTime extends Serializable<com.google.protobuf.Timestamp>
-    implements Comparable<DateTime> {
+public final class DateTime implements Comparable<DateTime>, Serializable {
```
aozarov: A toString would be nice here.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Cursor.java
mziccard: There already is one, see line 53.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DateTime.java
mziccard: Also here, one was already there, see line 45.

aozarov: Oh, right. fooled by partial collapse.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQuery.java
```diff
@@ -752,9 +751,24 @@ B mergeFrom(com.google.datastore.v1.Query queryPb) {
   }
 
   @Override
+  public String toString() {
+    return super.toStringHelper()
```
aozarov: Nit, unnecessary call to `super`.

aozarov:
LGTM

PR #1241, opened 2016-09-12
Expose static methods to get the default project ID and credentials

### LINKED ISSUE #1207: Google Cloud Dataflow/Apache Beam and gcloud-java-core
We have two external pull requests which add logic to support the active gcloud profile, and also to get the project from the service account which are already supported by gcloud-java-core.

Dataflow/Apache Beam would like to depend on gcloud-java-core for the project detection logic found here: https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-core/src/main/java/com/google/cloud/ServiceOptions.java?l=66

And also the auth credentials detection here: https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-core/src/main/java/com/google/cloud/ServiceOptions.java?l=237

Currently these are hidden away, is there a way for you to expose these concepts without needing to create a ServiceOptions subclass?

If not, how do you advise Dataflow/Apache Beam to integrate with gcloud-java-core?

Note, that Apache Beam would be interested in the python versions of that code, and in general this would be useful in the other language implementations as well.

mziccard:
This fixes #1207

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -149,7 +149,7 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
     }
 
     /**
-     * Sets project id.
+     * Sets project ID.
      *
      * @return the builder
      */
@@ -235,7 +235,18 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
     return true;
   }
 
-  private static AuthCredentials defaultAuthCredentials() {
+  /**
+   * Returns the default authentication credentials, or {@code null} if no default credentials could
+   * be found. This method returns the first available credentials among the following sources:
+   * <ol>
+   *   <li>App Engine credentials
+   *   <li>JSON credentials pointed to by the {@code GOOGLE_APPLICATION_CREDENTIALS} environment
+   *   variable
+   *   <li>Google Cloud SDK credentials
+   *   <li>Compute Engine credentials
+   * </ol>
+   */
+  public static AuthCredentials defaultAuthCredentials() {
     // Consider App Engine.
     if (appEngineAppId() != null) {
       try {
@@ -261,6 +272,21 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
   }
 
   protected String defaultProject() {
+    return defaultProjectId();
+  }
+
+  /**
+   * Returns the default project ID, or {@code null} if no default project ID could be found. This
+   * method returns the first available project ID among the following sources:
+   * <ol>
+   *   <li>App Engine project ID
+   *   <li>Project ID specified in the JSON credentials file pointed by the
+   *   {@code GOOGLE_APPLICATION_CREDENTIALS} environment variable
+   *   <li>Google Cloud SDK project ID
+   *   <li>Compute Engine project ID
+   * </ol>
+   */
+  public static String defaultProjectId() {
     String projectId = System.getProperty(PROJECT_ENV_NAME, System.getenv(PROJECT_ENV_NAME));
     if (projectId == null) {
       projectId =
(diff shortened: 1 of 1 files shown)


google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -261,6 +272,21 @@ protected String defaultHost() {
...
 
   protected String defaultProject() {
+    return defaultProjectId();
+  }
+
+  /**
+   * Returns the default project ID, or {@code null} if no default project ID could be found. This
+   * method returns the first available project ID among the following sources:
+   * <ol>
+   *   <li>App Engine project ID
```
lukecwik: It seems as though you get it from PROJECT_ENV_NAME, then LEGACY_PROJECT_ENV_NAME

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -235,7 +235,18 @@ protected boolean projectIdRequired() {
     return true;
   }
 
-  private static AuthCredentials defaultAuthCredentials() {
+  /**
+   * Returns the default authentication credentials, or {@code null} if no default credentials could
+   * be found. This method returns the first available credentials among the following sources:
+   * <ol>
+   *   <li>App Engine credentials
+   *   <li>JSON credentials pointed to by the {@code GOOGLE_APPLICATION_CREDENTIALS} environment
```
lukecwik: Is it wiser to instead give a link to the GoogleCredentials.defaultAuth method since this javadoc will become stale quickly if that method changes. I was under the impression that GoogleCredentials also supported cloud shell in addition to these other methods as well.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lukecwik: It seems as though the order here is wrong, we seem to get it from PROJECT_ENV_NAME and then LEGACY_PROJECT_ENV_NAME both for gcloud, then its AppEngine, then JSON credentials from service account, then gcloud configuration.

Where does the compute engine project id come from?

mziccard:
@lukecwik thanks for the feedback. I am putting this on hold as we are right now working on a better integration with https://github.com/google/google-auth-library-java/ that will eventually lead to us dropping our `AuthCredentials` classes.

lukecwik:
Does the same apply to the other google cloud languages?
I'm mainly interested in python as well but consistency of features across languages would be convenient.

mziccard:
Not 100% sure what python guys are doing. Also, the availability of libraries and tools across different platforms/languages is far from uniform, this makes the comparison hard if at all useful.

I believe that `google-cloud-python` is using credentials classes (`GoogleCredentials`) from https://github.com/google/oauth2client and they possibly provide some more utility functions (see https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/core/google/cloud/credentials.py).

lukecwik:
Any updates?

mziccard:
@lukecwik We are working on getting the features we need in `google-auth-library-java`, see [this PR](https://github.com/google/google-auth-library-java/pull/69) for instance. ETA is end of the month.

lukecwik:
Thanks for the update.

lukecwik:
Update?

mziccard:
PR to switch to `google-auth-library-java` is at https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1375, review is pending. You can start using `GoogleCredentials.getApplicationDefault()` by depending on [google-auth-library-java](https://github.com/google/google-auth-library-java) `0.6.0`. Once that is done I will then expose a static method for project discovery.

lukecwik:
Swapped to use `google-auth-library-java` `0.6.0` `GoogleCredentials.getApplicationDefault` in https://github.com/apache/incubator-beam/pull/1294

mziccard:
This was replaced by #1380

PR #1244, opened 2016-09-14
Add imports when possibile in BigQuery functional classes

mziccard:


lesv:
@tswast FYI
LGTM - This looks better -- Thanks.

PR #1259, opened 2016-09-16
Change Map<Str, Obj> -> Map<Str, ? extends Obj>

jcanizales:
This lets users write, e.g.:

``` java
builder.addRow("rowId", ImmutableMap.of("booleanField", false))
```

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/InsertAllRequest.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/InsertAllRequest.java
@@ -76,7 +76,7 @@ public final class InsertAllRequest implements Serializable {
     private final String id;
     private final Map<String, Object> content;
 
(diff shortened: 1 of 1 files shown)


jcanizales:
@lesv @mziccard

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/InsertAllRequest.java
```diff
@@ -76,7 +76,7 @@
     private final String id;
     private final Map<String, Object> content;
 
-    RowToInsert(String id, Map<String, Object> content) {
+    RowToInsert(String id, Map<String, ? extends Object> content) {
```
eamonnmcmanus: `<? extends Object>` is the same as just `<?>`.

jcanizales: 👍 done

PR #1262, opened 2016-09-17
Add Guava as a dependency of the examples project

jcanizales:
We're using it in one of the sample snippets to initialize maps in a sane way (using `ImmutableMap.of(...)`).

@lesv @mziccard


mziccard:
I'd prefer not to show Guava classes in snippets (Guava is not standard java and not all users use it or know it, adding Guava classes in snippets might just be confusing for people not familiar with them).

Also I believe this change is not necessary, Guava should already be available as it's a dependency of `google-cloud-core`.

PR #1273, opened 2016-09-21
Update grpc, protobuf, datastore and gax dependencies, regenerate SPI classes

garrettjonesgoogle:
- Updating grpc to 1.0.1 and protobuf to 3.0.0
- Converting PageAccessor to PagedListResponse
- Adding auto-generated tests for logging

### DIFF
--- a/google-cloud-errorreporting/pom.xml
+++ b/google-cloud-errorreporting/pom.xml
@@ -20,7 +20,7 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork17</version>
+      <version>1.1.33.Fork19</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/google-cloud-language/pom.xml
+++ b/google-cloud-language/pom.xml
@@ -20,7 +20,7 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork17</version>
+      <version>1.1.33.Fork19</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/google-cloud-logging/pom.xml
+++ b/google-cloud-logging/pom.xml
@@ -20,7 +20,7 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork17</version>
+      <version>1.1.33.Fork19</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/google-cloud-monitoring/pom.xml
+++ b/google-cloud-monitoring/pom.xml
@@ -20,7 +20,7 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork17</version>
+      <version>1.1.33.Fork19</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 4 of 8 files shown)


mziccard:
@garrettjonesgoogle I finally merged monitoring, does this affect this PR?

Build is failing, it seems that `gax 0.0.18` is not out yet.

garrettjonesgoogle:
Yes, monitoring affects this PR. I will update it. `gax 0.0.18` is released now so the next build should succeed.

garrettjonesgoogle:
Updated monitoring, PTAL

garrettjonesgoogle:
It turns out that datastore depended on generated proto stuff too, so I have updated the dependency. PTAL.

mziccard:
> It turns out that datastore depended on generated proto stuff too, so I have updated the dependency. PTAL.

Yes this was discussed [here](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1239#issuecomment-247201013).

Non related to this PR: I noticed that in all VeneerToolkit `pom.xml` there's the `auto-value` dependency, is it really needed?
```xml
<dependency>
  <groupId>com.google.auto.value</groupId>
  <artifactId>auto-value</artifactId>
  <version>1.1</version>
</dependency>
```

garrettjonesgoogle:
It looks like it wasn't needed; it got added once, and then stayed for the ride. I have removed it now.

garrettjonesgoogle:
I think this is ready to merge - all tests pass.

mziccard:
Was the latest merge commit necessary (bb52fa0)?

I have a bunch of PRs to merge, this will be the next one.

garrettjonesgoogle:
It was not strictly necessary; I was just making sure this PR was up to date. Does it cause a problem?

mziccard:
> It was not strictly necessary; I was just making sure this PR was up to date. Does it cause a problem?

I would not say it causes problems but if I merge this PR using a merge commit that merge commit will remain in the history of master and its unnecessary, if I squash the commits in this PR I get rid of that commit but I also lose the history (which I would prefer not to).

garrettjonesgoogle:
I see. For googleapis/, we have been doing squash commits to master in order to have a more concise history instead of including every update in each PR, so this merge was more appropriate for that model. I could rebase and force push to the PR so that the merge disappears, if you would like (just let me know).

mziccard:
I see now that all major changes are in one commit (the first one). I'll squash commits as history is not so meaningful. In cases like this, however, where we updated many different and unrelated dependencies as well as added tests, I prefer having different commits and keep them in the history. So in case something went wrong you can revert changes at a finer granularity.

mziccard:
I am adding the don't merge label as this PR causes all gRPC calls to fail with a `DEADLINE_EXCEEDED` error.

garrettjonesgoogle:
I have fixed the gRPC call failures. PTAL.

garrettjonesgoogle:
Do I need to worry about the appveyor failure?

google-cloud-core/pom.xml
```diff
@@ -101,22 +101,22 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
-      <version>3.0.0-beta-3</version>
+      <version>3.0.0</version>
```
mziccard: I see that 2 more recent versions already exist for `protobuf-java`.
- [3.0.2](https://repo.maven.apache.org/maven2/com/google/protobuf/protobuf-java/3.0.2/)
- [3.1.0](https://repo.maven.apache.org/maven2/com/google/protobuf/protobuf-java/3.1.0/)

Might be worth jumping straight to `3.1.0`. I run integration tests with no ugly surprises.

garrettjonesgoogle: I would prefer to just keep the stack consistent; grpc depends on 3.0.0, and all of the grpc-generated packages reference 3.0.0. I kind of want to get this in and move on.

mziccard: Not that I agree 100% but fine with me

PR #1275, opened 2016-09-21
Add of() method to DatasetInfo

### LINKED ISSUE #1266: DatasetInfo could use a static of() method
### What:
Add a static `of()` method to `DatasetInfo`, like `JobInfo` and `TableId` have.

### Why:
It looks nicer than explicitly constructing a builder and then calling `build()` on it. See sample I wrote for `create(DatasetInfo)` and how it could be simplified a bit.

mziccard:
Fixes #1266


lesv:
LGTM

PR #1280, opened 2016-09-23
DO NOT MERGE: Add streaming support to Speech API.

shinfan:
- This is an example PR that shows bi-directional streaming support in vkit java. 
  - If the streaming is one-directional (the request is not StreamObserver), parameter flattening will be applied as the same as other API calls.
- Right now we just expose the grpc StreamObserver directly.


shinfan:
Move to staging repo.

PR #1288, opened 2016-09-27
Make PubSub integration tests more robust

mziccard:
This should reduce some failures we have seen in travis from now and then

/cc @lesv


lesv:
LGTM

PR #1293, opened 2016-09-29
Add reset() method to LocalDatastoreHelper

### LINKED ISSUE #1292: Datastore Testing: Wiping data between tests?
I'm migrating from the App Engine Datastore API to the Cloud Datastore API (so that I can connect to Datastore from a VPS). A happy thing about the previous API is that it didn't spin up a separate emulator process (5+ seconds) when `helper.setUp()` is called. With the Cloud Datastore API, it's a different model, so it's a little slower.

I don't want to have to entirely restart my Datastore emulator between tests, but I want a clean slate. Can I either:
* Still have an in-process datastore implementation to test against, which is fast?
* Clear the remote emulator datastore, so it doesn't need to be fully restarted?

I've looked through `LocalDatastoreHelper`, `DatastoreOptions` and `Datastore` itself, but I can't find any immediately solutions to my problem.

Note: I can't just `@After` and delete a set of keys, since some tests involve letting Datastore choose the `id` of an inserted entity

mziccard:
This fixes #1292


lesv:
LGTM

PR #1303, opened 2016-10-03
Support using more recent datastore emulator versions via gcloud emulators

### LINKED ISSUE #1300: It is too slow testing Datastore locally when use LocalDatastoreHelper
It is too slow testing Datastore locally when we use `com.google.cloud.datastore.testing.LocalDatastoreHelper` in `google-cloud-0.4.0` with `cloud-datastore-emulator 1.2.1`.

I figure out  `LocalDatastoreHelper` define `private static final String GCD_VERSION = "1.2.0";` and lastest gcloud datastore version is `cloud-datastore-emulator 1.2.1` (`gcloud-tools` current version is `128.0.0`).  So, when we test by using LocalDatastoreHelper, it cannot find the `cloud-datastore-emulator 1.2.0` version in gcloud path, and as result, it ends up making download of `cloud-datastore-emulator 1.2.1` for each test case.

To solve this problem, I had to downgrade the `gcloud-tools` to `125.0.0`.

mziccard:
This fixes #1300 for good

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
@@ -54,9 +54,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


lesv:
LGTM - if only folks always kept versions like you expect them.  It's likely the emulator will keep the correct scheme, but lots of teams use 0.1.2-1.9.42, 0.2.3-beta, 0.2.3-SNAPSHOT, or 0.2.3-rev172 etc.

mziccard:
@lesv I was strict on the version format on purpose. I prefer the helper to download its own release version of the emulator rather than possibly using a beta/alpha version already installed via `gcloud`. I agree this should not happen with emulators.

PR #1308, opened 2016-10-04
Rename setters/getters/builders for Storage classes to meet proto conventions

mziccard:
This is the first of a series of PRs that update our naming conventions to meet the ones of proto/gRPC/VeneerToolkit.

Changes are:
- `property(Type property)` setters renamed to `setProperty(Type property)`
- `property()` getters renamed to `getProperty()`
- `builder()` factory methods renamed to `newBuilder()`

Previous methods are kept and deprecated for now.

### DIFF
--- a/google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITBucketSnippets.java
+++ b/google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITBucketSnippets.java
@@ -86,7 +86,7 @@ public class ITBucketSnippets {
     Bucket bucket = bucketSnippets.reload();
     assertNotNull(bucket);
     Bucket updatedBucket = bucketSnippets.update();
-    assertTrue(updatedBucket.getVersioningEnabled());
+    assertTrue(updatedBucket.versioningEnabled());
     Blob blob1 = bucketSnippets.createBlobFromByteArray(BLOB1);
     assertNotNull(blob1);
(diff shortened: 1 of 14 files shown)


lesv:
Is that a google standard or a vKit thing?

mziccard:
No Google standard that I know of (Guava in fact uses our current style). As per title, this is used in protos and VKIT. In order to be able to expose protos and VKIT classes without breaking Veneer look-and-feel we need to obey to their conventions on this.

lesv:
Sigh - the shorter form works better in IDE's.

mziccard:
@lesv I added 4 commits that do the following:
- Avoid deprecating package private methods, just rename them
- Update `TESTING.md` readme
- Update storage usage in `ITBigQueryTests`
- Update `google-cloud-nio` code to use renamed methods

lesv:
LGTM - I won't be using approve / commit again.   Not quite the behavior I was looking for.

BlobInfo seems public not "package private".  But I defer to your judgement.

mziccard:
> BlobInfo seems public not "package private". But I defer to your judgement.

Yes classes are public but methods I renamed without deprecating them should all be package-private. If that is not the case please point me to the methods so I can fix it.

lesv:
Looked closer, never found a public, so I'm happy.

LGTM

lesv:
LGTM

lesv:
LGTM

PR #1310, opened 2016-10-06
Rename setters/getters/builders for BigQuery classes to meet proto conventions

mziccard:


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
```diff
@@ -190,10 +285,25 @@ public boolean booleanValue() {
...
+   * @throws ClassCastException if the field is not a primitive type
+   * @throws NumberFormatException if the field's value could not be converted to {@link Long}
+   * @throws NullPointerException if {@link #isNull()} returns {@code true}
+   */
+  @SuppressWarnings("unchecked")
+  public long getTimestampValue() {
     // timestamps are encoded in the format 1408452095.22 where the integer part is seconds since
     // epoch (e.g. 1408452095.22 == 2014-08-19 07:41:35.220 -05:00)
-    return new Double(((Double.valueOf(stringValue())) * MICROSECONDS)).longValue();
+    return new Double(((Double.valueOf(getStringValue())) * MICROSECONDS)).longValue();
```
lesv: +1 this would be good to fix.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryStage.java
```diff
@@ -428,21 +559,21 @@ ExplainQueryStage toPb() {
 
   static QueryStage fromPb(com.google.api.services.bigquery.model.ExplainQueryStage stagePb) {
     Builder builder = new QueryStage.Builder();
```
lesv: Should this be newBuilder() (or is it private?)

mziccard: What is the problem with this? This is internal code that accesses directly the constructor rather than going through the public builder method, I don't see an issue with that.

mziccard: > in the CL's that large, it's difficult to figure out if something was changed or not, or go back and find the source.

I know and I am sorry for these big PRs. I wanted all the changes for one service to be contained in the same PR, to avoid having inconsistent states in master.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
mziccard: Fixed

PR #1312, opened 2016-10-10
Rename setters/getters/builders for PubSub classes to meet proto conventions

mziccard:


mziccard:
I updated example with a9e8397, did I miss something?

lesv:
No - I did.  Sigh.  I think that was the first file or two, saw it, but didn't register what it was.  Thanks.

lesv:
LGTM

PR #1315, opened 2016-10-13
Rename setters/getters/builders for Datastore classes to meet proto conventions

mziccard:


google-cloud-datastore/src/main/java/com/google/cloud/datastore/IncompleteKey.java
```diff
@@ -77,34 +77,77 @@ static IncompleteKey fromPb(com.google.datastore.v1.Key keyPb) {
    * Returns the key's parent.
    */
   @Override
+  @Deprecated
   public Key parent() {
```
lesv: I agree w/ codacy-bot

mziccard: The methods renamed are very simple (getters and setters in most cases) so I didn't see the need for calling the renamed one. I will updated the PRs accordingly though.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/BaseKeyTest.java
```diff
@@ -71,13 +86,29 @@ public void testBadDatasetInConstructor() throws Exception {
...
+    Builder builder = new Builder("d", "k");
     builder.projectId(" ");
   }
 
   @Test
   public void testNamespace() throws Exception {
     Builder builder = new Builder("ds", "k");
     BaseKey key = builder.build();
+    assertTrue(key.getNamespace() != null);
+    assertTrue(key.getNamespace().isEmpty());
```
mziccard: We are not using `Truth`. I believe that only the guys who developed the `google-cloud-nio` module introduced (and used) truth (only in that module). Switching to truth might be reasonable but it's a low priority task and out of the scope of this PR.

lesv: Totally understood -- I ask questions sometimes just to understand the choices rather than as an objection.  I wouldn't give an LGTM if I was objecting or focused on something.

mziccard: This kind of comments are valuable. In fact, using truth would have been a good choice since the beginning.

PR #1317, opened 2016-10-17
Add support for PubSub emulator host variable

### LINKED ISSUE #998: Support standard env vars for emulator testing
https://github.com/GoogleCloudPlatform/gcloud-python/blob/master/gcloud/environment_vars.py

mziccard:
This fixes #998. Datastore already support emulator host variable.


lesv:
LGTM - I had to do something like that last week.

PR #1326, opened 2016-10-19
Update spi classes with GAX 0.0.21

shinfan:
- Add grpc streaming support
- Rename ApiCallable to UnaryCallable
- Rename ApiCallSettings to UnaryCallSettings
- Retire "page streaming" and use "paged" instead


shinfan:
@mziccard Please review.

google-cloud-errorreporting/src/main/java/com/google/cloud/errorreporting/spi/v1beta1/PagedResponseWrappers.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.devtools.clouderrorreporting.v1beta1.ListGroupStatsResponse;
+
+// AUTO-GENERATED DOCUMENTATION AND CLASS
+/**
+ * Wrapper class to contain paged response types for page streaming methods. Each static class
+ * inside this wrapper class is used as the return type of one of an API method that implements the
+ * page streaming pattern.
+ */
+@javax.annotation.Generated("by GAPIC")
+public class PagedResponseWrappers {
```
mziccard: What is the purpose of these classes? Just hiding generics?

google-cloud-language/src/main/java/com/google/cloud/language/spi/v1beta1/PagedResponseWrappers.java
```diff
@@ -0,0 +1,23 @@
...
+package com.google.cloud.language.spi.v1beta1;
+
+// AUTO-GENERATED DOCUMENTATION AND CLASS
+/**
+ * Wrapper class to contain paged response types for page streaming methods. Each static class
+ * inside this wrapper class is used as the return type of one of an API method that implements the
+ * page streaming pattern.
+ */
+@javax.annotation.Generated("by GAPIC")
+public class PagedResponseWrappers {}
```
mziccard: Is it necessary to generate empty classes?

google-cloud-errorreporting/src/main/java/com/google/cloud/errorreporting/spi/v1beta1/PagedResponseWrappers.java
garrettjonesgoogle: Yes, that is the reason.

google-cloud-language/src/main/java/com/google/cloud/language/spi/v1beta1/PagedResponseWrappers.java
garrettjonesgoogle: No - we have an open issue to not generate them: https://github.com/googleapis/toolkit/issues/617

google-cloud-core/src/main/java/com/google/cloud/GrpcServiceOptions.java
```diff
@@ -226,7 +226,7 @@ protected GrpcServiceOptions(
   /**
    * Returns a builder for API call settings.
    */
-  protected ApiCallSettings.Builder apiCallSettings() {
+  protected UnaryApiCallSettings.Builder apiCallSettings() {
```
garrettjonesgoogle: Yeah this is inconsistent. @shinfan , could you address the inconsistency? Maybe removing the `Api` from the names would be best. We needed `Api` on `ApiCallable` because using `Callable` on its own would be confusing due to the existing type in Java, but the other names don't have that problem. 

@mziccard it's not clear which class you are referencing when you say "Should we consider renaming it?" I can see how `PageStreamingCallSettings` might seem more related to `StreamingCallSettings`. Maybe we should call the first one `PagedListCallSettings`.

Does this scheme work?
* `UnaryCallable` and `UnaryCallSettings`
  * `PagedListCallSettings`
* `StreamingCallable` and `StreamingCallSettings`

mziccard: > it's not clear which class you are referencing

Sorry, but you got it right I was suggesting renaming `PageStreamingCallSettings`.
Proposed naming sounds much better to me. I would also consider `PagedCallSettings`, any reason you ruled that out?

garrettjonesgoogle: We have used the "paged list" terminology internally in some places. Maybe you're right, simply `PagedCallSettings` might be better. We'll discuss internally.

shinfan: Renaming done.

shinfan:
Removed AgentTranslationService and its tests since the corresponding proto is no longer available from googleapis:
https://github.com/googleapis/googleapis/tree/master/google/monitoring/v3

mziccard:
@shinfan It seems that gax 0.0.21 is not out yet. Also, didn't we get rid of the unused imports error?

garrettjonesgoogle:
It seems that google-java-format isn't removing the particular import in question. It is an inner static class, which we reference in a semi-qualified manner in the code (ContainerClass.InnerClass).

google-cloud-errorreporting/src/main/java/com/google/cloud/errorreporting/spi/v1beta1/ErrorStatsServiceSettings.java
```diff
@@ -319,14 +348,21 @@ private Builder(ErrorStatsServiceSettings settings) {
...
       deleteEventsSettings = settings.deleteEventsSettings.toBuilder();
 
-      methodSettingsBuilders =
-          ImmutableList.<ApiCallSettings.Builder>of(
+      unaryMethodSettingsBuilders =
+          ImmutableList.<UnaryCallSettings.Builder>of(
               listGroupStatsSettings, listEventsSettings, deleteEventsSettings);
     }
 
+    private static ConnectionSettings.Builder s_getDefaultConnectionSettingsBuilder() {
```
mziccard: Is there a reason to favor these static methods over constants? Also, FYI, [google style guide](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names) suggests that underscores should only be used for test cases.

shinfan:
@mziccard There is some sync delay between the nexus and maven. Can you trigger the travis again? It should be available now.

mziccard:
@shinfan I restarted travis and everything seems fine now. Any comment on [this](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1326#discussion_r85392367)?

google-cloud-errorreporting/src/main/java/com/google/cloud/errorreporting/spi/v1beta1/ErrorStatsServiceSettings.java
shinfan: @garrettjonesgoogle Looks like we should fix the function names. I think Garrett is working on the refactor of these connection settings, so I suggest we can fix these with the refactoring (assuming this is a non-blocking issue).

shinfan: Done

PR #1327, opened 2016-10-20
shade everything but us in the shaded nio jar

jean-philippe-martin:
Update our shaded jar example to actually shade everything except what it needs (google-cloud-nio). This allows this jar to be used with recent versions of Spark that also bring in a conflicting version of netty.

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/pom.xml
+++ b/google-cloud-contrib/google-cloud-nio/pom.xml
@@ -112,6 +112,18 @@
               <shadedPattern>shaded.cloud-nio.org</shadedPattern>
             </relocation>
             <relocation>
+              <pattern>io</pattern>
+              <shadedPattern>shaded.cloud-nio.io</shadedPattern>
+            </relocation>
+            <relocation>
+              <pattern>javax</pattern>
+              <shadedPattern>shaded.cloud-nio.javax</shadedPattern>
+            </relocation>
+            <relocation>
+              <pattern>okio</pattern>
+              <shadedPattern>shaded.cloud-nio.okio</shadedPattern>
+            </relocation>
+            <relocation>
               <pattern>google</pattern>
               <shadedPattern>shaded.cloud-nio.google</shadedPattern>
             </relocation>


jean-philippe-martin:
I ran the nio example and it still works fine:

$ java -cp google-cloud-contrib/google-cloud-nio/target/google-cloud-nio-0.4.1-SNAPSHOT-shaded.jar:google-cloud-contrib/google-cloud-nio-examples/target/google-cloud-nio-examples-0.4.1-SNAPSHOT.jar com.google.cloud.nio.examples.ListFilesystems
Installed filesystem providers:
  file
  jar
  gs

jean-philippe-martin:
cc: @jart @mziccard

google-cloud-contrib/google-cloud-nio/pom.xml
```diff
@@ -112,6 +112,18 @@
               <shadedPattern>shaded.cloud-nio.org</shadedPattern>
             </relocation>
             <relocation>
+              <pattern>io</pattern>
```
jart: Would we be able to make these shading patterns more specific? By the looks of this, it seems like they might pick up things we don't necessarily want to shade.

jean-philippe-martin: I thought the only thing we really wanted to export was com.google.cloud. Isn't a general shading pattern a good thing, then?

jart: I'm honestly not familiar enough with how the shading plugin works. But this will probably be fine.

jart: What is the name of the artifact that's going to have the shading?

jean-philippe-martin: It's google-cloud-nio-0.4.1-shaded.jar

mziccard:
@jart is this a LGTM for you?

@jean-philippe-martin While looking at this PR I noticed other dependency-related issues:

1. `org.mockito:mockito-core` dependency should have `test` scope
2. `google-cloud-nio` now depends on the whole `google-cloud`, we should depend just on `google-cloud-storage`
3. To make 2. effective we need to also remove `google-cloud` dependency from `google-cloud-contrib` main `pom.xml`, we can let each child module pull its own dependencies.

Does this make sense to you? We can take care in a different PR, I just wanted to bring this up.

jart:
LGTM

jean-philippe-martin:
@mziccard that makes sense, let me work on those in a separate PR. I've filed #1332.

jean-philippe-martin:
>
> What is the name of the artifact that's going to have the shading?
>

It's google-cloud-nio-0.4.1-shaded.jar

PR #1330, opened 2016-10-21
Fix google-cloud-logging groupId in readme

meltsufin:


mziccard:
@meltsufin Thanks for this! Merged right away.

Issue #1332, opened 2016-10-24
remove unneeded dependencies from nio


jean-philippe-martin:
mziccard noticed the following with google-cloud-nio:

1.  org.mockito:mockito-core dependency should have test scope
2.  google-cloud-nio now depends on the whole google-cloud, we should depend just on google-cloud-storage
3.  To make 2. effective we need to also remove google-cloud dependency from google-cloud-contrib main pom.xml, we can let each child module pull its own dependencies.

mziccard:
Fixed with #1333

PR #1333, opened 2016-10-24
Remove unused dependencies from contrib, NIO and NIO examples

jean-philippe-martin:
As per issue #1332:
- move `mockito-core` to test scope
- change `google-cloud-nio` dependency to `google-cloud-storage` rather than `google-cloud`

The google-cloud-nio-example test passes, as expected.

```
$ java -cp google-cloud-contrib/google-cloud-nio/target/google-cloud-nio-0.4.1-SNAPSHOT-shaded.jar:google-cloud-contrib/google-cloud-nio-examples/target/google-cloud-nio-examples-0.4.1-SNAPSHOT.jar com.google.cloud.nio.examples.ListFilesystems
Installed filesystem providers:
  file
  jar
  gs
```

cc: @mziccard

### DIFF
--- a/google-cloud-contrib/google-cloud-nio-examples/pom.xml
+++ b/google-cloud-contrib/google-cloud-nio-examples/pom.xml
@@ -22,57 +22,7 @@
       <artifactId>google-cloud-storage</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>19.0</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>jsr305</artifactId>
-      <version>2.0.1</version>
-    </dependency>
-    <dependency>
-      <groupId>javax.inject</groupId>
-      <artifactId>javax.inject</artifactId>
-      <version>1</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.auto.service</groupId>
-      <artifactId>auto-service</artifactId>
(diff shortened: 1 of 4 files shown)


google-cloud-contrib/pom.xml
```diff
@@ -23,7 +23,7 @@
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>google-cloud</artifactId>
+      <artifactId>google-cloud-storage</artifactId>
```
mziccard: I believe we can get rid of this dependency, no need to import anything here, we do that in child modules. Makes sense?

google-cloud-contrib/google-cloud-nio-examples/pom.xml
```diff
@@ -71,6 +71,7 @@
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <version>1.9.5</version>
+      <scope>test</scope>
```
mziccard: Are test dependencies needed for the example module?

google-cloud-contrib/pom.xml
jean-philippe-martin: Indeed! I removed it.

google-cloud-contrib/google-cloud-nio-examples/pom.xml
jean-philippe-martin: You're right, I removed all of the test dependencies.

mziccard:
@jean-philippe-martin Thanks for this!

jean-philippe-martin:
Thanks for fixing the title, @mziccard !

PR #1341, opened 2016-10-27
Rename setters/getters/builders for core classes to meet proto conventions

mziccard:


lesv:
So, github won't even show me all 314 files to review.  I'm presuming this is all stuff I've reviewed before.  

If that is true, consider this an **LGTM**.  (I've looked at the subset that Github showed me)

mziccard:
Argh.. you cannot see the git diff even of the single commit? File limit should be 300, both commits are under that threshold.

Changes are new, nothing special actually, but they might still benefit your check.

mziccard:
@lesv in [1341.diff](https://gist.github.com/mziccard/fbc89bce05af116a2de6d800a4f2f808) I copy pasted the diff for files not show by github. Can you please have a look?

lesv:
LGTM - Again - mind-numbing and very superficial review.  (I looked at every line, but did check for missing `@deprecations` or use of a deprecated method in an unchanged line.

PR #1348, opened 2016-10-28
Fix regression in storage url signing of objects with slashes in names

### LINKED ISSUE #1346: Regression in storage url signing

mziccard:
This replaces #1347 and fixes #1346

@ostronom I took the liberty to cherry-pick your commits and apply some fixes, namely:
- Squash and rebase on top of master
- Handle `?` character: `UrlEscapers.urlFragmentEscaper()` does not escape `?`. Even though the use of `?` in blob names is discouraged (see [here](https://cloud.google.com/storage/docs/naming#objectnames)) we better handle it properly
- Add slashes and special characters to `signUrl` integration tests

@ostronom @lesv can you have a look?

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
@@ -1193,7 +1193,7 @@ public class ITStorageTest {
 
   @Test
   public void testGetSignedUrl() throws IOException {
-    String blobName = "test-get-signed-url-blob/with/slashes/and?special=char*";
+    String blobName = "test-get-signed-url-blob/with/slashes/and?special=!#$&'()*+,:;=?@[]";
     BlobInfo blob = BlobInfo.newBuilder(BUCKET, blobName).build();
     Blob remoteBlob = storage.create(blob, BLOB_BYTE_CONTENT);
     assertNotNull(remoteBlob);


mziccard:
@mickeyreiss this should fix the regression, feel free to comment on :)

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -528,7 +528,8 @@ public URL signUrl(BlobInfo blobInfo, long duration, TimeUnit unit, SignUrlOptio
     if (blobInfo.getName().startsWith("/")) {
       path.setLength(path.length() - 1);
     }
-    path.append(UrlEscapers.urlPathSegmentEscaper().escape(blobInfo.getName()));
+    String escapedName = UrlEscapers.urlFragmentEscaper().escape(blobInfo.getName());
+    path.append(escapedName.replace("?", "%3F"));
```
lesv: Do you need to do something with "&" here as well?

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
```diff
@@ -1323,7 +1323,8 @@ public void testSignUrlForBlobWithSpecialChars() throws NoSuchAlgorithmException
       String blobName = "/a" + specialChar + "b";
       URL url =
           storage.signUrl(BlobInfo.newBuilder(BUCKET_NAME1, blobName).build(), 14, TimeUnit.DAYS);
-      String escapedBlobName = UrlEscapers.urlPathSegmentEscaper().escape(blobName);
+      String escapedBlobName =
+          UrlEscapers.urlFragmentEscaper().escape(blobName).replace("?", "%3F");
```
lesv: Do we also need to substitute "&amp;" for "&" ??

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1193,7 +1193,7 @@ public void testWriteChannelExistingBlob() throws IOException {
 
   @Test
   public void testGetSignedUrl() throws IOException {
-    String blobName = "test-get-signed-url-blob";
+    String blobName = "test-get-signed-url-blob/with/slashes/and?special=char*";
```
lesv: Should this have a 2nd parameter &blah=this+is+a+test

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
mickeyreiss: Did you consider configuring your own URL encoder, rather than implementing your own here?

mziccard: @lesv Escaping `&` is not necessary as long as we properly escape all `?`. See the improved integration test I pushed with 988c1c4

@mickeyreiss yours is a good refactoring suggestion that I plan on taking, however for the moment I say kick this out and fix the bug :D

lesv:
LGTM

PR #1355, opened 2016-10-31
Add better javadoc and snippets for Datastore and Transaction

### LINKED ISSUE #1339: JavaDoc not clear as it could be, e.g. Datastore.add vs. Datastore.put
What's the difference between Datastore.add and Datastore.put?

Am I right in assuming that "add" will **not** clobber existing entities?  

Maybe a quick update to the javadoc to clarify--would be much appreciated!!!

mziccard:
This PR does the following:
- Add better javadoc for `put` and `add` in `Datastore`, `Batch` and `Transaction`
- Add snippets for `Transaction`'s `put`, `add`, `update`, `delete`, `addWithDeferredIdAllocation` and `putWithDeferredIdAllocation`
- Add snippets for `Datastore`'s `add`
- Add integration tests for all new snippets

Fixes #1339 

@unitydynamics feel free to comment :)

/cc @lesv

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreBatchWriter.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreBatchWriter.java
@@ -26,8 +26,8 @@ interface DatastoreBatchWriter extends DatastoreWriter {
 
   /**
    * Datastore add operation. This method will also allocate id for any entity with an incomplete
-   * key. As opposed to {@link #add(FullEntity)}, this method will defer any necessary id allocation
-   * to submit time.
+   * key. As opposed to {@link #add(FullEntity)} and {@link #add(FullEntity...)}, this method will
+   * defer any necessary id allocation to submit time.
    *
    * @throws IllegalArgumentException if any of the given entities is missing a key
    * @throws DatastoreException if a given entity with a complete key was already added to this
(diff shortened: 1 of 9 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreBatchWriter.java
```diff
@@ -37,12 +37,24 @@
...
+
+  /**
+   * {@inheritDoc}
    * For entities with complete keys that were marked for deletion in this writer the operation
    * will be changed to {@link #put}.
    *
    * @throws DatastoreException if a given entity with the same complete key was already added to
    *     this writer, if writer is not active or if id allocation for an entity with an incomplete
-   *     key failed.
+   *     key failed
```
lesv: Is there a reason you removed the "."?

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Transaction.java
```diff
@@ -176,6 +176,205 @@
   <T> QueryResults<T> run(Query<T> query);
 
   /**
+   * Datastore add operation. This method will also allocate id for any entity with an incomplete
+   * key. As opposed to {@link #add(FullEntity)} and {@link #add(FullEntity...)}, this method will
+   * defer any necessary id allocation to commit time.
```
lesv: Seems unclear.   Perhaps:
**This method will also allocate id for any entity with an incomplete key unlike {@link #add(FullEntity)} and {@link #add(FullEntity...)}, this method defers any necessary id allocation to commit time.**

mziccard: I feel like your proposed change alters the meaning of the sentence.

"unlike {@link #add(FullEntity)} and {@link #add(FullEntity...)}" must refer to the second part of the sentence (i.e. when id allocation is made), not to the first part.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreBatchWriter.java
mziccard: According to the [javadoc guide](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@exception) it's not needed (same holds for `@param` and `@return` descriptions).

lesv: Sg

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Transaction.java
lesv: Sg.

PR #1365, opened 2016-11-02
Reduce gRPC dependency footprint, get rid of duplicate classes

### LINKED ISSUE #1278: Datastore: duplicate classes in dependencies
Using the com.google.cloud:google-cloud-datastore jar version 0.3.0 as a dependency in my project resulted in a lot of transitive dependency jars. I ran the [duplicate-finder plugin](https://github.com/basepom/duplicate-finder-maven-plugin/wiki) and it came up with duplicate classes in multiple of the datastore dependenies. Might be something you can do to exclude some of the dependencies?
[duplicate-finder.txt](https://github.com/GoogleCloudPlatform/google-cloud-java/files/489209/duplicate-finder.txt)

?

mziccard:
This reduced gRPC dependencies that we include (in particular in the core module) fixes #1278 for good.

@anthmgoogle @garrettjonesgoogle any chance you can have a look? This can also be done at GAX layer to fix https://github.com/googleapis/gax-java/issues/133. Also, is there a reason why proto artifacts (as `grpc-google-iam-v1` or `grpc-google-pubsub-v1`) depend on the whole `grpc-all` package? I believe that the dependency list there can be heavily reduced.

### DIFF
--- a/google-cloud-core/pom.xml
+++ b/google-cloud-core/pom.xml
@@ -106,7 +106,7 @@
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-protobuf</artifactId>
-      <version>1.0.1</version>
+      <version>${grpc.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.api</groupId>
--- a/google-cloud-errorreporting/pom.xml
+++ b/google-cloud-errorreporting/pom.xml
@@ -41,17 +41,17 @@
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-netty</artifactId>
-      <version>1.0.1</version>
+      <version>${grpc.version}</version>
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
-      <version>1.0.1</version>
+      <version>${grpc.version}</version>
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-auth</artifactId>
-      <version>1.0.1</version>
+      <version>${grpc.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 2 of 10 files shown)


garrettjonesgoogle:
Maybe I should bump the priority of the dependency change in gax so that you don't have to make this mess.

mziccard:
Actually GAX is not a big issue. We use GAX in our `google-cloud-core` module, but even with the GAX issue fixed I would still want to exclude some of its gRPC dependency (namely all but `grpc-protobuf`) in`google-cloud-core` (to reduce its footprint for those modules that depende on it but don't use gRPC). Then in gRPC-based modules I would still have to re-add all grpc dependencies that I previously excluded.

What is causing this mess is the fact that all proto artifacts depend on `grpc-all` which is as surprising as unnecessary.

garrettjonesgoogle:
Ahh, the proto artifacts - how much would it help for me to update their dependencies too?

mziccard:
> Ahh, the proto artifacts - how much would it help for me to update their dependencies too?

My guess is that they only need `grpc-protubuf`, if this is really the case I can get rid of all `<exclude>` elements for them (there's one for each pom.xml in this PR and 2 in the core module's pom.xml).

It's not a problem to send out this PR and then remove the `<excludes>`. Releasing releasing new proto artifacts for all services might take you some time.

google-cloud-core/pom.xml
```diff
@@ -104,19 +104,42 @@
       <version>3.0.0</version>
     </dependency>
     <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-protobuf</artifactId>
+      <version>1.0.1</version>
```
garrettjonesgoogle: would be be worth defining this version in a central variable?

garrettjonesgoogle:
LGTM

mziccard:
@garrettjonesgoogle thanks for reviewing!

PR #1371, opened 2016-11-04
DO NOT MERGE: regenerating spi layer - GAX 0.0.22

garrettjonesgoogle:
This is a non-working PR that only includes updates to the SPI
classes. The hand-written layer still needs to be updated.

### DIFF
--- a/google-cloud-core/pom.xml
+++ b/google-cloud-core/pom.xml
@@ -106,7 +106,7 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax</artifactId>
-      <version>0.0.21</version>
+      <version>0.0.22</version>
     </dependency>
     <dependency>
       <groupId>com.google.api.grpc</groupId>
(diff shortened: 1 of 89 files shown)


mziccard:
@garrettjonesgoogle thanks! I started working on the integration on this changes into `google-cloud-java` but I realized that you removed some classes form GAX:
```java
import com.google.api.gax.testing.DownloadableEmulatorRunner;
import com.google.api.gax.testing.GCloudEmulatorRunner;
import com.google.api.gax.testing.LocalServiceHelper;
```

Those classes were used in our `LocalPubSubHelper` (which was originally written by you and @shinfan).
I understand that you are using a mocking framework to test autogen code. However, should we add those classes back to GAX or should I implement them in `google-cloud-java`?

garrettjonesgoogle:
The auto-generated testing code already covers everything in the spi layer that the emulator test was covering, so it had become redundant from our perspective. Emulator support has been removed from our testing strategy because it doesn't scale as well. If you still want these tests in `google-cloud-java` to support the pubsub tests, I would recommend copying them in.

mziccard:
Replaced by #1379

PR #1375, opened 2016-11-07
Remove AuthCredentials and related classes, use google-auth-library-java instead

mziccard:
This PR gets rid of our `AuthCredentials` classes and uses [google-auth-library-java](https://github.com/google/google-auth-library-java) instead.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -102,7 +102,7 @@ You can test against a remote Datastore emulator as well.  To do this, set the `
   DatastoreOptions options = DatastoreOptions.newBuilder()
       .setProjectId("my-project-id") // must match project ID specified on remote machine
       .setHost("http://<hostname of machine>:<port>")
-      .setNoCredentials()
+      .setCredentials(NoCredentials.getInstance())
       .build();
   Datastore localDatastore = options.getService();
   ```
@@ -209,7 +209,7 @@ endpoint to the hostname of the remote machine, like the example below.
   PubSubOptions options = PubSubOptions.newBuilder()
       .setProjectId("my-project-id") // must match project ID specified on remote machine
       .setHost("<hostname of machine>:<port>")
-      .setNoCredentials()
+      .setCredentials(NoCredentials.getInstance())
       .build();
   PubSub localPubsub = options.getService();
   ```
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/SerializationTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/SerializationTest.java
@@ -17,6 +17,7 @@
 package com.google.cloud.bigquery;
 
 import com.google.cloud.BaseSerializationTest;
+import com.google.cloud.NoCredentials;
 import com.google.cloud.Restorable;
 import com.google.cloud.bigquery.StandardTableDefinition.StreamingBuffer;
 import com.google.common.collect.ImmutableList;
@@ -228,7 +229,7 @@ public class SerializationTest extends BaseSerializationTest {
   protected Serializable[] serializableObjects() {
     BigQueryOptions options = BigQueryOptions.newBuilder()
         .setProjectId("p1")
-        .setNoCredentials()
+        .setCredentials(NoCredentials.getInstance())
         .build();
     BigQueryOptions otherOptions = options.toBuilder().setProjectId("p2").build();
     return new Serializable[]{DOMAIN_ACCESS, GROUP_ACCESS, USER_ACCESS, VIEW_ACCESS, DATASET_ID,
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -84,7 +84,6 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
   private final String serviceRpcFactoryClassName;
   private final String serviceFactoryClassName;
   private final Clock clock;
-  private final boolean noCredentials;
   private final Credentials credentials;
 
   private transient ServiceRpcFactory<ServiceRpcT, OptionsT> serviceRpcFactory;
@@ -106,7 +105,6 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
 
     private String projectId;
     private String host;
-    private boolean noCredentials;
     private Credentials credentials;
     private RetryParams retryParams;
     private ServiceFactory<ServiceT, OptionsT> serviceFactory;
@@ -118,7 +116,6 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
     protected Builder(ServiceOptions<ServiceT, ServiceRpcT, OptionsT> options) {
       projectId = options.projectId;
       host = options.host;
-      noCredentials = options.noCredentials;
       credentials = options.credentials;
       retryParams = options.retryParams;
       serviceFactory = options.serviceFactory;
@@ -214,29 +211,18 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
     }
 
     /**
-     * Sets the service authentication credentials. If this method or {@link #setNoCredentials() are
-     * not used on the builder, {@link GoogleCredentials#getApplicationDefault()} will be used to
-     * attempt getting credentials from the environment.
+     * Sets the service authentication credentials. If no credentials are set,
+     * {@link GoogleCredentials#getApplicationDefault()} will be used to attempt getting credentials
+     * from the environment. Use {@link NoCredentials#getInstance()} to skip authentication, this is
+     * typically useful when using local service emulators.
      *
      * @param credentials authentication credentials, should not be {@code null}
      * @return the builder
      * @throws NullPointerException if {@code credentials} is {@code null}. To disable
-     *     authentication use {@link Builder#setNoCredentials()}
+     *     authentication use {@link NoCredentials#getInstance()}
      */
     public B setCredentials(Credentials credentials) {
       this.credentials = checkNotNull(credentials);
-      this.noCredentials = false;
-      return self();
-    }
-
-    /**
-     * Sets that no credentials should be used. This is typically useful when using the local
-     * service emulators, such as {@code LocalDatastoreHelper}, {@code LocalPubsubHelper} and
-     * {@code LocalResourceManagerHelper}.
-     */
-    public B setNoCredentials() {
-      this.noCredentials = true;
-      this.credentials = null;
       return self();
     }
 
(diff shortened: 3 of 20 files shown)


mziccard:
@lesv @garrettjonesgoogle care to have a look?

README.md
```diff
@@ -115,40 +115,63 @@ Most `google-cloud` libraries require a project ID.  There are multiple ways to
...
+`google-cloud-java` uses
+[https://github.com/google/google-auth-library-java](https://github.com/google/google-auth-library-java)
+to authenticate requests. `google-auth-library-java` supports a wide range of authentication types,
+see the project's [README](https://github.com/google/google-auth-library-java/blob/master/README.md)
+and [javadoc](http://google.github.io/google-auth-library-java/releases/0.6.0/apidocs/) for more
+details.
+
+To access Google Cloud services, you first need to ensure that the necessary Google Cloud APIs are
+enabled for your project. To do this, follow the instructions on the
+[authentication document](https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/authentication/readme.md#authentication)
```
lesv: This probably should point at cloud-site  possibly: https://cloud.google.com/docs/authentication
@broady @jgeewax do you have a thought about this?

README.md
```diff
@@ -115,40 +115,63 @@ Most `google-cloud` libraries require a project ID.  There are multiple ways to
 Authentication
 --------------
 
-First, ensure that the necessary Google Cloud APIs are enabled for your project. To do this, follow the instructions on the [authentication document](https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/authentication/readme.md#authentication) shared by all the gcloud language libraries.
+`google-cloud-java` uses
+[https://github.com/google/google-auth-library-java](https://github.com/google/google-auth-library-java)
+to authenticate requests. `google-auth-library-java` supports a wide range of authentication types,
```
garrettjonesgoogle: use a semicolon instead of a comma here

README.md
broady: +1 for preference of cloud.google.com over README files on GitHub.

README.md
```diff
@@ -115,40 +115,63 @@ Most `google-cloud` libraries require a project ID.  There are multiple ways to
...
-        .setAuthCredentials(AuthCredentials.createForJson(new FileInputStream("/path/to/my/key.json"))
+        .setCredentials(ServiceAccountCredentials.fromStream(new FileInputStream("/path/to/my/key.json"))
         .build()
         .getService();
     ```
-  * If running locally for development/testing, you can use Google Cloud SDK.  Download the SDK if you haven't already, then login using the SDK (`gcloud auth login` in command line).  Be sure to set your project ID as described above.
-  * If you already have an OAuth2 access token, you can use it to authenticate (notice that in this case the access token will not be automatically refreshed):
+  * If running locally for development/testing, you can use the
+  [Google Cloud SDK](https://cloud.google.com/sdk/). Create Application Default Credentials with
+  `gcloud auth application-default login`, `google-cloud` will automatically detect such
```
garrettjonesgoogle: add "and then": "...login, and then google-cloud..."

README.md
```diff
@@ -115,40 +115,63 @@ Most `google-cloud` libraries require a project ID.  There are multiple ways to
...
+  credentials.
+  * If you already have an OAuth2 access token, you can use it to authenticate (notice that in this
+  case the access token will not be automatically refreshed):
```
garrettjonesgoogle: add a comma: "...in this case, the access token..."

README.md
```diff
@@ -115,40 +115,63 @@ Most `google-cloud` libraries require a project ID.  There are multiple ways to
...
+If no credentials are provided, `google-cloud` will attempt to detect them from the environment
+using `GoogleCredentials.getApplicationDefault()` which will search for Default Application
+Credentials in the following locations (in order):
 
-1. Credentials supplied when building the service options
-2. App Engine credentials
-3. Key file pointed to by the GOOGLE_APPLICATION_CREDENTIALS environment variable
-4. Google Cloud SDK credentials
-5. Compute Engine credentials
+1. Credentials file pointed to by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
```
garrettjonesgoogle: Add "the": "1. The Credentials file..."

TESTING.md
```diff
@@ -102,7 +102,7 @@ You can test against a remote Datastore emulator as well.  To do this, set the `
   DatastoreOptions options = DatastoreOptions.newBuilder()
       .setProjectId("my-project-id") // must match project ID specified on remote machine
       .setHost("http://<hostname of machine>:<port>")
-      .setAuthCredentials(AuthCredentials.noAuth())
+      .setNoCredentials()
```
garrettjonesgoogle: We don't want to have a "null object" to pass in for no credentials?

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/SerializationTest.java
```diff
@@ -229,11 +228,11 @@
   protected Serializable[] serializableObjects() {
     BigQueryOptions options = BigQueryOptions.newBuilder()
         .setProjectId("p1")
-        .setAuthCredentials(AuthCredentials.createForAppEngine())
+        .setNoCredentials()
         .build();
     BigQueryOptions otherOptions = options.toBuilder()
         .setProjectId("p2")
-        .setAuthCredentials(null)
+        .setCredentials(null)
```
garrettjonesgoogle: why allow `null` credentials when we have a `setNoCredentials` method?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -28,7 +28,7 @@
 
   private static final String BIGQUERY_SCOPE = "https://www.googleapis.com/auth/bigquery";
   private static final Set<String> SCOPES = ImmutableSet.of(BIGQUERY_SCOPE);
-  private static final long serialVersionUID = -8592198255032667206L;
+  private static final long serialVersionUID = -2437598817433266049L;
```
garrettjonesgoogle: This is more for my own information than anything: How do we figure out when we should update these version UID's?

README.md
mziccard: https://cloud.google.com/docs/authentication contains code for Java showing how to authenticate with the Java apiary. Referencing it for here might just be confusing.

TESTING.md
mziccard: If `credentials` is `null` in the `Builder` object  we attempt to do credentials discovery. We kind of need a special indicator for the "no credentials" case. This is the reason why we previously had `AuthCredentials.noAuth()` and now have the `setNoCredentials()` setter (and the private `noCredentials` boolean inside `ServiceOptions` and `ServiceOptions.Builder`).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
mziccard: See [Type Changes Affecting Serialization](http://docs.oracle.com/javase/7/docs/platform/serialization/spec/version.html#6678) for some info on this.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/SerializationTest.java
mziccard: Fixed in faa4476, also added some docs and a test case

README.md
mziccard: Fixed all wording issues in 96c2b60

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -213,13 +214,18 @@ public B setHost(String host) {
...
     /**
-     * Sets the service authentication credentials.
+     * Sets the service authentication credentials. If this method or {@link #setNoCredentials() are
+     * not used on the builder, {@link GoogleCredentials#getApplicationDefault()} will be used to
+     * attempt getting credentials from the environment.
      *
+     * @param credentials authentication credentials, should not be {@code null}
      * @return the builder
+     * @throws NullPointerException if {@code credentials} is {@code null}. To disable
+     *     authentication use {@link Builder#setNoCredentials()}
```
garrettjonesgoogle: Given this, how does one take an existing options instance and make it use the default credentials?

mziccard: Given that we now throw on `null`:
```java
options.toBuilder().setCredentials(GoogleCredentials.getApplicationDefault()).build();
```
If we revert the NPE on `null` you could also do the following (where `build()` does the discovery for you):
```java
options.toBuilder().setCredentials(null).build();
```
I have no personal preference, if you do I am happy to go with it.

garrettjonesgoogle: The `setNoCredentials` pattern feels odd to me. It's a little like the pattern we were previously using in GAX, where different entry points overwrote the same data. I'd prefer one entry point I think (`setCredentials`). Could we create these types?

* `NullCredentials` - no credentials are used; for the emulator case.
* `DefaultCredentials` - I know this feels a little redundant with `GoogleCredentials.getApplicationDefault()`, but maybe there will be a case of a service where ADC is not the right default credentials to use for it. Additionally, having the user call `GoogleCredentials.getApplicationDefault()` themselves is slightly different, because the point of call for credential acquisition is different.

mziccard: I would prefer not to add types other than the ones in `google-auth-library-java`. Users already struggle with all the different credentials we have and it makes little sense to adopt a library for authentication and then start adding custom types.

We could consider adding a `NullCredentials` to `google-cloud-java` (or `NoAuthCredentials`) to get rid of the `setNoCredentials` method for supporting the emulator case, this seems as a reasonable compromise but I am strongly against adding `DefaultCredentials` class.

garrettjonesgoogle: Ok I can live with that for now. Maybe we can think about other default credentials sources at the time that it comes up in the future (maybe it never will). 

Would it make sense to move `NullCredentials` up to `google-auth-library-java`? If so, I'm happy to defer it until later, but it would be good to decide on the "proper" home.

mziccard: I don't see a use for `NullCredentials` outside of the emulator world. Do you?

garrettjonesgoogle: I don't currently. Since that is the case, maybe it makes sense for it to be scoped only to `google-cloud-java`, if that's what you're implying.

garrettjonesgoogle:
LGTM

PR #1377, opened 2016-11-08
Add BaseEmulatorHelper class and tests, refactor helper classes

mziccard:


google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -0,0 +1,428 @@
...
+    this.emulator = emulator;
+    this.port = port > 0 ? port : DEFAULT_PORT;
+    this.projectId = projectId;
+  }
+
+  /**
+   * Returns the emulator runners supported by this emulator. Runners are evaluated in order, the
+   * first available runner is selected and executed
+   */
+  protected abstract List<EmulatorRunner> emulatorRunners();
```
garrettjonesgoogle: Maybe call this `getEmulatorRunners()`?

google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -0,0 +1,428 @@
...
+      }
+      return false;
+    }
+
+    private boolean isEmulatorUpToDate() throws IOException, InterruptedException {
+      Version currentVersion = installedEmulatorVersion(versionPrefix);
+      return currentVersion != null && currentVersion.compareTo(minVersion) >= 0;
+    }
+
+    private Version installedEmulatorVersion(String versionPrefix)
```
garrettjonesgoogle: `getInstalledEmulatorVersion`?

google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -0,0 +1,428 @@
...
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+/**
+ * Utility class to start and stop a local service which is used by unit testing.
+ */
+public abstract class BaseEmulatorHelper<T extends ServiceOptions> {
```
garrettjonesgoogle: I willing to bend on this because of time pressure, but I'd really rather not use abstract classes in this case. I think we could just as easily favor composition over inheritance, and make the code easier to follow.

mziccard: I don't agree. With composition we would end up with a class with the exact same interface, just not in the hierarchy, which is odd to me unless there's a good reason to do it.

You would also be forced to expose more classes as `public`, while here we can just hide them inside `BaseEmulatorHelper`, mark them as `protected` and make them available only to implementation. This ends up in less users trying to use such classes (hopefully only those users that are knowingly to extend `BaseEmulatorHelper`) which are meant for internal use only.

garrettjonesgoogle:
LGTM

PR #1379, opened 2016-11-09
Regenerating spi layer - GAX 0.0.22

mziccard:
This replaces #1371 and applies the necessary changes for the code to work with new generated SPI.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java
@@ -16,12 +16,15 @@
 
 package com.google.cloud.logging.spi;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
-
-import com.google.api.gax.core.ConnectionSettings;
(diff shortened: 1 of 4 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java
```diff
@@ -245,6 +250,7 @@ public void close() throws Exception {
     configApi.close();
     loggingApi.close();
     metricsApi.close();
+    providerManager.getChannel().shutdown();
```
garrettjonesgoogle: You can remove the close on the individual APIs now that you're using the provider manager - although, it doesn't hurt anything if they are called.

mziccard: You are right. Although I would keep the `close()` calls in case we end up adding other Api-scoped closeables.

garrettjonesgoogle: Sounds good.

PR #1380, opened 2016-11-09
Expose static method to get default project ID

### LINKED ISSUE #1207: Google Cloud Dataflow/Apache Beam and gcloud-java-core

mziccard:
This replaces #1241 and fixes #1207

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,12 +105,12 @@ Most `google-cloud` libraries require a project ID.  There are multiple ways to 
 
 `google-cloud` determines the project ID from the following sources in the listed order, stopping once it finds a value:
 
-1. Project ID supplied when building the service options
+1. The project ID supplied when building the service options
 2. Project ID specified by the environment variable `GOOGLE_CLOUD_PROJECT`
-3. App Engine project ID
-4. Project ID specified in the JSON credentials file pointed by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
-5. Google Cloud SDK project ID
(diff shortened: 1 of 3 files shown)


lukecwik:
Any progress?

lukecwik:
@garrettjonesgoogle ping?

garrettjonesgoogle:
I can take care of merging this. @lukecwik , I think @mziccard cc'd you to see if you had feedback - does it look good as is?

lukecwik:
I had reviewed #1241 previously which was a precursor to this. Those comments that were relevant to this change were already fixed so this looks good to me.

PR #1384, opened 2016-11-09
More cloud storage examples

jerjou:
Sample for making GCS object public-read
Add sample for moving blob.
@mziccard FYI

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/BlobSnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/BlobSnippets.java
@@ -173,6 +173,25 @@ public class BlobSnippets {
   }
 
   /**
+   * Example of moving a blob to a different bucket with a different name.
+   */
+  // [TARGET copyTo(String, String, BlobSourceOption...)]
+  // [VARIABLE "my_unique_bucket"]
+  // [VARIABLE "move_blob_name"]
+  public Blob moveTo(String destBucket, String destBlob) {
+    // [START storageMoveFile]
+    CopyWriter copyWriter = blob.copyTo(destBucket, destBlob);
+    Blob copiedBlob = copyWriter.getResult();
+    boolean deleted = blob.delete();
+    // [END storageMoveFile]
+    if (deleted) {
+      return copiedBlob;
+    } else {
+      return null;
+    }
+  }
+
+  /**
    * Example of reading the blob's content through a reader.
    */
   // [TARGET reader(BlobSourceOption...)]
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -591,6 +591,15 @@ public class Blob extends BlobInfo {
    * Blob copiedBlob = copyWriter.getResult();
    * }</pre>
    *
+   * <p>Example of moving a blob to a different bucket with a different name.
+   * <pre> {@code
+   * String destBucket = "my_unique_bucket";
+   * String destBlob = "move_blob_name";
+   * CopyWriter copyWriter = blob.copyTo(destBucket, destBlob);
+   * Blob copiedBlob = copyWriter.getResult();
+   * boolean deleted = blob.delete();
+   * }</pre>
+   *
    * @param targetBucket target bucket's name
    * @param targetBlob target blob's name
    * @param options source blob options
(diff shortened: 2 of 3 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/BlobSnippets.java
```diff
@@ -173,6 +173,22 @@ public Blob copyToStrings(String bucketName, String blobName) {
   }
 
   /**
+   * Example of moving a blob to a different bucket with a different name.
+   */
+  public Blob moveTo(Blob srcBlob, String destBucket, String destBlob) {
```
mziccard: You should use the `blob` field instead of passing a blob parameter.

This method is also missing `TARGET` and `VARIABLE` tags:

```
// [TARGET copyTo(String, String, BlobSourceOption...)]
// [VARIABLE "my_unique_bucket"]
// [VARIABLE "copy_blob_name"]
```

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -543,6 +543,13 @@ public boolean delete(BlobSourceOption... options) {
    * Blob copiedBlob = copyWriter.getResult();
    * }</pre>
    *
+   * <p>Example of moving a blob to a different bucket with a different name.
+   * <pre> {@code
+   * CopyWriter copyWriter = srcBlob.copyTo(destBucket, destBlob);
+   * Blob copiedBlob = copyWriter.getResult();
+   * boolean deleted = srcBlob.delete();
+   * }</pre>
+   *
```
mziccard: To add snippets to the javadoc use the script in `utilities/add_snippets_to_file.py` (it will check that all the tags are ther and fail otherwise).
In this case from the root of the project do:
```
python ./utilities/add_snippets_to_file.py google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/BlobSnippets.java google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
python ./utilities/add_snippets_to_file.py google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/BlobSnippets.java
jerjou: > You should use the blob field instead of passing a blob parameter.

It's just problematic because moving the blob includes deleting it, which makes testing it murkier (the way the tests are structured right now). ie if I move the `blob` field while testing `moveTo`, then the test for `delete` has nothing to delete. I figured my options were:

1. Move a blob that's created via one of the `copyTo` tests instead of the canonical `blob`
1. Change the structure of the tests such that each method tested is a separate test method, and the blob is created and torn down for each test
1. Create another test that creates, uses, and cleans up its own blob instead of using the one in `beforeClass`.

Clearly I went with the first option, which seemed the simplest. The second option seems like the Right Thing to do, but would involve rewrite a bunch of the other tests, and seems like it'd incur overhead, which I don't know how much your team cares about. The third option seems like a decent compromise between keeping the actual snippets file using the same `blob` convention and keeping changes to the test file to a minimum, but then the test file has inconsistent test conventions, which feels hacky.

Thoughts?

Also - I removed the `TARGET` and `VARIABLE` tags, first because I wasn't sure they were required, but more importantly because I didn't know how to specify a `VARIABLE` for a `Blob` type :-)

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
jerjou: Ah - thanks. Was wondering about that...

Also - it's not a `mvn` target? ;-)

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/BlobSnippets.java
jerjou: Also - I wasn't sure if this sample was super relevant for the javadocs, since it's not about how to use either `copyTo` or `delete`, but how to do a move, which there isn't a method for..

mziccard: > Clearly I went with the first option, which seemed the simplest.

This is a fine option to me, I suggest you do that by creating a new `BlobSnippets` class instead of passing a `blob` parameter, something like:

```Java
BlobSnippets moveBlobSnippets = new CopyBlobSnippets(copiedBlob);
copiedBlob = moveBlobSnippets.moveTo(copiedBlob, BUCKET, "moveBlob");
assertNotNull(copiedBlob);
copiedBlob.delete();
```
> I wasn't sure if this sample was super relevant for the javadocs

Yeah it's not super relevant but we can put it on top of `copyTo` and it won't harm anybody :)

jerjou: Hm... tried to create a new class, but it doesn't look like the javadoc script can handle having multiple snippets classes target a single javadoc file. Ended up just adding it to the existing snippets, and modifying the test to reset the blob between tests. PTAL.

jerjou:
PR updated - PTAL

mziccard:
@jerjou thanks for this!

Issue #1329, opened 2016-10-21
Allow customization for context of LogEntry instances created by LoggingHandler 


gregw:
The https://github.com/GoogleCloudPlatform/jetty-runtime needs to be able to generate stackdriver log messages with the label `traceid` set the the value obtained from the `x-cloud-trace-context` HTTP header.

However, the [LoggingHandler#getLogEntry](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/v0.4.0/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java#L296) method does not allow for easy extension to support traceid as the method is private rather than protected, thus to change behaviour more methods on the class need to be replaced to customize the LogEntry.

If this method could be made protected, then it would be easy to extend the handler to use a threadlocal label map to set additional labels (eg traceid).

Better yet, such a threadlocal context could be implemented as a standard feature of the LogHandler?

mziccard:
> If this method could be made protected, then it would be easy to extend the handler to use a threadlocal label map to set additional labels (eg traceid).

This sounds good to me.

> Better yet, such a threadlocal context could be implemented as a standard feature of the LogHandler?

I am not sure this is a feature that a lot of users would use/need.

gregw:
@mziccard Do you want a PR for the change to protected?

With regards to a threadlocal mechanism, I agree it is not entirely widely applicable, but neither does it have a zero use-case.   Other logging mechanisms (Eg [logback MDC](http://logback.qos.ch/manual/mdc.html) do provide this feature and indeed the stackdrive log API has the map of labels as an extensible mechanism for such logging that is essentially hidden by the JUL handler.

If there is zero interest from you guys for this feature, I'll drop it.  But if there is a hint of interest I'm happy to create a PR to show how it may be done with minimal cost to users that do not use it.

mziccard:
> If there is zero interest from you guys for this feature, I'll drop it. But if there is a hint of interest I'm happy to create a PR to show how it may be done with minimal cost to users that do not use it.

@gregw I'd be happy to have such a PR. If we can make its cost close-to-zero it'd be nice to have.

> @mziccard Do you want a PR for the change to protected?

Should we put this on hold while we evaluate the threadlocal mechanism?

gregw:
[...]

PR #1385, opened 2016-11-10
Issue #1329

gregw:
For #1329 

Allow extended classes to access the LogEntry building, so that
additional information may be encoded in each log entry (eg traceid).

Use addLabel rather than setLabels as it avoids the creation and copy
of a HashMap per log entry.

Signed-off-by: Greg Wilkins <gregw@webtide.com>

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -26,7 +26,6 @@ import com.google.common.collect.ImmutableSet;
 
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.logging.ErrorManager;
 import java.util.logging.Filter;


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -303,15 +303,17 @@ private LogEntry entryFor(LogRecord record) {
...
-        .setSeverity(severityFor(level))
-        .build();
+    LogEntry.Builder builder = LogEntry.newBuilder(Payload.StringPayload.of(payload))
+        .addLabel("levelName", level.getName())
+        .addLabel("levelValue", String.valueOf(level.intValue()))
+        .setSeverity(severityFor(level));
+    return buildEntryFor(record, builder);
   }
 
+  protected LogEntry buildEntryFor(LogRecord record, LogEntry.Builder builder) {
```
mziccard: Isn't adding this method enough for your needs? Why making also `entryFor` protected?

gregw: Correct buildEntryFor is sufficient for our needs.
I made `entryFor` protected as my first approach was going to be to replace that method.  I then left it protected as I thought others might like to extend it.   But we don't need it.   

Shall I remove it from this PR?

meltsufin:
@mziccard Work on the jetty-runtime is blocking on this. Is anything else needed to approve this for merging?

mziccard:
@meltsufin I am no longer involved in this project, please refer to @garrettjonesgoogle for help

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
garrettjonesgoogle: Hi @gregw , I find this signature to be a little unusual. Would you be able to give more details about why it needs to look this way in particular?

garrettjonesgoogle: I think the following would be a better abstraction:

```
protected void enhanceLogEntry(LogEntry.Builder builder, LogRecord record) {
  // no-op in this class
}
```

and do the `.build()` call in `entryFor`.

garrettjonesgoogle:
LGTM if Travis passes

PR #1387, opened 2016-11-10
Use returnImmediately=false and disable timeouts for pullAsync

mziccard:


mziccard:
This fixes #1157

/cc @garrettjonesgoogle @aozarov care to have a look?

garrettjonesgoogle:
LGTM

aozarov:
Looks good to me. One minor comment.

Also, another suggestion that even I am not too keen about (as it typically frown upon) but still would like to mention. Did you consider adding a finalize to this Future to "cancel" the request if was not completed and finalized?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
```diff
@@ -512,18 +512,13 @@ public SubscriptionId apply(String compositeSubscription) {
...
-  @Override
-  public Future<Iterator<ReceivedMessage>> pullAsync(final String subscription, int maxMessages) {
+  private Future<Iterator<ReceivedMessage>> pullAsync(final String subscription,
+      int maxMessages, boolean returnImmediately) {
     PullRequest request = PullRequest.newBuilder().setReturnImmediately(true)
         .setSubscription(
             SubscriberApi.formatSubscriptionName(getOptions().getProjectId(), subscription))
         .setMaxMessages(maxMessages)
-        .setReturnImmediately(true)
+        .setReturnImmediately(returnImmediately)
```
mziccard: Uh sure I missed it

mziccard:
> Did you consider adding a finalize to this Future to "cancel" the request if was not completed and finalized?

Do you think this would be a good choice? I wouldn't rely on `finalize` in general and I also don't want to give users the impression that they can rely on it to do the cleanup. As a matter of fact the service timeouts the request after 90seconds and returns an empty list of messages, so the request will not stay around forever.

aozarov:
> As a matter of fact the service timeouts the request after 90seconds and returns an empty list of messages, so the request will not stay around forever.

Perfect! Then surely there is no need for such (pretty lame) guard.

Feel free to merge after removing the redundant `setReturnImmediately`

PR #1390, opened 2016-11-10
Change BaseWriteChannel.position type to long to fix overflow

### LINKED ISSUE #1389: Integer Overflow In BaseWriteChannel
Hi there. I'm testing out com.google.cloud.storage.Storage.writer() for writing to GCS. It works very well for files smaller than 2GB, but right around the 2GB mark WriteChannel throws an exception indicating that there's something wrong with the Content-Range header.

`com.google.api.client.http.HttpResponseException: 400|Failed to parse Content-Range header.`

I noticed that the Content-Range header is set in com.google.cloud.storage.spi.DefaultStorageRpc.write() and is calculated using com.google.cloud.BaseWriteChannel.position, which is an int. The problem is that at the 2GB mark the position variable overflows and becomes negative, causing panic and mayhem.

Here are some screenshots to paint a fuller picture:

trace1

trace2

Any chance we can get a patch for this? Should I submit a pull request?

mziccard:
This fixes #1389 

@garrettjonesgoogle do you mind having a look?

/cc @Dima1224

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/BaseWriteChannel.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/BaseWriteChannel.java
@@ -41,7 +41,7 @@ public abstract class BaseWriteChannel<
   private final ServiceOptionsT options;
   private final EntityT entity;
   private final String uploadId;
-  private int position;
+  private long position;
   private byte[] buffer = new byte[0];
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
LGTM

PR #1394, opened 2016-11-11
bigquery: BigQuery.writer return type should be public.

tswast:
The `BigQuery.writer` function does not link to its return type in the
javadocs.

http://googlecloudplatform.github.io/google-cloud-java/0.5.1/apidocs/com/google/cloud/bigquery/BigQuery.html#writer(com.google.cloud.bigquery.WriteChannelConfiguration)

I believe this is because `TableDataWriteChannel` is package-private.
Since the snippets use `BaseWriteChannel<BigQueryOptions,
WriteChannelConfiguration>` as the variable type, I changed the type to
this public type rather than making `TableDataWriteChannel` public.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
@@ -18,6 +18,7 @@ package com.google.cloud.bigquery;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.google.cloud.BaseWriteChannel;
 import com.google.cloud.FieldSelector;
 import com.google.cloud.FieldSelector.Helper;
 import com.google.cloud.Page;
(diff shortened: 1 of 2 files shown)


tswast:
@mziccard @lesv Maybe this method should return http://googlecloudplatform.github.io/google-cloud-java/0.5.1/apidocs/com/google/cloud/WriteChannel.html instead? In that case, I'd want to update the snippets, too.

tswast:
Makes sense, though this isn't really a breaking change. No one could have used the return type it had before.

lesv:
If writer is fairly recent, then I'm ok with out the `@Deprecated`, but if it's old, then we should do it, despite it being unused.

tswast:
Closing based on https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1392#issuecomment-260079544

We'll probably make `TableDataWriteChannel` public instead.

PR #1395, opened 2016-11-14
Make TableDataWriteChannel expose Job when upload completed

### LINKED ISSUE #1392: BigQuery: support `/upload` method for loading data from local files.
See documentation: https://cloud.google.com/bigquery/loading-data-post-request

Python, PHP, Ruby, and Node.js provide an upload-from-file method on the Table object.

This is blocking documentation samples.

mziccard:
This fixes #1392


mziccard:
> Can we tell codacy-bot to ignore unused locals in snippets?

It is doable but configuration is not exactly easy.

mziccard:
@garrettjonesgoogle do you want to have a look?

tswast:
@lesv PTAL

lesv:
So the argument for not using `@Deprecated` and creating a new API is that no one could actually have used the earlier API.  Since that's the case, and presuming everyone agrees, LGTM.

PR #1396, opened 2016-11-14
Remove --quiet from mvn site to avoid Travis timeout

mziccard:
Travis times out on `mvn site --quiet -Djava.util.logging.config.file=logging.properties` as no log is produced for more than 10 minutes.

This PR enables back some logging to avoid timeouts.


garrettjonesgoogle:
LGTM if Travis passes :-)

PR #1399, opened 2016-11-14
ACL examples

jerjou:
* Example for getting ACL of bucket for specific user
* Example for getting ACL of blob for specific user

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
@@ -872,6 +872,21 @@ public class StorageSnippets {
   }
 
   /**
+   * Example of getting the ACL entry for a specific user on a blob.
+   */
+  // [TARGET getAcl(BlobId, Entity)]
+  // [VARIABLE "my_unique_bucket"]
+  // [VARIABLE "my_blob_name"]
+  // [VARIABLE "google-cloud-java-tests@java-docs-samples-tests.iam.gserviceaccount.com"]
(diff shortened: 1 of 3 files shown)


jerjou:
@mziccard FYI

garrettjonesgoogle:
@jerjou , you will want to cc me from this point forward.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -718,6 +718,19 @@ public Acl getBucketAcl(String bucketName) {
   }
 
   /**
+   * Example of getting the ACL entry for a specific user on a bucket.
+   */
+  // [TARGET getAcl(String, Entity)]
+  // [VARIABLE "my_unique_bucket"]
+  // [VARIABLE "google-cloud-java-tests@java-docs-samples-tests.iam.gserviceaccount.com"]
+  public Acl getBucketAcl(String bucketName, String userEmail) {
+    // [START storagePrintBucketAclForUser]
```
garrettjonesgoogle: `storagePrintBucketAclForUser` doesn't seem right, is that what you want?

jerjou: Yeah - there's a [nascent effort](go/veneer-docs-samples) to start standardizing on region tag names across languages, and that's what they wanted for this sample. Why? What seems odd about it?

garrettjonesgoogle: Oh it was specifically that it was more different from the snippet name than for other snippets. If it's intentional, then it's fine.

PR #1400, opened 2016-11-14
Encryption examples

jerjou:
Diffbase: #1399 

Adds two commits with examples for encrypting, decrypting, and updating the encryption key on blobs.

Also adds two methods to `Storage.java` for being able to set the target of a copy operation with a new encryption key, but without overwriting the other metadata.


jerjou:
@mziccard FYI

jerjou:
FWIW, the travis tests pass, and the appveyor CI failure seems like it's unrelated to this PR.

PR #1403, opened 2016-11-15
Pubsub resource names

michaelbausor:
Regenerate PubSub with generated resource name types


garrettjonesgoogle:
LGTM once Travis passes

PR #1406, opened 2016-11-16
Update BigQuery.writer snippets to get LoadStatistics from Job.

tswast:
Also, add an example for loading from a file, not just a string.

@lesv @mziccard PTAL


lesv:
Definitely be good to teach Circle to build this.

tswast:
I used Guava in this one, too. :-(

PR #1407, opened 2016-11-16
bigquery: show base64 encode and records in insertAll snippets.

tswast:
Showing a raw base64 string isn't great. Instead, show a way to encode
the bytes to a base64 string using Guava's BaseEncoding class.

Also, it wasn't clear how to insert values of type record, so show that,
too.

@lesv @mziccard PTAL


mziccard:
Using Guava in snippets is generally a bad idea. We shouldn't assume that users are familiar with it nor we should show code that depends on it to work

tswast:
Okay. I wasn't sure about that. I wish we could show imports in snippets.

What about using classes that are Java 8, only? http://docs.oracle.com/javase/8/docs/api/java/util/Base64.html There wasn't a built-in way to encode Base64 until then.

PR #1413, opened 2016-11-18
Using fake clock in AckDeadlineRenewerTest

garrettjonesgoogle:
This reduces the running time of the test from 3-4 minutes
down to 0.03 seconds.

### DIFF
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/FakeScheduledExecutorService.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/FakeScheduledExecutorService.java
@@ -19,7 +19,6 @@ package com.google.cloud.pubsub;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Vector;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Delayed;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/FakeScheduledExecutorService.java
```diff
@@ -0,0 +1,183 @@
...
+
+      if (exception != null) {
+        throw new ExecutionException(exception);
+      }
+
+      return result;
+    }
+
+    @Override
+    public Object get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException,
```
michaelbausor: Is this unused? Perhaps it could throw UnsupportedOperationException, like the unused schedule methods?

garrettjonesgoogle: I don't think it's ever called. But, since the main `get()` doesn't support blocking (i.e. when isDone = false), from that perspective, the timeout doesn't matter and this is a consistent implementation.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/FakeScheduledExecutorService.java
```diff
@@ -0,0 +1,183 @@
...
+      scheduledFuture.run();
+    }
+  }
+
+  private boolean cancel(FakeScheduledFuture toCancel) {
+    synchronized (this) {
+      Iterator<FakeScheduledFuture> iter = futures.iterator();
+      while (iter.hasNext()) {
+        FakeScheduledFuture scheduledFuture = iter.next();
+        if (scheduledFuture == toCancel) {
```
garrettjonesgoogle: This is intentional, I want reference equality.

PR #1414, opened 2016-11-19
Multithreaded prefetcher option for newReadChannel

jean-philippe-martin:
Benefits:
 - the caller can run computation in between read calls, and data is
   being fetched in parallel with that computation.
 - the caller can issue many small sequential reads, and the system
   instead sends larger reads, which is more efficient.

NIO gets better performance when issuing parallel requests, as we already know from the ParallelCountBytes example. This pull request adds a "SeekableByteChannelPrefetcherOptions" that we can pass to newReadChannel to enable parallel prefetching without the NIO library, so the caller gets the benefits of parallel requests without having to change the client code at all.

The PR also updates the tests so they check the prefetching code path in addition to the normal, non-prefetching one.

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/SeekableByteChannelPrefetcher.java
+++ b/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/SeekableByteChannelPrefetcher.java
@@ -191,6 +191,7 @@ public class SeekableByteChannelPrefetcher implements SeekableByteChannel {
     try {
       src = fetch(position);
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
       return 0;
     } catch (ExecutionException e) {
       throw new RuntimeException(e);
@@ -341,11 +342,12 @@ public class SeekableByteChannelPrefetcher implements SeekableByteChannel {
           }
         }
       } catch (InterruptedException e) {
-        System.out.println("Timed out while waiting for channels to close.");
+        Thread.currentThread().interrupt();
       }
       try {
         exec.awaitTermination(60, TimeUnit.SECONDS);
       } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
         exec.shutdownNow();
       }
       // Close all underlying channels


jean-philippe-martin:
Are we now requiring javadoc @return on all methods? It looks like this PR is failing on Travis because of (unrelated) missing Javadoc (?)

> [ERROR] Exit code: 1 - /Users/travis/build/GoogleCloudPlatform/google-cloud-java/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageConfiguration.java:36: warning: no @return
[ERROR] public abstract String workingDirectory();

jean-philippe-martin:
OK I've addressed most of the automatic review comments. Would like guidance as to what to throw for errors.

jean-philippe-martin:
Hopefully someone will be looking at this code soon? The code was initially submitted 12 days ago.

garrettjonesgoogle:
I will take a look at it. We are very resource-constrained at the moment.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -0,0 +1,474 @@
...
+            synchronized (idleWorkers) {
+            if (idleWorkers.size() >= workerCount) {
+              // every thread is idle, we're done.
+              break;
+            }
+            idleWorkers.wait(60_000);
+          }
+        }
+      } catch (InterruptedException e) {
+        System.out.println("Timed out while waiting for channels to close.");
```
garrettjonesgoogle: You need to either rethrow the InterruptedException or set the interrupted flag to true.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/SeekableByteChannelPrefetcher.java
```diff
@@ -0,0 +1,474 @@
...
+      }
+      try {
+        exec.awaitTermination(60, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        exec.shutdownNow();
```
garrettjonesgoogle: Either rethrow the InterruptedException or set the interrupted flag to true.

jean-philippe-martin:
Thank you, @garrettjonesgoogle, excellent! I was just worried that my PR has gone unnoticed, somehow.

jean-philippe-martin:
@garrettjonesgoogle I understand you're resource constrained. Could you review my shorter PRs before this one please? This may help us make the best of the situation.

jean-philippe-martin:
Thanks for the reviews on the smaller PRs! I'm going to close this one as I think we can do without it, and it'll free up some of your resources. We can reopen if we change our minds later.

PR #1415, opened 2016-11-19
Fix broken URL Links in documentation

jweaver:
- Looks like possibly commit 6fd7163 changed an external link that
   should have stayed the same, for gcloud-java-examples.
 - Other links just looked to be outdated, either due to refactors or
   external link changes.

-------

I found one broken link, and figured while I'm at it, might as well find any others too, and found a couple more.  I know it's small nits here, but whenever I come across a dead link I figure I might as well send a PR for it.


garrettjonesgoogle:
Nice, thanks!

PR #1416, opened 2016-11-21
Regenerate spi layer

michaelbausor:
Change ListLogEntries from projectIds to resourceIds
Remove incorrect format and parse methods
Minor format changes


garrettjonesgoogle:
LGTM once tests pass

PR #1417, opened 2016-11-22
SPI layer: renaming Api to Client

garrettjonesgoogle:
Also:

* Marking all SPI classes as `@ExperimentalApi`
* Updating the Language API from v1beta1 to v1


michaelbausor:
LGTM

PR #1436, opened 2016-11-30
Make defaultLibraryVersion() more robust

shinfan:
`defaultLibraryVersion()` now reads version from pom.properties

Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1434
Tested manually.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -682,20 +682,7 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
   private static String defaultLibraryVersion() {
     String version = getPomVersion();
     if (version == null) {
-      try {
-        Enumeration<URL> resources =
-            ServiceOptions.class.getClassLoader().getResources(JarFile.MANIFEST_NAME);
-        while (resources.hasMoreElements() && version == null) {
-          Manifest manifest = new Manifest(resources.nextElement().openStream());
-          Attributes manifestAttributes = manifest.getMainAttributes();
-          String artifactId = manifestAttributes.getValue(MANIFEST_ARTIFACT_ID_KEY);
-          if (artifactId != null && artifactId.equals(ARTIFACT_ID)) {
-            version = manifestAttributes.getValue(MANIFEST_VERSION_KEY);
-          }
-        }
-      } catch (IOException e) {
-        // ignore
-      }
+      version = getManifestVersion();
     }
     return version;
   }
@@ -715,5 +702,24 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
       // ignore
     }
     return null;
+  }
+
+  private static String getManifestVersion() {
+    String version = null;
+    try {
+      Enumeration<URL> resources =
+          ServiceOptions.class.getClassLoader().getResources(JarFile.MANIFEST_NAME);
+      while (resources.hasMoreElements() && version == null) {
+        Manifest manifest = new Manifest(resources.nextElement().openStream());
+        Attributes manifestAttributes = manifest.getMainAttributes();
+        String artifactId = manifestAttributes.getValue(MANIFEST_ARTIFACT_ID_KEY);
+        if (artifactId != null && artifactId.equals(ARTIFACT_ID)) {
+          version = manifestAttributes.getValue(MANIFEST_VERSION_KEY);
+        }
+      }
+    } catch (IOException e) {
+      // ignore
+    }
+    return version;
   }
 }


sai-pullabhotla:
I don't have a whole lot of background on this, but I'm curious - why can't we use something as simple as - 

`ServiceOptions.class.getPackage().getImplementationVersion();`

That should do the work of getting the implementation version from Jar's metadata.

shinfan:
@sai-pullabhotla I tried this approach but it did not work for me (the method simply returns null).

sai-pullabhotla:
@shinfan - as long as the class is loaded from JAR and JAR has manifest with the Implementation-Version defined, it should work. Below is what I get: 

```
Package pack = ServiceOptions.class.getPackage();
System.out.println("Impl Title: " + pack.getImplementationTitle());
System.out.println("Impl Version: " + pack.getImplementationVersion());
System.out.println("Spec Title: " + pack.getSpecificationTitle());
System.out.println("Spec Version: " + pack.getSpecificationVersion());

```
### Output

```
Impl Title: Google Cloud Core
Impl Version: 0.7.0
Spec Title: Google Cloud Core
Spec Version: 0.7.0
```

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -676,21 +680,40 @@ protected RetryParams defaultRetryParams() {
...
+        Properties properties = new Properties();
+        String mavenPropertiesPath = META_FILE_ROOT
+            + ServiceOptions.class.getPackage().getName() + "/"
+            + ARTIFACT_ID + "/pom.properties";
+        InputStream inputStream = ServiceOptions.class.getResourceAsStream(mavenPropertiesPath);
+        if (inputStream != null) {
+            properties.load(inputStream);
+            return properties.getProperty(META_VERSION_KEY, "");
+        }
+    } catch (Exception e) {
```
garrettjonesgoogle: inconsistent indentation

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -676,21 +680,40 @@ protected RetryParams defaultRetryParams() {
...
+      } catch (IOException e) {
+        // ignore
       }
-    } catch (IOException e) {
-      // ignore
     }
     return version;
   }
+
+  private static String getMavenVersion() {
```
garrettjonesgoogle: Better name: getPomVersion (we don't care what version of maven was used)

garrettjonesgoogle:
LIBRARY_NAME should be changed to "google-cloud-java". There are also a couple comments that reference the old name ("gcloud-java").

shinfan:
@sai-pullabhotla As I said, The method did not work under my environment and it returned null. If it happens to me it may happen to other users. My best guess is that my maven does not insert the correct manifest data into the JAR.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java

shinfan:
@garrettjonesgoogle  PTAL

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -676,21 +680,40 @@ protected RetryParams defaultRetryParams() {
...
-          ServiceOptions.class.getClassLoader().getResources(JarFile.MANIFEST_NAME);
-      while (resources.hasMoreElements() && version == null) {
-        Manifest manifest = new Manifest(resources.nextElement().openStream());
-        Attributes manifestAttributes = manifest.getMainAttributes();
-        String artifactId = manifestAttributes.getValue(MANIFEST_ARTIFACT_ID_KEY);
-        if (artifactId != null && artifactId.equals(ARTIFACT_ID)) {
-          version = manifestAttributes.getValue(MANIFEST_VERSION_KEY);
+    String version = getPomVersion();
+    if (version == null) {
+      try {
```

garrettjonesgoogle:
LGTM after Travis passes

shinfan:
Travis passed for oraclejdk7/8 and openjdk. There is one "no_language_set" pending for 30 mins. 
AppVeyor CI failed but I believe it is caused by flakiness.  I will go ahead and merge if no objection @garrettjonesgoogle

garrettjonesgoogle:
The "no_language_set" one is for Mac. I have often seen it take longer. I have no concerns about your CL on mac so go ahead and merge it.

mziccard:
> LIBRARY_NAME should be changed to "google-cloud-java". There are also a couple comments that reference the old name ("gcloud-java").

@garrettjonesgoogle did you check this with @omaray and @jgeewax? Last time we discussed this (~2months ago) it was suggested to stick to `gcloud-java` for the time being.

I feel like a creep when I step in like this :)

garrettjonesgoogle:
No, your feedback is always welcome :-)

We have agreed internally to change all languages to `google-cloud-java` for this value.

garrettjonesgoogle:
I checked again internally and I was wrong - this needs to be changed back to `gcloud-java`. @shinfan , could you fix that up? Sorry for the churn.

shinfan:
Acknowledged. Will do.

PR #1442, opened 2016-12-02
Pub/Sub High Perf client

davidtorres:
This is just for us to discuss, I just realized some conflicts with classes that are already in the repo that happened to have the same name as mine that need to be solved. 

Also need to discuss the dependencies introduced  and the data model (PubsubMessage vs Message)


davidtorres:
Thanks for merging it

PR #1444, opened 2016-12-05
Revert library name to gcloud-java

shinfan:


garrettjonesgoogle:
LGTM after tests pass

lesv:
Why was this done?

garrettjonesgoogle:
The internal tracking systems are all set up to track "gcloud-java" (and gcloud-* for other languages), and it would be some effort to switch them all over.

Issue #1388, opened 2016-11-10
Add support for DATE, TIME, and DATETIME schema types


esheffield:
BigQuery allows creation of tables with fields of type DATE, TIME, and DATETIME. However the Java client does not recognize these types. So for example, executing a load job to import JSON into a table an existing table with one of these field types in the schema will fail.

Schema:
```
    ItemDate:DATE,Description:STRING
```

JSON:
```
{"ItemDate": "2016-11-01", "Description": "Item 1"}
{"ItemDate": "2016-11-02", "Description": "Item 2"}
{"ItemDate": "2016-11-03", "Description": "Item 3"}
{"ItemDate": "2016-11-04", "Description": "Item 4"}
```

Create job: (Example from Scala repl but using the Java client)
```
val tableId = TableId.of("instant-bonbon-111304", "TestData", "Date_Test")
val uris = new java.util.ArrayList[String]()
uris.add("gs://ih_test_v1/date_test/date_test.json")

val jobConf = LoadJobConfiguration.newBuilder(tableId, uris).
  setWriteDisposition(JobInfo.WriteDisposition.WRITE_APPEND).
  setFormatOptions(FormatOptions.json).
  build

val jobInfo = JobInfo.of(jobConf)

var job : Job = null

try {
  job = bigquery.create(jobInfo)
} catch {
  case e: BigQueryException => println(e)
}
```

Error:
```
java.lang.IllegalArgumentException: No enum constant com.google.cloud.bigquery.Field.Type.Value.DATE
        at java.lang.Enum.valueOf(Enum.java:238)
        at com.google.cloud.bigquery.Field$Type$Value.valueOf(Field.java:75)
        at com.google.cloud.bigquery.Field.fromPb(Field.java:477)
        at com.google.cloud.bigquery.Field$1.apply(Field.java:46)
        at com.google.cloud.bigquery.Field$1.apply(Field.java:43)
        at com.google.common.collect.Lists$TransformingRandomAccessList.get(Lists.java:451)
        at java.util.AbstractList$Itr.next(AbstractList.java:358)
        at java.util.AbstractCollection.toArray(AbstractCollection.java:141)
        at java.util.ArrayList.<init>(ArrayList.java:177)
        at com.google.common.collect.Lists.newArrayList(Lists.java:119)
        at com.google.cloud.bigquery.Schema$Builder.setFields(Schema.java:86)
        at com.google.cloud.bigquery.Schema.of(Schema.java:189)
[...]

PR #1451, opened 2016-12-06
Adding QueryParameter support in BigQuery

### LINKED ISSUE #1411: BigQuery: Add support for query parameters
https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#request-body

A query can put placeholders in the query text, which can then be filled in with `queryParameters`. Depending on what `parameterMode` is set to, the parameter can be positional `?` or names `@somename`.

I'd expect these to be available on the [QueryRequest](http://googlecloudplatform.github.io/google-cloud-java/0.6.0/apidocs/com/google/cloud/bigquery/QueryRequest.html).

See also: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2551

garrettjonesgoogle:
Also, adding DATE, TIME, and DATETIME support for legacy SQL types.

Fixes #1411


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameter.java
```diff
@@ -0,0 +1,128 @@
...
+  private final String name;
+  private final QueryParameterValue parameterValue;
+
+  private QueryParameter(String name, QueryParameterValue parameterValue) {
+    this.name = name;
+    this.parameterValue = checkNotNull(parameterValue);
+  }
+
+  /** Returns the query parameter name. */
+  public String getName() {
```
lesv: Since this is public, shouldn't there be a full JavaDoc?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameter.java
```diff
@@ -0,0 +1,128 @@
...
+    return name;
+  }
+
+  /** Returns the typed query parameter value. */
+  public QueryParameterValue getParameterValue() {
```
lesv: ditto

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameter.java
aozarov: I think javadoc here is fine (though as said, I would mention the null case).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequest.java
```diff
@@ -102,6 +106,33 @@ public Builder setQuery(String query) {
...
 
     /**
+     * Adds a QueryParameter to the list of query parameters. See
+     * {@link #setQueryParameters(Iterable)} for more details on the input requirements.
+     */
+    public Builder addQueryParameter(QueryParameter queryParameter) {
+      if (queryParameters == null) {
+        queryParameters = Lists.newArrayList();
+      }
+      queryParameters.add(checkNotNull(queryParameter));
```
aozarov: Nit, but should we have an empty list if the given parameter was null (and therefore rejected)?
Looking at the code I see that you handle it later, so a very minor nit which could be ignored.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequest.java
```diff
@@ -102,6 +106,33 @@ public Builder setQuery(String query) {
...
+      return this;
+    }
+
+    /**
+     * Sets the query parameters for the query.
+     *
+     * The set of query parameters must either be all positional or all named parameters.
+     * Positional parameters are denoted in the query with a question mark (?), and named
```
aozarov: We could enforce that during compilation by having separate classes  for positional and name (could inherit from the same base class if needed). Then we can have separate method for adding them.
Also, not compile time but the single case of addQueryParameter (which also should have 2 overloaded methods) could throw IllegalArgumentException when we already have non empty list of the other type (positional or named).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -0,0 +1,433 @@
...
+
+    /** Creates a {@code QueryParameterValue} object. */
+    public QueryParameterValue build() {
+      return new QueryParameterValue(this);
+    }
+  }
+
+  private QueryParameterValue(Builder builder) {
+    checkArgument(
+        (builder.arrayValues != null) == (StandardSQLTypeName.ARRAY.equals(builder.type)));
```
aozarov: It would be nice to add a message for less obvious checkArgument

garrettjonesgoogle:
I decided to add the missing types TIME/DATE/DATETIME into LegacySQLTypeName to resolve #1388 .

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
aozarov: I think it would have been more readable if it was written this way:
if (builder.arrayValues != null) {
   ... apply array precondition with relevant description of failure
} else {
  ... apply non array preconditions with relevant description of failure
}

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/LegacySQLTypeName.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
garrettjonesgoogle: will do.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/LegacySQLTypeName.java
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigquery;
+
+/**
+ * A type used in legacy SQL contexts. For example, these types are used in query results,
+ * even if the query is a standard SQL query.
```
garrettjonesgoogle: The wording here isn't the best - I will make it clearer. I will also add the `@see` link.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameter.java
garrettjonesgoogle: Will do.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameter.java
```diff
@@ -0,0 +1,128 @@
...
+   */
+  public static QueryParameter of(QueryParameterValue value) {
+    return new QueryParameter(null, value);
+  }
+
+  /**
+   * Creates a named {@code QueryParameter} object with the given value.
+   */
+  public static QueryParameter named(String name, QueryParameterValue value) {
+    return new QueryParameter(name, value);
```
garrettjonesgoogle: I'll add that.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
garrettjonesgoogle: Ok I'll arrange it like this, and add messages for checkArgument failures.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequest.java
garrettjonesgoogle: Easy fix - I'll do the checkNotNull call first.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequest.java
garrettjonesgoogle: Do you think this provides value in this case? An issue is that half of what needs to be done is a textual query, which also has to be consistent with itself and with these parameters, and can't be checked at compile time. Someone could pass the compile check (and even runtime check) for the query parameters, but still have it rejected by the server for inconsistency. 

If we were going to go the route of enforcing consistency here, I might prefer to just hide the concept of QueryParameter and do something here like:

```
public Builder addPositionalParameter(QueryParameterValue value);
public Builder setPositionalParameters(Iterable<QueryParameterValue> values);

public Builder addNamedParameter(String name, QueryParameterValue value);
public Builder setNamedParameters(Map<String, QueryParameterValue> values);
```

How important do you think this is to address?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/StandardSQLTypeName.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigquery;
+
+/**
+ * A type used in standard SQL contexts. For example, these types are used in queries
+ * with query parameters, which requires usage of standard SQL.
+ */
```
garrettjonesgoogle: ok I'll add that.

aozarov:
> I decided to add the missing types TIME/DATE/DATETIME into LegacySQLTypeName to resolve #1388 

Update PR description?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequest.java
aozarov: Indeed query must match and there is nothing we can do about it, however we could help with mixing them if that is not supported and also make it a cleaner API by not having an object with attributes that are not relevant and will always be null in that case.

I am fine with your suggestion if you plan to return Map<String, QueryParameterValue> for getNamedParameters and List<QueryParameterValue> for positional. I would prefer that over the current approach.

garrettjonesgoogle:
Description updated.

garrettjonesgoogle:
Review comments addressed - PTAL

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -0,0 +1,439 @@
...
+    }
+    return QueryParameterValue.newBuilder()
+        .setArrayValues(listValues)
+        .setType(StandardSQLTypeName.ARRAY)
+        .setArrayType(type)
+        .build();
+  }
+
+  private static <T> StandardSQLTypeName classToType(Class<T> type) {
+    if (Boolean.class.isAssignableFrom(type)) {
```
aozarov: Fine as is though because all types here are final you could have kept a private immutablemap<Class, StandardSQLTypeName> and use a simple map.get(type) to get the StandardSQLTypeName value.

Also, any reason for not supporting here date, time and timestamp (e.g. by using the java.sql matching types)?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryRequest.java
```diff
@@ -102,6 +115,88 @@ public Builder setQuery(String query) {
...
+    }
+
+    /**
+     * Adds a named query parameter to the set of query parameters. See
+     * {@link #setNamedParameters(Map)} for more details on the input requirements.
+     *
+     * <p>A named parameter cannot be added after positional parameters have been added.
+     */
+    public Builder addNamedParameter(String name, QueryParameterValue value) {
+      checkNotNull(value);
```
aozarov: Nit, would be better to check name not null as well here (rather than to get an exception later when added to namedParameters).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
garrettjonesgoogle: The java.sql types aren't referenced anywhere else in `google-cloud-java`. Time-related data seems to only be exposed with a Long representing epoch in microseconds, but Long is already taken for Int64. Do you know why Date is not used?

aozarov: java.sql are part of the Java SDK. I am not sure why not used (lots of changes done to the bigquery client that I am unaware of). If you used Date what type will you pick (date, time, timestamp)?

garrettjonesgoogle: I'm not sure. I think that ambiguity makes it preferable to use the other entry point where the StandardSQLTypeName is specified explicitly.

aozarov: Ok, as said the java.sql do have a matching classes but it is fine to not supportit if documented.

I think you would be better of documenting such behavior in the method[s]  `of(T value, Class<T> type)` and `array(T[] array, Class<T> clazz)`  rather than in the class level javadoc.

garrettjonesgoogle: Ok that makes sense, I can move it.

Issue #1429, opened 2016-11-27
Integration tests flaky in AppVeyor


lesv:
#1421 AppVeyor failed on DNS for this, succeeded for Travis, I'm going to approve this.

```
Tests run: 35, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.391 sec <<< FAILURE! - in com.google.cloud.dns.testing.LocalDnsHelperTest
testListRecordSets(com.google.cloud.dns.testing.LocalDnsHelperTest)  Time elapsed: 0.032 sec  <<< FAILURE!
java.lang.AssertionError
	at com.google.cloud.dns.testing.LocalDnsHelperTest.testListRecordSets(LocalDnsHelperTest.java:822)

    assertNull(record.getType());
    assertNotNull(record.getTtl());
    assertNull(listResult.pageToken());
```

`getTtl()` appears to be Null.

garrettjonesgoogle:
This seems to have been happening on every PR.

garrettjonesgoogle:
We need to either fix the test or stop making this part of the build; it looks really bad to have a Beta product whose builds are failing. /cc @omaray

pongad:
I'll look into this.

pongad:
@garrettjonesgoogle I think this test file should simply be deleted. The goal of this repo is to create good client libraries. This file is not testing the DNS library. It's testing the DNS emulator, which was written just to test the client.

I think the emulator is a maintenance burden. It's an HTTP server, meaning it has to deal with synchronization. It also tries to behave like the DNS service. Arguably, it's more complicated than the code it's trying to test.

It's much easier to test the library by making sure that method calls create the right RPC requests and that RPC responses gets translated to the right response objects. We can integration test by calling a couple methods on the real service to make sure we set up HTTP and Auth layers properly. I think this would lead to much better tests.

garrettjonesgoogle:
I looked through more of the AppVeyor failures. It does look like about half are the particular failure mentioned by @lesv , but I also see other ones too. They occur in ITTranslateTest, ITLoggingTest, LocalPubSubHelperTest, ITBigQueryTest, and BlockingProcessStreamReaderTest. So, just disabling LocalDnsHelperTest won't help.  @mziccard , any insights? It seems like there is probably a fair amount of work necessary to work through all of the failures happening. 

* https://ci.appveyor.com/project/GoogleCloudPlatform/google-cloud-java/build/276 :

```
[...]

PR #1453, opened 2016-12-07
pubsub: fix testStartStopReset

pongad:
The test verifies that calling stop() actually stops the emulator.
The previous implementation stops the emulator by
- sending HTTP POST to /shutdown
- terminating the emulator process
The second step follows directly after the first.
It then queries the server it just shut down to make sure it does not
respond.

This behavior is incorrect for two reasons.
- The server might continue serving request after responding to
  /shutdown.
- The "emulator process" might not be the emulator at all.
  It might be a script that in-turn runs the emulator.
  On UNIX systems, killing a parent process does not kill any of its
  children; the actual emulator might run indefinitely after
  we kill the process we spawned.

This commit changes this behavior.
After POSTing to /shutdown, we simply wait for the spawned process
to terminate on its own.

Technically, there is still a race:
Another emulator could have been started on the same port.
When we send a request to verify that the old emulator has terminated,
we might get a response from the new one.
Since we execute tests sequentially though,
I believe this should not be a problem in practice.

Updates #1429.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
@@ -18,6 +18,8 @@ package com.google.cloud.testing;
 
 import com.google.cloud.ServiceOptions;
 import com.google.common.io.CharStreams;
+import com.google.common.util.concurrent.SettableFuture;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
@@ -43,10 +45,14 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
+import org.joda.time.Duration;
 
 /**
  * Utility class to start and stop a local service which is used by unit testing.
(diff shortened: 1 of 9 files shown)


google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -337,9 +337,8 @@ public void start() throws IOException {
     }
 
     @Override
-    public void stop() throws InterruptedException {
+    public void waitFor() throws InterruptedException {
       if (process != null) {
-        process.destroy();
         process.waitFor();
```
garrettjonesgoogle: This puts us at the mercy of the emulator exiting. I'd like to see a timeout so that we don't wait forever if the emulator ignores us or locks up. Unfortunately, it might be a tiny bit of a pain to do this in Java 7:

http://stackoverflow.com/questions/808276/how-to-add-a-timeout-value-when-using-javas-runtime-exec

pongad:
@garrettjonesgoogle PTAL, we now raises an exception if process has not finished in specified time.

google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -104,16 +109,50 @@ protected final void startProcess(String blockUntilOutput)
...
+      @Override
+      public void run() {
+        try {
+          exitValue.set(process.waitFor());
+        } catch (InterruptedException e) {
+          exitValue.setException(e);
+        }
+      }
+    });
+    waiter.run();
```
garrettjonesgoogle: I think Codacy caught a bug here - won't this execute your "threaded" code in the thread calling `run()`?

garrettjonesgoogle: Nevermind, you already fixed it.

PR #1454, opened 2016-12-07
Regenerate spi layer and update

michaelbausor:
- Regenerate spi layer with resource names
- Update to use resource names instead of formatting methods


garrettjonesgoogle:
LGTM after tests pass

Issue #1450, opened 2016-12-06
[NIO] Make CloudStorageFileSystem#newFileSystem more lenient in the URIs it accepts


tomwhite:
Currently the URIs must be of the form `gs://bucket`, and URIs with port, user info, path, query or fragment parts will fail.

This seems to be unnecessarily restrictive, and causes legitimate uses to fail - e.g. if we are [trying to find a FileSystem for a given URI](https://github.com/broadinstitute/gatk/blob/5e17764f74fdf110d4ea09cc0b5508fbad9a1305/src/main/java/org/broadinstitute/hellbender/utils/io/IOUtils.java) then we shouldn't have to know that GS is more restrictive than other providers, and remove everything from the URI except the scheme and host.

I suggest that the implementation accept the extra information in URIs and just ignore it.

jean-philippe-martin:
Fixed in #1470.

PR #1470, opened 2016-12-09
Allow path in URIs passed to newFileSystem

jean-philippe-martin:
This makes it easier for users who start with a URI describing a full
path to get a FileSystem that can work with that path, since they no
longer have to needlessly remove the path from the URI.

Note that Oracle's description of newFileSystem [1] puts no restriction
on the passed URI.

Addresses #1450.

[1]
https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystems.html#newFileSystem(java.net.URI,%20java.util.Map)

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
+++ b/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
@@ -170,7 +170,8 @@ public final class CloudStorageFileSystemProvider extends FileSystemProvider {
   }
 
   /**
-   * Returns Cloud Storage file system, provided a URI with no path, e.g. {@code gs://bucket}.
+   * Returns Cloud Storage file system, provided a URI, e.g. {@code gs://bucket}.
+   * The URI can include a path component (that will be ignored).
    *
    * @param uri bucket and current working directory, e.g. {@code gs://bucket}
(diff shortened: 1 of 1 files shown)


google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
```diff
@@ -176,8 +176,8 @@ public CloudStorageFileSystem getFileSystem(URI uri) {
    * @param env map of configuration options, whose keys correspond to the method names of
    *     {@link CloudStorageConfiguration.Builder}. However you are not allowed to set the working
    *     directory, as that should be provided in the {@code uri}
-   * @throws IllegalArgumentException if {@code uri} specifies a user, query, fragment, or scheme is
-   *     not {@value CloudStorageFileSystem#URI_SCHEME}
+   * @throws IllegalArgumentException if {@code uri} specifies a port, user, query, or fragment, or
+   *     if scheme is not {@value CloudStorageFileSystem#URI_SCHEME}
```
garrettjonesgoogle: The first line of this doc comment still says "URI with no path" - could you update that wording? Maybe say something about the path being ignored.

jean-philippe-martin: Good catch, thank you! Fixed.

PR #1473, opened 2016-12-12
fix more races in pubsub tests

pongad:
Previously BlockingProcessStreamReader has a terminate() method,
used to tell the Reader to stop reading from the emulator process.

This causes an inter-process race.
If the Reader stops before reading emulator's output,
the emulator process will hang as it tries to write to stdout/stderr
as there's no one to read from the other side of the pipe.

Since there is no way to safely stop the Reader,
this commit deletes the method and its associated test.

Additionally, the timeout for LocalSystemTest is increased to 3 minutes,
since the emulator, somehow, consistently takes just longer than a
minute to shut down.


garrettjonesgoogle:
LGTM

PR #1475, opened 2016-12-13
DO NOT SUBMIT; testing how deleting LocalSystemTest affects coverage

pongad:


pongad:
@garrettjonesgoogle Are you OK with this change in coverage? The one big change in coverage is in `DefaultPubSubRpc`. Most of the methods are just calling equivalent methods in `Publisher` or `SubscriberClient`. Tests might not reveal much anyway.

One way to test these methods is to mock the clients we generate from toolkit. Currently, this is not possible.
- All methods of the clients are final, so they cannot be mocked.
- We need to mock `UnaryCallable`, so that we can return `Future`s that is prepared for the test. We can't, because `UnaryCallable` is final.

I'm not sure how serious the problem is, but this means our users cannot mock our clients to test their code.

garrettjonesgoogle:
I'd like to see the code covered. I suggest trying to use the auto-generated mock server.

PR #1479, opened 2016-12-15
use the term "bundling" instead of "batching"

pongad:


garrettjonesgoogle:
LGTM - let @davidtorres take a look too .

PR #1480, opened 2016-12-15
use fewer guava classes in public surface

pongad:
- SubscriberImpl used to expose AbstractService.
  We just make this package-private.
- SubscriberStats defines its own Stats type.
- SubsriberStats and PublisherStats are implemented with AutoValue.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/SubscriberStats.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/SubscriberStats.java
@@ -16,33 +16,67 @@
 
 package com.google.cloud.pubsub;
 
-import com.google.auto.value.AutoValue;
+import javax.annotation.concurrent.Immutable;
 
 /**
  * A snapshot of the subscriber statistics at the time they were requested from the {@link
  * Subscriber}.
  */
 //TODO: Finish implementation.
-@AutoValue
-public abstract class SubscriberStats {
-  @AutoValue
-  public static abstract class Stats {}
+@Immutable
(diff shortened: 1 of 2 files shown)


pongad:
cc @davidtorres

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/SubscriberStats.java
```diff
@@ -16,65 +16,33 @@
...
-  }
-
-  /** Number of successfully published messages. */
-  public long getAckedMessages() {
-    return totalAckedMessages;
-  }
+@AutoValue
+public abstract class SubscriberStats {
+  @AutoValue
+  public static abstract class Stats {}
```
garrettjonesgoogle: doesn't this mean that code referencing Stats objects won't compile?

pongad: I'm not sure if I'm understanding you correctly, but...

No code references this yet, because the only way to get to Stats is through SubscriberStats, and the only place that returns SubscriberStats is not implemented (it just returns null for now).

garrettjonesgoogle: Oh! That makes sense then.

pongad: Sorry, I should have been explicit in the commit msg.

garrettjonesgoogle:
LGTM after tests pass

PR #1481, opened 2016-12-16
make AutoValue work in PubSub

pongad:
This commit digs us out of the AutoValue dependency hell.

Previously
- We require Guava 20 for its Stats class.
- Guava source has a `@CanIgnoreReturnValue` annotation,
  this annotation is in errorprone package.
- Errorprone package required Java8 to run. So, I just took dependency
  on it without running to make the build pass in both 7 and 8.
- Errorprone shades class javax.lang.model.type.IntersectionType.
  This class is Java8-only, but because it wasn't run, it didn't cause a problem.
- AutoValue conditionally loads IntersectionType.
  It catches Exception and behaves properly if the class is not found;
  this is why it works on Java 7.
  However, because of the previous step, we have the class in our path.
  Because the class is 8-only, JVM7 loading it results in
  UnsupportedVersionError. This is an Error and not an Exception,
  and so AutoValue (probably wisely) ignore it and crashes.

The fix is simple.
Since #1480
already removed Stats, we can just go back to Guava 19.

@garrettjonesgoogle is full of wisdom!

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PublisherStats.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PublisherStats.java
@@ -22,7 +22,6 @@ import com.google.auto.value.AutoValue;
  * A snapshot of the publisher statistics at the time they were requested from the {@link
  * Publisher}.
  */
-//TODO: Finish implementation.
 @AutoValue
 public abstract class PublisherStats {
   /** Number of successfully published messages. */
@@ -36,4 +35,21 @@ public abstract class PublisherStats {
 
   /** Total messages sent, equal to pending + acked + failed messages. */
   public abstract long getSentMessages();
+
+  public static Builder newBuilder() {
+    return new AutoValue_PublisherStats.Builder();
+  }
+
+  @AutoValue.Builder
+  public static abstract class Builder {
+    public abstract Builder setAckedMessages(long value);
+
+    public abstract Builder setFailedMessages(long value);
+
+    public abstract Builder setPendingMessages(long value);
+
+    public abstract Builder setSentMessages(long value);
+
+    public abstract PublisherStats build();
+  }
 }
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/SubscriberStats.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/SubscriberStats.java
@@ -22,10 +22,11 @@ import com.google.auto.value.AutoValue;
  * A snapshot of the subscriber statistics at the time they were requested from the {@link
  * Subscriber}.
  */
-//TODO: Finish implementation.
 @AutoValue
 public abstract class SubscriberStats {
+  
   @AutoValue
+  //TODO: Finish implementation.
   public abstract static class Stats {}
 
   /** Number of successfully published messages. */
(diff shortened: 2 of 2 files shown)


eamonnmcmanus:
I don't really understand the explanation. If ErrorProne shades IntersectionType, nothing else should see it, even if explicitly loading javax.lang.model.type.IntersectionType, right? Also, I don't see where it is that AutoValue loads that class.

pongad:
@eamonnmcmanus I'm really not a Java expert but here's what I pieced together. I ran
```shell
mvn dependency:build-classpath | tr : '\n' | grep error | xargs -l1 zipinfo | grep -e ^Ar -e IntersectionType
```
The command shows that `IntersectionType.class` is in `$HOME/.m2/repository/com/google/errorprone/javac/1.9.0-dev-r2973-2/javac-1.9.0-dev-r2973-2.jar`.
Maybe I was wrong to use the word "shade".

AutoValue seems to be using `IntersectionType` [here](https://github.com/google/auto/blob/da49ab599269c26cd3bd3e2f3c364142ae08893c/common/src/main/java/com/google/auto/common/MoreTypes.java#L280)

eamonnmcmanus:
To step back a bit, the reason I care is that it does not seem right to go back to an earlier version of Guava. You should be able to use the latest version always.

The code in com.google.auto.common that you reference *could* catch Throwable rather than Exception. We'd have to make new releases of auto.common and auto.value for that to apply.

I'm very puzzled at the dependency from Error Prone that you show. It really shouldn't be polluting the class path with what appear to be classes from the Java 9 javac. Perhaps @eaftan can clarify?

It's also not clear to me from the description why Error Prone needs to be in the picture at all. If it requires Java 8 to run, and you're supposed to be compatible with Java 7, then it sounds as if you are making life difficult for yourselves. Perhaps you could have a Java 8 configuration with Error Prone and a Java 7 configuration without?

pongad:
@eamonnmcmanus I agree that it's not an ideal approach. Thank you for working with me on this.

Including errorprone, in hindsight, seems to be a mistake. However, it seems to be necessary to use guava20. Without it `mvn compile` fails because `CanIgnoreReturnValue` was missing from the classpath.

Grepping the source code, the annotation is used in 184 files in guava20, while it is not used at all in guava19.

eaftan:
Hmm, we were aware that Error Prone is polluting the classpath for our non-Maven integrations: https://github.com/google/error-prone/issues/492

But we thought our Maven integration was working correctly.  Also, you shouldn't need Error Prone just to use Guava 20.  You should be able to depend on our error-prone-annotations artifact to get the `@CanIgnoreReturnValue` annotation you need for Guava 20.  I'll put a comment on the correct line in the PR.

google-cloud-pubsub/pom.xml
```diff
@@ -54,16 +54,6 @@
       <version>${grpc.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>20.0</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.errorprone</groupId>
-      <artifactId>error_prone_core</artifactId>
```
eaftan: here you should have `error_prone_annotations`, not `error_prone_core`.

pongad:
@eaftan Thank you for the suggestion, this seems to make everything work now.

@eamonnmcmanus @garrettjonesgoogle PTAL

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PublisherStats.java
```diff
@@ -16,44 +16,24 @@
...
-
-  PublisherStats(long sentMessages, long ackedMessages, long failedMessages, long pendingMessages) {
-    this.sentMessages = sentMessages;
-    this.ackedMessages = ackedMessages;
-    this.failedMessages = failedMessages;
-    this.pendingMessages = pendingMessages;
-  }
-
+@AutoValue
+public abstract class PublisherStats {
```
pongad: @eamonnmcmanus I thought of just getting it to compile then work on the rest later. I guess the work is small enough that it makes sense to roll in.

pongad:
PTAL. I have left SubscriberStats.Stats empty for now, since I'm still not sure what kind of stats we want.

PR #1482, opened 2016-12-16
Use resource_names field instead of deprecated project_ids field

andreweames:


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -508,7 +508,7 @@ public void write(Iterable<LogEntry> logEntries, WriteOption... options) {
   private static ListLogEntriesRequest listLogEntriesRequest(LoggingOptions serviceOptions,
       Map<Option.OptionType, ?> options) {
     ListLogEntriesRequest.Builder builder = ListLogEntriesRequest.newBuilder();
-    builder.addProjectIds(serviceOptions.getProjectId());
+    builder.addResourceNames("projects/" + serviceOptions.getProjectId());
```
garrettjonesgoogle: @michaelbausor could you give guidance on using the right resource name type here?

garrettjonesgoogle: @andreweames , This should be: `ProjectName.create(serviceOptions.getProjectId()).toString()`

garrettjonesgoogle:
@michaelbausor could you write a new PR with this change but that uses the resource name classes? (Closing out this one due to inactivity)

PR #1487, opened 2016-12-21
delete pull-related methods from PubSub

pongad:
Instead, provide a way to create the Subscriber object.

cc @davidtorres


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
```diff
@@ -512,110 +503,17 @@ public SubscriptionId apply(String compositeSubscription) {
...
-    get(nackAsync(subscription, ackIds));
-  }
-
-  @Override
-  public Future<Void> nackAsync(String subscription, Iterable<String> ackIds) {
-    return modifyAckDeadlineAsync(subscription, 0, TimeUnit.SECONDS, ackIds);
-  }
-
-  @Override
   public void modifyAckDeadline(String subscription, int deadline, TimeUnit unit, String ackId,
```
davidtorres: This method is also not needed with the Subscriber class

pongad:
@davidtorres PTAL

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PubSubImpl.java
```diff
@@ -512,110 +503,17 @@ public SubscriptionId apply(String compositeSubscription) {
...
-
-  @Override
-  public MessageConsumer pullAsync(String subscription, MessageProcessor callback,
-      PullOption... options) {
-    Map<Option.OptionType, ?> optionMap = optionMap(options);
-    return MessageConsumerImpl.builder(getOptions(), subscription, ackDeadlineRenewer, callback)
-        .maxQueuedCallbacks(MAX_QUEUED_CALLBACKS.getInteger(optionMap))
-        .executorFactory(EXECUTOR_FACTORY.getExecutorFactory(optionMap))
+  public Subscriber subscriber(SubscriptionInfo subscription, Subscriber.MessageReceiver receiver) {
+    // TODO(pongad): Provide a way to pass in the rest of the options.
```
garrettjonesgoogle: @pongad , we probably want the input of @davidtorres on this matter: how many executors and channels to we want, in relation to the number of objects in the client library?

garrettjonesgoogle:
A couple high-level things:

* We need to write down in our TO DO list to write new samples that replace the deleted ones here (in spirit)
* Does our new code have the same level of test coverage as the code that's being deleted?
* We might want to write down some migration details for anyone using the library currently. Could you draft up a list of equivalents to each method on the main surface that you're deleting?

pongad:
@garrettjonesgoogle 
- I made an issue for the TODO list: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1493
- According to https://coveralls.io/builds/9371615, we have 80%+ on most classes. `PollingSubscriberConnection` is an exception at 76%. The missed lines are for expoenential backoff. They can be tested relatively easily I think.
- The migration document is at `google-cloud-pubsub/HP-MIGRATION.md`, added to the last commit. There isn't that much to document, since the new surface is pretty small.

garrettjonesgoogle:
Add a TODO to settle the issue of executors & channels. Otherwise LGTM.

PR #1496, opened 2016-12-22
create FlowControlSettings

pongad:
FlowController uses it instead of accepting Optionals.

cc @davidtorres

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
@@ -16,11 +16,14 @@
 
 package com.google.cloud.pubsub;
 
+import com.google.auto.value.AutoValue;
 import com.google.cloud.AsyncPage;
 import com.google.cloud.GrpcServiceOptions.ExecutorFactory;
 import com.google.cloud.Page;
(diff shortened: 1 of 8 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PubSub.java
```diff
@@ -137,6 +140,49 @@ public static PullOption executorFactory(ExecutorFactory executorFactory) {
     }
   }
 
+  @AutoValue
+  public abstract static class FlowControlSettings {
```
garrettjonesgoogle: Why can't this be a GAX class? It could be made generic like BundlingSettings by renaming 'messages' to 'elements'. Then `maxOutstandingBytes` corresponds to `requestByteThreshold` and `maxOutstandingElements` corresponds to `elementCountThreshold`. In fact, a better matchup might be `requestByteMaxOutstanding` and `elementCountMaxOutstanding`.

pongad:
@garrettjonesgoogle I originally thought of doing it in two steps, the second moves things to GAX. Now that I think of it, that probably just adds more overhead with no real gain.

This should be superseded by https://github.com/googleapis/gax-java/pull/163 . Once that is merged, I'll close out this PR.

PR #1501, opened 2017-01-05
Regenerating SPI layer

garrettjonesgoogle:
* Converting Error Reporting and Monitoring to use resource name types
* Removing formatX/parseX methods from pubsub, converting usage of the
  same to resource name types
* New methods in Logging and PubSub


michaelbausor:
LGTM

PR #1504, opened 2017-01-05
Updating grpc dependency to 1.0.3

garrettjonesgoogle:
This is necessary to provide configuration of proxies using the GRPC_PROXY_EXP environment variable.


michaelbausor:
LGTM

PR #1505, opened 2017-01-05
make pubsub work with new gax

pongad:


garrettjonesgoogle:
LGTM after the build succeeds

PR #1508, opened 2017-01-06
delete handwritten PushConfig

pongad:


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PushConfig.java
```diff
@@ -1,363 +0,0 @@
...
- *
- * <p>In a push subscription, the Pub/Sub server sends a request to the subscriber application. A
- * {@code PushConfig} object can be used to configure the application endpoint. The subscriber's
- * HTTP response serves as an implicit acknowledgement: a success response indicates that the
- * message has been succesfully processed and the Pub/Sub system can delete it from the
- * subscription; a non-success response indicates that the Pub/Sub server should resend it
- * (implicit "nack").
- *
- * @see <a href="https://cloud.google.com/pubsub/subscriber">Subscriber Guide</a>
- */
```
garrettjonesgoogle: In the deleted comments here, are we losing any useful documentation?

pongad: I don't think we are. The link is in a bunch of other places. We can also add it to doc of MessagesProcessor if we want to.

garrettjonesgoogle: I'm not so much referring to this specific link as to all of the doc comments in the file. When we wrote this client, we added documentation that would help people when using the client. I would like to still make the information available somehow, even though we are losing these wrapper classes. That either means putting it into higher-level documentation, or modifying the original proto documentation.

PR #1515, opened 2017-01-09
add Publisher.Settings

pongad:
This class is used to configure Publisher.
User-visible options are made public.
Options that should be automatically populated,
like user credentials,
are left package-private for now.

Tests for input validation for BundlingSettings
and FlowController are removed.
They should be moved to gax instead.
https://github.com/googleapis/gax-java/issues/168

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PublisherImpl.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PublisherImpl.java
@@ -17,7 +17,6 @@
 package com.google.cloud.pubsub;
 
 import com.google.api.gax.bundling.FlowController;
-import com.google.auth.oauth2.GoogleCredentials;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.common.primitives.Ints;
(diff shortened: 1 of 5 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PublisherImpl.java
```diff
@@ -90,55 +83,41 @@
   private final Duration sendBundleDeadline;
   private ScheduledFuture<?> currentAlarmFuture;
 
-  PublisherImpl(Builder builder) throws IOException {
-    topic = builder.topic;
-
-    maxBundleMessages = builder.bundlingSettings.getElementCountThreshold();
-    maxBundleBytes = builder.bundlingSettings.getRequestByteThreshold();
-    maxBundleDuration = builder.bundlingSettings.getDelayThreshold();
+  PublisherImpl(String topic, PubSubOptions psOptions, Settings settings) throws IOException {
```
pongad: Option splitting logic:
- `topic` => we definitely need to ask for it, otherwise we don't know where to publish to
- `psOptions` => we want to infer these most of the time, so that the user can't push/pull to different endpoints by mistake
- `settings` => in simple cases, just use the default, might need to tweak for performance reasons

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/PublisherImpl.java
```diff
@@ -90,55 +84,42 @@
...
```
garrettjonesgoogle: Don't abbreviate the "ps" here - I would prefer spelled out, or maybe just `options`.

Bigger picture: It is very confusing to have both options and settings.

garrettjonesgoogle:
Is this obsolete or does it need to be rewritten?

pongad:
@garrettjonesgoogle Thanks for pointing this out. This is superseded by us deleting all the hand written parts.

PR #1516, opened 2017-01-10
Update version to 0.8.2-alpha-SNAPSHOT

garrettjonesgoogle:


google-cloud-bigquery/pom.xml
```diff
@@ -12,7 +12,7 @@
   <parent>
     <groupId>com.google.cloud</groupId>
     <artifactId>google-cloud-pom</artifactId>
-    <version>0.8.1-alpha</version>
+    <version>0.8.2-alpha-SNAPSHOT</version>
```
lesv: This looks so wrong.  SNAPSHOT is by it's very nature not released.  I kind of doubt you'll get the right result with this, as it's not properly defined in the format from the manual.

I would suggest the following `<version>0.8.2-SNAPSHOT</version><!-- alpha -->`

@tswast Thoughts?

tswast: Since we aren't going to have a `0.8.2` version, only a `0.8.2-alpha` version, `0.8.2-alpha-snapshot` seems the most correct.

This tagging of the release status in the version number is the root problem. It really is too bad we can't use a development status classifier like Python does instead of mucking with version numbers. https://packaging.python.org/distributing/#classifiers

lesv: Yes - but I don't think Maven Central will recognize -alpha-SNAPSHOT correctly as it doesn't match the standard.

garrettjonesgoogle: According to the Maven Complete Reference:

> If a version contains the string “-SNAPSHOT,” then Maven will expand this token to a date and time value converted to UTC (Coordinated Universal Time) when you install or release this component. For example, if your project has a version of “1.0-SNAPSHOT” and you deploy this project’s artifacts to a Maven repository, Maven would expand this version to “1.0-20080207-230803-1” if you were to deploy a release at 11:08 PM on February 7th, 2008 UTC. In other words, when you deploy a snapshot, you are not making a release of a software component; you are releasing a snapshot of a component at a specific time.

Notice that the expansion contains dashes in itself. To me this means that it won't cause a problem; the logic to determine if something is a snapshot looks for the "-SNAPSHOT" string, which won't be messed up by having extra dashes in the complete version.

PR #1520, opened 2017-01-11
Delete handwritten PubSub layer

pongad:
You have no idea how happy this PR makes me


google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/PubSubExample.java
```diff
@@ -1,665 +0,0 @@
...
-      if (param != null && !param.isEmpty()) {
-        actionAndParams.append(' ').append(param.replace("\n", "\n\t\t"));
-      }
-    }
-    System.out.printf("Usage: %s [<project_id>] operation [entity] <args>*%s%n",
-        PubSubExample.class.getSimpleName(), actionAndParams);
-  }
-
-  @SuppressWarnings("unchecked")
-  public static void main(String... args) throws Exception {
```
garrettjonesgoogle: We might want to consider keeping this file; it shows how to write a command-line app which calls PubSub. The command line options could all be preserved; the calls to the pubsub client would need to be converted. This could also be a good place to point people to who need to migrate existing code. 

You can address this in a separate PR.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/testing/LocalPubSubHelperTest.java
```diff
@@ -50,25 +46,10 @@ public void testCreate() {
...
-    LocalPubSubHelper helper = LocalPubSubHelper.create();
-    PubSubOptions options = helper.getOptions();
-    assertTrue(options.getProjectId().startsWith(PROJECT_ID_PREFIX));
-    assertTrue(options.getHost().startsWith("localhost:"));
-    assertSame(NoCredentials.getInstance(), options.getCredentials());
-  }
-
-  @Test
   public void testStartStopReset() throws IOException, InterruptedException, TimeoutException {
     LocalPubSubHelper helper = LocalPubSubHelper.create();
```
garrettjonesgoogle: Do we need LocalPubSubHelper still?

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/PubSubExample.java
pongad: Ack. With the change we do, I'll essentially have to rewrite the file anyway.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/testing/LocalPubSubHelperTest.java
pongad: I think the class can be deleted for reasons we already talked about. On the other hand, according to the doc, it's not only there to help test the client lib but also for any user who want to test their app against.

Up to you? I can be swayed both ways.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/testing/LocalPubSubHelperTest.java
```diff
@@ -50,25 +46,10 @@ public void testCreate() {
...
     helper.start();
-    PubSub pubsub = helper.getOptions().getService();
-    pubsub.create(TopicInfo.of(TOPIC));
-    assertNotNull(pubsub.getTopic(TOPIC));
```
garrettjonesgoogle: I think this should still call the equivalent methods in the GAPIC layer.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/testing/LocalPubSubHelperTest.java
garrettjonesgoogle: I guess let's keep it for now - it's not conceptually related to the existence of the handwritten layer.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/testing/LocalPubSubHelperTest.java
pongad: I think this buys us very little. This method is essentially testing two things:
- start, reset, stop
- getting service from the helper

We're still doing the first, but now that PubSub and PubSubOptions are gone, it's tough to do the second. Maybe we can provide a way to get PublisherClient and SubscriberClient from the helper? If we want to do it, I think that should be in a separate PR. Note that it still won't be usefully testing the clients. The clients are already tested with the generated unit and smoke tests. The test is only useful in verifying that the clients actually communicate with the created helper, as opposed to the real service.

garrettjonesgoogle: hmm ok, we can think about this separately.

garrettjonesgoogle:
LGTM

PR #1533, opened 2017-01-13
Set timestamp from LogRecord

gregw:
Set the timestamp for stackdriver logging from the  timestamp from the LogRecord


gregw:
@garrettjonesgoogle oops - pushed now

Issue #1532, opened 2017-01-13
Deprecate AsyncLoggingHandler?


gregw:
The fix for #1386 has made the normal LoggingHandler asynchronous!  Was this deliberate? If so, should the AsyncLoggingHandler be deprecated as it now no longer changes behaviour.

garrettjonesgoogle:
I think this needs to be cleaned up before GA.

garrettjonesgoogle:
We should delete this altogether, and support a setting for sync v async in the logging config.

michaelbausor:
Closed in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1716

gregw:
Could the AsyncLoggingHandler be left as a deprecated extension of LoggingHandler, as that will make transitioning usage easier?

PR #1535, opened 2017-01-13
Initialize the default MonitoredResource from a GAE environment

gregw:
This PR is based on work done for https://github.com/GoogleCloudPlatform/jetty-runtime/pull/81.
The creation of a MonitoredResource based on the GAE environment variables is probably general purpose enough for the conde to be included here.

### DIFF
--- a/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
+++ b/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
@@ -54,6 +54,7 @@ public class LoggingHandlerTest {
       .addLabel("levelName", "FINEST")
       .addLabel("levelValue", String.valueOf(Level.FINEST.intValue()))
       .addLabel("enhanced", "true")
+      .setTimestamp(123456789L)
       .build();
   private static final LogEntry FINER_ENTRY = LogEntry.newBuilder(StringPayload.of(MESSAGE))
       .setSeverity(Severity.DEBUG)
@@ -260,7 +261,7 @@ public class LoggingHandlerTest {
         new LoggingHandler(LOG_NAME, options, resource, Collections.singletonList(enhancer));
     handler.setLevel(Level.ALL);
     handler.setFormatter(new TestFormatter());
-    handler.publish(new LogRecord(Level.FINEST, MESSAGE));
+    handler.publish(newLogRecord(Level.FINEST, MESSAGE));
   }
   
   @Test


gregw:
@meltsufin FYI

meltsufin:
Thanks @gregw!

@garrettjonesgoogle Can you please take a look at this? We're trying to contribute the code we've developed for the `TracingLogHandler` back into the library, and this is just a part of it.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -310,6 +327,10 @@ private LogEntry entryFor(LogRecord record) {
         .addLabel("levelName", level.getName())
         .addLabel("levelValue", String.valueOf(level.intValue()))
         .setSeverity(severityFor(level));
+    if (gaeInstanceId != null) {
+      builder.addLabel("appengine.googleapis.com/instance_name", gaeInstanceId);
+    }
```
garrettjonesgoogle: I think I want to make the log entry enhancement more "pluggable" and do this work from a strategy object. This interacts with a previous PR of yours - https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1385 - which introduced `enhanceLogEntry`. The problem with that method is that if someone overrides that method, then they can't used AsyncLoggingHandler. I think I might do the conversion to a strategy pattern myself and send a PR for it.

gregw:
@garrettjonesgoogle 

in the case of gaeInstance, if the core of this PR is accepted (creating a GAE MonitoredResource), then it would be natural for the LoggingHandler to also set the gaeInstance. Also note #1532 that currently there is no need for AsyncLoggingHandler, as the base handler is now Async.

Either way I think a pluggable strategy would be good and better than `enhanceLogEntry` and I'd be happy to have a go at a PR for it.

I'm thinking:
```
interface LoggingEnhancer {
  MonitoredResource getDefaultResource();
  void enhanceLogEntry(LogEntry.Builder builder, LogRecord record);
}
```

Then we can add a GaeFlexLoggingEnhancer to the build so that a logging.properties file could look like:
```
handlers=com.google.cloud.logging.LoggingHandler
com.google.cloud.logging.LoggingHandler.enhancer=com.google.cloud.logging.GaeFlexLoggingEnhancer
```

garrettjonesgoogle:
You are correct that the base handler is Async, but I'm not sure if I want to keep it that way. My point is a larger one though - as soon as someone creates one subclass to customize some behavior, it prevents independent behavior customization through other subclasses.

I think the scope of enhancing MonitoredResource should actually be the same as enhancing LogEntry. So I would suggest this:

```
interface LoggingEnhancer {
  void enhanceMonitoredResource(MonitoredResource.Builder builder);
  void enhanceLogEntry(LogEntry.Builder builder, LogRecord record);
}
```

The current `getDefaultResource()` implementation should create a builder and add the project_id label to it, then call enhanceMonitoredResource to override/supplement values.

I would also suggest a list of enhancers, in case there are independent labels to add.

It would be great for you to pick up this enhancement if you are able.

gregw:
@garrettjonesgoogle main problem with the approach that you propose is that the type of the monitored resource type cannot not be enhanced, nor  does it lend itself to multiple enhancers (even if it were settable).

Perhaps the resource type can be another parameter passed either to the constructor or set in the property file:

```
handlers=com.google.cloud.logging.LoggingHandler
com.google.cloud.logging.LoggingHandler.resourceType=gae_app
com.google.cloud.logging.LoggingHandler.enhancers=com.google.cloud.logging.GaeFlexLoggingEnhancer,org.example.SomeOtherEnhancer
```

I'll prepare something along those lines first thing next week.

gregw:
@garrettjonesgoogle I have updated to have a Enhancer abstraction.  I've added a GaeFlexLoggingEnchancer implementation, which includes the code to be able to add a traceId to the logs.

@meltsufin The GAE log can now be setup with no code, although it is a little verbose in the logging.properties file:
```
handlers=com.google.cloud.logging.LoggingHandler
com.google.cloud.logging.LoggingHandler.log=gaeflex.log
com.google.cloud.logging.LoggingHandler.resourceType=gae_app
com.google.cloud.logging.LoggingHandler.enhancers=com.google.cloud.logging.GaeFlexLoggingEnhancer
com.google.cloud.logging.LoggingHandler.formatter = java.util.logging.SimpleFormatter
java.util.logging.SimpleFormatter.format=%3$s: %5$s%6$s
```

google-cloud-logging/src/main/java/com/google/cloud/logging/GaeFlexLoggingEnhancer.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package com.google.cloud.logging;
+
+import java.util.logging.LogRecord;
+
+import com.google.cloud.MonitoredResource.Builder;
+
+/**
+ * A Logging {@link Enhancer} that enhances the logging for
```
garrettjonesgoogle: "for" -> "for the"

google-cloud-logging/src/main/java/com/google/cloud/logging/GaeFlexLoggingEnhancer.java
```diff
@@ -0,0 +1,84 @@
...
+ * GAE Flex environment.    This enhance can be configured in
```
garrettjonesgoogle: "This enhance" -> "This enhancer"

google-cloud-logging/src/main/java/com/google/cloud/logging/GaeFlexLoggingEnhancer.java
```diff
@@ -0,0 +1,84 @@
...
+   * @return id The traceID
+   */
+  public static String getCurrentTraceId() {
+    return traceId.get();
+  }
+  
+  @Override
+  public void enhanceMonitoredResource(Builder builder) {
+    gaeInstanceId = System.getenv("GAE_INSTANCE");    // Are we running on a GAE instance?
+    if (gaeInstanceId!=null) {
```
garrettjonesgoogle: put spaces around `!=` (multiple places in this file)

google-cloud-logging/src/main/java/com/google/cloud/logging/GaeFlexLoggingEnhancer.java
```diff
@@ -0,0 +1,84 @@
...
+   * the current thread.
+   * @return id The traceID
+   */
+  public static String getCurrentTraceId() {
+    return traceId.get();
+  }
+  
+  @Override
+  public void enhanceMonitoredResource(Builder builder) {
+    gaeInstanceId = System.getenv("GAE_INSTANCE");    // Are we running on a GAE instance?
```
garrettjonesgoogle: I think it would be better to set this in the constructor instead of here, where we can't really predict how it will be called (e.g. from multiple threads)

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -131,9 +134,25 @@ public LoggingHandler(String log, LoggingOptions options) {
    *
    * @param log the name of the log to which log entries are written
    * @param options options for the Stackdriver Logging service
-   * @param monitoredResource the monitored resource to which log entries refer
+   * @param monitoredResource the monitored resource to which log entries refer. If null a default
```
garrettjonesgoogle: "If null a default" -> "If it is null, a default" (same in following doc also)

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -131,9 +134,25 @@ public LoggingHandler(String log, LoggingOptions options) {
...
+  
+  /**
+   * Creates a handler that publishes messages to Stackdriver Logging.
+   *
+   * @param log the name of the log to which log entries are written
+   * @param options options for the Stackdriver Logging service
+   * @param monitoredResource the monitored resource to which log entries refer. If null a default
+   * @param enhancers List of {@link Enhancer} instances.
+   * resource is created based on the project ID.  If a Google App Engine environment is detected
+   * then a more comprehensive default resource may be created.
```
garrettjonesgoogle: You seem to have put the `enhancers` doc in the middle of the `monitoredResource` doc

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -429,4 +476,14 @@ public synchronized long setFlushSize(long flushSize) {
   public static void addHandler(Logger logger, LoggingHandler handler) {
     logger.addHandler(handler);
   }
+  
+  /**
+   * A Log Enhancer.
+   * May be used to enhanced the {@link MonitoredResource} and/or the {@link LogEntry}
```
garrettjonesgoogle: "to enhanced" -> "to enhance"

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -131,9 +134,25 @@ public LoggingHandler(String log, LoggingOptions options) {
    *
    * @param log the name of the log to which log entries are written
    * @param options options for the Stackdriver Logging service
-   * @param monitoredResource the monitored resource to which log entries refer
+   * @param monitoredResource the monitored resource to which log entries refer. If it is null 
+   * then a default resource is created based on the project ID.  If a Google App Engine environment is detected
+   * then a more comprehensive default resource may be created.
```
garrettjonesgoogle: This doc is out of date now with the change you make to pull out `GaeFlexLoggingEnhancer`. (Next method too)

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -310,12 +354,15 @@ private LogEntry entryFor(LogRecord record) {
         .addLabel("levelName", level.getName())
         .addLabel("levelValue", String.valueOf(level.intValue()))
         .setSeverity(severityFor(level));
+
     enhanceLogEntry(builder, record);
     return builder.build();
   }
   
   protected void enhanceLogEntry(LogEntry.Builder builder, LogRecord record) {
```
garrettjonesgoogle: Make this private now, or better yet, just move the implementation to where it's called, because it doesn't need to be overridden anymore.

gregw: I kept it there because it is a released API.  But it only existed for a very short period of time...
How about I deprecate it and move the behaviour out of it.

PR #1538, opened 2017-01-16
make Publisher uses RetrySettings

pongad:
cc @davidtorres

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
@@ -146,28 +146,30 @@ public interface Publisher {
 
   /** A builder of {@link Publisher}s. */
   public final class Builder {
-    // Meaningful defaults.
-    static final long DEFAULT_MAX_BUNDLE_MESSAGES = 100L;
-    static final long DEFAULT_MAX_BUNDLE_BYTES = 1000L; // 1 kB
-    static final Duration DEFAULT_MAX_BUNDLE_DURATION = new Duration(1); // 1ms
-    static final Duration DEFAULT_REQUEST_TIMEOUT = new Duration(10 * 1000); // 10 seconds
     static final Duration MIN_TOTAL_TIMEOUT = new Duration(10 * 1000); // 10 seconds
     static final Duration MIN_RPC_TIMEOUT = new Duration(10); // 10 milliseconds
+
+    // Meaningful defaults.
+    static final long DEFAULT_ELEMENT_COUNT_THRESHOLD = 100L;
+    static final long DEFAULT_REQUEST_BYTES_THRESHOLD = 1000L; // 1 kB
+    static final Duration DEFAULT_DELAY_THRESHOLD = new Duration(1); // 1ms
+    static final Duration DEFAULT_RPC_TIMEOUT = new Duration(10 * 1000); // 10 seconds
+    static final Duration DEFAULT_TOTAL_TIMEOUT = MIN_TOTAL_TIMEOUT;
     static final BundlingSettings DEFAULT_BUNDLING_SETTINGS =
         BundlingSettings.newBuilder()
-            .setDelayThreshold(DEFAULT_MAX_BUNDLE_DURATION)
-            .setRequestByteThreshold(DEFAULT_MAX_BUNDLE_BYTES)
-            .setElementCountThreshold(DEFAULT_MAX_BUNDLE_MESSAGES)
+            .setDelayThreshold(DEFAULT_DELAY_THRESHOLD)
+            .setRequestByteThreshold(DEFAULT_REQUEST_BYTES_THRESHOLD)
+            .setElementCountThreshold(DEFAULT_ELEMENT_COUNT_THRESHOLD)
             .build();
     static final RetrySettings DEFAULT_RETRY_SETTINGS =
         RetrySettings.newBuilder()
-            .setTotalTimeout(MIN_TOTAL_TIMEOUT)
+            .setTotalTimeout(DEFAULT_TOTAL_TIMEOUT)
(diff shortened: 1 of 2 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -150,14 +151,24 @@
...
             .build();
+    static final RetrySettings DEFAULT_RETRY_SETTINGS =
+        RetrySettings.newBuilder()
+            .setTotalTimeout(MIN_TOTAL_TIMEOUT)
+            .setInitialRetryDelay(Duration.millis(5))
+            .setRetryDelayMultiplier(2)
+            .setMaxRetryDelay(Duration.millis(Long.MAX_VALUE))
+            .setInitialRpcTimeout(DEFAULT_REQUEST_TIMEOUT)
+            .setRpcTimeoutMultiplier(2)
+            .setMaxRpcTimeout(DEFAULT_REQUEST_TIMEOUT)
```
pongad: This is left the same with initial RPC timeout so preserve the old behavior that the value stays the same. I'm not too worried about it since it's easy to change later.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -150,14 +151,24 @@
...
+    static final Duration MIN_RPC_TIMEOUT = new Duration(10); // 10 milliseconds
     static final BundlingSettings DEFAULT_BUNDLING_SETTINGS =
         BundlingSettings.newBuilder()
             .setDelayThreshold(DEFAULT_MAX_BUNDLE_DURATION)
             .setRequestByteThreshold(DEFAULT_MAX_BUNDLE_BYTES)
             .setElementCountThreshold(DEFAULT_MAX_BUNDLE_MESSAGES)
             .build();
+    static final RetrySettings DEFAULT_RETRY_SETTINGS =
+        RetrySettings.newBuilder()
+            .setTotalTimeout(MIN_TOTAL_TIMEOUT)
```
garrettjonesgoogle: We should distinguish the concepts of minimum and default - make a new constant `DEFAULT_TOTAL_TIMEOUT` and use that here.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherImpl.java
```diff
@@ -150,17 +140,21 @@ public PublisherStats getStats() {
...
-    return maxBundleBytes;
+    return bundlingSettings.getRequestByteThreshold();
+  }
+
+  private boolean hasBundlingBytes() {
+    return getMaxBundleBytes() > 0;
   }
 
   @Override
   public long getMaxBundleMessages() {
```
garrettjonesgoogle: We should probably just have an accessor for bundling settings, because that is the type that is provided at build time.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -150,14 +151,24 @@
     static final long DEFAULT_MAX_BUNDLE_BYTES = 1000L; // 1 kB
     static final Duration DEFAULT_MAX_BUNDLE_DURATION = new Duration(1); // 1ms
     static final Duration DEFAULT_REQUEST_TIMEOUT = new Duration(10 * 1000); // 10 seconds
```
garrettjonesgoogle: This should be `DEFAULT_RPC_TIMEOUT` for consistency.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -150,14 +151,24 @@
     static final long DEFAULT_MAX_BUNDLE_BYTES = 1000L; // 1 kB
     static final Duration DEFAULT_MAX_BUNDLE_DURATION = new Duration(1); // 1ms
```
garrettjonesgoogle: Rename these other defaults to match the fields they are used to set (You can do this in a separate PR)

pongad: Done. MIN_{TOTAL,RPC}_TIMEOUT is kept unchanged, since they are actually the minimum we allow. The default just also happens to be the min.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherImpl.java
pongad: I'd like to defer this for when we make Publisher a concrete class. Then we'll have fewer places to change.

pongad:
@garrettjonesgoogle PTAL

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PublisherImpl.java
garrettjonesgoogle: sounds good.

garrettjonesgoogle:
LGTM - but put the failure here on your list to investigate: https://travis-ci.org/GoogleCloudPlatform/google-cloud-java/jobs/192903764 (it is a pubsub failure; it seems intermittent since 3 builds passed, this one failed)

pongad:
As discussed, we can merge to pubsub-hp, but we'll definitely hold off merging to master until we fix the flakes.

PR #1540, opened 2017-01-16
add back PubSub samples

pongad:


google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.pubsub.v1.PubsubMessage;
+import com.google.pubsub.v1.PushConfig;
+import com.google.pubsub.v1.SubscriptionName;
+import com.google.pubsub.v1.TopicName;
+
+/**
+ * A snippet for Google Cloud Pub/Sub showing how to create a Pub/Sub pull subscription and
+ * asynchronously pull messages from it.
+ */
+public class CreateSubscriptionAndPullMessages {
```
garrettjonesgoogle: This is extremely helpful for reviewing the overall API!

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public static void main(String... args) throws Exception {
+    TopicName topic = TopicName.create("test-project", "test-topic");
+    SubscriptionName subscription = SubscriptionName.create("test-project", "test-subscription");
+
+    try (SubscriberClient subscriberClient = SubscriberClient.create()) {
+      subscriberClient.createSubscription(subscription, topic, PushConfig.getDefaultInstance(), 0);
+    }
+
+    MessageReceiver receiver =
```
garrettjonesgoogle: Comparing to the existing snippet at https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java , it occurs to me that we could call this MessageProcessor and require fewer changes of existing code.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
```diff
@@ -0,0 +1,72 @@
...
+            public void failed(Service.State from, Throwable failure) {
+              System.err.println(failure);
+            }
+          },
+          MoreExecutors.directExecutor());
+      subscriber.startAsync().awaitRunning();
+      Thread.sleep(60000);
+    } finally {
+      if (subscriber != null) {
+        subscriber.stopAsync();
```
garrettjonesgoogle: Consider for a To Do: should we have subscriber implement AutoCloseable like the GAPIC clients do? That would make usage more consistent.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
```diff
@@ -0,0 +1,72 @@
...
+        };
+    Subscriber subscriber = null;
+    try {
+      subscriber = Subscriber.Builder.newBuilder(subscription, receiver).build();
+      subscriber.addListener(
+          new Service.Listener() {
+            @Override
+            public void failed(Service.State from, Throwable failure) {
+              System.err.println(failure);
+            }
```
garrettjonesgoogle: I can't tell what the purpose of this listener is. Why can't we support handling of this in `MessageReceiver`?

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
```diff
@@ -0,0 +1,72 @@
...
+          },
+          MoreExecutors.directExecutor());
+      subscriber.startAsync().awaitRunning();
```
garrettjonesgoogle: Why is starting asynchronous?

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
```diff
@@ -0,0 +1,72 @@
...
+      Thread.sleep(60000);
```
garrettjonesgoogle: Why do we have a sleep here if we are already called `awaitRunning` on the previous line?

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+/**
+ * A snippet for Google Cloud Pub/Sub showing how to create a Pub/Sub topic and asynchronously
+ * publish messages to it.
+ */
+public class CreateTopicAndPublishMessages {
+  public static void main(String... args) throws Exception {
+    TopicName topic = TopicName.create("test-project", "test-topic");
+    try (PublisherClient publisherClient = PublisherClient.create()) {
```
garrettjonesgoogle: We are losing sample code that gets the project id from the environment

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
```diff
@@ -0,0 +1,60 @@
...
+
+    Publisher publisher = null;
+    try {
+      publisher = Publisher.Builder.newBuilder(topic).build();
+      List<String> messages = Arrays.asList("first message", "second message");
+      List<ListenableFuture<String>> messageIds = new ArrayList<>();
+      for (String message : messages) {
+        messageIds.add(
+            publisher.publish(
+                PubsubMessage.newBuilder().setData(ByteString.copyFromUtf8(message)).build()));
```
garrettjonesgoogle: Reduce the embedding on this line

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
```diff
@@ -0,0 +1,60 @@
...
+      }
+      for (String messageId : Futures.allAsList(messageIds).get()) {
```
garrettjonesgoogle: The code you have here engages blocking, and is pretty close to purely synchronous code. With the existing bundling defaults, only the delay threshold would trigger a publish if we didn't have the blocking call count threshold functionality. Since the default threshold is 10 ms this isn't much of an issue here, but if someone bumped up that threshold, this code would block for that long. 

If we did retain blocking call count threshold, then blocking on the result (with the default value of 1) would be equivalent to a flush, and the pattern you have in this snippet would work well for whatever values the user set in the bundling settings (except for messing with the blocking call count threshold).

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -186,6 +187,10 @@ public static Builder newBuilder(String topic) {
       return new Builder(topic);
     }
 
+    public static Builder newBuilder(TopicName topic) {
```
garrettjonesgoogle: add javadoc

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -142,6 +143,10 @@ public static Builder newBuilder(String subscription, MessageReceiver receiver)
       return new Builder(subscription, receiver);
```
garrettjonesgoogle: Validation might be a good idea here

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -186,6 +187,10 @@ public static Builder newBuilder(String topic) {
       return new Builder(topic);
```
garrettjonesgoogle: Validation might be a good idea here

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
pongad: The name is already used by another class in David's implementation. We can just swap the two though...

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
pongad: They are for different errors. This listener is for cases where the error is so serious that the service cannot start properly. Maybe things like "kernel doesn't let us spawn more threads" and stuff like that.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
pongad: These primitives are provided by the [Service class](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/Service.html). We are essentially making this synchronous though since we `awaitRunning` directly after.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
pongad: We are sleeping here to let other threads (spawned by subscriber) process messages. If we don't sleep here, will wait for subscriber to start listening then immediately shut it down.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
pongad: As discussed, leaving this for now since stop may block

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
pongad: Done

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
pongad: Do you think we can put this somewhere central? Maybe com.google.cloud.CloudUtil? That way it would also benefit code that uses autogenerated code directly.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
pongad: As discussed offline, the threshold should in principle be small, so it should never really be an issue.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
pongad: I think the best idea is to delete the String overload altogether and just use TopicName. TopicName is then responsible for the validation.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
pongad: Only use SubscriptionName.

pongad:
@garrettjonesgoogle PTAL

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
garrettjonesgoogle: As discussed offline: we will keep `MessageReceiver` named as is. The existing `MessagesProcessor` will be renamed to `MessageDispatcher`.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
```diff
@@ -0,0 +1,60 @@
...
+
+    Publisher publisher = null;
+    try {
+      publisher = Publisher.Builder.newBuilder(topic).build();
+      List<String> messages = Arrays.asList("first message", "second message");
+      List<ListenableFuture<String>> messageIds = new ArrayList<>();
+      for (String message : messages) {
+        PubsubMessage pubsubMessage =
+            PubsubMessage.newBuilder().setData(ByteString.copyFromUtf8(message)).build();
+        messageIds.add(publisher.publish(pubsubMessage));
```
garrettjonesgoogle: Split this into two lines also.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
pongad: Tracking at https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1546

pongad:
@garrettjonesgoogle You mean splitting the ByteString off? PTAL

garrettjonesgoogle:
LGTM for merging purposes. Add a to do somewhere to make the samples simpler.

PR #1547, opened 2017-01-19
delete PublisherClientTest

pongad:
This is replaced by PublisherTest.
Eventually, we'll want to rename PublisherTest to PublisherClientTest,
so that PublisherImplTest and be renamed PublisherTest.

There will not be a quiz on renaming history.


garrettjonesgoogle:
LGTM

PR #1551, opened 2017-01-19
make sure channel round-robin cannot fail on overflow

pongad:
The AtomicInteger might overflow.
Previously, we'll just die since we'd try to index into an array
with negative number.
This commit makes sure we have a valid index to use.

Note that this is still not a "perfect" round robin.
If the number of channels does not divide 2^32,
we'll slightly prefer lower indices.
This is probably OK since the bias is small,
and the alternative requires a much more expensive lock.

cc @davidtorres


garrettjonesgoogle:
Hmm how about a test?

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #1552, opened 2017-01-19
fix improper use of iteration

pongad:
cc @davidtorres

The previous implementation modifies a HashMap while iterating through
its key set, causeing unpredictable iteration behavior.
This might be the reason our tests intermittently deadlock.

The new implementation uses a PriorityQueue.
The time complexity is O(M * log N),
where M is the number of expirations before the cut over time
and N is the total number of expirations.
I am not sure how this compares to O(N) intended in the previous
implementation.
If required, O(N) is also possible using an ArrayList.

Unfortunately, a new failure has emerged.
Instead of deadlocking, testModifyAckDeadline intermittently fails.
Maybe
- I have fixed the old bug and created a new one,
- I have fixed the old bug that was masking another one,
- The deadlock wasn't caused by the iteration. Now the tests just fail
  before they could deadlock,
or some combination thereof.
The incorrect iteration should be fixed regardless.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
@@ -114,7 +114,9 @@ class MessageDispatcher {
       for (AckHandler ah : ackHandlers) {
         ackIds.add(ah.ackId);
       }
-      return String.format("ExtensionJob {expiration: %s, nextExtensionSeconds: %d, ackIds: %s}", expiration, nextExtensionSeconds, ackIds);
+      return String.format(
+          "ExtensionJob {expiration: %s, nextExtensionSeconds: %d, ackIds: %s}",
+          expiration, nextExtensionSeconds, ackIds);
     }
   }
 
@@ -138,7 +140,9 @@ class MessageDispatcher {
     }
 
     public String toString() {
-      return String.format("extend %d sec: %s", deadlineExtensionSeconds, ackIds);
+      return String.format(
+          "PendingModifyAckDeadline{extension: %d sec, ackIds: %s}",
+          deadlineExtensionSeconds, ackIds);
     }
   }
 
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessagesProcessor.java
```diff
@@ -257,17 +248,16 @@ public void processReceivedMessages(List<com.google.pubsub.v1.ReceivedMessage> r
...
       ackHandlers.add(new AckHandler(pubsubMessage.getAckId(), messageSize));
     }
-    ExpirationInfo expiration =
-        new ExpirationInfo(
-            clock, now.plus(messageDeadlineSeconds * 1000), INITIAL_ACK_DEADLINE_EXTENSION_SECONDS);
+    Instant expiration = now.plus(messageDeadlineSeconds * 1000);
     synchronized (outstandingAckHandlers) {
-      addOutstadingAckHandlers(expiration, ackHandlers);
+      outstandingAckHandlers.add(
+          new ExtensionJob(expiration, INITIAL_ACK_DEADLINE_EXTENSION_SECONDS, ackHandlers));
```
pongad: It just occurred to me that this can cause a race. The AckDeadlineAlarm might access ackHandlers at the same time as this function iterates over it just below. For some reason, moving this synchronized-block just above the try-block below causes a deadlock 100% of the time. I don't fully understand this yet; I'll take another look in the morning.

garrettjonesgoogle:
I think @davidtorres should probably do the first pass, since he wrote the original code and understands it better.

pongad:
David's first pass is at https://github.com/GoogleCloudPlatform/google-cloud-java/commit/9dc57e8dc66bb44179de00ad7b76a2ac0f13ab48

I'm not sure why the review is not showing up here. @garrettjonesgoogle @davidtorres PTAL.

pongad:
David LGTMed in the linked PR. Still not sure why it's not here...

PR #1553, opened 2017-01-20
do not expose Service

pongad:
I'm not sure how ready this PR is. The general logic is to hide the implementation in a subclass then create a delegator. If the approach looks fine to you, I'll go ahead and document the methods.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -32,6 +32,8 @@ import com.google.common.util.concurrent.AbstractService;
 import com.google.common.util.concurrent.Service;
 import com.google.pubsub.v1.SubscriptionName;
 import io.grpc.ManagedChannelBuilder;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeoutException;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.grpc.netty.GrpcSslContexts;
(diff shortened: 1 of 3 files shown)


pongad:
@garrettjonesgoogle Please take a look. If possible, I'd like to merge this ASAP, since everything we do will conflict with this change.

garrettjonesgoogle:
It's hard to figure out what has changed and what has moved - can you point me to the points in the PR that need to actually be considered?

pongad:
@garrettjonesgoogle Essentially, all fields (non static ones) and all methods are moved into the inner `SubscriberImpl` class. `SubscriberImpl` still `extends AbstractService` so that we can the methods we want. The `Subscriber` class just re-declare methods in `SubscriberImpl`. It also re-declares `Service.State` and `Service.Listener` classes and translate between the original and the redeclaration.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -115,272 +117,402 @@
...
+    impl.stopAsync();
+    return this;
   }
 
-  @Override
-  protected void doStart() {
-    logger.debug("Starting subscriber group.");
-    startStreamingConnections();
-    notifyStarted();
+  public enum State {
```
pongad: Redeclaring `Service.State`.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -115,272 +117,402 @@
...
+    FAILED,
+    NEW,
+    RUNNING,
+    STARTING,
+    STOPPING,
+    TERMINATED
+  }
+
+  public abstract static class SubscriberListener {
```
pongad: Redeclare `Service.Listener`.

garrettjonesgoogle:
LGTM

PR #1555, opened 2017-01-22
hide Publisher and Subscriber's getStats()

pongad:
The methods are not yet implemented,
so there is no reason to expose it.
Another way out is to delete it for the time being.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -17,6 +17,8 @@
 package com.google.cloud.pubsub.spi.v1;
 
 import com.google.api.gax.bundling.FlowController;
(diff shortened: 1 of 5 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -356,7 +356,7 @@ public void run() {
     }
   }
 
-  public SubscriberStats getStats() {
+  private SubscriberStats getStats() {
```
garrettjonesgoogle: We are tracking separately - let's not have unnecessary empty getters. Keeping the classes (but package private as you have them) seems fine.

pongad: Done

garrettjonesgoogle:
LGTM

PR #1557, opened 2017-01-22
use ExecutorProvider

pongad:
Use ExecutorProvider instead of asking of an Executor directly.
This makes Publisher and Subscriber consistent with toolkit-generated
code.

This commit leaves Channels and Credentials alone for now,
since it is already getting big.


garrettjonesgoogle:
LGTM

PR #1558, opened 2017-01-23
fix race in TestReceiver

pongad:
cc @davidtorres

If TestReceiver is set so that messages must be explicitly acked,
test code acks a message by calling replyNextOutstandingMessage.
There is a race between calling the function and the message being
delivered.

Previously, if the function gets called before the message is delivered,
the test fails since polling an empty deque returns a null pointer.
This commit makes us wait until a message becomes available instead.

### DIFF
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java
@@ -19,6 +19,8 @@ package com.google.cloud.pubsub.spi.v1;
 import static com.google.cloud.pubsub.spi.v1.MessageDispatcher.PENDING_ACKS_SEND_DELAY;
 import static org.junit.Assert.assertEquals;
 
+import com.google.api.gax.grpc.FixedExecutorProvider;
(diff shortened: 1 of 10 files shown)


google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java
```diff
@@ -123,34 +122,43 @@ void waitForExpectedMessages() throws InterruptedException {
...
 
     public void replyAllOutstandingMessage() {
       Preconditions.checkState(explicitAckReplies);
+      for (; ; ) {
+        SettableFuture<AckReply> reply = outstandingMessageReplies.poll();
+        if (reply == null) {
+          return;
+        }
+        replyTo(reply);
+      }
```
davidtorres: nit: up to you but the following code looks more readable to me

SettableFuture<AckReply> reply;
while ((reply = test.poll()) != null) {
  replyTo(reply);
}

pongad: I somehow didn't know this worked in Java. Done.

PR #1560, opened 2017-01-24
fix race in mock publisher

pongad:
FakePublisherServiceImpl::publish has a race.
If the call to publish happen before a response can be placed,
the server will try to respond with a null object.

The fix is to either
- always set the response before calling
- make publish wait for the response
For propriety, this commit does both.

This fix reveals another flake.
Publisher uses exponential backoff with jitter.
The jitter randomly picks a number between 0 and a maximum.
If we pick low values too many times,
it will retry too often and the server will run out of canned
transient errors to respond back with.
The test still passed since it expected any Throwable.

This commit fixed the test to expect FakeException,
and use a fake "random" number generator to ensure
we don't run out of return values.

Retrying can still causes random test failures,
independently of above changes.
If a request fails due to DEADLINE_EXCEEDED,
the future is completed with a corresponding error.
However, the last RPC might not have been successfully cancelled.
When a new test starts, it gives canned response to the server.
The server might use some of these responses to respond to
RPCs of previous tests.
Consequently, a misbehaving test can fail every test that comes
after it.
This commit changes the test setup code so that it
creates a new fake server for every test to avoid this problem.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
@@ -23,6 +23,7 @@ import com.google.api.gax.grpc.ExecutorProvider;
 import com.google.api.gax.grpc.InstantiatingExecutorProvider;
 import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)


pongad:
cc @davidtorres , we still seem to have flakes in `SubscriberImplTest`. I'll take a look tomorrow.
@garrettjonesgoogle Do we want to make retry jitter in gax `(max/2, max)` instead of `(0, max)`? Please see the motivation in the paragraph below the bullets.

garrettjonesgoogle:
Couldn't you also mock the jitter, or the random number provider?

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM, after you update the PR description to match the latest logic.

pongad:
Description fixed in PR and commit

PR #1563, opened 2017-01-25
use ChannelProvider instead of ChannelBuilder

pongad:


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -159,37 +151,35 @@ private Publisher(Builder builder) throws IOException {
...
-              : NettyChannelBuilder.forAddress(
-                      PublisherSettings.getDefaultServiceAddress(),
-                      PublisherSettings.getDefaultServicePort())
-                  .negotiationType(NegotiationType.TLS)
-                  .sslContext(GrpcSslContexts.forClient().ciphers(null).build())
-                  .executor(executor)
-                  .build();
+          builder.channelProvider.needsExecutor()
+              ? builder.channelProvider.getChannel(executor)
+              : builder.channelProvider.getChannel();
```
garrettjonesgoogle: The ChannelProvider interface abstracts away whether a new channel should be created or a single channel should be reused; it occurs to me that re-using a channel violates the intent of this code. I wonder if we should only accept InstantiatingChannelProvider?

If not, maybe ProviderManager might be helpful here - it encapsulates a pair of ChannelProvider & ExecutorProvider.

garrettjonesgoogle: Hmm after offline discussion, it looks like only accepting `InstantiatingChannelProvider` might not make sense; it makes testing harder. We should just be clear in the documentation.

garrettjonesgoogle:
LGTM

PR #1572, opened 2017-01-27
use RpcFuture and remove old BundlingSettings

pongad:


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -107,9 +106,6 @@ public Void apply(WriteLogEntriesResponse input) {
 
   private static <I, O> Future<O> transform(Future<I> future,
       Function<? super I, ? extends O> function) {
-    if (future instanceof ListenableFuture) {
-      return Futures.transform((ListenableFuture<I>) future, function);
-    }
```
garrettjonesgoogle: Why does this logic become unnecessary?

google-cloud-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java
```diff
@@ -148,21 +145,25 @@ public DefaultLoggingRpc(LoggingOptions options) throws IOException {
...
-      public V apply(ApiException exception) {
-        if (returnNullOnSet.contains(exception.getStatusCode().value())) {
-          return null;
-        }
-        throw new LoggingException(exception, idempotent);
-      }
-    });
+    return from.catching(
+        ApiException.class,
+        new RpcFuture.Function<ApiException, V>() {
```
garrettjonesgoogle: Regarding the question in GAX about the location of `Function`: The notion of Function is not inherently tied to futures; what if we want to use them outside of futures?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
```diff
@@ -111,26 +108,54 @@ protected ChannelProvider getChannelProvider() {
   }
 
   private static final class PullFutureImpl
-      extends ForwardingListenableFuture.SimpleForwardingListenableFuture<PullResponse>
```
garrettjonesgoogle: Is it worth creating our own forwarding rpc future?

google-cloud-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java
pongad: Split into its own top-level interface.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java
pongad: Done. Unfortunately, it's still necessary to override the addCallback method.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
pongad: It has never been necessary. `Futures.transform` and `Futures.lazyTransform` do the same thing. The only difference is that `lazyTransform` moves the work to the thread running `get()` while `transform` does the work in the thread that completes the future.

We now return RpcFuture instead of ListenableFuture, so the path is never taken. Additionally, the function is so small it's unlikely this path would actually make a difference.

google-cloud-core/pom.xml
```diff
@@ -111,7 +111,7 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax</artifactId>
-      <version>0.0.25</version>
+      <version>0.0.28-SNAPSHOT</version>
```
pongad: Makes sense. I'll change this to `0.0.28` (without the `-SNAPSHOT`). If we don't update the version, the build still doesn't compile.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
mziccard: This was done to keep the dynamic type an instance of `ListenableFuture` if the input future was already a `ListenableFuture`.

- `Futures.transform` returns a `ListenableFuture` created via [`AbstractTransformFuture.create`](https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/AbstractTransformFuture.java#L37) 
- `Futures.lazyTrasform` returns an instance of `Future`

This was introduced in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1168 to allow users to use `JdkFutureAdapters.listenInPoolThread` without requiring additional threads. More context can be found in https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1148

garrettjonesgoogle: Since we are switching to using our own `RpcFuture` which has `addCallback` functionality, this workaround will no longer be necessary.

PR #1576, opened 2017-01-27
BigQuery: Add support to FormatOptions for AVRO

vam-google:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1441

Added new constant in FormatOptions and a corresponding factory method. Updated test cases. Confirmed that AVRO does not require special treatment (like CSV does), so no additional changes are required.

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/FormatOptionsTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/FormatOptionsTest.java
@@ -30,9 +30,12 @@ public class FormatOptionsTest {
     assertEquals(FormatOptions.JSON, options.getType());
     options = new FormatOptions(FormatOptions.DATASTORE_BACKUP);
     assertEquals(FormatOptions.DATASTORE_BACKUP, options.getType());
+    options = new FormatOptions(FormatOptions.AVRO);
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/FormatOptionsTest.java
```diff
@@ -30,23 +30,29 @@ public void testConstructor() {
     assertEquals(FormatOptions.JSON, options.getType());
     options = new FormatOptions(FormatOptions.DATASTORE_BACKUP);
     assertEquals(FormatOptions.DATASTORE_BACKUP, options.getType());
+    options = new FormatOptions(FormatOptions.AVRO);
+    assertEquals(FormatOptions.AVRO, options.getType());
   }
 
   @Test
+  @SuppressWarnings("deprecation")
```
garrettjonesgoogle: Can you change the `options.type()` calls to `options.getType()` here so that you don't need to suppress deprecation warnings?

vam-google: I believe those are used for an artificial test coverage. Leaving it as is, as we agreed.

garrettjonesgoogle:
LGTM - merge after the tests pass.

PR #1581, opened 2017-02-01
fix deadlock in testStreamAckDeadlineUpdate

### LINKED ISSUE #1577: deadlock in testStreamAckDeadlineUpdate
@garrettjonesgoogle @davidtorres @kir-titievsky

I tracked down another deadlock, but this one I cannot fix without changing the surface.

In the happy case, the test
- receives a message, and creates a future to respond
- advance time by 20s
- sets the future result to `ACK`
  - this updates the latency distribution
  - the next time we configure the stream latency, we'll set the latency to 20s
- advance time by 60s
  - we configure stream latency every minute
  - so, this tells the server to change our stream latency to 20s
- YAY!

The mechanism we use the update latency is that we add a callback to the `ListenableFuture` that is returned by `MessageReceiver::receiveMessage`. The logic is that once the user finishes with the work, the latency distribution is automatically updated.

In the bad case:
- the GRPC thread calls `MessageReciever::receive`
  - the test implementation creates a future, puts it in a queue to be retrieved by the test code.
- advance time by 20s
- the test thread takes the future above out the queue, and sets future result
- advance time by 60s
- **then**, the GRPC thread finally adds the callback (let's call this step *S1*)
  - since the future is already completed, we update the latency distribution to 80s
- the minute for configuring stream latency has also passed (let's call this step *S2*)
  - change stream latency to 80s

In an even worse case, *S1* and *S2* can switch order. Since there is no entries in the distribution, we don't update the server with anything. The latency on the server stays at 10s (the default).

Test code expects the latency value on the server [eventually be set to 20s](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/c46faf232dcde2e211f25cd5485d1de9216027f7/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java#L349). In either bad case, the test deadlocks.

I don't think this test can be fixed with the current surface. Since the Future is returned by `MessageReceiver`, it stands to reason that other threads can do arbitrary things to the future before the callback can be added. Note that this problem affects production code as well: the synchronization is still technically incorrect though the consequence is small enough that it probably will never matter.

I can think of a couple of ways to fix it:
**Current**
```java
// Definition
ListenableFuture<AckReply> receiveMessage(PubsubMessage message);

// Sample implementation
ListenableFuture<AckReply> receiveMessage(final PubsubMessage message) {
  return executor.submit(new Callable<AckReply>() {
    @Override
    AckReply call() {
      if (someLongComputation(message)) {
        return AckReply.ACK;
      }
      return AckReply.NACK;
    }
  });
}
```

**Option 1**
```java
// Definition
void receiveMessage(PubsubMessage message, SettableFuture<AckReply> response);

// Sample implementation
void receiveMessage(final PubsubMessage message, final SettableFuture<AckReply> response) {
  ListenableFuture<AckReply> future = executor.submit(new Callable<AckReply>() {
    @Override
    AckReply call() {
      if (someLongComputation(message)) {
        return AckReply.ACK;
      }
      return AckReply.NACK;
    }
  });
  response.setFuture(future);
}
```
In this way, we create the future for the user and can make sure the callback is set before the user can do anything with it. If the user takes a long time to respond to a job, job scheduling is still left to the user. Eg, if the user wants to process many messages in a threadpool, the user must create the threadpool themselves.

**Option 2**
```java
// Definition
AckReply receiveMessage(PubsubMessage message);
// This is essentially the same with Function<PubsubMessage, AckReply>

// Sample implementation
AckReply receiveMessage(PubsubMessage message) {
  return someLongComputation(message);
}
```
We can call the function from a threadpool ourselves. The function must be thread safe. We can increase the number of threads in the pool to cope with extra load. The user can always set the number of threads to suit their work load.

pongad:
cc @davidtorres

Fixes #1577. The detailed description of the bug is available in the
issue.

This commit fixes the bug by ensuring that the Future received by
MessageReceiver cannot be completed before a callback is added to it.

@garrettjonesgoogle I know that we are moving away from exposing Guava types. Since this branch has not been moved to `RpcFuture` yet, I think it's best to keep using Guava types for now. I'm mostly afraid that moving types might change the timing just enough that this bug become more difficult to reproduce.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageReceiver.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageReceiver.java
@@ -39,5 +39,5 @@ public interface MessageReceiver {
    * Called when a message is received by the subscriber. The implementation must arrange for {@code
    * reponse} to be set after processing the {@code message}.
    */
-  void receiveMessage(PubsubMessage message, SettableFuture<AckReply> response);
+  void receiveMessage(final PubsubMessage message, final SettableFuture<AckReply> response);
 }


garrettjonesgoogle:
LGTM, but let @davidtorres take a look too.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageReceiver.java
```diff
@@ -34,11 +34,10 @@
...
+
   /**
-   * Called when a message is received by the subscriber.
-   *
-   * @return A future that signals when a message has been processed.
+   * Called when a message is received by the subscriber. The implementation must arrange for {@code
+   * reponse} to be set after processing the {@code message}.
    */
-  ListenableFuture<AckReply> receiveMessage(PubsubMessage message);
+  void receiveMessage(PubsubMessage message, SettableFuture<AckReply> response);
```
davidtorres: make this:

void receiveMessage(final PubsubMessage message, final SettableFuture<AckReply> response);

garrettjonesgoogle: From what I understand, the 'final' doesn't change the signature - an interface can omit the `final`, and the implementations can add the `final`. 

http://stackoverflow.com/questions/5380177/final-arguments-in-interface-methods-whats-the-point

pongad:
@garrettjonesgoogle is right that adding `final` in interface signature doesn't actually do anything, but it could serve as a reminder for users to be careful since we're dealing with concurrent code.
@davidtorres PTAL

PR #1586, opened 2017-02-02
fix race in SubscriberImplTest::receiveMessage

pongad:
cc @davidtorres

This bug manifests as testBundleAcks deadlocking.

The test
1. publishes a series of messages to a mock pubsub server,
2. sets up a Subscriber client, waits to receive the messages back,
3. and acknowledges them.

For performance reasons, Subscriber does not immediately
send the acknowledgement request to the server.
Instead, it sends acks in batch every 100ms.
Using a fake clock, the test advances the time by 100ms,
sending the ack-batch, then verify that the mock server
receives the acks.

The bug is in step 2. The test thread waits by waiting on a
CountDownLatch, which is counted down by GRPC thread calling
receiveMessage().
However, the method decrements the latch before acking the message.

On a bad day, the test thread can [wake up](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/d325afbc4f0ed49d2a98c092992e5f7117f6a1da/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java#L461),
[advance the clock](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/d325afbc4f0ed49d2a98c092992e5f7117f6a1da/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java#L246),
and [send the ack-batch](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/d325afbc4f0ed49d2a98c092992e5f7117f6a1da/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java#L311)
before the GRPC thread could [add the message to the batch](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/d325afbc4f0ed49d2a98c092992e5f7117f6a1da/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java#L129).
The test then [waits ](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/d325afbc4f0ed49d2a98c092992e5f7117f6a1da/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java#L248)
for the server to receive an ack it never sent, deadlocking the test.

The fix is for receiveMessage() to ack the message before
decrementing the counter.

### DIFF
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java
@@ -115,10 +115,6 @@ public class SubscriberImplTest {
 
     @Override
     public void receiveMessage(PubsubMessage message, SettableFuture<AckReply> response) {
-      if (messageCountLatch.isPresent()) {
-        messageCountLatch.get().countDown();
-      }
-
       if (explicitAckReplies) {
         try {
           outstandingMessageReplies.put(response);
@@ -127,6 +123,10 @@ public class SubscriberImplTest {
         }
       } else {
         replyTo(response);
+      }
+      
+      if (messageCountLatch.isPresent()) {
+        messageCountLatch.get().countDown();
       }
     }
 


garrettjonesgoogle:
Your description is hard to follow. Could you add some qualifiers? If the following are incorrect, it indicates that I didn't understand correctly:

* (mock) server -> (mock) pubsub server
* client -> subscriber client

Add explicit references to where the following are performed in the code:

* "the test thread can [...] send the ack-batch"
* "before the GRPC thread could add the message to the batch"
* "The test then waits for the server to receive an ack"

It's hard to verify the flow if I have to do the work to correlate these descriptions with the code.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #1589, opened 2017-02-03
add zone to GAE Flex logging enhancer

gregw:
The google stackdriver logging console requires that the zone label be set in the monitored resource before it can aggregate logs by trace ID.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -177,8 +177,7 @@ public class LoggingHandler extends Handler {
       String resourceType = helper.getProperty(className + ".resourceType", "global");
       MonitoredResource resource = monitoredResource != null ? monitoredResource : getDefaultResource(resourceType);
       writeOptions = new WriteOption[]{WriteOption.logName(logName), WriteOption.resource(resource)};
-    }
-    catch (Exception ex) {
+    } catch (Exception ex) {
       reportError(null, ex, ErrorManager.OPEN_FAILURE);
       throw ex;
     }


meltsufin:
@garrettjonesgoogle Can you please take a look?

garrettjonesgoogle:
There is a release going out on Tuesday that I don't want to destabilize; I'll take a look after that goes out.

google-cloud-logging/src/main/java/com/google/cloud/logging/GaeFlexLoggingEnhancer.java
```diff
@@ -65,13 +70,29 @@ public GaeFlexLoggingEnhancer() {
...
+    if (System.getenv("GAE_SERVICE") != null) {
+      builder.addLabel("module_id", System.getenv("GAE_SERVICE"));
+    }
+    if (System.getenv("GAE_VERSION") != null) {
+      builder.addLabel("version_id", System.getenv("GAE_VERSION"));
+    }
+    try {
+      HttpURLConnection metadata = (HttpURLConnection) new URL(
+          "http://metadata.google.internal/computeMetadata/v1/instance/zone").openConnection();
+      metadata.setRequestProperty("Metadata-Flavor", "Google");
```
garrettjonesgoogle: In the one other place where we do a call like this, https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L409 , it sets `X-Google-Metadata-Request` - is that not necessary here?

gregw: I followed the documentation [here](https://cloud.google.com/compute/docs/storing-retrieving-metadata), which says

> Note: Previously, the `X-Google-Metadata-Request: True` header was required in requests. Both of these headers are still supported but it is recommended that you use the `Metadata-Flavor` header rather than the `X-Google-Metadata-Request: True`header.

I have verified that it works with and does not work without the `Metadata-Flavor` header.

Should we correct the other usage in this PR?

meltsufin:
@garrettjonesgoogle Would this make into the next release, and when is it scheduled to go out?

google-cloud-logging/src/main/java/com/google/cloud/logging/GaeFlexLoggingEnhancer.java
garrettjonesgoogle: Maybe just add a TODO at the other location - I don't want to perturb anything more than necessary right now.

garrettjonesgoogle:
This should make it into the next release. I'm intending on doing a release tomorrow (Friday).

gregw:
@garrettjonesgoogle I have fixed those issues. Note the unit tests are not passing, but neither are they in master (for me).  I'll try to investigate later today.

garrettjonesgoogle:
LGTM - once you figure out the tests I will merge it.

PR #1590, opened 2017-02-03
Fixing release script bugs, making formal releases manual

### LINKED ISSUE #1531: Convert to manually-triggered release process
For the last several releases, the release process hasn't run properly in Travis, and I have had to run various components manually. It's really hard to figure out what happened and didn't happen during the release process because there is no messaging. So, I would like to change the release scripts (specifically https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/utilities/after_success.sh ) so that they are designed to be run from the command line by a team member instead of by Travis. I intend to make the process a one-shot command so that only two steps are required in the happy case: 1. start the script, 2. verify it succeeded.

I am filing this as an issue to get any comments anyone has about this.

garrettjonesgoogle:
* Not auto-releasing after close (so that releases can be staged)
* Splitting release.sh into stage_release.sh and finalize_release.sh
* Stripping the qualifier from the SITE_VERSION for generating the gh-pages site
* Fixing deploy.sh (the input checking was wrong)
* Not running any tests in deploy.sh (they are already run in other contexts
  before this point)
* Updating RELEASING.md with the new process

Fixes #1531 .

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,18 +1,25 @@
 ### Overview
 
-Most of the release process is handled by the `after_success.sh` script, triggered after Travis CI successfully completes a non-PR build.  A new artifact will be released to Maven Central Repository via Travis CI when "-SNAPSHOT" is not included in the version (as listed in the base directory's `pom.xml`).  The website and README files will also be updated automatically in this case.  When "-SNAPSHOT" is included in the version, Travis only updates the artifact in the snapshot repository.
+The release process for SNAPSHOT versions is handled by the `after_success.sh` script, triggered after Travis CI successfully completes a non-PR build. A new SNAPSHOT artifact will be released to the snapshot repository as part of this script.
 
 ### To push a release version
 
-1. Run `utilities/update_pom_version.sh` from the repository's base directory.
-This script takes an optional argument denoting the new version.  By default, if the current version is X.Y.Z-SNAPSHOT, the script will update the version in all the pom.xml files to X.Y.Z.  If desired, another version can be supplied via command line argument instead.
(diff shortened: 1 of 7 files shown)


garrettjonesgoogle:
I need a new approval because the existing approval was on a completely different set of diffs.

RELEASING.md
```diff
@@ -1,18 +1,25 @@
...
 
-3. Before moving on, verify that the artifacts have successfully been pushed to the Maven Central Repository.  Open Travis CI, click the ["Build History" tab](https://travis-ci.org/GoogleCloudPlatform/google-cloud-java/builds), and open the second build's logs for Step 2's PR.  Be sure that you are not opening the "Pull Request" build logs.  When the build finishes, scroll to the end of the log and verify that the artifacts were successfully staged and deployed.  Search for `google-cloud` on the [Sonatype website](https://oss.sonatype.org/#nexus-search;quick~google-cloud) and check the latest version number.  In rare cases (when the Maven plugin that determines the version of the repository fails), the artifacts may not be deployed even if the version in the pom.xml files doesn't contain `SNAPSHOT`.  If the artifacts weren't deployed due to invalid version parsing or a flaky test, rerun the build.
+  ```
+  git commit -m "Release [VERSION HERE]"
```
shinfan: It seems always the case that we would create a commit everytime we update the version. Why not make `utilities/update_pom_version.sh` does the git commit automatically?

garrettjonesgoogle: This is the status quo - the preexisting release process requires it too. I would prefer to change this part later if we want to change it.

shinfan:
LGTM

PR #1591, opened 2017-02-06
merge master into pubsub-hp

pongad:
This commit merges master into pubsub-hp, in preparation for merging back.

This commit adds back previously deleted hand-written pubsub classes.
There is no new surface or feature, only renamings to reflect name changes in gax.

There are still ListenableFuture in the surface. I plan to move to RpcFuture in another PR since this is already humongous.


garrettjonesgoogle:
LGTM

PR #1592, opened 2017-02-06
fix race in FakeSubscriberServiceImpl

pongad:
This is the same race as
https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1495
but in sendStreamResponse rather than sendError.


garrettjonesgoogle:
LGTM

PR #1599, opened 2017-02-08
remove ListenableFuture from surface

pongad:
This won't compile yet, since it needs change in gax:
https://github.com/googleapis/gax-java/pull/200

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/AckReplyConsumer.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/AckReplyConsumer.java
@@ -17,7 +17,7 @@
 package com.google.cloud.pubsub.spi.v1;
 
 /**
- * Used by {@link MessageReceiver}, {@code AckReplyConsumer} accepts an {@link MessageReceiver.AckReply}, sending it to the service.
+ * Accepts a reply, sending it to the service.
  *
  * <p>Both the interface and its method is named after the Java 8's {@code BiConsumer} interface
  * to make migration to Java 8 and adopting its patterns easier.


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageReceiver.java
```diff
@@ -36,8 +35,18 @@
...
-   * reponse} to be set after processing the {@code message}.
+   * Accepts a reply, sending it to the service.
+   *
+   * <p>Both the interface and its method is named after the Java 8's {@code BiConsumer} interface
+   * to ease migration when we finally move.
    */
-  void receiveMessage(final PubsubMessage message, final SettableFuture<AckReply> response);
+  interface AckReplyConsumer {
+    void accept(AckReply ackReply, Throwable t);
+  }
```
garrettjonesgoogle: You're embedding again! Move this out.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageReceiver.java
```diff
@@ -36,8 +35,18 @@
   }
 
   /**
-   * Called when a message is received by the subscriber. The implementation must arrange for {@code
-   * reponse} to be set after processing the {@code message}.
+   * Accepts a reply, sending it to the service.
+   *
+   * <p>Both the interface and its method is named after the Java 8's {@code BiConsumer} interface
+   * to ease migration when we finally move.
```
garrettjonesgoogle: "when we finally move" -> "when we finally migrate to Java 8 and adopt its patterns"

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -287,7 +292,44 @@ public void run() {
           });
     }
 
-    return publishResult;
+    return new ListenableFutureDelegate<String>(publishResult);
+  }
+
+  private static class ListenableFutureDelegate<V> extends SimpleForwardingListenableFuture<V>
```
garrettjonesgoogle: This is copying what's in GAX, right? At least document it. We'll be able to get rid of it later, right?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageReceiver.java
pongad: Done

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageReceiver.java
pongad: Done. Moved `AckReply` out as well.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
pongad: No, this can only be copied. It is true that `ListenableFutureDelegate` class is also in GAX, but we cannot make the class public. If it were public, it becomes a part of GAX's surface and we defeat the purpose of not exposing `ListenableFuture`.

Unfortunately we'll just have to copy this to every package that needs it.

pongad:
@garrettjonesgoogle PTAL

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
garrettjonesgoogle: What I mean is that I think we can get rid of it when we convert `Publisher` to use `PublisherClient` since `PublisherClient` exposes `RpcFuture` instead of `ListenableFuture`.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/AckReplyConsumer.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.pubsub.spi.v1;
+
+/**
+ * Used by {@link MessageReceiver}, {@code AckReplyConsumer} accepts an {@link MessageReceiver.AckReply}, sending it to the service.
```
garrettjonesgoogle: I don't think you're supposed to use `@link`s in the first sentence of javadoc.

garrettjonesgoogle:
In the future I think it would be easier to do a PR first against `api-client-staging`, and then against `google-cloud-java`.

garrettjonesgoogle:
LGTM

PR #1608, opened 2017-02-10
pubsub: new publish snippet

pongad:
the subscriber side soon to follow

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
@@ -178,8 +178,7 @@ public class Publisher {
    * current thread until there are more resources available to publish.
    *
    * <p>Example of publishing a message.
-   *
-   * <pre>{@code
+   * <pre> {@code
    * String message = "my_message";
    * ByteString data = ByteString.copyFromUtf8(message);
    * PubsubMessage pubsubMessage = PubsubMessage.newBuilder().setData(data).build();
@@ -554,8 +553,7 @@ public class Publisher {
    * Constructs a new {@link Builder} using the given topic.
    *
    * <p>Example of creating a {@code Publisher}.
-   *
-   * <pre>{@code
+   * <pre> {@code
    * String projectName = "my_project";
    * String topicName = "my_topic";
    * TopicName topic = TopicName.create(projectName, topicName);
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
LGTM, but wait for @lesv to review and don't merge before Tue PST.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Example of publishing messages.
+   */
+  // [TARGET publish(PubsubMessage)]
+  // [VARIABLE "my_project_name"]
+  // [VARIABLE "my_topic_name"]
+  public void publish(String projectName, String topicName) throws Exception {
+    // [START publish]
+    Publisher publisher = Publisher.newBuilder(TopicName.create(projectName, topicName)).build();
+    List<String> messages = Arrays.asList("message1", "message2");
```
lesv: You might wish to mention that Publisher is a very long lived thing.  I'd imaging that Publisher and message list would be what's passed to the function (perhaps from some other function), and the activity just does it's thing.  You could even call it from the loop as I don't see how the loop adds anything useful to the example.

pongad:
@lesv PTAL. I'm not 100% sure I understand you correctly, but please let me know if this looks better. I also wanted to make sure the sample for creating the publisher also shuts it down. There're lots of threads to leak if you forget.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
```diff
@@ -0,0 +1,75 @@
...
+  public PublisherSnippets(Publisher publisher) {
+    this.publisher = publisher;
+  }
+
+  /**
+   * Example of publishing a message.
+   */
+  // [TARGET publish(PubsubMessage)]
+  // [VARIABLE "my_message"]
+  public void publish(String message) {
```
lesv: Looks a lot better.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -635,6 +622,30 @@ private Builder(TopicName topic) {
      * <p>For performance, this client benefits from having multiple channels open at once. Users
      * are encouraged to provide instances of {@code ChannelProvider} that creates new channels
      * instead of returning pre-initialized ones.
+     *
+     * <p>Example of creating a {@code Publisher}.
```
lesv: This looks good, though it might be in the wrong place ??

pongad:
@lesv PTAL. Looks like the snippet-insertion script doesn't work well with one-line comments or something.

garrettjonesgoogle:
LGTM with the understanding that tests will come next

PR #1613, opened 2017-02-13
add Subscriber snippet

pongad:
The snippet is very long, but the shortest I can make it without loosing
details.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/GrpcServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/GrpcServiceOptions.java
@@ -313,7 +313,7 @@ public abstract class GrpcServiceOptions<ServiceT extends Service<OptionsT>, Ser
     InstantiatingChannelProvider.Builder builder = InstantiatingChannelProvider.newBuilder()
         .setServiceAddress(hostAndPort.getHostText())
         .setPort(hostAndPort.getPort())
-        .setClientLibHeader(getLibraryName(), firstNonNull(getLibraryVersion(), ""));
+        .setClientLibHeader(getGoogApiClientLibName(), firstNonNull(getLibraryVersion(), ""));
     Credentials scopedCredentials = getScopedCredentials();
     if (scopedCredentials != null && scopedCredentials != NoCredentials.getInstance()) {
       builder.setCredentialsProvider(FixedCredentialsProvider.create(scopedCredentials));
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -74,6 +74,7 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
   private static final String MANIFEST_VERSION_KEY = "Implementation-Version";
   private static final String ARTIFACT_ID = "google-cloud-core";
   private static final String LIBRARY_NAME = "gcloud-java";
+  private static final String X_GOOGLE_CLIENT_HEADER_NAME = "gccl";
   private static final String LIBRARY_VERSION = defaultLibraryVersion();
   private static final String APPLICATION_NAME =
       LIBRARY_VERSION == null ? LIBRARY_NAME : LIBRARY_NAME + "/" + LIBRARY_VERSION;
@@ -599,6 +600,13 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
    */
   public String getLibraryName() {
     return LIBRARY_NAME;
+  }
+
+  /**
+   * Returns the library's name used by x-goog-api-client header as a string.
+   */
+  public String getGoogApiClientLibName() {
+    return X_GOOGLE_CLIENT_HEADER_NAME;
   }
 
   /**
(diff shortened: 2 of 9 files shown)


garrettjonesgoogle:
We shouldn't need locking for a simple example - how can we simplify this?

lesv:
It does seem overly complex and the locks shouldn't be needed in typical usage.  You really only need them if you are trying to run everything on the main thread, instead of thinking threaded and async.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
```diff
@@ -0,0 +1,100 @@
...
+    final AtomicBoolean done = new AtomicBoolean();
+
+    MessageReceiver receiver = new MessageReceiver() {
+      public void receiveMessage(final PubsubMessage message, final AckReplyConsumer consumer) {
+        System.out.println("got message: " + message);
+        consumer.accept(AckReply.ACK, null);
+        if (pendingReceives.decrementAndGet() != 0) {
+          return;
+        }
+        lock.lock();
```
lesv: We don't need lock(s).

Mention that this is processing the message.

pongad:
@lesv @garrettjonesgoogle PTAL. I'll add the tests for both Subscriber and Publisher in one go. They kind of belong in the same test anyway.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -207,6 +177,66 @@ public boolean isRunning() {
...
+   * final AtomicBoolean done = new AtomicBoolean();
+   *
+   * MessageReceiver receiver = new MessageReceiver() {
+   *   public void receiveMessage(final PubsubMessage message, final AckReplyConsumer consumer) {
+   *     System.out.println("got message: " + message);
+   *     consumer.accept(AckReply.ACK, null);
+   *     if (pendingReceives.decrementAndGet() != 0) {
+   *       return;
+   *     }
+   *     lock.lock();
```
garrettjonesgoogle: This still needs to be updated based on your change in `SubscriberSnippets`.

pongad:
@garrettjonesgoogle My mistake, PTAL

garrettjonesgoogle:
LGTM with the understanding that tests will come next

pongad:
@lesv PTAL

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
```diff
@@ -0,0 +1,80 @@
...
+        done.setException(failure);
+      }
+    }, new Executor() {
+      public void execute(Runnable command) {
+        command.run();
+      }
+    });
+    subscriber.startAsync();
+
+    done.get(10, TimeUnit.MINUTES);
```
lesv: You definitely need to comment this as most users won't need this except, perhaps on exit.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
```diff
@@ -0,0 +1,80 @@
...
+    // [START startAsync]
+    final AtomicInteger pendingReceives = new AtomicInteger(receiveNum);
+    final SettableRpcFuture<Void> done = new SettableRpcFuture<>();
+
+    MessageReceiver receiver = new MessageReceiver() {
+      public void receiveMessage(final PubsubMessage message, final AckReplyConsumer consumer) {
+        System.out.println("got message: " + message);
+        consumer.accept(AckReply.ACK, null);
+        if (pendingReceives.decrementAndGet() == 0) {
+          done.set(null);
```
lesv: This isn't actually a part of Async, as it's based on the assumption that we are dealing w/ n-things, which may or may not be active.  It might be better to move some of this into the test, and focus on the common / active thinngs.

pongad:
@lesv I think the updated snippets should do what you want. I think it gives us just enough room to properly test the samples too.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/MessageReceiverSnippets.java
```diff
@@ -0,0 +1,61 @@
...
+   * This example is for illustration. Implementations may directly process messages
+   * instead of sending them to queues.
+   */
+  // [TARGET receiveMessage(PubsubMessage, AckReplyConsumer)]
+  public MessageReceiver messageReceiver() {
+    // [START receiveMessage]
+    MessageReceiver receiver = new MessageReceiver() {
+      public void receiveMessage(final PubsubMessage message, final AckReplyConsumer consumer) {
+        if (blockingQueue.offer(message)) {
+          consumer.accept(AckReply.ACK, null);
```
lesv: Probably before the `.accept()`
```
// ...
// Process your message here
// ...
```

lesv: strike this.

pongad:
@lesv Thank you for the rigorous code review. I'll wait for Travis, then merge this.

PR #1625, opened 2017-02-15
pubsub: start with polling by default

pongad:
cc @davidtorres 

Previously, Subscribers start with streaming pull and fall back
to polling pull if streaming is unavailable.
To prevent fallback or streaming from causing any problem in production,
this commit starts Subscribers with polling pull directly.
Streaming pull will be re-enabled once the endpoint is working.

Since the ultimate fate of fallback is up in the air,
I chose to comment out the start-with-streaming code instead of
deleting.

This change also smoked out a logic bug that causes a race condition.
A message can be "nacked" either by calling AckReplyConsumer::accept
with an explicit NACK or a throwable signally an error.
The explicit NACK case works properly.

In case of a throwable, MessageDispatcher did not set the ["acked" flag](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/5311bb39021f97722bd8a549ed703e4c3d9b7c66/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java#L156).
(Acks and nacks share most of the same code path;
they might as well use the same flag.)
This causes two things to happen concurrently.

1. Since the message is being nacked, it is added to the [nack list](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/5311bb39021f97722bd8a549ed703e4c3d9b7c66/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java#L173)
to be reported to the pubsub service.

2. Pubsub client automatically [extends deadlines](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/5311bb39021f97722bd8a549ed703e4c3d9b7c66/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java#L406) of the messages
the client's user is processing. Since acked flag is not set,
the client also tries to extend the message's deadline.

If (1) happens first, the test code [sees](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/5311bb39021f97722bd8a549ed703e4c3d9b7c66/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java#L225) that the message is being
nacked and the test passes even though the client will later
incorrectly extend the message's deadline.

If (2) happens first, the test code sees the incorrect deadline
extension and fails.

The fix is simple: set the acked flag.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -113,7 +113,7 @@ public class Subscriber {
       20 * 1024 * 1024; // 20MB API maximum message size.
   private static final int INITIAL_ACK_DEADLINE_SECONDS = 10;
   private static final int MAX_ACK_DEADLINE_SECONDS = 600;
-  private static final int MIN_ACK_DEADLINE_SECONDS = 10;
+  static final int MIN_ACK_DEADLINE_SECONDS = 10;
   private static final Duration ACK_DEADLINE_UPDATE_PERIOD = Duration.standardMinutes(1);
   private static final double PERCENTILE_FOR_ACK_DEADLINE_UPDATES = 99.9;
 
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
@@ -90,6 +90,7 @@ final class PollingSubscriberConnection extends AbstractService implements AckPr
             flowController,
             executor,
             clock);
+    messageDispatcher.setMessageDeadlineSeconds(Subscriber.MIN_ACK_DEADLINE_SECONDS);
   }
 
   @Override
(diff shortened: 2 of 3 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -113,7 +113,7 @@
       20 * 1024 * 1024; // 20MB API maximum message size.
   private static final int INITIAL_ACK_DEADLINE_SECONDS = 10;
   private static final int MAX_ACK_DEADLINE_SECONDS = 600;
-  private static final int MIN_ACK_DEADLINE_SECONDS = 10;
+  static final int MIN_ACK_DEADLINE_SECONDS = 10;
```
garrettjonesgoogle: Add `@VisibleForTesting`

pongad: It's not just for tests though, it is also used in `PollingSubscriberConnection`.

garrettjonesgoogle: Ok I'll forget about it for now.

PR #1638, opened 2017-02-21
Update client name in header for Veneer layer

shinfan:
`gcloud-java` -> `gccl`

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -605,7 +605,7 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
   /**
    * Returns the library's name used by x-goog-api-client header as a string.
    */
-  public String getGrpcLibraryName() {
+  public String getGoogApiClientLibName() {
     return X_GOOGLE_CLIENT_HEADER_NAME;
   }
 
--- a/google-cloud-core/src/main/java/com/google/cloud/GrpcServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/GrpcServiceOptions.java
@@ -313,7 +313,7 @@ public abstract class GrpcServiceOptions<ServiceT extends Service<OptionsT>, Ser
     InstantiatingChannelProvider.Builder builder = InstantiatingChannelProvider.newBuilder()
         .setServiceAddress(hostAndPort.getHostText())
         .setPort(hostAndPort.getPort())
-        .setClientLibHeader(getGrpcLibraryName(), firstNonNull(getLibraryVersion(), ""));
+        .setClientLibHeader(getGoogApiClientLibName(), firstNonNull(getLibraryVersion(), ""));
     Credentials scopedCredentials = getScopedCredentials();
     if (scopedCredentials != null && scopedCredentials != NoCredentials.getInstance()) {
       builder.setCredentialsProvider(FixedCredentialsProvider.create(scopedCredentials));


google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -73,7 +73,7 @@
   private static final String MANIFEST_ARTIFACT_ID_KEY = "artifactId";
   private static final String MANIFEST_VERSION_KEY = "Implementation-Version";
   private static final String ARTIFACT_ID = "google-cloud-core";
-  private static final String LIBRARY_NAME = "gcloud-java";
```
garrettjonesgoogle: So I traced through the code, and this is being used in the User Agent string. I don't think we want to change this. It doesn't appear that Storage/Datastore/BigQuery are setting `x-goog-api-client`.

shinfan: So this PR is going to inject the 'gccl' header for Veneer/Vtk clients. I will submit a separate PR for Veneer only clients.
Are you suggesting we use 'gccl' in the x-google-api-client header and 'gcloud-java' in user agent? Wouldn't that be inconsistent and confusing?

garrettjonesgoogle: The only Veneer+Vtk client right now is Logging. It doesn't set any headers on the GAPIC client:

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java

@bjwatson Can you comment on the user-agent situation? We've been sending `gcloud-java` in user-agent for quite a while now, and I don't know if there are things that depend on it. We avoided changing it to `google-cloud-java` even though everything else was renamed several months ago.

shinfan: Wait..correct me if I am wrong though, `LoggingOptions` is a subclass of `GrpcServiceOptions` which is a subclass of `ServiceOptions`. If I make a change in `ServiceOptions` to set the header, Logging will get the header set automatically.

garrettjonesgoogle: As confirmed offline: `GrpcServiceOptions` sets the headers we expect (`x-goog-api-client`), but changing `LIBRARY_NAME` also affects the `User-Agent` value for non-gRPC clients. For now, let's separate the value used for gRPC and non-gRPC.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -602,6 +603,13 @@ public String getLibraryName() {
   }
 
   /**
+   * Returns the library's name used by x-goog-api-client header as a string.
+   */
+  public String getGrpcLibraryName() {
```

garrettjonesgoogle:
LGTM

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
bjwatson: @garrettjonesgoogle We definitely want to leave UA alone for the discovery-based APIs like BQ, Storage, and Translate. We're not updating how these are handled server-side for now.

For the gRPC-based APIs, we can stop doing anything in UA; no one is depending on what we have been doing. It would actually be beneficial to do so for server-side performance, but this is not a hard requirement for this round of header updates.

garrettjonesgoogle: Ok, that confirms what I thought. 

The gRPC-based APIs in Java never actually set the UA.

So for the indefinite future, we'll keep the UA for BQ, Storage, Datastore, and Translate, and only use `x-goog-api-client` for gRPC-based clients.

bjwatson: Is Datastore discovery-based on Java? If not, we'll need to use the custom `x-goog-api-client` header. Since this is not on GAPIC yet, we'll just need to implement the entire header string for Datastore or we won't have any usage metrics for it.

Using UA is correct for the other three services.

garrettjonesgoogle: Yes, Datastore is discovery-based in Java. Your double-negation didn't lead to a clear interpretation: are you stating that we need to implement `x-goog-api-client` for Datastore? Datastore is currently sending UA in Java.

bjwatson: If Datastore is discovery-based, then you can leave it alone. We're just concerned with gRPC-based client libraries right now (whether or not they use GAPIC).

Issue #1639, opened 2017-02-21
Enable compression for pubsub library


lucmult:
As a customer I would like to have communication compressed from pubsub client to pubsub service.

This allows saving on network bandwidth when running it from other Cloud providers for example.

garrettjonesgoogle:
Fixed in #1645 ; it will be included in the next release.

ryankillory:
Given that https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1645 appears to have been reverted, I was wondering whether this should be reopened.

Thanks!

georgespalding:
Seeing that it was a year ago someone asked about this, I'd like to ask again:
We would like to use compression at the Pub/Sub level to avoid doing it at application level.

The pull request and this issue mentions nothing to motivate why it was reverted. 
If there is a good reason for not allowing compression it would be nice to understand

ostlerc:
Adding this for posterity. the revert commit says grpc [doesn't support compression yet](https://github.com/googleapis/google-cloud-java/commit/e1ca0d55cbc7da58cc1310698beb7ea40fdf7129).

> Compression is not fully supported in gRPC, can't have it in the library
yet.

georgespalding:
The lack of this functionality does really hurt now that Google Cloud started charging for inter region pubsub egress network usage.

PR #1645, opened 2017-02-22
Enabling client-side compression for Pub/Sub

davidtorres:
Partially addressing #1639, changes will need to be made to the gax.grpc classes to support compression.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
@@ -39,6 +39,7 @@ import com.google.common.util.concurrent.SettableFuture;
 import com.google.pubsub.v1.PublishRequest;
 import com.google.pubsub.v1.PublishResponse;
 import com.google.pubsub.v1.PublisherGrpc;
+import com.google.pubsub.v1.PublisherGrpc.PublisherFutureStub;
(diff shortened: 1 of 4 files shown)


davidtorres:
@pongad @garrettjonesgoogle for your review

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -696,6 +705,15 @@ public Builder setExecutorProvider(ExecutorProvider executorProvider) {
       this.executorProvider = Preconditions.checkNotNull(executorProvider);
       return this;
     }
+    
+    /** 
+     * Gives the ability to disable client-side compression. 
+     * Note compression is enabled by default. 
+     */
+    public Builder setCompressionEnabled(boolean enabled) {
```
pongad: The name `setCompressionEnabled` sounds a little strange to me. That said there are java UI compoenents with method `setEnabled` so I think the name is not that jarring.

garrettjonesgoogle: I think this is the established pattern.

garrettjonesgoogle:
LGTM too.

tomerr-ma:
Hi, Is there any plan to evaluate compression/gzip support?

tomarikawa:
I could not quite get, why is this reverted? Could somebody explain, please.

PR #1648, opened 2017-02-23
BigQuery: Adding support for schema autodetection

gene-telligent:
This has been exposed through the protobuf API already; I just coded in a method to access it through the Java SDK.

See: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load.autodetect

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ExternalTableDefinition.java
@@ -61,6 +61,7 @@ public class ExternalTableDefinition extends TableDefinition {
   private final Integer maxBadRecords;
   private final Boolean ignoreUnknownValues;
   private final String compression;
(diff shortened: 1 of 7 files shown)


garrettjonesgoogle:
@tswast could you take a look?

garrettjonesgoogle:
@tswast I don't think we've been putting annotations on methods to match the support level of the underlying API. It gets tricky when the support level of the underlying API changes, because then we might have to keep revising the annotations in the client library to match. 

The tag we have been using for features that might change in the client library is `@ExperimentalApi` (defined in protobuf).

gene-telligent:
Hi all,

I've rebased the commit with the latest from master (waiting on CI builds to complete now, but internal running of the tests raised no errors). 

I didn't use `@ExperimentalApi` as it seems to be an annotation reserved for the SPI libraries.

tswast:
@garrettjonesgoogle could you merge? I haven't been tracking when releases are happening. I don't want to step on the release process again.

PR #1651, opened 2017-02-23
Reverting compression changes

davidtorres:
Sorry for the churn but we realized that the compression is not fully supported in gRPC (it is marked as experimental and there are still doubts of its API https://github.com/grpc/grpc-java/issues/1704) and we don't want to depend on it but until this is flesh out.


davidtorres:
@pongad @garrettjonesgoogle for your review

garrettjonesgoogle:
It's ok, we didn't do a release with it yet. LGTM.

PR #1659, opened 2017-02-24
Make sure dependencies converges

rborer:
Hi,

I'm planning to use the BigQuery client in one of my project where I have have enabled dependency convergence (see [1]). Unfortunately, it seems that the client has quite a few dependencies that do not converge, so I've started to fix them as I cannot build my project.

This is a work in progress, please let me know if this would make sense to you or if I have to sort things out at my project level only.

Also, the policy I've chosen when a conflict has been detected is to always pick the highest version number.

I've also added in the modified pom.xml (only core project for the moment) which lower versions are excluded for reference.

Thanks,
Reynald

[1]: http://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryExceptionTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryExceptionTest.java
@@ -26,13 +26,13 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 
+import com.google.api.client.http.HttpHeaders;
+import com.google.api.client.http.HttpResponseException;
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryHelper.RetryHelperException;
-
-import org.junit.Test;
-
 import java.io.IOException;
 import java.net.SocketTimeoutException;
+import org.junit.Test;
 
 public class BigQueryExceptionTest {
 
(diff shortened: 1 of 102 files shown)


rborer:
I signed it!

rborer:
Now that I'm doing it, I see that there a quite a few common deps that does not converge (guava, joda-time, ...). Another solution could be to have a dependencyManagement block at the root pom.xml to force those libraries to be at the specified version globally. Thoughts?

google-cloud-bigquery/pom.xml
```diff
@@ -51,7 +51,6 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.12</version>
       <scope>test</scope>
```
garrettjonesgoogle: I'm not a maven expert so it would help to get some explanation - what is the difference between not having this dependency in this file, and having it here with no version?

rborer: Right, sorry I rushed a bit to have this out and forgot to explain its behaviour ;-).

On multi-modules project, when you have a dependency A which is configured in the `dependencyManagement` block of the `root pom.xml`, it instruct maven to always use this version if any child module has A as a dependency (may it be directly declared or transitive!).

For this kind of project size, I find this mechanism quite convenient as it allow to have a central location to define common versions, and they are inherited (and enforced) in all child modules. Thus, I also tend to not specify in child modules the version (for `junit` for example, as it's in the `dependencyManagement` block of the `root pom.xml`, there is not need to specify the version within `google-cloud-bigquery/pom.xml` (it's also a good hint to indicate that the version is managed elsewhere :-) .

Let me know if I'm clear enough :-)

Cheers,
Reynald

rborer:
Any news on this one? Let me know if my explanation is not clear or if I need to improve my PR :-)

garrettjonesgoogle:
I intend on circling back - we're trying to get a lot of breaking changes through at the moment so we're concentrating on that.

rborer:
Thanks @garrettjonesgoogle I fully understand, it looks like things are moving at an interesting pace. Ping me when you intend to have a look at this one, there will probably be some conflicts to fix first. Also, I think we should compare the output of `mvn dependency:tree` before & after to spot differences in dependencies.

garrettjonesgoogle:
@rborer I went ahead and re-implemented this with the latest versions in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1943 .

garrettjonesgoogle:
Closing out because #1943 was submitted.

rborer:
@garrettjonesgoogle thanks!

PR #1665, opened 2017-02-27
properly stop MessageDispatcher if Subscriber fails

pongad:
When PollingSubscriberConnection fails, it calls `notifyFailed`.
Subscriber listens to this failure and stops all connections before
failing itself.
Stopping connections is implemented with `stopAsync`.

The problem is with the connection that originally failed.
Since it has already entered the FAILED state, it cannot be stopped
by `stopAsync`.
Consequently, this leaks a MessageDispatcher.

This commit makes the connection explicitly stop itself before notifying
the Subscriber that it has failed.

cc @davidtorres


garrettjonesgoogle:
LGTM

PR #1670, opened 2017-02-28
POSC support for Storage

shinfan:
Fixes: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1602

Tests: Unit tests, Integration tests
cc @omaray


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -357,6 +363,7 @@ static CustomerEncryption fromPb(StorageObject.CustomerEncryption customerEncryp
 
     abstract Builder setCustomerEncryption(CustomerEncryption customerEncryption);
 
+
```
michaelbausor: Nit: remove this new line

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobInfoTest.java
```diff
@@ -66,6 +66,8 @@
   private static final String KEY_SHA256 = "keySha";
   private static final CustomerEncryption CUSTOMER_ENCRYPTION =
       new CustomerEncryption(ENCRYPTION_ALGORITHM, KEY_SHA256);
+  private static final String STORAGE_CLASS = "COLDLINE";
```
michaelbausor: We should make storage class string constants available (as per https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/766/files#diff-e26b094819cdfbac4ddb7249471814abR23)

shinfan:
@michaelbausor  PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.storage;
+
+/**
+ * String constants for the names of the storage classes.
+ */
+public class StorageClass {
```
michaelbausor: Maybe `StorageClasses` instead of `StorageClass`? The name `StorageClass` sounds like it is a storage class, not that it contains them.

garrettjonesgoogle: Why isn't this an enum? Other APIs seem to use enums - e.g. Compute:

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-compute/src/main/java/com/google/cloud/compute/AttachedDisk.java#L81

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-compute/src/main/java/com/google/cloud/compute/NetworkConfiguration.java#L43

shinfan:
@garrettjonesgoogle PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.storage;
+
+/**
+ * Enums for the storage classes.
+ * See https://cloud.google.com/storage/docs/storage-classes for details.
+ */
+public enum StorageClass {
```
michaelbausor: I don't know if we want to use an enum for this. We are using valueOf to convert from string to enum, which will throw an exception on an unexpected value. So if storage adds a new storage class, then we will be broken.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -100,7 +101,7 @@ public Bucket createBucketWithStorageClassAndLocation(String bucketName) {
     // [START createBucketWithStorageClassAndLocation]
     Bucket bucket = storage.create(BucketInfo.newBuilder(bucketName)
         // See here for possible values: http://g.co/cloud/storage/docs/storage-classes
-        .setStorageClass("COLDLINE")
+        .setStorageClass(StorageClass.COLDLINE)
```
garrettjonesgoogle: According to the URL above this line, the storage classes are in lower case. Do the integration tests pass?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -416,14 +416,14 @@ void populateCondition(Rule.Condition condition) {
...
-    public abstract Builder storageClass(String storageClass);
+    public abstract Builder storageClass(StorageClass storageClass);
 
     /**
      * Sets the bucket's storage class. This defines how blobs in the bucket are stored and
      * determines the SLA and the cost of storage. A list of supported values is available
      * <a href="https://cloud.google.com/storage/docs/storage-classes">here</a>.
      */
-    public abstract Builder setStorageClass(String storageClass);
+    public abstract Builder setStorageClass(StorageClass storageClass);
```
garrettjonesgoogle: This is a breaking change - can we mark it deprecated for now and make an overload with the new type?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -416,14 +416,14 @@ void populateCondition(Rule.Condition condition) {
      * <a href="https://cloud.google.com/storage/docs/storage-classes">here</a>.
      */
     @Deprecated
-    public abstract Builder storageClass(String storageClass);
+    public abstract Builder storageClass(StorageClass storageClass);
```
garrettjonesgoogle: Leave this deprecated function as it was.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
garrettjonesgoogle: 2017

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
shinfan: After put some thought I think enum is better. The benefit of having enum instead of string lower the user effort of constructing the value themselves (enums can autocomplete by IDE), it would be also clearer to know which values are by reading the enum class. Also, it would be a breaking change anyway if storage adds a new type of storage class. Even though we use strings, a new string value can still potentially break users' code.

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -549,12 +549,12 @@ public Builder setDeleteRules(Iterable<? extends DeleteRule> rules) {
...
     @Override
     @Deprecated
-    public Builder storageClass(String storageClass) {
+    public Builder storageClass(StorageClass storageClass) {
       return setStorageClass(storageClass);
     }
 
     @Override
-    public Builder setStorageClass(String storageClass) {
+    public Builder setStorageClass(StorageClass storageClass) {
```
garrettjonesgoogle: same comments here as for `BucketInfo`

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
shinfan: Are we going to release GA with any deprecated methods? I thought we are going to remove all the deprecated methods before GA.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
shinfan: This would become isolated once my deprecated PR is merged. so I think it would be more efficient to merge the deprecated PR before this one.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
garrettjonesgoogle: No, we'll move the deprecated methods before GA.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
garrettjonesgoogle: Sure, but you'll have a merge conflict if you don't change this back.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
shinfan: Yes, the integration tests pass. According to this url: https://cloud.google.com/storage/docs/managing-buckets#manage-class the class values are in upper case.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
garrettjonesgoogle: There are a lot of other enums in the client library that you could say have the same risk. I think we should be consistent with the approach.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
shinfan: I would recommend to avoid adding any new deprecated methods before GA (since we are going to remove them soon anyway)

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
shinfan: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
shinfan: Done

shinfan:
@garrettjonesgoogle  PTAL

garrettjonesgoogle:
LGTM

PR #1672, opened 2017-02-28
Make class com.google.cloud.ServiceOptions$Builder public

rborer:
Hi,

I'm trying to use BigQuery client from clojure. Due to a [very old JVM bug](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4283544), the class com.google.cloud.ServiceOptions$Builder need to be flagged public to make configuration methods callable from clojure (clojure uses reflection).

Due to this bug, the following piece of code throws an exception:

```clojure
(-> (BigQueryOptions/newBuilder)
      (.setProjectId project-id))
```

=> 

`IllegalArgumentException Can't call public method of non-public class: public com.google.cloud.ServiceOptions$Builder com.google.cloud.ServiceOptions$Builder.setProjectId(java.lang.String)  clojure.lang.Reflector.invokeMatchingMethod (Reflector.java:88)
`

As replied on [the corresponding StackOverflow thread](http://stackoverflow.com/questions/38059977/cant-call-public-method-of-non-public-class-public-google-gcloud-library) there are work-arounds but they are non-trivial to implement.

Thanks in advance to consider this patch.

Cheers,
Reynald

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -104,7 +104,7 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, Service
    * @param <OptionsT> the {@code ServiceOptions} subclass corresponding to the service
    * @param <B> the {@code ServiceOptions} builder
    */
-  protected abstract static class Builder<ServiceT extends Service<OptionsT>, ServiceRpcT,
+  public abstract static class Builder<ServiceT extends Service<OptionsT>, ServiceRpcT,
       OptionsT extends ServiceOptions<ServiceT, ServiceRpcT, OptionsT>,
       B extends Builder<ServiceT, ServiceRpcT, OptionsT, B>> {
 


garrettjonesgoogle:
This was written before I started work on this project, and I can't figure out what the advantage is of having the `Builder` class be `protected`. I guess it's an attempt to prevent people from directly holding references to the `ServiceOptions$Builder` type? I can't figure out what kind of mistakes or inappropriate dependencies that would prevent. 

@aozarov maybe you have some insight here?

rborer:
maybe the purpose here was to prevent anyone to directly create an instance of this `Builder` class, but given that it's abstract I cannot figure out how one would be able to achieve that ;-)

aozarov:
Change looks fine.  

This class was "internal" as it is nasty from generics perspective and we wanted to force holding a 
reference to it and using its API only via its much cleaner sub-classes.  Not a big deal though.

garrettjonesgoogle:
Thanks, @aozarov !

rborer:
Thanks.

PR #1674, opened 2017-03-01
pubsub: fix potential deadlock in startConnections

pongad:
If a subscriber connection fails, awaitRunning will throw
IllegalStateException.
We must make sure that we count down the latch in either case;
otherwise startup will deadlock.

Also move the call to addListener above startAsync.
Otherwise, the listener might not see the connection failing
if it fails too quickly.

cc @davidtorres @mdietz94


garrettjonesgoogle:
Makes sense to me - @davidtorres can you take a look?

PR #1675, opened 2017-03-01
pubsub: make subscriber connections split requests properly

### LINKED ISSUE #1624: Subscriber uselessly partitions `PendingModifyAckDeadline` list
Looping in @davidtorres to make sure I understand this right.

In [this function](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/5311bb39021f97722bd8a549ed703e4c3d9b7c66/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java#L197), we seem to be splitting the list so that individual requests are under pubsub's API limit.

However, I think this is the wrong list to split. We create one RPC not per list of `PendingModifyAckDeadline` but per each `PendingModifyAckDeadline` object. If anything, we should be splitting the `modifyAckDeadline.ackIds`. My understanding is that `ackIds` can only be over the limit if the pubsub service gives us > 1000 messages in one call to pull. I don't think this can happen anyway, but would be happy to code defensively against it.

David, can you verify?

pongad:
Fixes #1624.

cc @davidtorres

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
@@ -40,6 +40,7 @@ import io.grpc.auth.MoreCallCredentials;
 import io.grpc.stub.ClientCallStreamObserver;
 import io.grpc.stub.ClientCalls;
 import io.grpc.stub.ClientResponseObserver;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -211,7 +212,7 @@ final class StreamingSubscriberConnection extends AbstractService implements Ack
   @Override
   public void sendAckOperations(
       List<String> acksToSend, List<PendingModifyAckDeadline> ackDeadlineExtensions) {
-    StreamingPullRequest[] requests =
+    List<StreamingPullRequest> requests =
         partitionAckOperations(acksToSend, ackDeadlineExtensions, MAX_PER_REQUEST_CHANGES);
     for (StreamingPullRequest request : requests) {
       requestObserver.onNext(request);
@@ -219,23 +220,23 @@ final class StreamingSubscriberConnection extends AbstractService implements Ack
   }
 
   @VisibleForTesting
-  static StreamingPullRequest[] partitionAckOperations(
+  static List<StreamingPullRequest> partitionAckOperations(
       List<String> acksToSend, List<PendingModifyAckDeadline> ackDeadlineExtensions, int size) {
     int numExtensions = 0;
     for (PendingModifyAckDeadline modify : ackDeadlineExtensions) {
       numExtensions += modify.ackIds.size();
     }
-    int numRequests = Math.max(numExtensions, acksToSend.size());
-    numRequests = numRequests / size + (numRequests % size == 0 ? 0 : 1);
+    int numChanges = Math.max(numExtensions, acksToSend.size());
+    int numRequests = numChanges / size + (numChanges % size == 0 ? 0 : 1);
 
-    StreamingPullRequest.Builder[] requests = new StreamingPullRequest.Builder[numRequests];
+    List<StreamingPullRequest.Builder> requests = new ArrayList<>(numRequests);
     for (int i = 0; i < numRequests; i++) {
-      requests[i] = StreamingPullRequest.newBuilder();
+      requests.add(StreamingPullRequest.newBuilder());
     }
 
     int reqCount = 0;
     for (List<String> acksChunk : Lists.partition(acksToSend, size)) {
-      requests[reqCount].addAllAckIds(acksChunk);
+      requests.get(reqCount).addAllAckIds(acksChunk);
       reqCount++;
     }
 
@@ -243,7 +244,8 @@ final class StreamingSubscriberConnection extends AbstractService implements Ack
     int ackCount = 0;
     for (PendingModifyAckDeadline modify : ackDeadlineExtensions) {
       for (String ackId : modify.ackIds) {
-        requests[reqCount]
+        requests
+            .get(reqCount)
             .addModifyDeadlineSeconds(modify.deadlineExtensionSeconds)
             .addModifyDeadlineAckIds(ackId);
         ackCount++;
@@ -254,9 +256,9 @@ final class StreamingSubscriberConnection extends AbstractService implements Ack
       }
     }
 
-    StreamingPullRequest[] ret = new StreamingPullRequest[requests.length];
-    for (int i = 0; i < requests.length; i++) {
-      ret[i] = requests[i].build();
+    List<StreamingPullRequest> ret = new ArrayList<>(requests.size());
+    for (StreamingPullRequest.Builder builder : requests) {
+      ret.add(builder.build());
     }
     return ret;
   }
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnectionTest.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnectionTest.java
@@ -19,102 +19,93 @@ package com.google.cloud.pubsub.spi.v1;
 import com.google.cloud.pubsub.spi.v1.MessageDispatcher.PendingModifyAckDeadline;
 import com.google.common.truth.Truth;
 import com.google.pubsub.v1.StreamingPullRequest;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import org.junit.Test;
 
 public class StreamingSubscriberConnectionTest {
-  private static List<StreamingPullRequest> partitionAckOperations(
-      List<String> acksToSend, List<PendingModifyAckDeadline> ackDeadlineExtensions, int size) {
-    List<StreamingPullRequest> requests = new ArrayList<>();
-    for (StreamingPullRequest req :
-        StreamingSubscriberConnection.partitionAckOperations(
-            acksToSend, ackDeadlineExtensions, size)) {
-      requests.add(req);
-    }
-    return requests;
-  }
-
   @Test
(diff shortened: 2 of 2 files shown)


google-cloud-pubsub/pom.xml
```diff
@@ -78,6 +78,12 @@
         <version>1.9.5</version>
         <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>0.30</version>
+      <scope>test</scope>
+    </dependency>
```
garrettjonesgoogle: Why is this being added?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
```diff
@@ -475,7 +473,7 @@ private void processOutstandingAckOperations(
       if (!pendingAcks.isEmpty()) {
         try {
           acksToSend = new ArrayList<>(pendingAcks);
-          logger.log(Level.INFO, "Sending {} acks", acksToSend.size());
+          logger.log(Level.INFO, "Sending " + acksToSend.size() + " acks");
```
garrettjonesgoogle: Why are you converting the log into string addition? The string formatting is more performant in the case that logging is not performed because it is higher than the log level (e.g. if it is set to `WARN`). (occurs in multiple places in this PR)

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
```diff
@@ -211,35 +211,45 @@ private boolean isAlive() {
...
-            requestBuilder
-                .addModifyDeadlineSeconds(modifyAckDeadline.deadlineExtensionSeconds)
-                .addModifyDeadlineAckIds(ackId);
-          }
-        }
-      }
-      if (ackChunksIt.hasNext()) {
-        List<String> ackChunk = ackChunksIt.next();
-        requestBuilder.addAllAckIds(ackChunk);
+  @VisibleForTesting
```
garrettjonesgoogle: You made this visible for testing but there doesn't appear to be any test changes to access it.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
```diff
@@ -211,35 +211,45 @@ private boolean isAlive() {
   @Override
   public void sendAckOperations(
       List<String> acksToSend, List<PendingModifyAckDeadline> ackDeadlineExtensions) {
+    StreamingPullRequest.Builder[] requests =
+        partitionAckOperations(acksToSend, ackDeadlineExtensions, MAX_PER_REQUEST_CHANGES);
+    for (StreamingPullRequest.Builder request : requests) {
+      requestObserver.onNext(request.build());
```
garrettjonesgoogle: It seems odd to have `partitionAckOperations` return an array of builders when the only thing it does is build them here. Why not call `build()` in `partitionAckOperations`?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
```diff
@@ -211,35 +211,45 @@ private boolean isAlive() {
...
+    int count = 0;
+    for (List<String> acksChunk : Lists.partition(acksToSend, size)) {
+      requests[count].addAllAckIds(acksChunk);
+      count++;
+    }
+
+    count = 0;
+    for (PendingModifyAckDeadline modify : ackDeadlineExtensions) {
+      for (String ackId : modify.ackIds) {
+        requests[count / size]
```
garrettjonesgoogle: what is `count / size` supposed to compute? If `size` is `10000` (`MAX_PER_REQUEST_CHANGES`), I have trouble seeing when this will ever be anything besides 0. 

Now I really want to see a test for this...

google-cloud-pubsub/pom.xml
pongad: This is for testing. I forgot to add the test file.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
pongad: Done.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
pongad: added test, forgot to `git add` last time

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
pongad: It was to avoid creating another array. That said, I now realized that the array is probably small. Creating the array is cheap enough that it shouldn't matter.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
pongad: I agree it's a lot. I can see this happening when we shut down though. Shutting down nacks all unprocessed messages, and nack is the same as setting deadline to 0.

pongad:
@garrettjonesgoogle PTAL

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnectionTest.java
```diff
@@ -0,0 +1,120 @@
...
+        partitionAckOperations(
+            Collections.<String>emptyList(), Collections.<PendingModifyAckDeadline>emptyList(), 3);
+    Truth.assertThat(requests).isEmpty();
+
+    requests =
+        partitionAckOperations(
+            Arrays.asList("a", "b", "c"), Collections.<PendingModifyAckDeadline>emptyList(), 3);
+    Truth.assertThat(requests)
+        .contains(
+            StreamingPullRequest.newBuilder().addAckIds("a").addAckIds("b").addAckIds("c").build());
```
garrettjonesgoogle: It would feel more intuitive to do a list comparison (if that is possible) instead of repeated containment comparisons.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnectionTest.java
```diff
@@ -0,0 +1,120 @@
...
+            3);
+    Truth.assertThat(requests)
+        .contains(
+            StreamingPullRequest.newBuilder().addAckIds("a").addAckIds("b").addAckIds("c").build());
+    Truth.assertThat(requests).contains(StreamingPullRequest.newBuilder().addAckIds("d").build());
+
+    requests =
+        partitionAckOperations(
+            Arrays.asList("a", "b", "c", "d"),
+            Arrays.asList(new PendingModifyAckDeadline(42, "A")),
```
garrettjonesgoogle: It's confusing to use a capital "A" here which is distinct from the lower-case "a" earlier - I'd prefer a distinct letter. Maybe start at "m" for ack deadline modifications.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnectionTest.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.List;
+import org.junit.Test;
+
+public class StreamingSubscriberConnectionTest {
+  private static List<StreamingPullRequest> partitionAckOperations(
+      List<String> acksToSend, List<PendingModifyAckDeadline> ackDeadlineExtensions, int size) {
+    List<StreamingPullRequest> requests = new ArrayList<>();
+    for (StreamingPullRequest req :
+        StreamingSubscriberConnection.partitionAckOperations(
+            acksToSend, ackDeadlineExtensions, size)) {
```
pongad: Originally thought array indexing would look more natural. Done.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnectionTest.java
pongad: Done.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnectionTest.java
pongad: Changed to w, x, y, z for symmetry

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StreamingSubscriberConnection.java
```diff
@@ -211,35 +211,54 @@ private boolean isAlive() {
...
-          }
+  @VisibleForTesting
+  static StreamingPullRequest[] partitionAckOperations(
+      List<String> acksToSend, List<PendingModifyAckDeadline> ackDeadlineExtensions, int size) {
+    int numExtensions = 0;
+    for (PendingModifyAckDeadline modify : ackDeadlineExtensions) {
+      numExtensions += modify.ackIds.size();
+    }
+    int numRequests = Math.max(numExtensions, acksToSend.size());
+    numRequests = numRequests / size + (numRequests % size == 0 ? 0 : 1);
```
pongad: Done

garrettjonesgoogle:
LGTM

Issue #1545, opened 2017-01-18
Storage Retry Logic


Dima1224:
Hi there. I'm using `com.google.cloud.storage.Storage.writer()` method of v0.8.0 of the google-cloud lib. I'm getting "503 backend error" from the GCS service when trying to upload files (stack trace below). I expect the retry logic in `com.google.cloud.storage.BlobWriteChannel` to retry the chunk, but it doesn't do that.

```java
  @Override
  protected void flushBuffer(final int length, final boolean last) {
    try {
      runWithRetries(callable(new Runnable() {
        @Override
        public void run() {
          getOptions().getRpc().write(getUploadId(), getBuffer(), 0, getPosition(), length, last);
        }
      }), getOptions().getRetryParams(), StorageImpl.EXCEPTION_HANDLER, getOptions().getClock());
    } catch (RetryHelper.RetryHelperException e) {
      throw StorageException.translateAndThrow(e);
    }
  }
```
The problem seems to lie in the logic for converting the `com.google.api.client.http.HttpResponseException` to a `com.google.cloud.storage.StorageException`

```java
public BaseServiceException(IOException exception, boolean idempotent) {
    super(message(exception), exception);
    int code = UNKNOWN_CODE;
    String reason = null;
    String location = null;
    String debugInfo = null;
    Boolean retryable = null;
    if (exception instanceof GoogleJsonResponseException) {
      GoogleJsonError jsonError = ((GoogleJsonResponseException) exception).getDetails();
      if (jsonError != null) {
        Error error = new Error(jsonError.getCode(), reason(jsonError));
        code = error.code;
        reason = error.reason;
        retryable = isRetryable(idempotent, error);
        if (reason != null) {
          GoogleJsonError.ErrorInfo errorInfo = jsonError.getErrors().get(0);
          location = errorInfo.getLocation();
          debugInfo = (String) errorInfo.get("debugInfo");
        }
      } else {
        code = ((GoogleJsonResponseException) exception).getStatusCode();
      }
    }
    this.retryable = MoreObjects.firstNonNull(retryable, isRetryable(idempotent, exception));
    this.code = code;
    this.reason = reason;
    this.idempotent = idempotent;
    this.location = location;
    this.debugInfo = debugInfo;
  }
```

Since the HttpResponseException is not an instanceof GoogleJsonResponseException, we end up with a code of 0, which the retry logic sees as a non-retryable exception. Here's a code snippet to easily reproduce the issue:

```java
[...]

PR #1677, opened 2017-03-01
fix retry logic on HttpResponseException (#1545)

neozwu:
In google-cloud-bigquery and google-cloud-storage, some API calls ("open" and "write" in BigQuery and Storage) are fired by using manually constructed http requests. When errors occur, these requests only throw `HttpResponseException` (i.e. not `GoogleJsonResponseException`). The current logic in its superclass, BaseServiceException, cannot parse out the "retryability" from `HttpResponseException`. Therefore, certain retryable errors (such as 503) are deemed non-retrybale. This PR fixes this issue by checking the status code in HttpResponseException in order to determine whether an HttpResponseException is retrybale or not.

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryExceptionTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryExceptionTest.java
@@ -16,17 +16,23 @@
 
 package com.google.cloud.bigquery;
 
+import static org.easymock.EasyMock.createMock;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.replay;
+import static org.easymock.EasyMock.verify;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
 import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.http.HttpResponseException;
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryHelper.RetryHelperException;
-import org.junit.Test;
-
 import java.io.IOException;
 import java.net.SocketTimeoutException;
-
-import static org.easymock.EasyMock.*;
-import static org.junit.Assert.*;
+import org.junit.Test;
 
 public class BigQueryExceptionTest {
 
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageExceptionTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageExceptionTest.java
@@ -16,18 +16,24 @@
 
 package com.google.cloud.storage;
 
+import static org.easymock.EasyMock.createMock;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.replay;
+import static org.easymock.EasyMock.verify;
+import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


garrettjonesgoogle:
@vam-google could you do the first pass on this PR?

google-cloud-core/src/main/java/com/google/cloud/BaseServiceException.java
```diff
@@ -140,20 +141,36 @@ public BaseServiceException(IOException exception, boolean idempotent) {
...
       } else {
-        code = ((GoogleJsonResponseException) exception).getStatusCode();
+        // In cases where an exception is an instance of HttpResponseException but not
+        // an instance of GoogleJsonResponseException, check the status code to determine whether it's retryable
+        code = ((HttpResponseException) exception).getStatusCode();
+        Set<Error> retryableErrors = getRetryableErrors();
+        if (retryableErrors != null && !retryableErrors.isEmpty()) {
+          for (Error e : retryableErrors) {
+            if (e.getCode() != null && e.getCode().equals(code)) {
+              Error error = new Error(code, null);
```
vam-google: The body of the loop does not depend on the loop itself (the "retriable" is assigned based on value of "code" and "idempotent", both of which are known outside of the loop), potentially a bug. 

Looks like the tests pass, because the actual "isRetryable(boolean idempotent, Error error)" call already iterates over the getRetriableErrors() in its implementation. So, probably the code can be simplified to the single direct call to that method:
retriable = isRetriable(idempotent, new Error(((HttpResponseException) exception).getStatusCode(), null));

neozwu: >The body of the loop does not depend on the loop itself (the "retriable" is assigned based on value of "code" and "idempotent", both of which are known outside of the loop), potentially a bug.

Maybe I didn't understand it, but I thought the body of the loop depends on the loop, because `retryable` depends on the `if` condition, which in turn depends on the loop variable.

>Looks like the tests pass, because the actual "isRetryable(boolean idempotent, Error error)" call already iterates over the getRetriableErrors() in its implementation. So, probably the code can be simplified to the single direct call to that method:
retriable = isRetriable(idempotent, new Error(((HttpResponseException) exception).getStatusCode(), null));

I did consider the one-single-call approach. But my original thought was to not convert any `HttpResponseException` to `Error` without knowing it's actually an `Error` - after all, exceptions and errors are different.  But now I look at it again, I think you are right - the `Error` class is generic enough, and the repeated iteration is indeed a code smell. I'll rewrite it using a direct call.

vam-google: Yes, sorry, I missed the fact that "not assigning at all" is another kind of result. But the isRetriable() still has same type of the loop inside, so looping twice over same set of data looks suspicious. If you  can rewrite it without looping here again, it would be great.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryExceptionTest.java
```diff
@@ -16,24 +16,18 @@
...
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryHelper.RetryHelperException;
-
 import org.junit.Test;
 
 import java.io.IOException;
 import java.net.SocketTimeoutException;
 
+import static org.easymock.EasyMock.*;
+import static org.junit.Assert.*;
```
garrettjonesgoogle: Don't use wildcard imports.

garrettjonesgoogle:
LGTM - @vam-google could you also approve?

PR #1678, opened 2017-03-01
Remove deprecated methods and corresponding tests

shinfan:
Update:

Fixed all malformed javadocs including:
- Deprecated method references
- Unqualified class names
- Typos
 

_Note: I intend to merge this PR soon since it may cause lots of conflicts._

Remove deprecated methods and tests in one pass by a self-made script, with a few exceptions:
- There are still plenty of manual work involved to update usages of deprecated methods and fields
- Manual corrections due to misspell of the word "deprecated"
- google-cloud-pubsub: Pubsub has a few deprecated packages and it affects its integrated tests as well. Removing deprecated packages is not in the scope of this PR.
- google-cloud-compute: Compute engine has a few "Deprecated" related service which causes lots of noise of the automated script. 
- google-cloud-datastore
  - In `ValueBuilder` `getExcludeFromIndexes` was marked as deprecated which is inconsistent with other getter/setter methods
  - Preserved the fields `meaning` in `ValueBuilder` since it is still wide used in the code. May require a separate PR to deprecate.

Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1569

Test: All unit tests passed locally.

PR #1687, opened 2017-03-03
Updates for flowcontrol

michaelbausor:
Updates to Pubsub and Logging to support integration of FlowControl into Bundling.

Note that these are changes in the Pubsub surface, in particular in Publisher.java


garrettjonesgoogle:
I think this improves the surface.

pongad:
The surface change LGTM. This seems to be failing on Travis though.

pom.xml
```diff
@@ -92,7 +92,7 @@
     <github.global.server>github</github.global.server>
     <google.auth.version>0.6.0</google.auth.version>
     <grpc.version>1.0.3</grpc.version>
-    <gax.version>0.1.3</gax.version>
+    <gax.version>0.1.5</gax.version>
```
garrettjonesgoogle: Since the bundling stuff is breaking, could we release the next gax at 0.2.0 instead?

michaelbausor:
Updated with generated surface refresh, PTAL. Note that this will continue to fail CI because the relevant GAX version has not been released. Once we have stabilized the changes to GAX, we can merge and release the new version, and then rerun CI here.

pom.xml
michaelbausor: Yes will do

garrettjonesgoogle:
LGTM, but let's wait until https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1678 is merged before merging this one.

shinfan:
#1678 is merged!

PR #1688, opened 2017-03-03
Bump the GAX version

michaelbausor:


garrettjonesgoogle:
Fix the version of gax that spanner uses while you're at it: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/pom.xml#L132 (use the variable instead of a fixed version here)

garrettjonesgoogle:
LGTM

michaelbausor:
Note: merging before Travis is complete because this is a low-risk change, and has passed on 3 platforms.

PR #1690, opened 2017-03-03
GAX rename: RpcFuture -> ApiFuture

garrettjonesgoogle:


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/deprecated/Message.java
```diff
@@ -75,7 +75,7 @@ public PubsubMessage apply(Message message) {
     }
 
     @Override
-    protected ByteString getByteString() {
+    public ByteString getByteString() {
```
vam-google: Is it related to the RpcFuture -> ApiFuture rename? I is not related to the change, but it seems strange that the class Message under the "deprecated" directory is not declared as @Deprecated .

garrettjonesgoogle: Oh this was a mistake, I didn't mean to include this change.

garrettjonesgoogle:
Feedback addressed, PTAL

garrettjonesgoogle:
As for the codacy-bot comments, it is wrong in this case. We specifically want println statements because it is just example code.

PR #1701, opened 2017-03-06
Vision moved to beta stage

vam-google:


google-cloud/pom.xml
```diff
@@ -92,7 +92,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>google-cloud-vision</artifactId>
-      <version>${project.version}</version>
```
garrettjonesgoogle: Leave the overall project at alpha.

vam-google: I was making it consistent with the other beta packages in the repo. For example for logging same file has:
    <dependency>
      <groupId>${project.groupId}</groupId>
      <artifactId>google-cloud-logging</artifactId>
      <version>${beta.version}</version>
    </dependency>

Do we really want to have it alpha here for vision?

garrettjonesgoogle: My mistake, I didn't look at the context to see that this was for vision. This is correct.

garrettjonesgoogle:
LGTM

PR #1708, opened 2017-03-08
Regenerate logging with updated class names

michaelbausor:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1047


shinfan:
LGTM after at lease one of the tests pass

PR #1709, opened 2017-03-08
Pub/Sub: Converting Future -> ApiFuture

garrettjonesgoogle:
Also, deleting PullFuture which is subsumed by ApiFuture.

### DIFF
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/deprecated/PubSubImplTest.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/deprecated/PubSubImplTest.java
@@ -1381,7 +1381,6 @@ public class PubSubImplTest {
     }
   }
 
-
   @Test
   public void testMessageConsumerWithOptions() throws Exception {
     pubsub = new PubSubImpl(options, renewerMock);


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/deprecated/spi/PubSubRpc.java
```diff
@@ -40,189 +41,149 @@
...
-     * @param error the execption that caused the {@link PullFuture} to fail
-     */
-    void failure(Throwable error);
-  }
-
-  /**
-   * A {@link Future} implementation for pull RPCs. This class also allows users to register
-   * callbacks via {@link #addCallback(PullCallback)}.
-   */
-  interface PullFuture extends Future<PullResponse> {
```
pongad: I'm a little concerned by this. This is definitely a breaking change, even if the fix is pretty obvious. If you think the cost of maintaining it outweighs the cost of the break, I'm good with this.

garrettjonesgoogle: This is an "internal" component of Pub/Sub - it's not exposed on the `PubSub` interface, so I'm pretty sure users aren't expected to interact with it.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/deprecated/PubSubImplTest.java
```diff
@@ -1407,6 +1381,7 @@ public PullFuture answer() throws Throwable {
     }
   }
 
+
```
michaelbausor: Nit: remove newline

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/deprecated/PubSubOptions.java
```diff
@@ -132,11 +132,6 @@ public Builder toBuilder() {
     return new Builder(this);
   }
 
-  @Deprecated
```
garrettjonesgoogle: I was getting an error in IntelliJ - it looks like it was just overlooked.

PR #1711, opened 2017-03-08
Spanner test fix

vkedia:
Fixed ITDatabaseAdminTest to only drop the databases it created rather than dropping all the databases in the instance.


garrettjonesgoogle:
LGTM. Be sure to use "squash and merge".

PR #1714, opened 2017-03-08
Update to README.md

danoscarmike:
Add Cloud Vision to Beta supported list


danoscarmike:
Thanks @vam-google.  Understood.  I don't know why it wasn't listed... I note that Speech and NL are also _not_ there.  @garrettjonesgoogle do you know if there is a reason why?

PR #1716, opened 2017-03-08
Add async/sync setting to logging handler

michaelbausor:
Two questions:
- I don't like WriteLogMethod much as a setting name. WriteSyncType? Others?
- I am using a switch in write(). Any preference to instead set a callable at construction time? Or something else?

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -53,44 +53,42 @@ import java.util.logging.SimpleFormatter;
  * <tr><td>FINEST</td><td>DEBUG</td></tr>
  * </table>
  *
- * <p>Original Java logging levels are added as labels (with {@code levelName} and {@code
- * levelValue} keys, respectively) to the corresponding Stackdriver Logging {@link LogEntry}. You
- * can read entry labels using {@link LogEntry#getLabels()}. To use logging levels that correspond
- * to Stackdriver Logging severities you can use {@link LoggingLevel}.
+ * <p>Original Java logging levels are added as labels (with {@code levelName} and
+ * {@code levelValue} keys, respectively) to the corresponding Stackdriver Logging {@link LogEntry}.
+ * You can read entry labels using {@link LogEntry#getLabels()}. To use logging levels that
+ * correspond to Stackdriver Logging severities you can use {@link LoggingLevel}.
  *
  * <p><b>Configuration</b>: By default each {@code LoggingHandler} is initialized using the
- * following {@code LogManager} configuration properties (that you can set in the {@code
- * logging.properties} file). If properties are not defined (or have invalid values) then the
+ * following {@code LogManager} configuration properties (that you can set in the
+ * {@code logging.properties} file). If properties are not defined (or have invalid values) then the
  * specified default values are used.
- *
  * <ul>
- *   <li>{@code com.google.cloud.logging.LoggingHandler.log} the log name (defaults to {@code
- *       java.log}).
- *   <li>{@code com.google.cloud.logging.LoggingHandler.level} specifies the default level for the
- *       handler (defaults to {@code Level.INFO}).
- *   <li>{@code com.google.cloud.logging.LoggingHandler.filter} specifies the name of a {@link
- *       Filter} class to use (defaults to no filter).
- *   <li>{@code com.google.cloud.logging.LoggingHandler.formatter} specifies the name of a {@link
- *       Formatter} class to use (defaults to {@link SimpleFormatter}).
- *   <li>{@code com.google.cloud.logging.LoggingHandler.flushSize} specifies the maximum size of the
- *       log buffer. Once reached, logs are transmitted to the Stackdriver Logging service (defaults
- *       to 1).
- *   <li>{@code com.google.cloud.logging.LoggingHandler.flushLevel} specifies the flush log level.
- *       When a log with this level is published, logs are transmitted to the Stackdriver Logging
- *       service (defaults to {@link LoggingLevel#ERROR}).
- *   <li>{@code com.google.cloud.logging.LoggingHandler.enhancers} specifies a comma separated list
- *       of {@link Enhancer} classes. This handler will call each enhancer list whenever it builds a
- *       {@link MonitoredResource} or {@link LogEntry} instance (defaults to empty list).
- *   <li>{@code com.google.cloud.logging.LoggingHandler.resourceType} the type name to use when
- *       creating the default {@link MonitoredResource} (defaults to "global").
- *   <li>{@code com.google.cloud.logging.Synchronicity} the synchronicity of the write method to use
- *       to write logs to the Stackdriver Logging service (defaults to {@link Synchronicity#ASYNC}).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.log} the log name (defaults to
+ *     {@code java.log}).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.level} specifies the default level for the
+ *     handler (defaults to {@code Level.INFO}).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.filter} specifies the name of a {@link Filter}
+ *     class to use (defaults to no filter).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.formatter} specifies the name of a
+ *     {@link Formatter} class to use (defaults to {@link SimpleFormatter}).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.flushSize} specifies the maximum size of the
+ *     log buffer. Once reached, logs are transmitted to the Stackdriver Logging service (defaults
+ *     to 1).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.flushLevel} specifies the flush log level.
+ *     When a log with this level is published, logs are transmitted to the Stackdriver Logging
+ *     service (defaults to {@link LoggingLevel#ERROR}).
+ * <li>{@code com.google.cloud.logging.LoggingHandler.enhancers} specifies a comma separated list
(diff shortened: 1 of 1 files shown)


michaelbausor:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1532

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -497,14 +499,21 @@ public void testLoggingHandler() throws InterruptedException {
...
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
     LoggingOptions options = logging().getOptions();
     MonitoredResource resource = MonitoredResource.of("gce_instance",
         ImmutableMap.of("project_id", options.getProjectId(),
             "instance_id", "instance",
             "zone", "us-central1-a"));
-    LoggingHandler handler = new AsyncLoggingHandler(logName, options, resource);
+    LoggingHandler handler = new LoggingHandler(logName, options, resource);
```
garrettjonesgoogle: Is there a programmatic way to control Sync vs Async?

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -497,14 +499,21 @@ public void testLoggingHandler() throws InterruptedException {
...
-  public void testAsyncLoggingHandler() throws InterruptedException {
-    String logName = formatForTest("test-async-logging-handler");
+  public void testSyncLoggingHandler() throws InterruptedException {
+    String logName = formatForTest("test-sync-logging-handler");
+    try {
+      LogManager.getLogManager().readConfiguration(
+          new ByteArrayInputStream(
+              "com.google.cloud.logging.LoggingHandler.writeLogMethod=SYNC".getBytes()));
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
garrettjonesgoogle: If you throw an exception, throw `LoggingException`

google-cloud-logging/src/main/java/com/google/cloud/logging/WriteLogMethod.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+public enum WriteLogMethod {
```
garrettjonesgoogle: I think `Synchronicity` would capture the concept.

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
michaelbausor: Removed these

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
michaelbausor: Added

google-cloud-logging/src/main/java/com/google/cloud/logging/WriteLogMethod.java
michaelbausor: Renamed

michaelbausor:
PTAL. Also, should we have tests to make sure that settings in the properties file are read correctly?

garrettjonesgoogle:
Yes, I think we should have a test for the properties file settings.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -296,6 +300,16 @@ Formatter getFormatterProperty(String name, Formatter defaultValue) {
       }
       return Collections.emptyList();
     }
+
+    Synchronicity getSynchronicityProperty(String name, Synchronicity defaultValue) {
```
garrettjonesgoogle: make private

google-cloud-logging/src/main/java/com/google/cloud/logging/Synchronicity.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+public enum Synchronicity {
```
garrettjonesgoogle: Add a doc

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -35,6 +33,7 @@
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 import java.util.logging.SimpleFormatter;
+import com.google.cloud.logging.Logging.WriteOption;
```
pongad: Import isn't sorted, trying running `git diff -U0 master | google-java-format-diff.py -p1 -i`? The python script is [here](https://github.com/google/google-java-format/tree/master/scripts).

google-cloud-logging/src/main/java/com/google/cloud/logging/Synchronicity.java
pongad: I might be in the minority here. I think `Sychronicity` is too specific: it only has a yes-or-no answer. We might as well use a boolean. If we think we might add more enum members later on, I think the name `WriteMode` is more appropriate.

garrettjonesgoogle: It's a lot more readable with a more specific name, though. `WriteMode` doesn't really match the intent here - it's more like `WaitMode`.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
michaelbausor: Thanks for this! I had been avoiding using the formatter because it touches a lot of the comments in LoggingHandler. I have applied it now - if you want I can revert the changes to comments.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
michaelbausor: Done. Should I also make similar methods such as `getEnhancerProperty` private?

google-cloud-logging/src/main/java/com/google/cloud/logging/Synchronicity.java
michaelbausor: Done

michaelbausor: I don't mind `WriteMode` or `WaitMode`, although `WriteMode` seems more intuitive to me. I prefer the enum to a boolean.

pongad: I agree that the enum is better than boolean, I just think another name will give us more forward flexibility.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
garrettjonesgoogle: Yeah let's revert the comment changes.

BTW in IntelliJ when you select code and use the formatter, it will only format that selection.

google-cloud-logging/src/main/java/com/google/cloud/logging/Synchronicity.java
garrettjonesgoogle: We can't predict what forward flexibility we will need right now. I'd prefer to not be too general with a concept we can't predict.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
garrettjonesgoogle: Oh actually I didn't notice that this was in the context of a private class. You can just make this method (`getSynchronicityProperty`) package-private to be consistent with the rest.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
michaelbausor: Done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
michaelbausor: Done

google-cloud-logging/src/main/java/com/google/cloud/logging/Synchronicity.java
michaelbausor: I have left it as `Synchronicity`.

michaelbausor:
PTAL

google-cloud-logging/src/main/java/com/google/cloud/logging/Synchronicity.java
garrettjonesgoogle: I feel pretty strongly on this one: I have seen over-extended enumerations cause quite a bit of pain. In particular, when you end up adding values which specify an independent semantic dimension from the original dimension, you end up with a combinatorial cluster-f*. If we need new options, I think we should create a new enum & setting, unless the new option fits neatly within the "synchronicity" concept.

garrettjonesgoogle:
LGTM

PR #1717, opened 2017-03-09
Add project role identities to the Identity helper

rybosome:
In addition to the standard IAM identities, project membership identities are
also supported. For instance, a role can be applied to only the Editors of a
project.


garrettjonesgoogle:
looks good, I'll merge after >=1 tests pass.

PR #1721, opened 2017-03-09
logging: reduce message size

### LINKED ISSUE #1712: Serialized size is too large
The following `LogEntry` closely resembles ones `LoggingHandler` produces.
```java
LogEntry.newBuilder()
    .setSeverity(LogSeverity.INFO)
    .setTextPayload("test payload")
    .setTimestamp(
        Timestamp.newBuilder()
            .setSeconds(nanoTime / BILLION)
            .setNanos((int) (nanoTime % BILLION)))
    .putLabels("levelName", Level.INFO.getName())
    .putLabels("levelValue", "" + Level.INFO.intValue())
    .build();
```
Its serialized size is 67 bytes.

However, if we have many entries batched in the same request with the same severity, we can "hoist" the two labels out of the entries and into the request object. (Entries with different severities can still override.) If we comment out the two `putLabels` lines, the serialized size reduces to 28, saving 39 bytes.

**Proposal**
Set `LoggingHandler` with a "native severity". The severity labels would be placed in the request object, and all logs with the same severity would not contain the label.

**Alternative**
Drop the labels altogether. After all, each log already has `LogSeverity` which is defined by the protobuf message itself. This is a breaking change though.

**Memory Impact**
With labels, building one `LogEntry` object allocates 47 objects and 1584 bytes (measured on `google-cloud:0.9.4-alpha`). Without labels, we allocate 31 objects and 1040 bytes. Fixing this should give a 33% decrease in memory consumption in small messages.

pongad:
This commit is based on top of https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1720 . Once  we merge that, I'll rebase to make this PR easier to review.

This PR fixes #1712 , using the method suggested in the issue.

To un-WIP this PR (in rough dependency-order):
- [ ] 1. ~~Think of a better name for "native level"~~
- [ ] 2. ~~Native level and setLevel works from the same config. When we know what to call the former, give it its own config.~~
- [x] 3. The config is too difficult to test, because it's too difficult to pass value in from test.
Maybe we need a config class? It can be pre-populated with the property values but let users override.
- [x] 4. Native level should be documented

@garrettjonesgoogle @michaelbausor What do you think of 1 and 3?

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -110,9 +110,14 @@ public class LoggingHandler extends Handler {
   private volatile Logging logging;
   private Level flushLevel;
   private long flushSize;
-  private final Level baseLevel;
   private Synchronicity synchronicity;
   private final List<Enhancer> enhancers;
+
+  // Logs with the same severity with the base could be more efficiently sent to Stackdriver.
+  // Defaults to level of the handler or Level.FINEST if the handler is set to Level.ALL.
+  // Currently there is no way to modify the base level, see
+  // https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1740 .
+  private final Level baseLevel;
 
   /**
    * Creates an handler that publishes messages to Stackdriver Logging.


michaelbausor:
Overall this looks good to me. Re your points:
1. Naming is hard :-( I don't mind native level. Or perhaps request level?
2. Does this need its own config? It is strictly for performance optimization, so having it match the Level setting makes sense since this is the most granular level of log that will be processed by the logger. At the least we should discourage users from setting it manually unless they really do know that they generate more logs at a higher level.
3. In https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1716 I added getters for flush level and size, matching the Handler parent class for its settings. So you could add a getter and setter for native level. I think that pulling these into a config class makes sense in general, but we will still be left with the getters and setters on the Handler parent class.

garrettjonesgoogle:
I think "native level" is hard to figure out. "request level" is better, but still potentially confusing. How about "default level"?

michaelbausor:
I think the problem with "default level" is that it is too similar to "level", which is a property of the Handler object. Maybe "default request level", but it still sounds like this setting has semantic meaning, rather than being just a performance optimization.

garrettjonesgoogle:
maybe "common request level".

pongad:
Answers inline

> Maybe "default request level", but it still sounds like this setting has semantic meaning, rather than being just a performance optimization

Maybe we should just call it `optimizeForLevel`? I think "common request level" still implies some kind of default.

> Does this need its own config?

You made a good point that since it's purely optimization, it might not need one. We could document the `.level` config that logs at that severity might be sent more efficiently. Maybe open a tracking issue where users can comment if they need this as a config.

> So you could add a getter and setter for native level

Doing this raises a question I don't have a good answer to yet.
Changing label changes the partition key used by bundling. Currently bundling creates a new thread for each partition key. So changing label midway feels a bit like a time bomb waiting to go off. I think we should add getter/setter, but only after we make bundling a little more resilient. (https://github.com/googleapis/gax-java/pull/194 should help some.)

I think we should do is:
**For this PR**
- Finish item 4, by setting native level to be the same as `.level` config
- Finish item 3, by reusing @michaelbausor 's change in #1716
- This should automatically help some of our users, and our users don't need to do anything

**Later**
- Harden bundling
- Now that we can change the native level midway, it'd be easier to do items 1 and 2
- This lets advanced users tweak the settings to fit their app just right

@michaelbausor @garrettjonesgoogle WDYT?

michaelbausor:
SGTM. Even once we _can_ change do 1 and 2, we may not want to give users that knob until we are sure they want it. I think the tracking issue you suggest is a good idea.

Re googleapis/gax-java#194 I am working on that change - will close out that PR and open a new one once it is ready for review.

pongad:
@garrettjonesgoogle @michaelbausor PTAL

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -67,6 +68,12 @@
  *     {@code java.log}).
  * <li>{@code com.google.cloud.logging.LoggingHandler.level} specifies the default level for the
  *     handler (defaults to {@code Level.INFO}).
+ *     <p> This configuration also sets the "base severity level".
+ *     Logs with the same severity with the base could be more efficiently sent to Stackdriver.
+ *     The base severity defaults to the level of of handler or {@code Level.FINEST}
```
garrettjonesgoogle: "level of of handler" -> "level of the handler"

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -67,6 +68,12 @@
  *     {@code java.log}).
  * <li>{@code com.google.cloud.logging.LoggingHandler.level} specifies the default level for the
  *     handler (defaults to {@code Level.INFO}).
+ *     <p> This configuration also sets the "base severity level".
```
garrettjonesgoogle: Do we really need to publicly document this implementation detail?

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
```diff
@@ -149,8 +147,15 @@
           .put("resourceType", "testResourceType")
           .put("synchronicity", "SYNC")
           .build();
+  private static final ImmutableMap<String, String> NATIVE_SEVERITY_MAP =
```
garrettjonesgoogle: -> `BASE_SEVERITY_MAP`

michaelbausor:
LGTM

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
pongad: done

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
pongad: Done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
pongad: I agree this might be a lot of details most users won't care about. On the other hand, I felt it might be useful to document, since it represents a sizable percentage change in RPC efficiency. Of course, most users will probably not log enough to see a big difference in absolute terms.

I'll be happy to delete this if you feel strongly about it.

pongad:
@garrettjonesgoogle I replied to your comment. Wanted to make sure it's good with you before merging. PTAL.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
garrettjonesgoogle: I feel like the information is useful, but at the user level it is a distraction since it can't be controlled. Could we move it somewhere where a maintainer would find it?

garrettjonesgoogle:
LGTM

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
garrettjonesgoogle: I'd prefer this to be taken out of the public javadoc, but I won't block you from submitting it as it is if you feel strongly about keeping it here.

pongad: Removed

garrettjonesgoogle:
ship it

pongad:
I'll hold until CI pass just to make sure :)

PR #1723, opened 2017-03-09
Removing transport from ServiceOptions hierarchy

### LINKED ISSUE #1601: Change service hierarchy so that http vs grpc isn't baked in
Specifically, look at the setup required to initialize the service object.

garrettjonesgoogle:
Updates #1601

This allows a client to support multiple transports in the future,
e.g. Datastore supports http now but grpc support will be added
later.

Related changes:

* Renaming HttpServiceOptions to HttpTransportOptions
* Renaming GrpcServiceOptions to GrpcTransportOptions
* Creating new TransportOptions interface

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
@@ -108,10 +108,6 @@ public class BigQueryOptions extends ServiceOptions<BigQuery, BigQueryRpc, BigQu
     return HttpTransportOptions.newBuilder().build();
   }
 
-  public HttpTransportOptions getHttpTransportOptions() {
-    return (HttpTransportOptions) getTransportOptions();
-  }
-
   @Override
   protected Set<String> getScopes() {
     return SCOPES;
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/DefaultBigQueryRpc.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/DefaultBigQueryRpc.java
@@ -80,7 +80,7 @@ public class DefaultBigQueryRpc implements BigQueryRpc {
   private final Bigquery bigquery;
 
   public DefaultBigQueryRpc(BigQueryOptions options) {
-    HttpTransportOptions transportOptions = options.getHttpTransportOptions();
+    HttpTransportOptions transportOptions = (HttpTransportOptions) options.getTransportOptions();
     HttpTransport transport = transportOptions.getHttpTransportFactory().create();
     HttpRequestInitializer initializer = transportOptions.getHttpRequestInitializer(options);
     this.options = options;
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/testing/RemoteBigQueryHelperTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/testing/RemoteBigQueryHelperTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.bigquery.testing;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import com.google.cloud.HttpTransportOptions;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQuery.DatasetDeleteOption;
 import com.google.cloud.bigquery.BigQueryOptions;
@@ -80,8 +81,10 @@ public class RemoteBigQueryHelperTest {
     RemoteBigQueryHelper helper = RemoteBigQueryHelper.create(PROJECT_ID, JSON_KEY_STREAM);
     BigQueryOptions options = helper.getOptions();
     assertEquals(PROJECT_ID, options.getProjectId());
-    assertEquals(60000, options.getHttpTransportOptions().getConnectTimeout());
-    assertEquals(60000, options.getHttpTransportOptions().getReadTimeout());
+    assertEquals(60000,
+        ((HttpTransportOptions) options.getTransportOptions()).getConnectTimeout());
+    assertEquals(60000,
+        ((HttpTransportOptions) options.getTransportOptions()).getReadTimeout());
     assertEquals(10, options.getRetryParams().getRetryMaxAttempts());
     assertEquals(6, options.getRetryParams().getRetryMinAttempts());
     assertEquals(30000, options.getRetryParams().getMaxRetryDelayMillis());
--- a/google-cloud-compute/src/main/java/com/google/cloud/compute/ComputeOptions.java
+++ b/google-cloud-compute/src/main/java/com/google/cloud/compute/ComputeOptions.java
@@ -108,10 +108,6 @@ public class ComputeOptions extends ServiceOptions<Compute, ComputeRpc, ComputeO
     return HttpTransportOptions.newBuilder().build();
   }
 
-  public HttpTransportOptions getHttpTransportOptions() {
-    return (HttpTransportOptions) getTransportOptions();
-  }
-
   @Override
   protected Set<String> getScopes() {
     return SCOPES;
--- a/google-cloud-compute/src/main/java/com/google/cloud/compute/spi/DefaultComputeRpc.java
+++ b/google-cloud-compute/src/main/java/com/google/cloud/compute/spi/DefaultComputeRpc.java
@@ -81,7 +81,7 @@ public class DefaultComputeRpc implements ComputeRpc {
   private final Compute compute;
 
   public DefaultComputeRpc(ComputeOptions options) {
-    HttpTransportOptions transportOptions = options.getHttpTransportOptions();
+    HttpTransportOptions transportOptions = (HttpTransportOptions) options.getTransportOptions();
     HttpTransport transport = transportOptions.getHttpTransportFactory().create();
     HttpRequestInitializer initializer = transportOptions.getHttpRequestInitializer(options);
     this.options = options;
--- a/google-cloud-compute/src/test/java/com/google/cloud/compute/testing/RemoteComputeHelperTest.java
+++ b/google-cloud-compute/src/test/java/com/google/cloud/compute/testing/RemoteComputeHelperTest.java
@@ -21,6 +21,7 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 
+import com.google.cloud.HttpTransportOptions;
 import com.google.cloud.compute.ComputeOptions;
 import com.google.cloud.compute.testing.RemoteComputeHelper.ComputeHelperException;
 import java.io.ByteArrayInputStream;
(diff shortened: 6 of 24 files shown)


google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -241,6 +254,7 @@ protected ServiceOptions(Class<? extends ServiceFactory<ServiceT, OptionsT>> ser
         getFromServiceLoader(rpcFactoryClass, getDefaultRpcFactory()));
     serviceRpcFactoryClassName = serviceRpcFactory.getClass().getName();
     clock = firstNonNull(builder.clock, Clock.defaultClock());
+    transportOptions = firstNonNull(builder.transportOptions, getDefaultTransportOptions());
```
vam-google: Minor (as it is a much older design decision, not related to your changes). I believe it is better to avoid calling non-private methods from constructors, as it is a bad practice, and potentially leads to premature usage of non-initialized subclasses  (for example http://stackoverflow.com/questions/15327417/is-it-ok-to-call-abstract-method-from-constructor-in-java).

In this case passing the default transport options and default rpc factory as ServiceOptions constructor parameters can fix the problem.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -61,6 +64,15 @@ private Builder(BigQueryOptions options) {
     }
 
     @Override
+    public Builder setTransportOptions(TransportOptions transportOptions) {
```
vam-google: There are 7 exact copies of this method, all for ServiceNameOptions classes, all extending ServiceOptions. Since it is already an extension (not implementation), please consider adding HttpServiceOptions class in the hierarcy (or keeping existing one, so HttpTransportOptions would be a completely new class, not a renamed one). setTransportOptions and getTransportOptions are potential candidates to be moved into that class.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -81,6 +93,19 @@ protected BigQueryRpcFactory getDefaultRpcFactory() {
...
   @Override
+  public TransportOptions getDefaultTransportOptions() {
+    return getDefaultHttpTransportOptions();
+  }
+
+  public static HttpTransportOptions getDefaultHttpTransportOptions() {
+    return HttpTransportOptions.newBuilder().build();
+  }
+
+  public HttpTransportOptions getHttpTransportOptions() {
```
vam-google: Looks suspicious and redundant, as there is already the getTransportOptions method (having two methods doing essentially same thing is potentially confusing for a user).

Please consider overriding getTransportOptions() here and returning HttpTransportOptions from it directly (also, similar to the previous comment, it can be moved to the HttpServiceOptions base class):

@Override
public /*works as of java spec 8.4.5.*/ HttpTransportOptions getTransportOptions() {
    return (HttpTransportOptions) super.getTransportOptions();
}

google-cloud-core/src/main/java/com/google/cloud/GrpcTransportOptions.java
```diff
@@ -281,22 +261,34 @@ public int getMaxTimeout() {
...
   }
 
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
   @Override
-  protected int baseHashCode() {
-    return Objects.hash(super.baseHashCode(), executorFactoryClassName, initialTimeout,
+  public int hashCode() {
```
garrettjonesgoogle: It has to do with the inheritance setup in `google-cloud-java`. The pattern is to use a `baseHashCode` function in abstract classes and `hashCode` in concrete classes (which call the `baseHashCode` of the parent class). Since this class is converting from abstract to concrete, that's why you are seeing this change.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
michaelbausor: Another possibility is to make `setTransportOptions` protected in `ServiceOptions` and make the signature of this method `public Builder setTransportOptions(HttpTransportOptions transportOptions)`. This can be changed to accept `TransportOptions` in future without breaking.

This doesn't work if someone holding a `ServiceOptions<?, ?, ?>` object needs to set the transport.

google-cloud-core/src/main/java/com/google/cloud/GrpcTransportOptions.java
```diff
@@ -281,22 +261,34 @@ public int getMaxTimeout() {
...
+    return Objects.hash(executorFactoryClassName, initialTimeout,
         timeoutMultiplier, maxTimeout);
   }
 
-  protected boolean baseEquals(GrpcServiceOptions<?, ?, ?> other) {
-    return super.baseEquals(other)
-        && Objects.equals(executorFactoryClassName, other.executorFactoryClassName)
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof GrpcTransportOptions)) {
```
garrettjonesgoogle: Ahh, you're right, I'll switch it. 

I think the implicit guideline is to use a `baseEquals` function for abstract classes, which assumes that more checks will be done in derived classes. According to http://stackoverflow.com/questions/596462/any-reason-to-prefer-getclass-over-instanceof-when-generating-equals , the leaf classes should do exact class comparison in `equals()`.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
garrettjonesgoogle: Oh you're totally right - I knew this but somehow overlooked it. It seems worth it to change now.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
garrettjonesgoogle: 1. That defeats the entire purpose of this change
2. In the future, Datastore will support both http & grpc, so its implementation of this function would not actually reject either of those transport option types.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
garrettjonesgoogle: Ahh of course, use a covariant return type. I'll do that.

(leaving the discussion of HttpServiceOptions to the comment thread above)

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/RemoteComputeHelper.java
```diff
@@ -103,10 +107,12 @@ public static RemoteComputeHelper create(String projectId, InputStream keyStream
    * credentials.
    */
   public static RemoteComputeHelper create() {
+    HttpTransportOptions transportOptions = ComputeOptions.getDefaultHttpTransportOptions();
```
garrettjonesgoogle: Since that's not a surface change, I think we can defer it until later.

garrettjonesgoogle: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1745

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
garrettjonesgoogle: @michaelbausor We probably want to allow the setting of transport options in `ServiceOptions`.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
garrettjonesgoogle: Hmm @michaelbausor you make a good point. I was trying to clean up the Default*Rpc classes with this change, but it encourages bad behavior to users.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -63,4 +70,13 @@ public void builder() {
     assertThat(options.getProjectId()).isEqualTo(projectId);
     assertThat(options.getPrefetchChunks()).isEqualTo(2);
   }
+
+  class DummyTransportOptions implements TransportOptions {
```
garrettjonesgoogle: In Spanner in particular I get the following:

`java.lang.ClassNotFoundException: com.google.auth.http.HttpTransportFactory`

I didn't want to mess around with dependencies, so this seemed like the easiest way to go.

garrettjonesgoogle:
Addressed PR feedback, PTAL

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/DefaultBigQueryRpc.java
```diff
@@ -80,7 +80,7 @@
   private final Bigquery bigquery;
 
   public DefaultBigQueryRpc(BigQueryOptions options) {
-    HttpTransportOptions transportOptions = options.getHttpTransportOptions();
+    HttpTransportOptions transportOptions = (HttpTransportOptions) options.getTransportOptions();
```
michaelbausor: Presumably once we add support for GrpcTransportOptions we will have to start checking for null here?

And regarding the naming for `DefaultBigQueryRpc`, does that mean we are going to end up with one class called `DefaultBigQueryRpc` and one called `GrpcBigQueryRpc`? Should we consider renaming to `HttpBigQueryRpc` now?

garrettjonesgoogle: This wouldn't be null - we would check for which type it was. 

Yes, the naming of `Default*Rpc` is a problem - I will take care of renaming it in a following PR; I also want to remove the parameterization of `ServiceRpcT`, because essentially `BigQueryOptions` is currently typed with `DefaultBigQueryRpc` (`HttpBigQueryRpc`), meaning http hasn't been removed from the hierarchy yet.

garrettjonesgoogle: I annotated #1723 regarding the `ServiceRpcT` issue.

michaelbausor:
LGTM

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -186,7 +186,10 @@ public int hashCode() {
 
   @Override
   public boolean equals(Object obj) {
-    if (!(obj instanceof HttpTransportOptions)) {
+    if (obj == null) {
```
vam-google: Typically there is also an if(this == obj) return true; before everything to avoid redundant checks for equality with itself.

garrettjonesgoogle:
For the record: The Codacy warnings were already cleaned up before any reviews started.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
garrettjonesgoogle: I'll leave this for another day. I doubt people are doing much comparison of these.

PR #1725, opened 2017-03-09
Fix logging test failure

michaelbausor:
Use `reset` instead of `readConfiguration` because of missing configuration file on Travis.


garrettjonesgoogle:
LGTM

PR #1728, opened 2017-03-10
logging: don't hard code config string

pongad:
This PR turns the hard coded config string into
a map and a rendering function.
This doesn't do anything straightaway,
but will simplify #1721 .


michaelbausor:
LGTM

Issue #1527, opened 2017-01-12
BigQuery 'JobConfiguration.Type' enum should be public


polleyg:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobConfiguration.java#L39

`JobConfiguration.Type` is package-private. Yet, there is a **public** method in its `Builder` that returns the type of the job configuration, which is kind of useless. Because the enum itself is package private, it's not accessible outside the package/API.

Thus, it makes it hard to:

1. Check the type returned e.g. is a `QUERY`, `EXTRACT`..
2. Write unit tests

For example, clients should be able to do this:

`assertTrue(jobInfo.getConfiguration().getType() == JobConfiguration.Type.QUERY);`

..and not have to do this:

`assertTrue(jobInfo.getConfiguration() instanceof QueryJobConfiguration);`

Is there a good reason why the enum is package-private, and not public?

tswast:
Seems like an easy fix that can increase usability. I'll mark as blocking GA, but @garrettjonesgoogle since you're more familiar with the Java client library surface than I am you can decide otherwise.

PR #1736, opened 2017-03-11
make JobConfiguration.Type enum public

neozwu:
make JobConfiguration.Type enum public, so it can be referenced from outside of the class. (#1527)


garrettjonesgoogle:
LGTM, merge when >= 1 tests pass.

PR #1738, opened 2017-03-11
Reconcile RetrySettings & Clock (google-cloud-java part)

vam-google:
Reconcile RetrySettings in GAX with RetryParams in google-cloud-java
GoogleCloudPlatform/google-cloud-java#1574
Reconcile NanoClock in GAX with Clock in google-cloud-java
GoogleCloudPlatform/google-cloud-java#1575

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateSubscriptionAndPullMessages.java
@@ -46,7 +46,7 @@ public class CreateSubscriptionAndPullMessages {
           @Override
           public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
             System.out.println("got message: " + message.getData().toStringUtf8());
-            consumer.accept(AckReply.ACK, null);
+            consumer.accept(AckReply.ACK);
           }
         };
     Subscriber subscriber = null;
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/MessageReceiverSnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/MessageReceiverSnippets.java
@@ -51,9 +51,9 @@ public class MessageReceiverSnippets {
     MessageReceiver receiver = new MessageReceiver() {
       public void receiveMessage(final PubsubMessage message, final AckReplyConsumer consumer) {
         if (blockingQueue.offer(message)) {
-          consumer.accept(AckReply.ACK, null);
+          consumer.accept(AckReply.ACK);
         } else {
-          consumer.accept(AckReply.NACK, null);
+          consumer.accept(AckReply.NACK);
         }
       }
     };
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/LoggingSettings.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/LoggingSettings.java
@@ -20,15 +20,15 @@ import static com.google.cloud.logging.spi.v2.PagedResponseWrappers.ListLogsPage
 import static com.google.cloud.logging.spi.v2.PagedResponseWrappers.ListMonitoredResourceDescriptorsPagedResponse;
 
 import com.google.api.MonitoredResourceDescriptor;
-import com.google.api.gax.bundling.BundlingSettings;
-import com.google.api.gax.bundling.RequestBuilder;
+import com.google.api.gax.batching.BatchingSettings;
+import com.google.api.gax.batching.RequestBuilder;
 import com.google.api.gax.core.FlowControlSettings;
 import com.google.api.gax.core.FlowController.LimitExceededBehavior;
 import com.google.api.gax.core.GoogleCredentialsProvider;
 import com.google.api.gax.core.RetrySettings;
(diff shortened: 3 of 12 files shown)


shinfan:
I am concerned with the CI build failure.

google-cloud-translate/src/test/java/com/google/cloud/translate/TranslateImplTest.java
```diff
@@ -78,27 +79,28 @@
...
+  private static final RetrySettings NO_RETRY_SETTINGS = ServiceOptions.getNoRetrySettings();
 
   private TranslateOptions options;
   private TranslateRpcFactory rpcFactoryMock;
   private TranslateRpc translateRpcMock;
   private Translate translate;
 
-  @Rule
-  public ExpectedException thrown = ExpectedException.none();
+  @Rule public ExpectedException thrown = ExpectedException.none();
```
vam-google: Reverted

pom.xml
```diff
@@ -92,7 +92,7 @@
     <github.global.server>github</github.global.server>
     <google.auth.version>0.6.0</google.auth.version>
     <grpc.version>1.0.3</grpc.version>
-    <gax.version>0.4.0</gax.version>
+    <gax.version>0.3.2-SNAPSHOT</gax.version>
```
vam-google: Oh, sorry, it was caused by my painful merge. It should be 0.5.0. Changed.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -77,17 +81,22 @@
   private static final String LIBRARY_VERSION = defaultLibraryVersion();
   private static final String APPLICATION_NAME =
       LIBRARY_VERSION == null ? LIBRARY_NAME : LIBRARY_NAME + "/" + LIBRARY_VERSION;
-  private static final long serialVersionUID = -5714029257168617973L;
 
   private static final String META_FILE_ROOT = "/META-INF/maven/";
   private static final String META_VERSION_KEY = "version";
+  private static final RetrySettings DEFAULT_RETRY_SETTINGS = getDefaultRetrySettingsBuilder()
+      .build();
+  private static final RetrySettings NO_RETRY_SETTINGS = getDefaultRetrySettingsBuilder()
```
vam-google: This would mean that those should be the "default" options for the generated clients and within the gax package in general, but it is not the case, since those are default only for gcj/manual clients.

google-cloud-core/src/main/java/com/google/cloud/RetryHelper.java
```diff
@@ -16,234 +16,47 @@
...
-import java.util.logging.Logger;
 
 /**
- * Utility class for retrying operations. For more details about the parameters, see
- * {@link RetryParams}. If the request is never successful, a {@link RetriesExhaustedException} will
- * be thrown.
- *
- * @param <V> return value of the closure that is being run with retries
+ * Utility class for retrying operations. For more details about the parameters, see {@link
+ * RetrySettings}.
```
vam-google: Added

google-cloud-core/src/main/java/com/google/cloud/ExceptionHandler.java
```diff
@@ -264,6 +261,12 @@ boolean shouldRetry(Exception ex) {
   }
 
   @Override
+  public TimedAttemptSettings createNextAttempt(Throwable prevThrowable,
+      TimedAttemptSettings prevSettings) {
+    return null;
```
vam-google: Added comment

google-cloud-core/src/main/java/com/google/cloud/BaseServiceException.java
```diff
@@ -295,12 +295,9 @@ private static String message(IOException exception) {
     return exception.getMessage();
   }
 
-  protected static void translateAndPropagateIfPossible(RetryHelper.RetryHelperException ex) {
+  protected static void translate(RetryHelper.RetryHelperException ex) {
     if (ex.getCause() instanceof BaseServiceException) {
       throw (BaseServiceException) ex.getCause();
     }
```
vam-google: No. This method is used in the set of other translateAndThrow() methods (for example DatastoreException.translateAndThrow()). If the cause is not BaseServiceException, it will be wrapped in a client-specific exception (i.e. will become wrapped in a concrete implementation of BaseServiceException). I think that method exists to avoid redundant wrapping of BaseServiceExceptions in another BaseServiceException (so there will be no useless layers).

vam-google:
Adressed PR comments. PTAL. The CI will not pass at this point, because it needs https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1775 PR to be merged first (without that the code is not compilable)

google-cloud-core/src/main/java/com/google/cloud/ExceptionHandler.java
```diff
@@ -33,7 +35,7 @@
 /**
  * Exception handling used by {@link RetryHelper}.
  */
-public final class ExceptionHandler implements Serializable {
+public final class ExceptionHandler implements ExceptionRetryAlgorithm, Serializable {
```
shinfan: After discussed offline, we decide to keep the name ExceptionHandler

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
shinfan: Ok makes sense

shinfan:
LGTM after at least two of the tests pass.

PR #1739, opened 2017-03-13
logging: update from proto

### LINKED ISSUE #481: Multiple instances of "all X goes through this method"
This is a protodoc issue.
See writeLogEntries for an example.

pongad:
Fix #481.

Changes are mostly comments so this should be straight forward.
The only non-comment change is the addition of smoke test.
Are we OK merging this into the repo?


garrettjonesgoogle:
Remove the smoke test from the PR. Currently we always have to manually do this; we still have to fix the code generation process so that it doesn't put the smoke test here.

garrettjonesgoogle:
LGTM

PR #1753, opened 2017-03-14
Remove deperecated methods from GCE

shinfan:
- Removed all references of deprecated methods from GCE
- Fixed a StackOverflow bug in Subnetwork.java where new getter/setters recursively calling themselves


google-cloud-compute/src/main/java/com/google/cloud/compute/Subnetwork.java
```diff
@@ -104,22 +95,12 @@ Builder setGatewayAddress(String gatewayAddress) {
 
     @Override
     public Builder setNetwork(NetworkId network) {
-      return setNetwork(network);
```
garrettjonesgoogle: Makes me wonder how many people are using google-cloud-compute...

shinfan:
I will try to wait for the tests to pass before merge.

garrettjonesgoogle:
Also fix the spelling in your commit message when you merge.

PR #1756, opened 2017-03-15
Fixing intermittent pubsub test failures

garrettjonesgoogle:
`StreamingSubscriberConnection` already handled this failure scenario correctly, but `PollingSubscriberConnection` did not. I have made them match each other.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
@@ -16,8 +16,6 @@
 
 package com.google.cloud.pubsub.spi.v1;
 
-import static com.google.cloud.pubsub.spi.v1.StatusUtil.isRetryable;
-
 import com.google.api.gax.core.FlowController;
 import com.google.api.stats.Distribution;
 import com.google.auth.Credentials;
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StatusUtil.java
```diff
@@ -17,21 +17,31 @@
...
+    StatusRuntimeException statusRuntimeException = (StatusRuntimeException) error;
+    switch (statusRuntimeException.getStatus().getCode()) {
       case DEADLINE_EXCEEDED:
       case INTERNAL:
       case CANCELLED:
       case RESOURCE_EXHAUSTED:
-      case UNAVAILABLE:
         return true;
+      case UNAVAILABLE:
+        if (statusRuntimeException.getMessage().contains("Server shutdownNow invoked")) {
```
michaelbausor: Does this message have special status?

garrettjonesgoogle: I'm not sure what you mean by "special status", but when `UNAVAILABLE` is thrown for this reason, it can't be retried. I'm not sure how stable the message is. This particular change isn't strictly necessary because the `!isAlive()` code path will exit `onFailure` before this logic is called. If you're concerned, I can remove it.

michaelbausor: Not particularly concerned, but was unsure whether this was something we could rely on. There doesn't seem to be any downside to not retrying in this case.

michaelbausor:
LGTM

PR #1758, opened 2017-03-15
Changing the AckReplyConsumer interface

davidtorres:
@garrettjonesgoogle  @pongad I'm changing the interface of AckReplyConsumer because it really does not buy us anything to be able to accept an exception along with the ackreply. It actually makes the interface ambiguous, i.e what does it mean to accept an AckReply and a non null exception. And moreover I'd like to avoid methods in which you can pass null parameters.


pongad:
LGTM from me. @garrettjonesgoogle can +2

garrettjonesgoogle:
LGTM - but why are there no Travis test results here? @davidtorres could you push a PR update to rerun the tests?

davidtorres:
Tests are running, travis already passing. Thanks for reviewing.

pongad:
LGTM, thank you!

PR #1759, opened 2017-03-16
Removing ServiceRpcT parameterization

### LINKED ISSUE #1601: Change service hierarchy so that http vs grpc isn't baked in

garrettjonesgoogle:
Fixes #1601 

Also:

* Moving spi/*Rpc classes to spi/{MAJOR_VERSION}/*Rpc
* Renaming Default*Rpc classes to Http*Rpc/Grpc*Rpc as appropriate

Note: There are references in google-cloud-examples to Rpc-specific stuff (Only Tuples I think); I will file a separate issue to remove those.


michaelbausor:
I re-read the issue, but I am still not sure why we need to remove the ServiceRpcT parameterization. As an example, currently `BigQueryOptions` extends `ServiceOptions<BigQuery, BigQueryRpc, BigQueryOptions>`. The `BigQueryRpc` interface is independent of http/grpc, even if `DefaultBigQueryRpc` (now `HttpBigQueryRpc`) is not.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/BigQueryRpc.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.cloud.bigquery.spi;
+package com.google.cloud.bigquery.spi.v2;
```
michaelbausor: Do we need to move the factories as well? If not, will `BigQueryRpcFactory` always construct the latest major version? And then, how would we create older versions? Or would it only be possible to construct a single major version with the factory, as the manual layer is tied to a specific one.

garrettjonesgoogle:
@michaelbausor , The Rpc interface (e.g. `BigQueryRpc`) is tied to old-platform vs proto and also the specific major version. When the Rpc class is old-platform, only http 1 is available. My understanding is that the design of using a fully wrapped surface means that old-platform vs proto and major versions shouldn't be exposed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/SpannerRpcFactory.java
```diff
@@ -20,4 +20,4 @@
 import com.google.cloud.spi.ServiceRpcFactory;
 
 /** Factory to create instances of {@code SpannerRpc}. */
-public interface SpannerRpcFactory extends ServiceRpcFactory<SpannerRpc, SpannerOptions> {}
+public interface SpannerRpcFactory extends ServiceRpcFactory<SpannerOptions> {}
```
michaelbausor: Nit: newline

google-cloud-compute/src/main/java/com/google/cloud/compute/spi/v1/HttpComputeRpc.java
```diff
@@ -71,16 +71,17 @@
 import com.google.cloud.HttpTransportOptions;
 import com.google.cloud.compute.ComputeException;
 import com.google.cloud.compute.ComputeOptions;
+import com.google.cloud.compute.spi.v1.ComputeRpc;
```
michaelbausor: Nit: fix this one

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/WatchdogInterceptorTest.java
```diff
@@ -25,6 +25,7 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import com.google.cloud.spanner.spi.v1.WatchdogInterceptor;
```
michaelbausor: Nit: Fix this one

google-cloud-compute/src/main/java/com/google/cloud/compute/ComputeOptions.java
```diff
@@ -49,7 +50,7 @@ public Compute create(ComputeOptions options) {
     private static final ComputeRpcFactory INSTANCE = new DefaultComputeRpcFactory();
 
     @Override
-    public ComputeRpc create(ComputeOptions options) {
+    public ServiceRpc create(ComputeOptions options) {
```
michaelbausor: Why can't we continue to return `ComputeRpc` here?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -112,7 +112,7 @@ public static HttpTransportOptions getDefaultHttpTransportOptions() {
     return SCOPES;
   }
 
-  protected BigQueryRpc getBigQueryRpc() {
+  protected BigQueryRpc getBigQueryRpcV2() {
```
michaelbausor: What is the benefit of this rename? Are we planning on having `getBigQueryRpcV3` as well when it is available? How would we implement that?

garrettjonesgoogle: There aren't specific plans for specific major versions, but this aligns the rpc implementation more closely with how we are doing the GAPIC layer. The class itself (`BigQueryRpc`) has the major version in its package, so multiple instances of the same name could co-exist in google-cloud-java; however, two getter functions with the same name can't coexist, so we have to disambiguate this way.

google-cloud-compute/src/main/java/com/google/cloud/compute/ComputeOptions.java
garrettjonesgoogle: It's not that we can't; the problem is that the `ComputeRpc` interface is tied to a specific major version of compute, and if we want to swap it out later for a different major version, that would be a breaking change. However, if the return type is `ServiceRpc`, we can swap it out without breaking direct callers.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/BigQueryRpc.java
garrettjonesgoogle: We don't need to move the `*RpcFactory` interfaces because they are actually empty interfaces, and aren't tied to a specific major version or transport.

google-cloud-compute/src/main/java/com/google/cloud/compute/ComputeOptions.java
michaelbausor: That makes sense, thanks.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
michaelbausor: After discussion offline, I am OK with this change

PR #1768, opened 2017-03-17
adding functions to manage life cycle of resources in ITComputeTest

neozwu:
In ITComputeTest, we use "JUnit timeout rule" to specify test case timeout. Timeout rule (unlike timeout parameter in @Test annotation) applies to the entire test fixture (that includes all @Before methods + @Test method + all @After methods). Therefore, when a test times out in the middle of execution, teardown method (@After method) won't be executed. This will result in resource leak, and accumulation of unused resources (instances, disks, snapshots, etc.) in google cloud compute engine.  One approach is to use timeout parameter rather than timeout rule and properly clean up resources in @After method.  However timeout parameter will force tests (@Test method) to run in a different thread, thus may cause potential thread-safety issue.  In this PR, a different approach is used, which registers all resources that are created by test cases in a static data structure, and try to delete all these resources after all tests are executed (i.e. in the @Afterclass method).  The implementation tries to be as generic as possible, but without introducing any change of existing resource objects.

### DIFF
--- a/google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
+++ b/google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
@@ -43,28 +43,27 @@ public class ResourceCleaner {
   }
 
   private class ManagedResource<T> {
-    final T res;
-    Function<T, Void> func;
+    final T resourceId;
+    Function<T, Void> deleteFunc;
 
-    private ManagedResource(T res, Function<T, Void> func) {
-      this.res = res;
-      this.func = func;
+    private ManagedResource(T resourceId, Function<T, Void> deleteFunc) {
+      this.resourceId = resourceId;
+      this.deleteFunc = deleteFunc;
     }
 
     public void delete() {
-      func.apply(res);
+      deleteFunc.apply(resourceId);
     }
   }
 
   public void cleanUp() {
     for (Map.Entry<Object, ManagedResource<?>> r : resources.entrySet()) {
-      System.out.println("Clean up compute engine resource: " + r.getKey().toString());
       r.getValue().delete();
     }
     resources.clear();
   }
 
-  public ResourceCleaner add(AddressId addressId) {
+  public void add(AddressId addressId) {
     ManagedResource<AddressId> r =
         new ManagedResource<>(
             addressId,
@@ -76,10 +75,9 @@ public class ResourceCleaner {
               }
             });
     resources.put(addressId, r);
-    return this;
   }
 
-  public ResourceCleaner add(DiskId diskId) {
+  public void add(DiskId diskId) {
     ManagedResource<DiskId> r =
         new ManagedResource<>(
             diskId,
@@ -91,10 +89,9 @@ public class ResourceCleaner {
               }
             });
     resources.put(diskId, r);
-    return this;
   }
 
-  public ResourceCleaner add(SnapshotId snapshotId) {
+  public void add(SnapshotId snapshotId) {
     ManagedResource<SnapshotId> r =
         new ManagedResource<>(
             snapshotId,
@@ -106,10 +103,9 @@ public class ResourceCleaner {
               }
             });
     resources.put(snapshotId, r);
-    return this;
   }
 
-  public ResourceCleaner add(NetworkId networkId) {
+  public void add(NetworkId networkId) {
     ManagedResource<NetworkId> r =
         new ManagedResource<>(
             networkId,
@@ -121,10 +117,9 @@ public class ResourceCleaner {
               }
             });
     resources.put(networkId, r);
-    return this;
   }
 
-  public ResourceCleaner add(SubnetworkId subnetworkId) {
+  public void add(SubnetworkId subnetworkId) {
     ManagedResource<SubnetworkId> r =
         new ManagedResource<>(
             subnetworkId,
@@ -136,10 +131,9 @@ public class ResourceCleaner {
               }
             });
     resources.put(subnetworkId, r);
-    return this;
   }
 
-  public ResourceCleaner add(ImageId imageId) {
+  public void add(ImageId imageId) {
     ManagedResource<ImageId> r =
         new ManagedResource<>(
             imageId,
@@ -151,10 +145,9 @@ public class ResourceCleaner {
               }
             });
     resources.put(imageId, r);
-    return this;
   }
 
-  public ResourceCleaner add(InstanceId instanceId) {
+  public void add(InstanceId instanceId) {
     ManagedResource<InstanceId> r =
         new ManagedResource<>(
             instanceId,
@@ -166,10 +159,9 @@ public class ResourceCleaner {
               }
             });
     resources.put(instanceId, r);
-    return this;
   }
 
-  public <T> ResourceCleaner remove(T resourceId) {
+  public ResourceCleaner remove(Object resourceId) {
     if (!resources.containsKey(resourceId)) {
         throw new NoSuchElementException(resourceId + " has not been added to managed resources");
     }
--- a/google-cloud-compute/src/test/java/com/google/cloud/compute/it/ITComputeTest.java
+++ b/google-cloud-compute/src/test/java/com/google/cloud/compute/it/ITComputeTest.java
@@ -115,7 +115,7 @@ public class ITComputeTest {
   }
 
   @AfterClass
-  public static void teardown() {
+  public static void tearDown() {
     resourceCleaner.cleanUp();
   }
 


google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
...
+import com.google.cloud.compute.Operation;
+import com.google.cloud.compute.Snapshot;
+import com.google.cloud.compute.SnapshotId;
+import com.google.cloud.compute.Subnetwork;
+import com.google.cloud.compute.SubnetworkId;
+import com.google.common.base.Function;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ManagedResourceHelper {
```
shinfan: Why is this only for Compute? It seems to me that this helper can be generalized and used for other APIs as well. e.g. We can make this helper a general class and create a API-specific resource deletion handler per API.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
...
+    final T res;
+    Function<T, Void> func;
+
+    private ManagedResource(T res) {
+      this.res = res;
+
+      @SuppressWarnings("unchecked")
+      Function<T, Void> localFunc = (Function<T, Void>) getDeleteMethod(res);
+
+      this.func = localFunc;
```
shinfan: this.func = (Function<T, Void>) getDeleteMethod(res);

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
...
+  public <T> ManagedResourceHelper addManagedResource(T res) {
+    ManagedResource<T> r = new ManagedResource<>(res);
+    resources.add(r);
+    return this;
+  }
+
+  public void cleanUpResources() {
+    for (ManagedResource<?> r : resources) {
+      r.delete();
+      //TODO (neowu): add logging if a resource has not been deleted in test fixture
```
shinfan: How do you detect if a resource has been deleted or not?

shinfan:
Thanks for fixing this! Just a few comments.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
...
+      return new Function<Address, Void>() {
+        @Override
+        public Void apply(Address address) {
+          try {
+            Operation operation = address.delete();
+            operation.waitFor();
+          } catch (Exception e) {
+
+          } finally {
+            return null;
```
garrettjonesgoogle: It doesn't make sense for this `return` to be in a `finally` block.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
...
+
+  private <T> Function<?, Void> getDeleteMethod(T res) {
+    if (res instanceof Address) {
+      return new Function<Address, Void>() {
+        @Override
+        public Void apply(Address address) {
+          try {
+            Operation operation = address.delete();
+            operation.waitFor();
+          } catch (Exception e) {
```
garrettjonesgoogle: We should print the exceptions that we see in deletes.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
...
+  public void cleanUpResources() {
+    for (ManagedResource<?> r : resources) {
+      r.delete();
+      //TODO (neowu): add logging if a resource has not been deleted in test fixture
+    }
+    resources.clear();
+  }
+
+  private <T> Function<?, Void> getDeleteMethod(T res) {
+    if (res instanceof Address) {
```
garrettjonesgoogle: With this setup, we won't know that a particular type isn't supported at compile time, yet we know the type at the call site of `addManagedResource`. I think instead we should have a separate `addManagedResource` overload for each type in Compute. Then, if we add a new one, we'll know at compile time if we don't support it yet.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
...
+
+      this.func = localFunc;
+    }
+
+    public void delete() {
+      func.apply(res);
+    }
+  }
+
+  public <T> ManagedResourceHelper addManagedResource(T res) {
```
garrettjonesgoogle: you could probably just shorten this to `add`.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
garrettjonesgoogle: I think the `helper` suffix is too general. Maybe name this class `ResourceCleaner`.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
```diff
@@ -0,0 +1,230 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
neozwu: resolved

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceCleaner.java
```diff
@@ -0,0 +1,210 @@
...
+  private <T> Function<?, Void> getDeleteMethod(T res) {
+
+    if (res instanceof AddressId) {
+      return new Function<AddressId, Void>() {
+        @Override
+        public Void apply(AddressId addressId) {
+          compute.deleteAddress(addressId);
+          return null;
+        }
+      };
```
garrettjonesgoogle: It would be much cleaner to put these delete `Function`s inside the corresponding `add` method for each type. Then, you never have to bother with an `instanceof` check. Additionally, I think it would mean you could remove the type parameterization (`<T>`) from MonitoredResource too.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
neozwu: I agree the problem here is more generic and not only applied to compute engine.  I'd like to have a specific version for compute engine for now, and create a more generic version if we need. The reason is if we make a more general class, I might need to change things in various libraries.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
neozwu: yeah, this is ugly. The reason I'd like to have a seemingly repeated statement is because then I can limit the scope of `@SuppressWarnings("unchecked")`

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceCleaner.java
```diff
@@ -0,0 +1,210 @@
...
+  public ManagedResourceCleaner add(AddressId addressId) {
+    ManagedResource<AddressId> r = new ManagedResource<>(addressId);
+    resources.put(addressId, r);
+    return this;
+  }
+
+  public ManagedResourceCleaner remove(AddressId addressId) {
+    resources.remove(addressId);
+    return this;
+  }
```
garrettjonesgoogle: You don't need an overload for `remove` - you can have just one implementation. Maybe throw an exception if the id is not found in `resources`.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
neozwu: done

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
neozwu: done

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
neozwu: now use overloading methods for all supported type, so we know whether it's supported at compile time.

google-cloud-compute/src/test/java/com/google/cloud/compute/it/ITComputeTest.java
```diff
@@ -1155,9 +1193,13 @@ public void testListDisksAndSnapshots() throws InterruptedException, TimeoutExce
     }
     assertEquals(2, count);
     compute.deleteDisk(firstDiskId);
+    managedResourceCleaner.remove(firstDiskId);
     compute.deleteDisk(secondDiskId);
+    managedResourceCleaner.remove(secondDiskId);
     compute.deleteSnapshot(firstSnapshotId);
+    managedResourceCleaner.remove(firstSnapshotId);
     compute.deleteSnapshot(secondSnapshotId);
+    managedResourceCleaner.remove(secondSnapshotId);
```
garrettjonesgoogle: For tests where there is no check after deletion, you could just remove the deletes altogether, since the resource cleaner will take care of it in `tearDown`.

neozwu:
@shinfan @garrettjonesgoogle PTAL

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceHelper.java
garrettjonesgoogle: I agree about just focusing on Compute to start with.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceCleaner.java
neozwu: done

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ManagedResourceCleaner.java
neozwu: done, although still have type parameterization.

google-cloud-compute/src/test/java/com/google/cloud/compute/it/ITComputeTest.java
neozwu: done

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
```diff
@@ -0,0 +1,179 @@
...
+    this.compute = compute;
+    this.resources = new HashMap<>();
+  }
+
+  public static ResourceCleaner create(Compute compute) {
+    return new ResourceCleaner(compute);
+  }
+
+  private class ManagedResource<T> {
+    final T res;
```
garrettjonesgoogle: rename: `resourceId` (this variable holds an id, not a full resource object)

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
```diff
@@ -0,0 +1,179 @@
...
+    Function<T, Void> func;
```
garrettjonesgoogle: rename: `deleteFunc` (`func` is much too general)

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public void delete() {
+      func.apply(res);
+    }
+  }
+
+  public void cleanUp() {
+    for (Map.Entry<Object, ManagedResource<?>> r : resources.entrySet()) {
+      System.out.println("Clean up compute engine resource: " + r.getKey().toString());
```
garrettjonesgoogle: I don't think we need to commit this println.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
```diff
@@ -0,0 +1,179 @@
...
+      r.getValue().delete();
+    }
+    resources.clear();
+  }
+
+  public ResourceCleaner add(AddressId addressId) {
```
garrettjonesgoogle: I think we should remove the return value from the `add` methods since it's not used (and I don't see a use for it).

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
```diff
@@ -0,0 +1,179 @@
...
+              public Void apply(InstanceId instanceId) {
+                compute.deleteInstance(instanceId);
+                return null;
+              }
+            });
+    resources.put(instanceId, r);
+    return this;
+  }
+
+  public <T> ResourceCleaner remove(T resourceId) {
```
garrettjonesgoogle: The type parameterization isn't necessary here - you can just accept `Object`.

google-cloud-compute/src/test/java/com/google/cloud/compute/it/ITComputeTest.java
```diff
@@ -109,6 +111,12 @@
   public static void beforeClass() {
     RemoteComputeHelper computeHelper = RemoteComputeHelper.create();
     compute = computeHelper.getOptions().getService();
+    resourceCleaner = ResourceCleaner.create(compute);
+  }
+
+  @AfterClass
+  public static void teardown() {
```
garrettjonesgoogle: Make the 'd' here upper case to stop codacy from yelling at you.

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
neozwu: done

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
neozwu: done

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
neozwu: the println is deleted

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
neozwu: done

google-cloud-compute/src/main/java/com/google/cloud/compute/testing/ResourceCleaner.java
neozwu: done

google-cloud-compute/src/test/java/com/google/cloud/compute/it/ITComputeTest.java
neozwu: done

garrettjonesgoogle:
LGTM

Issue #1769, opened 2017-03-17
Move general purpose classes from google-cloud-spanner to google-cloud-core


vkedia:
Few of the classes in google-cloud-spanner are general enough that they can be useful in existing apis (in the case of ByteArrays) or presumably might be useful in future for other apis (in the case of Date and Timestamp). We should move these to google-cloud-core.

vkedia:
cc @garrettjonesgoogle

vkedia:
Closed by #1770

PR #1770, opened 2017-03-18
Move some classes from spanner to core

vkedia:
Move Date and Timestamp to core
Removed ByteArrays and moved toBase64 and fromBase64 to ByteArray
Removed the fancy toString in ByteArrays since ByteArray.toString does a reasonable job.


vkedia:
Closes #1769

vkedia:
@garrettjonesgoogle Can you please take a look.

google-cloud-core/pom.xml
```diff
@@ -116,6 +116,11 @@
       <version>3.0.0</version>
     </dependency>
     <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java-util</artifactId>
+      <version>3.0.0</version>
```
garrettjonesgoogle: I noticed that protobuf-java-util has a dependency on Guava. It doesn't expose Guava on its surface, but at the same time, it there isn't a shaded variant of it either. We are currently working towards providing a fully shaded variant of google-cloud-java soon (alongside the current variant where Guava isn't shaded).

garrettjonesgoogle:
LGTM

PR #1771, opened 2017-03-20
pubsub: delete deprecated code

pongad:
Two integration tests are moved over to new code
to make sure the library isn't dead on arrival.

### DIFF
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/it/ITPubSubTest.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/it/ITPubSubTest.java
@@ -116,9 +116,9 @@ public class ITPubSubTest {
                   public void receiveMessage(
                       final PubsubMessage message, final AckReplyConsumer consumer) {
                     if (received.set(message)) {
-                      consumer.accept(AckReply.ACK, null);
+                      consumer.accept(AckReply.ACK);
                     } else {
-                      consumer.accept(AckReply.NACK, null);
+                      consumer.accept(AckReply.NACK);
                     }
                   }
                 })


pongad:
@garrettjonesgoogle I wanted to make sure I'm not being too nuke happy. I deleted both [`PubSubExample`](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/PubSubExample.java) and [`LocalPubSubHelper`](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/deprecated/testing/LocalPubSubHelper.java).

As discussed before, I have doubts about the usefulness of these classes, but they are advertised in the README. I think the best course of action is
- delete PubSubExample; it is redundant with gcloud command
- in separate CL, create a helper class that helps create clients that connect to emulator (don't write any code to start/stop emulator itself)

Please let me know what you think.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/it/ITPubSubTest.java
```diff
@@ -16,127 +16,129 @@
...
+    publisherClient.createTopic(topicName);
 
-  @Test
-  public void testNonExistingTopicPolicy() {
-    String topicName = formatForTest("test-non-existing-topic-policy");
-    assertNull(pubsub().getTopicPolicy(topicName));
-  }
+    Policy policy = publisherClient.getIamPolicy(topicName.toString());
+    Binding binding =
+        Binding.newBuilder().setRole("roles/viewer").addMembers("allAuthenticatedUsers").build();
```
garrettjonesgoogle: It's a shame we're losing the typed IAM classes (Role, Identity, Policy) - not something you can address in this PR, though.

PR #1773, opened 2017-03-20
Max ack ext

davidtorres:
@kamalaboulhosn tell me how does this PR looks to you before I send a pull request to the java lib folks


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
```diff
@@ -103,7 +109,7 @@ void extendExpiration(Instant now) {
       expiration = now.plus(Duration.standardSeconds(nextExtensionSeconds));
```
kamalaboulhosn: We should probably take into account the maxAckExtensionPeriod here and not modifyAckDeadline beyond that boundary.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
```diff
@@ -380,6 +392,12 @@ public void run() {
             && outstandingAckHandlers.peek().expiration.compareTo(cutOverTime) <= 0) {
           ExtensionJob job = outstandingAckHandlers.poll();
 
+          if (job.creation.plus(maxAckExtensionPeriod).compareTo(now) < 0) {
```
kamalaboulhosn: <=

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
davidtorres: This is done in line 395

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
davidtorres: Done

davidtorres:
As we spoke made the changes to preserve the stream ack deadline based on the subscriber ack latency distribution

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
kamalaboulhosn: I think there is still a problem. Imagine the job is created at time 0 and maxAckExtensionPeriod is ten minutes. If the message is not acked or nacked, we continue to modify the ack deadline an increasing amount each time. As a result, we could overshoot the maxAckExtensionPeriod. I believe we could, for example, end up extending an additional 5 minutes at minute 8, which means we would not stop modifying the ack deadline until minute 13, which is 3 minutes beyond maxAckExtensionPeriod. That's why I think you need to set expiration to min(now.plus(Duration.standardSeconds(nextExtensionSeconds)), creation.plus(maxAckExtensionPeriod)).

davidtorres: Ah good point, thanks for the clarifying. Fixed.

PR #1783, opened 2017-03-22
fix allocateId so it cannot be called on complete keys

### LINKED ISSUE #1567: Datastore: `allocateId` should only accept an incomplete key
The current behavior is to "trim" the ID or name from the final path element before sending the key to the server:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java#L114

This means you can call `allocateId` on a key with a prepopulated ID, and the server will actuall allocate a completely different ID. This behavior is documented, but it seems surprising. Would be safer to either reject the `allocateId` call if the provided key contains an ID/name (or just let the server reject the call).

Reference:
http://stackoverflow.com/questions/41878706/where-is-allocateids-in-the-google-cloud-datastore-java-library/41879429

neozwu:
fix #1567 to not allow `allocateId` to be called on `Key` object (i.e. complete keys), since:
* allocateId in google cloud datastore doesn't support the function of reserving a user-specified numeric Id (which is supported in GAE datastore)
* calling `allocateId` on key will simply erase the current id and assign a new random id, which may not be what users expect. (Users may expect calling allocatedId on a key will reserve the id in that key)
* the api service will throw an error if a complete key is passed in.

### DIFF
--- a/google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
+++ b/google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
@@ -525,8 +525,6 @@ public class ITDatastoreTest {
     IncompleteKey incompleteKey1 = keyFactory.newKey();
     IncompleteKey incompleteKey2 =
         keyFactory.setKind(KIND2).addAncestors(PathElement.of(KIND1, 10)).newKey();
-    Key key3 = keyFactory.newKey("name");
-    Key key4 = keyFactory.newKey(1);
     List<Key> result =
         DATASTORE.allocateId(incompleteKey1, incompleteKey2, incompleteKey1);
     assertEquals(3, result.size());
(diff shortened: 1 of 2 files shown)


google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -532,14 +528,11 @@ public void testAllocateIdArray() {
     Key key3 = keyFactory.newKey("name");
     Key key4 = keyFactory.newKey(1);
```
garrettjonesgoogle: Delete `key3` and `key4` since they aren't used any more. (Same thing in other tests too)

garrettjonesgoogle:
LGTM

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
neozwu: @garrettjonesgoogle deleted.  PTAL, thanks.

PR #1787, opened 2017-03-22
WIP: Renames: Function -> ApiFunction, RpcStreamObserver -> ApiStreamObserver

garrettjonesgoogle:
Build will fail because GAX won't be released until after this PR is provisionally approved.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,7 +92,7 @@
     <github.global.server>github</github.global.server>
     <google.auth.version>0.6.0</google.auth.version>
     <grpc.version>1.0.3</grpc.version>
-    <gax.version>0.5.0</gax.version>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -92,7 +92,7 @@
     <github.global.server>github</github.global.server>
     <google.auth.version>0.6.0</google.auth.version>
     <grpc.version>1.0.3</grpc.version>
-    <gax.version>0.5.0</gax.version>
+    <gax.version>0.6.0</gax.version>
```
shinfan: Isn't the next version should be 0.5.1?

garrettjonesgoogle: No, this is a breaking change to GAX.

shinfan:
LGTM

PR #1790, opened 2017-03-22
Add x-goog-api-client support to HTTP1.1

shinfan:
This PR depends on the GAX change: https://github.com/googleapis/gax-java/pull/250

cc/ @bjwatson

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
@@ -24,7 +24,6 @@ import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
-import com.google.api.gax.core.GaxPropertiesProvider;
 import com.google.auth.Credentials;
 import com.google.auth.http.HttpCredentialsAdapter;
 import com.google.auth.http.HttpTransportFactory;
@@ -165,11 +164,10 @@ public class HttpTransportOptions implements TransportOptions {
 
   String getXGoogApiClientHeader() {
     return String.format(
-        "gl-java/%s %s/%s gax/%s",
+        "gl-java/%s %s/%s",
         getJavaVersion(),
         ServiceOptions.getGoogApiClientLibName(),
-        ServiceOptions.getLibraryVersion(),
-        GaxPropertiesProvider.getGaxVersion());
+        ServiceOptions.getLibraryVersion());
   }
 
   private static String getJavaVersion() {
--- a/google-cloud-core/src/test/java/com/google/cloud/HttpTransportOptionsTest.java
+++ b/google-cloud-core/src/test/java/com/google/cloud/HttpTransportOptionsTest.java
@@ -64,7 +64,7 @@ public class HttpTransportOptionsTest {
 
   @Test
   public void testHeader() {
-    String expectedHeaderPattern = "^gl-java/.* gccl/.* gax/.*$";
+    String expectedHeaderPattern = "^gl-java/.* gccl/.*";
     assertTrue(Pattern.compile(expectedHeaderPattern)
         .matcher(DEFAULT_OPTIONS.getXGoogApiClientHeader())
         .find());


shinfan:
Note the build will fail until the GAX change is released.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -20,12 +20,15 @@
...
 import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.gax.utils.PropertiesProvider;
 import com.google.auth.Credentials;
 import com.google.auth.http.HttpCredentialsAdapter;
 import com.google.auth.http.HttpTransportFactory;
+import com.google.common.annotations.VisibleForTesting;
```
garrettjonesgoogle: I just realized this is from Guava... we'll have to scrub it out from everywhere.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -154,10 +157,29 @@ public void initialize(HttpRequest httpRequest) throws IOException {
...
         }
+
+        HttpHeaders headers = httpRequest.getHeaders();
+        headers.set("x-goog-api-client", xApiClientHeader());
       }
     };
   }
 
+  @VisibleForTesting
+  String xApiClientHeader() {
```
garrettjonesgoogle: This name is odd... maybe `getXGoogApiClientHeader` instead.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -154,10 +157,29 @@ public void initialize(HttpRequest httpRequest) throws IOException {
...
+    return String.format(
+        "gl-java/%s %s/%s gax/%s grpc/%s",
```
garrettjonesgoogle: It seems best to only include this format in GAX, and invoke it from here.

shinfan: This format is actually different from the ones used in GAPIC (GAPIC header has one extra 'gapic' field)

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
shinfan: Are we removing the dependency of Guava? We have lots of Guava usages in gcj.

garrettjonesgoogle: We are removing it from the surface, but not the implementation.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
garrettjonesgoogle: Ahh ok, that makes sense.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -154,10 +156,28 @@ public void initialize(HttpRequest httpRequest) throws IOException {
...
   }
 
+  String getXGoogApiClientHeader() {
+    return String.format(
+        "gl-java/%s %s/%s gax/%s grpc/%s",
+        getJavaVersion(),
+        ServiceOptions.getGoogApiClientLibName(),
+        ServiceOptions.getLibraryVersion(),
+        GaxPropertiesProvider.getGaxVersion(),
+        GaxPropertiesProvider.getGrpcVersion());
```
garrettjonesgoogle: Why do we care about the grpc version if we are using http?

shinfan: Hmmm...I don't think we do. I will remove it.

shinfan: Done

shinfan:
@garrettjonesgoogle  PTAL. CI passed.

bjwatson:
LGTM, but @garrettjonesgoogle has final approval.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -154,10 +156,27 @@ public void initialize(HttpRequest httpRequest) throws IOException {
...
+
+        HttpHeaders headers = httpRequest.getHeaders();
+        headers.set("x-goog-api-client", getXGoogApiClientHeader());
       }
     };
   }
 
+  String getXGoogApiClientHeader() {
+    return String.format(
+        "gl-java/%s %s/%s gax/%s",
```
garrettjonesgoogle: The gax version isn't relevant here either, right? The equivalent would be the version of google-api-java-client I think. cc @bjwatson

bjwatson: C# is reporting it, because they use the GAX library to support their HTTP calls. It really depends on whether Java does the same. If GAX is not used at all for these calls, then remove it.

garrettjonesgoogle: We may have a situation in the future where gax is used for certain http contexts, and in that case I think it will make sense. As of now, none of our http codepaths use gax.

shinfan: Removed

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
shinfan: Removed

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
shinfan: Done

pom.xml
```diff
@@ -92,7 +92,7 @@
     <github.global.server>github</github.global.server>
     <google.auth.version>0.6.0</google.auth.version>
     <grpc.version>1.0.3</grpc.version>
-    <gax.version>0.6.0</gax.version>
+    <gax.version>0.6.1</gax.version>
```
garrettjonesgoogle: Now you don't even need this :-) Not that it isn't beneficial - let's keep the update anyway.

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -154,10 +155,26 @@ public void initialize(HttpRequest httpRequest) throws IOException {
...
+
+        HttpHeaders headers = httpRequest.getHeaders();
+        headers.set("x-goog-api-client", getXGoogApiClientHeader());
       }
     };
   }
 
+  String getXGoogApiClientHeader() {
+    return String.format(
+        "gl-java/%s %s/%s",
```
garrettjonesgoogle: @bjwatson So... I know that maybe I'm being a huge pain here, but I'm concerned about the lack of indication within this header that we're using the old-platform stack. There is no 'gdcl' key, no key to indicate the version of google-api-java-client, etc. This concerns me for the case of Datastore, where we intend to switch from old-platform to grpc down the road. I know that the 'gdcl' key was going to be something we added later, but my understanding was that was because we weren't adding this header for old-platform APIs until later; now, we are adding the header.

pom.xml
shinfan: sgtm

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
bjwatson: The current lack of `gapic` or `grpc` in the header, and then the presence of one or both later, tells us this story.

You can feel free to add `gdcl` if you like. There's just no plan to track it unless it's the top-most client library in the stack (i.e. `gccl` is not present, since `gccl` always layers on `gdcl`).

bjwatson: Actually, my statement is logically false. What I meant to say is that `gdcl` never layers on `gccl`.

garrettjonesgoogle: Relying on the absence of a key is a very shaky strategy... we never know what variants we will cook up in the future, and none of them will be distinguishable when no key is present to disambiguate them. I feel it is quite important to know old-platform vs old-platform+gapic vs new-platform+grpc vs new-platform+http. I almost feel like that is more important than `gccl`.

bjwatson: Presumably, that what `Google-HTTP-Java-Client` tells us in the `user-agent` header, correct? My understanding is that is says the legacy client library for Java was involved in the API call.

Part of the reason I'm not in a rush to add `gdcl` to `x-goog-api-client` is that we have a long history of legacy client library versions (i.e. not GAPIC) out there that don't use `x-goog-api-client`. We need to keep tracking them, so `gdcl` is not super valuable in the short term.

garrettjonesgoogle: Hmm I suppose if this information is already in another header, we can get by without it in our header for now. Someone else will have to deal with the mess of analyzing the data...

garrettjonesgoogle:
LGTM

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
bjwatson: Yep, indirectly me and directly someone else. ;)

PR #1793, opened 2017-03-23
Changed a test to not fail if the ordering of map changed.

vkedia:


garrettjonesgoogle:
LGTM after tests all pass

PR #1794, opened 2017-03-23
Fix links in READMEs

davidxia:
Use Markdown formatting that creates links.


davidxia:
cc @garrettjonesgoogle @michaelbausor 

Same as https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1784

garrettjonesgoogle:
I don't remember all of these links being broken before. Something must have changed with the markdown processing. In any case, thanks!

Issue #1795, opened 2017-03-23
Fix LoggingHandler flush


pongad:
Logging handler's `flush()` calls `LoggingImpl.write` which eventually calls `LoggingClient.writeLogEntries` which uses bundling. So, handler flushing just moves messages from one buffer to another without actually making any RPC call.

## Proposed fix
LoggingHandler should keep a count of pending messages. The count increments with calls to publish and decrements when RPC returns.

Flush should force batcher to issue all buffered messages and just wait for the count to reach zero. No call to publish may succeed when flush is "in flight" otherwise flush could wait forever.

Encountering a log above the configured flush level should also force a flush.

garrettjonesgoogle:
@pongad could you reword the description in terms of a command instead of a problem? (e.g. "Fix logging flush")

pongad:
@garrettjonesgoogle I never seem to be able to remember. Done.

garrettjonesgoogle:
Also reword the extended description, recommending a course of action to take.

garrettjonesgoogle:
This properly waits for the RPCs to finish with @pongad 's last change; the triggering of RPCs by the call can be done after GA. Removing the release-blocking label.

pongad:
@garrettjonesgoogle I think I need to send this back to you. The generated client must expose a way to flush. I don't think this is possible at the moment.

pongad:
I'll resolve this as duplicate of #2796.

PR #1796, opened 2017-03-23
logging: remove handler's buffer

pongad:
LoggingHandler's buffer is redundant,
since it's already using the batching feature.

LoggingHandler.flush previously just flushes its own buffer
and put messages in the batcher's buffer,
without necessarily making RPC calls.
This PR does not fix this problem,
but it makes flush obviously wrong instead of subtly.

The test for flush size is also removed.
Flush size should be forwareded to the batcher,
which already has its own test.

Updates #1795.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -20,6 +20,8 @@ import static com.google.common.base.MoreObjects.firstNonNull;
 
 import com.google.cloud.MonitoredResource;
 import com.google.cloud.logging.Logging.WriteOption;
+import com.google.api.gax.core.ApiFutures;
+import com.google.api.gax.core.ApiFutureCallback;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -372,22 +373,9 @@ public void publish(LogRecord record) {
...
-        if (entry != null) {
-          buffer.add(entry);
-        }
-        if (buffer.size() >= flushSize || record.getLevel().intValue() >= flushLevel.intValue()) {
-          flushBuffer = buffer;
-          flushWriteOptions = writeOptions;
-          buffer = new LinkedList<>();
-        }
+      if (entry != null) {
+        write(entry, writeOptions);
```
garrettjonesgoogle: We are losing the ability to trigger a flush if we encounter a log at the configured flush level. I realize that the existing implementation doesn't actually flush, but I also don't want to toss this feature away. Add a note to #1795 about this.

pongad: Done

PR #1799, opened 2017-03-23
Bump the storage API version

rybosome:
Increases the storage API version from v1-rev85-1.22.0 to v1-rev100-1.22.0


garrettjonesgoogle:
Looks good, I'll merge after tests pass.

PR #1806, opened 2017-03-23
force query results to become strongly consistent

neozwu:
Outside transactions, global queries of google cloud datastore utilize eventually consistent model - this means query results are not guaranteed to be updated. To facilitate deterministic testing, a helper method is implemented to force strong consistency for global queries.  It leverages the fact that any queries with ancestor filter will become strongly consistent queries.

### DIFF
--- a/google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
+++ b/google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
@@ -141,7 +141,7 @@ public class ITDatastoreTest {
 
   private <T> Iterator<T> getStronglyConsistentResults (Query scQuery, Query query) throws InterruptedException {
     //scQuery is equivalent to query, but with an ancestor filter in it
-    //This makes scQuery strongly consistent
+    //this makes scQuery strongly consistent
     QueryResults<T> scResults = DATASTORE.run(scQuery);
     List<T> scResultsCopy = makeResultsCopy(scResults);
     Set<T> scResultsSet = new HashSet<>(scResultsCopy);


garrettjonesgoogle:
@vam-google could you do a first pass on this PR?

neozwu:
kindly ping @vam-google

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -137,6 +139,46 @@
...
+      List<T> resultsCopy = makeResultsCopy(results);
+      if (!haveSameSize(scResultsCopy.iterator(), resultsCopy.iterator())) {
+        continue;
+      }
+      for (T res: resultsCopy)  {
+        if (! scResultsSet.contains(res)) {
+          continue infiniteloop;
+        }
+      }
+      return resultsCopy.iterator();
```
vam-google: (Complementary to the codacy-bot comment).
There is no need to use labels and the "continue label" statement here, as it is the last statement in the whole method.

In this particular case the while(true) loop is used solely as a "label: [...] goto label;" construct. This can be avoided, using containsAll method (its implementation does exactly same thing in exactly same way):
```
while(true) {
  QueryResults<T> results = DATASTORE.run(query);
  List<T> resultsCopy = makeResultsCopy(results);
  if (!haveSameSize(scResultsCopy.iterator(), resultsCopy.iterator())) {
    continue;
  }
  if(csResultsSet.containsAll(resultsCopy)) {
    return resultsCopy.iterator();
  }      
}
```

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -137,6 +139,46 @@
...
+  private <T> Iterator<T> getStronglyConsistentResults (Query scQuery, Query query) {
+    QueryResults<T> scResults = DATASTORE.run(scQuery);
+    List<T> scResultsCopy = makeResultsCopy(scResults);
+    Set<T> scResultsSet = new HashSet<>(scResultsCopy);
+
+    infiniteloop:
+    while(true) {
+      QueryResults<T> results = DATASTORE.run(query);
+      List<T> resultsCopy = makeResultsCopy(results);
+      if (!haveSameSize(scResultsCopy.iterator(), resultsCopy.iterator())) {
```
vam-google: This method takes linear time to compare two collections, when it is not necessary here as both scResultsCopy and resultsCopy are lists. Please replace it with just constant time and simpler to understand version:
```
if(csResultsCopy.size() != resultsCopy.size()) {
    continue;
}
```

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -137,6 +139,46 @@
   private static final Entity ENTITY3 = Entity.newBuilder(ENTITY1).setKey(KEY3).remove("str")
       .set("null", NULL_VALUE).set("partial1", PARTIAL_ENTITY2).set("partial2", ENTITY2).build();
 
+  private <T> Iterator<T> getStronglyConsistentResults (Query scQuery, Query query) {
+    QueryResults<T> scResults = DATASTORE.run(scQuery);
+    List<T> scResultsCopy = makeResultsCopy(scResults);
+    Set<T> scResultsSet = new HashSet<>(scResultsCopy);
```
vam-google: This will work only if the query returns only distinct results. Is it guaranteed to return distinct result items all the time? This is also under assumption, that equals and hascode are properly defined fro the result items.

Please (at least) provide a comment mentioning that the items in the result must be unique.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -137,6 +139,46 @@
...
+  private <T> List<T> makeResultsCopy(QueryResults<T> scResults) {
+    Preconditions.checkNotNull(scResults);
+    List<T> results = new ArrayList<>();
+    while (scResults.hasNext()) {
+      results.add(scResults.next());
+    }
+    return results;
+  }
+
+  private <T, S> boolean haveSameSize(Iterator<T> it1, Iterator<S> it2) {
```
vam-google: Please remove this method as we can avoid using it.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -341,12 +373,14 @@ public void testRunGqlQueryNoCasting() throws InterruptedException {
     Query<Entity> query1 = Query.newGqlQueryBuilder(ResultType.ENTITY, "select * from " + KIND1)
         .setNamespace(NAMESPACE)
         .build();
-    QueryResults<Entity> results1 = DATASTORE.run(query1);
-    while (Iterators.size(results1) < 1) {
-      Thread.sleep(500);
-      results1 = DATASTORE.run(query1);
-    }
-    results1 = DATASTORE.run(query1);
+    Query<Entity> scQuery1 = Query.newEntityQueryBuilder()
```
vam-google: I assume this means that queries built using .newEntityQueryBuilder are guaranteed to be strongly consistent. Is it correct? This feels strange. Can you please provide a comment here, describing why it is strongly consistent (like documentation link, where it is stated).

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -137,6 +139,46 @@
   private static final Entity ENTITY3 = Entity.newBuilder(ENTITY1).setKey(KEY3).remove("str")
       .set("null", NULL_VALUE).set("partial1", PARTIAL_ENTITY2).set("partial2", ENTITY2).build();
 
+  private <T> Iterator<T> getStronglyConsistentResults (Query scQuery, Query query) {
+    QueryResults<T> scResults = DATASTORE.run(scQuery);
+    List<T> scResultsCopy = makeResultsCopy(scResults);
+    Set<T> scResultsSet = new HashSet<>(scResultsCopy);
+
+    infiniteloop:
+    while(true) {
```
vam-google: in case of any kind of unexpected problem this potentially can make the loop never succeed. Please consider putting maximum allowed attempts here, and throwing an exception if the limit is reached (so there will be no hanged tests).

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -137,6 +139,46 @@
   private static final Entity ENTITY3 = Entity.newBuilder(ENTITY1).setKey(KEY3).remove("str")
       .set("null", NULL_VALUE).set("partial1", PARTIAL_ENTITY2).set("partial2", ENTITY2).build();
 
+  private <T> Iterator<T> getStronglyConsistentResults (Query scQuery, Query query) {
```
vam-google: Please strongly consider adding a sleep interval between attempts (as if something is not consistent now, it is most likely to remain inconsistent if the next attempt is performed too quickly from now, and hitting the service too often potentially may lead to another problem - exceeding quotas).

neozwu: 500 millisecond sleep is add.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
neozwu: refactored the code to avoid using labels.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
neozwu: refactored the code to use sets to avoid unnecessary complexity.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
neozwu: The newEntityQueryBuilder() itself doesn't not guarantee strongly consistent query. It is the "ancestor filter" that makes the query an ancestor query, which is guaranteed to be strongly consistent.  A comment is added to make this clear.

neozwu:
@vam-google PTAL

google-cloud-datastore/src/test/java/com/google/cloud/datastore/it/ITDatastoreTest.java
```diff
@@ -137,6 +139,38 @@
   private static final Entity ENTITY3 = Entity.newBuilder(ENTITY1).setKey(KEY3).remove("str")
       .set("null", NULL_VALUE).set("partial1", PARTIAL_ENTITY2).set("partial2", ENTITY2).build();
 
+  private <T> Iterator<T> getStronglyConsistentResults (Query scQuery, Query query) throws InterruptedException {
+    //scQuery is equivalent to query, but with an ancestor filter in it
+    //This makes scQuery strongly consistent
```
vam-google: This -> this, or put a "." at the end of the previous line.

PR #1815, opened 2017-03-28
logging: make flush wait for writes

pongad:
Updates 1795.

This PR still isn't completely correct,
since it does not force any RPC to immediately be issued.
However, flush should now correctly wait for RPCs
representing prior calls to publish to complete
and any failures to be reported to ErrorManager
before returning.

This PR is WIP due to lack of tests.
They are forth-coming,
but all existing tests pass.

Fun fact:
I wrote this PR three times,
using `synchronized`, `ReentrantLock`, then `Monitor`
mirroring the Monitor class's documentation:
http://google.github.io/guava/releases/21.0/api/docs/com/google/common/util/concurrent/Monitor.html

### DIFF
--- a/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
+++ b/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.logging;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
 
 import com.google.api.gax.core.ApiFutures;
 import com.google.api.gax.core.SettableApiFuture;
@@ -408,12 +409,13 @@ public class LoggingHandlerTest {
     flushWaiter.start();
 
     // flushWaiter should be waiting for mockRpc to complete.
-    flushWaiter.join(100);
+    flushWaiter.join(1000);
     assertTrue(flushWaiter.isAlive());
 
     // With the RPC completed, flush should return, and the thread should terminate.
     mockRpc.set(null);
-    flushWaiter.join();
+    flushWaiter.join(1000);
+    assertFalse(flushWaiter.isAlive());
   }
 
   @Test


google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -457,28 +500,31 @@ void write(LogEntry entry, WriteOption... options) {
...
-          @Override
-          public void onFailure(Throwable t) {
-            if (t instanceof Exception) {
-              reportError(null, (Exception) t, ErrorManager.FLUSH_FAILURE);
-            } else {
-              reportError(null, new Exception(t), ErrorManager.FLUSH_FAILURE);
-            }
-          }
-        });
+        flushMonitor.enterWhenUninterruptibly(noFlushers);
```
michaelbausor: It looks like when `flush()` is called, all new writes are blocked until the flush is completed. Is that correct?

pongad: This is correct. Otherwise, new writes will keep incrementing the `numPendingWrites` and flushes might never complete. There are ways around it, but I think this is fine for now, since flushes should happen relatively rarely. (Or so it is hoped anyway.)

michaelbausor: I would be worried because we call `flush()` inside `publish()` when the logging level is higher than the flush threshold. So if that happens, any log writes on other threads will block until flush completes.

pongad:
@michaelbausor @garrettjonesgoogle PTAL

pongad:
@michaelbausor @garrettjonesgoogle Actually, please hold off. I'm no longer convinced this is right.

Consider the following events:
- A publish RPC is sent, its future recorded in the table
- The RPC fails. We invoke a callback, roughly calling `callback.onFailure(throwable)`.
- Since we only need the callback and the throwable, the future itself might not be needed. The future gets GCed
- Someone calls flush, since the future is already GCed, flush returns.
- **Then**, the callback calls `reportError`.

Thus, the flush might return before the error could be reported.

pongad:
@michaelbausor @garrettjonesgoogle I hope this is right now. Still needs tests and force-RPC-to-fire thing.

michaelbausor:
Looks good!

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -457,28 +469,64 @@ void write(LogEntry entry, WriteOption... options) {
...
+              @Override
+              public void onFailure(Throwable t) {
+                try {
+                  if (t instanceof Exception) {
+                    reportError(null, (Exception) t, ErrorManager.FLUSH_FAILURE);
+                  } else {
+                    reportError(null, new Exception(t), ErrorManager.FLUSH_FAILURE);
+                  }
+                } finally {
+                  onSuccess(null);
```
garrettjonesgoogle: Calling `onSuccess` from `onFailure` could be a hazard for future maintenance - it looks correct now, but anything added to `onSuccess` later would affect `onFailure` too, which could be very surprising. I would prefer both methods to call a shared separate function, or just duplicate the code.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -457,28 +469,64 @@ void write(LogEntry entry, WriteOption... options) {
...
-            }
-          }
-        });
+        final ApiFuture<Void> writeFuture = getLogging().writeAsync(entryList, options);
+        writeLock.lock();
+        try {
+          pendingWrites.add(writeFuture);
+        } finally {
+          writeLock.unlock();
+        }
```
garrettjonesgoogle: I don't get the impression that there is an advantage to using `ReentrantLock` over a simple `synchronized` block for protecting `pendingWrites`. Using a `synchronized` block would look much simpler. (Synchronize on an internal lock object, not `this`.)

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -457,28 +469,64 @@ void write(LogEntry entry, WriteOption... options) {
...
+            });
         break;
     }
   }
 
   @Override
   public void flush() {
     // BUG(1795): flush is broken, need support from batching implementation.
+
+    ArrayList<ApiFuture<Void>> writes = new ArrayList<>();
```
garrettjonesgoogle: maybe call this `writesToFlush`

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
pongad: Done. I made `writeLock` an Object and used `synchronized(writeLock)`. We can also just use `synchronized(pendingWrites)`. Do you have an opinion?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
pongad: Moved to shared function

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
pongad: Done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -457,28 +500,31 @@ void write(LogEntry entry, WriteOption... options) {
...
+        } finally {
+          flushMonitor.leave();
+        }
         break;
     }
   }
 
   @Override
   public void flush() {
     // BUG(1795): flush is broken, need support from batching implementation.
```
pongad: Still accurate but misleading. Fixed.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
```diff
@@ -381,6 +383,40 @@ public void testFlushLevel() {
...
+    });
+    flushWaiter.start();
+
+    // flushWaiter should be waiting for mockRpc to complete.
+    flushWaiter.join(100);
+    assertTrue(flushWaiter.isAlive());
+
+    // With the RPC completed, flush should return, and the thread should terminate.
+    mockRpc.set(null);
+    flushWaiter.join();
```
garrettjonesgoogle: Maybe add a timeout so that the test doesn't lock up forever given an implementation bug.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
```diff
@@ -381,6 +383,40 @@ public void testFlushLevel() {
...
+    Thread flushWaiter = new Thread(new Runnable() {
+      @Override
+      public void run() {
+        handler.flush();
+      }
+    });
+    flushWaiter.start();
+
+    // flushWaiter should be waiting for mockRpc to complete.
+    flushWaiter.join(100);
```
garrettjonesgoogle: Will this be flaky with a timeout this short (100 ms)?

pongad: If the timeout is too short, there shouldn't be a false failure but a false success. Increased to 1 sec.

google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
pongad: Good idea, done.

pongad:
@garrettjonesgoogle PTAL

PR #1817, opened 2017-03-28
Update spi classes

shinfan:


garrettjonesgoogle:
What about Pub/Sub?

shinfan:
@garrettjonesgoogle Pubsub involves a lot more changes (due to the proto update) than other APIs so I think it would be cleaner to have a separate PR. But I am ok including it in this PR as well. Do you have a preference?

garrettjonesgoogle:
I was only looking for your versioning changes to show up in pubsub, but it looks like pubsub wasn't touched at all here, and will retain the old version flaw.

shinfan:
As discussed offline @michaelbausor will generate the pubsub with new protos. I will rebase this PR on top of his.

shinfan:
Rebased with pubsub update. PTAL.

garrettjonesgoogle:
LGTM

shinfan:
AppVeyor tests passed. Going to merge.

PR #1818, opened 2017-03-28
Pubsub update

michaelbausor:
Refresh GAPIC layer for all APIs; update pubsub with new methods; update pubsub with renamed client classes


garrettjonesgoogle:
LGTM after tests pass

shinfan:
LGTM

PR #1819, opened 2017-03-28
Full support for DATE, TIME, and DATETIME in BigQuery

### LINKED ISSUE #1792: BigQuery: Support Date, Time, and DateTime types in FieldValue
Use types from threetenbp.

garrettjonesgoogle:
Also:

* Using Timestamp instead of Long for timestamp values.
* Renaming Date to CivilDate, naming new types in an analogous way, CivilTime and CivilDateTime

Fixes #1792

### DIFF
--- a/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/BaseGrpcServiceException.java
+++ b/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/BaseGrpcServiceException.java
@@ -17,6 +17,7 @@
 package com.google.cloud.grpc;
 
 import com.google.api.client.http.HttpResponseException;
+import com.google.api.core.InternalApi;
 import com.google.api.gax.grpc.ApiException;
 import com.google.cloud.BaseServiceException;
 import com.google.common.base.MoreObjects;
@@ -30,12 +31,14 @@ public class BaseGrpcServiceException extends BaseServiceException {
 
   private static final long serialVersionUID = -2685197215731335549L;
 
-  public BaseGrpcServiceException(String message, Throwable cause, int code, boolean retryable) {
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BaseGrpcServiceException(String message, Throwable cause, int code, boolean retryable) {
     super(ExceptionData.newBuilder().setMessage(message).setCause(cause).setCode(code)
         .setRetryable(retryable).build());
   }
 
-  public BaseGrpcServiceException(IOException exception, boolean idempotent) {
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BaseGrpcServiceException(IOException exception, boolean idempotent) {
     super(makeExceptionData(exception, idempotent));
   }
 
--- a/google-cloud-core-http/src/main/java/com/google/cloud/http/BaseHttpServiceException.java
+++ b/google-cloud-core-http/src/main/java/com/google/cloud/http/BaseHttpServiceException.java
@@ -19,6 +19,7 @@ package com.google.cloud.http;
 import com.google.api.client.googleapis.json.GoogleJsonError;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.client.http.HttpResponseException;
+import com.google.api.core.InternalApi;
 import com.google.cloud.BaseServiceException;
 import com.google.common.base.MoreObjects;
 import java.io.IOException;
@@ -32,7 +33,8 @@ public class BaseHttpServiceException extends BaseServiceException {
   private static final long serialVersionUID = -5793034110344127954L;
   public static final int UNKNOWN_CODE = 0;
 
-  public BaseHttpServiceException(IOException exception, boolean idempotent,
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BaseHttpServiceException(IOException exception, boolean idempotent,
       Set<BaseServiceException.Error> retryableErrors) {
     super(makeExceptionData(exception, idempotent, retryableErrors));
   }
@@ -80,7 +82,8 @@ public class BaseHttpServiceException extends BaseServiceException {
         .build();
   }
 
-  public BaseHttpServiceException(GoogleJsonError googleJsonError, boolean idempotent,
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BaseHttpServiceException(GoogleJsonError googleJsonError, boolean idempotent,
       Set<BaseServiceException.Error> retryableErrors) {
     super(makeExceptionData(googleJsonError, idempotent, retryableErrors));
   }
@@ -108,12 +111,14 @@ public class BaseHttpServiceException extends BaseServiceException {
     return exceptionData.build();
   }
 
-  public BaseHttpServiceException(int code, String message, String reason, boolean idempotent,
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BaseHttpServiceException(int code, String message, String reason, boolean idempotent,
       Set<BaseServiceException.Error> retryableErrors) {
     this(code, message, reason, idempotent, retryableErrors, null);
   }
 
-  public BaseHttpServiceException(int code, String message, String reason, boolean idempotent,
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BaseHttpServiceException(int code, String message, String reason, boolean idempotent,
       Set<BaseServiceException.Error> retryableErrors, Throwable cause) {
     super(ExceptionData.newBuilder()
         .setMessage(message)
--- a/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
+++ b/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
@@ -24,6 +24,7 @@ import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.core.InternalApi;
 import com.google.auth.Credentials;
 import com.google.auth.http.HttpCredentialsAdapter;
 import com.google.auth.http.HttpTransportFactory;
@@ -122,7 +123,7 @@ public class HttpTransportOptions implements TransportOptions {
     }
   }
 
-  protected HttpTransportOptions(Builder builder) {
+  private HttpTransportOptions(Builder builder) {
     httpTransportFactory = firstNonNull(builder.httpTransportFactory,
         ServiceOptions.getFromServiceLoader(HttpTransportFactory.class,
             DefaultHttpTransportFactory.INSTANCE));
--- a/google-cloud-core/pom.xml
+++ b/google-cloud-core/pom.xml
@@ -48,7 +48,6 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>api-common</artifactId>
-      <version>${api-common.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.api</groupId>
--- a/google-cloud-core/src/main/java/com/google/cloud/BaseService.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/BaseService.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud;
 
+import com.google.api.core.InternalApi;
 import com.google.cloud.ExceptionHandler.Interceptor;
 
 /**
@@ -51,6 +52,7 @@ public abstract class BaseService<OptionsT extends ServiceOptions<?, OptionsT>>
 
   private final OptionsT options;
 
+  @InternalApi("This class should only be extended within google-cloud-java")
   protected BaseService(OptionsT options) {
     this.options = options;
   }
--- a/google-cloud-core/src/main/java/com/google/cloud/BaseServiceException.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/BaseServiceException.java
@@ -253,7 +253,8 @@ public class BaseServiceException extends RuntimeException {
     }
   }
 
-  public BaseServiceException(ExceptionData exceptionData) {
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BaseServiceException(ExceptionData exceptionData) {
     super(exceptionData.getMessage(), exceptionData.getCause());
     this.code = exceptionData.getCode();
     this.reason = exceptionData.getReason();
--- a/google-cloud-core/src/main/java/com/google/cloud/BaseWriteChannel.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/BaseWriteChannel.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud;
 
+import com.google.api.core.InternalApi;
 import java.io.IOException;
 import java.io.Serializable;
 import java.nio.ByteBuffer;
@@ -107,6 +108,7 @@ public abstract class BaseWriteChannel<
     this.chunkSize = Math.max(getMinChunkSize(), chunkSize);
   }
 
+  @InternalApi("This class should only be extended within google-cloud-java")
   protected BaseWriteChannel(ServiceOptionsT options, EntityT entity, String uploadId) {
     this.options = options;
     this.entity = entity;
@@ -209,6 +211,7 @@ public abstract class BaseWriteChannel<
     protected final boolean isOpen;
     protected final int chunkSize;
 
+    @InternalApi("This class should only be extended within google-cloud-java")
     protected BaseState(Builder<ServiceOptionsT, EntityT> builder) {
       this.serviceOptions = builder.serviceOptions;
       this.entity = builder.entity;
@@ -238,6 +241,7 @@ public abstract class BaseWriteChannel<
       private boolean isOpen;
       private int chunkSize;
 
+      @InternalApi("This class should only be extended within google-cloud-java")
       protected Builder(ServiceOptionsT options, EntityT entity, String uploadId) {
         this.serviceOptions = options;
         this.entity = entity;
--- a/google-cloud-core/src/main/java/com/google/cloud/BatchResult.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/BatchResult.java
@@ -19,6 +19,7 @@ package com.google.cloud;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
+import com.google.api.core.InternalApi;
 import java.util.LinkedList;
 import java.util.List;
 
@@ -35,6 +36,10 @@ public abstract class BatchResult<T, E extends BaseServiceException> {
   private boolean completed = false;
   private E error;
   private final List<Callback<T, E>> toBeNotified = new LinkedList<>();
+
+  @InternalApi("This class should only be extended within google-cloud-java")
+  protected BatchResult() {
+  }
 
   /**
    * Returns {@code true} if the batch has been completed and the result is available; {@code false}
--- a/google-cloud-core/src/main/java/com/google/cloud/ByteArray.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ByteArray.java
@@ -40,12 +40,8 @@ public class ByteArray implements Iterable<Byte>, Serializable {
 
   private final ByteString byteString;
 
-  protected ByteArray(ByteString byteString) {
+  ByteArray(ByteString byteString) {
     this.byteString = byteString;
-  }
-
-  protected ByteArray(ByteArray byteArray) {
-    this.byteString = byteArray.getByteString();
   }
 
   @Override
@@ -115,11 +111,6 @@ public class ByteArray implements Iterable<Byte>, Serializable {
    */
   public final InputStream asInputStream() {
     return byteString.newInput();
-  }
-
-
-  protected ByteString getByteString() {
-    return byteString;
   }
 
   /**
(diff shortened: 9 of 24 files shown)


garrettjonesgoogle:
Note I have fixed the codacy errors - I'll push them with other PR updates.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -245,20 +251,43 @@ public static QueryParameterValue date(String value) {
...
+  /**
+   * Creates a {@code QueryParameterValue} object with a type of TIME.
+   */
+  public static QueryParameterValue time(CivilTime value) {
+    return of(value, StandardSQLTypeName.TIME);
+  }
+
+  /**
+   * Creates a {@code QueryParameterValue} object with a type of DATETIME.
+   * Must be in the format "yyyy-MM-dd HH:mm:ss.SSSSSS", e.g. ""2014-08-19 12:41:35.220000".
```
shinfan: remove the extra quote

google-cloud-core/src/main/java/com/google/cloud/CivilDate.java
```diff
@@ -22,17 +22,28 @@
...
 /**
- * Represents a Date without time, such as 2017-03-17. Date is timezone independent.
+ * Represents a date without time or timezone, e.g. 2017-03-17.
+ *
+ * CivilDate holds values for BigQuery DATE
+ * (https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#date-type)
+ * and Spanner DATE
+ * (https://cloud.google.com/spanner/docs/data-types#date-type).
+ *
+ * The DATE type represents a logical calendar date, independent of time zone. A DATE value
```
shinfan: Update the comment

google-cloud-core/src/main/java/com/google/cloud/CivilTime.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Represents a time without timezone, e.g. 07:41:35.220000.
+ *
+ * CivilTime holds values for BigQuery TIME
+ * (https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#time-type).0
```
shinfan: .0 ?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -330,32 +359,40 @@ public static QueryParameterValue dateTime(String value) {
...
       case TIMESTAMP:
-        if (value instanceof Long) {
-          return timestampFormatter.print(((Long) value) / 1000);
+        if (value instanceof Timestamp) {
+          return ((Timestamp) value).formatMicroCloudTime();
+        } else if (value instanceof Long) {
+          return Timestamp.ofTimeMicroseconds((Long) value).formatMicroCloudTime();
         } else if (value instanceof String) {
           // verify that the String is in the right format
-          timestampFormatter.parseMillis((String) value);
```
shinfan: Can we remove the xxxFormatter now? The functionality seems overlapping with the new date classes.

garrettjonesgoogle: it *is* removed...

google-cloud-core/src/main/java/com/google/cloud/CivilDate.java
garrettjonesgoogle: in what way?

google-cloud-core/src/main/java/com/google/cloud/CivilTime.java
garrettjonesgoogle: LOL I have no idea how that happened... I will fix it.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
garrettjonesgoogle: will do.

google-cloud-core/src/main/java/com/google/cloud/CivilDate.java
shinfan: Oh nvm I was confusing that the DATE should be replaced with CivilDate... but it seems like I misunderstood.

garrettjonesgoogle:
@shinfan PTAL

shinfan:
LGTM

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * @param date civil date
+   * @param time civil time
+   */
+  public static CivilDateTime ofDateAndTime(CivilDate date, CivilTime time) {
+    return new CivilDateTime(date, time);
+  }
+
+  /**
+   * @param dateTime Date plus time: (yyyy-MM-dd HH:mm:ss.SSSSSSSSS) or
```
vkedia: document that time here is optional by maybe putting the time part in []

google-cloud-core/src/main/java/com/google/cloud/CivilTime.java
```diff
@@ -0,0 +1,153 @@
...
+    Preconditions.checkArgument(seconds >= 0 && seconds < 60, "Invalid seconds: " + seconds);
+    Preconditions.checkArgument(nanos >= 0 && nanos <= 999_999_999, "Invalid nanos: " + nanos);
+    this.hours = hours;
+    this.minutes = minutes;
+    this.seconds = seconds;
+    this.nanos = nanos;
+  }
+
+  /**
+   * Constructs a new CivilDate instance.
```
vkedia: "new **CivilTime** instance"

google-cloud-core/src/main/java/com/google/cloud/CivilTime.java
```diff
@@ -0,0 +1,153 @@
...
+    int seconds = Integer.parseInt(matcher.group(3));
+    int nanos = 0;
+    if (matcher.groupCount() > 3) {
+      String fractional = matcher.group(4);
+      Integer fractionalNumber = Integer.parseInt(fractional);
+      if (fractional.length() == 3) {
+        nanos = (int) TimeUnit.MILLISECONDS.toNanos(fractionalNumber);
+      } else if (fractional.length() == 6) {
+        nanos = (int) TimeUnit.MICROSECONDS.toNanos(fractionalNumber);
+      } else {
```
vkedia: I think you should check for length==9 here and throw IAE in remaining cases.

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
```diff
@@ -0,0 +1,134 @@
...
+  public CivilDate getDate() {
+    return date;
+  }
+
+  /** Returns the time component of this date-time. */
+  public CivilTime getTime() {
+    return time;
+  }
+
+  /** Formats the date and time in the format yyyy-MM-dd HH:mm:ss.SSSSSSSSS */
```
vkedia: the format would be yyyy-MM-dd HH:mm:ss.**SSSSSS** (6Ss)

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
```diff
@@ -0,0 +1,134 @@
...
+  public String formatMicroCloudTime() {
```
vkedia: why formatMicroCloudTime and not formatMicroTime

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -58,6 +82,13 @@ private Timestamp(long seconds, int nanos) {
     this.nanos = nanos;
   }
 
+  public static Timestamp ofTimeMicroseconds(long microseconds) {
```
vkedia: why not just ofMicroseconds

google-cloud-core/src/main/java/com/google/cloud/CivilTime.java
vkedia: Actually we should not throw IAE but parse it correctly irrespective of the number of digits..

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
```diff
@@ -0,0 +1,134 @@
...
+    return new CivilDateTime(date, time);
+  }
+
+  /**
+   * @param dateTime Date plus time: (yyyy-MM-dd HH:mm:ss.SSSSSSSSS) or
+   * (yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS).
+   */
+  public static CivilDateTime parseDateTime(String dateTime) {
+    String splitToken = " ";
+    if (dateTime.contains("T")) {
```
vkedia: technically this can be lower case t as well.

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
```diff
@@ -0,0 +1,134 @@
...
+package com.google.cloud;
+
+import com.google.api.client.util.Lists;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Represents a date and time with no timezone, e.g. e.g. 2017-03-17 07:41:35.220000.
```
vkedia: Maybe use T in the example instead of " " since toString outputs using T

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
garrettjonesgoogle: It allows any number of digits up to 9 of them.

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
garrettjonesgoogle: My thought was that since it uses a space instead of a 'T', that should be distinguished somehow, thus the "Cloud" part.

google-cloud-core/src/main/java/com/google/cloud/CivilTime.java
garrettjonesgoogle: This was an implementation I wrote before I encountered the superior approach in Timestamp, but I forgot to circle back. I will fix it.

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: Because of the existing `ofTimeSecondsAndNanos`. I can go ahead and remove `Time` from both.

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
garrettjonesgoogle: ack you're right. Will fix.

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -43,12 +47,32 @@
       new Timestamp(253402300799L, (int) TimeUnit.SECONDS.toNanos(1) - 1);
 
   /** Regexp to split timestamps into date-hour-minute-second and fractional second components. */
-  private static final Pattern FORMAT_REGEXP = Pattern.compile("([^\\.]*)(\\.\\d{0,9})?Z");
+  private static final Pattern FORMAT_REGEXP = Pattern.compile("([^.]*)(\\.\\d{0,9})?Z");
+  private static final String DATE_GROUP = "(\\d{4}-\\d{1,2}-\\d{1,2})";
```
vkedia: why do you want to allow single digit months,dates and the various time parts below?
According to the RFC they have to be 2 digits
https://www.ietf.org/rfc/rfc3339.txt
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -43,12 +47,32 @@
...
+  private static final String TIME_GROUP = "(\\d{1,2}:\\d{1,2}:\\d{1,2})";
+  private static final String FRACT_GROUP = "\\.(\\d{1,9})";
+  private static final String TZ_GROUP = "([\\-+]\\d{1,2}(?::\\d{1,2})?)";
```
vkedia: Why "?::"

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -137,22 +169,119 @@ public static Timestamp parseTimestamp(String timestamp) {
...
-      String padded = Strings.padEnd(nanosPart.substring(1), 9, '0');
-      nanos = Integer.parseInt(padded);
+      nanos = parseFractToNanos(nanosPart.substring(1));
+    }
+    return ofTimeSecondsAndNanos(seconds, nanos);
+  }
+
+  /**
+   * Creates a Timestamp instance from the given string. String is in the format
+   * yyyy-MM-dd HH:mm:ss.SSSSSSZZ. If no timezone is given, UTC is assumed.
```
vkedia: Should it be 9Ss? Is ZZ the correct format specifier?

google-cloud-core/src/main/java/com/google/cloud/CivilTime.java
garrettjonesgoogle: Will fix.

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
```diff
@@ -0,0 +1,134 @@
...
+    this.time = time;
+  }
+
+  /**
+   * Constructs a new CivilDate instance with time set to the beginning of the day.
+   *
+   * @param date civil date
+   */
+  public static CivilDateTime ofDate(CivilDate date) {
+    return ofDateAndTime(date, CivilTime.ofHoursMinutesSecondsNanos(0, 0, 0, 0));
```
garrettjonesgoogle: Will do.

google-cloud-core/src/main/java/com/google/cloud/CivilDateTime.java
garrettjonesgoogle: Will do.

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -137,22 +169,119 @@ public static Timestamp parseTimestamp(String timestamp) {
...
+          tz = dateTimeFractMatcher.group(4);
+        } else {
+          throw new IllegalArgumentException("Cannot parse input: " + timestamp);
+        }
+      }
+    }
+
+    DateTime dateTime = new DateTime();
+    dateTime = parseTz(dateTime, tz);
+    dateTime = parseDate(dateTime, Preconditions.checkNotNull(date));
```
vkedia: checkNotNull here seems like Defensive programing!!

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -137,22 +169,119 @@ public static Timestamp parseTimestamp(String timestamp) {
...
+    dateTime = parseTime(dateTime, time);
+    int nanos = parseFractToNanos(fract);
+    return ofTimeSecondsAndNanos(TimeUnit.MILLISECONDS.toSeconds(dateTime.getMillis()), nanos);
+  }
+
+  private static DateTime parseDate(DateTime newDateTime, String dateString) {
```
vkedia: you can just call the variable dateTime

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -137,22 +169,119 @@ public static Timestamp parseTimestamp(String timestamp) {
...
+      nanos = parseFractToNanos(nanosPart.substring(1));
+    }
+    return ofTimeSecondsAndNanos(seconds, nanos);
+  }
+
+  /**
+   * Creates a Timestamp instance from the given string. String is in the format
+   * yyyy-MM-dd HH:mm:ss.SSSSSSZZ. If no timezone is given, UTC is assumed.
+   */
+  public static Timestamp parseCloudTimestamp(String timestamp) {
```
vkedia: This is pretty complex. Is there no existing formatter that will do this? And do we actually need to support all these variants right now?

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: According to the BigQuery docs (and Spanner docs), this doesn't appear to be RFC 3339:

https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#canonical-format_3

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: `(?:` starts a non-capturing group; the second `:` is a literal match.

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: Yes, 9 S's. I will expand the explanation about the time zone format.

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: Two things can't be supported with existing date parsers for Java 7:

1. The need to parse microsecond precision
2. The need to support time zones with a colon

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: That was a check that came along for the ride when I moved a bunch of code around. I will remove it.

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: Agreed. This is another thing that came along for the ride...

garrettjonesgoogle:
All so-far-agreed-to feedback has been addressed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -245,20 +248,43 @@ public static QueryParameterValue date(String value) {
...
+   */
+  public static QueryParameterValue date(CivilDate value) {
+    return of(value, StandardSQLTypeName.DATE);
+  }
+
+  /**
    * Creates a {@code QueryParameterValue} object with a type of TIME. Must be in the format
    * "HH:mm:ss.SSSSSS", e.g. "12:41:35.220000".
    */
   public static QueryParameterValue time(String value) {
```
vkedia: Do you still want to keep this variants around that take in String?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
```diff
@@ -189,19 +192,47 @@ public boolean getBooleanValue() {
...
    *
    * @throws ClassCastException if the field is not a primitive type
-   * @throws NumberFormatException if the field's value could not be converted to {@link Long}
+   * @throws NumberFormatException if the field's value could not be converted to a
+   *   {@link Timestamp}
    * @throws NullPointerException if {@link #isNull()} returns {@code true}
    */
   @SuppressWarnings("unchecked")
-  public long getTimestampValue() {
+  public Timestamp getTimestampValue() {
```
vkedia: Might be worth documenting that this Timestamp is micro precision and not nano

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
vkedia: So can you split on '.' and use an existing parser for the part before '.'. Similar to parseTimestamp?

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
vkedia: Technically we don't have to adhere to that here. We can be loser or stricter than that as long as we make sure that while calling the bigquery rpc we pass strings of the correct format. But its fine.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
garrettjonesgoogle: I think I do. Sometimes time in string format might be passed through from another source, and it can be really convenient for testing purposes (it looks a lot simpler).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
garrettjonesgoogle: Good idea, I'll do that.

garrettjonesgoogle:
I have pushed some doc improvements, PTAL

garrettjonesgoogle:
@vkedia do you have any other concerns?

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
vkedia: This comment is not yet addressed but I dont care a lot about this so up to you.

vkedia:
Looks good to me.

google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
garrettjonesgoogle: Ahh I missed this response. The problem is that there are two possible things before the '.':

1. Date only
2. Date and time

I would have to use two separate parsers like I am already doing.

PR #1820, opened 2017-03-28
Make logging overrides the default channel provider

shinfan:


google-cloud-core/src/main/java/com/google/cloud/GrpcTransportOptions.java
```diff
@@ -210,23 +210,23 @@ private GrpcTransportOptions(Builder builder) {
   }
 
   /**
-   * Returns a channel provider.
+   * Returns a channel provider from the given default provider.
    */
-  public static ChannelProvider getChannelProvider(
-      ServiceOptions<?, ?> serviceOptions) {
+  public static ChannelProvider setupChannelProvider(
```
garrettjonesgoogle: the 'u' should be capitalized (`setUp`)

google-cloud-core/src/main/java/com/google/cloud/GrpcTransportOptions.java
```diff
@@ -210,23 +210,23 @@ private GrpcTransportOptions(Builder builder) {
...
+      InstantiatingChannelProvider.Builder provider, ServiceOptions<?, ?> serviceOptions) {
```
garrettjonesgoogle: Maybe call this `providerBuilder`

garrettjonesgoogle:
LGTM

PR #1822, opened 2017-03-28
temporarily disable google_api_key to test ITs

neozwu:


neozwu:
experiment done. so close the PR.

Issue #1789, opened 2017-03-22
Remove obsolete PubSub references from docs


garrettjonesgoogle:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/TESTING.md
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/README.md

grep for any others too.

PR #1823, opened 2017-03-29
pubsub: remove obsolete doc references

pongad:
This PR removes references to the deprecated code of the docs.

The emulator section is rewritten.
Example for Subscriber is missing because it still uses ChannelBuilder.
We plan to migrate it to ChannelProvider soon.

Updates #1789.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -162,14 +162,11 @@ $ gcloud beta emulators pubsub env-init
 ```
 
 3. Point your client to the emulator.
-The code below assumes that you have set the `host` and `port`
-to the host and port the emulator is on.
 ```java
 ChannelProvider channelProvider =
     // SubscriptionAdminSettings works too.
     TopicAdminSettings.defaultChannelProviderBuilder()
-        .setServiceAddress(host)
-        .setPort(port)
+        .setEndpoint(System.getenv("PUBSUB_EMULATOR_HOST"))
         .setCredentialsProvider(
             FixedCredentialsProvider.create(NoCredentials.getInstance()))
         .build();


TESTING.md
```diff
@@ -145,44 +145,34 @@ uses the `RemoteLoggingHelper` to create a metric.
...
+    TopicAdminSettings.defaultChannelProviderBuilder()
+        .setServiceAddress(host)
+        .setPort(port)
+        .setCredentialsProvider(
+            FixedCredentialsProvider.create(NoCredentials.getInstance()))
+        .build();
+TopicAdminClient topicClient = TopicAdminClient.create(
+    TopicAdminSettings.defaultBuilder().setChannelProvider(channelProvider).build());
+SubscriptionAdminClient subscriptionClient = SubscriptionAdminClient.create(
+    SubscriptionAdminSettings.defaultBuilder().setChannelProvider(channelProvider).build());
```
garrettjonesgoogle: I think it would be better to just show one client - showing multiple clients seems like a distraction.

TESTING.md
```diff
@@ -145,44 +145,34 @@ uses the `RemoteLoggingHelper` to create a metric.
...
-  ```
+3. Point your client to the emulator.
+The code below assumes that you have set the `host` and `port`
+to the host and port the emulator is on.
+```java
+ChannelProvider channelProvider =
+    // SubscriptionAdminSettings works too.
+    TopicAdminSettings.defaultChannelProviderBuilder()
+        .setServiceAddress(host)
+        .setPort(port)
```
garrettjonesgoogle: Does the emulator always default to a particular port, e.g. 8080? Could we just show `"localhost"` and `8080` as the example values?

pongad: The emulator doesn't start on any specific port. I just tried starting it twice, the first bound with 8532 and the second with 8261. It *looks* like it just binds to a free port of the kernel's choosing.

After the emulator started, running `gcloud beta emulators pubsub env-init` prints the port the emulator is bound to. I'll just document this.

TESTING.md
pongad: Done. Should we keep the `Publisher`, since the pattern for building it is different?

I'm keeping it for now. If you want it removed, please let me know.

TESTING.md
```diff
@@ -145,44 +145,39 @@ uses the `RemoteLoggingHelper` to create a metric.
...
+$ gcloud beta emulators pubsub start
+```
 
-You can test against a remote Pub/Sub emulator as well. To do this, set the `PubSubOptions` project
-endpoint to the hostname of the remote machine, like the example below.
+To determine which host/port the emulator is running on:
+```shell
+$ gcloud beta emulators pubsub env-init
+# Sample output:
+#   export PUBSUB_EMULATOR_HOST=localhost:8759
```
pongad: Should we also document that this is a valid shell command and so the user can run `$(gcloud beta emulators pubsub env-init)`? I'm leaning towards no, since the format should be familiar to Unix users. @garrettjonesgoogle ?

pongad:
@garrettjonesgoogle PTAL

TESTING.md
pongad: Ack

TESTING.md
pongad: Done.

TESTING.md
```diff
@@ -145,44 +145,39 @@ uses the `RemoteLoggingHelper` to create a metric.
...
+    TopicAdminSettings.defaultChannelProviderBuilder()
+        .setServiceAddress(host)
+        .setPort(port)
+        .setCredentialsProvider(
+            FixedCredentialsProvider.create(NoCredentials.getInstance()))
+        .build();
+TopicAdminClient topicClient = TopicAdminClient.create(
+    TopicAdminSettings.defaultBuilder().setChannelProvider(channelProvider).build());
+Publisher publisher =
+    Publisher.newBuilder(topicName).setChannelProvider(channelProvider).build();
```
pongad: At this stage, I'm somewhat reluctant to do this. I think the gain is very small: The code sample for both `Publisher` and `Subscriber` both show the `.newBuilder(...).build()`. To the user, this should be immediately clear that we provide some default values so they don't have to fill everything in.

I am a lot more concerned that changing this name will leave documentations outdated.

If you think this should absolutely be done, I think this should be in a separate PR.

pongad:
@garrettjonesgoogle PTAL

TESTING.md
garrettjonesgoogle: I think it should be done. It is a little bewildering to users in IDEs, having to remember different names for the same thing in different contexts. We are still in breaking-change mode; we'll get all documentation in perfect shape after we're fully settled. I agree about doing it in a different PR.

garrettjonesgoogle:
LGTM

pongad:
Reported the remaining item at https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1853 . Will merge this when the release is finished.

PR #1824, opened 2017-03-29
DO NOT MERGE: pubsub: make Subscriber use ApiService

### LINKED ISSUE #1761: Create ApiService in GAX and use it in Subscriber
In `Subscriber`, we currently wrap `AbstractService` from Guava. We should instead create a wrapper `ApiService` in GAX and use that, so that if we have other cases for a Service, we don't have to duplicate code.

pongad:
Fixes #1761.

Requires googleapis/gax-java#256,
wait for merge and release before merging this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,16 +45,16 @@ If you are using Maven, add this to your pom.xml file
 <dependency>
   <groupId>com.google.cloud</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.10.0-alpha</version>
+  <version>0.11.0-alpha</version>
 </dependency>
 ```
 If you are using Gradle, add this to your dependencies
 ```Groovy
-compile 'com.google.cloud:google-cloud:0.10.0-alpha'
+compile 'com.google.cloud:google-cloud:0.11.0-alpha'
 ```
 If you are using SBT, add this to your dependencies
 ```Scala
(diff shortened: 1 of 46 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -77,7 +75,7 @@
  * <p>If no credentials are provided, the {@link Subscriber} will use application default
  * credentials through {@link GoogleCredentials#getApplicationDefault}.
  */
-public class Subscriber {
+public class Subscriber extends AbstractApiService {
```
garrettjonesgoogle: I didn't look over this file very carefully - am I correct in assuming that this is mostly a revert to the version that extended Guava's `AbstractService`?

pongad: Yes that is correct.

garrettjonesgoogle:
The GAX dependency needs to be switched to 0.8.0 for this.

pongad:
Done, and now things build again!
Merging since @garrettjonesgoogle already LGTMed.

Issue #1654, opened 2017-02-23
JUL - Logging should be zero-conf for most users.


lesv:
1. JUL `java.util.logging` should be zero configuration for most users.
1. Work on all platforms: GAE Std 8 / Flex / GKE / GCE without modification.
1. See #1130 for fluentd support as well. (in GAE Flex env)

(Java 7 GAE Std / Dataproc / Dataflow is NOT Release Blocking;)

I've asked @jabubake to help you with this.

lesv:
Also it doesn't buffer in most cases; work in a fluentd environment; etc.

garrettjonesgoogle:
@lesv Can you edit the title to be in command form instead of problem-description form?

garrettjonesgoogle:
I can't quite tell from the description what needs to be done.

jabubake:
FYI, @meltsufin and team have worked on JUL support for App Engine Flex.
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-logging/src/main/java/com/google/cloud/logging/GaeFlexLoggingEnhancer.java

jabubake:
Fluentd support needs to be better understood and scoped,  JUL is now auto-configured on GCE, GKE, Flex, Standard. Will be tracked separately in #1130

PR #1847, opened 2017-04-03
Logging refactor of JUL + monitored resource construction

jabubake:
Auto detection of select resources to get JUL working across GCP environments  and refactoring of code to help support more adapters like Slf4j.
Testing across GCP in progress.
See Issue [#1654](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1654)

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
@@ -196,7 +196,7 @@ public class LoggingHandler extends Handler {
                     String.valueOf(baseLevel.intValue())))
           };
 
-      getLogging().setFlushSeverity(severityFor(flushLevel, Severity.ERROR));
+      getLogging().setFlushSeverity(severityFor(flushLevel));
       getLogging().setWriteSynchronicity(config.getSynchronicity());
 
       this.enhancers = new LinkedList<>();
@@ -283,7 +283,7 @@ public class LoggingHandler extends Handler {
     LogEntry.Builder builder =
         LogEntry.newBuilder(Payload.StringPayload.of(payload))
             .setTimestamp(record.getMillis())
-            .setSeverity(severityFor(level, Severity.DEFAULT));
+            .setSeverity(severityFor(level));
 
     if (!baseLevel.equals(level)) {
       builder
@@ -330,7 +330,7 @@ public class LoggingHandler extends Handler {
    */
   public void setFlushLevel(Level flushLevel) {
     this.flushLevel = flushLevel;
-    getLogging().setFlushSeverity(severityFor(flushLevel, Severity.ERROR));
+    getLogging().setFlushSeverity(severityFor(flushLevel));
   }
 
   /**
@@ -355,7 +355,7 @@ public class LoggingHandler extends Handler {
     logger.addHandler(handler);
   }
 
-  private static Severity severityFor(Level level, Severity defaultSeverity) {
+  private static Severity severityFor(Level level) {
     if (level instanceof LoggingLevel) {
       return ((LoggingLevel) level).getSeverity();
     }
@@ -383,7 +383,7 @@ public class LoggingHandler extends Handler {
       case 1000:
         return Severity.ERROR;
       default:
-        return defaultSeverity;
+        return Severity.DEFAULT;
     }
   }
 
--- a/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
+++ b/google-cloud-logging/src/test/java/com/google/cloud/logging/LoggingHandlerTest.java
@@ -21,17 +21,12 @@ import static org.easymock.EasyMock.expectLastCall;
 import static org.easymock.EasyMock.replay;
 import static org.easymock.EasyMock.verify;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-
 import com.google.cloud.MonitoredResource;
 import com.google.cloud.logging.LogEntry.Builder;
 import com.google.cloud.logging.Logging.WriteOption;
 import com.google.cloud.logging.Payload.StringPayload;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
 import java.util.logging.ErrorManager;
@@ -39,7 +34,6 @@ import java.util.logging.Filter;
 import java.util.logging.Formatter;
 import java.util.logging.Handler;
 import java.util.logging.Level;
-import java.util.logging.LogManager;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 import org.easymock.EasyMock;
@@ -199,7 +193,7 @@ public class LoggingHandlerTest {
 
   @Before
   public void setUp() {
-    logging = EasyMock.createStrictMock(Logging.class);
+    logging = EasyMock.createMock(Logging.class);
     options = EasyMock.createStrictMock(LoggingOptions.class);
   }
 
@@ -479,58 +473,6 @@ public class LoggingHandlerTest {
     logger.setLevel(Level.ALL);
     LoggingHandler.addHandler(logger, handler);
     logger.log(newLogRecord(Level.FINEST, MESSAGE));
-  }
-
-  @Test
-  public void testPropertiesFile() throws IOException, InterruptedException {
-    expect(options.getProjectId()).andReturn(PROJECT).anyTimes();
-    expect(options.getService()).andReturn(logging);
-    MonitoredResource monitoredResource = MonitoredResource.of(
-            "testResourceType", ImmutableMap.of("project_id", PROJECT, "enhanced", "true"));
-    LogEntry entry =
-        LogEntry.newBuilder(Payload.StringPayload.of(MESSAGE))
-            .setSeverity(Severity.DEBUG)
-            .addLabel("enhanced", "true")
-            .setTimestamp(123456789L)
-            .build();
-
-    logging.setFlushSeverity(Severity.CRITICAL);
-    expectLastCall().once();
-    logging.setWriteSynchronicity(Synchronicity.SYNC);
-    expectLastCall().once();
-    logging.write(
-        ImmutableList.of(entry),
-        WriteOption.logName("testLogName"),
-        WriteOption.resource(monitoredResource),
-        WriteOption.labels(
-            ImmutableMap.of(
-                "levelName",
-                Level.FINEST.getName(),
-                "levelValue",
-                String.valueOf(Level.FINEST.intValue()))));
-    logging.getWriteSynchronicity();
-    expectLastCall().andReturn(Synchronicity.SYNC).once();
-    replay(options, logging);
-    LogManager.getLogManager()
-        .readConfiguration(new ByteArrayInputStream(renderConfig(CONFIG_MAP)));
-    LoggingHandler handler = new LoggingHandler(null, options, monitoredResource);
-    LogRecord record = new LogRecord(Level.FINEST, MESSAGE);
-    record.setMillis(123456789L);
-    handler.publish(record);
-    assertEquals(Level.ALL, handler.getLevel());
-    assertNotNull(handler.getFilter());
-
-    assertEquals(
-        "com.google.cloud.logging.LoggingHandlerTest$TestFilter",
-        handler.getFilter().getClass().getName());
-    assertNotNull(handler.getFormatter());
-    assertEquals(
-        "com.google.cloud.logging.LoggingHandlerTest$TestFormatter",
-        handler.getFormatter().getClass().getName());
-    assertEquals(LoggingLevel.CRITICAL, handler.getFlushLevel());
-    assertEquals(Synchronicity.SYNC, handler.getSynchronicity());
-
-    LogManager.getLogManager().reset();
   }
 
   @Test


google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceHelper.java
```diff
@@ -0,0 +1,132 @@
...
+        value = getAppEngineInstanceName();
+        break;
+      case "zone":
+        value = MetadataConfig.getZone();
+        break;
+      case "instance_id":
+        value = MetadataConfig.getInstanceId();
+        break;
+      case "cluster_name":
+        value = MetadataConfig.getClusterName();
```
duchuyctlk: Do you miss a `break` here? IMO without it, value will be null when `name` is equal to "cluster_name"

jabubake: good catch, tx. fixed.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -174,6 +174,7 @@ public LoggingHandler(String log, LoggingOptions options, MonitoredResource moni
       this.flushLevel = config.getFlushLevel();
       String logName = firstNonNull(log, config.getLogName());
 
+      LoggingOptions loggingOptions = (options != null) ? options : LoggingOptions.getDefaultInstance();
```
duchuyctlk: These four lines could be simplified as:
```java
if (options == null) {
    options = LoggingOptions.getDefaultInstance();
}
LoggingOptions loggingOptions = options;
```

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -107,25 +101,17 @@
...
-  private final LoggingOptions options;
-  private final WriteOption[] writeOptions;
-  private volatile Logging logging;
-  private Level flushLevel;
-  private long flushSize;
-  private Synchronicity synchronicity;
-  private final List<Enhancer> enhancers;
+  private LoggingHelper loggingHelper;
+  private List<Enhancer> enhancers;
+  private  ErrorHandler errorHandler;
```
michaelbausor: Nit: remove extra space

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -174,45 +157,47 @@ public LoggingHandler(String log, LoggingOptions options, MonitoredResource moni
...
+      setFormatter(config.getFormatter());
+      Level level = config.getLogLevel();
       setLevel(level);
       baseLevel = level.equals(Level.ALL) ? Level.FINEST : level;
+      this.flushLevel = config.getFlushLevel();
+      String logName = firstNonNull(log, config.getLogName());
+
+      LoggingOptions loggingOptions = (options != null) ? options : LoggingOptions.getDefaultInstance();
+      if (options == null) {
+        options = LoggingOptions.getDefaultInstance();
```
michaelbausor: Why do we need both loggingOptions and options? Won't they always be the same?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -174,45 +157,47 @@ public LoggingHandler(String log, LoggingOptions options, MonitoredResource moni
...
+      errorHandler = new ErrorHandler();
+      setFilter(config.getFilter());
+      setFormatter(config.getFormatter());
+      Level level = config.getLogLevel();
       setLevel(level);
       baseLevel = level.equals(Level.ALL) ? Level.FINEST : level;
+      this.flushLevel = config.getFlushLevel();
+      String logName = firstNonNull(log, config.getLogName());
+
+      LoggingOptions loggingOptions = (options != null) ? options : LoggingOptions.getDefaultInstance();
```
michaelbausor: We can use `firstNonNull` here, and in other places

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -250,182 +235,100 @@ private static boolean hasLoggingHandler(Logger logger) {
...
       return;
     }
-
-    if (inPublishCall.get() != null) {
-      // ignore all logs generated in the course of logging through this handler
+    LogEntry.Builder logEntryBuilder;
+    try {
+      logEntryBuilder = logEntryBuilderFor(record);
+    } catch (Exception ex) {
+      errorHandler.handleFormatError(ex);
```
michaelbausor: Isn't this handled at line 283?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -250,182 +235,100 @@ private static boolean hasLoggingHandler(Logger logger) {
...
+
+  /**
+   * Closes the handler and the associated {@link LoggingHelper} object.
+   */
+  @Override
+  public synchronized void close() throws SecurityException {
+      loggingHelper.close();
+  }
+
+  public synchronized void setFlushLevel(Level level) {
```
michaelbausor: We are losing the comments on this and similar methods

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
jabubake: I missed that when i rewrote some of this, thanks for catching, updated.

jabubake:
@michaelbausor @garrettjonesgoogle ran the Google java formatter. 
I renamed the class to LoggingService : its a wrapper class, if you have a preferred name suggestion, let me know.

pongad:
Codacy seems to be really mad at you, but the flushing code looks fine (at least not any worse than it was before :laughing: ). Lots of unit test failures though, are you setting default properties correctly?

meltsufin:
@gregw PTAL

google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
```diff
@@ -0,0 +1,76 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+public class MetadataConfig {
```
garrettjonesgoogle: @jabubake make a private constructor to address this.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,132 @@
...
+  /* Return a self-configured monitored resource.
+   */
+  static MonitoredResource getResource(String projectId, String resourceType) {
+    if (Strings.isNullOrEmpty(resourceType)) {
+      resourceType = getAutoDetectedResourceType();
+    }
+    // Currently, "gae_app" is the supported logging resource type, but we distinguish
+    // between "gae_app_flex", "gae_app_standard" to support zone id, instance id logging on flex VMs.
+    // This method trims "gae_app_flex", "gae_app_standard" to "gae_app"
+    String resourceName = (resourceType.startsWith("gae_app")) ? "gae_app" : resourceType;
```
garrettjonesgoogle: @jabubake remove the parentheses as Codacy suggests.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,132 @@
...
+      case "instance_name":
+        value = getAppEngineInstanceName();
+        break;
+      case "zone":
+        value = MetadataConfig.getZone();
+        break;
+      case "instance_id":
+        value = MetadataConfig.getInstanceId();
+        break;
+      case "cluster_name":
```
garrettjonesgoogle: I think you could turn all of these strings into an enum.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,132 @@
...
+        break;
+      case "gae_app_standard":
+        labels = new String[] {"module_id", "version_id"};
+        break;
+      case "container":
+        labels = new String[] {"cluster_name", "zone"};
+        break;
+      case "gce_instance":
+        labels = new String[] {"instance_id", "zone"};
+        break;
```
garrettjonesgoogle: I think I would prefer pre-populating a map of resource type -> string[], and then doing a get() in this function.

google-cloud-logging/src/main/java/com/google/cloud/logging/TraceEnhancer.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+/*
+ */
+public class TraceEnhancer implements Enhancer {
+
+  private static final ThreadLocal<String> traceId = new ThreadLocal<>();
```
garrettjonesgoogle: put a blank line after this one

google-cloud-logging/src/main/java/com/google/cloud/logging/Enhancer.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+/** A Log Enhancer. Used to enhance the {@link LogEntry} */
+public interface Enhancer {
```
garrettjonesgoogle: I would prefer the name `LoggingEnhancer`

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
```diff
@@ -0,0 +1,280 @@
...
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.api.gax.core.ApiFuture;
+import com.google.api.gax.core.ApiFutureCallback;
+import com.google.api.gax.core.ApiFutures;
+import com.google.cloud.MonitoredResource;
+import com.google.cloud.logging.Logging.WriteOption;
+import com.google.common.util.concurrent.Uninterruptibles;
+import java.util.*;
```
garrettjonesgoogle: Don't use `*` imports

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,132 @@
...
+      return "gae_app_flex";
+    }
+    if (System.getenv("KUBERNETES_SERVICE_HOST") != null) {
+      return "container";
+    }
+    if (System.getenv("GAE_MODULE") != null) {
+      return "gae_app_standard";
+    }
+    if (MetadataConfig.getInstanceId() != null) {
+      return "gce_instance";
```
garrettjonesgoogle: I think it would be best to make an enum out of resource type.

google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingErrorHandler.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+public interface LoggingErrorHandler {
```
jabubake: agree, i've moved this to LoggingService

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -107,29 +102,19 @@
...
-  private final LoggingOptions options;
-  private final WriteOption[] writeOptions;
-  private volatile Logging logging;
-  private Level flushLevel;
-  private long flushSize;
-  private Synchronicity synchronicity;
-  private final List<Enhancer> enhancers;
+  private LoggingService loggingService;
+  private List<LoggingEnhancer> enhancers;
+  private ErrorHandler errorHandler;
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingEnhancer.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+/** An enhancer for log entries.
+ * Used to add custom labels to the {@link LogEntry.Builder} */
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingConfig.java
```diff
@@ -0,0 +1,148 @@
...
+  private final LogManager manager = LogManager.getLogManager();
+  private final String className;
+  private static final String flushLevelTag = "flushLevel";
+  private static final String logFileNameTag = "log";
+  private static final String logLevelTag = "level";
+  private static final String filterTag = "filter";
+  private static final String formatterTag = "formatter";
+  private static final String synchronicityTag = "synchronicity";
+  private static final String resourceTypeTag = "resourceType";
+  private static final String enchancersTag = "enhancers";
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -18,103 +18,124 @@
 
 import com.google.cloud.MetadataConfig;
 import com.google.cloud.MonitoredResource;
+import com.google.cloud.ServiceOptions;
 import com.google.common.base.Strings;
-import java.util.List;
+
+import java.util.*;
```
jabubake: done. and fixed IDE setting.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,184 @@
...
+
+  private enum Label {
+    app_id,
+    cluster_name,
+    instance_id,
+    instance_name,
+    module_id,
+    project_id,
+    version_id,
+    zone
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,184 @@
...
+        Label.instance_id,
+        Label.zone);
+
+    addToMap(map, Resource.gae_app_standard, Label.app_id, Label.module_id, Label.version_id);
+
+    addToMap(map, Resource.container, Label.cluster_name, Label.zone);
+
+    addToMap(map, Resource.gce_instance, Label.instance_id, Label.zone);
+
+    resourceTypeWithLabels = Collections.unmodifiableMap(map);
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
```diff
@@ -0,0 +1,284 @@
...
+import com.google.cloud.logging.Logging.WriteOption;
+import com.google.common.util.concurrent.Uninterruptibles;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Set;
+
+public class LoggingService {
```
jabubake: Primarily to have level based flush handling and a simpler interface for the additional logging frameworks we're looking to support.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
```diff
@@ -0,0 +1,284 @@
...
+    synchronized (writeLock) {
+      writesToFlush.addAll(pendingWrites);
+    }
+    for (ApiFuture<Void> write : writesToFlush) {
+      try {
+        Uninterruptibles.getUninterruptibly(write);
+      } catch (Exception e) {
+        // Ignore exceptions, they are propagated to the error manager.
+      }
+    }
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
```diff
@@ -0,0 +1,284 @@
...
+      } catch (Exception ex) {
+        // ignore
+      }
+    }
+    logging = null;
+  }
+
+  public synchronized void setFlushSeverity(Severity severity) {
+    flushSeverity = severity;
+  }
```
jabubake: Agree, i missed seeing the mutability here. I've updated to new and removed the builder pattern.

google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private static final String metadataUrl = "http://metadata/computeMetadata/v1/";
```
jabubake: done

google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private static String getProjectAttribute(String attributeName) {
+    return getAttribute("project/", attributeName);
+  }
+
+  private static String getInstanceAttribute(String attributeName) {
+    return getAttribute("instance/", attributeName);
+  }
+
+  private static String getAttribute(String prefix, String attributeName) {
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -158,61 +143,64 @@ public LoggingHandler(String log, LoggingOptions options) {
...
-      this.enhancers = enhancers != null ? enhancers : helper.getEnhancerProperty(className + ".enhancers");
-      String resourceType = helper.getProperty(className + ".resourceType", "global");
-      MonitoredResource resource = monitoredResource != null ? monitoredResource : getDefaultResource(resourceType);
+      LoggingOptions loggingOptions = firstNonNull(options, LoggingOptions.getDefaultInstance());
 
-      writeOptions =
+      MonitoredResource resource =
+          firstNonNull(
+              monitoredResource, config.getMonitoredResource(loggingOptions.getProjectId()));
+      WriteOption[] labelOptions =
```
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -250,357 +240,144 @@ private static boolean hasLoggingHandler(Logger logger) {
...
-                    reportError(null, new Exception(t), ErrorManager.FLUSH_FAILURE);
-                  }
-                } finally {
-                  removeFromPending();
-                }
-              }
-            });
-        break;
+    for (LoggingEnhancer enhancer : enhancers) {
+      enhancer.enhanceLogEntry(builder);
```
jabubake: I agree, i've consolidated into a single place in LoggingHandler.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -20,99 +20,116 @@
...
-    app_id,
-    cluster_name,
-    instance_id,
-    instance_name,
-    module_id,
-    project_id,
-    version_id,
-    zone
-  }
+    appId("app_id"),
```
garrettjonesgoogle: Make the enums start with a capital letter.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,199 @@
...
+import com.google.common.collect.ImmutableMap;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Monitored resource construction utilities to detect resource type and add labels.
+ */
+abstract class MonitoredResourceUtil {
```
garrettjonesgoogle: Why is this abstract? I don't see any classes extending it.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,199 @@
...
+      return Resource.gaeAppStandard;
+    }
+    if (MetadataConfig.getInstanceId() != null) {
+      return Resource.gceInstance;
+    }
+    // default Resource type
+    return Resource.global;
+  }
+
+  protected static List<LoggingEnhancer> getResourceEnhancers() {
```
garrettjonesgoogle: This can just be package private - `protected` doesn't add anything for a static method.

google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
```diff
@@ -0,0 +1,71 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+public class MetadataConfig {
```
garrettjonesgoogle: Add a doc on the class since it's public

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingErrorHandler.java
garrettjonesgoogle: This still seems to exist

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
```diff
@@ -0,0 +1,200 @@
...
+import com.google.cloud.logging.Logging.WriteOption;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Set;
+
+/** Configurable logging service with level-based flush and error handling */
+public class LoggingService {
```
garrettjonesgoogle: Does this need to be public?

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingErrorHandler.java
jabubake: removed.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: i had done this to avoid instantiation. added private constructor instead. i've also marked this public. logging framework adapter implementations like slf4j (that will not be part of google-cloud-java) will need access to the resource construction : `getResource` and `getResourceEhancers` methods.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: logging framework adapter implementations like slf4j will need access to the resource construction : `getResource` and `getResourceEhancers` methods. moved to public + added javadoc

google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
jabubake: done.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
jabubake: Public to allow for logging adapter implementations like slf4j to use this surface.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
garrettjonesgoogle: I've been thinking about this name & structure, and here are my thoughts:

* Calling this `LoggingService` is confusing because then it's not clear how `Logging` and `LoggingService` are supposed to be related - they are both for calling the Logging service, and neither of them are "local" services.
* It seems like all of this functionality could just be rolled into `Logging`/`LoggingImpl`. There is one potential point of confusion, the fact that `Logging` has `write` and `writeAsync`. Maybe that distinction should just be removed, and the sync/async should only be controlled by the synchronicity setting, not by which function is called. The `publish` function on this class could become the new `write` function in `Logging`.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -0,0 +1,208 @@
...
+  private static String getAppEngineInstanceName() {
+    return System.getenv("GAE_INSTANCE");
+  }
+
+  private static List<LoggingEnhancer> getEnhancers(Resource resourceType) {
+    List<LoggingEnhancer> enhancers;
+    switch (resourceType) {
+      case GaeAppFlex:
+        enhancers = new ArrayList<>();
+        enhancers.add(new TraceLoggingEnhancer());
```
gregw: GaeAppStandard also has traceId available in the requests, although only GaeAppFlex will have the code to call the static setter

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
```diff
@@ -158,61 +144,65 @@ public LoggingHandler(String log, LoggingOptions options) {
...
+      loggingService = new LoggingService(
+          loggingOptions, errorHandler, logName, resource, writeOptions);
+      loggingService.setFlushSeverity(severityFor(flushLevel));
+      loggingService.setSynchronicity(config.getSynchronicity());
+
+      this.enhancers = firstNonNull(enhancers, config.getEnhancers());
+
+      List<LoggingEnhancer> loggingEnhancers = MonitoredResourceUtil.getResourceEnhancers();
+      if (loggingEnhancers != null) {
+        this.enhancers.addAll(loggingEnhancers);
```
gregw: The logging enhancer list returned at line 200 above may be immutable

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingService.java
jabubake: I agree, the separation made more sense when we had in-memory buffering which we may not have wanted to support as the default implementation. 
I'll take a pass at rolling this into Logging then.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: Thanks, I've updated to add trace enhancer on flex and standard.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java
jabubake: Thanks, updated to handle the condition

jabubake:
@garrettjonesgoogle , @pongad : I've moved the flush functionality to LoggingImpl. PTAL.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -478,10 +506,80 @@ private static WriteLogEntriesRequest writeLogEntriesRequest(LoggingOptions serv
...
+    // so the code below doesn't wait uselessly.
+    ArrayList<ApiFuture<Void>> writesToFlush = new ArrayList<>();
+    synchronized (writeLock) {
+      writesToFlush.addAll(pendingWrites);
+    }
+
+   try {
+      ApiFutures.allAsList(writesToFlush).get();
+    } catch (InterruptedException|ExecutionException e) {
+       throw new RuntimeException(e);
```
duchuyctlk: It seems that the codacy bot is not happy when you throw a generic exception. How about throwing the original one?

google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.logging;
+
+/*
+ */
```
garrettjonesgoogle: Add a comment

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -478,10 +506,80 @@ private static WriteLogEntriesRequest writeLogEntriesRequest(LoggingOptions serv
...
+    if (inWriteCall.get() != null) {
+      return;
+    }
+    inWriteCall.set(true);
+
+    try {
+      writeLogEntries(logEntries, options);
+      for (LogEntry logEntry : logEntries) {
+        // flush pending writes if log severity at or above flush severity
+        if (logEntry.getSeverity().compareTo(flushSeverity) >= 0) {
```
jabubake: I've moved to volatile here. The reasoning makes sense to me.

google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java
jabubake: Done.

jabubake:
@gregw, I've addressed your comments, PTAL.

google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
jabubake: that would require marking the logging interface to throw the named exceptions, and I did n't want to introduce that. @garrettjonesgoogle/@pongad thoughts ?

garrettjonesgoogle: You can throw LoggingException which is a subclass of RuntimeException.

meltsufin:
@jabubake You might want to take a look at [`grpc-context`](http://www.grpc.io/grpc-java/javadoc/io/grpc/Context.html) as an alternative thread local to carry the trace id.

jabubake:
@meltsufin : will look into it. @garrettjonesgoogle  recommends we do that it in a separate PR.

garrettjonesgoogle:
LGTM

meltsufin:
@jabubake and @garrettjonesgoogle thanks pulling this together! When can we expect a push to Maven?

garrettjonesgoogle:
We should have a release by the end of the week.

PR #1858, opened 2017-04-04
Add Speech v1.

shinfan:
cc @pbogle


pom.xml
```diff
@@ -115,7 +115,8 @@
     <module>google-cloud-pubsub</module>
     <module>google-cloud-resourcemanager</module>
     <module>google-cloud-spanner</module>
-    <module>google-cloud-speech</module>
+    <module>google-cloud-speech-v1</module>
+    <module>google-cloud-speech-v1beta1</module>
```
garrettjonesgoogle: I think I would prefer a single artifact that contains both major versions in it. Putting the major version in the artifact doesn't follow the patterns we have been setting up.

shinfan: Do we have any existing clients that contain more than one version? If we merge two versions into one, they would share a single pom. In the last discussion of releasing Speech, the plan is to have both versions sit side by side independently.

shinfan: After the second thought, I think merging two versions into one single artifact might be okay. The two versions will stay in separate packages and the users won't notice any differences. Will do.

shinfan: Done

garrettjonesgoogle:
LGTM

shinfan:
3 tests passed on Travis. Going to merge.

PR #1863, opened 2017-04-05
Make project.properties path independent per API

shinfan:
- Previously we generate all project.properties files into root which is
incorrect
- Now the caller API client needs to pass in package path to retrieve the
corresponding version
- Create project.properties for Veneer-only clients as well

The new properties file will be stored at:
`{ROOT}/com/google/cloud/{api}/project.properties`

Follow-up: Update Gapic clients

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -28,7 +28,6 @@ import com.google.api.gax.core.RetrySettings;
 import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.spi.ServiceRpcFactory;
-import com.google.common.base.Splitter;
 import com.google.common.collect.Iterables;
 import com.google.common.io.Files;
 import java.io.BufferedReader;
@@ -45,7 +44,6 @@ import java.lang.reflect.Method;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.charset.Charset;
-import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
 import java.util.ServiceLoader;
@@ -627,10 +625,6 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>,
   }
 
   private String getPackagePath() {
-    List<String> fullClassName = Splitter.on(".").splitToList(this.getClass().getCanonicalName());
-    if (fullClassName.size() > 2) {
-      return String.join("/", fullClassName.subList(0, fullClassName.size() - 1));
-    }
-    return DEFAULT_PACKAGE_PATH;
+    return this.getClass().getPackage().getName().replaceAll("\\.", "/");
   }
 }


google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -618,25 +620,13 @@ protected RetrySettings defaultRetrySettings() {
...
-          + ServiceOptions.class.getPackage().getName() + "/"
-          + ARTIFACT_ID + "/pom.properties";
-      InputStream inputStream = ServiceOptions.class.getResourceAsStream(mavenPropertiesPath);
-      if (inputStream != null) {
-        properties.load(inputStream);
-        return properties.getProperty(META_VERSION_KEY, "");
-      }
+      String projectPropertiesPath = PROPERTIES_ROOT
+          + artifactId + "/"
+          + PROPERTIES_FILE;
```
garrettjonesgoogle: correctness verification would be easier if `PROPERTIES_ROOT` didn't have the final slash, and the slash was put here: someone could easily see by eye that all of the necessary slashes were present.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -547,8 +545,12 @@ public static String getGoogApiClientLibName() {
   /**
    * Returns the library's version as a string.
    */
-  public static String getLibraryVersion() {
-    return LIBRARY_VERSION;
+  public static String getLibraryVersion(String artifactId) {
+    String version = getPomVersion(artifactId);
+    if (version == null) {
+      version = getManifestVersion();
```
garrettjonesgoogle: Should we try to get the manifest version of the artifact instead of google-cloud-core?

google-cloud-core/src/main/java/com/google/cloud/HttpTransportOptions.java
```diff
@@ -138,7 +138,8 @@ public HttpTransportFactory getHttpTransportFactory() {
    * Returns a request initializer responsible for initializing requests according to service
    * options.
    */
-  public HttpRequestInitializer getHttpRequestInitializer(ServiceOptions<?, ?> serviceOptions) {
+  public HttpRequestInitializer getHttpRequestInitializer(
+      ServiceOptions<?, ?> serviceOptions, final String artifactId) {
```
garrettjonesgoogle: I think everything would be a lot easier if the artifact id was put into `ServiceOptions` itself (passed into the constructor from the service-specific derived options class).

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -633,16 +641,16 @@ private static String getPomVersion(String artifactId) {
...
     String version = null;
     try {
       Enumeration<URL> resources =
           ServiceOptions.class.getClassLoader().getResources(JarFile.MANIFEST_NAME);
       while (resources.hasMoreElements() && version == null) {
         Manifest manifest = new Manifest(resources.nextElement().openStream());
         Attributes manifestAttributes = manifest.getMainAttributes();
-        String artifactId = manifestAttributes.getValue(MANIFEST_ARTIFACT_ID_KEY);
-        if (artifactId != null && artifactId.equals(ARTIFACT_ID)) {
+        String currentArtifactId = manifestAttributes.getValue(MANIFEST_ARTIFACT_ID_KEY);
```
garrettjonesgoogle: What does "current" mean here? This code is in google-cloud-core, which implies this is "core artifact id".

google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/LoggingSettings.java
```diff
@@ -43,6 +39,9 @@
 import com.google.api.gax.grpc.SimpleCallSettings;
 import com.google.api.gax.grpc.UnaryCallSettings;
 import com.google.api.gax.grpc.UnaryCallable;
+import com.google.cloud.logging.spi.v2.PagedResponseWrappers.ListLogEntriesPagedResponse;
+import com.google.cloud.logging.spi.v2.PagedResponseWrappers.ListLogsPagedResponse;
+import com.google.cloud.logging.spi.v2.PagedResponseWrappers.ListMonitoredResourceDescriptorsPagedResponse;
```
garrettjonesgoogle: revert this file

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -73,7 +74,8 @@
   public HttpBigQueryRpc(BigQueryOptions options) {
     HttpTransportOptions transportOptions = (HttpTransportOptions) options.getTransportOptions();
     HttpTransport transport = transportOptions.getHttpTransportFactory().create();
-    HttpRequestInitializer initializer = transportOptions.getHttpRequestInitializer(options);
+    HttpRequestInitializer initializer =
+        transportOptions.getHttpRequestInitializer(options);
```
garrettjonesgoogle: This file can be reverted (only whitespace changes) - there are several other Http*Rpc.java files that look like they can be reverted too.

google-cloud-core/src/main/java/com/google/cloud/GrpcTransportOptions.java
```diff
@@ -213,12 +213,13 @@ private GrpcTransportOptions(Builder builder) {
    * Returns a channel provider from the given default provider.
    */
   public static ChannelProvider setUpChannelProvider(
-      InstantiatingChannelProvider.Builder providerBuilder, ServiceOptions<?, ?> serviceOptions) {
+      InstantiatingChannelProvider.Builder providerBuilder,
+      ServiceOptions<?, ?> serviceOptions) {
```
garrettjonesgoogle: revert the line formatting - there are a few more line reformattings you can revert too.

google-cloud-language/pom.xml
```diff
@@ -97,10 +97,10 @@
           <execution>
             <phase>generate-resources</phase>
             <goals>
-                <goal>write-project-properties</goal>
+              <goal>write-project-properties</goal>
             </goals>
             <configuration>
-                <outputFile>${project.build.outputDirectory}/project.properties</outputFile>
+              <outputFile>${project.build.outputDirectory}/properties/${project.artifactId}/project.properties</outputFile>
```
garrettjonesgoogle: What version will the GAPIC client of google-cloud-language report on the header?

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
shinfan: I was looking for something that means `retrieved value from the manifest`. 
Changed to `value`.

google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/LoggingSettings.java
shinfan: Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
shinfan: Reverted Http*Rpc files

google-cloud-core/src/main/java/com/google/cloud/GrpcTransportOptions.java
shinfan: Done

google-cloud-language/pom.xml
shinfan: Right now it reports beta version. I am actually a bit confused about this. Should the gapic version always be alpha?

shinfan:
@garrettjonesgoogle PTAL

As discussed offline, I have modified the properties file path to be `/{package_path/project.properties`

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -629,34 +620,17 @@ protected RetrySettings defaultRetrySettings() {
...
-        if (value != null && value.equals(artifactId)) {
-          version = manifestAttributes.getValue(MANIFEST_VERSION_KEY);
-        }
-      }
-    } catch (IOException e) {
-      // ignore
+  private String getPackagePath() {
+    List<String> fullClassName = Splitter.on(".").splitToList(this.getClass().getCanonicalName());
+    if (fullClassName.size() > 2) {
+      return String.join("/", fullClassName.subList(0, fullClassName.size() - 1));
```
garrettjonesgoogle: `String.join` doesn't exist in java 7 (this caused build failures)

Here is a simpler way to do this method:

```
return this.getClass().getPackage().getName().replaceAll("\\.", "/");
```

garrettjonesgoogle:
LGTM if tests pass

shinfan:
3 of 4 tests passed and 1 failed for timeout (most likely due to the flakiness of pubsub) Going to merge

PR #1869, opened 2017-04-05
Setting a maximum period for ack extensions.

davidtorres:
This is a feature that we are porting from the Go library. We think is a great idea; it prevents subscribers that lose track of the ack reply handler from trapping messages.

Note: that the message will still be accounted in the flow controller which may lead to stuckness, we discussed this largely and think that is best over the potential of letting the process to OOM. Ultimately, losing track of the ack reply handler is a bug so letting it get blocked by the flow controller we believe will be a symptom easier to debug than OOMs.

### DIFF
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -153,16 +153,6 @@
       <version>2.9.2</version>
     </dependency>
     <dependency>
-      <groupId>com.google.auth</groupId>
-      <artifactId>google-auth-library-credentials</artifactId>
-      <version>0.4.0</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.auth</groupId>
-      <artifactId>google-auth-library-oauth2-http</artifactId>
-      <version>0.4.0</version>
-    </dependency>
-    <dependency>
       <groupId>com.google.api-client</groupId>
       <artifactId>google-api-client</artifactId>
       <version>1.20.0</version>
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <github.global.server>github</github.global.server>
-    <google.auth.version>0.6.0</google.auth.version>
+    <google.auth.version>0.6.1</google.auth.version>
     <grpc.version>1.0.3</grpc.version>
     <gax.version>0.8.1</gax.version>
     <generatedProto.version>0.1.5</generatedProto.version>


pongad:
@davidtorres I merged a bug fix that conflicts with this PR. I thought I'd do you a favor and resolve the conflict. Turns out the fix introduces a more serious bug, so I also reverted the conflict resolution. I'm sorry about the code churn.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
```diff
@@ -305,7 +315,7 @@ public void run() {
 
     synchronized (outstandingAckHandlers) {
       outstandingAckHandlers.add(
-          new ExtensionJob(expiration, INITIAL_ACK_DEADLINE_EXTENSION_SECONDS, ackHandlers));
+          new ExtensionJob(new Instant(clock.millis()), expiration, INITIAL_ACK_DEADLINE_EXTENSION_SECONDS, ackHandlers));
```
davidtorres: Done thanks

davidtorres:
@pongad Thanks for reviewing. PTAL to see we can commit this PR.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -479,6 +485,21 @@ public Builder setAckExpirationPadding(Duration ackExpirationPadding) {
...
     }
 
+    /**
+     * Set the maximum period a message ack deadline will be extended.
+     *
+     * <p>It is recommended to set this value to a reasonable upper bound of the subscriber time to
+     * process any message. This maximum period avoids messages to be <i>locked</i> by a subscriber
+     * in cases when the {@link AckReply} is lost.
+     *
+     * <p>A zero duration effectively disables auto deadline extensions.
```
pongad: I'm sorry I just realized this. We said "zero duration effectively disables auto deadline extensions" but we then check that the time is non-zero. I realize it's still possible to set the duration is something really small instead, but I think we should be consistent.

pongad:
One more nit, but I think that's it from me. @garrettjonesgoogle do you also want to take a look?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
davidtorres: Sorry I'm not following, where do we check it is strictly non-zero?

garrettjonesgoogle: This might be a misreading of the code - it is `>= 0`, not `> 0`...

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/SubscriberImplTest.java
```diff
@@ -276,8 +277,8 @@ public void testModifyAckDeadline() throws Exception {
     Subscriber subscriber =
         startSubscriber(
             getTestSubscriberBuilder(testReceiver)
-                .setAckExpirationPadding(Duration.standardSeconds(1)));
-
+                .setAckExpirationPadding(Duration.standardSeconds(1))
+                .setMaxAckExtensionPeriod(Duration.standardSeconds(13)));
```
garrettjonesgoogle: It seems like we should have another test where MaxAckExtensionPeriod is not set.

davidtorres: Just to clarify, you want me to test the default max expiration case?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
pongad: I'm stupid. LGTM

pongad:
LGTM, @garrettjonesgoogle can +2

davidtorres:
Not sure what happened here I never closed it.

pongad:
@davidtorres Did you remove/reset your branch? Unlike Google-internal code review and Gerrit, GitHub doesn't "snapshot". It just has your branch which gets merged when the PR is accepted.

davidtorres:
Yes, I reverted an unintended commit and that cancelled this PR, I sent you another PR #1898. Sorry for the mess.

pongad:
@davidtorres Not at all! I confuse myself every time I move code review tool!

PR #1876, opened 2017-04-06
Replace a constant of type Set with ImmutableSet

vkedia:


vkedia:
This is motivated by documentation of ImmutableCollection (the section "Interfaces not implementations"):
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ImmutableCollection.html

PR #1904, opened 2017-04-12
Update org.json to latest version

shinfan:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1502 is caused by a dependency conflict of an older version of org.json package (which is used by a external dependency). The solution is to override the org.json version or exclude the older version from the external dependency.

I don't think there is much we can do in our client to prevent such dependency issues. But since our org.json package is very old I think we can upgrade to the latest anyway.

The only place we use org.json is in [ServiceOptions](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/3f0f645861c6f56ef193245ba3c9efa60a87a9fd/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L428) class so this should not cause any issue.

### DIFF
--- a/google-cloud-core/pom.xml
+++ b/google-cloud-core/pom.xml
@@ -102,7 +102,7 @@
     <dependency>
       <groupId>org.json</groupId>
       <artifactId>json</artifactId>
-      <version>20151123</version>
+      <version>20160810</version>
       <scope>compile</scope>
     </dependency>
     <dependency>


garrettjonesgoogle:
LGTM after tests pass

shinfan:
3 tests passed on Travis. Going to merge.

PR #1905, opened 2017-04-12
 removing pubsubexample, adding snippet for async subscriber

jabubake:


google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
```diff
@@ -66,4 +69,37 @@ public void failed(Subscriber.State from, Throwable failure) {
     subscriber.stopAsync().awaitTerminated();
     // [END startAsync]
   }
+
+  private void createSubscriber() throws Exception {
+    // [START async_pull_subscription]
```
pongad: Where is this used? I grepped through the google-cloud-java repo and don't see it referenced.

pongad:
I have one comment but otherwise it looks good to me.

PR #1908, opened 2017-04-12
Make sure Storage does not do name normalization

shinfan:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1837

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
@@ -103,6 +103,7 @@ public class ITStorageTest {
       new SecretKeySpec(BaseEncoding.base64().decode(BASE64_KEY), "AES256");
   private static final byte[] COMPRESSED_CONTENT = BaseEncoding.base64()
       .decode("H4sIAAAAAAAAAPNIzcnJV3DPz0/PSVVwzskvTVEILskvSkxPVQQA/LySchsAAAA=");
+  private static final Map<String, String> BUCKET_LABELS = ImmutableMap.of("label1", "value1");
 
   @BeforeClass
   public static void beforeClass() throws NoSuchAlgorithmException, InvalidKeySpecException {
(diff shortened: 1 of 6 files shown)


jskeet:
Sorry, I'm on vacation until Tuesday so I probably won't be able to review this until then.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1440,4 +1443,19 @@ public void testReadCompressedBlob() throws IOException {
     }
     blob.delete();
   }
+
+  @Test
+  public void testBlobNameNormalization() {
+    for (String blobName : BLOB_NORMALIZATION.keySet()) {
+      BlobInfo blobInfoFormC = BlobInfo.newBuilder(BlobId.of(BUCKET, blobName))
```
garrettjonesgoogle: Why is this called `blobInfoFormC` if it can hold either form C or form D?

shinfan: Oops, I started with using hardcoded variables and switched to using this loop, then I forgot to change the name back....

Done.

shinfan:
@jskeet No worries :)

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1440,4 +1443,19 @@ public void testReadCompressedBlob() throws IOException {
...
   }
+
+  @Test
+  public void testBlobNameNormalization() {
+    for (String blobName : BLOB_NORMALIZATION.keySet()) {
+      BlobInfo blobInfo = BlobInfo.newBuilder(BlobId.of(BUCKET, blobName))
+          .setContentType("text/plain")
+          .setContentEncoding("UTF-8")
+          .build();
+      Blob blob = storage.create(blobInfo, BLOB_NORMALIZATION.get(blobName).getBytes());
```
jskeet: This isn't necessarily proving what we want it to. In particular, without checking the code, I don't know whether `Blob.getName()` reflects the value *returned* by Storage, or whether it's just using the `BlobInfo` we provided.

I would suggest using the known-to-demonstrate-both-cases existing storage files - check you can download both of those and get the right values.

shinfan:
Oops I updated this branch by mistake. I will fix this.

garrettjonesgoogle:
@shinfan could you maybe create a PR off of a different branch so that it isn't prone to getting overwritten?

PR #1913, opened 2017-04-13
Fix generated grpc package doc links

garrettjonesgoogle:
Superset of @pongad 's PR https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1910 .


garrettjonesgoogle:
Yes, they are all batch-generated and published, and the dependencies in google-cloud-java are updated; the only thing missing was the version for the generated docs.

Agreed, api-client-staging should be updated too (of course, separate PR since it's a separate repo).

garrettjonesgoogle:
merging (only appveyor hasn't finished yet)

PR #1915, opened 2017-04-13
Fixing executor and flow control issues in subscriber

### LINKED ISSUE #1855: Potential deadlock in Subscriber
@davidtorres @garrettjonesgoogle 

A lot of our code uses callback without specifying an executor. [`PollingSubscriberConnection`](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/88eff83126d60ff4fb83541f8c954afdc8da425f/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java#L137) is an example. IIUC, if we don't specify one, the library uses "directExecutor". Is this just the same executor we create the channel with? I think that it is.

If I'm right, this is dangerous. If we receive a bunch of messages, pullers will block on FlowController, waiting for messages to get processed. Messages can't get processed because all the threads in the pool are blocked on FlowController, resulting in a deadlock. I don't think this can happen in default config since we have 5 threads for each puller, but if user provides their own executor, all bets are off.

My current idea: Only use the executor for RPCs. Additionally ask for `numWorkerThreads` and create a new executor for processing messages. Assuming that the user-provided MessageReceiver cannot directly cause more tasks to be registered to the pullers, we should be able to avoid deadlocks.

What do you think?

### LINKED ISSUE #1865: Ensure that single-threaded message pulling works
i.e. make sure it doesn't deadlock.

### LINKED ISSUE #1868: PollingSubscriberConnection - allow number of messages delivered to be configurable
The value is hard coded to an upper bound of 1000 (which used to be 500).  Shouldn't it be configurable?

This doesn't work well for really large messages and for messages that take substantial time to process. 

I see no way to configure this value.  Doesn't look like the client has any control over the number of messages pulled from the subscription.  I was hoping that the FlowController logic would support but it seems focused on whether/when to go back the server to get more rather than limiting the number pulled.

Note that if FlowController is set to 1 element, deadlock is likely on the FlowController semaphore (for example, it being set to 1, while having just processed 10 messages).

davidtorres:
The current subscriber logic has a few concurrency issues explained as follows:

- The flow controller blocks the user's executor threads; so in the event of all the threads getting blocked because of flow control limits, the subscriber won't be able to dispatch messages to the user code hence unable to release flow control permits and to make forward progress; a deadlock. This change makes the flow control non-blocking and instead regulates the flow of messages by either suspending the streaming or not issuing more pulls but until permits are released from flow controller
- Lease extensions are handled under the user's executor, in the event of the user code monopolizing all the threads in the executor, the subscriber won't be able to execute any lease extension which will make the leases to expire and increase the chances of message redelivery; if there is a time when it is more important to maintain lease extensions is when the user code is busy working on the messages. This code change sets up a separate small (two threads) shared-across-subscribers executor for taking care of lease extensions, under my load tests (with >600 hundred subscriber objects) this executor performed well not showing signs of work backup, this is mainly because the work of scheduling and sending lease extensions is fairly cheap.
- Flow control permits were being drawn on a per batch basis while the messages were dispatch one at a time, this could cause a deadlock if ever receiving a batch larger than the limit of message in the flow controller. This change makes the code reserve flow control permits on a per messages basis as they are dispatched to the user code.

Lastly, I have also address a minor issues which is the inability of choosing the number of messages to pull per pull request, now such number is taken from the flow controller settings.

This change fixes #1868, fixes #1865 and fixes #1855

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
@@ -16,7 +16,7 @@
 
 package com.google.cloud.pubsub.spi.v1;
 
-import com.google.api.gax.core.ApiClock;
+import com.google.api.core.ApiClock;
 import com.google.api.gax.core.FlowController;
 import com.google.api.gax.core.FlowController.FlowControlException;
 import com.google.api.gax.grpc.InstantiatingExecutorProvider;
@@ -46,6 +46,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nullable;
 import org.joda.time.Duration;
 import org.joda.time.Instant;
 import org.joda.time.Interval;
@@ -61,10 +62,11 @@ class MessageDispatcher {
   @VisibleForTesting static final Duration PENDING_ACKS_SEND_DELAY = Duration.millis(100);
   private static final int MAX_ACK_DEADLINE_EXTENSION_SECS = 10 * 60; // 10m
 
-  private static final ScheduledExecutorService alarmsExecutor =
+  private static final ScheduledExecutorService SHARED_ALARMS_EXECUTOR =
       InstantiatingExecutorProvider.newBuilder().setExecutorThreadCount(2).build().getExecutor();
 
   private final ScheduledExecutorService executor;
+  private final ScheduledExecutorService alarmsExecutor;
   private final ApiClock clock;
 
   private final Duration ackExpirationPadding;
@@ -243,8 +245,10 @@ class MessageDispatcher {
       Distribution ackLatencyDistribution,
       FlowController flowController,
       ScheduledExecutorService executor,
+      @Nullable ScheduledExecutorService alarmsExecutor,
       ApiClock clock) {
     this.executor = executor;
+    this.alarmsExecutor = alarmsExecutor == null ? SHARED_ALARMS_EXECUTOR : alarmsExecutor;
     this.ackExpirationPadding = ackExpirationPadding;
     this.maxAckExtensionPeriod = maxAckExtensionPeriod;
     this.receiver = receiver;
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -46,6 +46,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nullable;
 import org.joda.time.Duration;
 
 /**
@@ -93,6 +94,7 @@ public class Subscriber extends AbstractApiService {
   private final Duration ackExpirationPadding;
   private final Duration maxAckExtensionPeriod;
   private final ScheduledExecutorService executor;
+  @Nullable private final ScheduledExecutorService alarmsExecutor;
   private final Distribution ackLatencyDistribution =
       new Distribution(MAX_ACK_DEADLINE_SECONDS + 1);
   private final int numChannels;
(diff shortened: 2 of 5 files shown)


davidtorres:
@pongad @garrettjonesgoogle all yours to review thanks

garrettjonesgoogle:
I will have @pongad review it first.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
```diff
@@ -275,28 +284,113 @@ public int getMessageDeadlineSeconds() {
...
+
+    public void addMessage(ReceivedMessage receivedMessage, AckHandler ackHandler) {
+      this.messages.add(new OutstandingMessage(receivedMessage, ackHandler));
+    }
+
+    public Deque<OutstandingMessage> messages() {
+      return messages;
+    }
+
+    public void done() {
```
pongad: I don't think this is used anywhere.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
```diff
@@ -57,6 +61,9 @@
   @VisibleForTesting static final Duration PENDING_ACKS_SEND_DELAY = Duration.millis(100);
   private static final int MAX_ACK_DEADLINE_EXTENSION_SECS = 10 * 60; // 10m
 
+  private static final ScheduledExecutorService alarmsExecutor =
```
pongad: I think this will block the JVM from exiting. [`Thread` class](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html) documents that unless `Runtime.exit` is called, the JVM keeps running if there is at least one non-daemon thread. [`Executors` class](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#defaultThreadFactory()) in turn documents that it creates non-daemon threads by default.

Could you test this by running a CLI application, maybe `CreateSubscriptionAndConsumeMessages.java`? I think unit tests call `Runtime.exit(0)` when all tests pass so it kills everything and we don't see the test hang.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
davidtorres: Gone

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
davidtorres: Done

garrettjonesgoogle:
Looks like there is some fallout from the merge that is causing the build to fail

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -119,7 +120,13 @@ private Subscriber(Builder builder) throws IOException {
             Ints.saturatedCast(ackExpirationPadding.getStandardSeconds()));
     clock = builder.clock.isPresent() ? builder.clock.get() : CurrentMillisClock.getDefaultClock();
 
-    flowController = new FlowController(builder.flowControlSettings);
+    flowController =
+        new FlowController(
+            builder
+                .flowControlSettings
+                .toBuilder()
+                .setLimitExceededBehavior(LimitExceededBehavior.ThrowException)
```
pongad: It is but I think this is the intended behavior. Setting it to `IGNORE` will almost definitely OOM the machine and setting it to `BLOCK` will probably deadlock the executor.

I think we need to either overwrite or check-and-error.

davidtorres: Michael is right, we can't allow the use of BLOCK or IGNORE.

So let me know what do you think is best, override the user setting or throw an IllegalArgumentException? I honestly think that overriding the value is fine, since message-receiving-actions are not initiated by the user so the limitexceededbehavior is basically meaningless to the user on the subscriber side.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/MessageDispatcher.java
```diff
@@ -275,28 +284,109 @@ public int getMessageDeadlineSeconds() {
...
+    public void addMessage(ReceivedMessage receivedMessage, AckHandler ackHandler) {
+      this.messages.add(new OutstandingMessage(receivedMessage, ackHandler));
+    }
+
+    public Deque<OutstandingMessage> messages() {
+      return messages;
+    }
+  }
+
+  public void processReceivedMessages(List<ReceivedMessage> messages, Runnable doneCallback) {
```
davidtorres: the done callback requires no input and not output, seems awkward to me to force a function with just Void inputs and outputs.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
garrettjonesgoogle: Let's go with what you have.

garrettjonesgoogle:
LGTM

PR #1927, opened 2017-04-17
Changes required for updated api-common

michaelbausor:
Tests currently failing because grpc packages need to be released with updated api-common

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
 
     <google.auth.version>0.6.1</google.auth.version>
     <grpc.version>1.2.0</grpc.version>
-    <gax.version>0.11.0</gax.version>
+    <gax.version>0.12.0</gax.version>
     <api-common.version>0.2.0</api-common.version>
     <generatedProto.version>0.1.7</generatedProto.version>
 


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -17,12 +17,12 @@
...
 import com.google.api.gax.batching.BatchingSettings;
-import com.google.api.gax.core.ApiFuture;
-import com.google.api.gax.core.ApiFutures;
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutures;
 import com.google.api.gax.core.FlowControlSettings;
 import com.google.api.gax.core.FlowController;
 import com.google.api.gax.core.RetrySettings;
-import com.google.api.gax.core.SettableApiFuture;
+import com.google.api.core.SettableApiFuture;
```
garrettjonesgoogle: Looks like the imports need to be resorted

michaelbausor: I don't want to do that in this PR: it would mean many unrelated changes. I can do that in a separate PR

garrettjonesgoogle:
The updates LGTM

PR #1928, opened 2017-04-17
Improve GCS signUrlOption javadoc

danielcompton:
This isn't quite ready for merge yet. I'm not sure if there is a better wording for the `withContentType` doc, and whether the `httpMethod` doc should actually go against `Storage.signUrl`?

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
@@ -806,7 +806,8 @@ public interface Storage extends Service<StorageOptions> {
     }
 
     /**
-     * The HTTP method to be used with the signed URL. Defaults to GET.
+     * The HTTP method to be used with the signed URL. 
+     * If this method is not called, defaults to GET.
      */
     public static SignUrlOption httpMethod(HttpMethod httpMethod) {
       return new SignUrlOption(Option.HTTP_METHOD, httpMethod.name());


garrettjonesgoogle:
@danielcompton , did you want to finish this PR?

danielcompton:
I think this is ready to merge if you're happy with the wording?

garrettjonesgoogle:
@andreamlin can you verify the correctness of the new comments and merge if it looks good?

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -806,7 +806,7 @@ Object getValue() {
     }
 
     /**
-     * The HTTP method to be used with the signed URL.
+     * The HTTP method to be used with the signed URL. Defaults to GET.
```
andreamlin: Does it make sense to say "Defaults to GET" for a setter method? In what case would this method be called, but the default GET is used?

danielcompton: I'm trying to communicate that GET is the default method if this isn't called, not quite sure what wording I should use here? Maybe this?

> The HTTP method to be used with the signed URL. If this is not set, defaults to GET.

andreamlin: Sounds good!

danielcompton: Done, ready for re-review I think?

PR #1931, opened 2017-04-17
Mark LocalDatastoreHelper as internal

shinfan:


garrettjonesgoogle:
LGTM

garrettjonesgoogle:
Update the commit description when you merge.

danieln100:
@shinfan Why was this class marked as "Internal testing use only"? I was looking for a way to write some unit tests for our GCD queries and this class seemed like the perfect candidate for that. Is there another class that I could use for that?

PR #1934, opened 2017-04-17
Removing timeout settings that are set but never used

garrettjonesgoogle:


michaelbausor:
LGTM - did these used to be used? Or have they never been used at all?

garrettjonesgoogle:
Yes, they used to be used, but @vam-google 's PR https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1738 stopped using them after `RetrySettings` was introduced. Now, going through `RetrySettings` (which can be set on `ServiceOptions`) is the correct thing to do.

PR #1937, opened 2017-04-18
Creating an expandable enum type, converting StorageClass

garrettjonesgoogle:
Design notes:

* I made it so that converting an enum into a StorageEnumValue is not disruptive at a code level (e.g. it has `valueOf` and `values`, same as Java enums)
* I am only converting one enum now, and after we agree on the design, I will do a follow-up PR to convert the rest.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/StringEnumValue.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/StringEnumValue.java
@@ -24,8 +24,11 @@ import com.google.common.base.Preconditions;
  * a context.
  */
 public abstract class StringEnumValue {
-  private String constant;
+  private final String constant;
 
+  /**
+   * Don't create subclasses outside of google-cloud-java.
+   */
   @InternalApi
   protected StringEnumValue(String constant) {
     this.constant = Preconditions.checkNotNull(constant);
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
@@ -29,7 +29,7 @@ public class StorageClass extends StringEnumValue {
     super(constant);
   }
 
-  private static ApiFunction<String, StorageClass> CONSTRUCTOR =
+  private static final ApiFunction<String, StorageClass> CONSTRUCTOR =
       new ApiFunction<String, StorageClass>() {
         @Override
         public StorageClass apply(String constant) {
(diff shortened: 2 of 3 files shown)


garrettjonesgoogle:
Updates https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1832

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -15,39 +15,82 @@
...
    */
-  DURABLE_REDUCED_AVAILABILITY,
+  public static StorageClass DURABLE_REDUCED_AVAILABILITY = type.createAndRegister(
+      "DURABLE_REDUCED_AVAILABILITY");
+
+  /**
+   * Get the StorageClass for the given String constant, and throw an exception if the constant is
+   * not recognized.
+   */
+  public static StorageClass valueOfStrict(String constant) {
```
shinfan: Can we move these methods (valueOf/valueOfStrict/values) into the parent class? It looks like they are common enough.

google-cloud-core/src/main/java/com/google/cloud/StringEnumValue.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * This represents a specific instance of a concept having a known set of acceptable String values,
+ * which can expand later due to new API features. Standard Java enums can't be used in such
+ * a context.
+ */
+public abstract class StringEnumValue {
+  private String constant;
+
+  @InternalApi
```
shinfan: Move this to L25?

google-cloud-core/src/main/java/com/google/cloud/StringEnumType.java
```diff
@@ -0,0 +1,80 @@
...
+@InternalApi
+public class StringEnumType<EnumT> {
+
+  private final String enumName;
+  private final ApiFunction<String, EnumT> constructor;
+  private Map<String, EnumT> knownValues = new HashMap<>();
+
+  public StringEnumType(String enumName, ApiFunction<String, EnumT> constructor) {
+    this.enumName = enumName;
+    this.constructor = constructor;
```
shinfan: Add Precondition.checkNotNull() here

google-cloud-core/src/main/java/com/google/cloud/StringEnumType.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * This represents a concept having a known set of acceptable String values, which can expand later
+ * due to new API features.
+ */
+@InternalApi
+public class StringEnumType<EnumT> {
+
+  private final String enumName;
+  private final ApiFunction<String, EnumT> constructor;
+  private Map<String, EnumT> knownValues = new HashMap<>();
```
shinfan: make this final as well

michaelbausor:
Design looks good to me too. Two points: (1) Is may be possible to push some functionality from `StorageClass` into `StringEnumValue` by making `StringEnumValue` parametrized recursively, e.g. `class StringEnumValue<T extends StringEnumValue<T>>`. This reduces the code in `StorageClass`, at the expense of more complexity. (2) Do we need tests for `StringEnumValue` and `StringEnumType`?

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
garrettjonesgoogle: No, they are static.

google-cloud-core/src/main/java/com/google/cloud/StringEnumValue.java
garrettjonesgoogle: `StringEnumValue` itself isn't internal - public classes extend it. I just don't want external code creating subclasses.

google-cloud-core/src/main/java/com/google/cloud/StringEnumType.java
garrettjonesgoogle: Ahh yes of course.

google-cloud-core/src/main/java/com/google/cloud/StringEnumType.java
garrettjonesgoogle: Will do.

garrettjonesgoogle:
(1) Which functionality can be moved up? The only public things in `StorageClass` are static. 
(2) It does make sense to make tests for `StringEnumValue` and `StringEnumType` - I'll add those in.

michaelbausor:
Re (1): you are right.

garrettjonesgoogle:
Added unit tests, PTAL

michaelbausor:
LGTM

PR #1938, opened 2017-04-18
Make SSLHandShakeException retryable

shinfan:
SSLHandShakeException should be retryable as long as it is not caused by a certificate issue.

Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1762


garrettjonesgoogle:
If you look at the original issue, it looks like host connection problems can cause this.

neozwu:
One thing I'm not quite sure is whether it is the random connection problem or it is the underlying SSL-related issues that cause the server to terminate the connection.  The later seems to be possible. (For example, see this [stackoverflow thread](http://stackoverflow.com/questions/28908835/ssl-peer-shut-down-incorrectly-in-java))

shinfan:
@neozwu It will not retry if it is caused by a cert issue, otherwise it should be retyable

PR #1940, opened 2017-04-18
Converting all Storage deserialized enum types to StringEnumValue

garrettjonesgoogle:
Converting everything Storage-related where X.valueOf is called on it during proto deserialization. 

Updates https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1832 .

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/Identity.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/Identity.java
@@ -280,10 +280,12 @@ public final class Identity implements Serializable {
   public static Identity valueOf(String identityStr) {
     String[] info = identityStr.split(":");
     Type type = Type.valueOf(CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_UNDERSCORE, info[0]));
-    if (type.equals(Type.ALL_AUTHENTICATED_USERS) || type.equals(Type.ALL_USERS)) {
+    if (info.length == 1) {
       return new Identity(type, null);
+    } else if (info.length == 2){
+      return new Identity(type, info[1]);
     } else {
-      return new Identity(type, checkNotNull(info[1]));
+      throw new IllegalArgumentException("Illegal identity string: \"" + identityStr + "\"");
     }
   }
 }
--- a/google-cloud-core/src/main/java/com/google/cloud/StringEnumType.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/StringEnumType.java
@@ -66,6 +66,9 @@ public class StringEnumType<EnumT> {
    * Get the enum object for the given String constant, and allow unrecognized values.
    */
   public EnumT valueOf(String constant) {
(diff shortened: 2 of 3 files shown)


google-cloud-core/src/main/java/com/google/cloud/Identity.java
```diff
@@ -256,27 +280,10 @@ public String strValue() {
...
-      case PROJECT_EDITOR:
-        return Identity.projectEditor(info[1]);
-      case PROJECT_VIEWER:
-        return Identity.projectViewer(info[1]);
-      default:
-        throw new IllegalStateException("Unexpected identity type " + type);
+    if (type.equals(Type.ALL_AUTHENTICATED_USERS) || type.equals(Type.ALL_USERS)) {
+      return new Identity(type, null);
+    } else {
+      return new Identity(type, checkNotNull(info[1]));
```
michaelbausor: You will get an array out of bounds error here if the user passes in a custom/new value that doesn't have a `:`

garrettjonesgoogle: Yeah, that was inherited from the existing code - do you feel I need to fix it in this PR?

michaelbausor: Don't need to, but if we don't fix now then should open an issue to track.

michaelbausor:
LGTM once array bounds issue is resolved or tracked in an issue

garrettjonesgoogle:
PTAL again, made some minor logic changes to account for the array out of bounds issue.

google-cloud-core/src/main/java/com/google/cloud/Identity.java
```diff
@@ -280,10 +280,12 @@ public String strValue() {
   public static Identity valueOf(String identityStr) {
     String[] info = identityStr.split(":");
     Type type = Type.valueOf(CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_UNDERSCORE, info[0]));
-    if (type.equals(Type.ALL_AUTHENTICATED_USERS) || type.equals(Type.ALL_USERS)) {
+    if (info.length == 1) {
```
michaelbausor: This allows someone to pass in `ALL_AUTHENTICATED_USERS` with an argument, or something other than the `ALL_*` types without an argument, and still get a result.

garrettjonesgoogle: Yes it does. I don't want to be very restrictive with server results though, because I can't predict what kind of things it will do. Users who want to pass in an Identity should be using the static factory methods anyway, which they should be drawn to because they are nice and descriptive.

michaelbausor: OK, so long as we are happy to accept the less restrictive behaviour.

michaelbausor:
LGTM

PR #1955, opened 2017-04-20
Fix credential NPE

### LINKED ISSUE #1565: ServiceOptions.getCredentials can return null and cause NPE in Logging and PubSub
In ServiceOptions.java is no credentials are explicitly specified, it would default to calling getApplicationDefault here:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L312

But that method can fail by throwing an exception if for eg the credentials file is missing. In that case ServiceOptions just catches the exception, ignores it and returns null:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L314

But reading the documentation on getCredentials, it is not clear that callers should be handling null return value. google-cloud-java seems to be not handling the null return value atleast in couple of places:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/00add6e26fc2e1a533b715df28a5568222483f65/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/DefaultLoggingRpc.java#L117
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/00add6e26fc2e1a533b715df28a5568222483f65/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/DefaultPubSubRpc.java#L146

neozwu:
fix #1565

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java
@@ -28,6 +28,7 @@ import com.google.api.gax.grpc.ProviderManager;
 import com.google.api.gax.grpc.UnaryCallSettings;
 import com.google.cloud.GrpcTransportOptions;
 import com.google.cloud.GrpcTransportOptions.ExecutorFactory;
+import com.google.cloud.NoCredentials;
 import com.google.cloud.logging.LoggingException;
 import com.google.cloud.logging.LoggingOptions;
 import com.google.logging.v2.CreateLogMetricRequest;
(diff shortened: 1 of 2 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java
```diff
@@ -82,7 +81,7 @@ public GrpcLoggingRpc(LoggingOptions options) throws IOException {
       ChannelProvider channelProvider;
       // todo(mziccard): ChannelProvider should support null/absent credentials for testing
       if (options.getHost().contains("localhost")
-          || options.getCredentials().equals(NoCredentials.getInstance())) {
+          || options.credentialIsNoCredentialInstance()) {
```
garrettjonesgoogle: Probably easier:

`NoCredentials.getInstance().equals(options.getCredentials())`

Then you don't need a dedicated method.

neozwu: Yes, you are right. Done this way.

neozwu:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #1961, opened 2017-04-21
fix credentials NPE bug 

neozwu:
#1565


google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java
```diff
@@ -82,7 +81,7 @@ public GrpcLoggingRpc(LoggingOptions options) throws IOException {
       ChannelProvider channelProvider;
       // todo(mziccard): ChannelProvider should support null/absent credentials for testing
       if (options.getHost().contains("localhost")
-          || options.getCredentials().equals(NoCredentials.getInstance())) {
+          || options.credentialIsNoCredentialInstance()) {
```
garrettjonesgoogle: So why not implement a proper equals() method on `NoCredentials`?

neozwu: I didn't implement `equals` on NoCredentials, since it is used as a singleton class.  I changed the code to use "==" to compare singleton object.

neozwu:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
I think I would prefer to add an `equals` method into NoCredentials which checks equality of classes. Note that NoCredentials is serializable; I feel like it's safer to not assume the singleton will be perfect, just in case. Also, using `equals` allows for changing comparison logic later inside of `NoCredentials`. Then, go back to the code that looks like:

`NoCredentials.getInstance().equals(options.getCredentials())`

garrettjonesgoogle:
LGTM after tests pass

PR #1968, opened 2017-04-21
Splitting core into core, core-http, and core-grpc

garrettjonesgoogle:
Depends on https://github.com/googleapis/gax-java/pull/288

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,8 +143,6 @@
     <bigquery.version>0.14.1-beta-SNAPSHOT</bigquery.version>
     <compute.version>0.14.1-alpha-SNAPSHOT</compute.version>
     <contrib.version>0.14.1-alpha-SNAPSHOT</contrib.version>
-    <core.version>1.0.1-rc1-SNAPSHOT</core.version>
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
```diff
@@ -34,8 +34,7 @@
       boolean retryable,
       @Nullable String message,
       @Nullable Throwable cause) {
-    super(code.getCode(), message, null /* reason */, false /* idempotent */, cause);
-    this.retryable = retryable;
+    super(message, cause, code.getCode(), retryable);
```
michaelbausor: Is the reshuffling of parameters necessary here?

garrettjonesgoogle: It's the SpannerException that has an odd order - the order of the base exception is the "normal" one (used in multiple places). I didn't want to mess with the SpannerException for now.

Issue #1959, opened 2017-04-21
Inconsistency in credentials specification for CPS Publisher and Subscriber


kamalaboulhosn:
Both the Publisher and the Subscriber class state "If no credentials are provided, the [Publisher|Subscriber] will use application default creentials through GoogleCredentials#getApplicationDefault. However, while Subscriber.Builder has a setCredentials method, Publisher.Builder does not. Moreover, from what I can see, Subscriber does not actually use the credentials set via Subscriber.Builder.setCredentials. If we are going to allow one to set credentials this way, we should be consistent between the Publisher and Subscriber and ensure we use the provided credentials. Otherwise, we should remove the method from the Subscriber.Builder.

was2:
I am seeing the same behavior with Subscriber.Builder.setCrecentials - even when valid creds are provided, I get an error complaining that app default credentials are not available. This appears to have started in 0.13.0.

garrettjonesgoogle:
gRPC has a way to provide credentials through `CallOptions`, which is currently marked with `@ExperimentalApi`, but that will be removed soon. This enables us to change how credentials are provided to GAPIC clients: we could move `CredentialsProvider` from`ChannelProvider` up to `ClientSettings` (so that it is a sibling of `ChannelProvider`). This would make the code to provide custom credentials a bit simpler. Then, this same design could be used in `Publisher`/`Subscriber`, where credentials are provided separately from the channel. 

Current code:

```
CredentialsProvider credentialsProvider =
    FixedCredentialsProvider
        .create(ServiceAccountCredentials.fromStream(new FileInputStream("credentials.json");
ChannelProvider channelProvider =
    TopicAdminSettings.defaultChannelProviderBuilder()
        .setCredentialsProvider(credentialsProvider).build();
TopicName topicName = TopicName.create("my-project-id", "my-topic-id");
Publisher publisher =
    Publisher.defaultBuilder(topicName).setChannelProvider(channelProvider).build();
```

Proposed:

```
CredentialsProvider credentialsProvider =
    FixedCredentialsProvider
        .create(ServiceAccountCredentials.fromStream(new FileInputStream("credentials.json");
TopicName topicName = TopicName.create("my-project-id", "my-topic-id");
Publisher publisher =
    Publisher.defaultBuilder(topicName).setCredentialsProvider(credentialsProvider).build();
```

@pongad what are your thoughts?

pongad:
[...]

PR #1975, opened 2017-04-24
pubsub: delete Subscriber.Builder::setCredentials

pongad:
The credentials provided to this method is not used.
Currently the credentials should be set in the ChannelProvider.

Updates #1959.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -27,7 +27,6 @@ import com.google.api.gax.core.Distribution;
 import com.google.api.gax.grpc.ChannelProvider;
 import com.google.api.gax.grpc.ExecutorProvider;
 import com.google.api.gax.grpc.InstantiatingExecutorProvider;
-import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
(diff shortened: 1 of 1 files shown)


pongad:
@kamalaboulhosn There is actually a way to change credentials: by changing the ChannelProvider. I'm not sure how to best document this.

@garrettjonesgoogle Do we have a central doc somewhere detailing how to change credentials? We can create a .md file somewhere in the repo and link to it in docs. WDYT?

garrettjonesgoogle:
@pongad could you nominate another person in our team to do this review?

pongad:
Nominating @vam-google since we were just talking about this :laughing:

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -27,7 +27,6 @@
 import com.google.api.gax.grpc.ChannelProvider;
 import com.google.api.gax.grpc.ExecutorProvider;
 import com.google.api.gax.grpc.InstantiatingExecutorProvider;
-import com.google.auth.Credentials;
```
vam-google: Probably the documentation about credentials on the line 75 about credentials should also be removed/modified.
(Writing here, since I could not find a way to add a comment on a line which was not modified/added/removed)

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -450,7 +449,6 @@ public void run() {
             .build();
 
     SubscriptionName subscriptionName;
-    Optional<Credentials> credentials = Optional.absent();
```
vam-google: Minor:
Not related to this specific change, but this removes an Optional<> field and there is one more Optional<> field in this class: Optional<ApiClock> clock. Using optionals for class fields is considered a bad practice (for example: http://stackoverflow.com/questions/23454952/uses-for-optional), plus Optional is not serializable and can cause problems for serializable classes. Can you also change clock's optional to a regular field as part of this PR?

Issue #1854, opened 2017-04-04
Front page has outdated API


danawillow:
http://googlecloudplatform.github.io/google-cloud-java/0.11.0/index.html has a code snippet:
```
DatastoreOptions options = DatastoreOptions.newBuilder()
  .setProjectId(PROJECT_ID)
  .setAuthCredentials(AuthCredentials.createForJson(
    new FileInputStream(PATH_TO_JSON_KEY))).build();
```

However, `setAuthCredentials` isn't a method on `DatastoreOptions.Builder`- it's `setCredentials`: http://googlecloudplatform.github.io/google-cloud-java/0.11.0/apidocs/com/google/cloud/datastore/DatastoreOptions.Builder.html. It also has it taking an `AuthCredentials` object- `GoogleCredentials` is probably the correct substitute here.

neozwu:
The example code is now updated.

PR #1976, opened 2017-04-24
update datastore snippet in front page

neozwu:
* Update outdated front page #1854 
* Snippet tested to compile and work
* "Try-catch block" is omitted in the snippet example


garrettjonesgoogle:
LGTM

Issue #1430, opened 2016-11-28
DEADLINE_EXCEEDED when using Cloud Pub/Sub library from Compute Engine


codemoped:
When using the Cloud Pub/Sub library in a Java application that has been deployed in a Google Compute Engine instance, a DEADLINE_EXCEEDED exception occurs. The application resides in a Docker container within a Container Engine-managed Compute Engine instance. The Pub/Sub topic and the Compute Engine instance are in the same project.

The Compute Engine service account has project editor permissions.

No problems with the old google-api-services-pubsub library.

The code to create the instance looks like this:
```java
InputStream credentialsStream = IOUtils.toInputStream(SERVICE_ACCOUNT_JSON, StandardCharsets.UTF_8);
Credentials credentials = ServiceAccountCredentials.fromStream(credentialsStream);
PubSubOptions options = PubSubOptions.newBuilder().setCredentials(credentials).build();
PubSub pubsub = options.getService();
```
Then doing something like this (or alternatively, create a topic):
```java
for (Subscription subscription : pubsub.listSubscriptions().getValues()) {
...
}
```
results in the following:
> com.google.cloud.pubsub.PubSubException: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED

Versions:
```
google-cloud-pubsub - v0.6.0
google-auth-library-oauth2-http - v0.6.0
google-auth-library-credentials - v0.6.0
google-* - v1.22.0
```

I also found this issue at [Stackoverflow](http://stackoverflow.com/questions/40554096/deadline-exceeded-when-publishing-to-a-cloud-pub-sub-topic-from-compute-engine)

colintemple:
+1. I was the one who asked the question on Stack Overflow.

In my case, the following simple code caused DEADLINE_EXCEEDED errors when run on a Compute Engine instance within the same project as the Pub/Sub topic I wanted to publish to:

    PubSub pubSub = PubSubOptions.getDefaultInstance().toBuilder()
            .build().getService();

    String messageId = pubSub.publish(topic, message);


I could communicate with Pub/Sub using google-api-services-pubsub library, or via gcloud from the command line on the instance.

garrettjonesgoogle:
I intend on investigating but it will take a couple days due to other commitments.

timkettering:
+1.  Encountering this issue with any type of call made w/ the PubSub client.

jonatanpettersson:
[...]

PR #1982, opened 2017-04-24
adding supported/unsupported platforms to README

jabubake:
#1430

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,8 +696,9 @@ Java 7 or above is required for using this client.
 Supported Platforms
 -------------------
 
-This client is supported on Mac OS X, Windows and Linux.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -693,10 +693,11 @@ Java Versions
...
 Java 7 or above is required for using this client.
 
-Android
--------
+Supported Platforms
+-------------------
 
-Android is not currently supported.
+This client is supported on Mac OS X, Windows and Linux.
+All other platforms, including Android and Alpine, are unsupported.
```
garrettjonesgoogle: Technically, Alpine is a variant of Linux. 

Also call out which environments are supported (GCE, GKE, GAE Flex) and which aren't (GAE).

Issue #1525, opened 2017-01-12
BigQuery: support labels field of Datasets and Table resources


tswast:
Docs (talks about datasets, but tables can have labels, too):
https://cloud.google.com/bigquery/docs/labeling-datasets

API references:
https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#labels
https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#labels

If the `patch` request is not yet supported, this will be needed, too.
https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/patch
https://cloud.google.com/bigquery/docs/reference/rest/v2/tables/patch

rouzwawi:
Any update on this issue?

We're looking to use the high-level `BigQuery` client to read labels for datasets. But neither `bigquery.listDatasets(...)` nor `bigquery.getDataset(...)` propagate the labels field from low level protobuf types up to `Dataset`.

I could open a PR as it seems to be a simple fix.

garrettjonesgoogle:
@rouzwawi we would welcome your contribution for this :-)

garrettjonesgoogle:
We have a project to prepare BigQuery for GA, and this feature is part of the list (P0 for dataset labels, P2 for table labels). I'm going to close this out and let the feature be tracked internally.

PR #1990, opened 2017-04-25
Expose BigQuery Dataset labels

rouzwawi:
This PR propagates the `labels` map from the underlying `com.google.api.services.bigquery.model.Dataset` payload into `DatasetInfo`. It also adds a `setLabels(Map<String, String>)` method to the corresponding builders.

Furthermore, it adds a `LABELS` value to the `DatasetField` enum. 

It does not change anything related to BigQuery Tables.

This should resolve part of #1525


rouzwawi:
I believe Spotify should already have a signed CLA.

garrettjonesgoogle:
We will wait until we have moved Datastore/Storage/Logging to 1.0.0 before looking at this.

rouzwawi:
@garrettjonesgoogle do you have a rough timeline for when that will happen?

garrettjonesgoogle:
It's scheduled for today.

NicolaSpreafico:
Hello, I can see implementation for Datasets, how about Tables?

Issue #1791, opened 2017-03-22
Remove usage of spi Tuple classes under google-cloud-examples


garrettjonesgoogle:
* https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/BigQueryExample.java uses `com.google.cloud.bigquery.spi.v2.BigQueryRpc.Tuple`
* https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java uses `com.google.cloud.storage.spi.v1.StorageRpc.Tuple`

These tuple definitions are identical; maybe we should just make a Tuple class under core for everything to share.

garrettjonesgoogle:
I'm marking this as "docs" because it is example code.

garrettjonesgoogle:
Completed in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1992

PR #1992, opened 2017-04-25
make one Tuple class aross all the projects

neozwu:
#1791


garrettjonesgoogle:
LGTM - we'll want to do another rc release with this before 1.0.0

garrettjonesgoogle:
Approving an early merge because the change is purely syntactic and all linux tests have passed.

PR #2002, opened 2017-04-26
Versioning cleanup:

vam-google:
- fix versioning script (grpc.version issue)
- better version resolution (via dependencyManagement instead of usage of properties defined in parent pom as dependencies version)
- should also fix LICENSE resolution by versioneye.com and dependency ci (google-cloud-* moudles should now display correct licenses instead of UKNOWN)
- fix link in readmes to versioneye.com (old link points to 0.7.1 which is not getting updated
- fix security issue dependency in google-cloud-dns
- removed "test" scope in dependencyManagement for various libs (considered bad practice, plus potentially causes an issue with datastore lib)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -515,7 +515,6 @@
         <groupId>com.google.api.grpc</groupId>
         <artifactId>grpc-google-iam-v1</artifactId>
         <version>${generatedProto.version}</version>
-        <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>com.google.auto.value</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -197,10 +503,21 @@
...
+      <dependency>
+        <groupId>com.google.api.grpc</groupId>
         <artifactId>proto-google-iam-v1</artifactId>
         <version>${generatedProto.version}</version>
       </dependency>
       <dependency>
+        <groupId>com.google.api.grpc</groupId>
+        <artifactId>grpc-google-iam-v1</artifactId>
+        <version>${generatedProto.version}</version>
+        <scope>test</scope>
```
garrettjonesgoogle: Shouldn't the test scope be removed here?

garrettjonesgoogle:
LGTM

PR #2003, opened 2017-04-26
Fix build setup of project.properties

shinfan:
- Add project.properties as a resource file
- Process the resource file and inject the version information in the building stage

In this way only the properties listed in the project.properties will be injected.

Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1999

### DIFF
--- a/google-cloud-core-grpc/pom.xml
+++ b/google-cloud-core-grpc/pom.xml
@@ -69,36 +69,4 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-  <build>
-    <resources>
-      <resource>
-        <directory>./..</directory>
-        <filtering>true</filtering>
-        <includes>
-          <include>project.properties</include>
-        </includes>
-      </resource>
-    </resources>
-    <plugins>
-       <plugin>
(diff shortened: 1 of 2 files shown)


google-cloud-core-grpc/pom.xml
```diff
@@ -70,22 +70,34 @@
...
-            <phase>generate-resources</phase>
-            <goals>
-              <goal>write-project-properties</goal>
-            </goals>
+            <phase>process-resources</phase>
             <configuration>
-              <outputFile>${project.build.outputDirectory}/com/google/cloud/project.properties</outputFile>
+               <tasks>
+                 <copy file="${project.build.outputDirectory}/project.properties"
+                   toFile="${project.build.outputDirectory}/com/google/cloud/project.properties"
```
garrettjonesgoogle: This is the same path as for google-cloud-core.

I think we should either put the properties files in distinct places for each module, or remove the properties from core/core-http/core-grpc.

shinfan: Remove the script from http/grpc. Done.

garrettjonesgoogle:
LGTM

PR #2006, opened 2017-04-26
Bumping to next snapshot versions

garrettjonesgoogle:


garrettjonesgoogle:
AppVeyor was failing that way yesterday too - it looks like a problem that occurs before anything in our package is executed.

Issue #1994, opened 2017-04-25
LoggingHandler throws exception on GAE Flexible "Resource gae_app has extra label: instance_name"


bmenasha:
Using the latest Logging Hander fails on App Engine Flexible because we are supplying the InstanceName label twice, likely one of them should be InstanceId:

From MonitoredResourceUtil.java

Resource.GaeAppFlex.getKey(),
 new Label[] {
   Label.InstanceName,
   Label.ModuleId,
   Label.VersionId,
   Label.InstanceName,
   Label.Zone
 })

This results in the error on GAE Flexible when using the LoggingHandler:

SEVERE: RuntimeException while executing runnable com.google.common.util.concurrent.Futures$6@76665fe6 with executor MoreExecutors.directExecutor()
java.lang.RuntimeException: com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Resource gae_app has extra label: instance_name
	at com.google.cloud.logging.LoggingImpl$7.onFailure(LoggingImpl.java:577)
	at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:52)
	at com.google.common.util.concurrent.Futures$6.run(Futures.java:1764)
	at com.google.common.util.concurrent.MoreExecutors$DirectExecutor.execute(MoreExecutors.java:456)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:817)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:753)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:634)
	at com.google.common.util.concurrent.Futures$AbstractChainingFuture.run(Futures.java:1405)
	at com.google.common.util.concurrent.MoreExecutors$DirectExecutor.execute(MoreExecutors.java:456)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:817)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:753)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:634)
	at com.google.common.util.concurrent.Futures$AbstractCatchingFuture.run(Futures.java:794)
	at com.google.common.util.concurrent.MoreExecutors$DirectExecutor.execute(MoreExecutors.java:456)
[...]

PR #2011, opened 2017-04-26
Fix : moving instance_name logging to log entry label

jabubake:
#1994  removing unsupported resource labels for standard, flex
Adding instance_name to log entries instead of resource for Flex.
@meltsufin

### DIFF
--- a/google-cloud-datastore/README.md
+++ b/google-cloud-datastore/README.md
@@ -12,9 +12,6 @@ Java idiomatic client for [Google Cloud Datastore](https://cloud.google.com/data
 -  [Homepage](https://googlecloudplatform.github.io/google-cloud-java/)
 -  [API Documentation](https://googlecloudplatform.github.io/google-cloud-java/apidocs/index.html?com/google/cloud/datastore/package-summary.html)
 
-> Note: This client is a work-in-progress, and may occasionally
-> make backwards-incompatible changes.
-
 Quickstart
 ----------
 If you are using Maven, add this to your pom.xml file
@@ -170,9 +167,7 @@ Versioning
 
 This library follows [Semantic Versioning](http://semver.org/).
 
-It is currently in major version zero (``0.y.z``), which means that anything
-may change at any time and the public API should not be considered
-stable.
+It is currently in major version one (``1.y.z``), which means that the public API should be considered stable.
 
 Contributing
 ------------
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/language/snippets/AnalyzeSentiment.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/language/snippets/AnalyzeSentiment.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * EDITING INSTRUCTIONS
+ * This file is referenced in READMEs. Any change to this file should be reflected in
+ * the project's READMEs.
+ */
+
+package com.google.cloud.examples.language.snippets;
+
+import com.google.cloud.language.spi.v1.LanguageServiceClient;
+
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.Document.Type;
+import com.google.cloud.language.v1.Sentiment;
+
+/**
+ * A snippet for Google Cloud Speech API showing how to analyze text message sentiment.
+ */
+public class AnalyzeSentiment {
+
+  public static void main(String... args) throws Exception {
+    // Instantiates a client
+    LanguageServiceClient language = LanguageServiceClient.create();
+
+    // The text to analyze
+    String[] texts = {"I love this!", "I hate this!"};
+    for (String text : texts) {
+      Document doc = Document.newBuilder().setContent(text).setType(Type.PLAIN_TEXT).build();
+      // Detects the sentiment of the text
+      Sentiment sentiment = language.analyzeSentiment(doc).getDocumentSentiment();
+
+      System.out.printf("Text: \"%s\"%n", text);
+      System.out.printf(
+          "Sentiment: score = %s, magnitude = %s%n",
+          sentiment.getScore(), sentiment.getMagnitude());
+    }
+  }
+}
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
@@ -17,6 +17,7 @@
 package com.google.cloud.examples.pubsub.snippets;
 
 import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutures;
 import com.google.cloud.pubsub.spi.v1.Publisher;
 import com.google.cloud.pubsub.spi.v1.TopicAdminClient;
 import com.google.protobuf.ByteString;
@@ -31,30 +32,54 @@ import java.util.List;
  * publish messages to it.
  */
 public class CreateTopicAndPublishMessages {
-  public static void main(String... args) throws Exception {
+
+  public static void createTopic() throws Exception {
     TopicName topic = TopicName.create("test-project", "test-topic");
     try (TopicAdminClient topicAdminClient = TopicAdminClient.create()) {
       topicAdminClient.createTopic(topic);
     }
+  }
 
+  public static void publishMessages() throws Exception {
+    // [START publish]
+    TopicName topicName = TopicName.create("test-project", "test-topic");
     Publisher publisher = null;
+    List<ApiFuture<String>> messageIdFutures = new ArrayList<>();
+
     try {
-      publisher = Publisher.defaultBuilder(topic).build();
+      // Create a publisher instance with default settings bound to the topic
+      publisher = Publisher.defaultBuilder(topicName).build();
+
       List<String> messages = Arrays.asList("first message", "second message");
-      List<ApiFuture<String>> messageIds = new ArrayList<>();
+
+      // schedule publishing one message at a time : messages get automatically batched
       for (String message : messages) {
         ByteString data = ByteString.copyFromUtf8(message);
+        // message data is converted to base64-encoding
         PubsubMessage pubsubMessage = PubsubMessage.newBuilder().setData(data).build();
+
+        // Once published, returns a server-assigned message id (unique within the topic)
         ApiFuture<String> messageIdFuture = publisher.publish(pubsubMessage);
-        messageIds.add(messageIdFuture);
-      }
-      for (ApiFuture<String> messageId : messageIds) {
-        System.out.println("published with message ID: " + messageId.get());
+        messageIdFutures.add(messageIdFuture);
       }
     } finally {
+      // wait on any pending publish requests.
+      List<String> messageIds = ApiFutures.allAsList(messageIdFutures).get();
+
+      for (String messageId : messageIds) {
+        System.out.println("published with message ID: " + messageId);
+      }
+
       if (publisher != null) {
+        // When finished with the publisher, shutdown to free up resources.
         publisher.shutdown();
       }
     }
+    // [END publish]
+  }
+
+  public static void main(String... args) throws Exception {
+    createTopic();
+    publishMessages();
   }
 }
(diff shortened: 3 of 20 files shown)


meltsufin:
@gregw PTAL

gregw:
@meltsufin I've build @jabubake's branch locally and run the https://github.com/GoogleCloudPlatform/jetty-runtime/pull/81 against it. The exception is gone and I can see the correct logs in the console - however the remote tests are failing as the logs are not being found by the API calls the test has made, so perhaps the filter is now wrong against the labels used.  

tl;dr; It's fixed the #1994 issue but I need to check if other labels are missing.

jabubake:
Greg, Mike,

The trace id label is now just trace_id ,  I noticed the naming was done
more in the format of logging  v1 labels, so I moved them to v2 style with
URL not being in the label name. I also wanted to set it standardized to
not be tied to app engine in the URL.

Let me know if this is an issue

gregw:
@jabubake Ah thanks.  Updating my tests and redeploying now...

gregw:
@jabubake @meltsufin the branch works fine with my tests now that I have updated the traceId and instance_name checks.

However, I do not know if the shorter name will prevent the log console from matching the nginx log entries (which still use the long name) with the gae_log entries.  The matching is currently not working for other reasons so I cannot test.

jabubake:
@meltsufin , @gregw  : given the complexity, I'll update this for now to be in sync with what used to exist. We can take the rename to shorter v2 type versions in a separate CL.

meltsufin:
@jabubake Sounds good, thanks!

jabubake:
@gregw @meltsufin  : let me know if this will do

jabubake:
@garrettjonesgoogle  : the changes address the bug fix to resource labels and also reverting to the  currently supported Stackdriver API name for trace_id. ptal.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -72,23 +74,23 @@ String getKey() {
     }
   }
 
+  private static String appEngineLabelPrefix = "appengine.googleapis.com/";
```
garrettjonesgoogle: Make final, use constant-naming

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -193,18 +195,48 @@ private static String getAppEngineInstanceName() {
   }
 
   private static List<LoggingEnhancer> getEnhancers(Resource resourceType) {
-    List<LoggingEnhancer> enhancers;
+    List<LoggingEnhancer> enhancers = new ArrayList<>();
     switch (resourceType) {
       // Trace logging enhancer is supported on GAE Flex and Standard.
-      case GaeAppStandard:
       case GaeAppFlex:
         enhancers = new ArrayList<>();
```
garrettjonesgoogle: This line is redundant now

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -193,18 +195,48 @@ private static String getAppEngineInstanceName() {
...
     return enhancers;
   }
+
+  /**
+   *  Adds additional resource-based labels to log entries.
+   *  Labels that can be provided with {@link MonitoredResource.Builder#addLabel(String, String)}
+   *  are restricted to a supported set per resource.
+   *  @see <a href="https://cloud.google.com/logging/docs/api/v2/resource-list">Logging Labels</a>
+   */
+  static class LabelLoggingEnhancer implements LoggingEnhancer {
```
garrettjonesgoogle: Make the class private

google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java
```diff
@@ -19,6 +19,11 @@
 /* Adds tracing support for logging with thread-local trace ID tracking. */
 public class TraceLoggingEnhancer implements LoggingEnhancer {
 
+  String prefix;
```
garrettjonesgoogle: make private

garrettjonesgoogle:
@vam-google I added some comments - could you finish the review?

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: done

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: done.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: done.

google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java
jabubake: done.

jabubake:
@vam-google ping :)

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -218,9 +221,11 @@ private static String getAppEngineInstanceName() {
    *  @see <a href="https://cloud.google.com/logging/docs/api/v2/resource-list">Logging Labels</a>
    */
   static class LabelLoggingEnhancer implements LoggingEnhancer {
```
vam-google: Why is this class here but not in in same package with `TraceLoggingEnhancer`?. `TraceLoggingEnhancer` is also used only in one place (this `MonitoredResourceUtil`). It looks like either both classes should be standalone public classes (like `TraceLoggingEnhancer` currently is), or both private static classes (like `LabelLoggingEnhancer` is declared).

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -218,9 +221,11 @@ private static String getAppEngineInstanceName() {
...
+    String prefix;
     List<Label> labels;
 
-    LabelLoggingEnhancer(List<Label> labels) {
+    LabelLoggingEnhancer(String prefix, List<Label> labels) {
+      this.prefix = prefix != null ? prefix : "";
       this.labels = labels;
```
vam-google: Storing a reference to an outside collection as a field in a class is usually undesirable (as the outside reference can be changed independently at any time, which can lead to unpredictable results, for example ConcurrentModificationException during iteration over the list).

Here, it looks like this enhancer needs a prefixed label name and a corresponding constant value (retrieved by MonitoredResourceUtil.getValue()). Both of these values are constants. I don't know how important performance is for this particular code, switching to a map (with `prefix + label.getKey()` as a key and `MonitoredResourceUtil.getValue(label)` as value), can solve the two potential issues, not affecting readability of the code: 
1) possibility of unexpected modification of the `labels` list;
2) slightly worse performance, caused by `prefix + label.getKey()` addition (note, two constants are getting added over and over again on each execution of the enhanceLogEntry method).

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -193,18 +195,47 @@ private static String getAppEngineInstanceName() {
   }
 
   private static List<LoggingEnhancer> getEnhancers(Resource resourceType) {
-    List<LoggingEnhancer> enhancers;
+    List<LoggingEnhancer> enhancers = new ArrayList<>();
```
vam-google: Please consider switching to new ArrayList<>(2), as 2 is the maximum size of the list here.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -193,18 +195,47 @@ private static String getAppEngineInstanceName() {
...
-        enhancers.add(new TraceLoggingEnhancer());
+        enhancers.add(new LabelLoggingEnhancer(
+            APPENGINE_LABEL_PREFIX, Collections.singletonList(Label.InstanceName)));
+        enhancers.add(new TraceLoggingEnhancer(APPENGINE_LABEL_PREFIX));
+        break;
+      case GaeAppStandard:
+        enhancers.add(new TraceLoggingEnhancer(APPENGINE_LABEL_PREFIX));
         break;
       default:
         enhancers = Collections.emptyList();
```
vam-google: Technically redundant

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -193,18 +195,47 @@ private static String getAppEngineInstanceName() {
   }
 
   private static List<LoggingEnhancer> getEnhancers(Resource resourceType) {
```
vam-google: Should it be build/create enhancers (not get, as it creates them every tiime)?

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -193,18 +195,47 @@ private static String getAppEngineInstanceName() {
...
+   *  Adds additional resource-based labels to log entries.
+   *  Labels that can be provided with {@link MonitoredResource.Builder#addLabel(String, String)}
+   *  are restricted to a supported set per resource.
+   *  @see <a href="https://cloud.google.com/logging/docs/api/v2/resource-list">Logging Labels</a>
+   */
+  private static class LabelLoggingEnhancer implements LoggingEnhancer {
+    String prefix;
+    List<Label> labels;
+
+    LabelLoggingEnhancer(String prefix, List<Label> labels) {
```
vam-google: Storing a reference to an outside collection as a field in a class is usually undesirable (as the outside reference can be changed independently at any time, which can lead to unpredictable results, for example ConcurrentModificationException during iteration over the list).

Here, it looks like this enhancer needs a prefixed label name and a corresponding constant value (retrieved by MonitoredResourceUtil.getValue()). Both of these values are constants. I don't know how important performance is for this particular code, switching to a map (with prefix + label.getKey() as a key and MonitoredResourceUtil.getValue(label) as value), can solve the two potential issues, not affecting readability of the code:

possibility of unexpected modification of the labels list;
slightly worse performance, caused by prefix + label.getKey() addition (note, two constants are getting added over and over again on each execution of the enhanceLogEntry method).

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -193,18 +195,47 @@ private static String getAppEngineInstanceName() {
...
     return enhancers;
   }
+
+  /**
+   *  Adds additional resource-based labels to log entries.
+   *  Labels that can be provided with {@link MonitoredResource.Builder#addLabel(String, String)}
+   *  are restricted to a supported set per resource.
+   *  @see <a href="https://cloud.google.com/logging/docs/api/v2/resource-list">Logging Labels</a>
+   */
+  private static class LabelLoggingEnhancer implements LoggingEnhancer {
```
vam-google: Why is this class here but not in in same package with TraceLoggingEnhancer?. TraceLoggingEnhancer is also used only in one place (this MonitoredResourceUtil). It looks like either both classes should be standalone public classes (like TraceLoggingEnhancer currently is), or both private static classes (like LabelLoggingEnhancer is declared).

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: Done. Semantics make sense.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: Done.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: Agree, done.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: LabelLoggingEnhancer is intended to be closely tied to monitored resource based labels and need n't be exposed. TraceLoggingEnhancer is publicly exposed to allow for trace_id labels being set from the runtimes. hence the distinction.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
jabubake: Agree, this was not a good fix given mutability and perf. I've updated this to just build the map in the constructor itself.

google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java
```diff
@@ -192,19 +196,54 @@ private static String getAppEngineInstanceName() {
...
+  /**
+   * Adds additional resource-based labels to log entries.
+   * Labels that can be provided with {@link MonitoredResource.Builder#addLabel(String, String)}
+   * are restricted to a supported set per resource.
+   *
+   * @see <a href="https://cloud.google.com/logging/docs/api/v2/resource-list">Logging Labels</a>
+   */
+  private static class LabelLoggingEnhancer implements LoggingEnhancer {
+
+    Map<String, String> labels;
```
vam-google: Please keep class fields private, unless it is absolutely necessary to do otherwise.

jabubake: Done.

PR #2014, opened 2017-04-27
Update PollingSubscriberConnection.java logging level

AntonZarutsky:
It looks like library try to recover in case of failed pull.
It make sense to change logging level to Warning in order to distinguish from critical errors.


garrettjonesgoogle:
Adding @pongad to review

pongad:
I thought I did this 😆  LGTM.

PR #2023, opened 2017-05-01
Pubsub snippets

jabubake:
Adding snippets for cloud site documentation around advanced configuration for publisher, subscriber (batch, retries, flow) + clean up on doc labels.

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
@@ -93,7 +93,7 @@ public class PublisherSnippets {
 
     Duration publishDelayThreshold = Duration.ofMillis(100); // default : 1 ms
 
-    // Batching thresholds have order precedence : request size  > message count > publish delay
+    // Publish request get triggered based on request size, messages count & time since last publish
     BatchingSettings batchingSettings = BatchingSettings.newBuilder()
         .setElementCountThreshold(messageCountBatchSize)
         .setRequestByteThreshold(requestBytesThreshold)
@@ -132,8 +132,8 @@ public class PublisherSnippets {
     int maxOutstandingBatches = 20;
     int maxOutstandingRequestBytes = 500000;
 
-    // override behavior on limits exceeded, default behavior is to throw an exception
-    LimitExceededBehavior limitExceededBehavior = LimitExceededBehavior.Block;
+    // override behavior on limits exceeded if needed, default behavior is to block
+    LimitExceededBehavior limitExceededBehavior = LimitExceededBehavior.ThrowException;
 
     FlowControlSettings flowControlSettings = FlowControlSettings.newBuilder()
         .setMaxOutstandingElementCount(maxOutstandingBatches)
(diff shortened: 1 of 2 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
```diff
@@ -54,24 +67,112 @@ public void onFailure(Throwable t) {
...
-    // [END publish]
     return messageIdFuture;
   }
 
   /** Example of creating a {@code Publisher}. */
   // [TARGET newBuilder(TopicName)]
   // [VARIABLE "my_project"]
   // [VARIABLE "my_topic"]
-  public static void newBuilder(String projectName, String topicName) throws Exception {
-    // [START newBuilder]
```
pongad: Is this tag no longer necessary?

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
```diff
@@ -54,24 +67,112 @@ public void onFailure(Throwable t) {
...
+
+  public Publisher getPublisherWithCustomBatchSettings(TopicName topicName) throws Exception {
+    // [START publisherBatchSettings]
+    // Batch settings control how the publisher batches messages
+    long requestBytesThreshold = 5000L; // default : 1kb
+    long messageCountBatchSize = 10L; // default : 100
+
+    Duration publishDelayThreshold = Duration.ofMillis(100); // default : 1 ms
+
+    // Batching thresholds have order precedence : request size  > message count > publish delay
```
pongad: I don't think this is right. We force-send an RPC if the request size grows too large, or we have too many messages, or if we've been waiting too long. I wouldn't call this a precedence: since any of them can force an RPC, they all have the same precedence.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
```diff
@@ -54,24 +67,112 @@ public void onFailure(Throwable t) {
...
+  }
+
+  public Publisher getPublisherWithCustomFlowControlSettings(TopicName topicName) throws Exception {
+    // [START publisherFlowControlSettings]
+
+    // Flow control settings restrict the number of outstanding publish requests
+    int maxOutstandingBatches = 20;
+    int maxOutstandingRequestBytes = 500000;
+
+    // override behavior on limits exceeded, default behavior is to throw an exception
```
pongad: The default is to block. Please see https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/batching/FlowControlSettings.java#L77 . I'd be OK with setting it to throw exception here though.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
jabubake: it was not used in the cloudsite docs, i cleaned up : so we know its clearly not.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
jabubake: Agree, the precedence sounds confusing, removed.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/PublisherSnippets.java
jabubake: I could have sworn I saw default as throw exception :) corrected. thanks for catching.

PR #2026, opened 2017-05-01
spanner: Enhance session pool behavior

vkedia:
- Change defaults for maxSessions and blockIfExhauseted
- Added tracking of leaked sessions
- Changed closeAsync to close leaked sessions as well.
- Changed closeAsync to close pool maintenance worker as well.
- Added a SessionPoolStressTester
- Changed Spanner#closeAsync to be blocking Spanner#close.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -749,7 +749,7 @@ final class SessionPool {
     }
     if (waiter != null) {
       logger.log(
-          Level.INFO,
+          Level.FINE,
           "No session available in the pool. Blocking for one to become available/created");
       sess = waiter.take();
     }
@@ -799,7 +799,7 @@ final class SessionPool {
     }
     if (waiter != null) {
       logger.log(
-          Level.INFO,
+          Level.FINE,
           "No session available in the pool. Blocking for one to become available/created");
       sess = waiter.take();
     }


vkedia:
@tomayles Please take a look

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -558,6 +594,9 @@ private void replenishPool() {
   @GuardedBy("lock")
   private int maxSessionsInUse = 0;
 
+  @GuardedBy("lock")
+  private Map<String, PooledSession> allSessions = new HashMap<>();
```
vkedia: This is consistent with other fields in this class. I will fix them all separately.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
```diff
@@ -0,0 +1,271 @@
...
+  private Exception getFailedError() {
+    synchronized (lock) {
+      return e;
+    }
+  }
+
+  @Test
+  public void stressTest() throws Exception {
+    int concurrentThreads = 10;
+    final int numOperationsPerThread = 1000;
```
vkedia: I think this makes sense to be in the test method as it is very specific to this test case.

vkedia:
Closed by mistake.

danielcompton:
Thanks for updating the javadocs. I'm still a little uneasy about having blocking RPC calls in the shutdown path though.

vkedia:
Thanks for your comments. It is almost essential that this method be
allowed to finish and delete all the sessions otherwise it can lead to
session lying around on the backend for upto an hour. Since there is a
[hard limit](https://cloud.google.com/spanner/docs/limits) on how many
sessions one can have on the backend, this can lead to one running out of
sessions which will cause requests to fail. That is the reason I removed
the async version. You almost always want to block for this method to
finish.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
```diff
@@ -35,8 +38,12 @@
...
-   * logging and tracking of leaked sessions.
+   * method will block till it can clean up all the resources. Specifically, it deletes all the
+   * underlying sessions (which involves rpcs) and closes all the gRPC channels. Once this method
+   * called, this object is no longer usable. It is strongly advised to call this method when you
+   * are done with the {@code Spanner} object, typically when your application shuts down. There is
+   * a hard limit on number of sessions in Cloud Spanner and not calling this method can lead to
+   * unused sessions piling up on the backend.
    */
-  ApiFuture<Void> closeAsync();
+  void close();
```
vam-google: Please consider: 
1) Making Spanner implement `java.io.Closeable` interface
2) Implementing finalize() in SpannerImpl and calling close() there.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -263,18 +260,14 @@ public DatabaseClient getDatabaseClient(DatabaseId db) {
...
-          }
-        },
-        directExecutor()));
+    try {
+      Futures.successfulAsList(closureFutures).get();
+    } catch (InterruptedException | ExecutionException e) {
+      throw SpannerExceptionFactory.newSpannerException(e);
+    }
+    for (ManagedChannel channel : getOptions().getRpcChannels()) {
+      channel.shutdown();
```
vam-google: Should the whole process fail in one channel.shutdown() fails (throws exception)?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1006,9 +1048,10 @@ public void run() {
                 if (closureFuture != null) {
                   closeSession = true;
                 } else {
-                  totalSessions++;
-                  Preconditions.checkState(totalSessions <= options.getMaxSessions());
-                  releaseSession(new PooledSession(session));
+                  Preconditions.checkState(totalSessions() <= options.getMaxSessions() - 1);
+                  PooledSession pooledSession = new PooledSession(session);
+                  allSessions.put(session.getName(), pooledSession);
```
vam-google: Are all session guaranteed to have unique names? It seem that allSessions can and should be a Set, relying on reference-equality of Session objects (PolledSession does not implement equals()/hashCode(), which fits well here).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -20,6 +20,8 @@
 
 /** Options for the session pool used by {@code DatabaseClient}. */
 public class SessionPoolOptions {
+  private static final int DEFAULT_MAX_SESSIONS = 2000;
```
tomayles: As discussed separately, this is higher than necessary given the restriction on number of outstanding RPCs per channel and a default of 4 channels. Shall we make it 400?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -20,6 +20,8 @@
...
+  private static final ActionOnExhaustion DEFAULT_ACTION = ActionOnExhaustion.FAIL;
```
tomayles: I'm not convinced by FAIL as the default: it could lead to unpredictable exceptions if the user application briefly blips up in concurrency, rather than a simple latency increase. How about keeping BLOCK, but having a log warning when this happens?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
```diff
@@ -16,10 +16,13 @@
...
 package com.google.cloud.spanner;
 
-import com.google.api.core.ApiFuture;
 import com.google.cloud.Service;
 
-/** An interface for Cloud Spanner. */
+/**
+ * An interface for Cloud Spanner. Typically, there would only be one instance of this for the
+ * lifetime of the application which must be closed by invoking {@link #close()} when it is no
+ * longer needed.
```
tomayles: + something like "Failure to do so may result in leaking session resources and exhausting session quota."

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
```diff
@@ -35,8 +38,12 @@
 
   /**
    * Closes all the clients associated with this instance and frees up all the resources. This
-   * method does not block. Return future will complete when cleanup is done. TODO(user): Add
-   * logging and tracking of leaked sessions.
+   * method will block till it can clean up all the resources. Specifically, it deletes all the
```
tomayles: Nit: till -> until

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -838,27 +888,29 @@ private void handlePrepareSessionFailure(SpannerException e, PooledSession sessi
...
-            Iterables.consumingIterable(Iterables.concat(readSessions, writePreparedSessions))) {
-          closeSessionAsync(session.delegate);
+      pendingClosure =
+          totalSessions() + numSessionsBeingCreated + 1 /* For pool maintenance thread */;
+      poolMaintainer.close();
+      readSessions.clear();
+      writePreparedSessions.clear();
+      for (final PooledSession session : ImmutableList.copyOf(allSessions.values())) {
+        if (session.leakedException != null) {
+          logger.log(Level.WARNING, "Leaked session", session.leakedException);
```
tomayles: Do you think it would be useful to expose some way of dumping this on demand? For example, if someone wanted to see what sessions were currently in use without shutting down the client entirely.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -186,10 +189,21 @@ public Timestamp getReadTimestamp() {
     }
   }
 
+  // Exception class used just to track the stack trace at the point when a session was handed out
+  // from the pool.
+  private final class LeakedSessionException extends Exception {
```
vkedia: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -186,10 +189,21 @@ public Timestamp getReadTimestamp() {
...
+    private static final long serialVersionUID = 1451131180314064914L;
+
+    private LeakedSessionException() {
+      super("Session was checked out from the pool at:");
```
vkedia: Oh, that at was intended for the stack trace that will be printed below it (and not for time). But I think adding time is a good idea.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -402,30 +417,53 @@ private Session take() throws SpannerException {
...
+                }
+              },
+              LOOP_FREQUENCY,
+              LOOP_FREQUENCY,
+              TimeUnit.MILLISECONDS);
+    }
+
+    void close() {
+      scheduledFuture.cancel(false);
+      if (!running) {
```
vkedia: The reason maintainPool needs to manage lock on its own is because we cannot hold locks for the entire duration of that method. But I went ahead and added locks to close and init as well just to be sure.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
vkedia: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
vkedia: The reason I did not make it closeable is because if people see closeable, they will think that the right thing to do is to use Spanner in a try with resource block. But that will not be a good idea, since we really want to just create one instance of spanner instead of creating and closing every time they want to make a new request to spanner.
I had thought about adding a finalize but since close can take a while depending on how many sessions you have and the concurrency of your executor, I was not sure if it is a good idea.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
vkedia: Fixed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
vkedia: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
vkedia: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
vkedia: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
vkedia: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
vkedia: I think it will be useful but as of now we don't even expose SessionPool object to the users so we will need to add this to DatabaseClient I guess. Let us discuss this more and if needed I will do in a separate change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -263,18 +260,15 @@ public DatabaseClient getDatabaseClient(DatabaseId db) {
...
-              channel.shutdown();
-            }
-            return null;
-          }
-        },
-        directExecutor()));
+    try {
+      Futures.successfulAsList(closureFutures).get();
+    } catch (InterruptedException | ExecutionException e) {
+      throw SpannerExceptionFactory.newSpannerException(e);
```
vkedia: Fixed.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
```diff
@@ -0,0 +1,271 @@
...
+    final CountDownLatch releaseThreads = new CountDownLatch(1);
+    final CountDownLatch threadsDone = new CountDownLatch(concurrentThreads);
+    final int writeOperationFraction = 5;
```
vkedia: Ditto.

vkedia:
@tomayles @vam-google Please take a look again.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Spanner.java
vam-google: Nothing prevents people from calling close() manually on every execution, without the try with resource block. If a user does not read documentation (which explicitly says here to use spanner in a singleton mode), there is a good chance that he'll just create spanner without even calling close method (so no closable, no try with resources, no manual close, and new instance of spanner on each call and an eventual program crash in a while).

I still believe that making it closable makes sense, but I do not insist.

PR #2029, opened 2017-05-03
datastore: force import grpc-google-common-protos:0.1.9

pongad:
Also removes grpc-all dependency from top-level pom.xml; it is unused.

Datastore is tested with `mvn verify`.
Depedency versions are shown to not conflict by
creating and compiling a test project that depends
only on google-cloud.
Since it does not inherit the version resolution
from google-cloud-java/pom.xml,
it more closely resembles user project.

### DIFF
--- a/google-cloud-datastore/pom.xml
+++ b/google-cloud-datastore/pom.xml
@@ -46,7 +46,7 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-common-protos</artifactId>
-      <version>0.1.9</version>
+      <version>${generatedProto.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>


google-cloud-datastore/pom.xml
```diff
@@ -38,12 +38,17 @@
...
-          <artifactId>guava-jdk5</artifactId>
           <groupId>com.google.guava</groupId>
+          <artifactId>guava-jdk5</artifactId>
         </exclusion>
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>grpc-google-common-protos</artifactId>
+      <version>0.1.9</version>
```
vam-google: Should it be ${generatedProto.version} instead (currently set to 1.9)?

pongad: You're right, done.

Deleted maven cache, and all tests still pass!

pongad:
@vam-google PTAL

google-cloud-datastore/pom.xml
```diff
@@ -38,12 +38,17 @@
...
           <groupId>com.google.guava</groupId>
+          <artifactId>guava-jdk5</artifactId>
         </exclusion>
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>grpc-google-common-protos</artifactId>
+      <version>${generatedProto.version}</version>
+    </dependency>
```
garrettjonesgoogle: I think we should move this to an exclusion under datastore-v1-proto-client and add a dependency on proto-google-common-protos.

pongad: This doesn't work. From a combination of experimenting and reading the manual, I think the fundamental reason is that dependency exclusion only excludes packages from the classpath, but version resolution still needs to work out. I think @vam-google knows Maven a lot more than I do, could you fact-check? 

The "right way" is for datastore to depend on `proto-google-common-protos`, but since it depends on `grpc`, I think we're still forced to override the dependency version.

vam-google: copy&pasting random xml from one pom.xml to another and doing "mvn install" to check if it works - this is pretty much what my maven skill is =)

But seriously speaking, I agree with @pongad.

The xsd schema for dependency <exclusion> element does not even support <version> sub-element, so I assume that exclusion should be used to exclude dependency completely, but not to modify its version. For our problem - we still need the dependency to be present, but with different version.

Modifying datastore-v1-protos-1.3.0.pom, and figuring out why this exists in that file:

```
    <!--
      Must depend on the same version of com.google.protobuf:protobuf-java
      that we do.
    -->
    <project.grpc-google-common-protos.version>0.1.0</project.grpc-google-common-protos.version>
```
 would be the right fix. But for now @pongad's fix seems good enough.

pongad:
Merging with @vam-google 's +2. @garrettjonesgoogle I'll follow up in email.

PR #2033, opened 2017-05-04
RELEASING.md: Adding verbiage about breaking changes

garrettjonesgoogle:


RELEASING.md
```diff
@@ -4,35 +4,35 @@ The release process for SNAPSHOT versions is handled by the `after_success.sh` s
 
 ### To push a release version
 
-1. Make sure the team agrees that it is time to release. Verify that all unit and integration tests for the last commit have passed.
+1. Make sure the team agrees that it is time to release. 
 
-2. Run `utilities/update_versions.sh` from the repository's base directory.
-This script takes optional arguments denoting the new versions for each qualifier (alpha, beta and/or GA). By default, if the current version is X.Y.Z-SNAPSHOT, the script will update the version in all the pom.xml and other relevant files to X.Y.Z. Please refer to the documentation in `utilities/update_versions.sh` for more details. Commit this version locally:
+2. Look over all of the commits since the last release and make sure there are no breaking changes on the public surface. If there are any breaking changes, create and merge a new PR to revert the surface back.
```
vam-google: Technically this asks to re-review all commits done since last release. This is simply impossible. Should it be instead a recommendation for each contributor before making a PR, described either in GA libraries readmes and/or in the CONTRIBUTING.md doc?

Another approach would be to add some automation on top of that (like running old tests against newer jar - we have most of that infrastructure as part of GAE work done, and it doesn't have to be automated, can be a manual run)

RELEASING.md
```diff
@@ -4,35 +4,35 @@ The release process for SNAPSHOT versions is handled by the `after_success.sh` s
...
+
+3. Verify that all unit and integration tests for the last commit have passed.
 
+4. Run `utilities/update_versions.sh` from the repository's base directory.
+This script takes optional arguments denoting the new versions for each qualifier (alpha, beta and/or GA). By default, if the current version is X.Y.Z-SNAPSHOT, the script will update the version in all the pom.xml and other relevant files to X.Y.Z. Please refer to the documentation in `utilities/update_versions.sh` for more details. Commit this version locally:
+  
   ```
   git commit -m "Release [VERSION HERE]"
```
vam-google: Is this technically incomplete, as files should also be git add'ed before commit? I usually commit from IDE anyways, so it is not a problem for me (and, assume, for others too), but still.

RELEASING.md
garrettjonesgoogle: The review should only check for changes that are observable in javadoc. No implementation changes need to be checked, no README changes need to be checked, no snippet changes need to be checked; it is just a quick scan for changes that will break compilation of users. I will clarify this.

RELEASING.md
garrettjonesgoogle: Yes you are correct, I will add a note about this.

garrettjonesgoogle:
I have addressed feedback, PTAL

PR #2037, opened 2017-05-05
Fix Datastore - call to setNamespace fails under certain Locales

vam-google:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1957

Also checked other usages of String.format(), they seem ok. One potentially problematic place is com.google.cloud.testing.Version#toString(), but it is testing code.


vam-google:
First I wanted to add a test, but then... What to test? We cannot test all possible outcomes from all possible locales. I believe what we should do is to avoid using String.format() to stings, which are consumed programmatically (like in this case). Since String.format() relies on Local, it means it is supposed to be used to construct human-readable messages only.

PR #2038, opened 2017-05-08
Remove project id requirement for creating a GCS client. 

anuraaga:
Project id is only required for creating buckets, not operating on them, so it doesn't make sense to make it required for all cases.

I currently have some code that only operates on an existing bucket, and literally set the project id to "devnull" to prevent users from having to provide a valid project id which is not necessary.


anuraaga:
@shinfan Shortened the PR title, though feel free to edit it further as you see fit.

Don't have write access so will need someone to merge :)

PR #2040, opened 2017-05-08
Cloud Spanner: Make Spanner value classes serializable

mairbek:
Is needed to get `PCollection<Struct>` in Apache Beam working.

### DIFF
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/StatementTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/StatementTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.spanner;
 import static com.google.common.testing.SerializableTester.reserializeAndAssert;
 import static com.google.common.truth.Truth.assertThat;
 
+import com.google.cloud.ByteArray;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.testing.EqualsTester;
 import org.junit.Rule;
@@ -39,6 +40,19 @@ public class StatementTest {
     assertThat(stmt.getSql()).isEqualTo(sql);
     assertThat(stmt.getParameters()).isEmpty();
     assertThat(stmt.toString()).isEqualTo(sql);
+    reserializeAndAssert(stmt);
+  }
+
+  @Test
+  public void serialization() throws Exception {
+    Statement stmt = Statement.newBuilder("SELECT * FROM table WHERE ")
+        .append("bool_field = @bool_field ").bind("bool_field").to(true)
+        .append("long_field = @long_field ").bind("long_field").to(1L)
+        .append("float_field = @float_field ").bind("float_field").to(1.)
+        .append("string_field = @string_field ").bind("string_field").to("abc")
+        .append("bytes_field = @bytes_field ").bind("bytes_field")
+        .to(ByteArray.fromBase64("abcd"))
+        .build();
     reserializeAndAssert(stmt);
   }
 


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -2522,4 +2518,44 @@ Double get(double[] array, int i) {
       return array[i];
     }
   }
+
+  private static class StringArray extends AbstractList<String> implements Serializable {
+    private static final long serialVersionUID = 695127243179520960L;
+    private final List<com.google.protobuf.Value> underlying;
+
+    private StringArray(List<com.google.protobuf.Value> underlying) {
```
shinfan: nit: change this to ListValue so it is consistent with other classes in this file.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -2522,4 +2518,44 @@ Double get(double[] array, int i) {
...
+    public int size() {
+      return underlying.size();
+    }
+  }
+
+  private static class BoolArray extends AbstractList<Boolean> implements Serializable {
+    private static final long serialVersionUID = -2850504708084921083L;
+    private final List<com.google.protobuf.Value> underlying;
+
+    private BoolArray(List<com.google.protobuf.Value> underlying) {
```
shinfan: ditto

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
mairbek: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
mairbek: done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -2474,6 +2466,8 @@ A toPrimitiveArray(int columnIndex) {
   }
 
   private static class Int64Array extends PrimitiveArray<Long, long[]> {
```
vkedia: Making all these arrays serializable in this way ties us very tightly to how they are implemented currently and makes it harder to change. For eg we might decide that we should change BooleanArray also to extend PrimitiveArray or maybe we decide that we should just simplify Int64Array and do away with all this business of bitsets. Or we might do away with all these classes entirely and just return an ArrayList<Integer>. It might be better to implement a custom serialization for GrpcStruct possibly using a serialization proxy as described in effective java:
https://books.google.com/books?id=ka2VUBqHiWkC&lpg=PA312&ots=yZLjMhu2M3&dq=serialization%20proxy&pg=PA312#v=onepage&q=serialization%20proxy&f=false

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Struct.java
```diff
@@ -41,7 +43,9 @@
  * <p>{@code Struct} instances are immutable.
  */
 @Immutable
-public abstract class Struct extends AbstractStructReader {
+public abstract class Struct extends AbstractStructReader implements Serializable {
```
vkedia: How do we ensure that if we ever add another subclass of Struct, that remains serializable?

mairbek:
Reverted Struct serialization

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Statement.java
```diff
@@ -50,7 +52,9 @@
  *
  * <p>{@code Statement} instances are immutable.
  */
-public final class Statement {
+public final class Statement implements Serializable {
+  private static final long serialVersionUID = -1967958247625065259L;
+
   private final ImmutableMap<String, Value> parameters;
```
vkedia: Value might contain a struct. And since Struct are not serializable, we cannot have Value as serializable implying Statement cannot be serializable.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Struct.java
mairbek: `Struct` is required to be s`Serializable`. In case of `GrpcStruct` it will use a regular one as a serialziation proxy.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Statement.java
mairbek: Changed `Struct` to be serializable using Serialization Proxy pattern.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Struct.java
mairbek: `GrpcStruct` is serializable. Since Struct's constructor is package protected, this should be fine.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -1829,9 +1694,85 @@ private SpannerException yieldError(SpannerException e) {
...
+          case TIMESTAMP:
+            builder.set(fieldName).to((Timestamp) value);
+            break;
+          case DATE:
+            builder.set(fieldName).to((Date) value);
+            break;
+          case ARRAY:
+            switch(fieldType.getArrayElementType().getCode()) {
+              case BOOL:
+                builder.set(fieldName).toBoolArray((Iterable<Boolean>) value);
```
vkedia: shouldnt we need to create a copy of the iterable here? In many cases this iterable are those inner classes like Int64Array etc which we did not want to serialize in the first place.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -1829,9 +1694,85 @@ private SpannerException yieldError(SpannerException e) {
...
+                builder.set(fieldName).toBytesArray((Iterable<ByteArray>) value);
+                break;
+              case TIMESTAMP:
+                builder.set(fieldName).toTimestampArray((Iterable<Timestamp>) value);
+                break;
+              case DATE:
+                builder.set(fieldName).toDateArray((Iterable<Date>) value);
+                break;
+              case STRUCT:
+                builder.add(fieldName, fieldType.getStructFields(), (Iterable<Struct>) value);
```
vkedia: Can you verify that this is recursively calling writeReplace on these inner structs as well? Maybe using a debugger.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
```diff
@@ -588,4 +593,53 @@ private static ResultSetMetadata makeMetadata(Type rowType) {
...
+        Value.stringArray(null),
+        Value.boolArray(new boolean[]{true, false}),
+        Value.boolArray((boolean[]) null),
+        Value.int64Array(new long[]{1, 2, 3}),
+        Value.int64Array((long[]) null),
+        Value.timestampArray(ImmutableList.of(Timestamp.MAX_VALUE, Timestamp.MAX_VALUE)),
+        Value.timestampArray(null),
+        Value.dateArray(ImmutableList.of(
+            Date.fromYearMonthDay(2017, 4, 17), Date.fromYearMonthDay(2017, 5, 18))),
+        Value.dateArray(null)
```
vkedia: can you create a struct array?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueTest.java
```diff
@@ -713,28 +718,56 @@ public void serialization() {
...
+    reserializeAndAssert(Value.stringArray(of));
     reserializeAndAssert(Value.stringArray(null));
 
     reserializeAndAssert(
-        Value.bytesArray(Arrays.asList(newByteArray("a"), newByteArray("b"))));
+        Value.bytesArray(BrokenSerializationList.of(newByteArray("a"), newByteArray("b"))));
     reserializeAndAssert(Value.bytesArray(null));
   }
+
+  @Test(expected = IllegalStateException.class)
```
vkedia: neat. I did not know about this usage.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/ValueTest.java
```diff
@@ -713,28 +718,56 @@ public void serialization() {
...
 
-    reserializeAndAssert(
-        Value.stringArray(Arrays.asList("a", "b")));
+    BrokenSerializationList<String> of = BrokenSerializationList.of("a", "b");
+    reserializeAndAssert(Value.stringArray(of));
     reserializeAndAssert(Value.stringArray(null));
 
     reserializeAndAssert(
-        Value.bytesArray(Arrays.asList(newByteArray("a"), newByteArray("b"))));
+        Value.bytesArray(BrokenSerializationList.of(newByteArray("a"), newByteArray("b"))));
```
vkedia: I think we had discussed that in certain cases Value does not create a copy so we might need to handle those cases in Value class.

mairbek: That was my intution as well, turns out that it does and the test verifies this. See https://github.com/GoogleCloudPlatform/google-cloud-java/blob/734a7c40ca5b35bf82e5c41e302e81e579b131bc/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java#L491

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
mairbek: These iterables are copied. `ValueTest` verifies that.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/GrpcResultSetTest.java
mairbek: Done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
mairbek: Tested with `GrpcResultSetTest#nestedStructSerialization`

google-cloud-spanner/src/test/java/com/google/cloud/spanner/StatementTest.java
```diff
@@ -38,6 +39,7 @@ public void basic() {
     assertThat(stmt.getSql()).isEqualTo(sql);
     assertThat(stmt.getParameters()).isEmpty();
     assertThat(stmt.toString()).isEqualTo(sql);
+    reserializeAndAssert(stmt);
```
vkedia: can you add a test for query with params as well.

mairbek: > Done

mairbek:
Ping?

PR #2044, opened 2017-05-09
BigQuery: add maximumBillingTier option

### LINKED ISSUE #2012: "maximumBillingTier" missing in BigQuery
The Java client library is missing the `maximumBillingTier` from the BigQuery rest API. There's no way in the SDK to set it on a per-query level for a query job configuration.

See here --> http://stackoverflow.com/questions/43647678/where-is-the-option-to-set-the-bigquery-maximum-billing-tier-in-the-new-java-sdk

And here --> https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.maximumBillingTier

It's available in the Python SDK, but not in the Java one. We need to be able to set this at a query level - **we do not want to set it project wide**.

The only workaround I see to this, is rolling back to the old version of the Java libs (the old version does expose/provide it), which is obviously a massive headache for us.

rborer:
Allow to pass the optional maximumBillingTier parameter (fixes #2012).

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
@@ -574,7 +574,8 @@ public final class QueryJobConfiguration extends JobConfiguration {
   public int hashCode() {
     return Objects.hash(baseHashCode(), allowLargeResults, createDisposition, destinationTable,
         defaultDataset, flattenResults, priority, query, tableDefinitions, useQueryCache,
-        userDefinedFunctions, writeDisposition, dryRun, useLegacySql, schemaUpdateOptions);
+        userDefinedFunctions, writeDisposition, dryRun, useLegacySql, maximumBillingTier,
+        schemaUpdateOptions);
   }
 
   @Override


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -504,6 +522,11 @@ public Boolean useLegacySql() {
   }
 
   /**
+   * Return the optional billing tier limit for this job.
+   */
+  public Integer maximumBillingTier() { return maximumBillingTier; }
```
shinfan:
```java
public Integer maximumBillingTier() {
  return maximumBillingTier; 
}
```

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -336,6 +342,17 @@ public Builder setUseLegacySql(Boolean useLegacySql) {
       return this;
     }
 
+    /**
+     * Limit the billing tier for this job. Queries that have resource usage beyond this tier will fail
```
shinfan: s/Limit/Limits/

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -504,6 +522,11 @@ public Boolean useLegacySql() {
   }
 
   /**
+   * Return the optional billing tier limit for this job.
```
shinfan: s/Return/Returns/

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -504,6 +522,13 @@ public Boolean useLegacySql() {
   }
 
   /**
+   * Returns the optional billing tier limit for this job.
+   */
+  public Integer maximumBillingTier() {
```
garrettjonesgoogle: This needs to be `getMaximumBillingTier`

shinfan: +1 Sorry I missed it

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -534,6 +559,7 @@ ToStringHelper toStringHelper() {
         .add("writeDisposition", writeDisposition)
         .add("dryRun", dryRun)
         .add("useLegacySql", useLegacySql)
+        .add("maximumBillingTier", maximumBillingTier)
```
garrettjonesgoogle: That reminds me, `hashCode` needs this too

rborer: Right, I forgot about it.

garrettjonesgoogle:
LGTM

polleyg:
Has this been merged and rolled out?

garrettjonesgoogle:
This is merged now. It should go out in the next release in the next few business days.

PR #2046, opened 2017-05-09
Pubsub snippets minor updates

jabubake:
Adding message handler in  subscriber snippet
Renaming constant project id, topic ids to be consistent across projects

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
@@ -39,7 +39,6 @@ import com.google.pubsub.v1.PubsubMessage;
 import com.google.pubsub.v1.SubscriptionName;
 import java.io.FileInputStream;
 import java.util.concurrent.Executor;
-import javax.xml.bind.DatatypeConverter;
 
 /** This class contains snippets for the {@link Subscriber} interface. */
 public class SubscriberSnippets {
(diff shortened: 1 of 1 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
```diff
@@ -86,8 +87,11 @@ private void createSubscriber() throws Exception {
     MessageReceiver receiver = new MessageReceiver() {
           @Override
           public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
-            // handle incoming message, then ack or nack the received message
-            // ...
+            // handle incoming message, then ack/nack the received message
+            System.out.println("Id : " + message.getMessageId());
+            // decode from base64 to view original data
+            System.out.println("Data : " +
+                new String(DatatypeConverter.parseBase64Binary(message.getData().toStringUtf8())));
```
pongad: Why do we need to decode base64? Is the original data base64 encoded? Pubsub works OK with binary data, doesn't it?

jabubake: When `PubSubMessage` gets constructed, we auto-encode the data to base64 : 
`PubsubMessage pubsubMessage = PubsubMessage.newBuilder().setData(data).build();` 
Should n't the user need to decode to process then ?

jabubake: Uggh, you are right. I'm sorry I got this confused with handling messages over a push endpoint where they get received base64 encoded.

jabubake: Updated. Thank you.

PR #2047, opened 2017-05-10
pubsub: make Polling work with Long element count

pongad:
Bump gax version to 1.0.1 and gax-grpc to 0.16.1.


garrettjonesgoogle:
LGTM after tests pass

PR #2048, opened 2017-05-10
bump version of api-common to 1.0.1

pongad:
This brings in the change in
https://github.com/googleapis/api-common-java/pull/28


garrettjonesgoogle:
LGTM after updating the commit description

Issue #2041, opened 2017-05-09
Pub/Sub Subscriber Deadlock w/ Executor pool size of 1


saturnism:
Continuing from #1865 - this deadlock case still exists.

saturnism:
See #1827 for another example

saturnism:
Looks like #1915 didn't separate poller executors from message dispatcher executor.

pongad:
Closed by #2049

Issue #1827, opened 2017-03-29
pubsub Limit the number of subscriber threads how?


fonzy2013:
Can someone please provide an example of limiting the number of pulling pubsub threads with the new API so we can start testing?

In the old API it was possible to limit the number of concurrent threads that pulled messages with the following code:

```java
PubSub.MessageConsumer consumer = pubsub.pullAsync(
        subscription,
        messageProcessor,
        PubSub.PullOption.maxQueuedCallbacks(maxQueuedCallbacks),
        PubSub.PullOption.executorFactory(executorFactory)
    );
```
`where maxQueuedCallbacks` was an integer amount of msgs to pull and
`executorFactory` was provided from this
```java
 /**
   * Create an ExecutorFactory instance for use with a single
   * {@link PubSub.MessageConsumer}.
   *
   * <p>The number of threads in the executor determines how many messages can be processed at one
   * time.</p>
   *
   * @see PubSub.PullOption#executorFactory(ExecutorFactory)
   */
  private static ExecutorFactory<ExecutorService> createExecutor(
      final int threads,
      final String threadNameFormat) {
    final ThreadFactory threadFactory = new ThreadFactoryBuilder()
        .setDaemon(true)
        .setNameFormat(threadNameFormat)
        .build();

    final ExecutorService executorService = Executors.newFixedThreadPool(threads, threadFactory);

    return new ExecutorFactory<ExecutorService>() {
      @Override
      public ExecutorService get() {
        return executorService;
      }

      @Override
      public void release(final ExecutorService executor) {
        executorService.shutdownNow();
      }
    };
  }
```

garrettjonesgoogle:
There is not a way to do this yet.

fonzy2013:
ok I see.

garrettjonesgoogle:
I take it back, it is possible to do this. Example:

```
ExecutorProvider executorProvider =
  InstantiatingExecutorProvider.newBuilder()
    .setExecutorThreadCount(1)
    .build();
ChannelProvider channelProvider =
  TopicAdminSettings.defaultChannelProviderBuilder()
    .setExecutorProvider(executorProvider)
    .build();
Publisher publisher =
  Publisher.newBuilder(topicName).setChannelProvider(channelProvider).build();
```

[...]

PR #2049, opened 2017-05-10
pubsub: use one thread to start/stop connections

pongad:
Previously we submit a bunch of tasks which is unnecessary
and deadlock-prone.

This does not solve the one-thread-deadlock problem,
but it changes the symptom.
In my testing, Subscriber now successfully pulls
one batch of messages before failing due to
DEADLINE_EXCEEDED error.
This is obviously undesirable but at least it doesn't
deadlock anymore.

This is my best attempt at reconstructing the events:

1. A bunch of tasks get registered to [GetSubscription](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/9623994f58199c79ee5b9f99ad0ff6d7fb69bd84/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java#L105).

2. At least one of these tasks succeed and continue to
   [pull messsages](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/9623994f58199c79ee5b9f99ad0ff6d7fb69bd84/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java#L132).

3. Something somewhere register a very long-running task
   (~30sec on my machine) to the pool.
   Stack trace (obtained by `jstack`) revealed that the pool is
   busy doing something, but I don't know what it is yet.

4. A message is pulled and the task of printing it to the screen
   is registered to the pool.
   It cannot run yet because the long-running task is using the
   only thread available.

5. At least one of tasks that called GetSubscription timed out.

6. Long running task timed out or completed.

7. Message finally printed to the screen.

8. Because time out on GetSubscription is not considered retryable,
   the Subscriber fails. It only failed after 30sec because the
   task to fail it is also blocked behind the long-running task.

Updates #1827 and #2041.

@davidtorres Please take a look.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -409,8 +409,9 @@ public class Subscriber extends AbstractApiService {
       try {
         subscriber.awaitTerminated();
       } catch (IllegalStateException e) {
-        // It is expected for some connections to be already in state failed so stop will
-        // throw this expection.
+        // If the service fails, awaitTerminated will throw an exception.
+        // However, we could be stopping services because at least one
+        // has already failed, so we just ignore this exception.
       }
     }
   }


saturnism:
are there anything else that could get onto the worker executors? the fact that polling and processing are in the same executor makes this a bit more deadlock prone

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -415,26 +402,16 @@ private void stopConnections(List<? extends ApiService> connections) {
...
-    try {
-      connectionsStopping.await();
-    } catch (InterruptedException e) {
-      throw new IllegalStateException(e);
+    for (ApiService subscriber : liveConnections) {
+      try {
+        subscriber.awaitTerminated();
+      } catch (IllegalStateException e) {
+        // It is expected for some connections to be already in state failed so stop will
+        // throw this expection.
```
garrettjonesgoogle: Why do you say `stop` will throw this exception, but the try/catch is around `awaitTerminated`?

pongad: I see the confusion. Updating comments.

pongad:
@garrettjonesgoogle PTAL

@saturnism I think them sharing an executor is a red herring. The behavior documented in the commit message is not symptomatic of a deadlock but a starvation.

PR #2055, opened 2017-05-12
pubsub: start connections slowly

pongad:
Starting a new connection seems to add a task which blocks
a thread in our thread pool.
It does not block for long, but Subscriber starts many connections,
slamming the pool.
If the number of threads the pool is low, the pool is blocked for
a long time, causing strange delays and RPC failures.

This PR alleviates this problem by starting connections one at a time.
- Each connection now make one successful call before declaring itself running.
- The Subscriber waits for a connection to be in running state
  before starting another.

Together, these changes limit the number of blocking tasks
registered to the pool.

Since starting the Subscriber can now take a significant amount of time,
and `doStart` should return promptly,
a new thread is used to start up the connections.

cc @davidtorres @saturnism

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -243,6 +243,12 @@ public class Subscriber extends AbstractApiService {
       throw new IllegalStateException(e);
     }
 
+    // When started, connections submit tasks to the executor.
+    // These tasks must finish before the connections can declare themselves running.
+    // If we have a single-thread executor and call startPollingConnections from the
+    // same executor, it will deadlock: the thread will be stuck waiting for connections
+    // to start but cannot start the connections.
+    // For this reason, we spawn a dedicated thread. Starting subscriber should be rare.
     new Thread(new Runnable() {
       @Override
       public void run() {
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -244,10 +243,17 @@ public void close() {
...
+      @Override
+      public void run() {
+        try {
+          startPollingConnections();
+          notifyStarted();
+        } catch (Throwable t) {
+          notifyFailed(t);
+        }
+      }
+    }).start();
```
garrettjonesgoogle: Would anyone find it objectionable for us to use thread pools for almost everything, but then start a raw thread for this?

davidtorres: IMHO it be better if we use the "system/library" executor for this. But since starting subscribers should be an operation that is not often executed, I'm fine if we spawn a separate thread on every execution.

pongad: Wrote comments explaining the dedicated thread.

pongad:
@garrettjonesgoogle @davidtorres PTAL

garrettjonesgoogle:
LGTM

PR #2062, opened 2017-05-15
add-max-retry-property-support

AntonZarutsky:
RetrySettings contain property `maxAttempt` which is not used by Publisher.
Added support for maxAttempt, which gives developer control a bit more over retries.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
@@ -509,7 +509,6 @@ public class Publisher {
    * Constructs a new {@link Builder} using the given topic.
    *
    * <p>Example of creating a {@code Publisher}.
-   *
    * <pre>{@code
    * String projectName = "my_project";
    * String topicName = "my_topic";
@@ -522,6 +521,7 @@ public class Publisher {
    *   publisher.shutdown();
    * }
    * }</pre>
+   *
    */
   public static Builder defaultBuilder(TopicName topicName) {
     return new Builder(topicName);


google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/spi/v1/PublisherImplTest.java
```diff
@@ -343,6 +420,8 @@ public void testPublishFailureRetries_nonRetryableFailsImmediately() throws Exce
     }
   }
 
+
+
```
pongad: Unnecessary new lines?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -16,29 +16,27 @@
...
-import com.google.pubsub.v1.PublishRequest;
-import com.google.pubsub.v1.PublishResponse;
-import com.google.pubsub.v1.PublisherGrpc;
-import com.google.pubsub.v1.PubsubMessage;
-import com.google.pubsub.v1.TopicName;
+import com.google.pubsub.v1.*;
 import io.grpc.ManagedChannel;
 import io.grpc.Status;
+import org.threeten.bp.Duration;
+
```
pongad: We try to conform to [Google Java style](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) when modifying code. Could you try running [`google-java-format`](https://github.com/google/google-java-format)? If you use the provided `script/google-java-format-diff.py`, you can avoid changing lines you don't modify.

AntonZarutsky: was formatted by my default intellij formatter. will update

PR #2083, opened 2017-05-22
More aggressive retry strategy for NIO

jean-philippe-martin:
(but only when OptionMaxChannelReopens is set)

Testing shows that sometimes storage will fail with one of the errors below when load is high. In these situations, the ChannelReopen policy is to attempt to re-open the channel after a delay.

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
+++ b/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
@@ -126,6 +126,12 @@ final class CloudStorageReadChannel implements SeekableByteChannel {
           } else if (exs.isRetryable() || exs.getCode() == 500 || exs.getCode() == 503) {
             retries++;
             if (retries > maxRetries) {
+              // this exception will be marked as retriable in most cases since
+              // it's based on the code. It may be confusing to see a retriable error
+              // that says "all retries failed" but understand this to mean:
+              // "While in principle you should be able to retry, we already did that
+              // for you a few times and it still didn't work so we wouldn't recommend
+              // further retries."
               throw new StorageException(exs.getCode(), "All retries failed", exs);
             }
             sleepForAttempt(retries);


google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -109,19 +109,24 @@ public int read(ByteBuffer dst) throws IOException {
           amt = channel.read(dst);
           break;
         } catch (StorageException exs) {
-          if (exs.getMessage().contains("Connection closed prematurely") && reopens < maxChannelReopens) {
-            // this error isn't marked as retryable since the channel is closed;
+          if (reopens < maxChannelReopens && (
+              exs.getMessage().contains("Connection closed prematurely")
+          || exs.getCause() instanceof javax.net.ssl.SSLHandshakeException
```
droazen: Should we trap the parent class exception `javax.net.ssl.SSLException` here instead of the more specific `javax.net.ssl.SSLHandshakeException`? I think we may have seen the more broad `javax.net.ssl.SSLException` in a few stack traces.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -109,19 +109,24 @@ public int read(ByteBuffer dst) throws IOException {
...
+          || exs.getCause() instanceof java.io.EOFException
+          || exs.getCause() instanceof java.net.SocketTimeoutException
```
droazen: Can we add `java.net.SocketException`? I believe we've seen `java.net.SocketException: Connection reset`

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -109,19 +109,24 @@ public int read(ByteBuffer dst) throws IOException {
...
+          )) {
```
droazen: Is it sufficient to check only `exs.getCause()`, or would it be more robust to check `exs.getCause().getCause()`, `exs.getCause().getCause().getCause()`, etc., down to the bottom of the chain?

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -109,19 +109,24 @@ public int read(ByteBuffer dst) throws IOException {
           amt = channel.read(dst);
           break;
         } catch (StorageException exs) {
-          if (exs.getMessage().contains("Connection closed prematurely") && reopens < maxChannelReopens) {
-            // this error isn't marked as retryable since the channel is closed;
+          if (reopens < maxChannelReopens && (
+              exs.getMessage().contains("Connection closed prematurely")
```
droazen: Is inspecting the actual text of the error message the only way to check for this particular condition, or is there a corresponding specific subclass of `StorageException` we could look for?

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -109,19 +109,24 @@ public int read(ByteBuffer dst) throws IOException {
...
             innerOpen();
             continue;
-          } else if ((exs.getCode() == 500 || exs.getCode() == 503)  && retries < maxRetries) {
+          } else if ((exs.isRetryable() || exs.getCode() == 500 || exs.getCode() == 503)  && retries < maxRetries) {
             retries++;
             sleepForAttempt(retries);
             continue;
           }
-          throw exs;
+          throw new StorageException(exs.getCode(), "Retry failed", exs);
```
droazen: Does this imply "All retries failed"?

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: Sure, ok.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: Sure, ok.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -136,8 +141,13 @@ public int read(ByteBuffer dst) throws IOException {
   }
 
   private void sleepForAttempt(int attempt) {
+    long delay = 500;
+    // exponential backoff
```
droazen: With maxChannelReopens == 3 and the exponential backoff as implemented here, will a starting delay of 500 typically provide enough backoff to recover from these errors, according to your testing?

droazen:
Added my feedback @jean-philippe-martin. Thanks for doing this, it will (hopefully) solve a major issue for us with our use of the gcloud NIO functionality!

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: I don't think that's necessary, I haven't seen anything like that. To put you at ease, though, I've added a check one level deeper.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: Yes. That's a better text, I'm putting that in :)

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
droazen: I ask because some of our stack traces show nesting of the `StorageException 3-4 layers deep (https://github.com/broadinstitute/gatk/issues/2685)

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: It has, yes, though I haven't been able to run the actual GenomicsDBImport. I was relying on someone else to run through my [test instructions](https://github.com/broadinstitute/gatk/issues/2685#issuecomment-302798685) before sending the PR for review, but the timeline's been compressed.

For extra peace of mind I've doubled the initial delay to one second. These errors represent abnormal behavior so we can make a case that a long delay is reasonable.

droazen: :+1:

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: That one's a bit old, I don't remember what was the exact error scenario. I looked into the original PR but it didn't have more information. Ideally I'd like to check for a specific class but I'm pretty sure that if I put in the text check it was because there was no other option.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +136,25 @@ public int read(ByteBuffer dst) throws IOException {
     }
   }
 
+  private static boolean isReopenable(Throwable exs) {
+    return exs != null && (exs.getMessage().contains("Connection closed prematurely")
+            || exs.getCause() instanceof javax.net.ssl.SSLException
+            || exs.getCause() instanceof java.io.EOFException
+            || exs.getCause() instanceof java.net.SocketException);
```
droazen: I think we need both `java.net.SocketException` and `java.net.SocketTimeoutException`, since the latter is not a subclass of the former.

droazen: (This is also consistent with a working trap we have for these exceptions in a different codebase here: https://github.com/broadinstitute/cromwell/blob/develop/engine/src/main/scala/cromwell/engine/io/IoActor.scala#L158)

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -101,27 +101,28 @@ public int read(ByteBuffer dst) throws IOException {
...
       while (true) {
         try {
           dst.reset();
           amt = channel.read(dst);
           break;
         } catch (StorageException exs) {
-          if (exs.getMessage().contains("Connection closed prematurely") && reopens < maxChannelReopens) {
-            // this error isn't marked as retryable since the channel is closed;
-            // but here at this higher level we can retry it.
+          if (reopens < maxChannelReopens && (isReopenable(exs) || isReopenable(exs.getCause()))) {
```
droazen: Reiterate my recommendation to iterate all the way down the chain of `getCause()` to the bottom, as it's not clear to me at all that the exceptions `isReopenable()` checks for will be at or near the top of the pile.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: my bad! Fixing.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: OK, if you insist.

droazen: Thanks -- as a nice side effect, now that you're iterating down the chain in `isReopenable()`, you shouldn't need to call both `isReopenable(exs)` and `isReopenable(exs.getCause())` here -- `isReopenable(exs)` alone should suffice.

jean-philippe-martin: oh right, sorry, forgot to change that. Fixed.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +135,33 @@ public int read(ByteBuffer dst) throws IOException {
     }
   }
 
+  private static boolean isReopenable(Throwable exs) {
+    Throwable throwable = exs;
+    while (throwable != null) {
+      if (throwable.getMessage().contains("Connection closed prematurely")
+          || throwable.getCause() instanceof javax.net.ssl.SSLException
```
vam-google: Is there any particular reason to use fully qualified names here? If no, please change to importing. Also, please consider assigning `Throwable cause = throwable.getCause()` and than use it like `cause instanceof SSLException`, just to make the huge if statement less scary (if you choose to do so, please also change line 148 to `throwable = cause`). 

Also, even though it is very unlikely, but there is a small chance that throwable-cause may form an infinite loop (if any child eventually becomes a parent of any of its ancestors).

droazen:
:+1: Latest version looks good to me -- hope we can get this merged soon!

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -101,27 +100,28 @@ public int read(ByteBuffer dst) throws IOException {
       checkOpen();
       int amt;
       int retries = 0;
-      int maxRetries = 3;
+      int maxRetries = Math.max(3, maxChannelReopens);
```
vam-google: What was the resason for this change? Seems reasonable, but now I'm wondering how big maxChannelReopens can be in typical case?

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +135,33 @@ public int read(ByteBuffer dst) throws IOException {
...
+    }
+    return false;
+  }
+
   private void sleepForAttempt(int attempt) {
+    long delay = 1000;
+    long maxDelay = 120_000;
+    // exponential backoff
+    for (int i=0; i<attempt; i++) {
+      delay *= 2;
```
vam-google: Not sure that 2 as an exponential backoff factor is good - it is too aggressive. Across google-cloud-java we usually use backoff factor of 1.2 or 1.3.

Also, it is probably better to calculate delay on each iteration, to avoid this loop on each wait. For example, the essential changes would be: 
- switch to `private void sleepForAttempt(double delay)`; 
- `delay = Math.min(delay * 1.3, maxDelay);` 
- `Thread.sleep((long) delay)`

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +135,33 @@ public int read(ByteBuffer dst) throws IOException {
...
+      throwable = throwable.getCause();
+    }
+    return false;
+  }
+
   private void sleepForAttempt(int attempt) {
+    long delay = 1000;
+    long maxDelay = 120_000;
+    // exponential backoff
+    for (int i=0; i<attempt; i++) {
```
vam-google: Please use proper formatting: `for (int i = 0; i < attempt; i++)`

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
vam-google: Just a note, if you choose to keep 2 as a backoff factor. For the current implementation, if the goal is to achieve `delay = 1000 * 2 ^ attempt`, there is no need to have a loop:
`delay = 1000 * (2 << attempt)`;

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: Personal preference only. I switched to imports and factored out getCause().

I also added a loop variant to ensure loop termination.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: done.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: (see below)

jean-philippe-martin: 2 is aggressive, yes, but this code is handling an uncommon scenario: we've been throttled to the point that we're not even getting a retryable 503, but something cut the connection entirely. It feels appropriate to have an uncommon backoff amount.

I changed the delay computation to match your latest suggestion. I like it best because this way we don't need to add more state to the CloudStorageReadChannel.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: maxChannelReopens is 0 by default, 3 in GATK (the application that's motivating this code in the first place). If a user is willing to reopen the channel more often, then it's probably also OK to retry normal errors more often.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +140,31 @@ public int read(ByteBuffer dst) throws IOException {
...
+      throwable = cause;
+    }
+    return false;
+  }
+
   private void sleepForAttempt(int attempt) {
+    long firstdelay = 1000;
+    long maxDelay = 120_000;
+    // exponential backoff
+    long delay = firstdelay * Math.min(1 << attempt, maxDelay);
```
vam-google: Sorry, this is not exactly what I suggested. If `Math.min(1 << attempt, maxDelay)` returns maxDelay (120,000), then `delay = 1000 * 120,000 = 120,000,000 milliseconds = 33.33 hours (!)`. 

If you need delay to be bounded by maxDelay, then:
```java
long delay = 1000L * (1L << Math.min(7, attempt)); // /2 ^ 7 = 128, closest to 120, so the maximum resulting delay will be 128_000, or 1000 * 2 ^ 7
```
note, the `firstdelay` and `maxDelay` seem redundant (also, pelase note that in the current code they are non-final constants).

or, if you really-really need `maxDelay` to be exactly 120_000:

```java
long delay = 1000L * (1L << Math.min(20, attempt));  //we need to bound power to some reasonable limit, to avoid the shift overflowing number of bits in long
delay = Math.min(delay, 120_000L);
```

vam-google: Also please add a comment about why an aggressive factor of 2 is appropriate here.

jean-philippe-martin: Sorry, I was trying to go too fast.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +140,29 @@ public int read(ByteBuffer dst) throws IOException {
...
+    }
+    return false;
+  }
+
   private void sleepForAttempt(int attempt) {
+    // exponential backoff, but let's bound it around 2min.
+    long delay = 1000L * (1L << Math.min(attempt, 7));
     try {
-      Thread.sleep((attempt - 1) * 500);
+      Thread.sleep(delay);
```
vam-google: It looks like the previous implementation would sleep 0 milliseconds for the first time (attempt - 1). That seems weird, so I like the new version more, but mentioning it here just in case.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +140,29 @@ public int read(ByteBuffer dst) throws IOException {
     }
   }
 
+  private static boolean isReopenable(Throwable exs) {
+    Throwable throwable = exs;
+    // ensures finite iteration
+    int maxDepth = 10;
+    while (throwable != null && maxDepth-- > 0) {
+      Throwable cause = throwable.getCause();
+      if (throwable.getMessage().contains("Connection closed prematurely")
```
vam-google: Sorry, I did not notice it during first iteration: this condition is independent of the loop, so please take it out and place before the while loop.

jean-philippe-martin: I moved the getMessage() test out of the loop, but then the test fails; this suggests that the message text does not contain the text for every cause exception. I moved the check back in the loop and tests pass again.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: Yes indeed, I noticed the same thing and decided we wanted to wait for every retry, not just starting at the second one.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: I see what may have made the code confusing. I simplified it (the only downside is we'll check the type of the outer exception which is redundant since it's a StorageException, but that very small performance cost is worth it from how it eases maintenance, I think).

jean-philippe-martin:
@vam-google I'm not sure which comment you mean. What I see is the last comment on the if statement is the concern about the loop not terminating but I have addressed that with `maxDepth`. I apologize, but could you please point me to the part that still needs a fix?

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -135,9 +146,29 @@ public int read(ByteBuffer dst) throws IOException {
     }
   }
 
+  private static boolean isReopenable(Throwable exs) {
+    Throwable throwable = exs;
+    // ensures finite iteration
+    int maxDepth = 10;
+    while (throwable != null && maxDepth-- > 0) {
+      if (throwable.getMessage().contains("Connection closed prematurely")
```
vam-google: Just in case, please also change this line to
(throwable.getMessage() != null && throwable.getMessage().contains("Connection closed prematurely")), as exceptions are not required to provide messages.

vam-google:
@jean-philippe-martin worry, I meant the comment about extracting check from the loop. But I was wrong about it, wo it is ok. But I added one more comment. Please address it, and then LGTM.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: Good call, done!

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
```diff
@@ -101,26 +105,33 @@ public int read(ByteBuffer dst) throws IOException {
...
+              throw new StorageException(exs.getCode(), "All reopens failed", exs);
+            }
             sleepForAttempt(reopens);
             innerOpen();
             continue;
-          } else if ((exs.getCode() == 500 || exs.getCode() == 503)  && retries < maxRetries) {
+          } else if (exs.isRetryable() || exs.getCode() == 500 || exs.getCode() == 503) {
             retries++;
+            if (retries > maxRetries) {
+              throw new StorageException(exs.getCode(), "All retries failed", exs);
```
vam-google: Ok, it is a new change... In most cases, this new StorageException will be initialized as retriable (.isRetryable() will return true, and currently there is no way to set retriable property explicitly, it is set based on the error code). It can be confusing for users (i.e. a code, after failed retries throwing an exception which says that it is still "retriabel == true"). Maybe at least document this. 

Also (not related to your change), isRetryable seems like a misspelling from our side (though I'm not even sure than either "retryable" or "retriable" exist in English dictionary at all=)

vam-google:
Ok, I'm going to merge this pull request. @jean-philippe-martin please confirm that you are done with your changes.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannel.java
jean-philippe-martin: I'm not sure on the spelling either! I added a comment to document this potentially confusing case.

jean-philippe-martin:
Now that I added the comment, I'm done! Thanks for the review and for doing the merge (since I cannot). I suppose we want to wait for the Travis test to pass, though, yes?

jean-philippe-martin:
Great, Travis is green! @vam-google I leave it to you to merge in the code. Thank you very much!

vam-google:
Merged. Thanks @jean-philippe-martin for your work!

PR #2088, opened 2017-05-26
pubsub add Subscriber.Builder.setCredentialsProvider

pongad:
@garrettjonesgoogle because `startPollingConnections` throws `IOException`, code dealing with streaming doesn't compile. Since the code is unused and untested, I thought they'd best be commented out with the expectation that we'll re-enable and re-test them later. Please let me know if this is sounds good to you. Note that when we enable streaming, we'll have to modify it anyway so that it doesn't fall back to polling.


garrettjonesgoogle:
LGTM

PR #2102, opened 2017-06-01
pubsub: remove publisher flowcontrol

pongad:


pongad:
@jabubake I need to remove `getPublisherWithCustomFlowControlSettings` because flowcontrol on publisher is going away. I can't find reference to it in this repo; is it used somewhere else?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java
```diff
@@ -179,10 +169,7 @@ public TopicName getTopicName() {
    * Schedules the publishing of a message. The publishing of the message may occur immediately or
    * be delayed based on the publisher batching options.
    *
-   * <p>Depending on chosen flow control {@link FlowControlSettings#getLimitExceededBehavior
-   * option}, the returned future might immediately fail with a {@link
-   * FlowController.FlowControlException} or block the current thread until there are more resources
-   * available to publish.
+   * <p>{@code publish} does not attempt to limit the number of pending messages.
```
garrettjonesgoogle: This doesn't seem worth saying at all.

PR #2104, opened 2017-06-01
pubsub: start subscriber connections quickly again

pongad:
This PR undoes part of #2055.
Refreshing credentials blocks thread for a period of time.
Previously, each channel has its own credentials,
so starting many channels spams executor with many blocking tasks.
but the downside is we take a long time to fully start:
about 2 minutes on my machine.

Now, all connections share the same credentials,
so refreshing only happens once.
We can start quickly again.

Also rename setLeaseAlarmsExecutorProvider,
since the executor is now also used for polling.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
@@ -335,9 +335,10 @@ public class Subscriber extends AbstractApiService {
       List<? extends ApiService> connections, final ApiService.Listener connectionsListener) {
     for (ApiService subscriber : connections) {
       subscriber.addListener(connectionsListener, executor);
-      // Starting each connection submits a blocking task to the executor.
-      // We start connections one at a time to avoid swamping executor with blocking tasks.
-      subscriber.startAsync().awaitRunning();
+      subscriber.startAsync();
+    }
+    for (ApiService subscriber : connections) {
+      subscriber.awaitRunning();
     }
   }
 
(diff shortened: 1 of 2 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
```diff
@@ -468,10 +469,10 @@ public Builder setCredentialsProvider(CredentialsProvider credentialsProvider) {
     }
 
     /**
-     * Gives the ability to set a custom executor for managing lease extensions. If none is provided
-     * a shared one will be used by all {@link Subscriber} instances.
+     * Gives the ability to set a custom executor for polling and managing lease extensions. If none
+     * is provided a shared one will be used by all {@link Subscriber} instances.
      */
-    public Builder setLeaseAlarmsExecutorProvider(ExecutorProvider executorProvider) {
+    public Builder setSystemExecutorProvider(ExecutorProvider executorProvider) {
```
garrettjonesgoogle: I agree that this should be renamed from "LeaseAlarms," but I don't know if "System" is the right term here.

pongad: I mainly went with this based on [David's suggestion](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2079#discussion_r118393132). Feels suboptimal, but I struggle to think of a better name.

The way I think about it: the "time" command gives you "user" and "system" time, vaguely split into (1) "what you want to do" and (2) "whatever the kernel needs to do so you can do 1". To us, it's (1) "what you want to do with the messages" and (2) "whatever we need to do so you can do 1". The word "system" is vague because the user doesn't generally need to care whatever it is that we do, the same way I don't generally care what the kernel does to read me the file.

Am I making any sense?

garrettjonesgoogle: Makes enough sense. I guess we can go with it :-)

garrettjonesgoogle:
LGTM

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Subscriber.java
pongad: :laughing:

Issue #2103, opened 2017-06-01
Cannot stop subscriber


pongad:
If we create a subscriber using with `FixedChannelProvider` and `FixedExecutorProvider` and do this
```java
subscriber.startAsync().awaitRunning();
System.err.println("running");
subscriber.stopAsync().awaitTerminated();
System.err.println("stopped");
for (;;) {
  Thread.sleep(1000);
}
```
we keep pulling messages forever. Stopping subscriber flushes waits for all messages to finish processing, but it does not stop the polling chain from continuing.

pongad:
The polling implementation is being removed in favor of streaming, and streaming shouldn't have this problem.

PR #2113, opened 2017-06-05
pubsub: eventually stop polling after termination

pongad:
Update #2103.

This does not completely fix the problem.
If an RPC is started before the shutdown, the RPC won't be canceled.
Depending on how the channels and executors are set up,
messages in the said RPCs might be lease-extended.

With this PR, at least we're not polling forever anymore.

@davidtorres

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
@@ -204,7 +204,7 @@ final class PollingSubscriberConnection extends AbstractApiService implements Ac
   }
 
   private boolean isAlive() {
-    // Read state only once. Because of threading, different calss can give different results.
+    // Read state only once. Because of threading, different calls can give different results.
     State state = state();
     return state == State.RUNNING || state == State.STARTING;
   }


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
```diff
@@ -201,7 +204,9 @@ public void run() {
   }
 
   private boolean isAlive() {
-    return state() == State.RUNNING || state() == State.STARTING;
+    // Read state only once. Because of threading, different calss can give different results.
```
garrettjonesgoogle: spelling(calss)

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
```diff
@@ -127,6 +127,9 @@ protected void doStop() {
   }
 
   private void pullMessages(final Duration backoff) {
+    if (!isAlive()) {
```
davidtorres: Good catch thanks!

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
pongad: Done

PR #2116, opened 2017-06-06
regenerate from toolkit

pongad:
New feature: easier credentials customization


garrettjonesgoogle:
Merging now - mac is failing, but before it even gets to run anything.

PR #2124, opened 2017-06-07
Fix Files.walk behavior

jean-philippe-martin:
Java's Files.walk() ends up in an infinite loop when iterating
over files on GCS because the iterator returns the directory
itself as the first result, so Files.walk() tries to cd there
and we're stuck.

The fix is to check for that path and not return it.

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
+++ b/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
@@ -99,7 +99,8 @@ public final class CloudStorageFileSystemProvider extends FileSystemProvider {
     private final CloudStorageFileSystem fileSystem;
     private final String prefix;
 
-    LazyPathIterator(CloudStorageFileSystem fileSystem, String prefix,
+    LazyPathIterator(CloudStorageFileSystem fileSystem,
+                     String prefix,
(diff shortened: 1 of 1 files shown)


jean-philippe-martin:
Travis error was a transient, nothing to do with the code:

    The command "git clone -b travis `git config --get remote.origin.url` target/travis" failed and exited with 128 during .

neozwu:
@jean-philippe-martin Thanks for fixing this.  It would have been better if this [code](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L279) could include the original `prefix` in the returned tuple, so the caller could easily remove the `prefix` (i.e. the start directory) from the combined list of `items` and `prefixes`.  
@garrettjonesgoogle Could you please also take a look ?

jean-philippe-martin:
@neozwu Are you suggesting changing the StorageRpc interface ? That would impact a lot of code I'm not very familiar with, and it's a public interface so clients might rely on it. I agree that it would be nice if it had been that way from the start and perhaps that's what you're saying?

jean-philippe-martin:
ah ok @neozwu, that makes sense now! Thank you!

droazen:
@garrettjonesgoogle @neozwu Anything more need to be done to get this one merged? This is a pretty serious bug that manifests itself in hard lockups on our end (as a client of `google-cloud-java`), so we'd be very glad to see this fixed in the next release.

PR #2127, opened 2017-06-08
pubsub: GetSubscription only once

### LINKED ISSUE #2111: Hitting quota limit when starting a large number of pub/sub subscribers simultaneously
I have been spinning up a lot of pods within large node-pools in ContainerEngine.  When the pods start subscribing to pub/sub I often get 
`RESOURCE_EXHAUSTED: Insufficient tokens for quota 'administrator' and limit 'CLIENT_PROJECT-100s'`
I *think* this is due to the subscriber starting multiple channels per core.  First, isn't core-count improperly biased in kubernetes - I have a machine that has 16 cores but might be sharing them with other pods. My use case results in 10-20 minute of processing per pub/sub message when using 16 cores. Isn't 160 channels overkill?  I am assuming that starting 10,000 to 20,000 channels simultaneously the reason I'm hitting the quota.  Is the large number of channels for use cases that need to consume large number of messages per second?

I'd like to add a configuration parameter to limit the number of channels.  Are the other considerations for having large numbers of channels?  Perhaps a minimum and/or maximum number of channels?

I can create a pull request but I'm relatively new to the pub/sub code base and wanted to make sure I was understanding things appropriately.

pongad:
Previous each subscriber connection makes one GetSubscription
call to get subscription detail.
This spams the service with many Get requests,
that will likely all return the same information.

This PR makes Subscriber call Get only once,
before starting all connections.

Fixes #2111.

cc @davidtorres

### DIFF
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/FakeSubscriberServiceImpl.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/FakeSubscriberServiceImpl.java
@@ -17,8 +17,8 @@
 package com.google.cloud.pubsub.v1;
 
 import com.google.api.client.util.Preconditions;
-import com.google.common.collect.ImmutableList;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.google.protobuf.Empty;
 import com.google.pubsub.v1.AcknowledgeRequest;
 import com.google.pubsub.v1.GetSubscriptionRequest;
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
@@ -47,7 +47,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.LinkedBlockingQueue;
-import org.threeten.bp.Duration;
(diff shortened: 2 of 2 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/PollingSubscriberConnection.java
```diff
@@ -46,23 +45,24 @@
  * Implementation of {@link AckProcessor} based on Cloud Pub/Sub pull and acknowledge operations.
  */
 final class PollingSubscriberConnection extends AbstractApiService implements AckProcessor {
+  static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(10);
```
pongad: Done.

pongad:
@davidtorres Sorry for the delay. PTAL

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/FakeSubscriberServiceImpl.java
```diff
@@ -234,6 +236,10 @@ public void getSubscription(
     responseObserver.onCompleted();
   }
 
+  public int getSubscriptionCalledNum() {
```
davidtorres: nit: getSubscriptionCalledCount

davidtorres: Could please add a comment, this is exposed for testing purposes

Issue #2123, opened 2017-06-07
Use of System ClassLoader didn't find custom logging enhancer


dclements:
When using the logback appender I tried adding my own `LoggingEnhancer`, but it failed to load with a ClassNotFound exception (which was swallowed).

Investigation revealed that the problem was that the logback appender was loading it using the system classloader. Switching this to the thread context classloader (with `ch.qos.logback.core.util.Loader#loadClass`) worked fine. 

Label: logging-logback

tcoffee-google:
@jabubake would you mind taking a look at this?

jabubake:
@dclements : I'll take a look and patch, if this is the case. Thanks for the suggested fix.

jabubake:
@dclements, can you also provide  the version of JDK you used ?

dclements:
Java 1.8.0_131
Scala 2.12.2
akka-http 10.0.7
akka 2.5.2

Executed using `java -jar`.

Thanks for taking a look!

PR #2130, opened 2017-06-08
Logback logging appender : class loader fix for custom enhancers

jabubake:
#2123


garrettjonesgoogle:
Could you give an explanation of this fix?

jabubake:
The use of `ClassLoader.getSystemClassLoader()` restricts the visibility to the classpath.
Frameworks (like Play) use dynamic class loaders. Hence delegating to logback's `Loader`.
Reference : http://www.javaworld.com/article/2077344/core-java/find-a-way-out-of-the-classloader-maze.html

garrettjonesgoogle:
LGTM

PR #2132, opened 2017-06-09
update_version.sh: make maven run in batch mode

pongad:
Newer versions of Maven has colored outputs, which messes with the script.
The -B (batch) option disables colors.

Could folks using older Mavens let me know if your versions accept
the batch option?


neozwu:
I'm using maven 3.3.9, and `mvn -B dependency:tree` works.

PR #2133, opened 2017-06-09
verify.sh: make verify more debug-friendly

pongad:
This commit adds two options to `maven verify`.

The first, `-fae` (fail at end) makes the tests keep going
after failures. If a test flakes, at least we'll still
see the result of other tests.

The second, `-DtrimStackTrace=false` prints full stack
trace to make debugging easier.

Finally, we dedup the command call.


garrettjonesgoogle:
LGTM

PR #2135, opened 2017-06-09
Removing .spi from generated package names

garrettjonesgoogle:


google-cloud-vision/src/main/java/com/google/cloud/vision/v1/PagedResponseWrappers.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.cloud.language.spi.v1;
```
jabubake: expected : language -> vision ?

google-cloud-language/src/main/java/com/google/cloud/language/v1/PagedResponseWrappers.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.cloud.vision.spi.v1;
```
jabubake: Is this expected : vision -> language ?

jabubake: Ignore, I just noticed the change in path.

garrettjonesgoogle: This is git getting confused. Notice that the destination file is language here too. If you browse the final state of the code after my change, you'll see it's in order.

google-cloud-vision/src/main/java/com/google/cloud/vision/v1/PagedResponseWrappers.java
jabubake: Ignore, is there a reason for this file swap ?

PR #2145, opened 2017-06-13
fix wrong logging filter format, which causes 502 error

neozwu:


google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -468,12 +468,14 @@ public void testLoggingHandler() throws InterruptedException {
     logger.addHandler(handler);
     logger.setLevel(Level.INFO);
     logger.info("Message");
+    String filter =
+            "logName = projects/" + logging().getOptions().getProjectId() + "/logs/" + logName;
```
garrettjonesgoogle: Use https://github.com/garrettjonesgoogle/api-client-staging/blob/master/generated/java/proto-google-cloud-logging-v2/src/main/java/com/google/logging/v2/LogName.java#L26 instead of composing the String manually.

neozwu: Use LogName now. We had some unfortunate name clash. The `logName` in the code should be "logId", whereas the [doc](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) suggests that "logName"  is a fully qualified name prefixed with resourceName.  PTAL.

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -335,26 +336,27 @@ public void testListMetricsAsync() throws ExecutionException, InterruptedExcepti
...
-    String filter =
-        "logName = projects/" + logging().getOptions().getProjectId() + "/logs/" + logName;
+    String logId = formatForTest("test-write-log-entries-log");
+    LoggingOptions loggingOptions = logging().getOptions();
+    LogName logName = LogName.create(loggingOptions.getProjectId(), logId);
     StringPayload firstPayload = StringPayload.of("stringPayload");
     LogEntry firstEntry = LogEntry.newBuilder(firstPayload)
         .addLabel("key1", "value1")
-        .setLogName(logName)
+        .setLogName(logId)
```
garrettjonesgoogle: It seems like `logName` should be passed here - I think the tests have been passing the wrong thing. 

https://github.com/googleapis/googleapis/blob/master/google/logging/v2/log_entry.proto#L39

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -335,26 +336,27 @@ public void testListMetricsAsync() throws ExecutionException, InterruptedExcepti
...
     LogEntry secondEntry = LogEntry.newBuilder(secondPayload)
         .addLabel("key2", "value2")
-        .setLogName(logName)
+        .setLogName(logId)
         .setOperation(Operation.of("operationId", "operationProducer"))
         .setResource(MonitoredResource.newBuilder("cloudsql_database").build())
         .build();
     logging().write(ImmutableList.of(firstEntry));
     logging().write(ImmutableList.of(secondEntry));
+    String filter = "logName = " + logName.toString();
```
garrettjonesgoogle: I scanned https://cloud.google.com/logging/docs/view/advanced_filters very quickly, and I think it might be a good idea to quote the log name inside the string because it has special characters (the `/`). Since you format the filter multiple times, it's probably worth making a helper method for it.

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -435,51 +438,53 @@ public void testWriteAndListLogEntriesAsync() throws InterruptedException, Execu
...
-    assertFalse(logging().deleteLog(logName));
+    String logId = formatForTest("test-delete-non-existing-log");
+    assertFalse(logging().deleteLog(logId));
   }
 
   @Test
   public void testDeleteNonExistingLogAsync() throws ExecutionException, InterruptedException {
-    String logName = formatForTest("test-delete-non-existing-log-async");
-    assertFalse(logging().deleteLogAsync(logName).get());
+    String logId= formatForTest("test-delete-non-existing-log-async");
```
garrettjonesgoogle: add space before the `=`

neozwu: done

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
neozwu: create a helper method now

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
neozwu: I think we should pass logId, and logName will be constructed (with logId encoded and resourceName prepended):
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java#L500

garrettjonesgoogle: Ok, the naming on the Logging surface is wrong - but now we can't change it because it would be breaking :-(

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -76,6 +77,11 @@
    */
   protected abstract String formatForTest(String resourceName);
 
+  /**
+   * Creates a comparison expression for logging filter.
+   */
+  protected abstract <V> String createComparisonExpression(String name, String op, V value);
```
garrettjonesgoogle: This isn't something that needs to be overridden, so just make a static utility method.

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -335,26 +341,27 @@ public void testListMetricsAsync() throws ExecutionException, InterruptedExcepti
...
     LogEntry secondEntry = LogEntry.newBuilder(secondPayload)
         .addLabel("key2", "value2")
-        .setLogName(logName)
+        .setLogName(logId)
         .setOperation(Operation.of("operationId", "operationProducer"))
         .setResource(MonitoredResource.newBuilder("cloudsql_database").build())
         .build();
     logging().write(ImmutableList.of(firstEntry));
     logging().write(ImmutableList.of(secondEntry));
+    String filter = createComparisonExpression("logName", "=", logName);
```
garrettjonesgoogle: For simplicity, maybe take the "=" parameter out and rename the function to `createEqualityFilter`.

neozwu: done

google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
neozwu: done

PR #2153, opened 2017-06-15
Release process: skipping tests (cobertura, surefire) [ci skip]

garrettjonesgoogle:
The reporting steps for the release process make it take way too long (> 1 hour). Without any tests run, the release process takes about 20 minutes. I think we can live without the official coverage reports on releases, since we get it on all PRs.


garrettjonesgoogle:
Note: this essentially un-resolves https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1578 .

vam-google:
LGTM, though having broken cubertura page feels not good =(

garrettjonesgoogle:
I think it would be best to have some kind of offline process that rebuilds the cobertura page after a release, instead of making it part of the release process.

PR #2171, opened 2017-06-21
use grpc 1.4.0

pongad:
requires https://github.com/googleapis/googleapis/pull/384
and https://github.com/googleapis/gax-java/pull/332

Update dependencies
protobuf: 3.2.0 -> 3.3.0
grpc: 1.2.0 -> 1.4.0

These dependencies also require grpc/protobuf, so they also need to be updated so things converge.
gax: 1.4.0 -> 1.4.1
gax-grpc: 0.21.0 -> 0.21.1
generated proto: 19.0 -> 20.0

Generated code change:
We no longer retry on UNAVAILABLE, since bug is grpc is resolved.

### DIFF
--- a/google-cloud-errorreporting/pom.xml
+++ b/google-cloud-errorreporting/pom.xml
@@ -22,7 +22,7 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork26</version>
+      <version>2.0.3.Final</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/google-cloud-language/pom.xml
+++ b/google-cloud-language/pom.xml
@@ -22,7 +22,7 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork26</version>
+      <version>2.0.3.Final</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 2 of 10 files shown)


pongad:
Testing on my machine, the test for spanner seems to be deadlocking. I'll investigate tomorrow.

pongad:
Update: not actually a deadlock, but looked like one, probably due to how surefire printed stuff. actually a dependency conflict on guava, will investigate more

pom.xml
```diff
@@ -584,6 +589,12 @@
         <groupId>com.google.api</groupId>
         <artifactId>gax-grpc</artifactId>
         <version>${gax-grpc.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>com.google.api.grpc</groupId>
+            <artifactId>proto-google-common-protos</artifactId>
```
garrettjonesgoogle: Why does this need to be excluded here? We have control over the gax-grpc -> proto-google-common-protos dependency and we can fix it up if there's a problem...

pongad: Done, but please see https://github.com/googleapis/toolkit/pull/1382 for context.

garrettjonesgoogle:
It's a little frustrating to sort through regeneration PRs (and later, commits to master) that include multiple completely unrelated updates. My preference is for them to be separate PRs or commits within a PR, or if that is impractical given the passage of time (your case here), then at least list out the various types of changes in the commit description.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

pongad:
@garrettjonesgoogle Sorry, tcnative needs to change so we don't fail at runtime. PTAL

garrettjonesgoogle:
Add the tcnative version update in the description.

Have you run the Logging and Pub/Sub integration tests?

LGTM after the description is updated and you confirm integration tests were run.

pongad:
@garrettjonesgoogle Will add tcnative update to desc. Integ passes for all APIs.

PR #2179, opened 2017-06-23
Retry on resource exhausted.

mairbek:
R: @vkedia


tomayles:
I wonder if this should be optional. Resource errors could include situations like having too much storage per node as well as transient overuse of node compute capacity. Most users would want to retry, but there are possibly cases where they would prefer to fail fast.

How is RESOURCE_EXHAUSTED handled in other google-cloud-java APIs?

vkedia:
In DataFlow bulk load case I feel that people might run into "disk exceeded" kind of resource_exhausted pretty often in which case they might want to fail fast. Can we figure out from the error message what resource got exhausted?

@vam-google for the question about what do other APIs do.

garrettjonesgoogle:
@vam-google could you answer the question in this PR so it can either move forward or be closed?

vkedia:
This can be closed. @mairbek has a different PR #2645 to handle this.

PR #2180, opened 2017-06-23
Fixing spanner headers consistency issue.

vam-google:
Previously spanner client was sending (example valuess):
```
user-agent: google-cloud-java/0.20.0-beta grpc-java-netty/1.2.0
x-goog-api-client: google-cloud-java/0.20.0-beta
```

After the changes (example values):
```
user-agent: grpc-java-netty/1.2.0,
x-goog-api-client: gl-java/1.8.0_112-google-v7 gccl/0.20.1-beta-SNAPSHOT gax/1.4.0 grpc/1.2.0
```

Also added missing spanner **quickstart** sample and updated **README** respectively.

The quickstart sample was taken from https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/spanner/cloud-client/src/main/java/com/example/spanner/QuickstartSample.java

### DIFF
--- a/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java
+++ b/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java
@@ -74,6 +74,16 @@ public class GrpcTransportOptions implements TransportOptions {
         }
       };
 
+  /**
+   * Returns a string value for x-goog-api-client HTTP header. The header is used to report version
+   * of the client and its protocol-specific dependencies.
+   *
+   * For internal use.
+   *
+   * @param libraryVersion version of the google-cloud-java library
+   * @return value of x-goog-api-client HTTP header, which should be provided with each request
+   */
+  @InternalApi
   public String getXGoogApiClientHeader(String libraryVersion) {
     return String.format(Locale.US,
         "gl-java/%s %s/%s gax/%s grpc/%s",
--- a/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
+++ b/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
@@ -169,6 +169,15 @@ public class HttpTransportOptions implements TransportOptions {
     };
   }
 
+  /**
+   * Returns a string value for x-goog-api-client HTTP header. The header is used to report version
+   * of the client and its protocol-specific dependencies.
+   *
+   * For internal use.
+   *
+   * @param libraryVersion version of the google-cloud-java library
+   * @return value of x-goog-api-client HTTP header, which should be provided with each request
+   */
   String getXGoogApiClientHeader(String libraryVersion) {
     return String.format(Locale.US,
         "gl-java/%s %s/%s",


google-cloud-examples/src/main/java/com/google/cloud/examples/spanner/snippets/DatabaseSelect.java
```diff
@@ -0,0 +1,67 @@
...
+
+    // Name of your instance & database.
+    String instanceId = args[0];
+    String databaseId = args[1];
+    try {
+      // Creates a database client
+      DatabaseClient dbClient = spanner.getDatabaseClient(DatabaseId.of(
+          options.getProjectId(), instanceId, databaseId));
+      // Queries the database
+      ResultSet resultSet = dbClient.singleUse().executeQuery(Statement.of("SELECT 1"));
```
vkedia: can you put this in a try with resource block

vam-google:
Addressed review feedback. PTAL.

google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java
```diff
@@ -71,6 +74,16 @@ public void close(ScheduledExecutorService instance) {
         }
       };
 
+  public String getXGoogApiClientHeader(String libraryVersion) {
```
shinfan: Comments

google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
```diff
@@ -163,22 +163,18 @@ public void initialize(HttpRequest httpRequest) throws IOException {
...
     };
   }
 
-  String getXGoogApiClientHeader(ServiceOptions<?, ?> serviceOptions) {
-    return String.format(
+  String getXGoogApiClientHeader(String libraryVersion) {
+    return String.format(Locale.US,
         "gl-java/%s %s/%s",
-        getJavaVersion(),
+        firstNonNull(Runtime.class.getPackage().getImplementationVersion(), ""),
```
shinfan: I would still prefer to keep this as a helper (even it is a single line) for readability

google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java
vam-google: Added

google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
vam-google: I removed the method, because otherwise I would have to have the same method in two places (GrpcTransportOptions and HttpTransportOptions). Having two methods which return "java version" (somthing constant for same instance of an application) seems weird. We could also make that a public method somewhere and reuse it, but it seems unnecessary (trying to not add anything to public api unless it is necessary).

google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java
```diff
@@ -72,6 +75,26 @@ public void close(ScheduledExecutorService instance) {
...
+   *
+   * For internal use.
+   *
+   * @param libraryVersion version of the google-cloud-java library
+   * @return value of x-goog-api-client HTTP header, which should be provided with each request
+   */
+  @InternalApi
+  public String getXGoogApiClientHeader(String libraryVersion) {
+    return String.format(Locale.US,
+        "gl-java/%s %s/%s gax/%s grpc/%s",
```
bjwatson: Is GAX being used by manual layers that do _not_ depend on GAPIC?

vam-google: I would say yes. Gapic is basically codegen clients, which depend on grpc. 
Currently for java client the "gapic/" part of header does not make sense unless at least some part of the client was generated by toolkit. Spanner currently does not have generated parts in it (there is a plan to migrate to gapic), that is why the gapic part is not there in the header.

More details about gax.

In java we recently split our gax into just "gax" and "gax-grpc". All clients (manual, semi-manual and generated) use gax (because it has some fundamental classes, like RetrySettings for example). Not all clients use gax-grpc (for example datastore does not use it).
I discussed that with @omaray, and as result we decided to put gax header if at least gax (not gax-grpc) is used. Note, currently gax and gax-grpc are always released together, so gax version always have an equal corresponding gax-grpc (so for clients, which use both, knowing version of gax is enough).

bjwatson: SGTM. Thanks for the explanation.

PR #2183, opened 2017-06-23
add platform information

neozwu:
*add logic to distinguish GAE standard java 7 and 8 on both DevServer and Prod
*pick right transport option for a given platform

Verification: deploy GAE java 7 and 8 apps, to both DevServer and PROD

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
@@ -29,7 +29,7 @@ public final class PlatformInformation {
   public static final String JETTY_LOGGER_ON_GAE8_PROD =
       "com.google.apphosting.runtime.jetty9.JettyLogger";
   public static final String JETTY_LOGGER_ON_GAE8_DEVSERVER =
-      " com.google.appengine.development.jetty9.JettyLogger";
+      " com.google.appengine.development.jetty9.JettyLogger"; //note the leading space
 
   private PlatformInformation() {}
 


google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud;
+
+public final class PlatformInformation {
```
garrettjonesgoogle: Mark as @InternalApi for now. (We can open it up later if there is demand.)

google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final String SPECIFICATION_VERSION =
+      System.getProperty("java.specification.version");
+  public static final String GAE_RUNTIME_VERSION =
+      System.getProperty("com.google.appengine.runtime.version");
+  public static final String RUNTIME_JETTY_LOGGER =
+      System.getProperty("org.eclipse.jetty.util.log.class");
+  public static final String JETTY_LOGGER_ON_GAE8_PROD =
+      "com.google.apphosting.runtime.jetty9.JettyLogger";
+  public static final String JETTY_LOGGER_ON_GAE8_DEVSERVER =
+      " com.google.appengine.development.jetty9.JettyLogger";
```
garrettjonesgoogle: there is a space at the beginning of this string

neozwu: This is unfortunately true. Verified this space in google3.

google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
neozwu: Done

garrettjonesgoogle: I recommend this one

google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
garrettjonesgoogle: Add a comment then - this is very unexpected, and others might think it is a mistake too.

google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
neozwu: done

google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
neozwu: done

garrettjonesgoogle:
LGTM from the coding perspective, I'll let @lesv or @jabubake speak to the logic being used.

lesv:
Checked both J7 & J8 on Std for both local and deployed.  I didn't check GAE Flex-compat.

PR #2200, opened 2017-06-28
Redirect to HTTPS version of the docs.

tswast:
Even if I go to https://googlecloudplatform.github.io/google-cloud-java/, I was being redirected to the HTTP version.


garrettjonesgoogle:
This will just get overwritten on the next release. The source of this link is in https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/utilities/create_site.sh .

PR #2212, opened 2017-07-03
pubsub: enable streaming pull

pongad:
Moving this to a branch.
If nothing else, we should perf test this before
declaring it release-ready.

This commit re-enables streaming pull.
Unlike previous implementation,
it does not fall back to polling if streaming is unavailable,
since the streaming pull endpoint should be working
by the time this is released.

This commit fixes a bug deadline modification code.
Previously we record an Instant we receive a message,
then call Instant::getNano to get the time in nanoseconds.
This is incorrect since getNano returns the nanosecond
from the beginning of that second, not since an epoch.
The fix is to simply save the time since epoch instead of Instant.

This commit also slightly changes how errors are logged,
so that errors that occur after the service is being shut down
don't spam the console.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
@@ -38,9 +38,11 @@ import com.google.common.primitives.Ints;
 import com.google.pubsub.v1.GetSubscriptionRequest;
 import com.google.pubsub.v1.SubscriberGrpc;
 import com.google.pubsub.v1.SubscriberGrpc.SubscriberFutureStub;
+import com.google.pubsub.v1.SubscriberGrpc.SubscriberStub;
 import com.google.pubsub.v1.Subscription;
 import com.google.pubsub.v1.SubscriptionName;
 import io.grpc.CallCredentials;
+import io.grpc.Channel;
 import io.grpc.ManagedChannel;
 import io.grpc.auth.MoreCallCredentials;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -284,6 +289,94 @@ protected void doStop() {
...
+                }
+                // It could happen that we are shutting down while some channels fail.
+              }
+            }
+          });
+    }
+
+    ackDeadlineUpdater =
+        executor.scheduleAtFixedRate(
+            new Runnable() {
```
garrettjonesgoogle: This is a somewhat large anonymous class - could you extract it out of this method?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -257,22 +259,25 @@ public void close() {
...
+              }
+            })
+        .start();
   }
 
   @Override
   protected void doStop() {
-    // stopAllStreamingConnections();
+    stopAllStreamingConnections();
     stopAllPollingConnections();
```
pongad: It becomes a noop if polling never started. Commenting it out is a good idea though.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
pongad: Done

pongad:
Sorry I accidentally merged this. I've reset the branch and re-posted this at https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2223

PR #2217, opened 2017-07-07
Added snippets for DatabaseClient

vkedia:


google-cloud-examples/src/main/java/com/google/cloud/examples/spanner/snippets/DatabaseClientSnippets.java
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  /**
+   * Example of blind write.
+   */
+  // [TARGET write(Iterable)]
+  // [VARIABLE my_singer_id]
+  public void write(long singerId) {
+    // [START write]
+    List<Mutation> mutations = new ArrayList<>();
```
vam-google: This is a size 1 (`Collections.singletonList()`) list. What is the reason to use `ArrayList` here (as well as in other places, like line 79)?

vkedia: No particular reason. I changed it now to `Collections.singletonList()`

PR #2223, opened 2017-07-10
pubsub: enable streaming pull

pongad:
Moving this to a branch.
If nothing else, we should perf test this before
declaring it release-ready.

This commit re-enables streaming pull.
Unlike previous implementation,
it does not fall back to polling if streaming is unavailable,
since the streaming pull endpoint should be working
by the time this is released.

This commit fixes a bug deadline modification code.
Previously we record an Instant we receive a message,
then call Instant::getNano to get the time in nanoseconds.
This is incorrect since getNano returns the nanosecond
from the beginning of that second, not since an epoch.
The fix is to simply save the time since epoch instead of Instant.

This commit also slightly changes how errors are logged,
so that errors that occur after the service is being shut down
don't spam the console.

cc @davidtorres


garrettjonesgoogle:
LGTM

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -257,23 +259,26 @@ public void close() {
...
-          notifyFailed(t);
-        }
-      }
-    }).start();
+    new Thread(
+            new Runnable() {
+              @Override
+              public void run() {
+                try {
+                  // startPollingConnections();
```
davidtorres: nit: is there any value of leaving this commented out?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -257,23 +259,26 @@ public void close() {
...
+            })
+        .start();
   }
 
   @Override
   protected void doStop() {
-    // stopAllStreamingConnections();
-    stopAllPollingConnections();
+    stopAllStreamingConnections();
+    // stopAllPollingConnections();
```
davidtorres: ditto

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -284,6 +289,90 @@ protected void doStop() {
...
+
+  private void stopAllStreamingConnections() {
+    stopConnections(streamingSubscriberConnections);
+    if (ackDeadlineUpdater != null) {
+      ackDeadlineUpdater.cancel(true);
+    }
+  }
+
   // Starts polling connections. Blocks until all connections declare themselves running.
   private void startPollingConnections() throws IOException {
```
davidtorres: should remove this method as well as stopPollingConnections?

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
```diff
@@ -205,6 +205,9 @@ public void testAckSingleMessage() throws Exception {
 
   @Test
   public void testGetSubscriptionOnce() throws Exception {
+    if (isStreamingTest) {
```
davidtorres: I assume you'll clean up the tests later? since there is no point on keeping the logic to test non streaming polling

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
pongad: Done

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
pongad: Done

pongad:
@davidtorres Sorry for the delay. PTAL

garrettjonesgoogle:
Actually I don't want to nuke polling just yet... could we keep the code for the moment?

pongad:
@garrettjonesgoogle Do you mean just keeping the Polling class around? I was thinking about the same thing, but then if we need to restore polling, we'll probably revert anyway right? This is also in a branch so we can test to make sure before releasing? Please let me know if you feel strongly about this.

garrettjonesgoogle:
@pongad I want to see if there is demand for Pub/Sub on http/json first - my request isn't really about reverting.

pongad:
@garrettjonesgoogle I'm not sure exactly what you want. I added back Polling class, but it's not tested as we removed the fallback logic. Is this good enough for you?

garrettjonesgoogle:
The diff is now showing all of the diffs from master since the head of your branch doesn't have those commits, making it very hard to read.

Could we add a package-private thing on the builder to trigger polling, and run all the relevant tests on both polling and streaming?

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
re-LGTM

PR #2230, opened 2017-07-11
Fix BaseKey.Builder usage via reflection issue

vam-google:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2226

apparently caused by [JDK bug opened from 1999!](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4283544)


garrettjonesgoogle:
LGTM

PR #2238, opened 2017-07-14
cache library version

### LINKED ISSUE #2234: ServiceOptions#getLibraryVersion() performance bottleneck/native memory leak.
Dear Google team,

During our internal performance testing of our application that is using google cloud java library (for accessing Google Datastore) we noticed abnormal native memory consumption (which was leading to java process being killed by Linux's OOM killer). It is caused by another bug in com.google.api.gax.core.PropertiesProvider (https://github.com/googleapis/gax-java/issues/337). But it wouldn't happen if class com.google.cloud.ServiceOption had a little bit different implementation.

In method com.google.cloud.ServiceOptions#getLibraryVersion() properties are loaded each time from classpath. This method is used very intensively (as far as I understood during each request to REST service). Loading classpath resource is not efficient especially when application has large classpath with a lot of jars and resources.

Proposal is to add caching of method execution result. Even dumb implementation with whole method synchronization is more than 10 time faster.

Example:

```java
public abstract class ServiceOptions<ServiceT extends Service<OptionsT>, OptionsT extends ServiceOptions<ServiceT, OptionsT>> implements Serializable {
    /**
    * Cached library version
    **/
    private String libraryVersion = null;

    //...

    public synchronized String getLibraryVersion() {
        if(this.libraryVersion != null) {
            return this.libraryVersion;
        }

        try {
            String version = this.getVersionProperty(this.getPackagePath());
            if(version == null) {
                version = this.getVersionProperty("com/google/cloud");
            }

            // Cache library version
            this.libraryVersion = version;

            return version;
        } catch (Exception var2) {
            return null;
        }
    }

     //...

}
```

BR,
Dionis

pongad:
so we don't need to read environment properties every call.

Fixes #2234.
This is not the approach proposed in the issue.
As implemented in this PR, we will read the property once
for every client instantiation, not once ever.
Still, this should be fast enough,
and we won't need to maintain double-check-locking mechanism.

### DIFF
--- a/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
+++ b/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
@@ -149,6 +149,7 @@ public class HttpTransportOptions implements TransportOptions {
     final HttpRequestInitializer delegate =
         scopedCredentials != null && scopedCredentials != NoCredentials.getInstance()
             ? new HttpCredentialsAdapter(scopedCredentials) : null;
+    final String xGoogHeader = getXGoogApiClientHeader(serviceOptions.getLibraryVersion());
     return new HttpRequestInitializer() {
       @Override
       public void initialize(HttpRequest httpRequest) throws IOException {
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
LGTM

PR #2239, opened 2017-07-14
Retry in more situations

jean-philippe-martin:
Change the code to retry in more situations (when copying or deleting files, checking file permissions, or reading file size). Also give a comprehensive exception message when we run out of retries, to make it abundantly clear that retries were attempted.

Those aggressive retries only happen if the user has set the `MaxChannelReopens` option. If unset, we still get the nicer error message but they'll say something like "All 0 reopens failed."

Note that this doesn't affect the normal retries done at a lower level by the `cloud.storage.Storage` object.

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
+++ b/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
@@ -269,7 +269,7 @@ public final class CloudStorageFileSystemProvider extends FileSystemProvider {
   private SeekableByteChannel newReadChannel(Path path, Set<? extends OpenOption> options)
       throws IOException {
     initStorage();
-    int maxChannelReopens = ((CloudStorageFileSystem) path.getFileSystem()).config().maxChannelReopens();
+    int maxChannelReopens = CloudStorageUtil.getMaxChannelReopensFromPath(path);
     for (OpenOption option : options) {
       if (option instanceof StandardOpenOption) {
         switch ((StandardOpenOption) option) {
@@ -408,7 +408,18 @@ public final class CloudStorageFileSystemProvider extends FileSystemProvider {
       }
       throw new CloudStoragePseudoDirectoryException(cloudPath);
     }
-    return storage.delete(cloudPath.getBlobId());
+
+    final CloudStorageRetryHandler retryHandler = new CloudStorageRetryHandler(CloudStorageUtil.getMaxChannelReopensFromPath(path));
+    // Loop will terminate via an exception if all retries are exhausted
+    while (true) {
+      try {
+        return storage.delete(cloudPath.getBlobId());
+      } catch (StorageException exs) {
+        // Will rethrow a StorageException if all retries/reopens are exhausted
+        retryHandler.handleStorageException(exs);
+        // we're being aggressive by retrying even on scenarios where we'd normally reopen.
+      }
+    }
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * @param exs StorageException to test
+   * @return true if exs is an error that can be resolved via a channel reopen, otherwise false
+   */
+  static boolean isReopenable(final StorageException exs) {
+    Throwable throwable = exs;
+    // ensures finite iteration
+    int maxDepth = 10;
+    while (throwable != null && maxDepth-- > 0) {
+      if ((throwable.getMessage() != null
```
jean-philippe-martin: We'll have to agree to disagree on that one, codacy-bot.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
```diff
@@ -515,40 +527,50 @@ public void copy(Path source, Path target, CopyOption... options) throws IOExcep
...
+                Storage.CopyRequest.newBuilder().setSource(fromPath.getBlobId());
+        if (wantReplaceExisting) {
+          copyReqBuilder = copyReqBuilder.setTarget(tgtInfo);
+        } else {
+          copyReqBuilder = copyReqBuilder.setTarget(tgtInfo, Storage.BlobTargetOption.doesNotExist());
         }
-        if (!setContentDisposition) {
-          tgtInfoBuilder.setContentDisposition(blobInfo.getContentDisposition());
+        CopyWriter copyWriter = storage.copy(copyReqBuilder.build());
+        copyWriter.getResult();
```
droazen: BUGBUG: There needs to be a `break;` after this line to end the enclosing loop when the copy is successful.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -0,0 +1,176 @@
...
+   * Otherwise (or if out of attempts), throws a StorageException if we've exhausted all retries.
+   *
+   * The sleep time is dependent on the retry number.
+   *
+   * @param exs The StorageException error that prompted this retry attempt.
+   */
+  public void handleRetryForStorageException(final StorageException exs) throws StorageException {
+    if (!isRetryable(exs)) {
+      throw exs;
+    }
```
droazen: Currently, `handleRetryForStorageException()` throws if the exception it's passed is not retryable, but `handleReopenForStorageException()` does not throw if the exception it's passed is not re-openable. This is a bit confusing/inconsistent.

My recommendation here would be to make both `handleRetryForStorageException()` and `handleReopenForStorageException()` private (forcing all clients to go through `handleStorageException()` instead), remove the `isRetryable()` check from `handleRetryForStorageException()`, and rely on `handleStorageException()` to check whether each error is retryable/reopenable and delegate to the right handler method.

droazen:
Replaced by https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2281, which addresses the issues raised above (@jean-philippe-martin is away for a month, which is why we chose to open a new PR). This can be closed.

jean-philippe-martin:
closing, we should merge #2281 instead.

PR #2252, opened 2017-07-18
use objenesis 2.6

neozwu:
This change excludes old version of objenesis from easymock and mockito, and use objenesis 2.6 as a dependency.


google-cloud-bigquery/pom.xml
```diff
@@ -58,9 +58,12 @@
     <dependency>
       <groupId>org.easymock</groupId>
       <artifactId>easymock</artifactId>
-      <version>3.4</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.objenesis</groupId>
+      <artifactId>objenesis</artifactId>
+    </dependency>
```
garrettjonesgoogle: shouldn't the objenesis dependency be scoped to `test`?

pom.xml
```diff
@@ -625,6 +628,24 @@
...
+            <groupId>org.objenesis</groupId>
+            <artifactId>objenesis</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>org.objenesis</groupId>
+        <artifactId>objenesis</artifactId>
+        <version>${objenesis.version}</version>
+        <scope>test</scope>
```
garrettjonesgoogle: I think the scope should be at the dependency declaration site, not the dependency management site.

google-cloud-bigquery/pom.xml
neozwu: Its scope is defined to be test in dependency management section in the parent pom.

pom.xml
neozwu: Since the scope for objenesis is always test, I think we should centralize it in dependency management.  We could always override it if it needs to be a different scope for a particular artifact. I think as long as the declared dependency can match the one in dependency management, it will inherit the defined scope. See [here](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html) (The dependency management section)

neozwu: hmmm, you are right. I didn't know the dependencyManagement takes precedence over scope mediation rules for transitive dependency.  Made scope declaration explicit now for each dependency. PTAL

garrettjonesgoogle:
LGTM

PR #2253, opened 2017-07-19
Support 'format' option in Translation

huangp:
The REST api of google translate [1] supports 'format' option as query parameter. The library does not come with this option. This pull request will add support for that.

[1] https://cloud.google.com/translate/docs/reference/translate

### DIFF
--- a/google-cloud-translate/src/main/java/com/google/cloud/translate/spi/v2/HttpTranslateRpc.java
+++ b/google-cloud-translate/src/main/java/com/google/cloud/translate/spi/v2/HttpTranslateRpc.java
@@ -126,7 +126,7 @@ public class HttpTranslateRpc implements TranslateRpc {
               .setSource(sourceLanguage)
               .setKey(options.getApiKey())
               .set("model", Option.MODEL.getString(optionMap))
-              .set(Option.FORMAT.value(), Option.FORMAT.getString(optionMap))
+              .setFormat(Option.FORMAT.getString(optionMap))
               .execute()
               .getTranslations();
       return Lists.transform(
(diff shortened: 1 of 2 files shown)


google-cloud-translate/src/main/java/com/google/cloud/translate/spi/v2/HttpTranslateRpc.java
```diff
@@ -126,6 +126,7 @@ private GenericUrl buildTargetUrl(String path) {
               .setSource(sourceLanguage)
               .setKey(options.getApiKey())
               .set("model", Option.MODEL.getString(optionMap))
+              .set(Option.FORMAT.value(), Option.FORMAT.getString(optionMap))
```
tcoffee-google: Shouldn't this be `"format"` instead of `Option.FORMAT.value()`?

google-cloud-translate/src/main/java/com/google/cloud/translate/spi/v2/HttpTranslateRpc.java
```diff
@@ -126,6 +126,7 @@ private GenericUrl buildTargetUrl(String path) {
               .setSource(sourceLanguage)
               .setKey(options.getApiKey())
               .set("model", Option.MODEL.getString(optionMap))
```
tcoffee-google: @mziccard @garrettjonesgoogle was there a reason to use the `set` method here instead of `setModel` (in https://github.com/GoogleCloudPlatform/google-cloud-java/commit/60270261fa763131f6cd6b9aacfc7819e0af1c3d#diff-1bf2355a417917946c22e0783005adf1R130 )?

huangp: I also see that's a setFormat method to use. I tested it and it works.

garrettjonesgoogle: I don't know personally - my best guess is that the code was written against a version of the generated translate library before `setModel` was present, but when it was supported on the server side. It seems fine to use `setModel` now.

PR #2254, opened 2017-07-19
pubsub: correct emulator sample

pongad:
Previously we use TopicAdminSettings.defaultBuilder().
The default channel uses secure connection,
which is incompatible with the emulator.
This commit corrects this.

We cannot yet provide a way for users to set plain text
in InstantiatingChannelProvider, since usePlaintext
is still a beta API.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -147,17 +147,30 @@ $ gcloud beta emulators pubsub env-init
 
 3. Point your client to the emulator.
 ```java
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
LGTM.

It occurred to me that we are calling `usePlaintext` from `GrpcLoggingRpc`, which is a problem (but outside the scope of this PR).

PR #2267, opened 2017-07-24
utilities: also create latest directory

pongad:
This commit makes create_site.sh create directory "latest",
containing the docs for the latest release.

This commit also fixes a bug.
Previously, create_site chooses to copy only files in directory
whose names begin with the current version.
Since the current version of google-cloud-java is 0.y.z,
libraries in alpha and beta are copied properly.
However, libraries in GA are not copied, because their versions
are x.y.z where x is not 0.

### DIFF
--- a/utilities/create_site.sh
+++ b/utilities/create_site.sh
@@ -15,32 +15,24 @@ fi
 git clone --branch gh-pages --single-branch https://github.com/GoogleCloudPlatform/google-cloud-java/ tmp_gh-pages
 mkdir -p tmp_gh-pages/$SITE_VERSION_BASE
 mvn site -DskipTests -Djava.util.logging.config.file=logging.properties
-mvn site:stage --quiet -Djava.util.logging.config.file=logging.properties -DtopSiteURL=http://googlecloudplatform.github.io/google-cloud-java/site/${SITE_VERSION_BASE}/
-cd tmp_gh-pages
 
-for dir in ../target/staging/*
+version_bases=(${SITE_VERSION_BASE} latest)
+rm -rf tmp_gh-pages/latest
+mkdir tmp_gh-pages/latest
+for version_base in ${version_bases[@]}
 do
-    cp -r ${dir}/* $SITE_VERSION_BASE/
+  mvn site:stage --quiet -Djava.util.logging.config.file=logging.properties -DtopSiteURL=http://googlecloudplatform.github.io/google-cloud-java/site/${version_base}/
+  cd tmp_gh-pages
+  cp -r ../target/staging/*/* $version_base
+  sed -i "s/{{SITE_VERSION}}/$SITE_VERSION/g" ${version_base}/index.html # Update "Quickstart with Maven" to reflect version change
+  git add $version_base
+  cd ..
 done
 
-sed -i "s/{{SITE_VERSION}}/$SITE_VERSION/g" ${SITE_VERSION_BASE}/index.html # Update "Quickstart with Maven" to reflect version change
-git add $SITE_VERSION_BASE
+cd tmp_gh-pages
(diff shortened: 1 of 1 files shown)


utilities/create_site.sh
```diff
@@ -30,4 +30,17 @@ git add index.html
 echo "<html><head><script>window.location.replace('/google-cloud-java/${SITE_VERSION_BASE}/apidocs' + location.search)</script></head><body></body></html>" > apidocs/index.html
 git add apidocs/index.html
 
+cd ..
+mvn site:stage --quiet -Djava.util.logging.config.file=logging.properties -DtopSiteURL=http://googlecloudplatform.github.io/google-cloud-java/site/latest/
```
pongad: I honestly can't tell if we need this or not. The documentation on `-DtopSiteURL` is a little fuzzy, so I re-stage here just in case.

utilities/create_site.sh
```diff
@@ -30,4 +30,17 @@ git add index.html
...
+cd tmp_gh-pages
+rm -rf latest
+mkdir latest
```
vam-google: Can you please clarify why we need this (latest dir) at all?

pongad: @omaray can probably answer this better than I can. I believe the main purpose is that people can write documentation that link to the latest version.

garrettjonesgoogle: I can answer I think - it's confusing to provide a link to GA library documentation (e.g. for pubsub, which is 1.x) but where the version in the link is the alpha version (0.x). Also, every single release we do, cloud docs have to be updated too, which is a big maintenance burden; having the /latest/ link is much easier to link to.

utilities/create_site.sh
```diff
@@ -30,4 +30,17 @@ git add index.html
...
+for dir in ../target/staging/*
```
vam-google: I'm confused by this.

It seems like this code is attempting to repeat the regular site creation process one more time, with only difference: it does not update global tmp_gh-pages/index.html to point to latest (it keeps pointing to the actual newest version, which seems reasonable). If that is the case please strongly consider modifying this code to something like (not sure it it is completely valid bash syntax, but...):

```bash
# [...] (lines 1-17)
version_bases=(${SITE_VERSION_BASE} latest)
for version_base in ${version_bases} 
do
  # All the stuff starting from "mvn site:stage" (line 18) to "git add $SITE_VERSION_BASE" (line 27)
  # but replacing ${SITE_VERSION_BASE} with ${version_base}
done
# [...] (lines 28-31)
```
I believe the following code structure reflects the best what the intent of the change is (and is also shorter, potentially easier to maintain and keep latest in real sync with version_base copy)

pongad:
Thank you @vam-google PTAL

utilities/create_site.sh
```diff
@@ -15,16 +15,21 @@ fi
 git clone --branch gh-pages --single-branch https://github.com/GoogleCloudPlatform/google-cloud-java/ tmp_gh-pages
 mkdir -p tmp_gh-pages/$SITE_VERSION_BASE
 mvn site -DskipTests -Djava.util.logging.config.file=logging.properties
-mvn site:stage --quiet -Djava.util.logging.config.file=logging.properties -DtopSiteURL=http://googlecloudplatform.github.io/google-cloud-java/site/${SITE_VERSION_BASE}/
-cd tmp_gh-pages
 
-for dir in ../target/staging/${SITE_VERSION_BASE}*
+version_bases=(${SITE_VERSION_BASE} latest)
+rm -rf tmp_gh-pages/latest
```
vam-google: Sorry, I missed this initially. How will this behave together with the corresponding "git add latest" for cases when there is already something in tmp_gh-pages/latest in github? I hope git is smart enough to remove what is not needed and update what was updated (no other commands needed, like explicit "git rm")?

pongad: git is smart enough to figure out what's required in this case.

PR #2281, opened 2017-07-27
google-cloud-nio: retry in more situations upon retryable errors

droazen:
There were a number of operations in google-cloud-nio that were prone
to intermittent failures (such as 503 errors) when run at scale, but
which were not wrapped in a retry loop or hooked up to the client settings
for number of retries/reopens. This rendered the google-cloud-nio library
effectively unusable at scale.

This commit wraps those additional operations in CloudStorageReadChannel
and CloudStorageFileSystemProvider within retry/reopen loops, with the number
of retries/reopens determined by the "maxChannelReopens" client settings.

If the client has not set "maxChannelReopens", they'll still get a nice
error message with the underlying error (such as a 503), plus an indication
that 0 retries/reopens were attempted.

To facilitate this change, we've added a CloudStorageRetryHandler that
keeps track of the total number of retry/reopen attempts, and throws
an appropriate exception when out of attempts.

Note that this change doesn't affect the normal retries done at a
lower level by the cloud.storage.Storage object.

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
+++ b/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage.contrib.nio;
+
+import com.google.cloud.storage.StorageException;
+
+import java.io.EOFException;
+import java.net.SocketException;
+import java.net.SocketTimeoutException;
+import javax.net.ssl.SSLException;
+
+/**
+ * Simple counter class to keep track of retry and reopen attempts when StorageExceptions are
+ * encountered. Handles sleeping between retry/reopen attempts, as well as throwing an exception
+ * when all retries/reopens are exhausted.
+ */
+public class CloudStorageRetryHandler {
+  private int retries;
+  private int reopens;
+  private long totalWaitTime; // in milliseconds
+  private final int maxRetries;
+  private final int maxReopens;
+
+  /**
+   * Create a CloudStorageRetryHandler with the maximum retries and reopens set to the same value.
+   *
+   * @param maxRetriesAndReopens value for both maxRetries and maxReopens
+   */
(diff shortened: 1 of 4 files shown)


droazen:
This PR solves a major problem for us that has cost our institute tens of thousands of dollars in wasted compute using the `google-cloud-nio` library, so we are extremely eager to see it merged! The problem was that there are many GCS operations in `google-cloud-nio` that are prone to intermittent failures, such as 503s, when running at scale. Only some of these operations were hooked up to the retry settings for the library and wrapped within retry loops. This PR simply wraps the remaining failure-prone operations within retry loops.

We've tested this patch at large scale (thousands of workers each accessing tens of thousands of files in GCS). Before this patch, we saw 10-20% of our jobs fail with 503s, SSL exceptions, and other transient errors. With this patch, we haven't seen any such failures, provided that the client code sets the number of retries/reopens to a large enough value.

This PR replaces both https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2239 and https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2219. Those two PRs were authored by us but submitted on our behalf by our Google collaborator @jean-philippe-martin due to CLA issues, and can now be closed. Since @jean-philippe-martin is away for a month, we've sorted out our CLA issues and re-submitted this change with some additional fixes.

Please let me know if there's anything we can do to expedite the merge of this patch! We'd love to see it included in the next `google-cloud-java` release.

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -0,0 +1,171 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
michaelbausor: s/2016/2017

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -0,0 +1,171 @@
...
+      // reset interrupt flag
+      Thread.currentThread().interrupt();
+    }
+  }
+
+  /**
+   * @param exs StorageException to test
+   * @return true if exs is a retryable error, otherwise false
+   */
+  static boolean isRetryable(final StorageException exs) {
```
michaelbausor: Can this be private rather than package-private? (same for isReopenable)

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
droazen: Fixed

google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
droazen: Yes, they can both be `private` -- fixed.

droazen:
@michaelbausor Thanks for the review! I've addressed your comments and rebased onto the latest master. I've also opened https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2375 for the task of creating those unit tests you requested for `CloudStorageRetryHandler`. Please let me know if this is good to merge now, or if more changes are needed.

PR #2283, opened 2017-07-27
google-cloud-nio: CloudStorageReadChannel: fix critical bug that could cause the channel position to overflow

droazen:
This fixes a critical bug in CloudStorageReadChannel in which the channel position
(a long) was getting down-cast to int in CloudStorageReadChannel.innerOpen(). This
could cause the channel position to become negative or a wildly incorrect positive
value if the channel position was greater than MAX_INT.

In practice, this bug would only manifest itself if the channel was constructed with
a non-zero initial position, or if the channel was reopened in response to an error.

Included a test case that fails without the fix, and passes with the fix.

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannelTest.java
+++ b/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannelTest.java
@@ -37,6 +37,7 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+import org.mockito.ArgumentCaptor;
 
 import javax.net.ssl.SSLHandshakeException;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


lbergelson:
The long nightmare is almost over!

pshapiro4broad:
👍

google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannelTest.java
```diff
@@ -191,4 +192,20 @@ public void testSetPosition() throws IOException {
...
+   * channel position (a long) was getting truncated to an int when seeking on the encapsulated
+   * ReadChannel in innerOpen(). This test case fails when the bad long -> int cast is present,
+   * and passes when it's removed.
+   */
+  @Test
+  public void testChannelPositionDoesNotGetTruncatedToInt() throws IOException {
+    // This position value will overflow to a negative value if a long -> int cast is attempted
+    long startPosition = 11918483280L;
+    ArgumentCaptor<Long> captor = ArgumentCaptor.forClass(Long.class);
+    final CloudStorageReadChannel channel = CloudStorageReadChannel.create(gcsStorage, file, startPosition, 1);
```
lbergelson: hey, codacy actually caught something

lbergelson:
This will fix https://github.com/broadinstitute/gatk/issues/2516 and https://github.com/broadinstitute/gatk/issues/3316

droazen:
@garrettjonesgoogle Could we please get an official review on this PR and https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2281? Without these patches `google-cloud-nio` is unusable at large scale, so as a large-scale user of this library we are naturally eager to see these make it into a future release! :)

This PR (#2283) fixes a nasty bug that could cause the channel position to overflow and become either negative or a wildly incorrect positive value, leading to corruption of the input stream from the client perspective. This is not a hypothetical problem -- we experienced it quite frequently!

The other PR (https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2281) wraps additional GCS operations within retry loops. Without that patch, 10-20% of our jobs were failing with transient 503 errors.

Thanks for your help!

google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/CloudStorageReadChannelTest.java
michaelbausor: Can we fix this codacy error?

droazen: Fixed!

droazen:
Thanks for merging this and https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2281 @michaelbausor!  Is there a timeline for the next `google-cloud-java` release?

michaelbausor:
Re the timing of the next release, @garrettjonesgoogle will know best

PR #2286, opened 2017-07-27
Adding @BetaApi to particular exception constructors

garrettjonesgoogle:
In https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2156 , the constructors in `BaseGrpcServiceException` and `LoggingException` that accepted `ApiException` were changed to accept `GrpcApiException`, which is technically a breaking change, but there is a magical loophole because the methods couldn't be called without referencing `ApiException` which was `@BetaApi` itself. I'm now adding `@BetaApi` to these methods so that the situation is clearer to users, and so that we don't depend on the loophole.


michaelbausor:
LGTM once tests pass

garrettjonesgoogle:
Merging because this is only an annotation change, and only the mac test hasn't run yet.

PR #2292, opened 2017-07-31
datastore: cache application name

### LINKED ISSUE #2290: ServiceOptions#getApplicationName() - performance issue
Dear Google Team,

During our company internal performance testing we found that there are some places for improvement (although it's not critical but very nice to have). Method com.google.cloud.ServiceOptions#getApplicationName() is used with each request (for example from code of method com.google.cloud.datastore.spi.v1.HttpDatastoreRpc#getHttpRequestInitializer() )
. This method is used each time during EACH request. Basically what it does under the hood: it loads some resource from classpath with EACH request. It is leading to not necessary memory and CPU consumption. It is especially notable on a high-loaded project with millions requests per day. 

Proposal is to cache result of method ServiceOptions#getApplicationName(). It's constant value anyway(by it's nature) - so there's no need to load it each time with each request (it won't change). Or load it once on startup.

Bug found in a library: com.google.cloud:google-cloud-datastore:1.2.3

Best Regards,
Dionis Argiri

pongad:
Compute the application name once per client and reuse.

Datastore seems to not use the client Builder pattern like
other HTTP clients which explains why I missed them earlier.

Fixes #2290.


garrettjonesgoogle:
LGTM

PR #2293, opened 2017-07-31
Release 0.21.1

shinfan:


google-cloud-contrib/google-cloud-nio-examples/README.md
```diff
@@ -22,12 +22,12 @@ To run this example:
 4.    Run the sample with:
 
     ```
-    java -cp google-cloud-contrib/google-cloud-nio/target/google-cloud-nio-0.21.1-alpha-SNAPSHOT-shaded.jar:google-cloud-contrib/google-cloud-nio-examples/target/google-cloud-nio-examples-0.21.1-alpha-SNAPSHOT.jar com.google.cloud.nio.examples.ListFilesystems
+    java -cp google-cloud-contrib/google-cloud-nio/target/google-cloud-nio-0.21.2-alpha-SNAPSHOT-shaded.jar:google-cloud-contrib/google-cloud-nio-examples/target/google-cloud-nio-examples-0.21.2-alpha-SNAPSHOT.jar com.google.cloud.nio.examples.ListFilesystems
```
michaelbausor: Is this one step aheat at 0.21.2 for a reason? Not necessarily now, but should we reconcile this with the rest of the version numbers?

shinfan:
@lesv @garrettjonesgoogle I thought the patch release would be okay since there is no breaking change but now I want to double check with you guys.

google-cloud-contrib/google-cloud-nio-examples/README.md
shinfan: ummm this is auto-changed by the script...I think this should be okay since we are bumping the version to 0.21.2-alpha-SNAPSHOT shortly after.

garrettjonesgoogle: This is intentional because people are expected to run the sample from source, so this should match the next snapshot version we will use.

garrettjonesgoogle:
I think that in the case of no new features in google-cloud-java, but updating a dependency with new features, should be a patch version bump. It's too prohibitive to do the analysis of all of the changes in dependencies and mirror the impact in google-cloud-java.

pom.xml
```diff
@@ -143,31 +143,31 @@
     <easymock.version>3.4</easymock.version>
     <objenesis.version>2.6</objenesis.version>
 
-    <cloud.version>0.21.1-alpha-SNAPSHOT</cloud.version>
+    <cloud.version>0.21.1-alpha</cloud.version>
     <bigquery.version>0.21.1-beta-SNAPSHOT</bigquery.version>
-    <compute.version>0.21.1-alpha-SNAPSHOT</compute.version>
-    <contrib.version>0.21.1-alpha-SNAPSHOT</contrib.version>
+    <compute.version>0.21.1-alpha</compute.version>
```
garrettjonesgoogle: I don't think that alpha should move a version without beta also moving a version. Also, logging & pubsub wouldn't pick up the corresponding dependency updates without updating them too (logging implies GA version update). I'd like to stick to patch bumps across everything instead of just across some version tracks.

shinfan: Done

shinfan:
@garrettjonesgoogle  PTAL

garrettjonesgoogle:
LGTM

PR #2298, opened 2017-08-02
P0 auth sample

lesv:


lesv:
Looks like 3 of 4 travis builds passed, the fourth appears to be bad because of my being on a fork, rather than a branch.  So, I'm going to merge.

PR #2304, opened 2017-08-03
fetch project id on app engine platforms

neozwu:
With GAE flex and standard J8 environment, the preferred way to get project id is through env var (if set ) or metadata server.
Verified on GAE J8 s+f.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -46,7 +46,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.Serializable;
-import java.lang.reflect.Method;
 import java.nio.charset.Charset;
 import java.util.Locale;
 import java.util.Objects;
@@ -391,31 +390,6 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>,
     String projectId = null;
     if (PlatformInformation.isOnGAEStandard7()) {
       projectId = getAppEngineProjectIdFromAppId();
-      if (projectId == null) {
-        try {
-          Class<?> factoryClass =
-                  Class.forName("com.google.appengine.api.appidentity.AppIdentityServiceFactory");
-          Class<?> serviceClass =
-                  Class.forName("com.google.appengine.api.appidentity.AppIdentityService");
-          Method method = factoryClass.getMethod("getAppIdentityService");
-          Object appIdentityService = method.invoke(null);
-          method = serviceClass.getMethod("getServiceAccountName");
-          String serviceAccountName = (String) method.invoke(appIdentityService);
-          int indexOfAtSign = serviceAccountName.indexOf('@');
-          projectId = serviceAccountName.substring(0, indexOfAtSign);
-        } catch (ClassNotFoundException exception) {
-          if (System.getProperty("com.google.appengine.runtime.version") != null) {
-            // Could not resolve appengine classes under GAE environment.
-            throw new RuntimeException(
-                    "Google App Engine runtime detected "
-                            + "(the environment variable \"com.google.appengine.runtime.version\" is set), "
-                            + "but unable to resolve appengine-sdk classes. "
-                            + "For more details see "
-                            + "https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/APPENGINE.md");
-          }
-        } catch (Exception ignore) {
-        }
-      }
     } else {
       //for GAE flex and standard Java 8 environment
       projectId = System.getenv("GOOGLE_CLOUD_PROJECT");


google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -381,32 +388,70 @@ private static boolean isWindows() {
...
+        }
+        return null;
+      } catch (Exception ignore) {
+        return null;
       }
-      return null;
-    } catch (Exception ignore) {
-      return null;
+    } else {
+      //for GAE flex and standard Java 8 environment
```
lesv: As I read this it's:
1. GCLOUD_PROJECT
2. GOOGLE_CLOUD_PROJECT
3. Metadata server
4. getAppEngineAppId()

If you are running locally, in an emulator, a Metadata server won't be available, so the Metadata request will timeout.  Ideally, you'd be able to get it from the `gcloud config get-value project`

neozwu: do you know is there a sys property or env var to tell whether it's on devserver or not ?

lesv: I would recommend preferring your getAppEngineAppId() to the Metadata server as it shouldn't care which thread you are on.

lesv: Actually, I think I'm ok as is.

lesv: You can tell if you are running in production by looking at these environment variables:
```
GAE_VERSION | gaeinfo:20170722t140621.402856547336086879
GAE_APPLICATION | s~lesv-003
GAE_RUNTIME | java8
GAE_ENV | standard
```
If GAE_VERSION is set, then you are in production, if unset you are not.

lesv: Argh - that's true for GAE Std, I need to check Flex.

lesv:
On Fetch, the following are available

Java System Properties
=
```
com.google.appengine.application.id | lesv-005
com.google.appengine.runtime.environment | Production
```
Environment Variables
=
```
APPENGINE_LOADBALANCER=
APPENGINE_LOADBALANCER_IP=
GAE_LONG_APP_ID=lesv-005
GAE_MINOR_VERSION=403158202882676881
GAE_APPENGINE_HOSTNAME | lesv-005.appspot.com
GAE_VM=true
GAE_PARTITION=s
GAE_DEPLOYMENT_ID= 403158202882676881
GOOGLE_CLOUD_PROJECT=lesv-005
GAE_SERVICE=default
GAE_INSTANCE= aef-default-20170804t141509-wj2q
GAE_VERSION= 20170804t141509
GCLOUD_PROJECT=lesv-005
GAE_MODULE_INSTANCE=0
```
Metadata
=
```
/computeMetadata/v1/project/numeric-project-id | 47006412909
/computeMetadata/v1/project/project-id | lesv-005
/computeMetadata/v1/instance/zone | projects/47006412909/zones/us-central1-b
```

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -381,32 +388,70 @@ private static boolean isWindows() {
...
+        if (projectId != null && projectId.contains(":")) {
+            int colonIndex = projectId.indexOf(":");
+            projectId = projectId.substring(colonIndex + 1);
+        }
+      }
+      return projectId;
     }
   }
 
+  private static String getAppEngineProjectIdFromMetadataServer() throws IOException {
```
lesv: Strike this

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -381,32 +388,70 @@ private static boolean isWindows() {
...
-      return null;
-    } catch (Exception ignore) {
-      return null;
+    } else {
+      //for GAE flex and standard Java 8 environment
+      String projectId = System.getenv("GCLOUD_PROJECT");
+      if (projectId == null) {
+        projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
+      }
+      if (projectId == null) {
```
lesv: I suggested the Metadata server, but really you can get it from GAE_APPLICATION on J8 standard and remove the prefix anything before & including the **~**

So the order becomes:
GCLOUD_PROJECT
GOOGLE_CLOUD_PROJECT
getAppEngineAppId()  i.e. System.getProperty("com.google.appengine.application.id"); 
GAE_APPLICATION (removing the prefix)

We need to document that users should set GOOGLE_CLOUD_PROJECT when running locally, until our tooling sets "com.google.appengine.application.id"

lesv:
I've checked GAE Standard J8 in a dasher account and it's behaving the same way as non-dasher.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
neozwu: I reordered the sequence to (1) GCLOUD_PROJECT , (2) GOOGLE_CLOUD_PROJECT, (3) getAppEngineAppId, (4) metadata server.  PTAL.

rrch: GOOGLE_CLOUD_PROJECT should come before GCLOUD_PROJECT. AFAICT the latter is deprecated.

lesv:
You might wish to make Travis happy, however.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -381,32 +388,71 @@ private static boolean isWindows() {
...
+    if (PlatformInformation.isOnGAEStandard7()) {
+      try {
+        Class<?> factoryClass =
+            Class.forName("com.google.appengine.api.appidentity.AppIdentityServiceFactory");
+        Class<?> serviceClass =
+            Class.forName("com.google.appengine.api.appidentity.AppIdentityService");
+        Method method = factoryClass.getMethod("getAppIdentityService");
+        Object appIdentityService = method.invoke(null);
+        method = serviceClass.getMethod("getServiceAccountName");
+        String serviceAccountName = (String) method.invoke(appIdentityService);
```
rrch: Checking for the availablity of the AppIdentityService class is not enough in GAE standard. If you call AppIdentityService.getServiceAccountName() from a thread that is not managed by the App Engine runtime, you'll get a ApiProxy.CallNotFoundException. Would it be possible to log a message in this case? Another way to detect this situation is to call ApiProxy.getCurrentEnvironment() and ensure the result is not null.

google-cloud-core/pom.xml
```diff
@@ -40,6 +40,22 @@
...
+        <exclusion>
+          <groupId>com.google.code.findbugs</groupId>
+          <artifactId>jsr305</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
```
ifigotin: Is findbugs needed as a compile time dependency?

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
neozwu: @rrch Done - GOOGLE_CLOUD_PROJECT coms before GCLOUD_PROJECT now.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
neozwu: I decided to remove reflection on appengine APIs, because the system prop approach doesn't have "thread access" issue and should work on both dev and prod servers for GAE J7.

google-cloud-core/pom.xml
neozwu: I didn't look into the details.  But google-http-client defines findbugs as a compile dependency.  A compile dependency of a compile dependency is a transitive compile dependency. So I think it's safe to make it compile time dependency.

neozwu:
@lesv @rrch PTAL.

Issue #2187, opened 2017-06-26
JavaDocs should explain the products rather than just say they are clients for...


lesv:
http://googlecloudplatform.github.io/google-cloud-java/0.20.0/apidocs/

Take the one line Marketing summary and put it as a description.

Currently:

| Class  | Description |
| ------ | ------------ |
| com.google.cloud	| Core classes for the google-cloud library. |
| com.google.cloud.bigquery	| A client to Google Cloud BigQuery. |
| com.google.cloud.bigquery.spi.v2	 | |
| com.google.cloud.compute	| A client to Google Cloud Compute. |
| com.google.cloud.compute.spi.v1	 | |
| com.google.cloud.datastore	| A client to the Google Cloud Datastore. |
| com.google.cloud.datastore.spi.v1	| 
| com.google.cloud.dns	| A client to the Google Cloud DNS. |
| com.google.cloud.dns.spi.v1	 | |

Suggested:

| Class  | Description |
| ------ | ------------ |
| com.google.cloud	| Core classes for the google-cloud library. |
| com.google.cloud.bigquery	| Bigquery – A fast, economical and fully-managed enterprise data warehouse for large-scale data analytics |
| com.google.cloud.bigquery.spi.v2	 | |
| com.google.cloud.compute	| Compute – Scalable, High-Performance Virtual Machines |
| com.google.cloud.compute.spi.v1	 | |
| com.google.cloud.datastore	| Datastore – Cloud Datastore is a highly-scalable NoSQL database for your web and mobile applications |
| com.google.cloud.datastore.spi.v1	| 
| com.google.cloud.dns	| Cloud DNS - Reliable, resilient, low-latency DNS serving from Google’s worldwide network |
| com.google.cloud.dns.spi.v1	 | |

garrettjonesgoogle:
Why? These are clients for those services...

lesv:
It helps people discover that we have those services.  (They might not have found it otherwise)

garrettjonesgoogle:
I don't think it's the responsibility of the client libraries to tell people what GCP services are available...

jabubake:
Agree with Les on making these more informative.  To the user who arrives at the page, it is our opportunity to inform users of the existing services without bouncing off to search for what that is.

lesv:
Let's talk about this - this afternoon.

garrettjonesgoogle:
[...]

PR #2312, opened 2017-08-07
Describing services in package-info files

garrettjonesgoogle:
I based the descriptions off of the official Cloud docs for each service, and also updated their names correspondingly. 

Updates #2187 .

Auto-gen services will be taken care of separately.

### DIFF
--- a/google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/package-info.java
+++ b/google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/package-info.java
@@ -15,8 +15,8 @@
  */
 
 /**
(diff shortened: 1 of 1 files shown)


google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/package-info.java
```diff
@@ -15,7 +15,8 @@
  */
 
 /**
- * A client to Google Cloud Resource Manager.
+ * A client for Cloud Resource Manager - Centrally manage resource containers for your
+ * projects.
```
lesv: Centrally manage all your projects, IAM, and resource containers.

PR #2313, opened 2017-08-07
Add missing fields to LogEntry

saicheems:
Resolves https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2201

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
@@ -363,7 +363,7 @@ public class LogEntry implements Serializable {
 
 
   /**
-   * Returns the source code location information associated with the log entry if any.
+   * Returns the source code location information associated with the log entry, if any.
    */
   public SourceLocation getSourceLocation() {
     return sourceLocation;
@@ -436,6 +436,18 @@ public class LogEntry implements Serializable {
     return new Builder(this);
   }
 
+  private static Timestamp timestampFromMillis(Long millis) {
+    Timestamp.Builder tsBuilder = Timestamp.newBuilder();
+    tsBuilder.setSeconds(millis / MILLIS_PER_SECOND);
+    tsBuilder.setNanos((int) (millis % MILLIS_PER_SECOND * NANOS_PER_MILLISECOND));
+    return tsBuilder.build();
+  }
+
+  private static Long millisFromTimestamp(Timestamp timestamp) {
+    return timestamp.getSeconds() * MILLIS_PER_SECOND
+        + timestamp.getNanos() / NANOS_PER_MILLISECOND;
+  }
+
   com.google.logging.v2.LogEntry toPb(String projectId) {
     com.google.logging.v2.LogEntry.Builder builder = payload.toPb();
     builder.putAllLabels(labels);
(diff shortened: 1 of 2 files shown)


saicheems:
Note, I'm missing doc strings -- I'll add those in, but I'm looking for some direction on how to write them.

saicheems:
Especially for fields which are "Output Only", as `receiveTimestamp` is.

saicheems:
Added the docstrings in, I just ripped them from the proto essentially. Lmk what else I need.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -303,6 +354,23 @@ public Operation getOperation() {
...
+   * Returns the resource name of the trace associated with the log entry, if any. If it contains a
+   * relative resource name, the name is assumed to be relative to `//tracing.googleapis.com`.
+   */
+  public String getTrace() {
+    return trace;
+  }
+
+
+  /**
+   * Returns the source code location information associated with the log entry if any.
```
vam-google: I'm not the English language expert, but the `getTrace()` method doc has comma before "if any", and this one does not.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -432,6 +518,13 @@ static LogEntry fromPb(com.google.logging.v2.LogEntry entryPb) {
         builder.setTimestamp(millis);
       }
     }
+    if (entryPb.hasReceiveTimestamp()) {
```
vam-google: Optional. Taking into account that this weird time computation logic is copied in two places now, please consider having a private method for that to avoid duplication.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -377,6 +451,12 @@ public Builder toBuilder() {
       tsBuilder.setNanos((int) (timestamp % MILLIS_PER_SECOND * NANOS_PER_MILLISECOND));
       builder.setTimestamp(tsBuilder.build());
     }
+    if (receiveTimestamp != null) {
```
vam-google: Optional. Taking into account that this weird time computation logic is copied in two places now, please consider having a private method for that to avoid duplication.

google-cloud-logging/src/main/java/com/google/cloud/logging/SourceLocation.java
```diff
@@ -0,0 +1,163 @@
...
+        .add("line", line)
+        .add("function", function)
+        .toString();
+  }
+
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  LogEntrySourceLocation toPb() {
```
vam-google: The access modifiers of `toPb() / fromPb()` methods seem inconsistent with the other similar entities (like `Operation`, where they are both package-private or `MonitoredResource`, where they are both public). Please make them at least consistent with each other (either both public or both package-private)

google-cloud-logging/src/test/java/com/google/cloud/logging/SourceLocationTest.java
```diff
@@ -0,0 +1,73 @@
...
+package com.google.cloud.logging;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class SourceLocationTest {
+
+  private static final String FILE = "file";
+  private static final Long LINE = 42L;
```
vam-google: Now we know which logs to search for the ultimate question answer =).

google-cloud-logging/src/main/java/com/google/cloud/logging/SourceLocation.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Objects;
+
+/**
+ * Additional information about the source code location that produced the log entry.
+ */
+public final class SourceLocation implements Serializable {
+
+  private static final long serialVersionUID = 8502955858162387374L;
+
+  String file;
```
saicheems: None, other than that I forgot to add the modifiers...

Done.

google-cloud-logging/src/main/java/com/google/cloud/logging/SourceLocation.java
saicheems: Done, I had omitted it because `LogEntry`'s `toPb` function did, but didn't really think beyond that :P

I've changed it to `public`, and the one in `LogEntry` as well.

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
saicheems: Done

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
saicheems: Created the helper `timestampFromMillis`

google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
saicheems: Created the helper `timestampFromMillis`

google-cloud-logging/src/test/java/com/google/cloud/logging/SourceLocationTest.java
saicheems: :)

Issue #2240, opened 2017-07-16
java.lang.IllegalArgumentException: Jetty ALPN/NPN has not been properly configured.


vscatarino:
I'm trying run a code that use a cloud-vision.
Everything works normally when I use my desktop(ubuntu / windows10). But when I use a raspberry appear the following error:

> Jul 16, 2017 5:52:39 PM io.grpc.internal.ChannelExecutor drain
> WARNING: Runnable threw exception in ChannelExecutor
> java.lang.IllegalArgumentException: Jetty ALPN/NPN has not been properly configured.
> 	at io.grpc.netty.GrpcSslContexts.selectApplicationProtocolConfig(GrpcSslContexts.java:174)
> 	at io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:151)
> 	at io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:139)
> 	at io.grpc.netty.GrpcSslContexts.forClient(GrpcSslContexts.java:109)
> 	at io.grpc.netty.NettyChannelBuilder.createProtocolNegotiatorByType(NettyChannelBuilder.java:335)
> 	at io.grpc.netty.NettyChannelBuilder.createProtocolNegotiator(NettyChannelBuilder.java:308)
> 	at io.grpc.netty.NettyChannelBuilder$NettyTransportFactory$DynamicNettyTransportParams.getProtocolNegotiator(NettyChannelBuilder.java:499)
> 	at io.grpc.netty.NettyChannelBuilder$NettyTransportFactory.newClientTransport(NettyChannelBuilder.java:448)
> 	at io.grpc.internal.CallCredentialsApplyingTransportFactory.newClientTransport(CallCredentialsApplyingTransportFactory.java:61)
> 	at io.grpc.internal.InternalSubchannel.startNewTransport(InternalSubchannel.java:209)
> 	at io.grpc.internal.InternalSubchannel.obtainActiveTransport(InternalSubchannel.java:186)
> 	at io.grpc.internal.ManagedChannelImpl$SubchannelImplImpl.obtainActiveTransport(ManagedChannelImpl.java:806)
> 	at io.grpc.internal.GrpcUtil.getTransportFromPickResult(GrpcUtil.java:568)
> 	at io.grpc.internal.DelayedClientTransport.reprocess(DelayedClientTransport.java:296)
[...]

PR #2314, opened 2017-08-07
Updating and expanding explanation of supported platforms

garrettjonesgoogle:
Updates #2240 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -877,7 +877,7 @@ Supported Platforms
 -------------------
 
 Clients in this repository use either HTTP or gRPC for the transport layer. All
-HTTP-based clients should work in all environments (theoretically).
+HTTP-based clients should work in all environments.
 
 For clients that use gRPC, the supported platforms are constrained by the platforms
 that [Forked Tomcat Native](http://netty.io/wiki/forked-tomcat-native.html) supports,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -876,12 +876,28 @@ Java 7 or above is required for using the clients in this repository.
 Supported Platforms
 -------------------
 
-This client is supported on Mac OS X, Windows and Linux (excluding Android and Alpine).
-Google Cloud Platform environments currently supported include GCE, GKE and GAE Flex.
-GAE Standard is not currently supported.
+Clients in this repository use either HTTP or gRPC for the transport layer. All
+HTTP-based clients should work in all environments (theoretically).
```
vam-google: I like the "theoretically" part in the readme, but probably it should not be there =). Otherwise it looks like we ourselves don't know how our libraries work (which is true, but it should remain a secret...)

README.md
```diff
@@ -876,12 +876,28 @@ Java 7 or above is required for using the clients in this repository.
...
+threading restrictions.
+
+Thus, the following are not supported:
+
+- Android
+- Alpine Linux (due to netty-tcnative requiring glibc, which is not present on Alpine)
+- Raspberry Pi (since it runs on the ARM architecture)
+- Google App Engine Standard Java 7
+
+The following environments should work (among others):
```
vam-google: Maybe it would make sense to specify standalone linux/windows/mac environments as a separate item(s) here to avoid making impression that this is a google-managed environments only library (there are libraries like that, for example the app-engine-std libs/clients). I see that there is "(among others)" note, but people may easily skip it, while reading.

README.md
garrettjonesgoogle: Done.

README.md
garrettjonesgoogle: Done.

PR #2317, opened 2017-08-08
Creating google-cloud-compat-checker

garrettjonesgoogle:
So far I am just bootstrapping this to check os & architecture. There is a lot more we could check, which we can add later. 

Currently, users have to build this themselves, because there have been no releases of it yet. Later, we can switch the instructions to use the maven artifact somehow. 

I intend to use this to diagnose https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2266 .

### DIFF
--- a/google-cloud-util/README.md
+++ b/google-cloud-util/README.md
@@ -1,0 +1,5 @@
+Google Cloud Java Utils
+==========================
+
+Google Cloud Java Utils is a module containing other modules with
+utilities for use with Google Cloud Java.
--- a/google-cloud-util/google-cloud-compat-checker/README.md
+++ b/google-cloud-util/google-cloud-compat-checker/README.md
@@ -1,6 +1,10 @@
 Google Cloud Java Compatibility Checker
 ==========================
 
+The Google Cloud Java Compatibility Checker is a utility to check
+compatibility of your environment with grpc-based clients in Google
+Cloud Java (e.g. Pub/Sub, Vision, Speech, and others).
+
 Usage:
 
 ```
--- a/google-cloud-util/google-cloud-compat-checker/pom.xml
+++ b/google-cloud-util/google-cloud-compat-checker/pom.xml
@@ -11,7 +11,7 @@
   </description>
   <parent>
     <groupId>com.google.cloud</groupId>
-    <artifactId>google-cloud-pom</artifactId>
+    <artifactId>google-cloud-util</artifactId>
     <version>0.21.2-alpha-SNAPSHOT</version>
   </parent>
   <properties>
--- a/google-cloud-util/pom.xml
+++ b/google-cloud-util/pom.xml
@@ -2,7 +2,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.google.cloud</groupId>
-  <artifactId>google-cloud-dep-util</artifactId>
+  <artifactId>google-cloud-util</artifactId>
   <version>0.21.2-alpha-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>Google Cloud Util</name>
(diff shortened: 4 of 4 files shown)


google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
```diff
@@ -0,0 +1,72 @@
...
+      ImmutableSet.of("linux-x86_64", "osx-x86_64", "windows-x86_64");
+
+  private static class OsDetector extends Detector {
+    public void detect(Properties props, List<String> classifierWithLikes) {
+      super.detect(props, classifierWithLikes);
+    }
+
+    @Override
+    protected void log(String s) {
+      // meh
```
vam-google: Really?

google-cloud-compat-checker/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative-boringssl-static</artifactId>
+      <version>2.0.3.Final</version>
+    </dependency>
+    <dependency>
+      <!-- It may look weird to have a compile dependency on a plugin,
+	   but we want to execute code from it at runtime, in order
+	   to not duplicate the logic ourselves -->
+      <groupId>kr.motd.maven</groupId>
```
vam-google: Do we really need this strange dependency? Usually it is possible to get enough of information about the running environment by means of core java with combination of environment variables and system properties, plus, maybe some more advanced (but still built in stuff). For example:

```java
PrintStream out = new PrintStream(System.out);

out.append("\n\nSYSTEM PROPERTIES:\n\n");
System.getProperties().list(out);

//out.append("\nENVIRONMENT VARIABLES:\n\n");
//Map<String, String> envVars = System.getenv();
//for (Map.Entry<String, String> entry : envVars.entrySet()) {
//  out.append(entry.getKey()).append("=").append(entry.getValue()).append('\n');
//}

```

Among other garbage, that outputs stuff like:
```
os.name=Linux
os.version=4.4.0-83-generic
os.arch=amd64

```

google-cloud-compat-checker/README.md
```diff
@@ -0,0 +1,25 @@
+Google Cloud Java Compatibility Checker
```
vam-google: What do you think about renaming this into something more generic (as I believe more utility stuff can/will be added here in the future)?

google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
garrettjonesgoogle: Yeah, the implementation inside the plugin just has log.info lines in its implementation.

google-cloud-compat-checker/pom.xml
garrettjonesgoogle: Sure, it's possible to get the information that way. *However*, the selection of netty-tcnative uses this logic. It seems both risky and wasteful to duplicate the logic, when all we want to do use get the exact result of the logic. Remember, we are trying to diagnose library loading errors here, so the more we diverge, the more likely we are to mismatch the actual library loading.

google-cloud-compat-checker/README.md
garrettjonesgoogle: I would probably want other utility stuff to be in a separate module - this module is likely to have weird dependencies (for example, a compile dependency on a maven plugin package).

vam-google: Utility classes often have weird dependencies, so I believe it is ok. I just feel uncomfortable with creating ad-hoc modules which do only one thing, as it contaminates the google-cloud java library (user most probably expects a module per client, various utility modules give wrong impression that that is some real google client to some mysterious google cloud service).

garrettjonesgoogle:
Then how about having a google-cloud-util module, and then put google-cloud-compat-checker under that?

google-cloud-compat-checker/README.md
garrettjonesgoogle: Then how about having a google-cloud-util module, and then put google-cloud-compat-checker under that?

vam-google: Sounds good to me

garrettjonesgoogle:
Moved the compat checker under google-cloud-util, PTAL

vam-google:
I meant some (optional) description in readme of what this module is about, since for utility modules it is not that clear. It doesn't have to be called "purpose".

PR #2320, opened 2017-08-08
Add numDmlAffectedRows to QueryResponse

### LINKED ISSUE #1895: "getNumDmlAffectedRows()" not avaiable in QueryResult class
I can see this method is already available in REST v2 API as well as api client libaray.

Can it be ported over to this libarary as well?

saicheems:
Resolves #1895

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
@@ -190,7 +190,7 @@ public class QueryResponse implements Serializable {
 
   @Override
   public final int hashCode() {
-    return Objects.hash(jobId);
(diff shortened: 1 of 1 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
```diff
@@ -191,6 +206,7 @@ public final boolean equals(Object obj) {
         && Objects.equals(etag, response.etag)
         && Objects.equals(result, response.result)
         && Objects.equals(jobId, response.jobId)
+        && Objects.equals(numDmlAffectedRows, response.numDmlAffectedRows)
```
garrettjonesgoogle: `hashCode` and `equals` are seriously mismatching - we need to align them.

saicheems: Fixed

garrettjonesgoogle:
LGTM

PR #2323, opened 2017-08-09
Using Jobs.insert instead of Jobs.query for queries

### LINKED ISSUE #1588: BigQuery - remove uses of jobs.query method.
The [jobs.query](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query) method should not be used in the client libraries.

Reasons: the query can time out, but it still might actually run, costing money & resources. Unless you poll for the results (not obvious you have to do this) you won't see them.

Also, the `jobs/query` API does not accept a job ID. This means that a retry on failure could result in duplicate queries being run, causing extra resource usage & charges.

Proposal:

- Remove all uses of `jobs/query`. Instead use [`jobs/insert`](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert).
- Deprecate `BigQuery.query(QueryRequest)`. Providing a synchronous method without an explicit job ID is dangerous for the reasons listed above. Alternatively, we could generate a job ID client-side, which would protect against duplicate queries from retries from the client libraries.
- Add a method `BigQuery.query(QueryRequest, JobId)` so users can explicitly provide a job ID for safe retries (even against program crashes if they save the job ID somewhere -- important for really big/expensive queries).

garrettjonesgoogle:
Fixes #1588 .

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -39,7 +39,6 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -602,7 +601,8 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
       if (request.getPageSize() != null) {
         options.add(QueryResultsOption.pageSize(request.getPageSize()));
       }
-      return getQueryResults(job.getJobId(), options.toArray(new QueryResultsOption[0]));
+      return getQueryResults(job.getJobId(),
+          options.toArray(new QueryResultsOption[options.size()]));
 
     } catch (RetryHelper.RetryHelperException e) {
       throw BigQueryException.translateAndThrow(e);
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -987,6 +987,19 @@ public static QueryResultsOption maxWaitTime(long maxWaitTime) {
   QueryResponse query(QueryRequest request);
```
tswast: Are we open to making breaking changes here? I'd somewhat like to do away with QueryRequest/QueryResponse altogether if possible and return a job here, especially since Job has `waitFor` but QueryResponse doesn't.

We tell users to run their own wait loop for QueryResponse

```
 while (!response.jobCompleted()) {
   Thread.sleep(1000);
   response = bigquery.getQueryResults(response.getJobId());
 }
```

We may need a special QueryJob class though so that query jobs can poll for completion using `getQueryResults` per BQ team's go/bq-client-reqs requirements doc.

garrettjonesgoogle: We are indeed open to breaking changes, but I wanted to limit the scope of this current PR because it's under the context of our fix-it week. @vam-google will be making more extensive changes for BigQuery in the coming weeks.

garrettjonesgoogle:
How about I just make the new function `@BetaApi`, since we'll go and break it later?

Issue #1683, opened 2017-03-02
Add ability to configure threads at least in PubSub and where appropriate


saturnism:
Currently, PubSub Subscriber creates multiple threads based on Runtime.availableProcessors(). This is a reasonable default but should be configurable, specifically when there are other competing threads pools, and also when running in Docker containers, where Runtime.availableProcessors() may not return the allocated CPUs (see https://bugs.openjdk.java.net/browse/JDK-6515172)

PR #2324, opened 2017-08-10
pubsub: provide a way to change number of streams

pongad:


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -589,6 +589,15 @@ public Builder setSystemExecutorProvider(ExecutorProvider executorProvider) {
       return this;
     }
 
+    /**
+     * Sets the number of pullers used to pull messages from the subscription. Defaults to the
+     * number of available processors.
+     */
+    public Builder setNumPullers(int numPullers) {
```
garrettjonesgoogle: This naming is a little opaque (for Java). Maybe: `setParallelPullCount` or something like that.

pongad: Do you have preference between `setParallelPullCount` and `setNumStreams`? I like the latter for being a little shorter but that's not a very important metric.

garrettjonesgoogle: I think the first is more meaningful from the user's perspective - they don't have to know that pubsub is implemented with streams exactly. In Java, length is a less important metric than other languages :-P

garrettjonesgoogle:
Side question - would this PR resolve #1683 ?

pongad:
Yes it does!

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #2329, opened 2017-08-10
Fix retrying logic for HTTP responses with "Content-Type" other than "application/json"

vam-google:
This should fix https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2098

### DIFF
--- a/google-cloud-core-http/src/main/java/com/google/cloud/http/BaseHttpServiceException.java
+++ b/google-cloud-core-http/src/main/java/com/google/cloud/http/BaseHttpServiceException.java
@@ -62,6 +62,7 @@ public class BaseHttpServiceException extends BaseServiceException {
           }
         } else {
           code = ((GoogleJsonResponseException) exception).getStatusCode();
+          retryable = BaseServiceException.isRetryable(code, null, idempotent, retryableErrors);
         }
       } else {
         // In cases where an exception is an instance of HttpResponseException but not


garrettjonesgoogle:
How does it get to the `exception instanceof GoogleJsonResponseException` branch if Content-Type is not application/json?

vam-google:
This is caused by (I believe) some key design decisions in [google-api-client](https://github.com/google/google-api-java-client). When you make a http request you may define it as a JSON request (the request class extends `AbstractGoogleJsonClientRequest`, and it is the case for `StorageRequest`). Unfortunately, the code is structured in a way that if your request is a JsonRequest, then it assumes that your **error** responses are also in Json. 

Places, where things start going in a wrong direction are: 
- [AbstractGoogleClientRequest#executeUnparsed()](https://github.com/google/google-api-java-client/blob/dev/google-api-client/src/main/java/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java#L432) - **requests** decides how to handle **response** error (is it request's responsibility?)
- [AbstractGoogleJsonClientRequest.newExceptionOnError()](https://github.com/google/google-api-java-client/blob/dev/google-api-client/src/main/java/com/google/api/client/googleapis/services/json/AbstractGoogleJsonClientRequest.java#L112) - the Json implementation of the request, makes an **assumption** that if the request is json, then the error response should also be json (seems like a wrong assumption)

But regardless of these details, I believe the proposed fix is not only the simplest but at the same time the best we can have practically. In our clients retries are configured by error codes, not by content type. The current code, for some reason starts ignoring the error code if it cannot parse the message. Why? Error code should sill be treated as error code. It is ok to try to get more information from the message (it potentially allows to handle error better), bit if that does not succeed, we should fallback to checking just error code.

garrettjonesgoogle:
ok, LGTM

Issue #2307, opened 2017-08-04
Create BOM for google-cloud-java


garrettjonesgoogle:
Example BOM:

https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-dependencies/pom.xml

Here is how Spring Cloud uses the BOM for their dependencies:

https://spring.io/blog/2016/09/26/spring-cloud-camden-release-and-brixton-sr6-are-available

The work for us is the following:

1. Extract dependencyManagement and pluginManagement to a dedicated pom.xml (e.g., artifact: google-cloud-java-bom)
2. In the parent pom.xml, import the BOM

Then consumers can import the BOM too, instead of specifying individual versions.

saturnism:
@meltsufin @joaoandremartins who's implemented for spring-cloud-gcp

vam-google:
`google-cloud-pom ` itself can be used as BOM. Added documentation in main README.md about how it can be done. Closing this issue.

saturnism:
i know it can _also_ be used as a BOM, but it's not necessarily the best practice. I don't feel we should ask people to import from a parent POM.  Parent POM is meant to be the parent of our projects.

saturnism:
@garrettjonesgoogle can we reopen this issue to track?

PR #2332, opened 2017-08-11
Extracted dependencies to a BOM

saturnism:
Example for For #2307. May still need some work:
1. scm/release not tested
1. consider moving test library dependencies out
1. reconsider what to put in the BOM, e.g., for consumers, the BOM should simply include our API artifacts


vam-google:
Hi Ray,

Thanks for cleaning the datastore dependencies problem! I missed that, I thought that all exact versions were already migrated to dependencyManagement portion of parent pom.sml.

The remaining work items, which you mentioned, were the main reason why it was chosen to use the parent pom as BOM instead of having a separate module. Specifically it breaks releasing process and makes things more complicated. The proposed change breaks releasing process completely and makes google-cloud-java modules structure inconsistent.

I understand the concern about the test dependencies, but it is always possible to override them either explicitly in `<dependencies>` or in `<dependencyManagement>` of the client pom. And it is not necessarily required, I believe having testing tools in sync with the client is also a good thing. At the same time google-cloud-pom does do not dictate them, because by default every client pom is expected to specify their test dependency versions explicitly. 

As for how BOM is done in spring, I believe the spring bom is generated during build, and in general its construction is quite complex. It is understandable, since spring is a huge framework (platform?) with tons of pluggable/optional parts. Google-cloud-java is just a client library, much more specific, smaller and simpler. 

So I agree that having separate bom is cleaner. At the same time: 
- it significantly complicates the releasing process 
- it makes modules structure in google-cloud-java inconsistent (the new google-cloud-dependencies is in a subfolder, listed in google-clou-pom modules list but is not a child of google-cloud-pom)
- at the same time using google-cloud-pom as BOM should work fine (after your datastore fix); I have hard time justifying those complications and inconsistencies when there is a cheaper alternative.

As for surface problem. I dont' understand what is wrong with having google-cloud-pom on the surface? It is already public and deployed. It describes the libraries and its dependencies, has links on documentation and version properties. 

According to [maven documentation](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Importing_Dependencies), one (and the first suggested) of the ways to implement a BOM is to use it as a parent for your project. Taking into account the structure of google-cloud-java I believe that would be a preferred way (compared to importing it independently in dependencyManagement section). It will also be cleaner from our releasing process perspective. The only problem with that is it would mean reparenting our google-cloud-pom, and since we already have several released GA libraries, I'm hesitant to do so.

garrettjonesgoogle:
Superseded by https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2679

PR #2335, opened 2017-08-11
Add `downloadTo` to `Blob`

### LINKED ISSUE #2107: [Feature Request] Add downloadTo to Storage Client
# Feature Request
The Java Cloud Storage client `google-cloud-storage` has a [sample to download a Blob](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java#L358) (shown below). The feature request is to add this functionality into the sample shown into an additional method `downloadTo` for [Class Blob](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java) to reduce complexity of downloading a blob using the Java Storage client. At the moment [other clients](https://cloud.google.com/storage/docs/object-basics#download) use this approach and allow a user to download a given Blob to a specified file.
 
xxnxbgjccop
 
# Language Comparison:
In other [client libraries](https://cloud.google.com/storage/docs/object-basics#download) (google-cloud-ruby) there exists a method to download a Blob to a specified file. 

h4q4mswiyt0
 
Another example using .NET:
 
bbfhutvhz08
 
# Example of usage
 
```
try {
  String bucketName = ”BUCKET_NAME”;
  String blobName = ”BLOB_NAME”;
 
  Blob blob = BlobId.of(bucketName, blobName);
  Path path = FileSystems.getDefault().getPath("output.ext”);
 
  blob.downloadTo(path);
} catch (IOException io) {
   //...
}
```

saicheems:
Resolves #2107

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -35,13 +35,14 @@ import com.google.cloud.storage.spi.v1.StorageRpc;
 import com.google.common.base.Function;
 import com.google.common.io.BaseEncoding;
 
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
+import java.io.OutputStream;
 import java.net.URL;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.Key;
 import java.util.Arrays;
@@ -73,6 +74,8 @@ public class Blob extends BlobInfo {
           return Blob.fromPb(pb.x(), pb.y());
         }
       };
+
+  private static final int DEFAULT_CHUNK_SIZE = 2 * 1024 * 1024;
 
   /**
    * Class for specifying blob source options when {@code Blob} methods are used.
@@ -196,20 +199,23 @@ public class Blob extends BlobInfo {
    * Downloads this blob to the given file path.
    *
    * @param path destination
-   * @throws IOException upon failure
+   * @throws StorageException upon failure
    */
-  public void downloadTo(Path path) throws IOException {
-    FileOutputStream outputStream = new FileOutputStream(path.toFile());
-    try (ReadChannel reader = reader()) {
+  public void downloadTo(Path path) throws StorageException {
+    try (
+        OutputStream outputStream = Files.newOutputStream(path);
+        ReadChannel reader = reader()
+    ) {
       WritableByteChannel channel = Channels.newChannel(outputStream);
-      ByteBuffer bytes = ByteBuffer.allocate(64 * 1024);
+      ByteBuffer bytes = ByteBuffer.allocate(DEFAULT_CHUNK_SIZE);
       while (reader.read(bytes) > 0) {
         bytes.flip();
         channel.write(bytes);
         bytes.clear();
       }
+    } catch (IOException e) {
+      throw new StorageException(e);
     }
-    outputStream.close();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -187,6 +193,26 @@ public static BlobSourceOption decryptionKey(String key) {
...
+   * Downloads this blob to the given file path.
+   *
+   * @param path destination
+   * @throws IOException upon failure
+   */
+  public void downloadTo(Path path) throws IOException {
+    FileOutputStream outputStream = new FileOutputStream(path.toFile());
+    try (ReadChannel reader = reader()) {
+      WritableByteChannel channel = Channels.newChannel(outputStream);
+      ByteBuffer bytes = ByteBuffer.allocate(64 * 1024);
```
vam-google: I'm afraid that 64k is too small for such operation. The actual reader chooses the maximum of the capacity of the supplied ByteBuffer and [DEFAULT_CHUNK_SIZE](https://github.com/vam-google/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadChannel.java#L41), which is 2Mb. I believe it would make sence to make the byte buffer to match the default chunk size (I see that 64 * 1024 was chosen from the documentation of the `Blob#read()`, but the doc is for external usage, for internal usage we can stick to the actually used buffer sizes).

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -187,6 +193,26 @@ public static BlobSourceOption decryptionKey(String key) {
   }
 
   /**
+   * Downloads this blob to the given file path.
+   *
+   * @param path destination
+   * @throws IOException upon failure
+   */
+  public void downloadTo(Path path) throws IOException {
```
vam-google: I believe the convention is to throw StorageException upon failure. Also it is considered a good practice to avoid throwing of checked exceptions when applicable (there is no universal agreement, but checked exceptions have bad reputation these days, for example http://www.artima.com/intv/handcuffs.html).

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -187,6 +193,26 @@ public static BlobSourceOption decryptionKey(String key) {
...
+    FileOutputStream outputStream = new FileOutputStream(path.toFile());
```
vam-google: Probably it is worth making this code fully `java.nio` (without mixing `nio` and `pre-nio` file writing techniques and classes). Specifically `Files.newOutputStream(path)`. I'm also not sure what should be done with buffering here. None of these methods creaet buffered streams, but looks like we don't need it, as we have buffer on the outside.

Also, please use the "try with resources" block for the stream too (not only for the reader).

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
saicheems: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
saicheems: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
saicheems: Done

lesv:
One of three Travis builds failed - it looks like this can be either retested or merged.

vam-google:
Our Mac builds are broken (not related to this change).

PR #2336, opened 2017-08-11
Removing Cloud Storage section of main README

garrettjonesgoogle:
Just starting with one API. I intend to move out all of the API-specific sections from the main readme because it has gotten too long.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Java idiomatic client for [Google Cloud Platform][cloud-platform] services.
 This client supports the following Google Cloud Platform services at a [GA](#versioning) quality level:
 -  [Stackdriver Logging](#stackdriver-logging-ga) (GA)
 -  [Cloud Datastore](#google-cloud-datastore-ga) (GA)
--  [Cloud Storage](https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-storage) (GA)
+-  [Cloud Storage](google-cloud-storage) (GA)
 -  [Cloud Translation](#google-translation-ga) (GA)
 
 This client supports the following Google Cloud Platform services at a [Beta](#versioning) quality level:
--- a/google-cloud-storage/README.md
+++ b/google-cloud-storage/README.md
@@ -120,7 +120,7 @@ String contentString = new String(content, UTF_8);
 A complete example for accessing blobs can be found at
 [CreateBlob.java](./google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/CreateBlob.java).
 
-### Updating data
+#### Updating data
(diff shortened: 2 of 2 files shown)


google-cloud-storage/README.md
```diff
@@ -99,19 +99,45 @@ Bucket bucket = storage.create(BucketInfo.of(bucketName));
...
-String blobContent = new String(blob.getContent(), UTF_8);
+BlobId blobId = BlobId.of(bucketName, "my_blob_name");
+byte[] content = storage.readAllBytes(blobId);
+String contentString = new String(content, UTF_8);
 ```
 
+A complete example for accessing blobs can be found at
+[CreateBlob.java](./google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/CreateBlob.java).
+
+### Updating data
```
vam-google: Needs one more `#`

google-cloud-storage/README.md
```diff
@@ -99,19 +99,45 @@ Bucket bucket = storage.create(BucketInfo.of(bucketName));
...
+  System.out.println(new String(prevContent, UTF_8));
+  WritableByteChannel channel = blob.writer();
+  channel.write(ByteBuffer.wrap("Updated content".getBytes(UTF_8)));
+  channel.close();
+}
+```
+
+The complete source code can be found at
+[UpdateBlob.java](./google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/UpdateBlob.java).
+
```
vam-google: (Commenting here, since I don't know how to comment on an unchanged line).
This is the comment for line 162.
It says "we put together all the code shown above into one program", but the new update example is not added in the snippet. Please update the snippet also (and probably rename it, since since it will be doing an update too).

README.md
```diff
@@ -15,7 +15,7 @@ Java idiomatic client for [Google Cloud Platform][cloud-platform] services.
 This client supports the following Google Cloud Platform services at a [GA](#versioning) quality level:
 -  [Stackdriver Logging](#stackdriver-logging-ga) (GA)
 -  [Cloud Datastore](#google-cloud-datastore-ga) (GA)
--  [Cloud Storage](#google-cloud-storage-ga) (GA)
+-  [Cloud Storage](https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-storage) (GA)
```
vam-google: This  breaks the "isolation" of the forks. Is it possible to do something like just `#tree/master/google-cloud-storage` ?

garrettjonesgoogle: It looks like I can simply do `google-cloud-java` and it works.

google-cloud-storage/README.md
garrettjonesgoogle: I have fixed it by explicitly describing the scope instead of saying it has all the code above.

PR #2347, opened 2017-08-16
Checking more things in google-cloud-compat-checker

garrettjonesgoogle:


google-cloud-util/google-cloud-compat-checker/README.md
```diff
@@ -23,7 +23,19 @@ OS details:
   os.detected.classifier: linux-x86_64
   os.detected.release: ubuntu
   os.detected.release.version: 14.04
+JVM details:
+  Java version: 1.8.0_144-b01
+  Java specification version: 1.8
```
vam-google: This looks redundant

google-cloud-util/google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
```diff
@@ -43,30 +44,96 @@ protected void logProperty(String s, String s1) {
   }
 
   public static void main(String[] args) {
+    Boolean checkResult = check();
+    if (checkResult != null && !checkResult) {
+      System.exit(1);
```
vam-google: I agree with codacy-bot. This looks more like a bash script now. Why not just throwing an exception with a meaningful message?

google-cloud-util/google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
```diff
@@ -43,30 +44,96 @@ protected void logProperty(String s, String s1) {
...
+    }
+  }
+
+  public static Boolean check() {
```
vam-google: Please split this method. Potentially grouping OS, JVM and OpenSSL things (potentially changing output order too: first OS details + check, and only then JVM details + check etc). Codacy-bot complains that it this method is too complex and it is right. Also, splitting makes adding more stuff easier (and probably there will be more stuff).

google-cloud-util/google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
```diff
@@ -43,30 +44,96 @@ protected void logProperty(String s, String s1) {
...
+      System.out.println("  prevent you from using grpc-based APIs.");
+      if (warnings) {
+        System.out.println("  However, there were some warnings to watch out for.");
+      }
+    }
+
+    if (!compatible) {
+      return false;
+    } else {
+      return null; // don't want to promise the answer yet
```
vam-google: I believe it is ok to return true. Nobody expects this method to guarantee anything, it checks what it can check and it is ok. There can be more checkers outside, and checks are usually done via &&, so even if this returns true and some other check returns false it will be still false.

google-cloud-util/google-cloud-compat-checker/README.md
garrettjonesgoogle: We would build logic on the specification version, but probably not the full java version, but I think it's useful to have some extra details in the output just for debugging purposes.

google-cloud-util/google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
garrettjonesgoogle: This is `main`, throwing exceptions is not helpful. We already have meaningful messages printed in `check()`. This is indeed meant to be run like a script, but it has to be in java to make the right calls and version checks.

google-cloud-util/google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
garrettjonesgoogle: I don't think there is value in splitting this yet... it's just a straight run of checks; putting structure on this requires more design thought than I think is merited for this problem.

google-cloud-util/google-cloud-compat-checker/src/main/java/com/google/cloud/compatchecker/GoogleCloudCompatChecker.java
garrettjonesgoogle: Ok I'll switch this to native `boolean`.

PR #2356, opened 2017-08-17
Adding some unit tests for cloud spanner

vkedia:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -205,7 +207,8 @@ public void cancelled(Context context) {
         logger.log(Level.FINE, "Retryable exception, will sleep and retry", e);
         backoffSleep(context, backOff);
       } catch (Exception e) {
-        throw Throwables.propagate(e);
+        Throwables.throwIfUnchecked(e);
+        throw new RuntimeException(e);
```
vam-google: I agree with codacy-bot (even though the replaced Throwables.propagate() is doing exactly same. Can the exception be a non-retriable SpannerException instead of being raw RuntimeException?

garrettjonesgoogle:
Don't merge this until I'm able to do a release with our current code in master

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
vkedia: Done

vkedia:
@garrettjonesgoogle Can this be merged now?

garrettjonesgoogle:
Yes it can be merged now.

PR #2359, opened 2017-08-18
Fixing retries of Storage.create

garrettjonesgoogle:
* For `Blob create(BlobInfo blobInfo, InputStream content, BlobWriteOption... options)`: Turning off retries and marking `@Deprecated`, because once the `InputStream` has been consumed, it can't be consumed again.
* For `Blob create(BlobInfo blobInfo, byte[] content, BlobTargetOption... options)`: Re-creating the `InputStream` on every call, so that the same data is resent to the server.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
@@ -18,7 +18,6 @@ package com.google.cloud.storage;
 
 import static com.google.cloud.storage.testing.ApiPolicyMatcher.eqApiPolicy;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.easymock.EasyMock.eq;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -606,7 +605,7 @@ public class StorageImplTest {
   }
 
   @Test
-  public void testCreateBlobRetryableException() throws IOException {
+  public void testCreateBlobFromStreamRetryableException() throws IOException {
     Capture<ByteArrayInputStream> capturedStream = Capture.newInstance();
     ByteArrayInputStream fileStream = new ByteArrayInputStream(BLOB_CONTENT);
     BlobInfo.Builder infoBuilder = BLOB_INFO1.toBuilder();
@@ -631,7 +630,7 @@ public class StorageImplTest {
     // shouldn't retry.
     thrown.expect(StorageException.class);
     thrown.expectMessage("internalError");
-    storage.create(BLOB_INFO1, fileStream);
+    storage.create(infoWithHashes, fileStream);
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -132,24 +133,29 @@ public Blob create(BlobInfo blobInfo, byte[] content, BlobTargetOption... option
...
   @Override
+  @Deprecated
   public Blob create(BlobInfo blobInfo, InputStream content, BlobWriteOption... options) {
     Tuple<BlobInfo, BlobTargetOption[]> targetOptions = BlobTargetOption.convert(blobInfo, options);
-    return create(targetOptions.x(), content, targetOptions.y());
+    StorageObject blobPb = targetOptions.x().toPb();
+    Map<StorageRpc.Option, ?> optionsMap = optionMap(targetOptions.x(), targetOptions.y());
+    InputStream inputStreamParam = firstNonNull(content, new ByteArrayInputStream(EMPTY_BYTE_ARRAY));
+    // retries are not safe when the input is an InputStream, so we can't retry.
+    return Blob.fromPb(this, storageRpc.create(blobPb, inputStreamParam, optionsMap));
```
vam-google: Another possibility would be to use InputStream.mark()/InputStream.reset() methods combination. Though even with those methods it is hard to provide robust implementation. At the same time with those methods we can reuse inputstream version of create() method for the byte[] version (ByteArrayInputStream must obviously support mark()/reset() in the most efficient way possible).

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -1460,6 +1463,7 @@ public static Builder newBuilder() {
    * @return a [@code Blob} with complete information
    * @throws StorageException upon failure
    */
+  @Deprecated
```
vam-google: Now I'm wondering is it possible to undeprecate something in the future? It seems like an essential method and we will eventually want to undeprecate it in the future (for example if we manage to implement it reliably with mark()/reset() calls)? Or, I know that we talked about it, but now I think maybe we should not mark this method as deprecated as we do not provide a non-deprecated alternative to this method (the byte[] version is not enough).

garrettjonesgoogle: I would like to create a `RestartableInputStream` abstraction and create a new entry point that requires that. It would not have the drawbacks of `mark()`/`reset()`, although of course it would require users to understand a new concept. I think it's the least worst solution for this problem. I will do that later, though.

nafg:
So for now we're not supposed to use it and are required to either load everything into RAM or write a lot of boilerplate?

frankyn:
Ping @garrettjonesgoogle and @andreamlin for user feedback on this PR. Is there a plan to introduce `RestartableInputStream` as stated?

garrettjonesgoogle:
I have filed https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2620 to track the work of introducing `RestartableInputStream`.

PR #2360, opened 2017-08-18
Disabling failing integration test

garrettjonesgoogle:


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -794,7 +794,8 @@ public void testCopyBlobUpdateMetadata() {
     assertTrue(storage.delete(BUCKET, targetBlobName));
   }
 
-  @Test
+  //Re-enable this test when it stops failing
+  //@Test
```
vam-google: Why not `@Ignore` ?

garrettjonesgoogle: That requires an import and I'm not on a good connection with an IDE to make that easy :-P

vam-google: :scream:

digger:
@garrettjonesgoogle , is there a reason this test is still disabled? What was that backend issue that made the test fail?

We were looking for a way to update storage class of a blob using this client and found this test. Does this mean the client can't be used to change storage classes?

garrettjonesgoogle:
@digger it looks like we never followed up. I would suggest filing a new issue to have it looked at.

PR #2376, opened 2017-08-24
removing "synchronously" from handler javadoc

jabubake:
Synchronicity is default async now, and can be controlled using the `Synchronicity` field.


garrettjonesgoogle:
LGTM

PR #2382, opened 2017-08-25
Directs users to cloud-trace-java for instrumentation

mtwo:
Trying to reduce user confusion


garrettjonesgoogle:
I don't know that we've settled how we want the landscape of trace libraries to look yet...

saturnism:
there def needs to be a raw client, like this one.

the other "trace libraries" like opencensus and zipkin/brave, is really an instrumentation and distributed tracing framework.  Both of which may still require this client library to work.

PR #2387, opened 2017-08-28
Fix visibility of setRequesterPays in BucketInfo

shinfan:


jabubake:
@shinfan  can you merge ?

PR #2390, opened 2017-08-29
Fix confusing Translation sample code

vam-google:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2389

### DIFF
--- a/google-cloud-translate/README.md
+++ b/google-cloud-translate/README.md
@@ -126,7 +126,7 @@ import com.google.cloud.translate.Translate.TranslateOption;
 import com.google.cloud.translate.Translation;
 ```
 
-Then add the following code to translate a text, specifying English as its source language and the previously obtained `detectedLanguage` as the target language (in this concrete example it will be Spanish):
+Then add the following code to translate a text, specifying English as its source language and the previously obtained `detectedLanguage` as the target language (in this concrete example it will be Spanish (`"es"`)):
 
 ```java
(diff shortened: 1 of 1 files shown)


google-cloud-translate/README.md
```diff
@@ -126,13 +126,13 @@ import com.google.cloud.translate.Translate.TranslateOption;
...
 ```
 
-Then add the following code to translate a text (specifying its source language):
+Then, for example to translate from English to Spanish, add the following code:
 
 ```java
 Translation translation = translate.translate(
     "World",
     TranslateOption.sourceLanguage("en"),
-    TranslateOption.targetLanguage(detectedLanguage));
```
garrettjonesgoogle: In the larger context of the README, it seems like we should use `detectedLanguage` (for source language). I took a peek at https://github.com/GoogleCloudPlatform/google-cloud-java/blob/70d98a29e36474a73be93781b49f4e3d54bef8ad/google-cloud-examples/src/main/java/com/google/cloud/examples/translate/snippets/DetectLanguageAndTranslate.java , and it looks like its code doesn't match the README code (which the README claims it does). We should fix that too in this PR.

vam-google:
PTAL

vam-google:
Replacing with a proper PR from the repo fork (as now it has more than one file) https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2394. Closing this PR without merging.

PR #2394, opened 2017-08-30
Fix confusing Translation sample code

vam-google:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2389

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/translate/snippets/DetectLanguageAndTranslate.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/translate/snippets/DetectLanguageAndTranslate.java
@@ -49,7 +49,7 @@ public class DetectLanguageAndTranslate {
     Detection detection = translate.detect(mysteriousText);
     String detectedLanguage = detection.getLanguage();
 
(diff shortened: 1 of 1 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/translate/snippets/DetectLanguageAndTranslate.java
```diff
@@ -42,16 +42,19 @@ public static void main(String... args) {
...
-    Detection detection = translate.detect("Hola");
+    // Text of an "unknown" language to detect and then translate into English
+    final String mysteriousText = "Hola Mundo";
+
+    // Detect the language of the mysterious text
+    Detection detection = translate.detect(mysteriousText);
     String detectedLanguage = detection.getLanguage();
 
-    // Translate some text
+    // Translate the mysterious text from Spanish to English
```
garrettjonesgoogle: It's odd to say it is "mysterious text" and also that it is Spanish at the same time (which means it's not so mysterious :-P) (maybe just remove the "from Spanish" part here)

vam-google:
PTAL

PR #2403, opened 2017-09-02
Fixes typo in javadocs for Datastore's BaseEntity class.

TonyRippy:
Replaces "...if not such property" with "...if no such property."


TonyRippy:
I am a Googler, covered by the Corporate CLA.

PR #2409, opened 2017-09-06
pubsub: cap stream reconnect time

### LINKED ISSUE #2406: [0.22.0-beta] StreamingSubscriberConnection unlimited increasing the channelReconnectBackoff duration
The Pub/Sub Subscriber of version 0.22.0-beta uses streaming as its subscription connection by default. This setting cannot be changed, because of the default access of the Subcriber.Builder#setUseStreaming method.
The Problem with the StreamingSubscriberConnection is, that on each re-connection (which occurs quite often) the channelReconnectBackoff duration is doubled. There is no upper limit, nor any reset of the channelReconnectBackoff duration on successful messages.

pongad:
Previously we exponentially grow stream reconnect backoff time
without bound, so unlucky users will wait forever waiting for
a connection.
This commit caps the reconnect at 10s (the same as polling).

Previously we only reset the polling time after the stream closes
without error. This usually doesn't happen because the streams are
long running and don't close until some error is encountered.
This commit resets the polling time if we receive any response
from the stream.

Updates #2406.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
@@ -52,7 +52,8 @@ final class StreamingSubscriberConnection extends AbstractApiService implements 
   private static final Duration MAX_CHANNEL_RECONNECT_BACKOFF = Duration.ofSeconds(10);
   private static final int MAX_PER_REQUEST_CHANGES = 10000;
 
-  private final AtomicLong channelReconnectBackoffMillis = new AtomicLong(INITIAL_CHANNEL_RECONNECT_BACKOFF.toMillis());
+  private final AtomicLong channelReconnectBackoffMillis =
+      new AtomicLong(INITIAL_CHANNEL_RECONNECT_BACKOFF.toMillis());
 
   private final SubscriberStub asyncStub;
 
@@ -189,10 +190,8 @@ final class StreamingSubscriberConnection extends AbstractApiService implements 
             logger.log(Level.WARNING, "Terminated streaming with exception", cause);
             if (StatusUtil.isRetryable(cause)) {
               long backoffMillis = channelReconnectBackoffMillis.get();
-              long newBackoffMillis = backoffMillis*2;
-              if (newBackoffMillis > MAX_CHANNEL_RECONNECT_BACKOFF.toMillis()) {
-                newBackoffMillis = MAX_CHANNEL_RECONNECT_BACKOFF.toMillis();
-              }
+              long newBackoffMillis =
+                  Math.min(backoffMillis * 2, MAX_CHANNEL_RECONNECT_BACKOFF.toMillis());
               channelReconnectBackoffMillis.set(newBackoffMillis);
 
               executor.schedule(


pongad:
@davidtorres Looking at this more, I'm not sure if this class is properly synchronized. Do we need a mutex around requestObserver?

I think `MessageDispatcher` could try to ack at the same time `initialize` tries to reestablish a stream. I think we can get a read-write race on the `requestObserver` variable?

Andy2003:
Additionally the Builder for the Subscriber should be fixed, so that the user can chosse between polling or streaming. See: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/ce59d34fa59c66add4c8370f1ca760f127a91b56/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java#L607 (should be public)

pongad:
@Andy2003 The private access was intentional. We'd like to migrate to streaming exclusively. Do you have a use case that streaming does not solve?

Andy2003:
A solution to work-around the current bug would be to switch back to pulling. But this is not possible, so the only choice was to downgrade to 0.21.1-beta.

pongad:
If you are running into this in production, I think the right answer is to roll back instead of forward. It does take some time for us to cut a new release so even with this PR merged, we might not be able to cut one immediately.

That said, it might make sense, in the next release, to make this method public and `@Deprecated` until more bugs in streaming pull is flushed out. @davidtorres What do you think of this?

mdietz94:
@davidtorres is OOO this week FYI. I would be against marking deprecated, the message should be that StreamingPull is the new preferred API - the first users to upgrade to a new version will always be taking on some amount of risk, and I think rolling back makes more sense than keeping the old version of the code alive just in case there are bugs - we discuss more offline if you feel strongly.

pongad:
@mdietz94 I agree with you but wanted to bring others into the conversation since I don't want to be a dictator :)

Would you like to review this change? Or should we wait for David?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -185,8 +188,13 @@ public void onFailure(Throwable cause) {
             }
             logger.log(Level.WARNING, "Terminated streaming with exception", cause);
             if (StatusUtil.isRetryable(cause)) {
-              long backoffMillis = channelReconnectBackoff.toMillis();
-              channelReconnectBackoff = channelReconnectBackoff.plusMillis(backoffMillis);
+              long backoffMillis = channelReconnectBackoffMillis.get();
+              long newBackoffMillis = backoffMillis*2;
```
mdietz94: nit: Math.min(backoffMillis * 2, MAX_CHANNEL_RECONNECT_BACKOFF.toMillis());

pongad:
@mdietz94 PTAL. The failing test seems to be coming from logging and unrelated to this change. Do you have any opinion about the potential race I called out earlier?

mdietz94:
I agree it looks like there could be a race, unless there is some other sort of guarantee that we won't try to ack while restarting a stream. I think a Mutex would definitely fix it, but maybe it's worth looking through the initial review, I didn't take part in it and maybe they discussed why this wouldn't e a problem then.

If it isn't a race it definitely needs a comment explaining why.

pongad:
@mdietz94 I looked into this more. I'm almost certain there's a race. I'll merge this PR now and send out a fix separately.

PR #2415, opened 2017-09-08
travis: move to trusty

pongad:
Since Precise has already reached EOL, we should move.


landrito:
This looks LGTM but I don't think I'm the right person to give approval to this. @vam-google can you take a look?

pongad:
I didn't do anything. Do you happen to know how long ago did it not work? I *think* Travis fixed whatever the problem was.

PR #2416, opened 2017-09-08
pubsub: fix race condition in streaming connection

pongad:


pongad:
@davidtorres @mdietz94

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -52,14 +54,15 @@
...
   private final SubscriberStub asyncStub;
-
   private final String subscription;
   private final ScheduledExecutorService executor;
   private final MessageDispatcher messageDispatcher;
+
+  private final AtomicLong channelReconnectBackoffMillis =
+      new AtomicLong(INITIAL_CHANNEL_RECONNECT_BACKOFF.toMillis());
+
+  private final Lock lock = new ReentrantLock();
```
mdietz94: I think in general we prefer ReadWriteLock to ReentrantLock, it should be possible to guarantee we won't have re-entry, and I think it's generally preferred to disallow it in order to simplify the concurrency model.

If you feel strongly and/or this is used widely elsewhere feel free to disregard.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -101,23 +104,35 @@ protected void doStart() {
   @Override
   protected void doStop() {
     messageDispatcher.stop();
-    requestObserver.onError(Status.CANCELLED.asException());
-    notifyStopped();
+
+    lock.lock();
+    try {
+      requestObserver.onError(Status.CANCELLED.asException());
```
mdietz94: Unless you think requestObserver could be null, I don't believe it's possible for this to throw an exception.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -101,23 +104,35 @@ protected void doStart() {
...
     StreamingPullResponseObserver(SettableFuture<Void> errorFuture) {
       this.errorFuture = errorFuture;
     }
 
     @Override
     public void beforeStart(ClientCallStreamObserver<StreamingPullRequest> requestObserver) {
-      StreamingSubscriberConnection.this.requestObserver = requestObserver;
-      requestObserver.disableAutoInboundFlowControl();
+      thisRequestObserver = requestObserver;
+      lock.lock();
```
mdietz94: This should not require a lock, we have the guarantee here that nothing could have killed the stream since this is beforeStart, we cannot have had our error callback hit. We can also just use thisRequestObserver since it does not need to be lock protected.

Also why the change from this.requestObserver to thisRequestObserver?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -128,9 +143,22 @@ public void onNext(StreamingPullResponse response) {
...
+               * outer class. By the time this Runnable is called, the stream might have already
+               * failed and the outer class might have created a new stream. In this case, we should
+               * request more messages from the already-failed stream (a no-op), otherwise the
+               * current observer might get more messages than it should.
+               */
               if (isAlive()) {
-                requestObserver.request(1);
+                lock.lock();
+                try {
+                  thisRequestObserver.request(1);
```
mdietz94: This can also be done under a reader lock, the only writer lock we should need to hold is while we are aborting and restarting the stream.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -220,8 +260,13 @@ public void sendAckOperations(
       List<String> acksToSend, List<PendingModifyAckDeadline> ackDeadlineExtensions) {
     List<StreamingPullRequest> requests =
         partitionAckOperations(acksToSend, ackDeadlineExtensions, MAX_PER_REQUEST_CHANGES);
-    for (StreamingPullRequest request : requests) {
-      requestObserver.onNext(request);
+    lock.lock();
```
mdietz94: Can be a reader lock, like above.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -271,9 +316,14 @@ public void sendAckOperations(
 
   public void updateStreamAckDeadline(int newAckDeadlineSeconds) {
     messageDispatcher.setMessageDeadlineSeconds(newAckDeadlineSeconds);
-    requestObserver.onNext(
-        StreamingPullRequest.newBuilder()
-            .setStreamAckDeadlineSeconds(newAckDeadlineSeconds)
-            .build());
+    lock.lock();
```
mdietz94: Can be a reader lock.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
pongad: I agree. Though it seems like doing this is a good practice in case we add something else to this in the future? If nothing else, it is recommended by the [docs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html)

I guess my view of this is "I know the gun isn't loaded but why point it at my own foot anyway", when the cost of pointing it away is only a few characters.

I don't feel super strongly about this though. Do you?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
pongad: You're right about the lock. Done.

I documented `thisRequestObserver` in the `onNext` method. I moved the doc here so it reads better.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
pongad: I don't think a read lock is right here. `StreamObserver`s don't need to be thread-safe. If we make this a read-lock, it is possible for us to call `request` and `onNext` concurrently.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
mdietz94: Ah okay, fair enough, based on name alone I thought it may be re-entrant specifically.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
mdietz94: I don't feel strongly, if you do feel free to change it back.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
mdietz94: If you see ClientCall.java line 177, while StreamObserver is not necessarily thread-safe, the request method in particular is guaranteed to be able to be called concurrently without external synchronization.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
mdietz94: I stand corrected here, onNext is not guaranteed to be thread-safe.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
mdietz94: I stand corrected here, onNext is not guaranteed to be thread-safe.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
pongad: Ah I see. Makes sense. Done

pongad:
@mdietz94 PTAL

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -148,12 +148,7 @@ public void onNext(StreamingPullResponse response) {
             public void run() {
               // Only request more if we're not shutdown.
               if (isAlive()) {
-                lock.lock();
-                try {
-                  thisRequestObserver.request(1);
-                } finally {
-                  lock.unlock();
-                }
+                thisRequestObserver.request(1);
```
mdietz94: This can be called after onError has been called on thisRequestObserver. Are we certain that's valid?

pongad:
Thinking about this more, I think I found more issues with streams. I'll follow up with you and David on email.

pongad:
@mdietz94 @davidcavazos Please let me know if this looks OK to you. The error handling is still obviously wrong. However, I think this is a big enough improvement that we can submit this as an intermediate state.

- At least we don't have the race condition anymore.
- If `onNext` throws, it might stop us from scheduling the next deadline extension. After this change, at least it can't do crazy things like that anymore. It will make us loose some acks and stuff like that, but that's still a lot better.

davidcavazos:
Did you mean @davidtorres ?

pongad:
Yes I did! Sorry for the spam!

pongad:
@davidtorres

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -131,9 +149,18 @@ public void onNext(StreamingPullResponse response) {
...
             @Override
             public void run() {
-              // Only if not shutdown we will request one more batches of messages to be delivered.
-              if (isAlive()) {
-                requestObserver.request(1);
+              // Only request more if we're not shutdown.
+              // If errorFuture is done, the stream has either failed or hung up,
+              // and we don't need to request.
+              if (isAlive() && !errorFuture.isDone()) {
+                lock.lock();
```
mdietz94: This should only be a reader lock, as described earlier

pongad: I think a normal lock here is OK. There's only one thread that can call `request` at a given time, so making it a read lock shouldn't make anything go faster.

mdietz94: Why can only one thread call request at a time?

pongad: The callback is called every message *batch* processed. Eg, if the `StreamingPullResponse` we receive has 20 messages. We'll call the callback once after all 20 has been processed. The `request(1)` here means one response, not one message.

For each connection, we request one response, process all messages, then execute the callback which requests another. So there's no way this method can be executed in parallel.

I ran the loadtest, apart from https://github.com/GoogleCloudPlatform/pubsub/pull/124, we observed no regression.

PR #2420, opened 2017-09-08
Update to GAX 1.7.0 and 0.24.0

shinfan:
- Sync with api-client-staging
- Make compatibility changes to GAX 1.7.0 and 0.24.0 (Explicit exception types)

### DIFF
--- a/google-cloud-core-grpc/src/test/java/com/google/cloud/grpc/BaseGrpcServiceExceptionTest.java
+++ b/google-cloud-core-grpc/src/test/java/com/google/cloud/grpc/BaseGrpcServiceExceptionTest.java
@@ -27,6 +27,7 @@ import static org.junit.Assert.assertTrue;
 
 import com.google.api.gax.grpc.GrpcStatusCode;
 import com.google.api.gax.rpc.ApiException;
+import com.google.api.gax.rpc.InternalException;
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryHelper;
 import io.grpc.Status.Code;
@@ -74,8 +75,8 @@ public class BaseGrpcServiceExceptionTest {
     assertNull(serviceException.getDebugInfo());
 
     Exception cause = new IllegalArgumentException("bad arg");
-    ApiException apiException =
-        new ApiException(MESSAGE, cause, GrpcStatusCode.of(Code.INTERNAL), NOT_RETRYABLE);
+    InternalException apiException =
+        new InternalException(MESSAGE, cause, GrpcStatusCode.of(Code.INTERNAL), NOT_RETRYABLE);
     serviceException = new BaseGrpcServiceException(apiException);
     assertFalse(serviceException.isRetryable());
     assertEquals(MESSAGE, serviceException.getMessage());


google-cloud-core-grpc/src/test/java/com/google/cloud/grpc/BaseGrpcServiceExceptionTest.java
```diff
@@ -73,8 +74,8 @@ public void testBaseServiceException() {
     assertNull(serviceException.getDebugInfo());
 
     Exception cause = new IllegalArgumentException("bad arg");
-    GrpcApiException apiException =
-        new GrpcApiException(MESSAGE, cause, Code.INTERNAL, NOT_RETRYABLE);
+    ApiException apiException =
```
vam-google: Shouln't this be an InternalException? 
This is more of a question why do individual exceptions allow to set any different status code (it seems like there is 1:1 relationship between status code and exception, but currently it is possible to have InternalException with, for example, ABORTED status code)?

Is it intentional? If no, then I believe it would make sence to update gax (remove status code from individual exceptions parameters and hardcode the individual codes for each exception in constructor implementations).

shinfan: Yes this should be internal exception.

The reason that each exception can still accept a status code because the status code can either be a GrpcStatusCode or HttpStatusCode which can have completely different code values. Otherwise it would be hard-coding both status code into each constructor and the user will figure out which code to use. Current approach is acceptable because user will be rarely constructing exceptions by themselves so this is just the implementation details underlying gax.

shinfan:
The travis failed but I don't think it is caused by this particular PR and I could not reproduce the failure on my local environment (all tests passed on my machine):
```
java.lang.AssertionError: 
  Unexpected method call Logging.write([LogEntry{logName=null, resource=null, timestamp=123456789, receiveTimestamp=null, severity=DEBUG, insertId=null, httpRequest=null, labels={levelName=FINEST, levelValue=300}, operation=null, trace=null, sourceLocation=null, payload=StringPayload{type=STRING, data=message}}], WriteOption{name=LOG_NAME, value=java.log}, WriteOption{name=RESOURCE, value=MonitoredResource{type=gce_instance, labels={instance_id=2185756361539269989, project_id=project, zone=us-central1-c}}}, WriteOption{name=LABELS, value={levelName=INFO, levelValue=800}}):
    Logging.write([LogEntry{logName=null, resource=null, timestamp=123456789, receiveTimestamp=null, severity=DEBUG, insertId=null, httpRequest=null, labels={levelName=FINEST, levelValue=300}, operation=null, trace=null, sourceLocation=null, payload=StringPayload{type=STRING, data=message}}], WriteOption{name=LOG_NAME, value=java.log}, WriteOption{name=RESOURCE, value=MonitoredResource{type=global, labels={project_id=project}}}, WriteOption{name=LABELS, value={levelName=INFO, levelValue=800}}): 
```

Adding @pongad since PR https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2420/ is having a similar issue.

pongad:
@shinfan I'm not sure what's causing the error, but I agree that we can be pretty sure it's not this PR.

shinfan:
This is blocking the release. I will go ahead and merge soon if there is no objections. @vam-google

PR #2445, opened 2017-09-15
Add Pub/Sub emulator, synchronous pull snippets

jabubake:


pongad:
This looks OK to me, but Travis seems to not like this change. Could you fix?

jabubake:
@pongad , fixed missing license in source file.

pongad:
Travis Java 7 failed, but this should be independent of this change. LGTM.

PR #2446, opened 2017-09-15
Implement missing features and fix found bugs for BigQuery Client Library

vam-google:
0) Access query results row cells by name. Now `row.get(1)` and `get("firstColumnName")` are supported (currently supported only for results returned by `BigQuery.query()` and not supported by results returned `BigQuery.listTableData()` because there is no schema and it would require additional request to the server to get schema (but still can be added in the future easily with cost of an extra network call)). Same syntax is supported for record-type fields (i.e. nested fields can be accessed by name too).
1) Refactor `BigQuery.query()` to use `jobs.insert/jobs.getQueryResults` combination instead of `jobs.query`. `QueryRequest` class was removed, now `query()` method directly accepts `QueryJobConfiguration` instead. The remaining properties of removed `QueryRequest` class now are passed in a form of `QueryOption` vararg argument. `QueryOption` combines query and waiting options (waiting options are necessary because query() is now waiting for query to complete).
2) Make `BigQuery.query()` call synchronous (waits for query completion before return result), with jittered exponential backoff for polling (based on `gax.retrying` package)
3) Add client side job id generation (`JobId.of()`)
4) Replace `WaitForOption` with `RetryOption` (based on `RetrySettings` and is consistent with the rest of the codebase). This affected compute and spanner clients too.
5) Rewrite `Job.wait()` to use jittered exponentiall backoff polling (based on`gax.retrying` package). Polling is performed differently depending on the type of job: for query jobs on each poll `getQueryResults()` is called, for all other job types `getJob()` is used instead.
6) Use standard SQL as default for all queries
7) Fix wrong query results iteration samples code used all over the place in documentation.
8) Various smaller changes/refactorings

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
@@ -865,16 +865,16 @@ public interface BigQuery extends Service<BigQueryOptions> {
    * <pre> {@code
    * String datasetName = "my_dataset_name";
    * String tableName = "my_table_name";
-   * Page<FieldValues> tableData =
+   * Page<FieldValueList> tableData =
    *     bigquery.listTableData(datasetName, tableName, TableDataListOption.pageSize(100));
-   * for (FieldValues row : tableData.iterateAll()) {
+   * for (FieldValueList row : tableData.iterateAll()) {
    *   // do something with the row
    * }
    * }</pre>
    *
    * @throws BigQueryException upon failure
    */
-  Page<FieldValues> listTableData(String datasetId, String tableId,
+  Page<FieldValueList> listTableData(String datasetId, String tableId,
       TableDataListOption... options);
 
   /**
@@ -885,16 +885,16 @@ public interface BigQuery extends Service<BigQueryOptions> {
    * String datasetName = "my_dataset_name";
    * String tableName = "my_table_name";
    * TableId tableIdObject = TableId.of(datasetName, tableName);
-   * Page<FieldValues> tableData =
+   * Page<FieldValueList> tableData =
    *     bigquery.listTableData(tableIdObject, TableDataListOption.pageSize(100));
-   * for (FieldValues row : rowIterator.hasNext()) {
+   * for (FieldValueList row : rowIterator.hasNext()) {
    *   // do something with the row
    * }
    * }</pre>
    *
    * @throws BigQueryException upon failure
    */
-  Page<FieldValues> listTableData(TableId tableId, TableDataListOption... options);
+  Page<FieldValueList> listTableData(TableId tableId, TableDataListOption... options);
 
   /**
    * Returns the requested job or {@code null} if not found.
@@ -1007,7 +1007,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
    *   // handle errors
    * }
    * QueryResult result = response.getResult();
-   * for (FieldValues row : result.iterateAll()) {
+   * for (FieldValueList row : result.iterateAll()) {
    *   // do something with the data
    * }
    * }</pre>
@@ -1024,7 +1024,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
    *   // handle errors
    * }
    * QueryResult result = response.getResult();
-   * for (FieldValues row : result.iterateAll()) {
+   * for (FieldValueList row : result.iterateAll()) {
    *   // do something with the data
    * }
    * }</pre>
@@ -1078,8 +1078,8 @@ public interface BigQuery extends Service<BigQueryOptions> {
    *   // handle errors
    * }
    * QueryResult result = response.getResult();
-   * Iterator<FieldValues> rowIterator = result.iterateAll();
-   * for (FieldValues row : result.iterateAll()) {
+   * Iterator<FieldValueList> rowIterator = result.iterateAll();
+   * for (FieldValueList row : result.iterateAll()) {
    *   // do something with the data
    * }
    * }</pre>
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -107,7 +107,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
     }
   }
 
-  private static class TableDataPageFetcher implements NextPageFetcher<FieldValues> {
+  private static class TableDataPageFetcher implements NextPageFetcher<FieldValueList> {
 
     private static final long serialVersionUID = -8501991114794410114L;
     private final Map<BigQueryRpc.Option, ?> requestOptions;
@@ -123,13 +123,13 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
     }
 
     @Override
-    public Page<FieldValues> getNextPage() {
+    public Page<FieldValueList> getNextPage() {
       return listTableData(table, serviceOptions, requestOptions);
     }
   }
 
   private static class QueryResultsPageFetcherImpl
-      implements NextPageFetcher<FieldValues>, QueryResult.QueryResultsPageFetcher {
+      implements NextPageFetcher<FieldValueList>, QueryResult.QueryResultsPageFetcher {
 
     private static final long serialVersionUID = -9198905840550459803L;
     private final Map<BigQueryRpc.Option, ?> requestOptions;
@@ -443,17 +443,17 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
   }
 
   @Override
-  public Page<FieldValues> listTableData(String datasetId, String tableId,
+  public Page<FieldValueList> listTableData(String datasetId, String tableId,
       TableDataListOption... options) {
     return listTableData(TableId.of(datasetId, tableId), getOptions(), optionMap(options));
   }
 
   @Override
-  public Page<FieldValues> listTableData(TableId tableId, TableDataListOption... options) {
+  public Page<FieldValueList> listTableData(TableId tableId, TableDataListOption... options) {
     return listTableData(tableId, getOptions(), optionMap(options));
   }
 
-  private static Page<FieldValues> listTableData(final TableId tableId,
+  private static Page<FieldValueList> listTableData(final TableId tableId,
       final BigQueryOptions serviceOptions, final Map<BigQueryRpc.Option, ?> optionsMap) {
     try {
       final TableId completeTableId = tableId.setProjectId(serviceOptions.getProjectId());
@@ -474,15 +474,15 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
     }
   }
 
-  private static List<FieldValues> transformTableData(
+  private static List<FieldValueList> transformTableData(
       Iterable<TableRow> tableDataPb, Schema schema) {
     final FieldList schemaFields = schema != null ? schema.getFields() : null;
     return ImmutableList.copyOf(
         Iterables.transform(tableDataPb != null ? tableDataPb : ImmutableList.<TableRow>of(),
-            new Function<TableRow, FieldValues>() {
+            new Function<TableRow, FieldValueList>() {
               @Override
-              public FieldValues apply(TableRow rowPb) {
-                return FieldValues.fromPb(rowPb.getF(), schemaFields);
+              public FieldValueList apply(TableRow rowPb) {
+                return FieldValueList.fromPb(rowPb.getF(), schemaFields);
               }
             }));
   }
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
@@ -35,7 +35,7 @@ import java.util.Objects;
  * List<BigQueryError> executionErrors = response.getExecutionErrors();
  * // look for errors in executionErrors
  * QueryResult result = response.getResult();
- * for(FieldValues row : result.iterateAll()) {
+ * for(FieldValueList row : result.iterateAll()) {
  *   // do something with row
  * }
  * }</pre>
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
@@ -36,8 +36,7 @@ import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetId;
 import com.google.cloud.bigquery.DatasetInfo;
 import com.google.cloud.bigquery.Field;
-import com.google.cloud.bigquery.FieldValue;
-import com.google.cloud.bigquery.FieldValues;
+import com.google.cloud.bigquery.FieldValueList;
 import com.google.cloud.bigquery.FormatOptions;
 import com.google.cloud.bigquery.InsertAllRequest;
 import com.google.cloud.bigquery.InsertAllResponse;
@@ -440,11 +439,11 @@ public class BigQuerySnippets {
   // [TARGET listTableData(String, String, TableDataListOption...)]
   // [VARIABLE "my_dataset_name"]
   // [VARIABLE "my_table_name"]
-  public Page<FieldValues> listTableData(String datasetName, String tableName) {
+  public Page<FieldValueList> listTableData(String datasetName, String tableName) {
     // [START listTableData]
-    Page<FieldValues> tableData =
+    Page<FieldValueList> tableData =
         bigquery.listTableData(datasetName, tableName, TableDataListOption.pageSize(100));
-    for (FieldValues row : tableData.iterateAll()) {
+    for (FieldValueList row : tableData.iterateAll()) {
       // do something with the row
     }
     // [END listTableData]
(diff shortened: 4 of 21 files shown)


vam-google:
@vkedia please review spanner's `Operation.waitFor()`, as it was rewrittent as part of this PR.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
```diff
@@ -16,6 +16,7 @@
 
 package com.google.cloud.bigquery;
 
+import com.google.cloud.bigquery.BigQuery.QueryOption;
```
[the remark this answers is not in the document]
vam-google: QueryOption is used in documentation

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -440,6 +443,72 @@ public static QueryResultsOption maxWaitTime(long maxWaitTime) {
...
+
+    public QueryResultsOption getQueryResultsOption() {
+      return option instanceof QueryResultsOption ? (QueryResultsOption) option : null;
+    }
+
+    public RetryOption getRetryOption() {
+      return option instanceof RetryOption ? (RetryOption) option : null;
+    }
+
+    public static QueryResultsOption[] filterQueryResutlsOptions(QueryOption... options) {
```
pongad: "Results"?

Why arrays? I'd think lists are more idiomatic?

vam-google: Yes, we should use collections as return types instead of arrays if possible. In this particular case arrays are used because our manual clients use Options in a form of vararg arguments (someMethod(QueryResultsOption... options), so the output of this method is most likely to be passed to a method, which accepts a vararg argument of QueryResultsOption, and we can pass an array as a vararg argument (and it would work as expected). If it was a list we would have to convert it to an array first. So I decided to return array directly to avoid unnecessary conversions.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
```diff
@@ -571,22 +572,18 @@ public boolean cancelJobFromId(String jobName) {
...
     // [START runQuery]
-    QueryRequest request = QueryRequest.of(query);
-    QueryResponse response = bigquery.query(request);
-    // Wait for things to finish
-    while (!response.jobCompleted()) {
-      Thread.sleep(1000);
-      response = bigquery.getQueryResults(response.getJobId());
-    }
+    QueryJobConfiguration queryConfig =
+        QueryJobConfiguration.newBuilder(query).setUseLegacySql(true).build();
```
tswast: Ideally the runQuery example would show a standard SQL sample and we'd have a separate sample for explicitly setting to legacy SQL.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
```diff
@@ -596,25 +593,20 @@ public QueryResponse runQuery(String query) throws InterruptedException {
...
    * Example of running a query with query parameters.
    */
-  // [TARGET query(QueryRequest)]
+  // [TARGET query(QueryJobConfiguration, QueryOption...)]
   // [VARIABLE "SELECT distinct(corpus) FROM `bigquery-public-data.samples.shakespeare` where word_count > @wordCount"]
   public QueryResponse runQueryWithParameters(String query) throws InterruptedException {
     // [START runQueryWithParameters]
-    QueryRequest request = QueryRequest.newBuilder(query)
+    QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query)
         .setUseLegacySql(false) // standard SQL is required to use query parameters
```
tswast: No longer required to explicitly set legacy to false now, correct?

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/InsertDataAndQueryTable.java
```diff
@@ -79,20 +83,16 @@ public static void main(String... args) throws InterruptedException {
     }
 
     // Create a query request
-    QueryRequest queryRequest =
-        QueryRequest.newBuilder("SELECT * FROM my_dataset_id.my_table_id")
-            .setMaxWaitTime(60000L)
-            .setPageSize(1000L)
-            .build();
+    QueryJobConfiguration queryConfig =
+        QueryJobConfiguration.newBuilder("SELECT * FROM my_dataset_id.my_table_id").build();
```
tswast: We should not be encouraging people to [use `select *` with BigQuery](https://stackoverflow.com/a/23991604).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Operation.java
```diff
@@ -120,43 +135,57 @@
...
+            public boolean shouldRetry(
+                Throwable prevThrowable, com.google.longrunning.Operation prevResponse) {
+              if (prevResponse != null) {
+                return !prevResponse.getDone();
+              }
+              if (prevThrowable instanceof BaseServiceException) {
+                if (!(prevThrowable instanceof SpannerException)) {
+                  return ((BaseServiceException) prevThrowable).isRetryable();
+                }
+                return ((SpannerException) prevThrowable).getErrorCode() != ErrorCode.NOT_FOUND;
```
vkedia: I don't think this is correct. We should only retry if SpannerException.isRetryable is true otherwise we should just bubble up the exception.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
pongad: ah, sgtm

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
```diff
@@ -245,16 +98,31 @@ public Builder setName(String name) {
...
-     * Sets the value of the field.
+     * Sets the type of the field.
      *
      * @see <a href="https://cloud.google.com/bigquery/preparing-data-for-bigquery#datatypes">
      *     Data Types</a>
      */
-    public Builder setType(Type type) {
-      this.type = checkNotNull(type);
-      return this;
+    public Builder setType(LegacySQLTypeName type, Field... subFields) {
```
garrettjonesgoogle: It seems weird to require a "Legacy" SQL type name for non-legacy use cases.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Fields.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private final List<Field> fields;
+  private final Map<String, Integer> nameIndex;
+
+  private Fields(Iterable<Field> fields) {
+    this.fields = ImmutableList.copyOf(fields);
+    ImmutableMap.Builder<String, Integer> nameIndexBuilder = ImmutableMap.builder();
+    int index = 0;
+    for (Field field : fields) {
+      nameIndexBuilder.put(field.getName(), index++);
```
garrettjonesgoogle: I much prefer the increment to be on its own line.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/FieldValueTest.java
```diff
@@ -50,75 +50,75 @@
 
   @Test
   public void testFromPb() {
-    FieldValue value = FieldValue.fromPb(BOOLEAN_FIELD);
+    FieldValue value = FieldValue.fromPb(BOOLEAN_FIELD, null);
```
garrettjonesgoogle: If it's common to pass `null` as a second parameter, do we want to create a `fromPb` overload that only requires the first parameter?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -473,19 +468,21 @@ public Rows apply(RowToInsert rowToInsert) {
           }, serviceOptions.getRetrySettings(), EXCEPTION_HANDLER, serviceOptions.getClock());
       String cursor = result.x();
       return new PageImpl<>(new TableDataPageFetcher(tableId, serviceOptions, cursor, optionsMap),
-          cursor, transformTableData(result.y()));
+          cursor, transformTableData(result.y(), null));
```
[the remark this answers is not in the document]
vam-google: Theoretically yes, but the current implementation is based on retrieving the actual schema. With record field types and multiple layers of schema nesting the things get complicated (that is why a real schema is needed).

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
vam-google: I agree, but I believe this should be made as a separate PR. The goal of this is to modify the bigquery client itself (changing snippets only if they are broken).

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
vam-google: Yes. I removed the setter but left and edited the corresponding comment (to emphasize that parameters work only with Std SQL).

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/InsertDataAndQueryTable.java
vam-google: I did not know that. What should be used instead? Explicit list of column names?
I did not create that query, only modified the surrounding code, that is how that line is showed as "added". Leaving it as is.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Operation.java
vam-google: Seems reasonable, but the old code was doing this:

        if (e.getErrorCode() == ErrorCode.NOT_FOUND) {
          return null;
        }
        if (!e.isRetryable()) {
          throw e;
        }

I.e. it was essentially terminating in case if the error code was not found, and this is what is done here.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Field.java
vam-google: 100% agree. Though it is an older problem, just made more visible with this PR. Before, there was a wrapper type, which "encapsulated" legacy type (basically one more wrapper around it, but it did nto have word "Legacy" in it). But the wrapper type was still using the legacy terminology (for example RECORD (legacy) instead of STRUCT (std)).

So I just removed a wrapper which was essentially doing nothing, it was not hiding the legacy nature of the types, since the legacy terminology was used. 

Currently we have LegacySQLTypeName referencing std types, so when you know legacy type, you can easy find a mapped std type. But it doesn't work in the opposite direction.

As we discussed offline switching to StandardSQLTypeName can be done in separate PR

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Fields.java
vam-google: Well, I like using inline incrementing (feels like "true programming" :), but ok, changed.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/FieldValueTest.java
vam-google: So I was thinking about same thing it during making this change. I'll add the overloaded version.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Operation.java
```diff
@@ -120,43 +135,57 @@
...
+          waitSettings,
+          new BasicResultRetryAlgorithm<com.google.longrunning.Operation>() {
+            @Override
+            public boolean shouldRetry(
+                Throwable prevThrowable, com.google.longrunning.Operation prevResponse) {
+              if (prevResponse != null) {
+                return !prevResponse.getDone();
+              }
+              if (prevThrowable instanceof BaseServiceException) {
+                if (!(prevThrowable instanceof SpannerException)) {
```
vkedia: why are you checking for this not to be instance of SpannerException?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Operation.java
vkedia: In the old code, if the exception is SpannerException and is retryable, it would retry. Here if the exception is SpannerException and is not NOT_FOUND it would retry since it returns true in that case.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/InsertDataAndQueryTable.java
tswast: `SELECT my_column FROM my_dataset_id.my_table_id` would be appropriate.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Operation.java
vam-google: Oh, I think the idea was to "support" other retriable exceptions, which theoretically can be added in the future. 

But you are right, because of that I did it wrong, and all spanner exceptions will be retried to oblivion regardless if they are retriable or no. I've almost introduced a a time bomb bug =). Thanks!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/Operation.java
vam-google: Yes, you are right. I reversed the up the order of condition checks, because I thought it would make code shorter. But it just made it wrong =).

vam-google:
Addressed review feedback. PTAL.

@vkedia I fixed the retry logic (sorry for for breaking it). Hopefully now it is correct (matches exactly the original one in functionality). Please note that it theoretically may refuse to retry a retriable BaseServiceException, which is retriable but is not a SpannerException. This is only a theoretical concern, since all BaseServiceExceptions in this context will always be SpanneExceptions too.
Also, please note that with this change all unknown unchecked exceptions will also be wrapped in SpannerException (different from the original code). I think it makes sense here. Please let me know if that should be changed too.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -440,6 +443,72 @@ public static QueryResultsOption maxWaitTime(long maxWaitTime) {
...
```
garrettjonesgoogle: spelling (Resutls)

Also why is it necessary for this to be public?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -939,74 +994,80 @@ public static QueryResultsOption maxWaitTime(long maxWaitTime) {
...
-   * QueryResponse response = bigquery.query(request);
-   * // Wait for things to finish
-   * while (!response.jobCompleted()) {
-   *   Thread.sleep(1000);
-   *   response = bigquery.getQueryResults(response.getJobId());
-   * }
+   * String query = "SELECT distinct(corpus) FROM `bigquery-public-data.samples.shakespeare`";
+   * QueryJobConfiguration queryConfig = QueryJobConfiguration.of(query);
+   *
+   * // To run the legacy syntax queries use the following code instead:
```
garrettjonesgoogle: Also doesn't the query have slightly different syntax too?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValues.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Google BigQuery Table Field Values class, which represents a row in returned query result
+ * (table row). Tables rows can be retrieved as a result of a query or when listing table data.
+ *
+ * Depending on how a corresponding query was executed, each row (an instance of {@code FieldValues})
+ * may or may not contain related schema. If schema is not provided, the individual cells of the row
+ * will still be accessible by index but not by name.
+ */
+public class FieldValues extends AbstractList<FieldValue> implements Serializable {
```
garrettjonesgoogle: It seems like this should be called `FieldValueList`

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Fields.java
```diff
@@ -0,0 +1,118 @@
...
+import java.io.Serializable;
+import java.util.AbstractList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Google BigQuery Table schema fields (columns). Each field has a unique name and index. Fields
+ * with duplicate names are not allowed in BigQuery schema.
+ */
+public final class Fields extends AbstractList<Field> implements Serializable {
```
garrettjonesgoogle: It seems like this should be called `FieldList`

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
vam-google: Fixed the typo. It doesn't have to be public, but I thought it would not hurt to be public. Changed to package-private.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
vam-google: Yes, updated the comment with corresponding legacy syntax.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValues.java
vam-google: I like FieldValues and Fields more, as they are shorter and still clear. But I changed to FieldList and FieldValueList to make clear that those are valid lists.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Fields.java
vam-google: I like FieldValues and Fields more, as they are shorter and still clear. But I changed to FieldList and FieldValueList to make clear that those are valid lists.

garrettjonesgoogle:
LGTM

PR #2449, opened 2017-09-18
bump versions for release

neozwu:


michaelbausor:
LGTM, assuming that we are happy to make these changes part of a point release (I believe that is our current convention, but just wanted to check)

neozwu:
@michaelbausor , after discussing with @vam-google , I pushed another commit to increment minor versions. Do you think this makes sense ?

michaelbausor:
Yes, SGTM

PR #2456, opened 2017-09-20
Change to point to J8 Quickstart

lesv:
Please merge after approval.

### DIFF
--- a/APPENGINE.md
+++ b/APPENGINE.md
@@ -1,37 +1,6 @@
-Use Google Cloud Java Client with Google App Engine Java 8
+Use Google Cloud Java Client with Google App Engine
 =================================================
 
 ** Note: Google App Engine Java 7 is not supported by the google-cloud-java library. **
 
-** Note: the following instructions are not required for App Engine Flex and Java 8 standard environment. **
-
-Quickstart
-----------
-This file provides extra instructions to set up Google Cloud Java Client with Google App Engine.
-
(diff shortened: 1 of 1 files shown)


APPENGINE.md
```diff
@@ -1,4 +1,4 @@
-Use Google Cloud Java Client with Google App Engine
+Use Google Cloud Java Client with Google App Engine Java 8
```
garrettjonesgoogle: I'd prefer not to have the title mention Java 8; that implies that Java 7 instructions are somewhere else. We should probably have the content below state that Java 7 is not supported.

lesv:
PTAL

APPENGINE.md
garrettjonesgoogle: Can we remove the "Java 8" from the title here?

lesv: I was trying to re-enforce this is the way.  I'm not a fan of subtle messages when we are getting issues about this.

garrettjonesgoogle: Actually I just noticed that the title and the 2nd note below conflict... the title says these are instructions for Java 8, and the note says that the instructions aren't required for Java 8.

garrettjonesgoogle:
@lesv do you want to address my most recent feedback so that this PR can be completed?

lesv:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
I feel like this file can be entirely deleted now that it is so stripped down. You would also need to update the main README which currently points to this file.

PR #2460, opened 2017-09-21
increasing response header size, adding instructions for snippets

snehashah16:


vkedia:
Spanner changes look good.
Also in general it is better to separate unrelated changes into different pull requests.

snehashah16:
merge

Issue #2439, opened 2017-09-13
[Storage] Add Requester Pays support to Blob.downloadTo.


frankyn:
Issue:
`Blob.downloadTo` doesn't support requester pays buckets as a blob source.

Noted solution in [PR#2393](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2393):

Add parameter `BlobSourceOption` to `Blob.downloadTo` to support Requester Pays when downloading a Blob to a local file.

vam-google:
Fix merged, closing.

Issue #2392, opened 2017-08-29
Add `.userProject()` method to `Blob.BlobSourceOption` to allow using Requester Pays feature on `Blob.reader()`


jabubake:
Add support in [Blob.BlobSourceOption](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java#L83) similar to [Storage.BlobSourceOption](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L637) to set the billing project ID.
 It will not be able to download blobs using `Blob.reader()` without defining this parameter when Requester Pays is enabled for a bucket. 

**Use case**: downloading a larger blob using the `Blob.reader()`

[Reference API discovery doc](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest).

[Updated by @frankyn]

frankyn:
@vam-google I have updated the description of this issue.

Issue #2391, opened 2017-08-29
Add `RequesterPays` as `BucketField` entry


jabubake:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L95

Currently, if I retrieve by `ID`, I receive null on `requesterPays` even when it has been enabled on a bucket.
Please add `RequesterPays` as `BucketField` entry (unless there is another recommended approach to querying `requesterPays` status of a bucket)

jabubake:
@garrettjonesgoogle : this blocks the samples, unless there is another preferred way to retrieve the `requesterPays` status.

vam-google:
Fix merged, closing.

PR #2464, opened 2017-09-26
Fix requester pays issues: #2392, #2391 and #2439.

vam-google:
Fixes for #2392, #2391 and #2439

To get requesterPays info on a bucket:
```java
Bucket bucket = storage.get(bucketName, Storage.BucketGetOption.fields(BucketField.BILLING));
```

To download a bucket 
```java
blob.downloadTo(destFilePath, Blob.BlobSourceOption.userProject(projectId));
```
or directly use `Blob.reader(BlobSourceOption... options)` method.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketTest.java
@@ -693,7 +693,7 @@ public class BucketTest {
     assertEquals(STORAGE_CLASS, bucket.getStorageClass());
     assertEquals(VERSIONING_ENABLED, bucket.versioningEnabled());
     assertEquals(BUCKET_LABELS, bucket.getLabels());
-    assertEquals(VERSIONING_ENABLED, bucket.requesterPays());
+    assertEquals(REQUESTER_PAYS, bucket.requesterPays());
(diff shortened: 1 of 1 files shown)


jabubake:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketTest.java#L696 seems wrong
Also, given this change, can you update
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/82fa63d01fe19b5158f258f2a2991378e074ab0f/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java#L1507
to use the `BILLING` field and assert that `requesterPays` is `false`.

vam-google:
Fixed the `BucketTest`, left `ITStorageTest` untouched as we discussed. 
Note, `requesterPays()` will still return null if it is set to null on server (i.e. even if we request a bucket with `BucketField.BILLING` option specified, but on server it is set to null, the client will return `null`, not `false`) . In such cases it is essentially has same meaning `false`, but client still returns null to stay consistent and provide accurate information about data returned from server.

vam-google:
PTAL

jabubake:
LGTM

frankyn:
Thanks for picking this up! +1 LGTM.

PR #2470, opened 2017-09-29
pubsub: fix race in messageDeadlineSeconds

pongad:
The variable is concurrently accessed,
so it should be made atomic.


pongad:
cc @davidtorres @mdietz94

PR #2471, opened 2017-09-29
Datastore: addNullBinding and setNullBinding

andreamlin:
Resolves https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1559.

### DIFF
--- a/google-cloud-datastore/src/test/java/com/google/cloud/datastore/SerializationTest.java
+++ b/google-cloud-datastore/src/test/java/com/google/cloud/datastore/SerializationTest.java
@@ -52,12 +52,6 @@ public class SerializationTest extends BaseSerializationTest {
       .addBinding(20)
       .setNamespace("ns1")
       .build();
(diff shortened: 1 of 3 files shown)


google-cloud-datastore/src/test/java/com/google/cloud/datastore/SerializationTest.java
```diff
@@ -52,6 +52,12 @@
       .addBinding(20)
       .setNamespace("ns1")
       .build();
+  private static final Query<Entity> GQL3 =
+      Query.newGqlQueryBuilder(
+          Query.ResultType.ENTITY, "select * from kind1")
```
garrettjonesgoogle: It doesn't seem like this query would make use of the null binding.

garrettjonesgoogle:
@andreamlin do you want to add tests for this PR, or do you want to abandon it?

andreamlin:
I'll add tests!

andreamlin:
Added test.

PR #2478, opened 2017-10-04
pubsub: reverse the padding direction

### LINKED ISSUE #2465: Pub/Sub streamingPull subscriber: large number of duplicate messages, modifyAckDeadline calls observed
Large number of duplicate messages is observed with Pub/Sub streamingPull client library using code that pulls from a Pub/Sub subscription and inserts messages into BigQuery with a synchronous blocking operation, with flowControl set to max 500 outstanding messages.  See [1] for code. 

For the same code, we also observe an excessive number of modifyAckDeadline operations (>> streamingPull message operations).  And tracing a single message, we see modifyAcks and Acks in alternating order for the same message (modAck, modAck, Ack, modAck, Ack) [2]. This suggest that the implementation might fail to remove Ack'ed messages from a queue of messages to process and keep re-processing messages already on the client.  This also suggests that ack requests may not actually be sent.  

[2] https://docs.google.com/spreadsheets/d/1mqtxxm0guZcOcRy8ORG0ri787XLQZNF_FLBujAiayFI/edit?ts=59cac0f0#gid=2139642597

[1] <pre> 
package kir.pubsub;

import com.google.api.gax.batching.FlowControlSettings;
import com.google.cloud.bigquery.*;
import com.google.cloud.pubsub.v1.Subscriber;
import com.google.pubsub.v1.PubsubMessage;
import com.google.pubsub.v1.SubscriptionName;
import com.google.cloud.pubsub.v1.MessageReceiver;
import com.google.cloud.pubsub.v1.AckReplyConsumer;

import java.text.DateFormat;
import java.text.SimpleDateFormat;
import java.util.*;
import java.util.concurrent.atomic.AtomicInteger;

public class Sub {

    // Instantiate an asynchronous message receiver

    public static void main(String... args) throws Exception {
        final String projectId = args[0];
        final String subscriptionId = args[1];

        final BigQuery bq = BigQueryOptions.getDefaultInstance().getService();
        final String datasetName = "pubsub_debug";
        final String tableName = "gke_subscriber";
        final AtomicInteger messageCount = new AtomicInteger(0);
        final DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
        MessageReceiver receiver = new MessageReceiver() {
                    public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
                        // handle incoming message, then ack/nack the received message
                        System.out.printf("%s,\t%s,\t%d\n",message.getData().toStringUtf8()
                                , message.getMessageId()
                                , messageCount.incrementAndGet());
                        Map<String,Object> row = new HashMap<String,Object>();
                        long timestampMs = message.getPublishTime().getSeconds()*1000 + message.getPublishTime().getNanos() / 1000000;
                        Date timestampDate = new Date(timestampMs);

                        row.put("messageId", message.getMessageId());
                        row.put("messageData", message.getData().toStringUtf8());
                        row.put("messagePublishTime", dateFormat.format(timestampDate));
                        // a version of this code without the bq.insert was ran, where consumer.ack() 
                        // was called immediately. The results were the same.
                        InsertAllResponse response = bq.insertAll(
                                    InsertAllRequest.newBuilder(TableId.of(projectId, datasetName, tableName)).addRow(row).build()
                        );
                        if (response.hasErrors()) {
                            System.err.println("Error inserting into BigQuery " + response.toString() );
                            consumer.nack();
                        } else{
                            consumer.ack();
                        }
                    }

                };

        SubscriptionName subscriptionName = SubscriptionName.create(projectId, subscriptionId);
        Subscriber subscriber = Subscriber.defaultBuilder(subscriptionName, receiver).setFlowControlSettings(
                FlowControlSettings.newBuilder().setMaxOutstandingElementCount(1000L).build()
        ).build();
        subscriber.startAsync();
        subscriber.awaitRunning();
        System.out.println("Started async subscriber.");
        subscriber.awaitTerminated();
    }
}
</pre>

pongad:
Previously, padding works by sending deadline extension before
messages expire.
Eg, if message expires in 10 seconds, we try to send extension
in 9 seconds.
A problem is with high latency, highly congested, or highly buffered
connections.
If we want the server to expire messages in 10 seconds,
there is no way to set 10 seconds for more of padding.

This commit reverse the padding direction.
Eg, if we want the message to expire in 10 seconds, we send the
extension at 10 seconds, but we ask the server for 11 seconds.
In this way, if we want the padding to be 10 seconds, we just
ask the server for 20 seconds.

This logic only works for streaming pull though,
as streaming pull allows deadline to be changed individually
for each stream.
Consequently, the tests are changed to reflect the new logic
and the relevant tests are disabled for polling pull.

This commit also increases the default padding to 5 seconds,
since high latency has been observed in production.

Fixes #2465.

### DIFF
--- a/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
+++ b/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
@@ -303,7 +303,6 @@ public class SubscriberTest {
     if (!isStreamingTest) {
       return;
     }
-    System.err.println("testModifyAckDeadline " + isStreamingTest);
     Subscriber subscriber =
         startSubscriber(
             getTestSubscriberBuilder(testReceiver)


pongad:
@davidtorres @mdietz94

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
```diff
@@ -300,6 +300,10 @@ public void testBatchAcksAndNacks() throws Exception {
 
   @Test
   public void testModifyAckDeadline() throws Exception {
+    if (!isStreamingTest) {
+      return;
+    }
+    System.err.println("testModifyAckDeadline " + isStreamingTest);
```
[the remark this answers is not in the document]
pongad: oh whoops

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -278,8 +279,8 @@ public void stop() {
     processOutstandingAckOperations();
   }
 
-  public void setMessageDeadlineSeconds(int messageDeadlineSeconds) {
-    this.messageDeadlineSeconds.set(messageDeadlineSeconds);
+  public void setMessageDeadline(Duration dur) {
```
garrettjonesgoogle: `dur` is a very Go-like variable name :-P

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -318,18 +340,66 @@ public void sendAckOperations(
...
     } catch (Exception e) {
       logger.log(Level.WARNING, "failed to set deadline", e);
     } finally {
       lock.unlock();
     }
   }
+
+  /*
+    We define 3 forms of deadline:
+    1. intendedDeadline: The amount of time we take to process a message.
```
garrettjonesgoogle: maybe `processingDeadline` would be clearer

pongad:
pubsub team is going with a different solution. closing this.

PR #2492, opened 2017-10-09
pubsub: make logs quieter

pongad:
Retryable stream closure happens quite often.
This commit lowers the log level to FINE to not spam the console.


garrettjonesgoogle:
LGTM too

PR #2510, opened 2017-10-12
bigquery: don't return null schema

### LINKED ISSUE #2477: Java client BigQuery.getTable returns a table with null schema when the table has no field
To reproduce,
1. Create an empty BigQuery table without any field on its schema. Let's say it's table "tableA" on dataset "dataset1".
2. Running this code fails the assertion.
```
    BigQuery bigQuery = BigQueryOptions.getDefaultInstance().getService();
    Table table = bigQuery.getTable("dataset1", "tableA");
    Assert.assertNotNull(table.getDefinition().getSchema());
```

Is this expected behavior? Because we in fact can't create a TableDefinition with null schema as it's null-checked in `com.google.cloud.bigquery.TableDefinition.Builder`.
```
    /**
     * Sets the table schema.
     */
    public B setSchema(Schema schema) {
      this.schema = checkNotNull(schema);
      return self();
    }
```
I think `BigQuery.getTable` on a table with empty schema should return a Schema instance with no field within its TableDefinition.

pongad:
Fixes #2477

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableDefinition.java
@@ -147,7 +147,7 @@ public abstract class TableDefinition implements Serializable {
      * Sets the table schema.
      */
     public B setSchema(Schema schema) {
-      this.schema = checkNotNull(schema);
+      this.schema = schema;
       return self();
     }
 


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -1093,8 +1094,11 @@ public void testExtractJob() throws InterruptedException, TimeoutException {
     Job remoteExtractJob = bigquery.create(JobInfo.of(extractConfiguration));
     remoteExtractJob = remoteExtractJob.waitFor();
     assertNull(remoteExtractJob.getStatus().getError());
-    assertEquals(CSV_CONTENT,
-        new String(storage.readAllBytes(BUCKET, EXTRACT_FILE), StandardCharsets.UTF_8));
+
+    // Don't compare strings directly; BigQuery might reorder.
+    String gotCsvContent = new String(storage.readAllBytes(BUCKET, EXTRACT_FILE), StandardCharsets.UTF_8);
+    Truth.assertThat(gotCsvContent.split("\n")).asList().containsExactly(CSV_CONTENT.split("\n"));
```
pongad: The test fails after I made my change but the root cause seems unrelated to the change. Tim, can you verify that export to CSV can reorder rows?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableDefinition.java
```diff
@@ -204,7 +206,7 @@ final boolean baseEquals(TableDefinition tableDefinition) {
 
   Table toPb() {
     Table tablePb = new Table();
-    if (schema != null) {
+    if (schema.getFields().size() != 0) {
```
pongad: Without this change, we'll try to set schema of a view table we want to create, which is not allowed.
This concerns me a bit. Previously we have null to say "we don't know what the schema is" and drop it from the request. 
Perhaps we should let the field be nullable and do a null check on `getSchema`? I'm not sure what the right answer is.

tswast: This check before populating is similar to how it is done in Python, not that that makes it right.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
tswast: I don't think BigQuery gives any guarantees about row order in an export job, but I can check with the backend folks.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableDefinition.java
vam-google: Minor: please change it to `!schema.getFields().isEmpty()`. Here it will be essentially same as `size() != 0`, but technically collections are not required to provide constant time `size()` method and theoretically it can be linear.

pongad: Good catch, done

pongad:
@vam-google For (2), [AutoValue](https://github.com/google/auto/blob/master/value/CHANGES.md#13--14) also rejects `setFoo(null)` so there's so precedent here. The rest of your points are valid though; I don't think I understand BigQuery enough to know what the right answer is. @tswast Do you have an opinion?

tswast:
(3) You're right that allowing nulls is the most correct way to fix this.

The BigQuery API represents a schema as an object (with just one property: fields). This means that there is a difference between a null schema and an empty schema, though I don't know of a case where an empty schema would be returned by the API. I suppose there could be a case in the future where it makes a difference (such as if another property besides fields was added to schema)

pongad:
@tswast From [this issue](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2477) it looks like BigQuery returns null schema if a table is created with no fields. If we allow null in client lib, is this something we can fix on the server side?

tswast:
What would we fix server-side? It's a problem with the way the API is designed so making schema non-nullable server-side would be a breaking change.

pongad:
@tswast Ah yes, sorry. I was being stupid. Shall we fix this by allowing user to set Schema to null then?

tswast:
Yeah, allowing null is the safest thing to do in regards to avoiding problems in the future due to a difference in how schemas are handled in the client versus the API.

It does feel a bit yucky as a user, but better to err on the side of avoiding a disconnect between API representation and client library.

pongad:
Thank you @tswast sounds good. I updated the PR. PTAL

tswast:
LGTM.

PR #2513, opened 2017-10-12
bigquery: add simple benchmarks

pongad:
The benchmark executes a series of queries and report
1. how long until we get the fist result back
2. how long we take to iterate through all results.

### DIFF
--- a/google-cloud-bigquery/src/benchmark/java/com/google/cloud/bigquery/benchmark/Benchmark.java
+++ b/google-cloud-bigquery/src/benchmark/java/com/google/cloud/bigquery/benchmark/Benchmark.java
@@ -30,7 +30,7 @@ import org.threeten.bp.Instant;
 
 public class Benchmark {
 
-  private static final double BILLION = 1000*1000*1000;
+  private static final double NS_PER_SECOND = 1000 * 1000 * 1000;
 
   private Benchmark() {}
 
@@ -71,9 +71,8 @@ public class Benchmark {
       }
       Duration total = Duration.between(start, clock.instant());
 
-      double firstByteSec = (double)(firstByte.getNano())/BILLION + firstByte.getSeconds();
-      double totalSec = (double)(total.getNano())/BILLION + total.getSeconds();
-
+      double firstByteSec = (double) (firstByte.getNano()) / NS_PER_SECOND + firstByte.getSeconds();
+      double totalSec = (double) (total.getNano()) / NS_PER_SECOND + total.getSeconds();
 
       System.out.println(
           String.format(


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/README.md
```diff
@@ -0,0 +1,9 @@
+# BigQuery Benchmark
```
garrettjonesgoogle: Is it maybe more appropriate to put this under src/test or some other directory besides src/main?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/Benchmark.java
```diff
@@ -0,0 +1,79 @@
...
+          firstByte = Duration.between(start, clock.instant());
+        } else if (cols != row.size()) {
+          throw new IllegalStateException(
+              String.format("expected %d cols, found %d", cols, row.size()));
+        }
+      }
+      Duration total = Duration.between(start, clock.instant());
+
+      double firstByteSec = (double)(firstByte.getNano())/1000/1000/1000 + firstByte.getSeconds();
+      double totalSec = (double)(total.getNano())/1000/1000/1000 + total.getSeconds();
```
garrettjonesgoogle: Could we use `Duration` for more readable time calculations?

pongad: We're already using `Duration` right? The lines don't "create new durations", it just turns the existing durations into second form so they can be reported.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/README.md
pongad: Do you have a directory in mind? I was thinking `src/benchmark` but that felt a little odd since it's not a standard Maven directory.

Another thought: create another project `google-cloud-benchmarks`.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/Benchmark.java
garrettjonesgoogle: So why not use `firstByte.getSeconds()` and `total.getSeconds()`?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/README.md
garrettjonesgoogle: It's true that `src/benchmark` is not a standard Maven directory, but it's also not standard to do benchmarks. I think it follows the spirit of Maven though. I'd prefer to keep `google-cloud-*` directories mostly limited to APIs.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/Benchmark.java
pongad: We do call `getSeconds` at the end of the statement. It only reports whole seconds though. If your duration is 12.5s, `getSeconds` will only report the 12. You need to get the 0.5 from `getNano` which is what we're doing here.

garrettjonesgoogle: Oh derp, I didn't look close enough. 

I don't like the error-prone repetition of 1000. Maybe at least make a constant out of the divisor.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/README.md
pongad: Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/benchmark/Benchmark.java
pongad: Done

pongad:
@garrettjonesgoogle PTAL

google-cloud-bigquery/src/benchmark/java/com/google/cloud/bigquery/benchmark/Benchmark.java
```diff
@@ -66,8 +71,8 @@ public static void main(String[] args) throws Exception {
       }
       Duration total = Duration.between(start, clock.instant());
 
-      double firstByteSec = (double)(firstByte.getNano())/1000/1000/1000 + firstByte.getSeconds();
-      double totalSec = (double)(total.getNano())/1000/1000/1000 + total.getSeconds();
+      double firstByteSec = (double)(firstByte.getNano())/BILLION + firstByte.getSeconds();
+      double totalSec = (double)(total.getNano())/BILLION + total.getSeconds();
```
garrettjonesgoogle: "BILLION" isn't terribly helpful - how about `NS_PER_SECOND` or something like that?

pongad:
@garrettjonesgoogle PTAL

PR #2518, opened 2017-10-12
Add retries for insertAll

vam-google:
Retries are allowed only if all insertIds are set (same as it is done in C# client).

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -45,7 +45,6 @@ import com.google.common.collect.Maps;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuery {
 


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -430,16 +433,39 @@ public InsertAllResponse insertAll(InsertAllRequest request) {
...
     requestPb.setRows(rowsPb);
-    return InsertAllResponse.fromPb(
-        bigQueryRpc.insertAll(tableId.getProject(), tableId.getDataset(), tableId.getTable(),
-            requestPb));
+
+    TableDataInsertAllResponse responsePb;
+    if (allInsertIdsSet[0]) {
+      // allowing retries only if all row insertIds are set (used for deduplication)
+      try {
+        responsePb = runWithRetries(
```
garrettjonesgoogle: Can we test the retry?

vam-google: It is very hard to test it in integration tests. In unit tests this behavior is not tested for all other retrieable bigquery methods. The `setUp()` method in `BigQueryImplTest` has this call:
`setRetrySettings(ServiceOptions.getNoRetrySettings())`

garrettjonesgoogle:
LGTM after you address Codacy's complaint

PR #2531, opened 2017-10-17
pubsub: translate exception to ApiException

### LINKED ISSUE #2512: Pub/Sub on failures must return `ApiException` and not `io.grpc.StatusRuntimeException`
Currently, the following code on publishing to a non-existent topic returns `io.grpc.StatusRuntimeException` instead of an `ApiException`.
```
ApiFuture<String> future = publishMessage(publisher, "message");
ApiFutures.addCallback(future, new ApiFutureCallback<String>() {
    @Override
    public void onFailure(Throwable throwable) {
        if (throwable instanceof ApiException) {
         ApiException apiException = (ApiException) throwable;
         System.out.println("Error publishing message : " + apiException.getStatusCode());
       }
   }

   @Override
   public void onSuccess(String s) {
       System.out.println("Message published with ID : " + s);
   }
 });
```

pongad:
Fixes #2512.


jabubake:
LGTM

garrettjonesgoogle:
LGTM

PR #2532, opened 2017-10-17
Firestore: Adding a transaction test with contention

schmidt-sebastian:


google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -424,24 +427,55 @@ public String updateCallback(Transaction transaction) {
...
+        firestore.runTransaction(
+            new Function<String>() {
+              @Override
+              public String updateCallback(Transaction transaction)
+                  throws ExecutionException, InterruptedException {
+                attempts.incrementAndGet();
+
+                DocumentSnapshot documentSnapshot = transaction.get(documentReference).get();
+                firstLock.acquire();
+                secondLock.release();
```
pongad: nit: can we release before acquire?

My reasoning is if both transactions release before acquire, the synchronization should still work, but if both acquire before release, we'll deadlock. I think release-before-acquire makes it easier to prove deadlock-freedom in our heads.

schmidt-sebastian: Updated this to use the much more suitable CountDownLatch.

pongad:
LGTM. I didn't know `CountDownLatch` works OK if current count is zero!

PR #2547, opened 2017-10-23
bigquery: prepare to use tabledata.list

pongad:
Benchmark shows that tabledata.list is significantly
faster than getQueryResults. This commit begins the work
to migrate to the former.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
@@ -57,6 +57,7 @@ public class QueryResponse implements Serializable {
   private final boolean cacheHit;
   private final Schema schema;
   private final long totalBytesProcessed;
+  private final long totalRows;
 
   static final class Builder {
 
@@ -69,6 +70,7 @@ public class QueryResponse implements Serializable {
     private boolean cacheHit;
     private Schema schema;
     private long totalBytesProcessed;
+    private long totalRows;
 
     private Builder() {}
 
@@ -117,6 +119,11 @@ public class QueryResponse implements Serializable {
       return this;
     }
 
+    Builder setTotalRows(long totalRows) {
+      this.totalRows = totalRows;
+      return this;
+    }
+
     QueryResponse build() {
       return new QueryResponse(this);
     }
@@ -135,6 +142,7 @@ public class QueryResponse implements Serializable {
     this.cacheHit = builder.cacheHit;
     this.schema = builder.schema;
     this.totalBytesProcessed = builder.totalBytesProcessed;
+    this.totalRows = builder.totalRows;
   }
 
   /**
@@ -145,8 +153,8 @@ public class QueryResponse implements Serializable {
     if (!jobCompleted()) {
       return null;
     }
+    // return bigquery.listTableData
     return null;
-    // return result;
   }
 
   /** Returns the hash of the {@code QueryResponse} resource or {@code null} if not set. */
@@ -221,6 +229,15 @@ public class QueryResponse implements Serializable {
     return totalBytesProcessed;
   }
 
+  /**
+   * Returns the total number of rows in the complete query result set, which can be more than the
+   * number of rows in the first page of results returned by {@link #getValues()}. Returns {@code 0}
+   * if the query was a dry run.
+   */
+  public long getTotalRows() {
+    return totalRows;
+  }
+
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
@@ -232,6 +249,7 @@ public class QueryResponse implements Serializable {
         .add("cacheHit", cacheHit)
         .add("schema", schema)
         .add("totalBytesProcessed", totalBytesProcessed)
+        .add("totalRows", totalRows)
         .toString();
   }
 
@@ -245,7 +263,7 @@ public class QueryResponse implements Serializable {
         executionErrors,
         cacheHit,
         schema,
-        totalBytesProcessed);
+        totalBytesProcessed, totalRows);
   }
 
   @Override
@@ -264,6 +282,7 @@ public class QueryResponse implements Serializable {
         && Objects.equals(executionErrors, response.executionErrors)
         && Objects.equals(schema, response.schema)
         && totalBytesProcessed == response.totalBytesProcessed
+        && totalRows == response.totalRows
         && cacheHit == response.cacheHit;
   }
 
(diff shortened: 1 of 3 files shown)


pongad:
Reviewers:
I decided to put this work in a branch. The changes are likely to be substantial.

The idea is that we'll use `getQueryResults` to wait for the job to finish (`QueryResponse`), then use `tabledata.list` to actually read the results (`QueryResult`).

I wanted to make this PR early to make sure we agree on general approach; please focus on the general design for now.

Some of my concerns:
- It seems funny that `getQueryResults` method doesn't actually call `getQueryResults` RPC (the method has a public overload). I'm not sure this actually matters; does anyone think differently?
- This change materially changes `QueryResponse` even though the API surface looks about the same. In particular, `getResults` could now throw.
- Right now QueryResult has [a few fields](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/72ae1ffadc2a0c056177ff3c0cf563372b94a186/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResult.java#L32) populated by data available from `getQueryResults` but not `tabledata.list`. Can these fields be removed or hoisted into `QueryResponse`?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
```diff
@@ -117,9 +123,32 @@ private QueryResponse(Builder builder) {
...
+      }
+      throw new RuntimeException(resultException);
+    } finally {
+      resultLock.unlock();
+    }
+    // return result;
   }
 
+  private QueryResult loadResult() {
+    throw new UnsupportedOperationException();
```
pongad: call `tabledata.list` here.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
```diff
@@ -117,9 +123,32 @@ private QueryResponse(Builder builder) {
    * false}.
    */
   public QueryResult getResult() {
```
tswast: Is `QueryResult` still the most appropriate type for this?

tswast:
- `totalBytesProcessed`, `cacheHit`, and `schema` are available in the job resource, so it needn't be hoisted from the `getQueryResults` response.
- `totalRows` is available in `tabledata.list` response.

I believe the `QueryResults` class may be unnecessary. Could the returned type be the same as that returned by whichever function is exposed for listing rows directly from a table?

pongad:
Thank you for your thoughts @tswast , PTAL.

In general, we need to wait for the query to finish, so we need to call `getQueryResult` RPC anyway right? IIUC, a better opportunity is to not call `getJob` when we already know properties of the job, eg if we recently inserted the job.

The tests are a little messy: details will probably change soon so I didn't want to spend too much time cleaning up the transient state. If you think we're on the right track, let's submit? We can sweat the details in follow up CLs.

tswast:
I would still call `getJob` since it will have always have more detailed statistics than those exposed by `getQueryResults`.

The only reason we are calling `getQueryResults` is for the behavior where it hangs until timeout or completion. The compromise I made in Python was to only call `getJob` after `getQueryResults` says the job is complete.

Take care to make sure any timeout values are propagated correctly to `getQueryResults` as I had to do in https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4209.

tswast:
If we are going to keep the current interface of having the user call `getQueryResults` (the Java function) to wait for a query to finish, this change LGTM.

pongad:
@tswast Sorry I'm having a little difficulty parsing your last response. By "current interface", do you mean the slightly-changed interface proposed by this PR?

This PR actually changes `getQueryResults` (Java function); it's just that `query` calls `getQueryResults` (Java function) after inserting the job.

Sorry for the numerous round trips.

tswast:
Yes, I meant "if we are trying to keep the interface pretty close to the same". It seems to me that the `QueryResponse` object in this Java library corresponds to information about a query job resource from the BigQuery API, so it makes sense to me to move the properties up to this class.

In fact, I'm a little surprised that there aren't more properties on QueryResponse already, since there are many properties relevant to query jobs that I would expect to find there. For example, after you create a query job, if you don't specify a [destination table](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.destinationTable), one will come back when you get the job. This will be one property that you need to do the tabledata.list.

vam-google:
BigQuery has plans to migrate `getQueryResults` on same platform as` tabledata.list` is so it should perform closer to `tabledata.list`. Initially using `tabledata.list` instead of `getQueryResults` was one of the P0 items in bigquery Beta2 preparation. It was removed from requirements later because of the plans to migrate `getQueryResults` so it will be able to achieve similar performance to tabledata.list.

I understand that `tabledata.list` is faster (as expected), but our client lib is already doing stuff, which I wish it never did: it **fakes** actual calls. For example `BigQuery.query()` is not `query` api call anymore, it is now a wrapper around job insert and job status polling. `Job.waitFor()` is not getting job status anymore for all jobs, but for query jobs it actually calls `getQueryResults` instead when polls. 
There are other things like this, and `getQueryResults -> tabledata.list` will be one more "fraudulent call" like the ones above.

These things are additional levels of complication in the client library (and honestly speaking just dirt in our libraries). I would really like to not add more complications and "fradulent calls" to BigQuery client unless it is absolutely necessary. We need to confirm with BigQuery team, but if they really going to migrate `getQueryResults` (i.e. improve its performance) in any foreseeable future, I would prefer to avoid substituting it with `tabledata.list`.

tswast:
Even if we don't make `getQueryResults -> tabledata.list`, I would like `query()` to return something that is more clearly a query job. I like that this PR adds some of those properties to QueryResponse, but maybe the better decision is to make query() return a different type than QueryResponse

pongad:
@vam-google I understand the concern. Since Java is not *that* slow and should soon get faster, I'm OK with not migrating to `tabledata.list` at least for now. Still, I believe we should do some variation of the surface changes this PR introduces so that we have the option to migrate if we need to later.

This seems to be what @tswast wants as well. Tim, do you have a proposal in mind?

tswast:
Honestly, in the current state, I'm going to have a hard time recommending the `query()` method in docs, even though it is begging to be called when browsing the code.

My ideal interface would have
- `BigQuery.query()` return a Job.
- `Job.getQueryResults()` return an iterator over rows.

The problem with this is it does not address the requirement for "One method that takes SQL text and options, runs the query, waits for completion, and returns the result rows".

In my opinion, the high level method needs to make it easy to cancel the query job if something goes wrong or times out. The way this is done in PHP is to include the job information in the thrown exception. The way we did it in Python is to catch the exception and try to cancel.

pongad:
As discussed, we're going with slightly different route. I'll resend PR.

PR #2554, opened 2017-10-26
Requester pays: expose userProject

andreamlin:
Expose the userProject parameter for the List.Buckets operation.
Bump google-api-client dependency to 1.23.0.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -252,7 +252,7 @@ public class HttpStorageRpc implements StorageRpc {
           .setMaxResults(Option.MAX_RESULTS.getLong(options))
           .setPageToken(Option.PAGE_TOKEN.getString(options))
           .setFields(Option.FIELDS.getString(options))
-          .setUserProject(Option.FIELDS.getString(options))
+          .setUserProject(Option.USER_PROJECT.getString(options))
           .execute();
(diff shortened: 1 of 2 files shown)


andreamlin:
Not ready for review; just running CI

andreamlin:
PTAL

andreamlin:
Hmm bumping the google-api-* dependency to 0.23.1 seems to be an issue (#2453). Looking into that.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -252,6 +252,7 @@ public StorageObject create(StorageObject storageObject, final InputStream conte
           .setMaxResults(Option.MAX_RESULTS.getLong(options))
           .setPageToken(Option.PAGE_TOKEN.getString(options))
           .setFields(Option.FIELDS.getString(options))
+          .setUserProject(Option.FIELDS.getString(options))
```
vam-google: Should this use called `Option.FIELDS` (it is suspicious, because it looks like a copied value from the previous line)

garrettjonesgoogle:
@vam-google updating the underlying http and apiary clients should be fine - they don't form part of the surface of google-cloud-storage.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
andreamlin: Wow that was a glaring error! Thanks for catching that! The unit test is updated too.

andreamlin:
Fixed bugs. PTAL

PR #2557, opened 2017-10-26
Refreshing generated clients

garrettjonesgoogle:
Breaking changes:

- In Firestore, DatabaseName is renamed to DatabaseRootName
- in Video Intelligence, annotateVideoAsync has parameters removed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,7 +142,7 @@
     <api-common.version>1.1.0</api-common.version>
     <gax.version>1.9.1</gax.version>
     <gax-grpc.version>0.26.1</gax-grpc.version>
-    <generatedProto.version>0.1.21</generatedProto.version>
+    <generatedProto.version>0.1.22</generatedProto.version>
     <google.auth.version>0.8.0</google.auth.version>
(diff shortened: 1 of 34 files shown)


pom.xml
```diff
@@ -731,11 +731,6 @@
       </dependency>
       <dependency>
         <groupId>com.google.api.grpc</groupId>
-        <artifactId>grpc-google-common-protos</artifactId>
```
pongad: I don't remember what this is for, but are we sure we can delete this? Datastore's pom.xml declares dependency on it but nowhere else.

garrettjonesgoogle: I checked the contents, and it is *identical* to grpc-google-longrunning-v1. Thus, we aren't losing anything at all. 

Datastore declares a dependency on it because that was the original package name we used. Later we split out proto-google-common-protos, which contained the protos that datastore *actually* needed. Datastore doesn't need the long-running Grpc stub.

garrettjonesgoogle:
I was just trying to get at least one review, which I have now :-)

PR #2564, opened 2017-10-30
Regenerating clients (breaking changes)

garrettjonesgoogle:
This is a large regeneration to clean up the surface of generated clients. 

Corresponds to https://github.com/googleapis/toolkit/pull/1659 and depends on https://github.com/googleapis/gax-java/pull/396 being released.

### DIFF
--- a/pom.xml~
+++ b/pom.xml~
@@ -140,8 +140,8 @@
     <api-client.version>1.22.0</api-client.version>
 
     <api-common.version>1.1.0</api-common.version>
-    <gax.version>1.11.0</gax.version>
-    <gax-grpc.version>0.28.0</gax-grpc.version>
+    <gax.version>1.10.0</gax.version>
+    <gax-grpc.version>0.27.0</gax-grpc.version>
     <generatedProto.version>0.1.22</generatedProto.version>
     <google.auth.version>0.8.0</google.auth.version>
     <grpc.version>1.7.0</grpc.version>


google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/BaseGrpcServiceException.java
```diff
@@ -72,8 +72,8 @@ public BaseGrpcServiceException(ApiException apiException) {
         .setMessage(apiException.getMessage())
         .setCause(apiException)
         .setRetryable(apiException.isRetryable())
-        .setCode(((GrpcStatusCode) apiException.getStatusCode()).getCode().value())
-        .setReason(((GrpcStatusCode) apiException.getStatusCode()).getCode().name())
+        .setCode(apiException.getStatusCode().getCode().getHttpStatusCode())
```
pongad: I might be missing something. Why is this HTTP code? This is for gRPC right?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -168,11 +171,18 @@ public void close() throws IOException {
...
+    }
+    if (channelProvider.needsHeaders()) {
+      channelProvider = channelProvider.withHeaders(builder.headerProvider.getHeaders());
+    }
+    this.channelProvider = channelProvider;
     credentialsProvider = builder.credentialsProvider;
 
     numChannels = builder.parallelPullCount;
-    channels = new ArrayList<ManagedChannel>(numChannels);
+    channels = new ArrayList(numChannels);
```
pongad: `new ArrayList<>(numChannels)`

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/PublisherImplTest.java
```diff
@@ -16,40 +16,39 @@
...
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 import org.threeten.bp.Duration;
 
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
+
+import static org.junit.Assert.*;
```
pongad: Which formatting are we using for this repo? This doesn't conform to Google style guide; the style guide doesn't allow spaces between imports.

google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/BaseGrpcServiceException.java
garrettjonesgoogle: Yes, but the base class `BaseServiceException` expects an integer code. It was previously using the proto field number for grpc exceptions, but I don't think that was useful to anyone.

google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/PublisherImplTest.java
garrettjonesgoogle: Ack that was an IDE problem, I'll fix it.

pongad:
This LGTM, but I think at least one more person should approve.

PR #2565, opened 2017-10-30
Omitting WriteResults for DocumentTransforms

schmidt-sebastian:
We have to split up a write that contains a ServerTimestamp into two distinct writes. Since the user is unaware of this, we should only return one WriteResult in our response.

This is only relevant for WriteBatches and a port of https://github.com/googleapis/nodejs-firestore/pull/45

I also ran google-java-format on the codebase.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
@@ -459,7 +459,7 @@ abstract class UpdateBuilder<T extends UpdateBuilder> {
 
   /** Commit the current batch. */
   ApiFuture<List<WriteResult>> commit(@Nullable ByteString transactionId) {
-    // We create our own copy of this array since we need to access it when processing the response.
+    // We create our own copy of this list since we need to access it when processing the response.
     final ImmutableList<Write> writeRequests = ImmutableList.copyOf(this.writes);
 
     CommitRequest.Builder request = CommitRequest.newBuilder();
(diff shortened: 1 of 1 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -484,7 +484,7 @@ public String updateCallback(Transaction transaction) {
   @Test
   public void deleteDocument() throws Exception {
     doReturn(beginResponse())
-        .doReturn(commitResponse(0, 0))
+        .doReturn(commitResponse(2, 0))
```
pongad: Forgive my ignorance, I'm not sure why this is changing.

schmidt-sebastian:
> To make sure I understand it right:
> If the user makes N writes, we could previously return more than N responses because some writes are split. Now we always return N responses assuming everything succeeds?

The server always returns n write responses for n write requests. The problem is that there is no 1 to 1 mapping between user requests and write requests (some user requests might turn into two writes). This PR removes the extra writes, so that the user can correlate its requests to the responses.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
schmidt-sebastian: It's required because of the newly added Precondition: 
https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2565/files#diff-837e4dce39e3a89ffb585a67cddc45cbR486

This test deletes two documents.

pongad:
I see. That makes sense.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
pongad: got it, thank you

schmidt-sebastian:
Garrett, please let me know offline if there are any concerns.

PR #2566, opened 2017-10-31
bigquery: make query return a Job

pongad:
Eventually we'll delete QueryResponse completely so that we can
use the faster listTableData RPC.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
@@ -27,7 +27,7 @@ import java.util.Objects;
 /**
  * Google Cloud BigQuery Query Response. This class contains the results of a Query Job
  * ({@link BigQuery#getQueryResults(JobId, BigQuery.QueryResultsOption...)}) or of a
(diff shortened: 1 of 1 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -584,24 +583,15 @@ public Boolean call() {
...
-      JobInfo jobInfo = JobInfo.newBuilder(configuration).setJobId(jobId).build();
-      Job job = create(jobInfo);
-      RetrySettings waitSettings =
-          RetryOption.mergeToSettings(
-              Job.DEFAULT_QUERY_JOB_WAIT_SETTINGS, QueryOption.filterRetryOptions(options));
-      return job.waitForQueryResults(waitSettings, QueryOption.filterQueryResultsOptions(options));
-    } catch (RetryHelper.RetryHelperException e) {
-      throw BigQueryException.translateAndThrow(e);
-    }
+    return create(JobInfo.of(jobId, configuration), options);
```
vam-google: Please make query synchronous (wait for job completion  before returning).

PR #2572, opened 2017-11-02
pubsub: create experimental message dispatcher

pongad:
EXPERIMENTAL
I pulled it into a separate class so it may coexist with the original.
We can submit this, but it needs more testing before production use.

We emphasize simplicity, not optimal performance.
Of course, performance is still important,
hence the many concurrent data structures used here.

Major differences from the original and some justifications:
- It modacks immediately after receiving messages
  to take advantage of server-side duplicate-mitigation
  feature that should soon be available.
- In the original impl, we create many dispatchers.
  They share some data structures and keep others separate.
  I find this too confusing.
  The new impl creates only one dispatcher, to be shared
  among all connections. This obviously increases contention
  but should be at least partially alleviated by some lock-free
  algorithms used here.
- It makes deadline a constant of 1 minute.
  With the dup-mitigation feature, the server might need to
  wait on the order of minutes before it could redeliver messages.
  I opine that setting the deadline to 1 minute shouldn't drastically
  worsen the redelivery latency.
  Also unlike the original, it does not periodically adjust deadline.
  I have some ideas on how this could be simply implemented;
  we can add this feature back if necessary.
- Modack time is also set to 1 minute and doesn't exponentially back
  off. Since the deadline is already 1 minute, it seems silly to
  bicker over a few extra seconds. [1]
- Modacks run on fixed schedule, giving 15 seconds padding,
  and modacks all pending messages, not just the ones about to expire.
  While clearly suboptimal, it's not very expensive since it only
  happens once every 45 seconds.

[1] This caused a bug. If the padding is set too large, we'd
schedule modacks to occur in the past, creating a modack storm.

I believe the benefits of reduced complexity outweighs the cost.
Load test shows the current implementation still has no trouble
catching up with the publisher.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
@@ -36,6 +36,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import com.google.auto.value.AutoValue;
 
 class MessageDispatcher2 {
 
@@ -51,7 +52,7 @@ class MessageDispatcher2 {
 
   private final ConcurrentLinkedQueue<WorkItem> workQueue = new ConcurrentLinkedQueue<>();
   private final ConcurrentLinkedQueue<String> idsToAck = new ConcurrentLinkedQueue<>();
-  private final ConcurrentLinkedQueue<String> idsToNack = new ConcurrentLinkedQueue<>();
+  private final ConcurrentLinkedQueue<ModAckItem> modAcks = new ConcurrentLinkedQueue<>();
 
   // Boolean, not Void, because the map doesn't allow null values.
   private final ConcurrentHashMap<String, Boolean> extensionSet = new ConcurrentHashMap<>();
@@ -60,9 +61,19 @@ class MessageDispatcher2 {
 
   private ScheduledFuture<?> ackNackJob;
   private ScheduledFuture<?> extensionJob;
-  
+
   interface Connection {
     void send(StreamingPullRequest request);
+  }
+
+@AutoValue
+  static abstract class ModAckItem {
+    abstract String ackId();
+    abstract int seconds();
+
+    static ModAckItem create(String ackId, int seconds) {
+      return new AutoValue_MessageDispatcher2_ModAckItem(ackId, seconds);
+    }
   }
 
   private static class WorkItem {
@@ -108,7 +119,7 @@ class MessageDispatcher2 {
     @Override
     public void nack() {
       complete("nack");
-      idsToNack.add(workItem.message.getAckId());
+      modAcks.add(ModAckItem.create(workItem.message.getAckId(), 0));
     }
 
     void throwException(Throwable t) {
@@ -118,7 +129,7 @@ class MessageDispatcher2 {
           Level.WARNING,
           "MessageReceiver failed to processes ack ID: " + ackId + ", the message will be nacked.",
           t);
-      idsToNack.add(ackId);
+          modAcks.add(ModAckItem.create(ackId, 0));
     }
   }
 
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
```diff
@@ -0,0 +1,278 @@
...
+
+    AtomicInteger remaining = new AtomicInteger(messages.size());
+    for (ReceivedMessage message : messages) {
+      workQueue.add(new WorkItem(message, remaining, doneCallback));
+      extensionSet.put(message.getAckId(), true);
+    }
+    workMessages();
+  }
+
+  private synchronized void workMessages() {
```
pongad: This being synchronized is a little worrisome, since this is a hot path. Since the load test shows no problem, I think we should leave this as-is. I can think of a somewhat-lock-free way to implement this, but it's rather complicated. We can do that if this turns out to be a bottleneck.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  void processReceivedMessages(List<ReceivedMessage> messages, Runnable doneCallback) {
+    if (messages.isEmpty()) {
+      doneCallback.run();
+      return;
+    }
+
+    for (ReceivedMessage message : messages) {
+      modAcks.add(ModAckItem.create(message.getAckId(), DEADLINE_EXTENSION_SEC));
```
pongad: if we process messages fast, we could end up both modacking and acking the same message in one request. I don't think that's a problem right?

mdietz94: That's fine

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
```diff
@@ -0,0 +1,284 @@
...
+    while (builder.getModifyDeadlineAckIdsCount() < MAX_CHANGE_PER_REQUEST) {
+      ModAckItem modAck = modAcks.poll();
+      if (modAck == null) {
+        break;
+      }
+      builder.addModifyDeadlineAckIds(modAck.ackId());
+      builder.addModifyDeadlineSeconds(modAck.seconds());
+    }
+
+    return builder.getAckIdsCount() == MAX_CHANGE_PER_REQUEST
```
mdietz94: I would sum these, since they are effectively equally expensive. Better to do the sum relative to the max change, plus that way you can fill the requests with more mod acks if there are few acks.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
```diff
@@ -0,0 +1,284 @@
...
+              }
+            }
+          });
+    }
+  }
+
+  private void sendRequest(StreamingPullRequest request) {
+    Connection connection = null;
+    try {
+      connection = connections.take();
```
mdietz94: This logic is confusing, it's important we send acks and mod acks through the same connections that received the messages.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
pongad: From David's implementation, I thought that pubsub service requires that we send at most 1000 acks and at most 1000 modacks. Is this not the case? What determines the max request size?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher2.java
pongad: I thought I was being smart round-robin-ing the connections :(

Could you provide some context here? The current logic seems to work fine on both desktop and loadtest. Might it not work in the future?

mdietz94:
It's unfortunately much more expensive server side to process messages from
a different stream. We do support it so users can continue to process
messages after a stream dies, but we won't be able to scale if every stream
is doing this.

pongad:
superseded by #2580

PR #2574, opened 2017-11-03
Update google-cloud-datastore protos + client

wsh:
google-cloud-datastore v1.5.1 includes support for [ID reservation](https://cloud.google.com/datastore/docs/reference/rest/v1/projects/reserveIds) and a fix for an [HTTP connection handling issue](https://github.com/GoogleCloudPlatform/google-cloud-datastore/issues/185).


garrettjonesgoogle:
Looks like there was a build problem: "Failed to collect dependencies at com.google.cloud.datastore:datastore-v1-proto-client:jar:1.5.1"

PR #2578, opened 2017-11-05
Refreshing auto-generated code for create() -> of() rename

garrettjonesgoogle:
This depends on the proto-* package updates in https://github.com/googleapis/api-client-staging/pull/432 being released.


pongad:
LGTM. Of course, wait for the packages update first :)

PR #2583, opened 2017-11-06
Regenerating auto-gen clients - builder parameterization

garrettjonesgoogle:
Depends on https://github.com/googleapis/gax-java/pull/414

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/v2/MetricsSettings.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/v2/MetricsSettings.java
@@ -181,7 +181,7 @@ public class MetricsSettings extends ClientSettings<MetricsSettings> {
     return defaultGrpcTransportProviderBuilder().build();
   }
 
-  @BetaApi("The surface for customizing headers is not stable yet and my change in the future.")
+  @BetaApi("The surface for customizing headers is not stable yet and may change in the future.")
   public static ApiClientHeaderProvider.Builder defaultApiClientHeaderProviderBuilder() {
     return ApiClientHeaderProvider.newBuilder()
         .setGeneratorHeader(DEFAULT_GAPIC_NAME, getGapicVersion())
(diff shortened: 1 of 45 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/v2/MetricsSettings.java
```diff
@@ -176,13 +174,14 @@ public static String getDefaultEndpoint() {
...
   public static InstantiatingGrpcChannelProvider.Builder defaultGrpcTransportProviderBuilder() {
-    return InstantiatingGrpcChannelProvider.newBuilder().setEndpoint(getDefaultEndpoint());
+    return InstantiatingGrpcChannelProvider.newBuilder();
   }
 
   public static TransportChannelProvider defaultTransportChannelProvider() {
     return defaultGrpcTransportProviderBuilder().build();
   }
 
+  @BetaApi("The surface for customizing headers is not stable yet and my change in the future.")
```
michaelbausor: s/my/may here and elsewhere

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -141,6 +141,9 @@ private Publisher(Builder builder) throws IOException {
     if (channelProvider.needsHeaders()) {
       channelProvider = channelProvider.withHeaders(builder.headerProvider.getHeaders());
     }
+    if (channelProvider.needsEndpoint()) {
```
michaelbausor: From a review perspective, I think it would be good to put the manual changes in a separate PR, although then it may be impossible to get tests to pass?

google-cloud-logging/src/main/java/com/google/cloud/logging/v2/MetricsSettings.java
garrettjonesgoogle: This is just embarrassing at this point

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
garrettjonesgoogle: Correct.

Maybe one process we can follow in the future is to include two commits in the PR - one with only autogen changes, and the next with the manual fixups.

PR #2585, opened 2017-11-07
pubsub: add keepalive

### LINKED ISSUE #2506: Pub/Sub StreamingSubscriberConnection  : Terminated streaming with exception
See related issue on the spring-cloud-gcp project at https://github.com/spring-guides/gs-messaging-gcp-pubsub/issues/4.

pongad:
If streaming pull connection is idle for a long time,
a firewall/proxy might terminate the connection.
The keepalive should prevent this from happening,
or at least let us notice that the connection is terminated so we don't
wait forever.

Fixes #2506


pongad:
We seem to be getting some deadlock in CI. Analysis and experience suggest these flakes are caused by the test themselves; the libraries aren't doing anything fundamentally wrong.

I'm planning to simply pubsub implementation and tests in upcoming weeks which, I hope, will make these flakes go away. I think we can merge this for now.

PR #2593, opened 2017-11-08
bigquery: Job.getQueryResults returns QueryResult

pongad:
This is the first step in removing QueryResponse.

This PR first creates QueryResult then retrieves the Response.
Eventually, we'll directly create Result from tabledata.list RPC.

### DIFF
--- a/google-cloud-examples/src/test/java/com/google/cloud/examples/bigquery/snippets/ITBigQuerySnippets.java
+++ b/google-cloud-examples/src/test/java/com/google/cloud/examples/bigquery/snippets/ITBigQuerySnippets.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import com.google.api.gax.paging.Page;
@@ -37,6 +38,7 @@ import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.QueryResponse;
 import com.google.cloud.bigquery.QueryResult;
 import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.JobStatus;
 import com.google.cloud.bigquery.StandardTableDefinition;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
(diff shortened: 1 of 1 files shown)


michaelbausor:
This LGTM, although I don't have complete context so others should also approve

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -263,13 +263,20 @@ public Job waitFor(RetryOption... waitOptions) throws InterruptedException {
...
-  public QueryResponse getQueryResults(QueryResultsOption... options) {
+  public QueryResult getQueryResults(QueryResultsOption... options) {
     if (getConfiguration().getType() != Type.QUERY) {
       throw new UnsupportedOperationException(
           "Getting query results is supported only for " + Type.QUERY + " jobs");
     }
 
-    return bigquery.getQueryResults(getJobId(), options);
+    QueryResponse response = bigquery.getQueryResults(getJobId(), options);
+    if (!response.jobCompleted()) {
```
tswast: We could check that the [job status.state](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#status.state) is DONE rather than looking at the `getQueryResults` response.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -263,13 +263,20 @@ public Job waitFor(RetryOption... waitOptions) throws InterruptedException {
...
+      throw new IllegalStateException("the job hasn't completed yet");
+    }
+    if (response.hasErrors()) {
```
tswast: Ditto that we could check on the errors at [job status.errors](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#status.errors).

There is a chance that any errors populated here are just warnings according to the docs

> Errors here do not necessarily mean that the job has completed or was unsuccessful.

This means I think maybe we shouldn't be throwing an exception in all cases. Let's check with the BQ eng team.

pongad: I checked the Go implementation. You're right. Removed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
pongad: Added TODO. I have a little question/concern about this. I'll note this in a followup PR

pongad:
Forgot to edit samples :) I'll look into this tomorrow

PR #2599, opened 2017-11-08
getPath() should be relative to the database root

schmidt-sebastian:
This is technically a breaking API change, but getPath() is spec'ed to return the relative path to the database root and not the fully-qualified name.


google-cloud-firestore/src/main/java/com/google/cloud/firestore/ResourcePath.java
```diff
@@ -98,26 +98,46 @@ String getId() {
...
-  public String toString() {
+  String getPath() {
     StringBuilder result = new StringBuilder();
-    result.append(getDatabaseName());
 
-    ImmutableList<String> parts = getSegments();
-    if (parts.size() > 0) {
-      result.append("/documents");
-      for (String part : parts) {
+    boolean first = true;
```
pongad: Isn't this `String.join("/", getSegments())`?

schmidt-sebastian: It is, but using the Java 7 API. Did we move to Java 8?

pongad: Oh duh, sorry.

PR #2601, opened 2017-11-08
Bumping language to GA, 4 others to Beta

garrettjonesgoogle:


google-cloud-language/pom.xml
```diff
@@ -2,7 +2,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <artifactId>google-cloud-language</artifactId>
-  <version>0.27.1-beta-SNAPSHOT</version>
+  <version>1.9.1-SNAPSHOT</version>
```
michaelbausor: Where is the 1.9.1 coming from (instead of 1.0.0?)

garrettjonesgoogle: We are using coordinated versions among GA clients as a set - it makes it easier for users to get versions right.

PR #2606, opened 2017-11-09
Adds support for attaching labels to cloud spanner sessions

vkedia:


vkedia:
cc @tomayles

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -153,6 +159,14 @@ public Builder setSessionPoolOption(SessionPoolOptions sessionPoolOptions) {
     }
 
     /**
+     * Sets the labels to add to all Sessions created in this client. 
+     */
+    public Builder setSessionLabels(Map<String, String> sessionLabels) {
```
tomayles: Do we allow null values (and coerce them to empty) or require that the user give us empty string values in the first place?

Also, if we are given an invalid label, how is that revealed to the user?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -205,6 +219,10 @@ public SessionPoolOptions getSessionPoolOptions() {
     return sessionPoolOptions;
   }
 
+  public Map<String, String> getSessionLabels() {
+	  return sessionLabels;
```
garrettjonesgoogle: indentation

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
vkedia: I was thinking that we will not do anything fancy and just pass the map as it is. The CreateSession call would fail if there is something illegal here and users would get a SpannerException whenever they do any operation on DatabaseClient (just like they do in case of errors like permission denied).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -75,7 +77,8 @@ public ServiceRpc create(SpannerOptions options) {
   private final int prefetchChunks;
   private final int numChannels;
   private final String userAgent;
-
+  private final ImmutableMap<String, String> sessionLabels;
```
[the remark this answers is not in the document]
vkedia: Ignoring this to be consistent.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
vkedia: Added null check and linked to the sessions doc for valid syntax.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerRpc.java
```diff
@@ -325,14 +326,18 @@ public Operation getOperation(String name) throws SpannerException {
...
-  public com.google.spanner.v1.Session createSession(
-      String databaseName, @Nullable Map<Option, ?> options) {
-    CreateSessionRequest request =
-        CreateSessionRequest.newBuilder().setDatabase(databaseName).build();
+  public Session createSession(
+      String databaseName, @Nullable Map<String, String> labels, @Nullable Map<Option, ?> options) {
+    CreateSessionRequest.Builder request =
+        CreateSessionRequest.newBuilder().setDatabase(databaseName);
+    if (labels != null && !labels.isEmpty()) {
+    	  Session.Builder session = Session.newBuilder().putAllLabels(labels);
```
vam-google: I'm missing context here, but this seems weird. In `creteSession()` method, in case if there are any labels provided you _create_ some additional session on client side, to use it as an input parameter for server call, which will return the actual session? In general, why is `session` a field in `CreateSessionRequest` object?

vkedia: We are not creating additional session. CreateSession rpc would create exactly 1 session irrespective of whether the session field is set or not. If there are no labels, we do not need to set the session field. 
I think session being a field in CreateSessionRequest is like any other Create* rpc.
https://cloud.google.com/apis/design/standard_methods#create

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -153,6 +159,22 @@ public Builder setSessionPoolOption(SessionPoolOptions sessionPoolOptions) {
...
     /**
+     * Sets the labels to add to all Sessions created in this client.
+     * 
+     * @param sessionLabels Map from label key to label value. Label value cannot be null. For
+     *     more information on valid syntax see
+     *     <a href="https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Session">
+     *     api docs </a>.
+     */
+    public Builder setSessionLabels(Map<String, String> sessionLabels) {
+      for (String value : sessionLabels.values()) {
```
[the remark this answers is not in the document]
vkedia: Changed to create an immutable copy here itself. Also added some tests for the exception case.

vkedia:
@tomayles Does this look good?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -153,6 +159,23 @@ public Builder setSessionPoolOption(SessionPoolOptions sessionPoolOptions) {
     }
 
     /**
+     * Sets the labels to add to all Sessions created in this client.
+     * 
+     * @param sessionLabels Map from label key to label value. Label value cannot be null. For
```
tomayles: Keys and values cannot be null.

vkedia: Done

PR #2611, opened 2017-11-10
Changes default value for writeSessionFraction to 0.

### LINKED ISSUE #2540: Spanner java client + service account without write permission causes high error rate (403)
Steps to reproduce:

1. Create a service account without write permission.
2. Initialize an instance of DatabaseClient with default options and credentials from above.
3. Check the API dashboard (console.cloud.google.com/apis/dashboard), there will be a bunch of 403 errors.

I dug into this a bit and found out that it is due to the default settings of 

    com.google.cloud.spanner.SessionPoolOptions 

Which sets write session fraction to 0.2. These sessions will repeatedly try to make begin transaction API calls but fail since the service account does not have permission.

vkedia:
Closes #2540

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -81,7 +81,7 @@ public class SessionPoolOptions {
     private int minSessions;
     private int maxSessions = DEFAULT_MAX_SESSIONS;
     private int maxIdleSessions;
-    private float writeSessionsFraction = 0.f;
+    private float writeSessionsFraction;
     private ActionOnExhaustion actionOnExhaustion = DEFAULT_ACTION;
     private int keepAliveIntervalMinutes = 30;
 
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolOptionsTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolOptionsTest.java
@@ -13,55 +13,62 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package com.google.cloud.spanner;
+
+import static com.google.common.truth.Truth.assertThat;
 
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameter;
-import org.junit.runners.Parameterized.Parameters;
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-
-import static junit.framework.TestCase.assertEquals;
+import org.junit.runners.JUnit4;
 
 /**
  * Unit tests for {@link com.google.cloud.spanner.SessionPoolOptions}
  */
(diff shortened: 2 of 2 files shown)


vkedia:
cc @tomayles

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -81,7 +81,7 @@ public static Builder newBuilder() {
     private int minSessions;
     private int maxSessions = DEFAULT_MAX_SESSIONS;
     private int maxIdleSessions;
-    private float writeSessionsFraction = 0.2f;
+    private float writeSessionsFraction = 0.f;
```
tomayles: I don't think we want to change this default, since maintaining some write sessions does in the general case improve performance.

vkedia: That's correct. But I think anyone who cares about performance improvements provided by this parameter would anyways need to tune this according to there read/write ratio. 0.2 is pretty arbitrary.
OTOH setting this to >0 does lead to a bad experience in the case where BeginTransaction is guaranteed to fail such as permission denied.

tomayles:
On Mon, Nov 13, 2017 at 2:26 PM, Vikas Kedia <notifications@github.com>
wrote:

> *@vkedia* commented on this pull request.
> ------------------------------
>
> In google-cloud-spanner/src/main/java/com/google/cloud/spanner/
> SessionPoolOptions.java
> <https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2611#discussion_r150684852>
> :
>
> > @@ -81,7 +81,7 @@ public static Builder newBuilder() {
>      private int minSessions;
>      private int maxSessions = DEFAULT_MAX_SESSIONS;
>      private int maxIdleSessions;
> -    private float writeSessionsFraction = 0.2f;
> +    private float writeSessionsFraction = 0.f;
>
> That's correct. But I think anyone who cares about performance
> improvements provided by this parameter would anyways need to tune this
> according to there read/write ratio. 0.2 is pretty arbitrary.
> OTOH setting this to >0 does lead to a bad experience in the case where
> BeginTransaction is guaranteed to fail such as permission denied.
>
It's arbitrary, but likely a reasonable guess for a lot of users based on
the mix of read-write to read-only transaction usage we see. I would
strongly prefer that the libraries enable good performance by default,
without needing to discover the tuning parameters. As an alternative,
should we disable the feature if we see that 100% of the first N BeginTxn
calls fail with this error?

vkedia:
@tomayles We can do what you suggested but I am not sure adding that complexity is worth it.
Here is another idea. What if we figure out on the client based on the usage what percent of sessions need to be write prepared and then we prepare that many sessions. We can adaptively change the ratio based on usage over some window. Then we can remove this writePrepared parameter and have a single boolean parameter to turn this feature on/off (with default being on). That will help the client home in to the correct ratio and also resolve this issue since the ratio will be 0 in this case.

tomayles:
That sounds good too. At first glance, it seems that it could be more
complex, but having the client auto-tune is very appealing. If we want to
preserve the existing API, we could keep writefraction as an override, but
have it default to a sentinel value that means auto.

vkedia:
Ok. I will then close this pull request and file an issue for auto tuning write fractions

vkedia:
Filed #2624

PR #2613, opened 2017-11-10
support datastore transaction options

neozwu:


garrettjonesgoogle:
Removing myself as reviewer since this is WIP and @vam-google is secondary on Datastore.

neozwu:
Kindly ping

google-cloud-datastore/src/main/java/com/google/cloud/datastore/TransactionImpl.java
```diff
@@ -118,4 +127,9 @@ public void rollback() {
   public Datastore getDatastore() {
     return datastore;
   }
+
+  @Override
+  public ByteString getTransactionId() {
+    return transaction;
```
vam-google: The name was chosen before this code change, but please rename this to `transactionId`.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -62,22 +63,84 @@ public Batch newBatch() {
...
+          setPrevTransactionId(transaction);
+        }
+      }
+    }
+
+    void setPrevTransactionId(Transaction transaction) {
+      options =
+          options
+              .toBuilder()
+              .setReadWrite(
```
vam-google: Please format this  somewhat prettier (one option to consider is to create local variables, for example for the value passed to `.setReadWrite()`).

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -62,22 +63,84 @@ public Batch newBatch() {
...
+          transaction.rollback();
+        }
+        if (options.getModeCase().equals(TransactionOptions.ModeCase.READ_WRITE)) {
+          setPrevTransactionId(transaction);
+        }
+      }
+    }
+
+    void setPrevTransactionId(Transaction transaction) {
+      options =
```
vam-google: This seems not thread-safe.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/Datastore.java
```diff
@@ -45,6 +53,17 @@
...
   }
 
+  /**
+   * A wrapper around {@link TransactionCallable} with {@link TransactionOptions}.
+   * If {@link TransactionOptions} mode is set to ReadWrite, its previous transaction id will
+   * be populated each time the underlying {@link TransactionCallable} is retried.
+   *
+   * @param <T> the type of the return value
+   */
+  interface ReadWriteTransactionCallable<T> {
```
vam-google: It does not look like this interface should be exposed. For example, both versions of runIntransaction() method (with and without options) accept `TransactionCallable<>`. It like the wrapper is an implementation detail (and should be). If that is the case, please remove it from the client's surface.

neozwu: I have removed `ReadWriteTransactionCallable` in the interface.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -62,22 +63,84 @@ public Batch newBatch() {
...
+        transaction.commit();
+        return value;
+      } catch (Exception ex) {
+        transaction.rollback();
+        throw DatastoreException.propagateUserException(ex);
+      } finally {
+        if (transaction.isActive()) {
+          transaction.rollback();
+        }
+        if (options.getModeCase().equals(TransactionOptions.ModeCase.READ_WRITE)) {
```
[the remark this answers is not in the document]
neozwu: Fixed

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -62,22 +63,84 @@ public Batch newBatch() {
...
+          setPrevTransactionId(transaction);
+        }
+      }
+    }
+
+    void setPrevTransactionId(Transaction transaction) {
```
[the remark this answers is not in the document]
neozwu: Done.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
neozwu: Fixed.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
neozwu: Done.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -86,6 +149,20 @@ public T call() throws DatastoreException {
     }
   }
 
+  @Override
+  public <T> T runInTransaction(final TransactionCallable<T> callable, TransactionOptions options) {
```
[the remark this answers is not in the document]
neozwu: Done.

google-cloud-datastore/src/main/java/com/google/cloud/datastore/TransactionImpl.java
neozwu: Done.

neozwu:
PTAL @vam-google

google-cloud-datastore/src/test/java/com/google/cloud/datastore/DatastoreHelperTest.java
```diff
@@ -159,72 +153,4 @@ public void testFetchWithTransaction() throws Exception {
     assertEquals(entity2, values.get(1));
     verify(transaction);
   }
-
```
vam-google: Why was this removed?

google-cloud-datastore/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>com.google.cloud.datastore</groupId>
       <artifactId>datastore-v1-proto-client</artifactId>
-      <version>1.4.1</version>
+      <version>1.5.1</version>
```
vam-google: should this match datastore-v1-protos (i.e. either both 1.5.0 or both 1.5.1)? I know it did not match before, but it seems like it was wrong originally.

garrettjonesgoogle: This version should just be removed at this point, it is in google-cloud-bom now.

google-cloud-datastore/src/test/java/com/google/cloud/datastore/DatastoreHelperTest.java
neozwu: After I cleaned up the surface, I got rid of the DatastoreHelper "default method", hence deleting the tests. The new impl logics are tested in DatastoreTest and ITs.

garrettjonesgoogle:
@neozwu would you be able to finish this up?

google-cloud-datastore/pom.xml
neozwu: Removed.

neozwu:
@garrettjonesgoogle should be ready to merge once API goes public

garrettjonesgoogle:
I added the "status: blocked" tag since this is blocked on the server side.

PR #2615, opened 2017-11-13
pubsub: make connection log less aggressively

### LINKED ISSUE #2614: pubsub: streaming connection logs too aggressively
First reported [here](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1827#issuecomment-343760699).

The connection logs a warning even though the subscriber is shutting down. When shutting down, this error is normal. The subscriber waits to process all messages, but by the time this happens, another message might have already arrived. Trying to ack this message could error: it races with us trying to shutdown streaming channel.

This is sub-optimal but "OK". The message will be redelivered to another machine. The fix for this is to not log such errors when the the subscriber is shutting down.

pongad:
Fixes #2614

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
@@ -275,9 +275,8 @@ final class StreamingSubscriberConnection extends AbstractApiService implements 
         requestObserver.onNext(request);
       }
     } catch (Exception e) {
-      if (isAlive()) {
-        logger.log(Level.WARNING, "failed to send ack operations", e);
-      }
+      Level level = isAlive() ? Level.WARNING : Level.FINER;
+      logger.log(level, "failed to send ack operations", e);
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
```diff
@@ -275,7 +275,9 @@ public void sendAckOperations(
         requestObserver.onNext(request);
       }
     } catch (Exception e) {
-      logger.log(Level.WARNING, "failed to send acks", e);
+      if (isAlive()) {
+        logger.log(Level.WARNING, "failed to send ack operations", e);
+      }
```
garrettjonesgoogle: Maybe log at a verbose level if it's not alive?

pongad: Done

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #2618, opened 2017-11-13
remove tests from SubscriberTest

pongad:
These tests are moved, in simpler forms, to MessageDispatcherTest
and ITPubSubTest.

Eventually, we should test streaming reconnection logic independently
and remove SubscriberTest altogether.


pongad:
@mdietz94

PR #2625, opened 2017-11-15
Rejecting conflicting field paths

schmidt-sebastian:
Second attempt for https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2582

This is the same PR, but I removed the clashing `isPrefixOf`.


pongad:
This LGTM but wasn't #2582 already merged? Was it reverted or something?

schmidt-sebastian:
> This LGTM but wasn't #2582 already merged? Was it reverted or something?

Yeah, I had to revert it because it broke the google-cloud-java build.

Issue #2591, opened 2017-11-08
bigquery: tabledata.list tracking


pongad:
This issue tracks things we need to migrate query to use `tabledata.list` instead of `getQueryResults` RPC.
Work is in `bq-perf` branch.

- [x] ~Remove~ InternalApi `QueryResponse` type
- [x] ~Remove~ InternalApi `BigQuery.getQueryResults`
  - Use `getJob().getQueryResults` instead
- [x] Give `QueryResults` fewer fields (#2592)
  - To get details of a job, use `Job`
- [x] Should `Job.getQueryResults` wait for query to finish?
- [x] What should `BigQuery.query` do?
  - alias of `create`?
  - return completed job?
  - return iterator?
    - What option should it accept?
    - Accept option for iterating, use default option for waiting?
- [x] Consider unified types of `Job.getQueryResults` and `BigQuery.listTableData`
- [x] Make sure benchmark actually improves
- [x] Make sure we don't retrogress on small queries

pongad:
@vam-google @tswast I'm just copying from notes of meeting we had a couple of weeks ago. Could you check that we're on the same page?

tswast:
:+1:

Regarding "What should BigQuery.query do?" my preference is "alias of create". But the other options may be acceptable if any exceptions thrown while waiting for the job make it easy to get the full job resource so that the user can investigate errors further.

pongad:
Everything done now. We're going to merge back to master.

PR #2637, opened 2017-11-17
bigquery: read query results only if query finished

pongad:


pongad:
Tim, this addresses your concern from #2593.

I have one concern though: wouldn't it be easier to just make `Job.getQueryResults` wait for the job to complete? Even when we move to `tabledata.list`, we'll still need to call `getQueryResults` RPC to read the schema of the table.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -268,12 +267,14 @@ public QueryResult getQueryResults(QueryResultsOption... options) {
...
     }
-
-    // TODO(pongad): check job stats.state instead.
-    QueryResponse response = bigquery.getQueryResults(getJobId(), options);
-    if (!response.jobCompleted()) {
-      throw new IllegalStateException("the job hasn't completed yet");
+    if (getStatus().getState() != JobStatus.State.DONE) {
+      throw new BigQueryException(BigQueryException.UNKNOWN_CODE, "the job hasn't completed yet");
     }
+    if (getStatus().getError() != null) {
```
tswast: I assume `getError()` corresponds to [job.status.errorResult](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#status.errorResult)?

pongad: Yes it is.

pongad:
@tswast I thought about this a little more. Making `getQueryResults()` (the Java method, not RPC) wait for response can be a little confusing, since the `Job` won't be updated to reflect the fact that the query has already finished.

I've noted this down in #2591, but I'll submit this for now. It's in a branch so we can change our minds later.

PR #2638, opened 2017-11-17
pubsub: clean out dead code

pongad:


garrettjonesgoogle:
Could you give some context for why this code is dead?

pongad:
Ah sorry. Most of these tests are moved to `MessageDispatcherTest`. The `isStreaming` is dead because only tests left here tests streaming, so it's always true.

The special logic in MessageReceiver is dead because no test depends on that logic anymore.

pongad:
@garrettjonesgoogle Ping

pongad:
cc @jba

PR #2639, opened 2017-11-17
pubsub: re-implement periodic deadline adjustment

pongad:
@mdietz94 Do you also want to take a look?

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
@@ -82,7 +82,7 @@ class MessageDispatcher {
 
   // The deadline should be set before use. Here, set it to something unreasonable,
   // so we fail loudly if we mess up.
-  private final AtomicInteger messageDeadlineSeconds = new AtomicInteger(-1234);
+  private final AtomicInteger messageDeadlineSeconds = new AtomicInteger(60);
   private final AtomicBoolean extendDeadline = new AtomicBoolean(true);
   private final Lock jobLock;
   private ScheduledFuture<?> backgroundJob;
@@ -218,7 +218,8 @@ class MessageDispatcher {
       // Do not adjust deadline concurrently with extendDeadline or processOutstandingAckOperations.
       // The following sequence can happen:
       //  0. Initially, deadline = 1 min
-      //  1. Thread A (TA) wants to send receipts, reads deadline = 1m, but stalls before actually sending request
+      //  1. Thread A (TA) wants to send receipts, reads deadline = 1m, but stalls before actually
+      // sending request
       //  2. Thread B (TB) adjusts deadline to 2m
       //  3. TB calls extendDeadline, modack all messages to 2m, schedules next extension in 2m
       //  4. TA sends request, modacking messages to 1m.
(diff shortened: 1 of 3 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -78,12 +80,12 @@
...
 
-  private final Lock alarmsLock;
-  // The deadline should be set by the subscriber connection before use,
-  // but set it to some reasonable value just in case.
-  private final AtomicInteger messageDeadlineSeconds = new AtomicInteger(10);
-  private ScheduledFuture<?> ackDeadlineExtensionAlarm;
-  private ScheduledFuture<?> pendingAcksAlarm;
+  // The deadline should be set before use. Here, set it to something unreasonable,
+  // so we fail loudly if we mess up.
+  private final AtomicInteger messageDeadlineSeconds = new AtomicInteger(-1234);
```
mdietz94: I think it would be better to set it to a reasonable default. I think it would be better to be defensive rather than fail loudly in the client library.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -399,6 +401,23 @@ public void run() {
...
+    int sec = Ints.saturatedCast(secLong);
+
+    // Use Ints.constrainToRange when we get guava 21.
+    if (sec < Subscriber.MIN_ACK_DEADLINE_SECONDS) {
+      sec = Subscriber.MIN_ACK_DEADLINE_SECONDS;
+    } else if (sec > Subscriber.MAX_ACK_DEADLINE_SECONDS) {
+      sec = Subscriber.MAX_ACK_DEADLINE_SECONDS;
+    }
+
+    messageDeadlineSeconds.set(sec);
```
mdietz94: It seems like this should either set messageDeadlineSeconds or return the variable, callers should be responsible for dealing with the return value if they need it and they can also set messageDeadlineSeconds.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -139,10 +135,6 @@ private Subscriber(Builder builder) {
     cachedSubscriptionNameString = subscriptionName.toString();
     ackExpirationPadding = builder.ackExpirationPadding;
```
mdietz94: We should verify that ackExpirationPadding is less than 10 seconds. Otherwise it would be possible to schedule a deadline negative seconds.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
pongad: Done

pongad:
Thank you @mdietz94 PTAL

PR #2642, opened 2017-11-20
bigquery: use tabledata.list for query

pongad:
This sped up query benchmark from 470s to 360s.

This breaks "schema injection" however,
so users cannot call `row.get(String)`.
This will be fixed in a follow up.

Updates #2591.

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
@@ -1117,6 +1117,7 @@ public class BigQueryImplTest {
             .setJobReference(queryJob.toPb())
             .setId(JOB)
             .setStatus(new com.google.api.services.bigquery.model.JobStatus().setState("DONE"));
+    jobResponsePb.getConfiguration().getQuery().setDestinationTable(TABLE_ID.toPb());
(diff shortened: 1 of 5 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -1144,21 +1145,24 @@ public void testQueryRequestCompleted() throws InterruptedException {
...
-    QueryResult result =
+    // TODO(pongad) revert back to QueryResult
+    Page<FieldValueList> result =
         bigquery
             .query(QUERY_JOB_CONFIGURATION_FOR_QUERY, queryJob)
             .getQueryResults(QueryResultsOption.pageSize(42L));
-    assertEquals(null, result.getSchema());
-    assertEquals(1L, result.getTotalRows());
+    // assertEquals(null, result.getSchema());
+    // assertEquals(1L, result.getTotalRows());
```
tswast: `tabledata.list` returns `totalRows` as well. I'm slightly surprised there isn't a custom type already for row data that includes this property. I've seen feedback on the Python API that people do use this property.

pongad: I agree. Noted in #2591.

PR #2645, opened 2017-11-21
Treat RESOURCE_EXHAUSTED with retry information as aborted.

mairbek:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
```diff
@@ -48,4 +55,22 @@ public ErrorCode getErrorCode() {
   enum DoNotConstructDirectly {
     ALLOWED
   }
+
+  /**
+   * Return the retry delay for transaction in milliseconds. Return -1 if this does not specify any
+   * retry delay. In that case, clients should fall back to a locally computed retry delay.
+   */
+  public static long extractRetryDelay(Throwable cause) {
```
vkedia: this need not be public.

vkedia: Can it take SpannerException instead of Throwable?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
```diff
@@ -48,4 +55,22 @@ public ErrorCode getErrorCode() {
   enum DoNotConstructDirectly {
     ALLOWED
   }
+
+  /**
+   * Return the retry delay for transaction in milliseconds. Return -1 if this does not specify any
+   * retry delay. In that case, clients should fall back to a locally computed retry delay.
```
vkedia: `In that case, clients should fall back to a locally computed retry delay.`
This statement is only true in the specific case of ABORTED not in all retries. So please remove from here and put it on getRetryDelayMillis

google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbortedException.java
```diff
@@ -43,21 +36,7 @@
       DoNotConstructDirectly token, @Nullable String message, @Nullable Throwable cause) {
     super(token, ErrorCode.ABORTED, IS_RETRYABLE, message, cause);
   }
-
-  /**
-   * Return the retry delay for transaction in milliseconds. Return -1 if this does not specify any
-   * retry delay. In that case, clients should fall back to a locally computed retry delay.
-   */
```
vkedia: I think this Javadoc still needs to be there

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -1296,15 +1296,12 @@ TransactionSelector getTransactionSelector() {
 
     @Override
     public void onError(SpannerException e) {
-      if (e.getErrorCode() == ErrorCode.ABORTED) {
-        long delay = -1L;
-        if (e instanceof AbortedException) {
-          delay = ((AbortedException) e).getRetryDelayInMillis();
-        }
+      long delay = SpannerException.extractRetryDelay(e.getCause());
+      if (e.getErrorCode() == ErrorCode.ABORTED || (e.getErrorCode() ==
```
vkedia: I don't think adding this condition here is correct since this is used to decide whether to retry the whole transaction (and not just the individual rpc). But RESOURCE_EXHAUSTED should behave like UNAVAILABLE and only retry that specific rpc. You already made the change in `isRetryable` in `SpannerExceptionFactory` which will achieve that effect except that you need to also use the delay specified.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java
```diff
@@ -48,4 +55,26 @@ public ErrorCode getErrorCode() {
   enum DoNotConstructDirectly {
     ALLOWED
   }
+
+  /**
+   * Return the retry delay for transaction in milliseconds. Return -1 if this does not specify any
```
vkedia: "retry delay for operation" since it need not have been a transaction

google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerExceptionFactoryTest.java
```diff
@@ -51,6 +51,33 @@ public void connectionClosedIsRetryable() {
...
+    Metadata trailers = new Metadata();
+    Metadata.Key<RetryInfo> key = ProtoUtils.keyForProto(RetryInfo.getDefaultInstance());
+    RetryInfo retryInfo =
+        RetryInfo.newBuilder()
+            .setRetryDelay(Duration.newBuilder().setNanos(1000000).setSeconds(1L))
+            .build();
+    trailers.put(key, retryInfo);
+    SpannerException e =
+        SpannerExceptionFactory.newSpannerException(new StatusRuntimeException(status, trailers));
+    assertThat(e.isRetryable()).isTrue();
```
vkedia: can you also assert that the correct delay is returned by getRetryDelay

mairbek: Done

PR #2649, opened 2017-11-21
Storage: Download Public Data

yihanzhen:
Add a method to create an unauthenticated client. Also, add integration tests that an unauthenticated client have read access to a public bucket.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
@@ -132,7 +132,7 @@ public class StorageOptions extends ServiceOptions<Storage, StorageOptions> {
   /**
    * Returns a unauthenticated {@code StorageOptions} instance.
    */
-  public static StorageOptions getUnauthenticated() {
+  public static StorageOptions getUnauthenticatedInstance() {
     return newBuilder().setCredentials(NoCredentials.getInstance()).build();
   }
 
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
@@ -24,6 +24,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static com.google.common.truth.Truth.assertThat;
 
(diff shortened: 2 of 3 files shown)


hzyi-google:
Please take a look.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -128,6 +129,13 @@ public static StorageOptions getDefaultInstance() {
     return newBuilder().build();
   }
 
+  /**
+   * Returns a unautenticated {@code StorageOptions} instance.
```
andreamlin: sp: "authenticated"

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -128,6 +129,13 @@ public static StorageOptions getDefaultInstance() {
     return newBuilder().build();
   }
 
+  /**
+   * Returns a unauthenticated {@code StorageOptions} instance.
+   */
+  public static StorageOptions getUnauthenticated() {
```
pongad: `getUnauthenticatedInstance`?
In vacuum, I'm OK with the name, but a "parallel" method is called `getDefaultInstance` so I think it's better to have parallel names as well.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1317,6 +1318,55 @@ public void testGetBlobs() {
...
+    assertEquals(7903, bytes.length);
+    int numBlobs = 0;
+    Iterator<Blob> blobIterator = unauthorizedStorage
+      .list(landsatBucket, Storage.BlobListOption.prefix(landsatPrefix))
+      .iterateAll().iterator();
+    while (blobIterator.hasNext()) {
+      numBlobs++;
+      blobIterator.next();
+    }
+    assertEquals(13, numBlobs);
```
pongad: @garrettjonesgoogle @andreamlin Is there guidance for whether we use Truth or JUnit for asserts?

For new tests, I think Truth makes sense, since it gives better error messages.

garrettjonesgoogle: Let's go with Truth for new tests.

hzyi-google:
PTAL

pongad:
Apart from @andreamlin 's comment, LGTM. I'll let Andrea +2

PR #2653, opened 2017-11-22
bigquery: simplify query types

pongad:
Updates #2591.

Previous PR, #2642, broke Job.getQueryResults in two ways:
1. It wasn't possible to get total number of rows.
2. The returned iterator doesn't record schema, so it wasn't possible
   to call `row.get(String)`.

This PR prepares the fix.
Actually landing the fix will change tests,
so it is punted to another PR to keep changes small.

From previous meeting, it was our decision to remove
- BigQuery.getQueryResults method
- QueryResponse class

However, the method is being used internally by the client,
and interfaces only have public members.
Removing them altogether entails a big surgery,
so this PR instead mark them @InternalApi.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
@@ -434,8 +434,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
 
     /**
      * Returns an option that sets how long to wait for the query to complete, in milliseconds,
-     * before returning. Default is 10 seconds. If the timeout passes before the job completes,
-     * {@link QueryResponse#jobCompleted()} will be {@code false}.
+     * before returning. Default is 10 seconds.
      */
     public static QueryResultsOption maxWaitTime(long maxWaitTime) {
       checkArgument(maxWaitTime >= 0);
@@ -1064,7 +1063,7 @@ public interface BigQuery extends Service<BigQueryOptions> {
   /**
    * Returns results of the query associated with the provided job.
    *
-   * <p>Users are encouraged to use {@link Job.getQueryResults(QueryResultsOption...)} instead.
+   * <p>Users are encouraged to use {@link Job#getQueryResults(QueryResultsOption...)} instead.
    */
   @InternalApi
   QueryResponse getQueryResults(JobId jobId, QueryResultsOption... options);
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryError.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryError.java
@@ -23,13 +23,15 @@ import com.google.common.base.MoreObjects;
 import java.io.Serializable;
 import java.util.Objects;
 
+// TODO(pongad): Clean up this doc, depending on what we decide to do with query().
+
 /**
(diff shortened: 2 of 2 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResult.java
```diff
@@ -16,71 +16,35 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.api.gax.paging.Page;
+import com.google.common.base.Function;
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.Iterables;
 import java.util.Objects;
 
-public class QueryResult extends PageImpl<FieldValueList> {
+// TODO(pongad): Serializable ?
```
pongad: @tswast I don't completely understand what being `Serializable` requires. Does this class needs to be changed in some way?

I haven't seen any errors, but thought I should actually think about it instead of blindly typing :laughing:

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -278,6 +278,8 @@ public Job waitFor(RetryOption... waitOptions) throws InterruptedException {
 
     TableId table = ((QueryJobConfiguration) getConfiguration()).getDestinationTable();
     // TODO(pongad): return QueryResult so we can inject schema.
+    // QueryResponse response = bigquery.getQueryResults(getJobId()); // should return immediately
+    // return new QueryResult(response.getSchema(), response.getTotalRows(), bigquery.listTableData(table));
     return bigquery.listTableData(table);
```
tswast: Can you call `.withSchema()` here? (I guess not due to pagination?)

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResult.java
```diff
@@ -16,71 +16,35 @@
...
-  private QueryResult(Builder builder) {
-    super(builder.pageFetcher,
-        builder.cursor,
-        builder.results != null ? builder.results : ImmutableList.<FieldValueList>of());
-    this.schema = builder.schema;
-    this.totalRows = builder.totalRows;
+  private final Page<FieldValueList> pageNoSchema;
+  private final Function<FieldValueList, FieldValueList> addSchemaFunc;
+
+  QueryResult(final Schema schema, long totalRows, Page<FieldValueList> pageNoSchema) {
```
tswast: It might be good to get the `totalRows` from the call to `tabledata.list`, but this does introduce the issue that the total rows won't be populated until you start iterating.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
pongad: Right now `listTableData` just returns a `Page` so we don't have a good method to put the schema in.

We can do that once we uncomment the code above it though.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResult.java
pongad: We should do this a little later. We want to also return `totalRows` from `listTableData` (method, not RPC) anyway. After we do that, it will be easy to do this.

PR #2658, opened 2017-11-23
Code Drop: Adding the Collection Package

schmidt-sebastian:
This add the immutable tree data structure that we use in the Mobile Clients. It will be used by Firestore Watch.

This is based on the same Collection package that is already open-sourced here: https://github.com/firebase/firebase-admin-java/tree/master/src/main/java/com/google/firebase/database/collection

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
@@ -32,8 +32,8 @@ import java.util.Map;
  * avoid degrading performance with increasing collection size it will automatically convert to a
  * RBTreeSortedMap after an insert call above a certain threshold.
  *
- * Note: This package is copied from https://github.com/firebase/firebase-admin-java/tree/master/src
- * /main/java/com/google/firebase/database/collection
+ * Note: This package is copied from https://github.com/firebase/firebase-admin-java/tree/master/
+ * src/main/java/com/google/firebase/database/collection
  */
 @InternalApi
 public class ArraySortedMap<K, V> extends ImmutableSortedMap<K, V> {
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ImmutableSortedMapIterator.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ImmutableSortedMapIterator.java
@@ -25,8 +25,8 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Stack;
 
-// Note: This package is copied from https://github.com/firebase/firebase-admin-java/tree/master/src
-// /main/java/com/google/firebase/database/collection
+// Note: This package is copied from https://github.com/firebase/firebase-admin-java/tree/master/
+// src/main/java/com/google/firebase/database/collection
 @InternalApi
 public class ImmutableSortedMapIterator<K, V> implements Iterator<Map.Entry<K, V>> {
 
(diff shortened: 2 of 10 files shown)


pongad:
I'm good with this but not sure how I feel about all these collections being public. I like that they are in different package though. @schmidt-sebastian Do we actually expose these types to the user? If not, I think we should at least `@InternalApi` all of them.

@garrettjonesgoogle Do you have an opinion?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
```diff
@@ -0,0 +1,289 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
```
pongad: I think we're turning `Google Inc` into `Google LLC`

schmidt-sebastian: I just merged in Master and didn't see these changes. Should we update the clients in one big PR rather than do it incrementally?

schmidt-sebastian:
> I'm good with this but not sure how I feel about all these collections being public. I like that they are in different package though. @schmidt-sebastian Do we actually expose these types to the user? If not, I think we should at least @InternalApi all of them.

The package is only public because of how Java's visibility scope works. I marked them all as `@InternalApi`.

schmidt-sebastian:
> We should really consider making a separate/independent artifact out of this. Currently this code is duplicated in Android SDK, Admin SDK and now here. @bjornick and I were planning to make a separate Maven artifact from this as part of an ongoing open sourcing effort. Lets talk about this when we get some free cycles.

I don't see the Android RTDB SDK ever taking advantage of this. The Firestore SDK could, but we would likely need to ship with ProGuard.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
pongad: Big PR is ok with me

pongad:
This is fine with me but I think @garrettjonesgoogle should do a quick pass.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
```diff
@@ -0,0 +1,291 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore.collection;
+
+import com.google.api.core.InternalApi;
+import java.util.*;
```
garrettjonesgoogle: Don't use star imports (applies to multiple files)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
```diff
@@ -0,0 +1,291 @@
...
+
+/**
+ * This is an array backed implementation of ImmutableSortedMap. It uses arrays and linear lookups
+ * to achieve good memory efficiency while maintaining good performance for small collections. To
+ * avoid degrading performance with increasing collection size it will automatically convert to a
+ * RBTreeSortedMap after an insert call above a certain threshold.
+ */
```
garrettjonesgoogle: If these were copied from somewhere else, could we include a link to the original implementation?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
schmidt-sebastian: This is from the original source. I updated the files here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/collection/ArraySortedMap.java
schmidt-sebastian: They are copied out of Google3, but are available elsewhere. I added a note that points to the location in the Firebase Admin SDK.

PR #2659, opened 2017-11-24
Fix typo in subscription id on CreateSubscriptionAndConsumeMessages example

fmachado091:
Fix for what I believe is a typo in the mock id on the `CreateSubscriptionAndConsumeMessages` example.


garrettjonesgoogle:
Good catch!

PR #2662, opened 2017-11-25
Adding v1beta1 Ordering Semantics

schmidt-sebastian:
This is required for Watch.

Where applicable, the semantics are copied from Node and the Android SDK. The Unit test is copied 1:1 from the Node SDK.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BasePath.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BasePath.java
@@ -95,7 +95,7 @@ abstract class BasePath<B extends BasePath<B>> implements Comparable<B> {
   }
 
   /**
-   * Compare the current path against another Path object.
+   * Compare the current path lexicographically against another Path object.
    *
    * @param other The path to compare to.
    * @return -1 if current < other, 1 if current > other, 0 if equal


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BasePath.java
```diff
@@ -93,6 +94,24 @@ boolean isPrefixOf(B path) {
     return true;
   }
 
+  /**
+   * Compare the current path against another Path object.
```
pongad: Maybe also note that this is lexographic ordering?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
```diff
@@ -0,0 +1,241 @@
...
+        case REFERENCE_VALUE:
+          return REF;
+        case GEO_POINT_VALUE:
+          return GEO_POINT;
+        case ARRAY_VALUE:
+          return ARRAY;
+        case MAP_VALUE:
+          return OBJECT;
+        default:
+          throw new IllegalStateException("Could not detect value type for " + value);
```
pongad: IllegalArugmentException?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
```diff
@@ -0,0 +1,241 @@
...
+   * @return -1 is left < right, 0 if left == right, otherwise 1
+   */
+  public int compare(@Nonnull Value left, @Nonnull Value right) {
+    // First compare the types.
+    TypeOrder leftType = TypeOrder.fromValue(left);
+    TypeOrder rightType = TypeOrder.fromValue(right);
+    int typeComparison = leftType.compareTo(rightType);
+    if (typeComparison != 0) {
+      return typeComparison < 0 ? -1 : 1;
+    }
```
pongad: There are lots of these "double compares". Do we care that deeply that the returned values are only -1,0,or 1?

schmidt-sebastian: I rely on this in the test case. But it makes much more sense to convert the end result in the test case itself, so I moved that conversion there.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
schmidt-sebastian: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BasePath.java
schmidt-sebastian: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
schmidt-sebastian: I moved this conversion to the test class.

schmidt-sebastian:
Review addressed. Thanks.

pongad:
LGTM; @garrettjonesgoogle should +2

PR #2665, opened 2017-11-27
Adding Firestore Watch

schmidt-sebastian:
This PR adds snapshot listeners to the Firestore Server API for Java. It's combination of an Android port (DocumentSet, DocumentChange, ExponentialBackoff, ListenerRegistration) and a Node port (Watch).

The files ported from Android are almost exact copies. ExponentialBackoff has some constants added and DocumentSet & DocumentChange has some unneeded methods removed.

This PR also includes the files that are currently pending review at 
- https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2662 (Sorting)
- https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2658 (Collections package)

If you have comments about Order.java, OrderTest.java or any of the classes under collection/, please comment in the corresponding PRs. Otherwise, it might make sense to merge these PRs first (or at least one of them so I can create a diffbase).

I also rebased this so that all Watch-only changes are in 8f96297.

Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2577

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
@@ -67,26 +67,15 @@ class Watch implements ApiStreamObserver<ListenResponse> {
    */
   private static final int WATCH_TARGET_ID = 0xD0;
 
-  private static int RETRY_INITIAL_DELAY_MS = 1000;
-
-  /** The maximum backoff time in milliseconds. */
-  private static int RETRY_MAX_DELAY_MS = 60 * 1000;
-
-  /** The factor to increase the backup by after each failed attempt. */
-  private static double RETRY_BACKOFF_FACTOR = 1.5;
-
   private static RetrySettings RETRY_SETTINGS =
       RetrySettings.newBuilder()
-          .
           // The initial backoff time in seconds after an error.
           // Set to 1s according to https://cloud.google.com/apis/design/errors.
-          setInitialRetryDelay(Duration.ofSeconds(1))
-          .
+          .setInitialRetryDelay(Duration.ofSeconds(1))
           // The maximum backoff time in minutes.
-          setMaxRetryDelay(Duration.ofMinutes(1))
-          .
+          .setMaxRetryDelay(Duration.ofMinutes(1))
           // The factor to increase the backup by after each failed attempt.
-          setRetryDelayMultiplier(1.5)
+          .setRetryDelayMultiplier(1.5)
           .setJittered(true)
           .build();
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/spi/v1beta1/GrpcFirestoreRpc.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/spi/v1beta1/GrpcFirestoreRpc.java
@@ -193,6 +193,7 @@ public class GrpcFirestoreRpc implements FirestoreRpc {
   @Override
   public BidiStreamingCallable<ListenRequest, ListenResponse> listenCallable() {
     return firestoreStub.listenCallable();
+  }
 
   // This class is needed solely to get access to protected method setInternalHeaderProvider()
   private static class FirestoreSettingsBuilder extends FirestoreSettings.Builder {


schmidt-sebastian:
Ping :)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSet.java
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Returns a copy of the documents in this set as array. This is O(n) in the size of the set.
+   */
+  List<DocumentSnapshot> toList() {
+    List<DocumentSnapshot> documents = new ArrayList<>(size());
+    for (DocumentSnapshot document : this) {
+      documents.add(document);
+    }
+    return documents;
```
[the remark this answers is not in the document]
schmidt-sebastian: DocumentSet is not a collection unfortunately. It merely implements an iterator.

Note that this code is also taken from the Android SDK (as well as most of ExponentialBackoff and DocumentChange).

schmidt-sebastian:
This PR is rebased against `master` and no longer contains the changes from #2662 and
#2658.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSet.java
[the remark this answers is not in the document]
pongad: Oh derp, you're right.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentChange.java
```diff
@@ -0,0 +1,83 @@
...
+
+package com.google.cloud.firestore;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A DocumentChange represents a change to the documents matching a query. It contains the document
+ * affected and a the type of change that occurred (added, modifed, or removed).
+ */
+public class DocumentChange {
```
pongad: Does it make sense to make this `@AutoValue`? If you'd rather keep it Android-like, it's fine with me.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
```diff
@@ -391,6 +392,53 @@ public void remove() {
...
+  public ListenerRegistration addSnapshotListener(
+      @Nonnull Executor executor, @Nonnull final EventListener<DocumentSnapshot> listener) {
+    return Watch.forDocument(this)
+        .runWatch(
+            executor,
+            new EventListener<QuerySnapshot>() {
+              @Override
+              public void onEvent(
+                  @Nullable QuerySnapshot value, @Nullable FirestoreException error) {
+                if (value != null) {
```
pongad:
```java
if (value == null) {
  listener.onEvent(null, error);
  return;
}
```
?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
```diff
@@ -0,0 +1,580 @@
...
+ * This class is thread-compatible when called through the methods defined in ApiStreamObserver.
+ * It synchronizes on its own instance so it is advisable not to use this class for external
+ * synchronization.
+ */
+class Watch implements ApiStreamObserver<ListenResponse> {
+  /**
+   * Target ID used by watch. Watch uses a fixed target id since we only support one target per
+   * stream.
+   */
+  private static final int WATCH_TARGET_ID = 0xD0;
```
pongad: Does the value 0xD0 actually mean anything?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
```diff
@@ -0,0 +1,580 @@
...
+    target.getDocumentsBuilder().addDocuments(documentReference.getName());
+    target.setTargetId(WATCH_TARGET_ID);
+
+    return new Watch(
+        (FirestoreImpl) documentReference.getFirestore(),
+        target.build(),
+        new Comparator<DocumentSnapshot>() {
+          @Override
+          public int compare(DocumentSnapshot o1, DocumentSnapshot o2) {
+            return 0;
```
pongad: I don't really understand why. Is it because we'll only ever compare a DocumentSnapshot against itself?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
```diff
@@ -0,0 +1,580 @@
...
+                    status, "Backend ended Listen stream: " + change.getCause().getMessage()));
+            break;
+          case CURRENT:
+            current = true;
+            break;
+          case RESET:
+            resetDocs();
+            break;
+          default:
+            closeStream(FirestoreException.invalidState("Encountered invalid target change type"));
```
pongad: also report the type we see?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
```diff
@@ -0,0 +1,580 @@
...
+      return true;
+    }
+
+    for (int targetId : targetIds) {
+      if (targetId == currentId) {
+        return true;
+      }
+    }
+
+    return false;
```
pongad: return targetIds.contains(currentId)?

schmidt-sebastian:
> I have to admit I don't really understand how Watch works...

If you are willing to invest more time, you can find out more here: https://docs.google.com/document/d/1PjoomZav75_wevZxTarHWqkAlcLnNBAhLBJ4NN_eCPs/edit#bookmark=id.861ohp7wefbp

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
```diff
@@ -0,0 +1,580 @@
...
+
+    if (throwable instanceof StatusRuntimeException) {
+      status = ((StatusRuntimeException) throwable).getStatus();
+    } else if (throwable instanceof StatusException) {
+      status = ((StatusException) throwable).getStatus();
+    }
+    return status;
+  }
+
+  /** Determines whether we need to initiate a longer backoff due to system overload.b */
```
schmidt-sebastian: Typo.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentChange.java
schmidt-sebastian: I think it's better for us in the long run if we share code where we can. It should also make the chunks of code that need thorough reviewing a lot smaller.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
schmidt-sebastian: SGTM

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
schmidt-sebastian: No, I just chose it to be different from Node Watch. Maybe one day it will help us debug something on the backend. 

Added comment.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
schmidt-sebastian: This should actually never need to get called because the number of documents in a DocumentSnapshot is always just one. I added an assert and a comment.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
schmidt-sebastian: D'oh. Thanks.

schmidt-sebastian:
First round of comments addressed.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
```diff
@@ -0,0 +1,572 @@
...
+          }
+        });
+  }
+
+  /**
+   * Checks if the current target id is included in the list of target ids. Returns true if no
+   * targetIds are provided.
+   */
+  private boolean affectsTarget(List<Integer> targetIds, int currentId) {
+    return targetIds == null || targetIds.size() == 0 || targetIds.contains(currentId);
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

pongad:
LGTM but @garrettjonesgoogle should take a look

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ExponentialBackoff.java
```diff
@@ -0,0 +1,107 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/** Helper to implement exponential backoff. */
+class ExponentialBackoff {
```
garrettjonesgoogle: It's slightly annoying that we have yet another exponential backoff implementation here. GAX already has one (https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java). Maybe it's ok since it is package-private; maybe add a note in the documentation stating that this shouldn't be exposed in the future as public due to concerns about duplicating surface?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreException.java
```diff
@@ -73,4 +78,10 @@ static FirestoreException networkException(IOException exception, boolean retrya
   static FirestoreException apiException(ApiException exception) {
     return new FirestoreException(exception);
   }
+
+  @VisibleForTesting
```
garrettjonesgoogle: Use `@InternalApi` (from GAX) instead of `@VisibleForTesting` (which is from Guava, which isn't supposed to be exposed on the surface of google-cloud-java)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ExponentialBackoff.java
schmidt-sebastian: I'm not opposed to migrating to this implementation. Let me take a look at how involved this is.

schmidt-sebastian:
@garrettjonesgoogle I updated the code to use GAX's ExponentialBackoff. The change is here: https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2665/commits/255642c5df6239425254ea3f663d098b412245bd

garrettjonesgoogle:
Glad to see that integrating it wasn't too difficult!

LGTM

PR #2679, opened 2017-12-04
Creating BOM (google-cloud-bom)

### LINKED ISSUE #2307: Create BOM for google-cloud-java

garrettjonesgoogle:
Fixes #2307


garrettjonesgoogle:
cc @saturnism

garrettjonesgoogle:
Fixes #2307

pom.xml
```diff
@@ -136,804 +136,21 @@
...
-        <artifactId>truth</artifactId>
-        <version>0.30</version>
-      </dependency>
-      <dependency>
-        <groupId>com.google.api</groupId>
-        <artifactId>gax-grpc</artifactId>
-        <version>${gax-grpc.version}</version>
-	<classifier>testlib</classifier>
+        <artifactId>google-cloud-dependencies</artifactId>
+        <version>${google-cloud-parent.version}</version>
```
vam-google: This is confusing: the artifact is called `google-cloud-dependencies`, but the version is `google-cloud-parent.version`

google-cloud-dependencies/pom.xml
```diff
@@ -0,0 +1,810 @@
+<?xml version="1.0"?>
```
vam-google: Should be this module explicitly called `google-cloud-bom`?

garrettjonesgoogle: Yeah I think that is clearer.

pom.xml
garrettjonesgoogle: Will fix.

garrettjonesgoogle:
I have addressed feedback and verified that update_versions.sh works. PTAL

garrettjonesgoogle:
I agree with the comment you made in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2332 , that it's good to have compile and test dependencies versioned together. I think we can consider pulling out test deps as a later cleanup. Anyway, Spring (our spiritual leader in this matter) has both compile and test deps together: https://github.com/spring-io/platform/blob/master/platform-bom/pom.xml .

garrettjonesgoogle:
I have moved pure test dependencies back to the main pom.xml, PTAL

joaoandremartins:
This needs documentation that shows how to use the BOM (with Maven at least, also possible with Gradle with a special plugin), which version of the BOM to use, etc.

garrettjonesgoogle:
@jabubake already requested a readme in https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2708 (let's continue any discussion there)

Issue #2620, opened 2017-11-14
Add Storage.create() method which takes a restartable input stream


garrettjonesgoogle:
In https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2359 , retries had to be turned off for `Blob create(BlobInfo blobInfo, InputStream content, BlobWriteOption... options)` because the `InputStream` can only be consumed once.

We should create a new abstraction named `RetryableInputStream` which can produce a new input stream on the original content, and add a new create method which accepts it, so that it is easier for users to create blobs without loading the entire content into memory.

hzyi-google:
Move to backlog.

PR #2682, opened 2017-12-05
Add restartableInputStream, first commit

yihanzhen:
PTAL -
@garrettjonesgoogle Not finished yet, just to make sure the idea aligns with #2620

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/RestartableInputStream.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/RestartableInputStream.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ *
+ */
+public class RestartableInputStream {
+
+    InputStream inputStream;
+
+    public RestartableInputStream(InputStream inputStream) {
+        this.inputStream = inputStream;
+        this.inputStream.mark(Integer.MAX_VALUE);
+    }
+
+    public InputStream getInputStream() {
+        return this.inputStream;
+    }
+
+    public void restart() {
+        try {
+            this.inputStream.reset();
+            this.inputStream.mark(Integer.MAX_VALUE);
+        }
+        catch (IOException e) {
+            throw new IOExpection();
+        }
+    }
+}
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/RetryableInputStreamContent.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/RetryableInputStreamContent.java
@@ -1,0 +1,49 @@
+package com.google.cloud.storage;
+
+import com.google.api.client.http.AbstractInputStreamContent;
+import sun.nio.ch.IOUtil;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public final class RetryableInputStreamContent extends AbstractInputStreamContent {
+
+    private final RestartableInputStream restartableInputStream;
+    private boolean retrySupported = true;
+
+    public RetryableInputStreamContent(String type, RestartableInputStream restartableInputStream) {
+        super(type);
+        this.restartableInputStream = restartableInputStream;
+    }
+
+    @Override
+    public InputStream getInputStream() throws IOException {
+        return restartableInputStream.getInputStream();
+    }
+
+    @Override
+    public long getLength() {
+        return this.length;
+    }
+
+    @Override
+    public boolean retrySupported() {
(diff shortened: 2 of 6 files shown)


garrettjonesgoogle:
@vam-google , could you review this?

google-cloud-storage/src/main/java/com/google/cloud/storage/RestartableInputStream.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ *
+ */
+public class RestartableInputStream {
+
+    InputStream inputStream;
+
+    public RestartableInputStream(InputStream inputStream) {
+        this.inputStream = inputStream;
+        this.inputStream.mark(Integer.MAX_VALUE);
```
vam-google: Can you please share you basic idea/design here? Using `.mark()` seems promising but it is not as straightforward. For example, not each `InputStream` supports  `mark/reset` methods, and the default implementation does not support it: [java.io.InputStream.markSupported()](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/io/InputStream.java#367).

google-cloud-storage/src/main/java/com/google/cloud/storage/RestartableInputStream.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public InputStream getInputStream() {
+        return this.inputStream;
+    }
+
+    public void restart() {
+        try {
+            this.inputStream.reset();
+            this.inputStream.mark(Integer.MAX_VALUE);
```
vam-google: `Integer.MAX_VALUE` here and in the constructor seems too generous. It depends on the implementation, but there is a good chance, that the `InputStream`, which supports `mark/reset` does it by buffering its content in memory. Calling `mark(Integer.MAX_VALUE)` on it would cause it to buffer the whole content of the stream each time (even for the cases, when there were no retry needed, which should be 99% of the cases at least). For big streams (gigabytes+) this would consume too much resources and can easily throw `OutOfMemoryException`. We need to find an acceptable way to guarantee that it will not happen.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -244,6 +246,29 @@ public StorageObject create(StorageObject storageObject, final InputStream conte
     }
   }
 
+  @Override
+  public StorageObject create(StorageObject storageObject, final RestartableInputStream content,
```
vam-google: You need to choose on which level retires should be performed: `Storage` or the `StorageRpc`. It seems the retries normally performed on Storage level (and this is what you do calling `runWithRetries()` in `StorageImpl.create(BlobInfo, RestartableInputStream, BlobWriteOption...)` implementation above.The `Storage.create()` calls `StorageRpc.create()`, if you retry (or have retry-like code, which you have here with `RetryableInputStreamContent`) you can potentially have `n^2` retires instead of `n` retries.

Most probably  you do not need neither this method nor `RetryableInputStreamContent` class.

google-cloud-storage/src/main/java/com/google/cloud/storage/RetryableInputStreamContent.java
```diff
@@ -0,0 +1,49 @@
+package com.google.cloud.storage;
+
+import com.google.api.client.http.AbstractInputStreamContent;
+import sun.nio.ch.IOUtil;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public final class RetryableInputStreamContent extends AbstractInputStreamContent {
```
vam-google: Please check the comment for `HttpStorageRpc.create()`. There is a chance that you do not need this class.

google-cloud-storage/src/main/java/com/google/cloud/storage/RestartableInputStream.java
hzyi-google: The idea here is using `mark()` to mark the start of the `InputStream` so that when retrying, calling `reset()` would make the stream go back to where we want to start. I was not quite confident with this design here either - so besides mark()/reset(), is there a better way to cache the bytes that have been read previously?

google-cloud-storage/src/main/java/com/google/cloud/storage/RestartableInputStream.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+        catch (IOException e) {
+            throw new IOExpection();
```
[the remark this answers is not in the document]
hzyi-google: Yes it is. Since I was not quite confident with the design, this PR was just intended to check how the design was so it did not compile.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
hzyi-google: The only reason that I created another level of indirection is that it is explicitly said that the abstract class `InputStreamContent` does not support retry per [here](https://cs.corp.google.com/piper///depot/google3/java/com/google/api/client/http/InputStreamContent.java?q=InputStreamContent+lang:java&sq=package:piper+file://depot/google3+-file:google3/experimental&dr=CSs&l=23), so reusing `InputStreamContent` might be causing ambiguity and confusion. If that won't be a problem, I can remove this method and the `RetryableInputStreamContent` class.

google-cloud-storage/src/main/java/com/google/cloud/storage/RestartableInputStream.java
hzyi-google: In this case, how about taking the idea from `PushbackInputStream` but instead caching the bytes in memory, creating a new `FileOutputStream` to save the cached bytes to file system, and when retrying, creating a new `FileInputStream` to read back those bytes?

garrettjonesgoogle:
@hzyi-google could you address @vam-google 's comments?

hzyi-google:
A lot of changes need to be made for this, so abandoning. With submit a different PR when finished.

PR #2685, opened 2017-12-07
bigquery: make query() return iterator

pongad:
This also makes queries require fewer RPCs if detailed job statuses are
not required.

If a query fails, then query() throws an exception detailing why
it fails and provides the job ID so users can retrieve more details
themselves.

Updates #2591.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
@@ -250,7 +250,7 @@ public class Job extends JobInfo {
    *
    * <p>If the job hasn't finished, this method waits for the job to complete. However, the state of
    * the current {@code Job} instance is not updated. To get the new state, call {@link
-   * #waitFor(RetryOption...)} or {@link #reload()}.
+   * #waitFor(RetryOption...)} or {@link #reload(JobOption...)}.
    *
    * <p>Example of getting the results of a query job.
    *


pongad:
I still have to fix the examples. I'll do that in a follow up commit. Putting this for review now to reduce review latency.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -560,15 +561,15 @@ public Boolean call() {
   }
 
   @Override
-  public Job query(QueryJobConfiguration configuration, JobOption... options)
-      throws InterruptedException {
+  public QueryResult query(QueryJobConfiguration configuration, JobOption... options)
+      throws InterruptedException, JobException {
```
tswast: Does `JobException` need to move up to the interface, too?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
```diff
@@ -33,6 +34,8 @@
 
   abstract long getTotalRows();
 
+  abstract ImmutableList<BigQueryError> getErrors();
```
tswast: I'm not sure about plumbing errors here, since it doesn't apply to reading results directly from a table. I guess no matter what we want a way to expose these from the `query()` method so we'll want something like `QueryResults`, anyway. I guess it makes sense to add here.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
pongad: Good catch. Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
pongad: What do you think `QueryResults` should look like? How would it relate to the existing `QueryResult` (without the 's')?

You're right that this error doesn't apply for reading directly from table, but we're not using this class to read right? `getQueryResults` returns this when the query completes; `QueryResponse` is a bit of misnomer.

tswast: My ideal is to have a shared class for listing table data so that we can expose properties like `totalRows()` on both, but honestly the split between the two (query results and table listing) does make some sense. I was just thinking aloud.

I do like the current name `QueryResults` better than `QueryResponse`.

pongad: I see what you're saying. I think having a shared class make sense; it'll let users inject schema when iterating table too.

I'm thinking of something like this.
- Keep `QueryResponse` an "internal" class as it is. Just use it to wait for query and get errors.
- Rename `QueryResult` -> `TableResult` or something like it. Use it for both `query` and `listTableData`.
  - Currently we read `totalRows` from `getQueryResults` RPC, we'll change it to read from `tabledata.list` instead.

If this sounds about right to you, let's get this PR merged for now. I'll work on these in a follow up.

pongad:
@tswast PTAL

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryResponse.java
tswast: Yeah, that sounds about right.

PR #2690, opened 2017-12-08
Add support for custom static headers

vam-google:
Dependent on https://github.com/googleapis/gax-java/pull/432 (will fail build until that is released)

Note, PubSub is supported only partially. Full support is dependent on migration of PubSub from Grpc stubs to Gapic stubs. 

The spanner's `SpannerOptions.Builder.setUserAgentPrefix()` method is replaced with `ServiceOptions.Builder.setHeaderProvider()`  and providing "user-agent" value there (it will be prefixing the header provider as explained in `ServiceOptions.Builder.setHeaderProvider()` documentation, see example below).

Usage:
```java
// For manual/semi-manual clients:
//     Resource Manager
//     Translate
//     Compute
//     Storage
//     Datastore
//     Dns
//     BigQuery
//
//     Logging
//     Firestore
//     Spanner
HeaderProvider headerProvider =
    FixedHeaderProvider.create("user-agent", "my-client-name", "custom-header", "custom stuff");
DatastoreOptions options =
    DatastoreOptions.newBuilder().setHeaderProvider(headerProvider).build();
Datastore datastore = options.getService();

// For generated clients:
//    All others, except Pub/Sub
HeaderProvider headerProvider =
    FixedHeaderProvider.create("user-agent", "my-client-name", "custom-header", "custom stuff");
LanguageServiceSettings settings =
    LanguageServiceSettings.newBuilder().setHeaderProvider(headerProvider).build();
LanguageServiceClient client = LanguageServiceClient.create(settings);

// Note, PubSub is supported only partially. Full support is dependent on migration of PubSub from Grpc stubs to Gapic stubs
```

### DIFF
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -127,9 +127,9 @@
     <api-client.version>1.23.0</api-client.version>
 
     <api-common.version>1.2.0</api-common.version>
-    <gax.version>1.15.1-SNAPSHOT</gax.version>
-    <gax-grpc.version>1.15.1-SNAPSHOT</gax-grpc.version>
-    <gax-httpjson.version>0.32.1-SNAPSHOT</gax-httpjson.version>
+    <gax.version>1.16.0</gax.version>
+    <gax-grpc.version>1.16.0</gax-grpc.version>
+    <gax-httpjson.version>0.33.0</gax-httpjson.version>
     <generated-proto-beta.version>0.1.26</generated-proto-beta.version>
     <generated-proto-ga.version>1.0.2</generated-proto-ga.version>
     <google.auth.version>0.9.0</google.auth.version>


google-cloud-examples/src/main/java/com/google/cloud/examples/datastore/snippets/CreateEntity.java
```diff
@@ -22,27 +22,38 @@
...
-/**
- * A snippet for Google Cloud Datastore showing how to create an entity.
- */
+/** A snippet for Google Cloud Datastore showing how to create an entity. */
 public class CreateEntity {
 
   public static void main(String... args) {
-    Datastore datastore = DatastoreOptions.getDefaultInstance().getService();
+    HeaderProvider headerProvider =
+        FixedHeaderProvider.create("user-agent", "my-client-name", "custom-header", "custom stuff");
```
garrettjonesgoogle: I don't think it makes sense to include a custom header example in all of these samples...

vam-google: Oh, that was accidental. This was just me testing this new functionality....

google-cloud-spanner/pom.xml
```diff
@@ -18,38 +18,9 @@
...
-                   toFile="${project.build.outputDirectory}/com/google/cloud/spanner/project.properties"
-                   overwrite="true" />
-              </tasks>
-            </configuration>
-            <goals>
-              <goal>run</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
```
vkedia: this seems unrelated?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -82,7 +84,7 @@ public ServiceRpc create(SpannerOptions options) {
   private SpannerOptions(Builder builder) {
     super(SpannerFactory.class, SpannerRpcFactory.class, builder, new SpannerDefaults());
     numChannels = builder.numChannels;
-    userAgent = builder.userAgentPrefix;
```
vkedia: what happens if the user also specifies this userAgentPrefix?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerRpc.java
```diff
@@ -97,25 +100,23 @@
...
-  private static final Metadata.Key<String> API_CLIENT_KEY =
-      Metadata.Key.of("x-goog-api-client", Metadata.ASCII_STRING_MARSHALLER);
-  private static final Metadata.Key<String> RESOURCE_PREFIX_KEY =
-      Metadata.Key.of("google-cloud-resource-prefix", Metadata.ASCII_STRING_MARSHALLER);
   private static final Pattern DATABASE_PATTERN =
-      Pattern.compile("^(?<database>projects/[^/]*/instances/[^/]*/databases/[^/]*)(.*)?");
+      Pattern.compile("^(?<headerValue>projects/[^/]*/instances/[^/]*/databases/[^/]*)(.*)?");
   private static final Pattern INSTANCE_PATTERN =
-      Pattern.compile("^(?<instance>projects/[^/]*/instances/[^/]*)(.*)?");
+      Pattern.compile("^(?<headerValue>projects/[^/]*/instances/[^/]*)(.*)?");
```
vkedia: why not just define these two in SpannerHeaderProvider and use there instead of passing from here?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerClientHeaderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+          Metadata.Key.of(entry.getKey(), Metadata.ASCII_STRING_MARSHALLER), entry.getValue());
+    }
+    return headersAsMetadataBuilder.build();
+  }
+
+  Map<Key<String>, String> getHeadersAsMetadata() {
+    return headersAsMetadata;
+  }
+
+  Map<Key<String>, String> getResourceHeadersAsMetadata(
```
vkedia: this can be a static method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerClientHeaderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.api.gax.rpc.HeaderProvider;
+import com.google.common.collect.ImmutableMap;
+import io.grpc.Metadata;
+import io.grpc.Metadata.Key;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+class SpannerClientHeaderProvider implements HeaderProvider {
```
vkedia: this implements `HeaderProvider` but I don't think `getHeaders` from this class is being called anywhere.
It might be better to just put constructHeadersAsMetadata and getResourceHeadersAsMetadata directly in GrpcSpannerRpc class

google-cloud-spanner/pom.xml
vam-google: This is related. This change also simplifies the way `x-goog-api-client` header is handled differently now (just gets data from `METADATA.MF` and the project.properties is not needed anymore, thus removing its generation (same cleanup was done for all other poms in this repo).

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
vam-google: Good catch. That is a dead code now (user can specify userAgentPrefix in the builder, but it will be ignored). I've removed the rest of the related code.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerRpc.java
vam-google: This is possible. I was considering this too. Moved to SpannerHeaderProvider

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerClientHeaderProvider.java
vam-google: It is easier to test it if it is a separate class. Also I would like to keep headers stuff isolated, this has a good chance to be useful during Spanner GA work, when (if?) it gets migrated to gapic stubs.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerClientHeaderProvider.java
vam-google: It can be, but I'm trying to make spanner headers as similar in structure to general gapic headers logic for consistency and easier migration go gapic stubs in the future.

vam-google:
@vkedia PTAL

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerClientHeaderProvider.java
vkedia: Ok, it can be a separate class but there is no need to implemented HeaderProvider for now. That can be added later if needed.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
vkedia: Can you give an example of how users would now specify customer user agent with spanner library? Ideally that should be prefixed to the default UA and not replace it completely.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerRpc.java
```diff
@@ -130,18 +125,19 @@ public GrpcSpannerRpc(SpannerOptions options) {
...
+    ApiClientHeaderProvider internalHeaderProvider =
+        internalHeaderProviderBuilder
+            .setClientLibToken(
+                ServiceOptions.getGoogApiClientLibName(),
+                GaxProperties.getLibraryVersion(options.getClass()))
+            .build();
+
+    HeaderProvider mergedHeaderProvider = options.getMergedHeaderProvider(internalHeaderProvider);
+    this.headerProvider = SpannerHeaderProvider.create(mergedHeaderProvider.getHeaders(),
+        internalHeaderProviderBuilder.getResourceHeaderKey());
```
vkedia: This key is a constant so I should be able to get to it by referring to a static constant rather than as a member method on ApiClientHeaderProvider. Ideally there should be no need to pass it to SpannerHeaderProvider. It should just refer to the constant somewhere.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerHeaderProvider.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.google.api.gax.rpc.HeaderProvider;
+import com.google.common.collect.ImmutableMap;
+import io.grpc.Metadata;
+import io.grpc.Metadata.Key;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+class SpannerHeaderProvider implements HeaderProvider {
```
vkedia: can  you add some class level doc.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerHeaderProvider.java
```diff
@@ -0,0 +1,81 @@
...
+    for (Pattern pattern : RESOURCE_TOKEN_PATTERNS) {
+      Matcher m = pattern.matcher(resourceTokenTemplate);
+      if (m.matches()) {
+        resourceToken = m.group("headerValue");
+        break;
+      }
+    }
+
+    return ImmutableMap.of(
+        Key.of(resourceHeaderKey, Metadata.ASCII_STRING_MARSHALLER), resourceToken);
```
vkedia: This key can be a constant.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerRpc.java
```diff
@@ -474,19 +470,17 @@ public void start(Listener<RespT> responseListener, Metadata metadata) {
...
-    metadata.put(API_CLIENT_KEY, xGoogApiClientHeader);
-    return metadata;
-  }
 
-  @VisibleForTesting
-  String extractHeader(String resource) {
-    Matcher m = DATABASE_PATTERN.matcher(resource);
-    if (m.matches()) {
-      return m.group("database");
+    Map<Metadata.Key<String>, String> headers = headerProvider.getHeadersAsMetadata();
```
vkedia: I think it might be cleaner if this whole logic is put inside SpannerHeaderProvider. It can have a method like
void populateHeaders(Metadata metadata, String resource, String project);

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
vam-google: It is prefixed. In fact grpc itself does not allow to completely replace User-Agent: https://github.com/grpc/grpc-java/blob/c9b02db276403db4794c6e5ffc78b46889cd4ce8/netty/src/main/java/io/grpc/netty/Utils.java#L113

As for the example: it is exactly same as the examples in the original comment to this PR (the manual clients section). The provided User-Agent will be merged, not completelly replaced. It is also documented in the `ServiceOptions.setHeaderProvider()` method.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerRpc.java
vam-google: On gapic level this value is treated as the one with default value, but not constant. In this particular case, the value keeps the default value.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerHeaderProvider.java
vam-google: This is a package-private class.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerHeaderProvider.java
vam-google: Converted `SpannerHeaderProvider` to `SpannerMetadataProvider`, which now has only one method `newMetadata()`, with everything encapsulated there.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GrpcSpannerRpc.java
vam-google: (same comment as above): Converted `SpannerHeaderProvider` to `SpannerMetadataProvider`, which now has only one method `newMetadata()`, with everything encapsulated there.

vam-google:
@garrettjonesgoogle PTAL

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
vkedia: Is there some common documentation (not Javadoc) somewhere that we can point our customers to on how to set these headers.

vam-google: No. I hope it should not be more difficult than it was before this change. If somebody tries to set header he/she would be searching for some header setter method in the `SpannerOptions.Builder` (same as it was before) and will find `setHeadersProvider()` method (instead of `setUserAgentPrefix()`). It would be slightly less straightforward than before, but I believe should be still ok.

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProviderTest.java
```diff
@@ -0,0 +1,76 @@
...
+import io.grpc.Metadata;
+import io.grpc.Metadata.Key;
+import java.util.Map;
+import org.junit.Test;
+
+public class SpannerMetadataProviderTest {
+  @Test
+  public void testGetHeadersAsMetadata() {
+    Map<String, String> headers = ImmutableMap.of("header1", "value1", "header2", "value2");
+    SpannerMetadataProvider headerProvider = SpannerMetadataProvider.create(headers, "header3");
```
garrettjonesgoogle: rename the variable to match the type (`metadataProvider`)

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProvider.java
```diff
@@ -0,0 +1,80 @@
...
+package com.google.cloud.spanner.spi.v1;
+
+import com.google.common.collect.ImmutableMap;
+import io.grpc.Metadata;
+import io.grpc.Metadata.Key;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+class SpannerMetadataProvider {
```
garrettjonesgoogle: Add a comment about this being package-private

google-cloud-examples/src/main/java/com/google/cloud/examples/logging/snippets/CreateAndListMetrics.java
```diff
@@ -35,7 +37,15 @@
...
     // Create a service object
     // Credentials are inferred from the environment
-    try(Logging logging = LoggingOptions.getDefaultInstance().getService()) {
+
+
+    HeaderProvider headerProvider =
+        FixedHeaderProvider.create("user-agent", "my-client-name", "custom-header", "custom stuff");
+
+    LoggingOptions options =
+        LoggingOptions.newBuilder().setHeaderProvider(headerProvider).build();
```
garrettjonesgoogle: revert

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/InsertDataAndQueryTable.java
```diff
@@ -53,7 +53,7 @@ public static void main(String... args) throws InterruptedException {
     BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
 
     // Create a dataset
-    String datasetId = "my_dataset_id";
+    String datasetId = "my_dataset_id_1";
```
garrettjonesgoogle: revert

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -535,40 +552,70 @@ public TransportOptions getTransportOptions() {
...
+   * Returns the application's name as a string in the format {@code gcloud-java/[version]},
+   * optionally prepended with externally supplied User-Agent header value (via setting custom
+   * header provider).
    */
   public String getApplicationName() {
     String libraryVersion = getLibraryVersion();
-    return libraryVersion == null ? LIBRARY_NAME : LIBRARY_NAME + "/" + libraryVersion;
-  }
 
+    // We have to do it, since underlying layers often do not appreciate User-Agent provided as a
```
garrettjonesgoogle: Better wording: "We have to do the following since ..."

google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProviderTest.java
vam-google: Done

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/SpannerMetadataProvider.java
vam-google: Done

google-cloud-examples/src/main/java/com/google/cloud/examples/logging/snippets/CreateAndListMetrics.java
vam-google: Done

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/InsertDataAndQueryTable.java
vam-google: Done

google-cloud-bom/pom.xml
```diff
@@ -14,8 +14,9 @@
     <api-client.version>1.23.0</api-client.version>
 
     <api-common.version>1.2.0</api-common.version>
-    <gax.version>1.15.0</gax.version>
-    <gax-grpc.version>1.15.0</gax-grpc.version>
+    <gax.version>1.15.1-SNAPSHOT</gax.version>
+    <gax-grpc.version>1.15.1-SNAPSHOT</gax-grpc.version>
+    <gax-httpjson.version>0.32.1-SNAPSHOT</gax-httpjson.version>
```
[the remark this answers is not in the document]
vam-google: Will do

vam-google:
@garrettjonesgoogle PTAL

vam-google:
@garrettjonesgoogle PTAL

PR #2697, opened 2017-12-12
 bigquery: allow schema-ful data listing

pongad:
listTableData and query now use the same classes.

### DIFF
--- a/google-cloud-examples/src/test/java/com/google/cloud/examples/bigquery/snippets/ITBigQuerySnippets.java
+++ b/google-cloud-examples/src/test/java/com/google/cloud/examples/bigquery/snippets/ITBigQuerySnippets.java
@@ -36,7 +36,6 @@ import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobId;
 import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.QueryResponse;
-import com.google.cloud.bigquery.QueryResult;
 import com.google.cloud.bigquery.Schema;
 import com.google.cloud.bigquery.JobStatus;
 import com.google.cloud.bigquery.StandardTableDefinition;


tswast:
Looks like it's missing an update to the integration tests.

```
[ERROR] /home/travis/build/GoogleCloudPlatform/google-cloud-java/google-cloud-examples/src/test/java/com/google/cloud/examples/bigquery/snippets/ITBigQuerySnippets.java:[39,33] cannot find symbol

[ERROR]   symbol:   class QueryResult
[ERROR]   location: package com.google.cloud.bigquery
```

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -445,46 +446,72 @@ public TableDataInsertAllResponse call() throws Exception {
...
+  public TableResult listTableData(
+      String datasetId, String tableId, TableDataListOption... options) {
+    return listTableData(TableId.of(datasetId, tableId), options);
   }
 
   @Override
-  public Page<FieldValueList> listTableData(TableId tableId, TableDataListOption... options) {
-    return listTableData(tableId, getOptions(), optionMap(options));
+  public TableResult listTableData(TableId tableId, TableDataListOption... options) {
+    return listTableData(tableId, null, options);
```
tswast: I'm having trouble following which overloaded method this is going to. If it's the one above, it seems to be using `tableId` as the `datasetId`.

I suppose it's intended that the Schema is null, but I'm not sure how Java knows that.

Also, what was BigQueryOptions for and why is it going away?

pongad: This is calling overload on line 466.
The first argument has type `TableId` which disqualifies L449 and L460, so L466 is the only one left with the right number of args.

The `BigQueryOptions` isn't going away, it's just being populated on L468.

I think this means we have too many overloads, but I'm not sure if we should require the user to pass null schema.

pongad:
Build failure fixed now. Sorry, I keep missing them.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
tswast: Thanks for clarifying. Yeah, too many overloads. But we shouldn't force them to explicitly pass in null when they don't know the schema.

PR #2700, opened 2017-12-12
Bump versions to SNAPSHOT

andreamlin:


google-cloud-bom/pom.xml
```diff
@@ -138,36 +138,36 @@
...
-    <container.version>0.32.0-beta</container.version>
-    <contrib.version>0.32.0-alpha</contrib.version>
+    <cloud.version>0.32.1-alpha-SNAPSHOT</cloud.version>
+    <bigtable.version>0.32.1-beta-SNAPSHOT</bigtable.version>
+    <bigquery.version>0.32.1-beta-SNAPSHOT</bigquery.version>
+    <bigquerydatatransfer.version>0.32.1-beta-SNAPSHOT</bigquerydatatransfer.version>
+    <compute.version>0.32.1-alpha-SNAPSHOT</compute.version>
+    <container.version>0.32.1-beta-SNAPSHOT</container.version>
+    <contrib.version>0.32.1-alpha-SNAPSHOT</contrib.version>
     <core.version>1.14.0</core.version>
```
garrettjonesgoogle: Why aren't these versions getting bumped?

andreamlin: Fixed.

andreamlin:
PTAL

garrettjonesgoogle:
If you have verified by grep that 1.14.0 and 0.32.0 have been updated to the SNAPSHOT versions in all pom locations, then LGTM

PR #2703, opened 2017-12-12
update versions in bom pom

andreamlin:
Add a line in update_versions.sh to update the versions properties in the bom pom.xml


andreamlin:
PTAL

garrettjonesgoogle:
@vam-google could you do this review?

vam-google:
This will also require change renaming `google-cloud-bom.version` property to just `bom.property` to fix the script completelly. I'll do it in a separate PR

PR #2705, opened 2017-12-12
Core: Resolving Project Id

yihanzhen:
Check if application is on GAE before trying to resolve ProjectId from GAEAppId

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -388,6 +388,9 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>,
 
   protected static String getAppEngineProjectId() {
     String projectId = null;
+    if (!PlatformInformation.isOnGAE()) {
+      return projectId;
+    }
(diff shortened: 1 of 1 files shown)


hzyi-google:
PTAL @neozwu @pongad

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -388,6 +388,9 @@ private static boolean isWindows() {
 
   protected static String getAppEngineProjectId() {
     String projectId = null;
+    if (!PlatformInformation.isOnGAE()) {
```
neozwu: Can you check whether `PlatformInformation.isOnGAE()` works for GAE flex environ?

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -388,6 +388,9 @@ private static boolean isWindows() {
 
   protected static String getAppEngineProjectId() {
```
pongad: @neozwu Correct me if I'm wrong. I think some of these checks also work on GCE and GKE right?

PR #2713, opened 2017-12-15
bigquery: proper handle options to Job.getQueryResults

pongad:


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -272,13 +275,40 @@ public TableResult getQueryResults(QueryResultsOption... options)
     }
 
     TableId table = ((QueryJobConfiguration) getConfiguration()).getDestinationTable();
-    // TODO(pongad): merge options?
+
+    List<QueryResultsOption> waitOptions = new ArrayList<>();
+    waitOptions.add(QueryResultsOption.pageSize(0L));
+    List<TableDataListOption> listOptions = new ArrayList<>();
```
pongad: I'm somewhat concerned by this. If/When we eventually move to streaming RPC, most of these options will be useless.

tswast: That's going to be a problem across languages. We'll figure out how to transition if/when we get a streaming RPC API for this.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -272,13 +275,40 @@ public TableResult getQueryResults(QueryResultsOption... options)
...
+          listOptions.add(TableDataListOption.startIndex((Long) option.getValue()));
+          break;
+        case TIMEOUT:
+          hasWaitTime = true;
+          waitOptions.add(QueryResultsOption.maxWaitTime((Long) option.getValue()));
+          break;
+      }
+    }
+    if (!hasWaitTime) {
+      waitOptions.add(QueryResultsOption.maxWaitTime(Duration.ofMinutes(1).toMillis()));
```
tswast: I think the API default is 10 minutes. Is there a reason to explicitly set it if it's not provided?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
pongad: SGTM

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
pongad: I did it to keep existing behavior, but you're right that it seems odd to second-guess the API. Removed.

PR #2715, opened 2017-12-18
pubsub: respect total expiration while modacking

pongad:
Previously we use total expiration only to decide whether to send
new modacks. If we send modacks close to the expiration, we might
extend the lease beyond the total expiration time.

This commit shortens the lease when we approach the expiration time
so we never go over.

@mdietz94

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
@@ -432,33 +432,35 @@ class MessageDispatcher {
   @InternalApi
   void extendDeadlines() {
     int extendSeconds = getMessageDeadlineSeconds();
+    List<PendingModifyAckDeadline> modacks = new ArrayList<>();
     PendingModifyAckDeadline modack = new PendingModifyAckDeadline(extendSeconds);
-    ArrayListMultimap<Long, String> expiringModack = ArrayListMultimap.create();
     Instant now = now();
     Instant extendTo = now.plusSeconds(extendSeconds);
 
+    int count = 0;
     Iterator<Map.Entry<String, Instant>> it = pendingMessages.entrySet().iterator();
     while (it.hasNext()) {
       Map.Entry<String, Instant> entry = it.next();
       String ackId = entry.getKey();
       Instant totalExpiration = entry.getValue();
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -426,21 +431,37 @@ int computeDeadlineSeconds() {
...
+      String ackId = entry.getKey();
+      Instant totalExpiration = entry.getValue();
+      if (totalExpiration.isAfter(extendTo)) {
+        // Most extensions won't reach total expiration, so avoid map lookup.
+        modack.ackIds.add(ackId);
+        continue;
       }
+      it.remove();
+      if (totalExpiration.isAfter(now)) {
+        expiringModack.put(now.until(totalExpiration, ChronoUnit.SECONDS), ackId);
```
mdietz94: Maybe we should have some threshold here, if it's 10 milliseconds, we are going to nack the message, which doesn't seem quite right. We should probably mod ack either the max of 1 second and this value, or just ignore it if it is going to expire within a second anyway.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -426,21 +431,37 @@ int computeDeadlineSeconds() {
 
   @InternalApi
   void extendDeadlines() {
-    List<String> acksToSend = Collections.<String>emptyList();
-    PendingModifyAckDeadline modack = new PendingModifyAckDeadline(getMessageDeadlineSeconds());
+    int extendSeconds = getMessageDeadlineSeconds();
+    PendingModifyAckDeadline modack = new PendingModifyAckDeadline(extendSeconds);
+    ArrayListMultimap<Long, String> expiringModack = ArrayListMultimap.create();
```
mdietz94: Why use this multimap instead of just adding these to the modacks list directly?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java

pongad:
@mdietz94 PTAL

pongad:
@mdietz94 Ping

mdietz94:
Sorry this got lost in my inbox while I was OOO

pongad:
No problem!

PR #2729, opened 2017-12-20
Census tracing integration for Firestore.

danchia:
@bogdandrutu PTAL. The dependency management was in particular quite nasty, and ideas / feedback around that area greatly appreciated.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -247,6 +247,8 @@ class FirestoreImpl implements Firestore {
       final Transaction.Function<T> transactionCallback,
       final SettableApiFuture<T> resultFuture,
       final TransactionOptions options) {
+    // span is intentionally not ended here. It will be ended by runTransactionAttempt on success
+    // or error.
     Span span = tracer.spanBuilder("CloudFirestore.Transaction").startSpan();
     try (Scope s = tracer.withSpan(span)) {
       runTransactionAttempt(transactionCallback, resultFuture, options, span);
@@ -339,7 +341,6 @@ class FirestoreImpl implements Firestore {
                   span);
             } else {
               span.setStatus(TOO_MANY_RETRIES_STATUS);
-              span.end();
               rejectTransaction(
                   FirestoreException.serverRejected(
                       Status.ABORTED, "Transaction was cancelled because of too many retries."));
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/TraceUtil.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/TraceUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +21,7 @@ import io.opencensus.contrib.grpc.util.StatusConverter;
 import io.opencensus.trace.Status;
 
 /** Census tracing utilities. */
-class TraceUtil {
+final class TraceUtil {
   private TraceUtil() {}
 
   public static Status statusFromApiException(ApiException exception) {


google-cloud-firestore/pom.xml
```diff
@@ -63,6 +75,40 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-auth</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-api</artifactId>
+      <version>0.10.1</version>
```
garrettjonesgoogle: We should use a single opencensus version from google-cloud-bom.

vkedia: Yeah I just added it there while integrating tracing into spanner. You will not be able to use 0.10.1 since it depends on grpc-context 1.8.0 while we depend on grpc-context 1.7.0.

danchia: Ah, I see Cloud Spanner just added one. I will incorporate that change in.

vkedia: Ah I see you excluded grpc-context. Ok you might need to do the same in google-cloud-spanner/pom.xml if you increasing the version of opencensus in google-cloud-bom/pom.xml

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TraceUtil.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.google.cloud.firestore;
+
+import com.google.api.gax.rpc.ApiException;
+import io.opencensus.contrib.grpc.util.StatusConverter;
+import io.opencensus.trace.Status;
+
+/** Census tracing utilities. */
+class TraceUtil {
```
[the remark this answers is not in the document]
danchia: I'm not quite sure what else I should do here, advice appreciated.

danchia:
I updated the PR to use census from BOM instead.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -196,11 +206,24 @@ public DocumentReference document(@Nonnull String documentPath) {
...
+          int numResponses;
+
           @Override
           public void onNext(BatchGetDocumentsResponse response) {
             DocumentReference documentReference;
             DocumentSnapshot documentSnapshot;
 
+            numResponses++;
+            if (numResponses == 1) {
+              Tracing.getTracer()
```
bogdandrutu: why do you not use the tracer here? same comment everywhere.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -59,6 +67,8 @@
   private static final String AUTO_ID_ALPHABET =
       "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
 
+  public static final Tracer tracer = Tracing.getTracer();
```
bogdandrutu: why public?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -349,19 +398,28 @@ public void run() {
...
+              runTransactionAttempt(
                   transactionCallback,
                   resultFuture,
                   new TransactionOptions(
-                      attemptsRemaining, options.getExecutor(), transaction.getTransactionId()));
+                      attemptsRemaining, options.getExecutor(), transaction.getTransactionId()),
+                  span);
             } else {
+              span.setStatus(
+                  io.opencensus.trace.Status.ABORTED.withDescription("too many retries"));
```
bogdandrutu: you can have the status with your description as static private to avoid allocations.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -231,11 +254,13 @@ public void onNext(BatchGetDocumentsResponse response) {
 
           @Override
           public void onError(Throwable throwable) {
+            Tracing.getTracer().getCurrentSpan().addAnnotation("Firestore.BatchGet: Error");
```
bogdandrutu: Maybe set an error status.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -196,11 +206,24 @@ public DocumentReference document(@Nonnull String documentPath) {
...
+
           @Override
           public void onNext(BatchGetDocumentsResponse response) {
             DocumentReference documentReference;
             DocumentSnapshot documentSnapshot;
 
+            numResponses++;
+            if (numResponses == 1) {
+              Tracing.getTracer()
+                  .getCurrentSpan()
```
bogdandrutu: I couldn't determine if the context is propagated to this point from the "runTransaction". Is the onNext executed in the same thread as runTransaction?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TraceUtil.java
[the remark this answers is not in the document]
bogdandrutu: Also make the class final.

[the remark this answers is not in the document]
bogdandrutu: You did not make this class final. Or you did not push the change.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TraceUtil.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
```
bogdandrutu: You need probably to update all years.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -213,11 +247,26 @@ public void onCompleted() {
       final Transaction.Function<T> transactionCallback,
       final SettableApiFuture<T> resultFuture,
       final TransactionOptions options) {
+    Span span = tracer.spanBuilder("CloudFirestore.Transaction").startSpan();
```
bogdandrutu: Please comment that the "span" is intentionally not ended here and that it will be ended if an error occurs or success.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -279,19 +330,27 @@ public void run() {
...
+              runTransactionAttempt(
                   transactionCallback,
                   resultFuture,
                   new TransactionOptions(
-                      attemptsRemaining, options.getExecutor(), transaction.getTransactionId()));
+                      attemptsRemaining, options.getExecutor(), transaction.getTransactionId()),
+                  span);
             } else {
+              span.setStatus(TOO_MANY_RETRIES_STATUS);
+              span.end();
```
bogdandrutu: Don't need to call end here because you call in the rejectTransaction(). Or if you want to have this specific status and end here probably better to pass a boolean to rejectTransaction to end or not the Span.

The call of the setStatus will be ignored if happens after end, also the second end is a no-op. But better to not do that in my opinion to make the code look better.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -126,11 +138,24 @@ public DocumentReference document(@Nonnull String documentPath) {
 
     ApiStreamObserver<BatchGetDocumentsResponse> responseObserver =
         new ApiStreamObserver<BatchGetDocumentsResponse>() {
+          int numResponses;
```
bogdandrutu: Is onNext guaranteed to run on one thread for all responses? If not you probably need to make this atomic to make sure the increment is correct + the latest value is visible to other threads.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
danchia: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
danchia: I removed the `span.end()` here. I left the status, since I think it's useful to preserve "too many retries" information.
Status shouldn't be double set by rejectTransaction since FirestoreException is not an instanceof ApiException.

danchia:
@pongad do you have any other feedback?

pongad:
@danchia Sorry for the late response. I was transferring and dropped a bunch of things. This LGTM. @garrettjonesgoogle Do you have any other comments?

PR #2741, opened 2018-01-03
logging: flowcontrol should block

### LINKED ISSUE #2714: Default FlowControlSettings LimitExceededBehavior
Hi I'm having an issue with my code, I recently started getting
com.google.api.gax.batching.FlowController$FlowControlRuntimeException: The maximum number of batch bytes: 10485760 have been reached.

I can see that's the default behaviour for LoggingOptions.getDefaultInstance().getService(); but the documentation isn't very clear on why its that way or what the alternatives are.
I can see from com.google.api.gax.batching.FlowControlSettings that there are 2 other alternatives Block and Ignore.

I think it would be helpful if somewhere in the documentation these options and their implications were explained and also how to switch between them.

I'd be interested to know if I can switch to Block and what that would mean for my program.

My program is pulling data from an activeMQ queue, logging it to google and then further processing it.
I have to restart the program or else the error continues on indefinitely. Could this error be caused by a message from the queue being above the limit? and if so if set to block will it process the message anyway?

How should this exception be handled so that I can log the data?

Thanks

pongad:
Fixes #2714.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java
@@ -20,7 +20,6 @@ import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
 import com.google.api.gax.batching.BatchingSettings;
-import com.google.api.gax.batching.FlowControlSettings;
 import com.google.api.gax.batching.FlowController.LimitExceededBehavior;
 import com.google.api.gax.core.BackgroundResource;
 import com.google.api.gax.core.GaxProperties;
@@ -75,7 +74,6 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Set;
 import java.util.concurrent.ScheduledExecutorService;
-import org.threeten.bp.Duration;
 
 public class GrpcLoggingRpc implements LoggingRpc {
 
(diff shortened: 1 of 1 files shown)


google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java
```diff
@@ -138,6 +142,22 @@ public Void apply(UnaryCallSettings.Builder<?, ?> builder) {
...
+                  .setElementCountThreshold(1000L)
+                  .setRequestByteThreshold(1048576L)
+                  .setDelayThreshold(Duration.ofMillis(50))
+                  .setFlowControlSettings(
+                      FlowControlSettings.newBuilder()
+                          .setMaxOutstandingElementCount(100000L)
+                          .setMaxOutstandingRequestBytes(10485760L)
+                          .setLimitExceededBehavior(LimitExceededBehavior.Block)
+                          .build())
+                  .build());
```
garrettjonesgoogle: Is there no way to get the default batching settings and only change the LimiteExceededBehavior? This basically overrides whatever is in the GAPIC config for logging.

pongad: It's terrible but possible. Good point though.

pongad:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #2742, opened 2018-01-03
bigquery: properly detect statistics type

### LINKED ISSUE #2357: Invalid Statistics type for Load job
Hi,
  I have following code snippet in my project.
 `Credentials c = new ServiceAccountCredentials();  // hiding credentials here
  BigQueryOptions options = BigQueryOptions.newBuilder().setCredentials(c).setProjectId('id').build();
  BigQuery bq = options.getService();
  TableId id = TableId.of('project id', 'dataset id', 'table id');
  Table t = bq.getTable(id);
  t.load(FormatOptions.csv(), "gs://bucket/file1.csv");
  System.out.println(loadJob.getConfiguration().getType().name());`

I was expecting the job to be of 'Load' type but its of 'Copy' type.

Can anyone tell me where am I doing wrong?  Indeed, data is getting inserted into table but I am missing the output statistics (no of rows, no of bytes, etc) due to wrong type being returned.

pongad:
Fixes #2357.

If the job is incomplete, the statistics itself doesn't give us enough
info about what kind of job we're seeing.
However, the configuration should, since the request is required to
populate it.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
@@ -104,7 +104,9 @@ public abstract class JobStatistics implements Serializable {
 
       private Builder(com.google.api.services.bigquery.model.JobStatistics statisticsPb) {
         super(statisticsPb);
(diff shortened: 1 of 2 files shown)


pongad:
Done. This actually caught NPE bugs. I don't like nulls :(

PR #2743, opened 2018-01-03
bigquery: add setUseLegacySql to ViewDefinition

### LINKED ISSUE #2720: bigquery: setUseLegacySql missing in java API
Original bug created here: https://issuetracker.google.com/issues/70662064

```
The new Java API does not offer the setUseLegacySql() method that had the old API.
That means we can only create views in LegacySQL.

Could you please add this method back to the new API?
```

pongad:
Fixes #2720.


tswast:
The test failure seems to be a problem loading plugins.

pongad:
SGTM. I made https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2748 so I don't forget.

PR #2750, opened 2018-01-04
pubsub: document Subscriber running forever

### LINKED ISSUE #2733: Pub/Sub Subscriber service that should never stop listening for new messages
Hi,

I think I've now looked at all Pub/Sub Java client examples in the official docs and seems all of them have an idea of stopping Subscriber service at some point with `subscriber.stopAsync();` call. But what if I want to implement a service that should listen for Pub/Sub messages on a subscription and never exit. What would be a recommended way of doing it?
Maybe you can explain a common idea or provide such an example here and also add it to the docs?
In case there is already such an example, could you please point me to it.

pongad:
Fixes #2733.


dzlier-gcp:
LGTM

PR #2754, opened 2018-01-06
Refactor

ashutoshcp:
1. typo fix
2. made values as variable
3. used simpler method for string concat
4. assert, so that NPE can be avoided


ashu01:
I signed it!

pongad:
@ashu01 This seems to fail `RemoteBigQueryHelperTest`. I think this PR changes `maxRetryDelay` from 30 to 60 seconds. Could you edit this?

The rest of the PR looks good to me. @garrettjonesgoogle PTAL.

ashu01:
sure, will do this by tomorrow.

PR #2760, opened 2018-01-09
logging: clear null trace_id

### LINKED ISSUE #2746: Add option to clear traceId thread-local in TraceLoggingEnhancer
When the `traceId` thread-local needs to be removed, calling `setCurrentTraceId(null)` is not appropriate because it still leaves a thread-local entry. Instead, and indirect call to `ThreadLocal.remove()` should be exposed. One option is to provide a method called `removeCurrentTraceId`. Another alternative, is to detect the `null` argument to `setCurrentTraceId` and call `remove()` instead of `set(null)` in that case.

pongad:
To ThreadLocal, set(null) and remove() are functionally similar but
remove() clears out the map-entry so it's more space-efficient.

Fixes #2746.


jabubake:
LGTM

PR #2761, opened 2018-01-09
add grpc-testing to bom

igorbernstein2:
GrpcServerRule is very useful when writing clients. By including it in the bom, it makes it possible to keep in sync with the target grpc version


garrettjonesgoogle:
LGTM

PR #2764, opened 2018-01-10
Fix typo in sample code (in javadoc)

lama0206:
This pull request fixes a minor error in sample code contained in the javadoc of the v1/SpeechClient. Since this is generated, you might consider fixing it in the original source...


pongad:
@lama0206 Thank you for the PR. Since this is in a generated file, I think we shouldn't fix it here since the change will get overwritten the next generation-cycle. I have filed https://github.com/googleapis/toolkit/issues/1829 to track this issue.

I'll close this PR in favor of fixing it at the root cause.

lama0206:
👍

PR #2765, opened 2018-01-11
logback: log stack trace

### LINKED ISSUE #2735: Exceptions logged with Logback + SLF4J don't show up in Stackdriver
I'm using the Logback + SLF4J logging configuration described in the official documentation [here](https://cloud.google.com/logging/docs/setup/java).
Everything works fine with simple log messages like:
```java
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
...
private static final Logger logger = LoggerFactory.getLogger(MyClass.class);
...
logger.info("Hello world");
```

The problem appears when I try to log exceptions like this:
```java
logger.error("Something went wrong:", ex);
```

I see only my description message in Stackdriver but not the stack trace from the provided exception.
What can be the problem here?

pongad:
Fixes #2735.

### DIFF
--- a/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
+++ b/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
@@ -230,7 +230,7 @@ public class LoggingAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
     return builder.build();
   }
 
-  @InternalApi
+  @InternalApi("Visible for testing")
   static void writeStack(IThrowableProxy throwProxy, String prefix, StringBuilder payload) {
     if (throwProxy == null) {
       return;


google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
```diff
@@ -206,10 +206,15 @@ Logging getLogging() {
   }
 
   private LogEntry logEntryFor(ILoggingEvent e) {
-    String payload = e.getFormattedMessage();
+    StringBuilder payload = new StringBuilder(e.getFormattedMessage());
+    for (StackTraceElement stackEl : e.getCallerData()) {
```
jabubake: is `e.getCallerData()` null safe ?

garrettjonesgoogle: `element` or `stackElement` would be more idiomatic for the variable name

pongad: Done

pongad: It seems to be, but the documentation never explicitly stated this. I'll add a check just in case.

pongad:
@jabubake I'm sorry. Looking at this closer, I don't think this is right. The `getCallerData` returns the trace of the running thread, not of the exception. I don't see a way to get exception trace out of `ILoggingEvent`. Can you advise?

jabubake:
`ILoggingEvent` has a `getThrowableProxy` method.
see examples here : http://www.javased.com/?api=ch.qos.logback.classic.spi.IThrowableProxy

pongad:
@garrettjonesgoogle @jabubake PTAL. I think this actually works now :)

pongad:
@garrettjonesgoogle @jabubake Ping

google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
```diff
@@ -231,6 +230,34 @@ private LogEntry logEntryFor(ILoggingEvent e) {
     return builder.build();
   }
 
+  @InternalApi
```
garrettjonesgoogle: Add text to the annotation ("Visible for testing")

google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
```diff
@@ -231,6 +230,34 @@ private LogEntry logEntryFor(ILoggingEvent e) {
...
+    if (throwProxy == null) {
+      return;
+    }
+    payload
+        .append(prefix)
+        .append(throwProxy.getClassName())
+        .append(": ")
+        .append(throwProxy.getMessage())
+        .append('\n');
+    StackTraceElementProxy[] trace = throwProxy.getStackTraceElementProxyArray();
```
garrettjonesgoogle: Just as a sanity check, do we know that this is a stable API that we are using? (Or maybe it doesn't matter since google-cloud-logging-logback is alpha...)

pongad: I believe it is stable. We depend on version `1.2.3`.

google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
pongad: Done

PR #2766, opened 2018-01-11
bigquery: use standard SQL by default

### LINKED ISSUE #2748: make standard sql the default

pongad:
Fixes #2748.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ViewDefinition.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ViewDefinition.java
@@ -44,7 +44,7 @@ public class ViewDefinition extends TableDefinition {
 
     private String query;
     private List<UserDefinedFunction> userDefinedFunctions;
-    private Boolean useLegacySql;
+    private Boolean useLegacySql = false;
 
     private Builder() {
       super(Type.VIEW);
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/ViewDefinitionTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/ViewDefinitionTest.java
@@ -17,6 +17,7 @@
 package com.google.cloud.bigquery;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ViewDefinition.java
```diff
@@ -44,7 +44,7 @@
 
     private String query;
     private List<UserDefinedFunction> userDefinedFunctions;
-    private Boolean useLegacySql;
+    private Boolean useLegacySql = false;
```
vam-google: It seems like `null` is supported (handled in `useLegacySql()` method properly and allowed on `setUseLegacySql()` method). In such cases there is no need to set this to `false`, default `null` value should be fine and is cleaner.

This comment is applicable only if the other one (Line 204) remains unaddressed (otherwise this must be set to `false`, because interpretation of `null` will be changed).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ViewDefinition.java
```diff
@@ -201,7 +201,11 @@ Table toPb() {
       viewDefinition.setUserDefinedFunctionResources(Lists.transform(userDefinedFunctions,
           UserDefinedFunction.TO_PB_FUNCTION));
     }
-    viewDefinition.setUseLegacySql(useLegacySql);
+    if (useLegacySql == null) {
```
vam-google: This changes the interpretation of `null`. After this change on Apiary level `null` for a flag named `useLegacySql` will be still interpreted as "use legacy SQL", but on veneer level it will be "use standard SQL". This is a type of confusion which can be and should be avoided (as it may lead to stupid bugs in the future). 

Please strongly consider just having `private Boolean useLegacySql = false;` and keeping everything else as is. This should allow having standard SQL by default, but if somebody wants to change it, the behavior will be consistent with the rest.

pongad:
Makes sense. Only default to `false` it is.

pongad:
@tswast `QueryJobConfiguration` already defaults to false. I didn't realize this when I updated `View`.

PR #2767, opened 2018-01-11
Pulling in the latest version of the Conformance Tests

schmidt-sebastian:
This:
- Updates to the latest version of the conformance tests
- Rejects deep deletes for set(mergePaths:..) if the delete is not part of the list of paths
- Uses the test suite file rather than the binprotos file and gets rid of varint parsing


garrettjonesgoogle:
Go ahead and merge given the LGTMs of @jba and @pongad

PR #2770, opened 2018-01-11
Update google-cloud-datastore proto driver to v1.6.0.

wsh:
Removes dependency on old proto artifact for improved compatibility.


garrettjonesgoogle:
LGTM

PR #2787, opened 2018-01-18
date CHANGES.md

pongad:


CHANGES.md
```diff
@@ -1,4 +1,4 @@
-# RELEASE DATE HERE
+# 2018/01/18
```
garrettjonesgoogle: This should be keyed to releases I think, not dates. If someone depending on google-cloud-java is choosing a version to depend on, they have no idea what date they were released on.

pongad: good point; done

PR #2792, opened 2018-01-18
Stub settings (do not check in)

andreamlin:
Pubsub sample of stub settings class.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
@@ -20,13 +20,21 @@ import static com.google.cloud.pubsub.v1.PagedResponseWrappers.ListTopicsPagedRe
 
 import com.google.api.core.ApiFunction;
 import com.google.api.core.BetaApi;
+import com.google.api.gax.core.GaxProperties;
+import com.google.api.gax.core.GoogleCredentialsProvider;
+import com.google.api.gax.core.InstantiatingExecutorProvider;
+import com.google.api.gax.grpc.GaxGrpcProperties;
+import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.api.gax.rpc.ApiClientHeaderProvider;
 import com.google.api.gax.rpc.BatchingCallSettings;
 import com.google.api.gax.rpc.ClientContext;
 import com.google.api.gax.rpc.ClientSettings;
 import com.google.api.gax.rpc.PagedCallSettings;
+import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.api.gax.rpc.UnaryCallSettings;
 import com.google.cloud.pubsub.v1.stub.PublisherStub;
 import com.google.cloud.pubsub.v1.stub.PublisherStubSettings;
+import com.google.common.collect.ImmutableList;
 import com.google.iam.v1.GetIamPolicyRequest;
 import com.google.iam.v1.Policy;
 import com.google.iam.v1.SetIamPolicyRequest;
@@ -44,6 +52,7 @@ import com.google.pubsub.v1.PublishResponse;
 import com.google.pubsub.v1.Topic;
 import com.google.pubsub.v1.UpdateTopicRequest;
 import java.io.IOException;
+import java.util.List;
 import javax.annotation.Generated;
 
 // AUTO-GENERATED DOCUMENTATION AND CLASS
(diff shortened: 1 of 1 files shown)


andreamlin:
PTAL
I'm not sure if TopicAdminSettings is correct, now that it is refactored into a wrapper over the contained PublisherStubSettings that does literally nothing else.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
```diff
@@ -106,131 +78,74 @@
...
   /** Returns the object with the settings used for calls to updateTopic. */
   public UnaryCallSettings<UpdateTopicRequest, Topic> updateTopicSettings() {
-    return updateTopicSettings;
+    return stubSettings.updateTopicSettings();
   }
 
   /** Returns the object with the settings used for calls to publish. */
   public BatchingCallSettings<PublishRequest, PublishResponse> publishSettings() {
-    return publishSettings;
+    return stubSettings.publishSettings();
```
garrettjonesgoogle: Any surface for the `publish` API should be hidden on `TopicAdminSettings`. In the code generator, this should be driven by the surface_treatments.visibility config ( https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub_gapic.yaml#L515 ).

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
```diff
@@ -526,147 +179,54 @@ protected Builder() {
...
-      builder.setTransportChannelProvider(defaultTransportChannelProvider());
-      builder.setCredentialsProvider(defaultCredentialsProviderBuilder().build());
-      builder.setInternalHeaderProvider(defaultApiClientHeaderProviderBuilder().build());
-      builder.setEndpoint(getDefaultEndpoint());
-      return initDefaults(builder);
+    @Override
+    public Builder setTransportChannelProvider(TransportChannelProvider channelProvider) {
+      super.setTransportChannelProvider(channelProvider);
+      stubBuilder.setTransportChannelProvider(channelProvider);
+      return this;
```
garrettjonesgoogle: This makes a confusing situation where it's unclear which object holds the providers that will be used at client construction time. Also, I want to avoid generating general methods at the API-specific level. This probably means we need to support composition at the Gax level itself: we should probably have the ClientSettings of the TopicAdminSettings object hold the PublisherStubSettings object instead of TopicAdminSettings holding it.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
andreamlin: Does that mean that this method should be package-private, as specified in that gapic config?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
andreamlin: That is a problem for sure. So the ClientSettings should contain another ClientSettings field? That seems confusing too.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
garrettjonesgoogle: Yes.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
garrettjonesgoogle: Ok I think the rabbit hole is going slightly deeper - it appears that we need to create a StubSettings class and move the ClientSettings implementation into it, while leaving the ClientSettings surface itself intact. (This is just a parallel change to the API-specific classes.)

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
andreamlin: Ok, made this method and the builder method package private.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
andreamlin: lol the things we do for abstraction
gax-java PR at https://github.com/googleapis/gax-java/pull/459

still updating this PR to reflect gax changes...

andreamlin:
The pubsub tests pass when using the version of gax in https://github.com/googleapis/gax-java/pull/459

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
```diff
@@ -106,131 +75,69 @@
...
-        .equals(GrpcTransportChannel.getGrpcTransportName())) {
-      return GrpcPublisherStub.create(this);
-    } else {
-      throw new UnsupportedOperationException(
-          "Transport not supported: " + getTransportChannelProvider().getTransportName());
-    }
-  }
-
-  /** Returns a builder for the default ExecutorProvider for this service. */
-  public static InstantiatingExecutorProvider.Builder defaultExecutorProviderBuilder() {
```
andreamlin: Can a TopicAdminSettings take in any StubSettings, or does it only take in a PublisherStubSettings? I'm not sure if these public static methods should remain in this class or the stub settings class.

garrettjonesgoogle: Only PublisherStubSettings. 

I think they should remain here - they are part of the surface of settings classes for GA clients.

andreamlin: Ok, left the public static methods in TopicAdminSettings. However, deferred implementation of the methods to PublisherStubSettings, because the stub settings object is responsible for setting those default fields in its own builder.

garrettjonesgoogle:
I would say after you address the feedback I had on OperationsSettings in the gax-java review, you can go ahead with updating toolkit to generate this.

PR #2798, opened 2018-01-22
bigquery: support null values in inserts

### LINKED ISSUE #2505: How to support insert NULL values in bigquery.insertAll
I followed the official example of [Streaming insert examples](https://cloud.google.com/bigquery/streaming-data-into-bigquery), and came across with a problem:
the _insertAll_ command, does not accept **a null value** in the Map I pass it.

**Code example:**
        Map<String, Object> rowContent = new HashMap<>();
        rowContent.put("fname","David");
        rowContent.put("lname",null); //here's the problem

        TableId tableId = TableId.of(dataSetName, tableName);
        InsertAllResponse response = bigquery.insertAll(InsertAllRequest.newBuilder(tableId)
                .addRow( rowContent)
                .build());

        // If any of the insertions failed, this lets you inspect the errors
        if (response.hasErrors()) {
            for (Map.Entry<Long, List<BigQueryError>> entry : response.getInsertErrors().entrySet()) {
                logger.error(entry.toString());
            }
        }

I investigated the libraries the code is using, and I saw there're several checks regarding NULL. 
For example:

@GwtCompatible
final class CollectPreconditions {
    CollectPreconditions() {
    }

    static void checkEntryNotNull(Object key, Object value) {
        if (key == null) {
            throw new NullPointerException("null key in entry: null=" + value);
        } else if (value == null) {
            throw new NullPointerException("null value in entry: " + key + "=null");
        }
    }
...
}

**Question:**
How/What can can I do in order to insert records, which some of the values might be NULL.
**I'm using the Java client.

pongad:
ImmutableMap does not support null values.
This plays strangely with BigQuery which supports null values in some
fields.

Fixes #2505.

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/InsertAllRequestTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/InsertAllRequestTest.java
@@ -20,10 +20,10 @@ import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.HashMap;
-import java.util.Collections;
 import org.junit.Test;
 
 public class InsertAllRequestTest {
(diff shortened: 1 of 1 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/InsertAllRequestTest.java
```diff
@@ -215,14 +212,25 @@ public void testEquals() {
...
+  @Test(expected=UnsupportedOperationException.class)
+  public void testImmutable() {
+    InsertAllRequest.RowToInsert row = InsertAllRequest.RowToInsert.of(new HashMap<String,Object>());
+    row.getContent().put("zip", "zap");
+  }
+
+  @Test
+  public void testNullOK() {
+    InsertAllRequest.RowToInsert.of(Collections.singletonMap("foo", null));
+  }
```
garrettjonesgoogle: This tests that nulls are ok, but I'm not seeing a test that verifies that the inserted row is null (or at least that whatever is stored or passed along is null).

pongad:
@garrettjonesgoogle PTAL

PR #2819, opened 2018-01-30
compute: deprecate

### LINKED ISSUE #2732: Is google-cloud-java GCE misrepresented in root README
From b/70926158

In the README for GCE:

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-compute/README.md

"Note: This client is no longer receiving updates; new features in the Compute API will not be added to this client. Check https://cloud.google.com/compute/docs/api/libraries for the recommended Java client library to use for accessing Compute."

Yet this is listed in the main README as having "alpha" support:

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/README.md

"This client supports the following Google Cloud Platform services at an Alpha quality level:
- Cloud Compute (Alpha)"

This seems misleading, and that the GCE component should be in a separate section, or marked somehow at this level, especially since the following page points people to this component as the "preferred" client library, in general:

https://cloud.google.com/apis/docs/cloud-client-libraries

pongad:
Fixes #2732.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ This library supports the following Google Cloud Platform services with clients 
 
 These libraries are deprecated and no longer receive updates:
 
--  [Cloud Compute](google-cloud-compute)
+-  [Cloud Compute](google-cloud-compute) (Deprecated)
 
 Quickstart
 ----------


README.md
```diff
@@ -34,14 +34,17 @@ This library supports the following Google Cloud Platform services with clients
...
 -  [Cloud Dataproc](google-cloud-dataproc) (Alpha)
 -  [Cloud DNS](google-cloud-dns) (Alpha)
 -  [Cloud OS Login](google-cloud-os-login) (Alpha)
 -  [Cloud Resource Manager](google-cloud-resourcemanager) (Alpha)
 -  [Cloud Speech](google-cloud-speech) (Alpha)
 -  [Dialogflow](google-cloud-dialogflow) (Alpha)
 
+These libraries are deprecated and no longer receive updates:
+
+-  [Cloud Compute](google-cloud-compute)
```
garrettjonesgoogle: Use the same pattern as above - add `(Deprecated)` to the end of the line.

fkorotkov:
@pongad @garrettjonesgoogle what are the alternatives? I though `google-cloud-java` is the new set of libraries which will be the official ones and now libraries are deprecated before even getting to a release state. How can I and other customers be sure other libraries won't get deprecated?

garrettjonesgoogle:
Everything that is promoted to -beta will eventually be promoted to GA, and everything that is GA will be supported long-term. The only clients that may get deprecated and deleted are -alpha libraries.

As for Compute, we are planning on generating a new Compute client that will have much higher API coverage than the deprecated one, and will stay up to date over time.

fkorotkov:
Thanks for a quick response! I was pretty scared since one of our core functionality depends on this library. I think it's worth to mention that there will be a new client and it's not just dropped out of support.

PR #2822, opened 2018-01-30
Validate project-id returned from metadata server

yihanzhen:
Validate project-id returned from metadata server to handle cases in which a descriptive failure html page is returned instead of the project-id itself from server when projects are not running in google cloud machines.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -453,7 +453,7 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>,
             .setHeaders(new HttpHeaders().set("Metadata-Flavor", "Google"));
     HttpResponse response = request.execute();
     String projectId = response.parseAsString();
-    return (projectId != null && isValidProjectId(projectId))? projectId : null;
+    return projectId != null && isValidProjectId(projectId)? projectId : null;
   }
 
   protected static String getServiceAccountProjectId() {


google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -476,6 +477,18 @@ static String getServiceAccountProjectId(String credentialsPath) {
     return project;
   }
 
+  /* 
+   * Returns true if the projectId is valid.
+   * projectId must be between 6 and 30 characters
+   * projectId can have lowercase letters, digits or hyphens
+   * and must start with a lowercase letter
```
pongad: I think we should add a reference for this. Is there a web page where this comes from? So we can easily look it up if the requirement ever changes.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -476,6 +477,18 @@ static String getServiceAccountProjectId(String credentialsPath) {
...
+   */
+  private boolean isValidProjectId(String projectId) {
+    Pattern p = Pattern.compile("^[a-z][a-z0-9-]*$");
```
pongad: I think a `static final` member might be more appropriate here.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -476,6 +477,18 @@ static String getServiceAccountProjectId(String credentialsPath) {
     return project;
   }
 
+  /* 
+   * Returns true if the projectId is valid.
+   * projectId must be between 6 and 30 characters
+   * projectId can have lowercase letters, digits or hyphens
+   * and must start with a lowercase letter
+   */
+  private boolean isValidProjectId(String projectId) {
```
pongad: You asked on IM about test. I think a test is a good idea. You can make this `@InternalApi` so it's visible to test class.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
hzyi-google: Just to make sure, the signature should be `static final boolean ...`, not `private static final boolean ...`, is that right?

hzyi-google:
PTAL

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -86,6 +86,8 @@
   private static final RetrySettings NO_RETRY_SETTINGS = getDefaultRetrySettingsBuilder()
       .setMaxAttempts(1).build();
 
+  private static final Pattern projectIdPattern = Pattern.compile("^[a-z][a-z0-9-]*[a-z0-9]+$");
```
pongad: I think there is a bug here. `a-b-c` should be a valid project (I think) but the regex wouldn't match.

I think the right one would be `^[a-z][a-z0-9-]*$`

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -313,4 +314,16 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
...
+  @Test
+  public void testValidateProjectId() throws Exception {
+    String validProjectId = "abc-123";
+    String invalidProjectId1 = "abc=123";
+    String invalidProjectId2 = "abc123-";
+    String invalidProjectId3 = "1abc-23";
+    assertThat(ServiceOptions.isValidProjectId(validProjectId)).isTrue();
+    assertThat(ServiceOptions.isValidProjectId(invalidProjectId1)).isFalse();
+    assertThat(ServiceOptions.isValidProjectId(invalidProjectId2)).isFalse();
+    assertThat(ServiceOptions.isValidProjectId(invalidProjectId3)).isFalse();
```
pongad: I think you can just do `assertThat(ServiceOptions.isaValidProjectId("1abc-23")).isFalse()`. No need for indirection.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
hzyi-google: Sorry it is. How about `^[a-z]([a-z0-9]*-*)*[a-z0-9]$`? `^[a-z][a-z0-9]*$` would match "abc-", but trailing hyphens are not allowed.

pongad: Ah. Honestly regex is getting...difficult. Do you think we can just do this?

```java
for (char c : s.toCharArray()) {
  if (!isLetter(c) && !isDigit(c)) return false;
}
return s.length() > 0 && isLetter(s.charAt(0)) && !s.endsWith("-")
```

You'll need to implement isLetter and isDigit but those are easy.

hzyi-google: Yes. I agree with you on this. Will have a push soon

hzyi-google:
PTAL

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -476,6 +477,29 @@ static String getServiceAccountProjectId(String credentialsPath) {
...
+    for (char c : projectId.toCharArray()) {
+      if (!isLowerCase(c) && !isDigit(c) && c != '-') {
+        return false;
+      }
+    }
+    return projectId.length() > 0 && isLowerCase(projectId.charAt(0))
+      && !projectId.endsWith("-");
+  }
+
+  private static boolean isLowerCase(char c) {
```
wsh: https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isLowerCase(char)

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -476,6 +477,29 @@ static String getServiceAccountProjectId(String credentialsPath) {
...
+    return c >= 'a' && c <= 'z';
+  }
+
+  private static boolean isDigit(char c) {
```
wsh: https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isDigit(char)

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -476,6 +477,29 @@ static String getServiceAccountProjectId(String credentialsPath) {
...
+   * and does not end with a hyphen, but does not check the length of projectId. This
+   * method is primarily used to protect against DNS hijacking.
+   */
+  static boolean isValidProjectId(String projectId) {   
+    for (char c : projectId.toCharArray()) {
+      if (!isLowerCase(c) && !isDigit(c) && c != '-') {
+        return false;
+      }
+    }
+    return projectId.length() > 0 && isLowerCase(projectId.charAt(0))
```
wsh: Wouldn't that second condition already be verified by the previous loop?

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -476,6 +477,29 @@ static String getServiceAccountProjectId(String credentialsPath) {
...
   }
 
+  /* 
+   * Returns true if the projectId is valid. This method checks whether the projectId
+   * contains only lowercase letters, digits and hyphens, starts with a lowercase letter
+   * and does not end with a hyphen, but does not check the length of projectId. This
+   * method is primarily used to protect against DNS hijacking.
+   */
+  static boolean isValidProjectId(String projectId) {   
+    for (char c : projectId.toCharArray()) {
```
wsh: This doesn't validate that the *first* character isn't a hyphen, either. Also, some project IDs have colons in them (i.e. Googleplex apps). This validation will reject them, incorrectly. What are the constraints here based on? Is there a spec somewhere?

hzyi-google: There are constraints about project-id here[https://github.com/GoogleCloudPlatform/google-cloud-java/blob/536af840f93622a8743d7c129104e72909aeab0a/google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/ProjectInfo.java#L18] and here[https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects]. Also trying to create a project through cloud console with a project-id that does not follow the constraints will not be allowed. I haven't checked Googleplex apps though. I'll make the changes if colons are allowed in that case.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
hzyi-google: Project-id can only start with a lowercase letter (not a digit or a hyphen). The condition checks this constraint.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
hzyi-google: Please correct me if I am wrong. It seems Java's isLowerCase would return true for letters in languages other English, which is not allowed for project-id.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
hzyi-google: Same as previous.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
wsh: Oh, so you're using `isLowerCase` to verify that it's a letter. You can see why that's confusing.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
wsh: Neither of those is an actual, documented constraint. They're both implementation details of APIs. How do you know that pre-existing projects meet those constraints, for example? Assuming _isn't_ enough, especially when it will break customers, as *this change will* for Googleplex apps.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
wsh: Technically speaking, it's not about _English_, but non-ASCII characters--and furthermore, I don't see that in the references you mention, only "letters". What's actually allowed for a project ID?

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
wsh: What's a digit in this context?

PR #2828, opened 2018-01-31
datastore: better document emulator functions

### LINKED ISSUE #2800: improve LocalDatastoreHelper docs.
It would be nice to improve the javadocs for
http://googlecloudplatform.github.io/google-cloud-java/0.9.2/apidocs/com/google/cloud/datastore/testing/LocalDatastoreHelper.html

Especially the docs for reset look like generated docs string :). Please tell when one would call reset.

E.g. Does this setup for junit make sense? What bout including the code snippet into the docs?
``` java
  private static final LocalDatastoreHelper datastoreHelper = ...

  @BeforeClass
  public static void setUpOnce() throws Exception {
    datastoreHelper.start();
  }

  @AfterClass
  public static void tearDownOnce() throws Exception {
    datastoreHelper.stop(Duration.ofSeconds(5));
  }

  @Before
  public void setUp() throws Exception {
    datastoreHelper.reset();
  }
```

For start()/stop() it would be interesting to be explicit and tell if there is going to be any state (e.g. should I call reset() before calling stop()).

pongad:
Fixes #2800.

The API is unstable since we're still working on how to better
expose emulators.

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
@@ -198,6 +198,7 @@ public class LocalDatastoreHelper extends BaseEmulatorHelper<DatastoreOptions> {
    * <p>When running tests, one might {@code reset()} before each test, so earlier tests would not
    * affect later ones.
    */
+  @Override
   public void reset() throws IOException {
     sendPostRequest("/reset");
   }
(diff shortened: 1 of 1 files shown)


google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
```diff
@@ -193,14 +193,22 @@ public void start() throws IOException, InterruptedException {
...
   }
 
   /**
    * Stops the Datastore emulator.
+   *
+   * <p>It is important to stop the emulator. Since the emulator runs in its own process, not
+   * stopping it might cause it to become orphan.
+   *
+   * <p>It is not required to call {@link #reset()} before {@code stop()}.
    */
```
ensonic: Can we say anything about duration. 5 seconds seems to work for me, 1 second did not. I guess this is to wait the emulator being stopped to avoid weird test failures on subsequent runs if the emulator got stuck.

pongad: Sounds good. PTAL

google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
```diff
@@ -193,14 +193,25 @@ public void start() throws IOException, InterruptedException {
...
+   *
+   * <p>It is important to stop the emulator. Since the emulator runs in its own process, not
+   * stopping it might cause it to become orphan.
+   *
+   * <p>It is not required to call {@link #reset()} before {@code stop()}.
+   *
+   * @param timeout The duration to wait for the emulator process to stop. It is recommended to set
+   *     this value high to ensure proper shutdown, like 5 seconds or more.
    */
   public void stop(Duration timeout) throws IOException, InterruptedException, TimeoutException {
```
igorbernstein2: I'm not sure that this should be configureable. I don't think end users will have a better idea of how long it takes the emulator to shutdown than we do.

pongad: Unfortunately it must be, it is required by the base class. I'll add a no-config overload.

pongad:
@igorbernstein2 PTAL

igorbernstein2:
LGT<

PR #2829, opened 2018-01-31
CircleCI integration tests: adding more APIs

garrettjonesgoogle:
The following are known to fail - I want to push them so that we can track them:

- dns
- bigquery
- logging

### DIFF
--- a/utilities/verify_single_it.sh
+++ b/utilities/verify_single_it.sh
@@ -21,7 +21,7 @@ echo "----- building and installing parent pom -----"
 mvn -B -N install -DskipTests
 
 echo "----- building and installing shared modules -----"
-mvn -B -pl google-cloud-core,google-cloud-core-http,google-cloud-core-grpc,google-cloud-storage install -DskipTests
+mvn -B -pl google-cloud-core,google-cloud-core-http,google-cloud-core-grpc,google-cloud-storage,google-cloud-pubsub install -DskipTests
 
 echo "----- running integration tests -----"
 mvn -B -pl $MODULE -DtrimStackTrace=false -fae verify


garrettjonesgoogle:
PTAL

.circleci/config.yml
```diff
@@ -6,37 +6,140 @@ anchor_docker: &anchor_docker
       environment:
         TERM: dumb
 
+anchor_auth_vars: &anchor_auth_vars
+  environment:
+    GCLOUD_PROJECT: gcloud-devel
+    GOOGLE_APPLICATION_CREDENTIALS: .circleci/it-service-account.json
```
andreamlin: Even though this is encrypted, is it ok to host in a public repo?

garrettjonesgoogle: I think so. Other languages are already doing it. (I know, great excuse, isn't it?)

utilities/verify_single_it.sh
```diff
@@ -21,7 +21,7 @@ echo "----- building and installing parent pom -----"
 mvn -B -N install -DskipTests
 
 echo "----- building and installing shared modules -----"
-mvn -B -pl google-cloud-core,google-cloud-core-http,google-cloud-core-grpc install -DskipTests
+mvn -B -pl google-cloud-core,google-cloud-core-http,google-cloud-core-grpc,google-cloud-storage,google-cloud-pubsub install -DskipTests
```
andreamlin: why did storage and pubsub get put in this command?

garrettjonesgoogle: It was the easiest way to deal with the fact that bigquery depends on storage, and notification depends on pubsub (and storage).

Issue #2813, opened 2018-01-28
Compatibility Issue with Play 2.5.9


playtowork:
I have a Play framework project (version 2.5.9).

I append the following in its `build.sbt`'s `libraryDependencies`:
"com.google.cloud"        % "google-cloud-vision"      % "1.15.0"

Then I run `sbt clean && sbt "project frontend" run`. Instead of the app starting, it errors with the following:

--- (Running the application, auto-reloading is enabled) ---

java.lang.NoSuchFieldError: DEFAULT_MAX_PENDING_TASKS
	at io.netty.channel.nio.NioEventLoop.<init>(NioEventLoop.java:141)
	at io.netty.channel.nio.NioEventLoopGroup.newChild(NioEventLoopGroup.java:127)
	at io.netty.channel.nio.NioEventLoopGroup.newChild(NioEventLoopGroup.java:36)
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:84)
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:58)
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:47)
	at io.netty.channel.MultithreadEventLoopGroup.<init>(MultithreadEventLoopGroup.java:49)
	at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:77)
	at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:72)
	at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:59)
	at play.core.server.NettyServer.<init>(NettyServer.scala:77)
	at play.core.server.NettyServerProvider.createServer(NettyServer.scala:279)
	at play.core.server.NettyServerProvider.createServer(NettyServer.scala:278)
	at play.core.server.DevServerStart$$anonfun$mainDev$1.apply(DevServerStart.scala:235)
	at play.core.server.DevServerStart$$anonfun$mainDev$1.apply(DevServerStart.scala:65)
	at play.utils.Threads$.withContextClassLoader(Threads.scala:21)
	at play.core.server.DevServerStart$.mainDev(DevServerStart.scala:64)
	at play.core.server.DevServerStart$.mainDevHttpMode(DevServerStart.scala:54)
[...]

PR #2832, opened 2018-02-01
use grpc-netty-shaded

pongad:
Updates #2813 but does not close, since Spanner still requires unshaded netty.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -31,9 +31,9 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.grpc.ClientInterceptor;
 import io.grpc.ManagedChannel;
-import io.grpc.netty.GrpcSslContexts;
-import io.grpc.netty.NettyChannelBuilder;
-import io.netty.handler.ssl.SslContext;
+import io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts;
+import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
+import io.grpc.netty.shaded.io.netty.handler.ssl.SslContext;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.List;
@@ -79,7 +79,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
   private final int prefetchChunks;
   private final int numChannels;
   private final ImmutableMap<String, String> sessionLabels;
-  
+
   private SpannerOptions(Builder builder) {
     super(SpannerFactory.class, SpannerRpcFactory.class, builder, new SpannerDefaults());
     numChannels = builder.numChannels;
@@ -159,7 +159,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
 
     /**
      * Sets the labels to add to all Sessions created in this client.
-     * 
+     *
      * @param sessionLabels Map from label key to label value. Label key and value cannot be null.
      *     For more information on valid syntax see
      *     <a href="https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Session">
@@ -173,7 +173,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
       this.sessionLabels = ImmutableMap.copyOf(sessionLabels);
       return this;
     }
-    
+
     /**
      * Specifying this will allow the client to prefetch up to {@code prefetchChunks} {@code
      * PartialResultSet} chunks for each read and query. The data size of each chunk depends on the
@@ -224,7 +224,7 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
   public Map<String, String> getSessionLabels() {
     return sessionLabels;
   }
-  
+
   public int getPrefetchChunks() {
     return prefetchChunks;
   }
--- a/google-cloud-spanner/pom.xml
+++ b/google-cloud-spanner/pom.xml
@@ -53,10 +53,6 @@
   </build>
   <dependencies>
     <dependency>
-      <groupId>io.netty</groupId>
-      <artifactId>netty-tcnative-boringssl-static</artifactId>
-    </dependency>
-    <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>google-cloud-core</artifactId>
     </dependency>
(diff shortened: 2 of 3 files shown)


google-cloud-bom/pom.xml
```diff
@@ -127,14 +127,13 @@
     <api-client.version>1.23.0</api-client.version>
 
     <api-common.version>1.2.0</api-common.version>
-    <gax.version>1.17.0</gax.version>
-    <gax-grpc.version>1.17.0</gax-grpc.version>
-    <gax-httpjson.version>0.34.0</gax-httpjson.version>
+    <gax.version>1.17.1-SNAPSHOT</gax.version>
+    <gax-grpc.version>1.17.1-SNAPSHOT</gax-grpc.version>
+    <gax-httpjson.version>0.34.1-SNAPSHOT</gax-httpjson.version>
```
pongad: Change these to the right version after GAX is released.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -294,25 +292,25 @@ private static ManagedChannel createChannel(String rootUrl, RpcChannelFactory fa
...
-          NettyChannelBuilder.forAddress(host, port)
-            .sslContext(newSslContext())
+      ManagedChannelBuilder builder =
+          ManagedChannelBuilder.forAddress(host, port)
+            // .sslContext(newSslContext())
             .intercept(interceptors)
-            .maxHeaderListSize(MAX_HEADER_LIST_SIZE)
-            .maxMessageSize(MAX_MESSAGE_SIZE);
+            // .maxHeaderListSize(MAX_HEADER_LIST_SIZE)
+            .maxInboundMessageSize(MAX_MESSAGE_SIZE);
```
pongad: @vkedia How bad is this?

For context, many people are having problems with netty version conflicts. So gRPC provided `grpc-netty-shaded` in gRPC 1.9. Using it helps with the conflict, but it means we can't/shouldn't use Netty here.

google-cloud-util/google-cloud-compat-checker/pom.xml
```diff
@@ -21,6 +21,7 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
+      <version>2.0.7.Final</version>
```
pongad: @garrettjonesgoogle If netty is already shaded, do we need the compat checker at all?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
garrettjonesgoogle: Potentially we should leave Spanner with grpc-netty and file an issue to get it migrated to grpc-netty-shaded, so that we don't block everything else.

google-cloud-util/google-cloud-compat-checker/pom.xml
garrettjonesgoogle: There are fewer issues for it to check, but it will still tell you if your platform isn't supported.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
garrettjonesgoogle: Nevermind, I don't think this will work. Anyone who wanted to use both Pub/Sub and Spanner could only have one of grpc-netty or grpc-netty-shaded. We need to solve Spanner's problem here before we can switch over to grpc-netty-shaded.

vkedia: This cannot be removed since in certain cases cloud spanner would send a bigger header which will cause gRPC to throw an exception if if we do not increase the header size here. I think gprc should expose this functionality on ManagedChannelBuilder as well.

pongad: @garrettjonesgoogle According to [gRPC release doc](https://github.com/grpc/grpc-java/releases/tag/v1.9.0), it is safe to have both `grpc-netty` and `grpc-netty-shaded` right?

garrettjonesgoogle: If someone needs to use both Pub/Sub and Spanner, and they also can't get `grpc-netty` working such that they need to use `grpc-netty-shaded`, then Spanner won't work for them even if it's ok to have both `grpc-netty` and `grpc-netty-shaded` on the classpath. @ejona86 , do I have it right?

garrettjonesgoogle: But anyway, it appears that we could improve the situation for non-Spanner while not making Spanner worse off.

pongad: Reverted changes to spanner

pongad:
@garrettjonesgoogle PTAL
@vkedia Spanner is no longer affected.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
ejona86: Yes it is safe to depend on both at the same time. I will mention that we don't specify a relative priority between the two; gRPC will just pick one and hope for the best.

pongad:
@garrettjonesgoogle PTAL

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
vkedia: @ejona86 are there any plans to add these methods to ManagedChannelBuilder so that spanner can also move off using NettyChannelBuilder. Is there any other way we can do this without having to depend on NettyChannelBuilder. I would love to have spanner also depend on netty shaded since we have heard from multiple customers with this being an issue.

ejona86: @vkedia, by "these methods" do you mean only `maxHeaderListSize`? There's not currently any feature request open to add something like `maxHeaderListSize` to `ManagedChannelBuilder`. Maybe you want to create one?

I'll note that maxInboundMessageSize is currently experimental. https://github.com/grpc/grpc-java/issues/2563 is its tracking issue.

garrettjonesgoogle: Filed an issue: https://github.com/grpc/grpc-java/issues/4050

pongad:
@garrettjonesgoogle PTAL. Changed Spanner like we talked about

amerpersonal:
Any progress on this? Shading somehow won't work for me. I'm using ScalaPB in SBT project. This issue is really frustrating

pongad:
@amerpersonal The PR has already merged. Could you open a separate issue with the problem you have?

PR #2837, opened 2018-02-02
use GAE_RUNTIME to detect java 7 on GAE

### LINKED ISSUE #2805: PlatformInformation.isOnGAEStandard7 returns erroneous false positive
Seen from a Java 8 app in Flex. (Discovered because it's trying to construct a URLFetch transport for API clients.) Can provide further details, of course, but tbh I think there are better ways to detect this information than what that class does right now.

pongad:
Fixes #2805.

@wsh

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
@@ -20,33 +20,15 @@ import com.google.api.core.InternalApi;
 
 @InternalApi("This class should only be used within google-cloud-java")
 public final class PlatformInformation {
-  public static final String SPECIFICATION_VERSION =
-      System.getProperty("java.specification.version");
-  public static final String GAE_RUNTIME_VERSION =
-      System.getProperty("com.google.appengine.runtime.version");
-  public static final String RUNTIME_JETTY_LOGGER =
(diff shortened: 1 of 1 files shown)


google-cloud-core/src/main/java/com/google/cloud/PlatformInformation.java
```diff
@@ -38,10 +39,7 @@ public static boolean isOnGAE() {
...
 
   public static boolean isOnGAEStandard7() {
-    //edge case: when a Java 7 GAE app is deployed to DevServer running on J8, SPECIFICATION_VERSION will be "1.8",
-    //but RUNTIME_JETTY_LOGGER is unset
-    return isOnGAE() && SPECIFICATION_VERSION.equals("1.7")
-        || isOnGAE() && RUNTIME_JETTY_LOGGER == null;
+    return GAE_JAVA7;
   }
 
   public static boolean isOnGAEStandard8() {
```
wsh: This is also an opportunity to cheaply simply GAE Standard Java 8 detection, too. :smile: ([Reference](https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/444).)

pongad: Thank you for letting me know!

pongad:
@wsh PTAL

Issue #2413, opened 2017-09-07
Adding Snippets has errors


michaelbausor:
script: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/utilities/add_snippets_to_file.py

1. Method signature with @Nullable does not work.
Fix:
At this line in add_snippets_to_file.py#L99, replace the existing:
    ',\s+'.join(['(final\s+)?{}\s+{}'.format(
with this:
    ',\s+'.join(['(@Nullable\s+)?(final\s+)?{}\s+{}'.format(

2. for all single line comments, the script either fails (when its the last method in the file) or appends javadocs to the successive method.

example:
/** some javadoc */
public String method ( String args );

hzyi-google:
@pongad It seems your PR will resolve this issue, so I am giving this to you.

pongad:
Should be fixed by #2838.

PR #2838, opened 2018-02-02
RFC: new approach for snippet injection

pongad:
I recently thought about this problem and quickly wrote this up so we can have a concrete discussion. Comments welcome; hence the RFC.

The current snippet injector does not work properly
with google-java-format, because GJF formats short javadoc comments
on one line, eg "/** comment */".
However, the injector script looks for "/**" on a line by itself.
The script will also not work if/when we move to Java 8, due to lack
of parser support.

This PR takes a different approach of not caring about Java syntax and
copy-paste everything in the SNIPPET block. While less powerful, it is
more robust.

As written, the script is also easier to use. There's no need to tell it
what file contains snippets and where to copy the snippets to. The
script recursively scan given directories.

Updates #2413.

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
@@ -753,13 +753,19 @@ public interface BigQuery extends Service<BigQueryOptions> {
    * Updates dataset information.
    *
    * <p>Example of updating a dataset by changing its description.
-   * <pre> {@code
-   * String datasetName = "my_dataset_name";
-   * String newDescription = "some_new_description";
-   * Dataset oldDataset = bigquery.getDataset(datasetName);
-   * DatasetInfo datasetInfo = oldDataset.toBuilder().setDescription(newDescription).build();
-   * Dataset newDataset = bigquery.update(datasetInfo);
-   * }</pre>
+   * <!--SNIPPET bigquery_update_table_description-->
+   * <pre>{@code
+   *    // String datasetName = "my_dataset_name";
+   *    // String tableName = "my_table_name";
+   *    // String newDescription = "new_description";
+   *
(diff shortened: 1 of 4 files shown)


garrettjonesgoogle:
@jadekler I think it would be good for you to take a look at this

jadekler:
I'm not sure I get what's going on here. Is there a place I can see the changes live?

Also, what does this sentence mean?

> This PR takes a different approach of not caring about Java syntax and copy-paste everything in the SNIPPET block.

Not caring about java syntax? As in, the formatter somehow doesn't look at your syntax?

Sorry, out of the loop on these issues :(

jadekler:
Ahhh!! Very cool. Seems like a great change.

tswast:
I like that it is more robust. I do wish it supported START and END tags so that we can continue to use code snippets in cloud.google.com, though. I'm fine with the change to simplify, as that will actually make it easier to make sure the code snippets stay consistent with what ends up getting displayed at cloud.google.com.

tswast:
Actually, if it could just strip out START and END lines before including them in javadocs that would be sufficient. Then the SNIPPET lines could wrap START/END where needed and START/END wouldn't be necessary for the samples that will never be included on cloud.google.com. (Extra start/end tags throw off our samples tracker, so it's actually a feature if the syntax is a bit different as you're doing here)

tswast:
Soft ping to review, now that folks should be back from July 4th travels.

jabubake:
LGTM, and this seems like a good change ! Thanks @pongad @tswast

pongad:
Is `getSnip` still useful? If not, I think we can just use `getCloud` everywhere.

tswast:
@pongad Yes, it is useful to have `getSnip`. @alixhami and I have noticed that many samples we want in the JavaDocs don't make sense to include on cloud.google.com (mostly because they use redundant overloads). `getSnip` is useful to keep our cloud.google.com sample tracker from getting cluttered up with unpublished samples.

pongad:
LGTM then :)

PR #2839, opened 2018-02-02
CircleCI: Adding Storage

garrettjonesgoogle:


garrettjonesgoogle:
This is the last integration test to add. (I thought I might have to do something special for signing, but it turns out I didn't need to.)

PR #2841, opened 2018-02-02
Bigtable: 00. Add a couple of developer helper scripts

igorbernstein2:
This is to ease updating the gapic client and updating googleapis

This has been reviewed by Kevin, however the following changes were made since:
- paths have been updated from the overlay to the in tree updates
- file lists have been added to allow regeneration
- `bin` was renamed to `scripts`
- a readme was added


pongad:
LGTM. @garrettjonesgoogle Do you want to take a look?

google-cloud-bigtable/pom.xml
```diff
@@ -104,6 +104,32 @@
...
       </properties>
     </profile>
+
+    <profile>
+      <id>bigtable-dev</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.theoryinpractise</groupId>
+            <artifactId>googleformatter-maven-plugin</artifactId>
```
garrettjonesgoogle: I would prefer to apply this to all of google-cloud-java instead of just google-cloud-bigtable

google-cloud-bigtable/scripts/autogen_files.lst
```diff
@@ -0,0 +1,27 @@
+src/test/java/com/google/cloud/bigtable/data/v2/MockBigtable.java
```
garrettjonesgoogle: It seems like the .lst files should be up one directory

google-cloud-bigtable/scripts/regenerate.sh
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env bash
+
+# This script is used to regenerate protos & GAPIC stubs.
+# It stages all of the generated code in deps/ and then installs it into the local maven repo.
+# It is meant to be used to test changes to the gapic config before publishing
+# To pull in upstream changes, please use update-deps.sh
+
+set -eE # same as: `set -o errexit -o errtrace`
```
garrettjonesgoogle: If you're going to spill the ink to say what the command is, why not just use the extended command? :-P

google-cloud-bigtable/scripts/regenerate.sh
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env bash
```
garrettjonesgoogle: I am hesitant to have these bigtable-specific scripts checked into google-cloud-java. My main concern is them becoming abandoned over time and becoming deadweight because nobody knows how to use them. I'm ok with having them checked in here if we commit to generalizing them for other API owners (which would be an awesome result); otherwise, I would prefer to keep them outside of the repo.

google-cloud-bigtable/scripts/update-deps.sh
```diff
@@ -0,0 +1,80 @@
...
+function update() {
+  for DIR in "${DIRS[@]}"; do
+    log "Updating $DIR"
+    DIR="$WORKDIR/$DIR"
+
+    log "Fetching upstream"
+    git -C "$DIR" fetch origin >>$LOG 2>&1
+
+    log "Dropping uncommitted local changes"
+    git -C "$DIR" reset --hard HEAD >>$LOG 2>&1
```
garrettjonesgoogle: This feels like a grenade. Could we have the script fail if the user has any uncommitted local changes?

google-cloud-bigtable/pom.xml
igorbernstein2: Extracted to https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2851

google-cloud-bigtable/scripts/regenerate.sh
igorbernstein2: I found them very useful when when testing gapic config changes against the client. I can work to generalize them, but I think it would require a sync on your plans for autogenerating clients in general.
Can we compromise on leaving them here for now and I will  circle back to them?

google-cloud-bigtable/scripts/update-deps.sh
igorbernstein2: I originally wrote it that way, but in practice it became an annoyance and I changed it to this. Most of the time, the repos are dirty because they contain generated code from the changes you are testing.

igorbernstein2:
Responded to all feedback. PTAL

google-cloud-bigtable/scripts/update-deps.sh
garrettjonesgoogle: how hard would it be to have it automatically clear out auto-generated edits but not manual edits? (You have the lst files...)

google-cloud-bigtable/scripts/regenerate.sh
garrettjonesgoogle: Yes, I just want to make sure we circle back :-)

google-cloud-bigtable/scripts/update-deps.sh
igorbernstein2: That should work. But would prefer to delay that as a feature add later. Can we merge this and open an issue for me to generalize the scripts and add safer regen guards?

garrettjonesgoogle: Sure.

igorbernstein2:
Created an issue:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2852

Please merge when ready

PR #2845, opened 2018-02-03
release 0.33.1/1.15.1

neozwu:


garrettjonesgoogle:
- We need to do a minor version bump instead of a patch bump (the stub settings change was breaking to `@BetaApi` classes)
- You need to manually fix up the versions in the BOM

neozwu:
Drop this release bump since I have created another minor version bump PR.

PR #2846, opened 2018-02-03
Add initial opencensus instrumentation code in Cloud Storage.

HailongWen:
This PR adds [OpenCensus-Java](https://github.com/census-instrumentation/opencensus-java) instrumentation to monitor and record events in Cloud Storage client and propagate the trace context via HTTP headers to google service.

Note that there is another PR to instrument underlying library google-http-java-client is also [instrumented](https://github.com/google/google-http-java-client/pull/393) so as to provide detailed span break down. In case that we don't catch up new release of google-http-java-client, in this PR I still added helper class to do context propagation.

There are 3 major changes in this PR:
* Added a helper class `CensusHttpModule` to do trace context propagation through `HttpHeaders`. (might be temporary - can be removed if google-http-java-client release new version)
* Added a helper class `HttpStorageRpcSpans`, which contains all span names and  a method to register them all for collection.
* Added instrumentation code in `HttpStorageRpc`.  The instrumented code starts a new `Span` the main logic of the method begins, set it as current span with `Tracer.withSpan`. If exception occurs during the execution, the span status will be set to `Status.UNKNOWN` to indicate unparsed exception. The span and its scope will be ended before method exits.

All the above `Span`s will be stored into local span store, and can be optionally exported if user explicitly configure an exporter (e.g. [stackdriver exporter](https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/trace/stackdriver)).

Please @hzyi-google and @bogdandrutu kindly help review.

### DIFF
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -1049,11 +1049,6 @@
       </dependency>
       <dependency>
         <groupId>io.opencensus</groupId>
-        <artifactId>opencensus-impl</artifactId>
-        <version>${opencensus.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>io.opencensus</groupId>
         <artifactId>opencensus-contrib-grpc-util</artifactId>
         <version>${opencensus.version}</version>
       </dependency>
--- a/google-cloud-core-http/pom.xml
+++ b/google-cloud-core-http/pom.xml
@@ -129,10 +129,6 @@
     </dependency>
     <dependency>
       <groupId>io.opencensus</groupId>
-      <artifactId>opencensus-impl</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>io.opencensus</groupId>
       <artifactId>opencensus-contrib-http-util</artifactId>
     </dependency>
   </dependencies>
--- a/google-cloud-core-http/src/test/java/com/google/cloud/http/CensusHttpModuleTest.java
+++ b/google-cloud-core-http/src/test/java/com/google/cloud/http/CensusHttpModuleTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.http;
 import static com.google.common.truth.Truth.assertThat;
 
 import static org.easymock.EasyMock.createMock;
+import static org.easymock.EasyMock.createMockBuilder;
 import static org.easymock.EasyMock.replay;
 import static org.easymock.EasyMock.verify;
 
@@ -30,11 +31,18 @@ import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.javanet.NetHttpTransport;
 
 import io.opencensus.common.Scope;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.SpanContext;
+import io.opencensus.trace.SpanId;
+import io.opencensus.trace.TraceId;
+import io.opencensus.trace.TraceOptions;
 import io.opencensus.trace.Tracer;
 import io.opencensus.trace.Tracing;
 import io.opencensus.trace.propagation.TextFormat;
 
 import java.io.IOException;
+import java.util.EnumSet;
+import java.util.Random;
 
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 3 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -91,6 +101,13 @@ public HttpStorageRpc(StorageOptions options) {
     HttpTransport transport = transportOptions.getHttpTransportFactory().create();
     HttpRequestInitializer initializer = transportOptions.getHttpRequestInitializer(options);
     this.options = options;
+
+    // Open Census initialization
+    censusHttpModule = new CensusHttpModule(tracer, true);
+    initializer = censusHttpModule.getHttpRequestInitializer(initializer);
+    batchRequestInitializer = censusHttpModule.getHttpRequestInitializer(null);
```
HailongWen: Two initializers. One is the default one which wraps the original initializer. The other one wraps nothing and will be used in batch request (see comments below in line 133).

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -112,7 +129,8 @@ public HttpStorageRpc(StorageOptions options) {
     private DefaultRpcBatch(Storage storage) {
       this.storage = storage;
       batches = new LinkedList<>();
-      batches.add(storage.batch());
+      // add OpenCensus HttpRequestInitializer
+      batches.add(storage.batch(batchRequestInitializer));
```
HailongWen: `storage.batch()` can take an `HttpRequestInitializer` as parameter. If not given, it will use null.
Since the original code does not `initializer` for some reason, here I use another initializer which wrapper nothing to maintain the same logic.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -162,12 +180,22 @@ public void addGet(StorageObject storageObject, RpcBatch.Callback<StorageObject>
 
     @Override
     public void submit() {
```
HailongWen: I didn't instrument `addGet`, `addDelete`, `addPatch` since I think what they do is just add something into the queue. Please correct me if I am wrong.

google-cloud-core-http/src/main/java/com/google/cloud/http/CensusHttpModule.java
```diff
@@ -0,0 +1,168 @@
...
+          propagationTextFormat.inject(
+              spanContext, request.getHeaders(), propagationTextFormatSetter);
+        }
+      }
+    }
+  }
+
+  /**
+   * An {@link HttpRequestInitializer} implementation to set {@link CensusHttpExecuteInterceptor} as
+   * intercptor.
```
hzyi-google: typo: "interceptor"

hzyi-google: typo: "interceptor"

HailongWen: Thanks.

pongad:
@HailongWen I see the WIP in the title. Is this ready to review now?

HailongWen:
@pongad Yes. Sorry for the confusion. I have changed the title.

hzyi-google:
LGTM +1

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
pongad: I believe you're correct.

google-cloud-bom/pom.xml
```diff
@@ -1052,11 +1052,21 @@
         <artifactId>opencensus-api</artifactId>
         <version>${opencensus.version}</version>
       </dependency>
+      <dependency>
```
bogdandrutu: We should never depend on the impl artifact. The end user should do this.

HailongWen: Should we consider GCS team as the end user? I mean, they might want to collect the data by default. I figure it is better to involve GCS team member in the discussion here. @jabubake @frankyn

bogdandrutu: The end user is the user that uses this client to talk to GCS.

HailongWen: Yes. And I agree that it provides more flexibility. I'll remove it for now and leave that decision for GCS team.

pongad: I don't understand the package structure of opencensus very well. What's the consequence of removing the impl artifact?

HailongWen: We have opencensus-api which provides basic API and no-op implementation for trace and stats, while opencensus-impl, if present in classpath, will replace the op with real implementation.

So the idea here is that we should always depend on no-op api by default, and let end user to optionally enable the tracing/stats by adding opencensus-impl.

pongad: I see, this makes sense. Thank you for confirming.

HailongWen:
Thank you all for the help and support!

PR #2860, opened 2018-02-05
Fixing Logging integration tests

garrettjonesgoogle:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2835


google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java
```diff
@@ -273,17 +284,29 @@ public void testWriteAndListLogEntries() throws InterruptedException {
...
+    int allowedDeleteAttempts = 5;
+    boolean deleted = logging().deleteLog(logId);
+    while (!deleted && deleteAttempts < allowedDeleteAttempts) {
+      deleteAttempts++;
+      System.out.println("Not deleted - " + deleteAttempts + " of "
+          + allowedDeleteAttempts + " attempted so far.");
+      Thread.sleep(1000);
+      deleted = logging().deleteLog(logId);
+    }
+    assertTrue(deleted);
```
pongad: I think we should use `rateLimiter.acquire()` instead of `sleep`.
I also don't think it's necessary to print, but if you think that's worthwhile, then I think `System.err` is a better choice.

This might just be me; I find it easier to do
```java
boolean deleted = false;
while (...) {
  rateLimiter.acquire();
  deleted = logging().deleteLog(logId);
}
```
so we only need to write the delete once. Up to you though.

garrettjonesgoogle:
It turns out the high quota usage was calls to ListLogMetrics and ListMonitoredResourceDescriptors. I fixed the first by clearing out the metrics in the Cloud Console. I fixed the second by bumping the page size from 1 to 100. The rate limiter is basically not necessary, so I removed it. I am retaining the retry on deletion, since that is regularly needed when I run the test locally. 

PTAL

PR #2863, opened 2018-02-06
core: allow custom roles

### LINKED ISSUE #2687: Can't attach custom role to service account or container
I created a custom role that consists of some permission but when I tried to attach it to service account or container programatically it always give error. Is it not supported yet? If it is supported how can I attach it to service account or container?

My code are as follow:
In Google Cloud Storage
```
Policy policy = storage.getIamPolicy(bucketName);

Policy updatedPolicy = storage.setIamPolicy(bucketName, policy.toBuilder().addIdentity(Role.of("roles/CustomRole341"), Identity.serviceAccount(emailServiceAccount)).build());
```

```
ServiceAccount user = findServiceAccount(iam, username, projectName);
List<String> member = new ArrayList<String>();
member.add(Identity.serviceAccount(user.getEmail()).toString());

Binding s = new Binding();
s.setRole("roles/CustomRole341");
s.setMembers(member);

List<Binding> bindings = new ArrayList<Binding>();
bindings.add(s);

Policy a = new Policy();
a.setBindings(bindings);

SetIamPolicyRequest req = new SetIamPolicyRequest();
req.setPolicy(a);

iam.projects().serviceAccounts().setIamPolicy("projects/" + projectName + "/serviceAccounts/" + username +  "@" + projectName + ".iam.gserviceaccount.com" , req).execute()
```

pongad:
Role previously mandated that all role strings starts with "role/".
If the given string doesn't start with the prefix, Role prepends it.

This is incorrect, since custom roles have format "projects/X/roles/Y".
This PR makes Role a little smarter at prepending:
we only prepend if the given string contains no slash; this keeps the
previously-valid use valid.

Fixes #2687.


vchudnov-g:
Cool. Could you link to an official reference for this role behavior?

pongad:
@vchudnov-g I'm not sure what you mean by "behavior" here. Do you mean how it automatically prepends "roles/"?

If so, this is not a "standard behavior". The problem is `Roles` is trying too hard to be useful. If you run `Roles.of("owner")`, "owner" is not a valid role but "roles/owner" is, so `Roles` helpfully prepends.

However this makes custom roles impossible. Since custom role is in format "projects/X/roles/Y", calling `Roles.of("projects/X/roles/Y")` creates role "roles/projects/X/roles/Y". This PR makes it possible to refer to custom roles.

Am I understanding you correctly?

vchudnov-g:
That explanation is helpful. My comment was about including a link to an official reference to what you explain about "roles/owner" vs "projects/X/roles/Y", just for documentation purposes.

google-cloud-core/src/test/java/com/google/cloud/RoleTest.java
```diff
@@ -28,28 +28,31 @@
 
   @Test
   public void testOf() {
-    assertEquals("roles/viewer", VIEWER.getValue());
-    assertEquals("roles/editor", EDITOR.getValue());
-    assertEquals("roles/owner", OWNER.getValue());
+    assertThat(VIEWER.getValue()).isEqualTo("roles/viewer");
```
vchudnov-g: Is the convention to have expected values first or second? In other code bases, I've seen (expected, actual). I'm fine following whatever is the convention here.

pongad: Convention of `assertEquals` is to put expected value first. However, convention of `assertThat` is to [put actual first](https://google.github.io/truth/faq#order).

> That explanation is helpful. My comment was about including a link to an official reference to what you explain about "roles/owner" vs "projects/X/roles/Y", just for documentation purposes.

Ah I see. I can do that.

Issue #2440, opened 2017-09-15
Datastore - Datastore object creation fails with version 1.5.1


sai-pullabhotla:
I'm trying to upgrade from version 1.2.3 of the Datastore API to the latest version, 1.5.1. The below sample code fails creating the Datastore object/service. 

```
Datastore datastore = DatastoreOptions.newBuilder().build().getService();
```

Tests are running on my laptop with gcloud installed. Version 1.2.3 works fine by fetching the defaults (e.g. project/credentials) from Google Cloud SDK. Version 1.5.1 is producing the below exception: 

```
Exception in thread "main" java.lang.IllegalArgumentException: java.net.URISyntaxException: Illegal character in path at index 45: https://datastore.googleapis.com/v1/projects/<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html><head><meta http-equiv="refresh" content="0;url=http://finder.cox.net/main?ParticipantID=96e687opkbv4scrood8k84drs6gw5duf&FailedURI=http%3A%2F%2Fmetadata.google.internal%2FcomputeMetadata%2Fv1%2Fproject%2Fproject-id&FailureMode=1&Implementation=&AddInType=4&Version=pywr1.0&ClientLocation=us"/><script type="text/javascript">url="http://finder.cox.net/main?ParticipantID=96e687opkbv4scrood8k84drs6gw5duf&FailedURI=http%3A%2F%2Fmetadata.google.internal%2FcomputeMetadata%2Fv1%2Fproject%2Fproject-id&FailureMode=1&Implementation=&AddInType=4&Version=pywr1.0&ClientLocation=us";if(top.location!=location){var w=window,d=document,e=d.documentElement,b=d.body,x=w.innerWidth||e.clientWidth||b.clientWidth,y=w.innerHeight||e.clientHeight||b.clientHeight;url+="&w="+x+"&h="+y;}window.location.replace(url);</script></head><body></body></html>
	at com.google.datastore.v1.client.DatastoreFactory.validateUrl(DatastoreFactory.java:118)
	at com.google.datastore.v1.client.DatastoreFactory.buildProjectEndpoint(DatastoreFactory.java:104)
	at com.google.datastore.v1.client.DatastoreFactory.newRemoteRpc(DatastoreFactory.java:111)
[...]

PR #2868, opened 2018-02-06
Add validation for project-id returned by metadata server

yihanzhen:
Check whether the `HttpResponse` from metadata server contains a field `Metadata-Flavor` in its header when requesting project-id. 
If yes, the response will be parsed as the project-id; 
If not, the response will be considered invalid and `ServiceOptions` will continue to look for project-id elsewhere.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -457,7 +457,8 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>,
 
   @InternalApi("Visible for testing")
   static boolean headerContainsMetadataFlavor(HttpResponse response) {
-    return "Google".equals(response.getHeaders().getFirstHeaderStringValue("Metadata-Flavor"));
+    String metadataFlavorValue = response.getHeaders().getFirstHeaderStringValue("Metadata-Flavor");
+    return "Google".equals(metadataFlavorValue);
   }
 
   protected static String getServiceAccountProjectId() {
--- a/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
+++ b/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
@@ -44,11 +44,12 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
 import com.google.common.io.Files;
 import org.junit.Rule;
 import org.junit.Test;
@@ -328,17 +329,21 @@ public class ServiceOptionsTest {
   }
 
   @Test
-  public void testResponseHeaderContainsMetaDataFlavor() throws Exception {  
-    Map<String, String> headers = new HashMap<String, String>();
-    HttpResponse httpResponse = createHttpResponseWithHeader(headers);
-    assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isFalse();
-
+  public void testResponseHeaderContainsMetaDataFlavor() throws Exception {
+    Multimap<String, String> headers = ArrayListMultimap.create();
     headers.put("Metadata-Flavor", "Google");
-    httpResponse = createHttpResponseWithHeader(headers);
+    HttpResponse httpResponse = createHttpResponseWithHeader(headers);
     assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isTrue();
   }
+
+  @Test
+  public void testResponseHeaderDoesNotContainMetaDataFlavor() throws Exception {  
+    Multimap<String, String> headers = ArrayListMultimap.create();
+    HttpResponse httpResponse = createHttpResponseWithHeader(headers);
+    assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isFalse(); 
+  }
   
-  private HttpResponse createHttpResponseWithHeader(final Map<String, String> headers) throws Exception {
+  private HttpResponse createHttpResponseWithHeader(final Multimap<String, String> headers) throws Exception {
     HttpTransport mockHttpTransport = new MockHttpTransport() {
       @Override
       public LowLevelHttpRequest buildRequest(String method, String url) throws IOException {
@@ -346,7 +351,7 @@ public class ServiceOptionsTest {
           @Override
           public LowLevelHttpResponse execute() throws IOException {
             MockLowLevelHttpResponse response = new MockLowLevelHttpResponse();
-            for (Map.Entry<String, String> entry : headers.entrySet()) {
+            for (Map.Entry<String, String> entry : headers.entries()) {
               response.addHeader(entry.getKey(), entry.getValue());
             }            
             return response;


hzyi-google:
This fixes #2440

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -313,4 +324,26 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
 
     assertNull(ServiceOptions.getServiceAccountProjectId(credentialsFile.getPath()));
   }
+
+  @Test
+  public void testResponseHeaderContainsMetaDataFlavor() throws Exception {
```
wsh: Great test! I think it might be simpler to mock the `HttpRequest` itself (e.g. `when(request).execute().thenReturn(mockResponse)`.

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -313,4 +324,26 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
 
     assertNull(ServiceOptions.getServiceAccountProjectId(credentialsFile.getPath()));
   }
+
+  @Test
```
wsh: It's also worth testing the _negative_ case, e.g. when the response is invalid (like the Cox hijacked response). In particular this would reveal the NPE bug in the current code.

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -452,7 +452,13 @@ private static String getAppEngineProjectIdFromMetadataServer() throws IOExcepti
...
             .setHeaders(new HttpHeaders().set("Metadata-Flavor", "Google"));
     HttpResponse response = request.execute();
-    return response.parseAsString();
+    return headerContainsMetadataFlavor(response) ? response.parseAsString() : null;
+  }
+
+  @InternalApi("Visible for testing")
+  static boolean headerContainsMetadataFlavor(HttpResponse response) {
+    return response.getHeaders()
+      .getFirstHeaderStringValue("Metadata-Flavor").equals("Google");
```
wsh: Per [the docs for getFirstHeaderStringValue](https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/HttpHeaders#getFirstHeaderStringValue(java.lang.String)), this method returns _null_ when the header isn't found, so this method will either return true (inside GCP) _or_ blow up with a `NullPointerException` elsewhere. (A negative test case, as mentioned below, would catch this bug).

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -313,4 +326,30 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
...
+    //         response.setStatusCode(200);
+    //         return response;
+    //       }
+    //     };
+    //   }
+    // };
+    HttpTransport mockHttpTransport = Mockito.mock(MockHttpTransport.class);
+    MockLowLevelHttpResponse mockResponse = new MockLowLevelHttpResponse();
+    mockResponse.addHeader("Metadata-Flavor", "Google");
+    Mockito.when(mockHttpTransport.buildRequest(Matchers.anyString(), Matchers.anyString()).execute()).thenReturn(mockResponse);
```
hzyi-google: This won't work, due to the fact that `buildRequest(String, String)` has protected access in `com.google.api.client.http.HttpTransport`

google-cloud-core/pom.xml
```diff
@@ -87,5 +87,10 @@
       <artifactId>guava-testlib</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
+      <version>1.9.5</version>
```
wsh: This should have a <scope>test</scope> modifier so that it's not included in the non-test JAR. :)

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -457,8 +457,8 @@ private static String getAppEngineProjectIdFromMetadataServer() throws IOExcepti
 
   @InternalApi("Visible for testing")
   static boolean headerContainsMetadataFlavor(HttpResponse response) {
-    return response.getHeaders()
-      .getFirstHeaderStringValue("Metadata-Flavor").equals("Google");
+    String metadataFlavorValue = response.getHeaders().getFirstHeaderStringValue("Metadata-Flavor");
+    return metadataFlavorValue != null && metadataFlavorValue.equals("Google");
```
wsh: Nice!

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
wsh: https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/testing/http/MockHttpTransport to the rescue!

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -452,7 +452,13 @@ private static String getAppEngineProjectIdFromMetadataServer() throws IOExcepti
...
             .setHeaders(new HttpHeaders().set("Metadata-Flavor", "Google"));
     HttpResponse response = request.execute();
-    return response.parseAsString();
+    return headerContainsMetadataFlavor(response) ? response.parseAsString() : null;
+  }
+
+  @InternalApi("Visible for testing")
+  static boolean headerContainsMetadataFlavor(HttpResponse response) {
+    String metadataFlavorValue = response.getHeaders().getFirstHeaderStringValue("Metadata-Flavor");
+    return metadataFlavorValue != null && metadataFlavorValue.equals("Google");
```
pongad: Just do `"Google".equals(metadataFlavorValue)`. Null will compare unequal.

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -313,4 +326,35 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
...
+    Map<String, String> headers = new HashMap<String, String>();
+    HttpResponse httpResponse = createHttpResponseWithHeader(headers);
+    assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isFalse();
+
+    headers.put("Metadata-Flavor", "Google");
+    httpResponse = createHttpResponseWithHeader(headers);
+    assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isTrue();
+  }
+  
+  private HttpResponse createHttpResponseWithHeader(final Map<String, String> headers) throws Exception {
```
hzyi-google: I am not sure if `Map<String, String>` is the best practice here. `HttpResponse` seems to allow duplicate fields in some cases... Any suggestions?

hzyi-google:
Can you take a look? @wsh @pongad Thanks!

google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -457,8 +457,7 @@ private static String getAppEngineProjectIdFromMetadataServer() throws IOExcepti
 
   @InternalApi("Visible for testing")
   static boolean headerContainsMetadataFlavor(HttpResponse response) {
-    String metadataFlavorValue = response.getHeaders().getFirstHeaderStringValue("Metadata-Flavor");
-    return metadataFlavorValue != null && metadataFlavorValue.equals("Google");
+    return "Google".equals(response.getHeaders().getFirstHeaderStringValue("Metadata-Flavor"));
```
wsh: I _personally_ think it's clearer to keep the actual header in its own variable, but /shrug. Matter of taste. :)

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -329,27 +329,32 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
...
-    MockLowLevelHttpResponse mockResponse = new MockLowLevelHttpResponse();
-    mockResponse.addHeader("Metadata-Flavor", "Google");
-    Mockito.when(mockHttpTransport.buildRequest(Matchers.anyString(), Matchers.anyString()).execute()).thenReturn(mockResponse);
+    Map<String, String> headers = new HashMap<String, String>();
+    HttpResponse httpResponse = createHttpResponseWithHeader(headers);
+    assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isFalse();
 
+    headers.put("Metadata-Flavor", "Google");
+    httpResponse = createHttpResponseWithHeader(headers);
+    assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isTrue();
```
wsh: These should be separate test cases. :)

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
wsh: It does! If you wanted to allow for that, you could use a `Multimap` from Guava.

hzyi-google:
Another look? @wsh

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -328,25 +329,29 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
...
+  public void testResponseHeaderContainsMetaDataFlavor() throws Exception {
+    Multimap<String, String> headers = ArrayListMultimap.create();
     headers.put("Metadata-Flavor", "Google");
-    httpResponse = createHttpResponseWithHeader(headers);
+    HttpResponse httpResponse = createHttpResponseWithHeader(headers);
     assertThat(ServiceOptions.headerContainsMetadataFlavor(httpResponse)).isTrue();
   }
+
+  @Test
+  public void testResponseHeaderDoesNotContainMetaDataFlavor() throws Exception {
```
wsh: Awesome! :)

google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -328,25 +329,29 @@ public void testGetServiceAccountProjectId_nonExistentFile() throws Exception {
...
+  private HttpResponse createHttpResponseWithHeader(final Multimap<String, String> headers) throws Exception {
     HttpTransport mockHttpTransport = new MockHttpTransport() {
       @Override
       public LowLevelHttpRequest buildRequest(String method, String url) throws IOException {
         return new MockLowLevelHttpRequest() {
           @Override
           public LowLevelHttpResponse execute() throws IOException {
             MockLowLevelHttpResponse response = new MockLowLevelHttpResponse();
-            for (Map.Entry<String, String> entry : headers.entrySet()) {
+            for (Map.Entry<String, String> entry : headers.entries()) {
```
wsh: Just as an FYI--it's not like it makes a difference here, but just so you know :)--this simplifies to one line using streams:

`headers.entries().forEach(e -> response.addHeader(e.getKey(), e.getValue());`

pongad: We need Java 7 compat, so unfortunately so streams for us :(

wsh: :(
Will this change once the App Engine Standard Java 7 runtime is decommissioned?

PR #2870, opened 2018-02-06
BigQuery lib. Add support for customer supplied encryptiion keys

alexblyzn:
More details for this change in https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2866

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/CopyJobConfiguration.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/CopyJobConfiguration.java
@@ -19,7 +19,6 @@ package com.google.cloud.bigquery;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.services.bigquery.model.JobConfigurationTableCopy;
-import com.google.cloud.bigquery.LoadJobConfiguration.Builder;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects.ToStringHelper;
 import com.google.common.collect.ImmutableList;
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/EncryptionConfiguration.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/EncryptionConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Google LLC
+ * Copyright 2018 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,8 @@ import java.io.Serializable;
 import java.util.Objects;
 
 public final class EncryptionConfiguration implements Serializable {
+  private static final long serialVersionUID = -100849671239964L;
+
   private final String kmsKeyName;
 
   public String getKmsKeyName() {
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
@@ -24,7 +24,6 @@ import com.google.api.services.bigquery.model.QueryParameter;
 import com.google.cloud.bigquery.JobInfo.CreateDisposition;
 import com.google.cloud.bigquery.JobInfo.WriteDisposition;
 import com.google.cloud.bigquery.JobInfo.SchemaUpdateOption;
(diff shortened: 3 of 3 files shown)


alexblyzn:
I have integration tests for the change as well, but I don't have a good way to commit them. They require KMS key setup as dependency, it seems somewhat unreasonable to make everyone go through that setup if they want to run tests.

The tests:
https://github.com/alexblyzn/google-cloud-java/commit/7a34991f2c5b434e31d05dbd59e03a51ebdfcd77#diff-af63ea32c5525b76f3837bc5aa8e5a25

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/CopyJobConfiguration.java
```diff
@@ -19,6 +19,7 @@
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.services.bigquery.model.JobConfigurationTableCopy;
+import com.google.cloud.bigquery.LoadJobConfiguration.Builder;
```
pongad: How is this used? I find it a little strange that a copy configuration uses load config builder.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/EncryptionConfiguration.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2016 Google LLC
```
pongad: 2018?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/EncryptionConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigquery;
+
+import com.google.common.base.MoreObjects;
+import java.io.Serializable;
+import java.util.Objects;
+
+public final class EncryptionConfiguration implements Serializable {
```
pongad: Consider using `@AutoValue` for this

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
```diff
@@ -24,6 +24,7 @@
 import com.google.cloud.bigquery.JobInfo.CreateDisposition;
 import com.google.cloud.bigquery.JobInfo.WriteDisposition;
 import com.google.cloud.bigquery.JobInfo.SchemaUpdateOption;
+import com.google.cloud.bigquery.LoadJobConfiguration.Builder;
```
pongad: same comment about Builder

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/CopyJobConfiguration.java
alexblyzn: not used, removed.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/EncryptionConfiguration.java
alexblyzn: Updated

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/EncryptionConfiguration.java
alexblyzn: It needs to be parsed and serialized to proto. It is easier for me to follow existing pattern here.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java
alexblyzn: Removed

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/EncryptionConfiguration.java
pongad: OK with me. I suppose it's not hard to change later.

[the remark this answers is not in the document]
pongad: I think we should do this.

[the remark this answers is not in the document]
alexblyzn: Added.

alexblyzn:
Is there anything else I need to change before you can accept the request?

PR #2872, opened 2018-02-07
Fixing intermittent Storage integration test failures

garrettjonesgoogle:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2864


pongad:
What's the long-term plan for this? If we know API sometimes return invalid-argument, shouldn't we take catch it and return false?

garrettjonesgoogle:
We should probably file an issue with the API team; I feel like the service should throw the same error given the same inputs. I'm not actually sure what the correct behavior is, so I don't want to make a change to the client yet.

garrettjonesgoogle:
I have filed an internal bug. @neozwu could you take a look?

PR #2873, opened 2018-02-07
bigquery: add table label

### LINKED ISSUE #2865: BigQuery table info is missing Labels
The big query table info is missing the labels map

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/b7b3f21b784a3524f7830f34148abfd3e39677d4/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java

It is present on the dataset and on the underlying model object so it seems to just have been missed out on the mapping

pongad:
Fixes #2865.

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/LabelsTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/LabelsTest.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/TableInfoTest.java
```diff
@@ -241,7 +242,7 @@ private void compareTableInfo(TableInfo expected, TableInfo value) {
     assertEquals(expected.getGeneratedId(), value.getGeneratedId());
     assertEquals(expected.getLastModifiedTime(), value.getLastModifiedTime());
     assertEquals(expected.getSelfLink(), value.getSelfLink());
-    assertEquals(expected.getDefinition(), value.getDefinition());
+    assertEquals(expected.getLabels(), value.getLabels());
```
pongad: Definition already compared on L236.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableInfo.java
```diff
@@ -227,6 +241,12 @@ public Builder setDefinition(TableDefinition definition) {
       return this;
     }
 
+    @Override
+    public Builder setLabels(Map<String, String> labels) {
+      this.labels = ImmutableMap.copyOf(labels);
```
tswast: ImmutableMap won't work here. We want folks to be able to explicitly set a label to `null`, because this is the way to unset a label when using a `PATCH` request.

tswast:
> I see some EncryptionConfiguration properties snuck in too. :-)

I guess I was just seeing the merge commit.

pongad:
Yup those are commit merges :)

PR #2877, opened 2018-02-07
Adding basic builds to CircleCI config

garrettjonesgoogle:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2364

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ jobs:
       - image: maven:3.5.2-jdk-7
         environment:
           TERM: dumb
+          MAVEN_OPTS: -Xmx256m
     steps:
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
@pongad PTAL

I will disable Travis completely in a separate PR, because Circle will now subsume Travis' work.

.circleci/config.yml
```diff
@@ -1,5 +1,13 @@
 version: 2.0
 
+anchor_build: &anchor_build
+  name: Build
+  command: mvn -B -DtrimStackTrace=false -fae --quiet -DskipTests install
+
+anchor_unittest: &anchor_unittest
+  name: Run unit tests
+  command: mvn -B -DtrimStackTrace=false -fae --quiet -DskipITs install -P release
```
pongad: Do we actually need anchor_build? IIUC, we always build then unittest right after, so we might as well not do anchor_build and just unittest?

garrettjonesgoogle: It produces a nice result on the CircleCI build page - you can see at a glance if the build passed and the unit tests failed, or if the build failed before it got to the unit tests. A failure of the build phase is a very strong signal that the pull requester messed up :-)

PR #2887, opened 2018-02-09
bigquery: try getting job if create fails

### LINKED ISSUE #2867: BigQuery: Job already exists, probably for internal backoff or something
Hi, 
I'm currently using library 0.33.0-beta for an App Engine Std project,

When I need to execute a QueryJob, I have this method
```
public static TableResult executeQuery(QueryJobConfiguration queryConfig) throws IOException {
	BigQuery service = getService();

	JobId jobId = JobId.of(UUID.randomUUID().toString() + "-" + System.nanoTime());
	Job queryJob = service.create(JobInfo.newBuilder(queryConfig).setJobId(jobId).build());

	// Wait for job to finish
	queryJob = job.waitFor(retryOption);

	// Get the results
	TableResult result = queryJob.getQueryResults();
	
	// TODO Do something with the results...
}
```

You can see from the code itself the JobID is unique, due it has a random part and a time part.

But sometimes I get this error:
```
{
  "code": 409,
  "errors": [
    {
      "domain": "global",
      "message": "Already Exists: Job xxxxxxxxxxxxxxxx:EU.80760581-e8df-4c87-a538-1efbcf917778-1517936402050522782",
      "reason": "duplicate"
    }
  ],
  "message": "Already Exists: Job  xxxxxxxxxxxxxxxx:EU.80760581-e8df-4c87-a538-1efbcf917778-1517936402050522782"
}
```

and here is the stacktrace
```
com.google.cloud.bigquery.BigQueryException: Already Exists: Job xxxxxxxxxxxxxxxx:EU.80760581-e8df-4c87-a538-1efbcf917778-1517936402050522782
	at com.google.cloud.bigquery.spi.v2.HttpBigQueryRpc.translate(HttpBigQueryRpc.java:83)
	at com.google.cloud.bigquery.spi.v2.HttpBigQueryRpc.create(HttpBigQueryRpc.java:167)
	at com.google.cloud.bigquery.BigQueryImpl$3.call(BigQueryImpl.java:186)
	at com.google.cloud.bigquery.BigQueryImpl$3.call(BigQueryImpl.java:183)
	at com.google.api.gax.retrying.DirectRetryingExecutor.submit(DirectRetryingExecutor.java:89)
	at com.google.cloud.RetryHelper.run(RetryHelper.java:74)
	at com.google.cloud.RetryHelper.runWithRetries(RetryHelper.java:51)
	at com.google.cloud.bigquery.BigQueryImpl.create(BigQueryImpl.java:183)
	--- my class ---
```

Because I'm sure that my code generates every time a random/unique JobId (even if i Try multiple times the same query the id will be different) I'm wondering the conditions on how this kind of error is genereated.

I'm open this issue because I'm thinking about an internal backoff functionality (I can see the method **runWithRetries** from the stacktrace, this is why i'm suggesting this), where the jobs.create call ends w/o response (so the backoff try again to run the job) but the job successfully started the first time. In this way, trying to start a new job (same id) a second time, will produce this error.

It does not happens very often, so probably are temporary errors related to BigQuery infrastructure (and because of this is hard to replicate behaviour).

Is this a known behaviour?

pongad:
If we create a job with randomly generated ID, it's highly unlikely
that it will with another job created by someone else.
When create fails, it is possible that the job was created successfully
anyway.
Thus, we can be more failure-tolerant by trying to retrieve the job,
if create throws.

Fixes #2867.

### DIFF
--- a/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
+++ b/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
@@ -912,11 +912,11 @@ public class BigQueryImplTest {
         .andReturn(newJobPb());
     EasyMock.replay(bigqueryRpcMock);
 
-    BigQuery.JobOption options = BigQuery.JobOption.fields(BigQuery.JobField.USER_EMAIL);
+    BigQuery.JobOption jobOptions = BigQuery.JobOption.fields(BigQuery.JobField.USER_EMAIL);
 
     bigquery = options.getService();
-    bigquery.create(newJobPb(), options);
-    String selector = (String) capturedOptions.getValue().get(options.getRpcOption());
+    bigquery.create(JobInfo.of(QueryJobConfiguration.of("SOME QUERY")), jobOptions);
+    String selector = (String) capturedOptions.getValue().get(jobOptions.getRpcOption());
 
     // jobReference and configuration are always sent; the RPC call won't succeed otherwise.
     assertThat(selector.split(","))


google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -874,71 +877,95 @@ public void testListTableDataWithOptions() {
     assertArrayEquals(TABLE_DATA.toArray(), Iterables.toArray(page.getValues(), List.class));
   }
 
-  @Test
-  public void testCreateQueryJob() {
```
pongad: Note: these old tests are essentially change detectors. They don't really do much beyond calling `toPb` twice gives the same result.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -175,19 +177,68 @@ public Table create(TableInfo tableInfo, TableOption... options) {
...
+    }
+
+    // If create RPC fails, it's still possible that the job has been successfully created,
+    // and get might work.
+    // We can only do this if we randomly generated the ID. Otherwise we might mistakenly
+    // fetch a job created by someone else.
+    try {
+      return getJob(jobInfo.getJobId());
+    } catch (BigQueryException e) {
+      System.err.println("rethrow");
```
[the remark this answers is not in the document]
tswast: Leftover println from debugging?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -590,7 +641,8 @@ public Boolean call() {
   @Override
   public TableResult query(QueryJobConfiguration configuration, JobOption... options)
       throws InterruptedException, JobException {
-    return query(configuration, JobId.of(), options);
+    Job.checkNotDryRun(configuration, "query");
```
tswast: Is the reason for checkNotDryRun that a TableResult is not valid / useful in this case? Is this limitation documented in the interface?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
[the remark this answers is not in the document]
tswast: Leftover println from debugging?

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -874,71 +877,95 @@ public void testListTableDataWithOptions() {
...
+    String id = "testCreateJobSuccess-id";
+    JobId jobId = JobId.of(id);
+    String query = "SELECT * in FOO";
+
+    Capture<com.google.api.services.bigquery.model.Job> jobCapture = EasyMock.newCapture();
     EasyMock.expect(
-            bigqueryRpcMock.create(
-                JobInfo.of(LOAD_JOB_CONFIGURATION_WITH_PROJECT).toPb(), EMPTY_RPC_OPTIONS))
-        .andReturn(COMPLETE_LOAD_JOB.toPb());
+            bigqueryRpcMock.create(EasyMock.capture(jobCapture), EasyMock.eq(EMPTY_RPC_OPTIONS)))
```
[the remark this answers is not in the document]
tswast: Aside: I find this warning really annoying (harder to review the PR). I wonder if we should remove those static imports? I like your style here better with the explicit EasyMock.capture(...).

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -874,71 +877,95 @@ public void testListTableDataWithOptions() {
...
-        .andReturn(COMPLETE_COPY_JOB.toPb());
+                EasyMock.anyObject(com.google.api.services.bigquery.model.Job.class),
+                EasyMock.capture(capturedOptions)))
+        .andReturn(newJobPb());
     EasyMock.replay(bigqueryRpcMock);
+
     bigquery = options.getService();
-    Job job = bigquery.create(COPY_JOB);
-    assertEquals(new Job(bigquery, new JobInfo.BuilderImpl(COMPLETE_COPY_JOB)), job);
+    Job job = bigquery.create(QUERY_JOB, JOB_OPTION_FIELDS);
```
tswast: It's not so good to have a test that depends on a value defined elsewhere. Could you set the options that you're looking for below here?

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -874,71 +877,95 @@ public void testListTableDataWithOptions() {
...
+    String selector = (String) capturedOptions.getValue().get(JOB_OPTION_FIELDS.getRpcOption());
+    assertThat(selector.split(","))
+        .asList()
+        .containsExactly("jobReference", "configuration", "user_email");
```
tswast: I only see USER_EMAIL in JOB_OPTION_FIELDS, I guess jobReference and configuration are required for correct behavior of the method? We should have a comment in this test for that / adjust the test name.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -175,19 +177,68 @@ public Table create(TableInfo tableInfo, TableOption... options) {
 
   @Override
   public Job create(JobInfo jobInfo, JobOption... options) {
+    Supplier<JobId> idProvider =
+        new Supplier<JobId>() {
+          @Override
+          public JobId get() {
+            return JobId.of();
```
tswast: Isn't there a way to pass in a prefix for the JobId, too? I recall that was a requirement from the BQ team. Or is there already a supplier for that?

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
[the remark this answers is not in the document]
pongad: I agree. Sending an email to you and team about this.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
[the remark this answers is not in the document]
pongad: Thank you. Done

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
pongad: The primary problem is that dry-run doesn't execute the query, so there's no table to return from the API. We'd essentially be forced to either return a null or artificially-empty table. I thought failing explicitly made more sense.

Clarified this in doc.

google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
pongad: Done.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
pongad: According to the docs, the recommended way is `JobId.of("my-prefix-" + UUID.randomUUID())`. Unfortunately it doesn't benefit from this change.

This is a good point, but I'm not sure what the right design is here.

pongad:
@tswast PTAL

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
tswast: Perhaps let’s follow up in a separate issue/PR and add another static constructor to JobId with a prefix, maybe `JobId.withPrefix(String)`

pongad: In current design, this would still be considered "provided JobId" though and we won't be able to call `get` if `create` fails. Are you OK with this?

If we want `JobId.withPrefix` to benefit from this, we'd actually need the `JobId` to know whether it was randomly generated.

tswast: You're right. `withPrefix` wouldn't solve the problem. Rather, we'd want to be able to provide a prefix to the factory / supplier constructor.

Issue #1370, opened 2016-11-03
bigquery: support loading a POJO or Map like GSON does


tswast:
The code for processing the results of a query is a bit convoluted.

Example from https://github.com/GoogleCloudPlatform/java-docs-samples/blob/2e3f285b7dccfd978b14033f9869e728e8c13e56/bigquery/cloud-client/src/main/java/com/example/bigquery/SimpleApp.java#L53

```
    QueryResult result = response.result();

    while (result != null) {
      Iterator<List<FieldValue>> iter = result.iterateAll();

      while (iter.hasNext()) {
        List<FieldValue> row = iter.next();
        List<FieldValue> titles = row.get(0).repeatedValue();
        System.out.println("titles:");

        for (FieldValue titleValue : titles) {
          List<FieldValue> titleRecord = titleValue.recordValue();
          String title = titleRecord.get(0).stringValue();
          long uniqueWords = titleRecord.get(1).longValue();
          System.out.printf("\t%s: %d\n", title, uniqueWords);
        }

        long uniqueWords = row.get(1).longValue();
        System.out.printf("total unique words: %d\n", uniqueWords);
      }

      result = result.nextPage();
```

For a fixed query, I'd love to be able to fill in the object the way that I can in GSON. Then I'd be able to do something like:

```
    QueryResult result = response.result();

    while (result != null) {
      Iterator<List<FieldValue>> iter = result.iterateAll();

      while (iter.hasNext()) {
        List<FieldValue> row = iter.next();
        // I can't do this yet, but I'd love to be able to load an object
        // like I can with GSON.
        // I'm thinking maybe it only makes sense per-row rather than for
        // a whole query, due to paging.
        WordCounts counts = row.as(WordCounts.class);
        System.out.println("titles:");

        for (Title title : counts.getTitles()) {
          System.out.printf("\t%s: %d\n", title.getTitle(), title.getUniqueWords());
        }

        System.out.printf("total unique words: %d\n", counts.getTotalWords());
      }

      result = result.nextPage();
```

For context, here is a similar issue in the Go libraries: https://github.com/GoogleCloudPlatform/google-cloud-go/issues/399

mziccard:
@tswast yeah this would be nice to have. It's in the roadmap but we haven't had time to work on it yet.

danoscarmike:
[...]

PR #2891, opened 2018-02-12
bigquery: make FieldValue publicly construct-able

### LINKED ISSUE #2880: FieldValue.of is not public
For some reason FieldValue "of" method is not "public static" which means I can't create an instance of FieldValue. I can create instances of other, similar objects like Field, Schema, etc. though.

This is major problem, because it means you can't create bigquery result objects for things like testing. Could we please make this available?

pongad:
so make testing easier for users. Fixes #2880.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FieldValue.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
 import com.google.api.client.util.Data;
+import com.google.api.core.BetaApi;
 import com.google.common.base.MoreObjects;
 import com.google.common.io.BaseEncoding;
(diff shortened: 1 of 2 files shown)


pongad:
@vam-google Are you OK with this being public?

tswast:
Hmm, I think something like https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1370 would better cover the desired use case for testing specific query / table values than exposing this as public would. Albeit, that is a bigger project than changing visibility.

pongad:
I agree; doing that would definitely be better. I guess there are 2 options.

1. We can go ahead with this PR, and mark the newly-public methods `@BetaApi`. When POJO works, we deprecate them.
2. We close #2880 as dup of #1370, which is now tracked by feature wiki.

I'll defer to you on the decision. However, I'm slightly leaning towards (1). My reasoning is that, right now, users have to deal `FieldValueList.get` anyway. If you know what to expect from `get`, calling `of` isn't too terrible.

tswast:
Yeah, (1) makes sense. I agree to keep it as `@BetaApi` for now.

PR #2894, opened 2018-02-12
datastore: document DatastoreWriter.update exception

### LINKED ISSUE #2892: documentation issues: for transaction.update()
https://googlecloudplatform.github.io/google-cloud-java/0.34.0/apidocs/com/google/cloud/datastore/DatastoreWriter.html#update-com.google.cloud.datastore.Entity...-

the methods return value is void and it is documented to not throw anything. Still the documentation says "The operation will fail if an entity with the same key does not already exist.".

How will it fail?

pongad:
Fixes #2892.

@wsh please take a look


ensonic:
Thanks!

PR #2903, opened 2018-02-14
Accept strings in where(FieldPath.documentId(), ...)

schmidt-sebastian:
We currently support Strings and DocumentReferences for query.orderBy(FieldPath.documentId()).startAt(...), but only DocumentReferences for the corresponding where() call.

Addresses: https://stackoverflow.com/questions/48788805/query-by-documentid-in-firebase-function

Port of https://github.com/googleapis/nodejs-firestore/pull/136


schmidt-sebastian:
@jba - for potential tests cases :)

PR #2905, opened 2018-02-14
Make download in StorageExamples more straightforward

### LINKED ISSUE #2899: External user feedback regarding the GCS Download Object sample
This feedback came to the GCS Docs as bug 68217327.

The feedback applies to the Java sample for downloading objects from GCS, found at https://cloud.google.com/storage/docs/object-basics#download

The body of the feedback was:
"The example to download an object already has a populated blobId. What is a blobId? How do I make one?"

I believe they're referring to the fact that the download sample takes a BlobId parameter and assumes that users of the sample will know what a BlobId is.

yihanzhen:
DownloadAction example originally takes `blobId` as an argument which caused some misunderstanding. Change the signature to `run(Storage storage, String bucketName, String objectName, Path downloadTo)` to make the example more straightforward. This fixes #2899


hzyi-google:

PR #2906, opened 2018-02-15
Add documentation to FakeStorageRpc and LocalStorageHelper

### LINKED ISSUE #2902: Method listing the contents of a bucket in FakeStorageRpc does not use bucket name
Hello,

The implementation of the method `list(String bucket, Map<Option, ?> options)` in `FakeStorageRpc` does not use `bucket` parameter to retrieve the files inside the bucket passed as parameter (see [FakeStorageRpc.java](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/v0.34.0/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java#L122)). I'm using version 0.34.0, but it seems like in previous versions the behaviour was similar. Am I missing something ?

As a consequence, when using `LocalStorageHelper` (using `FakeStorageRpc`) for local testing purposes, the following code does not retrieve only the files in the bucket passed as parameter, but all files in all created buckets : 

```
Storage storage = LocalStorageHelper.customOptions(true).getService()

[...] // creating dummy files in two separate buckets (b1 and b2)

storage.list("b1") // returns all files in b1 and b2
```

I double checked in the doc, listing the contents of a bucket is however [defined as a supported operation](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/v0.34.0/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java#L53).

yihanzhen:
Document on method list(String bucket, Map options) since its behavior
is not implied but the method itself.
Fixes #2902

### DIFF
--- a/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
+++ b/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
@@ -117,7 +117,7 @@ class FakeStorageRpc implements StorageRpc {
   public Tuple<String, Iterable<Bucket>> list(Map<Option, ?> options) throws StorageException {
     throw new UnsupportedOperationException();
   }
-  
+
   @Override
   public Tuple<String, Iterable<StorageObject>> list(String bucket, Map<Option, ?> options)
       throws StorageException {


google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
```diff
@@ -41,7 +41,7 @@
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
- * A bare-bones in-memory implementation of Storage, meant for testing.
+ * A bare-bones in-memory implementation of StorageRpc, meant for testing of CloudStorageFileSystem.
```
pongad: Two things:

1. I don't think you need the "of CloudStorageFileSystem" here. It's OK to use this to test other things if you don't care about buckets.

2. I think this is a wrong place to document this; the class is package-private. I think it's better to document [this](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/LocalStorageHelper.java) instead, since more people are likely to see it.

hzyi-google:
PTAL @pongad @garrettjonesgoogle

pongad:
LGTM.

Issue #2796, opened 2018-01-20
LoggingHandler hangs the thread


sparhomenko:
Our application is running on Google Kubernetes Engine using `gcr.io/google-appengine/jetty` image and used `com.google.cloud.logging.LoggingHandler` to publish logs on Stackdriver. We noticed some worker threads becoming unresponsive over time. When the pod is shutting down we can see the following exception for each:
```java
java.lang.RuntimeException: java.lang.InterruptedException
	at com.google.cloud.logging.LoggingImpl.flush(LoggingImpl.java:545)
	at com.google.cloud.logging.LoggingImpl.write(LoggingImpl.java:525)
	at com.google.cloud.logging.LoggingHandler.publish(LoggingHandler.java:273)
	at java.util.logging.Logger.log(Logger.java:738)
	at org.slf4j.impl.JDK14LoggerAdapter.log(JDK14LoggerAdapter.java:582)
	at org.slf4j.impl.JDK14LoggerAdapter.error(JDK14LoggerAdapter.java:500)
        ...
Caused by: java.lang.InterruptedException
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:449)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:79)
	at com.google.common.util.concurrent.ForwardingFuture.get(ForwardingFuture.java:63)
	at com.google.cloud.logging.LoggingImpl.flush(LoggingImpl.java:543)
	... 30 more
```
We'll try to extract a thread dump to see why the future never completes, but the issue seems dangerous by itself: `LoggingImpl.java:543` uses the non-timeout version of `Future.get()` which can cause any logger call to block the current thread forever unless interrupted. Would it be possible to use the timeout version with a reasonably big timeout, e.g. 60 seconds?

pongad:
This seems related to #1795. The underlying client provides no way to initiate a flush.

@sparhomenko Do you have a sense of how long we're waiting? The above bug might cause us to wait a few seconds, but it shouldn't cause a prolonged hang. I agree that the wait with no timeout is a problem, but I want to understand the problem better before sending fixes.

pongad:
@sparhomenko Yikes! 2.5 hours is WAY beyond reasonable. I now suspect some kind of propagation problem somewhere. Maybe we don't properly complete the future for something. I'll try to dig into this as well, but a repro would be greatly appreciated.

tamersaleh:
[...]

PR #2910, opened 2018-02-15
RFC: logging: experimental batching writer

pongad:
Manually written batching implementation. Unlike GAX batching,
it implements flushing and does not deal with partition keys.

If we're OK with where this is heading,
we should make this work with batching settings (shouldn't be hard)
and load-test before migrating things to it.

If not, this can serve as a starting point for a better GAX batching
implementation.

In an experiment, I publish 1M messages of 300 bytes each.
Using LoggingHandler gives me ~14,000 msg/s.
BatchingWriter gives ~67,000, using similar BatchingSettings.
Letting users configure these settings might be important though;
I have observed ~300K msg/s given enough CPU and memory.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/BatchingWriter.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/BatchingWriter.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.logging;
+
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutureCallback;
+import com.google.api.core.ApiFutures;
+import com.google.common.base.Preconditions;
+import com.google.logging.v2.WriteLogEntriesRequest;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+
+class BatchingWriter {
+
+  interface Rpc {
+    ApiFuture<Void> call(WriteLogEntriesRequest request);
+  }
+
+  private final Rpc rpc;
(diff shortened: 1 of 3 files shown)


pongad:
Updates #2796

google-cloud-logging/src/main/java/com/google/cloud/logging/BatchingWriter.java
```diff
@@ -0,0 +1,152 @@
...
+      currentBatch.add(entry);
+      if (currentBatch.size() == batchSize) {
+        final WriteLogEntriesRequest request =
+            requestPrototype.toBuilder().addAllEntries(currentBatch).build();
+        currentBatch.clear();
+
+        // Whoever calls send serializes the proto; so we do it off-thread.
+        // This gives better CPU utilization if there are few producer threads
+        // on a many-core machine.
+        executor.execute(
```
pongad: Actually, a little bug here. If we add, then initFlush, initFlush might happen before send. A fix: create a "surrogate" future and put it in the map before unlocking. I'll perf test this.

igorbernstein2:
I still think that we should try refactoring gax code instead of writing batching code from scratch. There are a lot of useful features in the gax implementation that are lost here.

Before the bigtable client is ready I will need the following features in batching:
- request byte size flow control
- ability to know when either the the whole batch request failed or when an entry failed
- ability to retry individual entries

General feedback on the PR:
- add() should return a future so that the caller can find out the result of the write
- writer should be closeable & throw an error if any rpcs failed
- It would be nice to provide a means to do nonblocking back pressure: adding an bool isReady() & onReady(Callback) on the Writer
- there should be a flush() method that can send all pending RPCs and wait for all of them to return

pongad:
@igorbernstein2 Thank you for your thoughts. I agree that putting this in gax will be valuable. I'm mostly using this to figure out what logging needs from whatever ends up in gax. This is why it's an RFC after all.

To answer your points:

> request byte size flow control

I think this is easily done by changing Semaphore to FlowController right?

> ability to know when either the the whole batch request failed or when an entry failed
> ability to retry individual entries
> add() should return a future so that the caller can find out the result of the write

Please correct me if I'm wrong. I think these 3 are somewhat related. At least it's predicated on the ability to identify individual requests from a batch. Pubsub needs this too.

However, logging doesn't (writes are all or nothing, and failures are reported to error manager). I haven't measured how much CPU work this would cost. If it's significant, we should find a way for logging to not pay for this cost.

> writer should be closeable & throw an error if any rpcs failed

I don't think this makes sense to logging. [Doc](https://docs.oracle.com/javase/7/docs/api/java/util/logging/ErrorManager.html) explicitly discourages throwing exceptions.

> there should be a flush() method that can send all pending RPCs and wait for all of them to return

Definitely makes sense as a convenience method. I believe it can be implemented by
```java
initFlush();
for (Future<?> future : pendingRpcs()) future.get(); // error handling elided.
```

> It would be nice to provide a means to do nonblocking back pressure: adding an bool isReady() & onReady(Callback) on the Writer

Does BigTable require single- or multiple-producer per writer? I think isReady makes sense in single, but could cause thundering herd problem in multiple. I need to think more about onReady.

pongad:
@garrettjonesgoogle Are you OK with me submitting this as-is to a branch?

pongad:
I'll merge this into a branch

PR #2915, opened 2018-02-16
Bigtable: bug fix - Make ranges mutable

igorbernstein2:
Range are used with filters pretty heavily, and the filters dsl is mutable. So range should be mutable to be consistent.


igorbernstein2:
Bigtable veneer client's very first bugfix. I think this issue should be bronzed :)

PR #2919, opened 2018-02-17
Bigtable: cleanup - tweak filters

igorbernstein2:
Minor filter DSL updates as a result of their usage in the current bigtable client

- add a `raw` filter that wraps a protobuf
- remove `wrapString` and use `ByteString. copyFromUtf8()` directly 
- misc cleanups

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/wrappers/FiltersTest.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/wrappers/FiltersTest.java
@@ -66,12 +66,7 @@ public class FiltersTest {
     RowFilter actualProto = FILTERS.chain().toProto();
 
     RowFilter expectedFilter =
-        RowFilter.newBuilder()
-            .setChain(
-                Chain.newBuilder()
-                    .addFilters(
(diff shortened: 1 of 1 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/wrappers/Filters.java
```diff
@@ -151,6 +149,15 @@ public LimitFilter limit() {
   }
 
   // Miscellaneous filters without a clear target.
+  /**
+   * Wraps protobuf representation of a filter.
+   *
+   * <p>For advanced use only.
+   */
+  public Filter raw(RowFilter rowFilter) {
```
pongad: This is going "raw" to "nice" right? In that case, maybe call this "fromRaw" instead? The name "raw" seems to imply "raw representation of the argument".

igorbernstein2: You are right, but maybe `fromProto` is better? If you agree, please merge :)

igorbernstein2:
Sorry last minute addition to support existing client:
https://github.com/sduskis/cloud-bigtable-client/commit/8f58d502a0dbc052ccf2bd1f74868c7618bd19b1

PR #2925, opened 2018-02-20
Bigtable - improvement: Rewrite ReframingResponseObserver

igorbernstein2:
The new version uses atomics to ensure mutual exclusion instead of synchronization.
This allows the delivery loop to be simpler since there is no need to need to defer actions until after exiting the synchronization.
Also, move all interactions with the Reframer to the delivery loop, which ensures proper error handling if the Reframer detects an error.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
@@ -20,6 +20,7 @@ import com.google.api.gax.rpc.ResponseObserver;
 import com.google.api.gax.rpc.StateCheckingResponseObserver;
 import com.google.api.gax.rpc.StreamController;
 import com.google.common.base.Preconditions;
+import com.google.common.math.IntMath;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
@@ -159,10 +160,8 @@ public class ReframingResponseObserver<InnerT, OuterT>
       if (current == Integer.MAX_VALUE) {
         return;
       }
-      // Prevent overflow
-      int maxRequest = Integer.MAX_VALUE - current;
-      int newValue = maxRequest > count ? current + count : Integer.MAX_VALUE;
 
+      int newValue = IntMath.saturatedAdd(current, count);
       if (numRequested.compareAndSet(current, newValue)) {
         break;
       }
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserverTest.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserverTest.java
@@ -320,7 +320,7 @@ public class ReframingResponseObserverTest {
         new DasherizingReframer(1) {
           @Override
           public void push(String response) {
-            if (response.equals("boom")) {
+            if ("boom".equals(response)) {
               throw new IllegalStateException("fake error");
             }
             super.push(response);
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/gaxx/testing/FakeStreamingApi.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/gaxx/testing/FakeStreamingApi.java
@@ -120,7 +120,7 @@ public class FakeStreamingApi {
     private ResponseObserver<ResponseT> actualObserver;
     private RequestT actualRequest;
     private List<ResponseT> responseList;
-    private BlockingQueue<StreamControllerStash<ResponseT>> calls = Queues.newLinkedBlockingQueue();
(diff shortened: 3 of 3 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
```diff
@@ -158,11 +154,18 @@ private void onRequest(int count) {
...
 
-    synchronized (lock) {
-      int maxCount = Integer.MAX_VALUE - numPending;
-      count = Math.min(maxCount, count);
+    while (true) {
+      int current = numRequested.get();
+      if (current == Integer.MAX_VALUE) {
+        return;
+      }
+      // Prevent overflow
```
pongad: Consider using this? http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/math/IntMath.html#saturatedAdd-int-int-

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
```diff
@@ -209,12 +217,9 @@ protected void onResponseImpl(InnerT response) {
    */
   @Override
   protected void onErrorImpl(Throwable t) {
-    synchronized (lock) {
-      if (error == null) {
-        error = t;
-      }
-      closeOnDone = true;
-    }
+    // order of assignment matters
```
pongad: I don't completely understand why this isn't synchronized anymore.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
```diff
@@ -226,155 +231,150 @@ protected void onErrorImpl(Throwable t) {
...
+
+    // Try to move the new item into the reframer
+    InnerT newUpstreamItem = newItem.getAndSet(null);
+    if (newUpstreamItem != null) {
+      reframer.push(newUpstreamItem);
+      // and reset the awaiting flag, if the item arrived or upstream closed
+      awaitingInner = false;
+    } else if (localDone) {
+      awaitingInner = false;
+    }
```
pongad: Isn't this just

```java
if (newUpstreamItem != null) {
  reframer.push(newUpstreamItem);
}
awaitingInner = false;
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
igorbernstein2: Only the grpc thread can set this value and it always does this before writing the volatile `done` which ensure cache coherency for whatever thread might be in the delivery loop.

Previously `cancel` would share the `error` variable which meant that there were 2 writers: grpc + an application thread explicitly setting it via cancel. This is longer the case as the cancellation uses its own Atomic

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
igorbernstein2: I don't think so. Though you might be seeing something I'm not. The general order of operations in delviery is:

1. check if we are awaitingInner, which we aren't
2. notice that delivered < demand
3. ask the reframer for a full frame, which it would not have
4. check that we aren't awaitingInner, & request(1) from inner controller
5. go back to the top of the outer loop
6. here we have a race condition: grpc might or might have called onResponse and/or onComplete
  a. if it didn't, then awaitingInner is true and should remain so
  b. if it did, then we have a newupstreamItem and we should reset awaitingInner
  c. if grpc called onComplete, then awaitinginner should be reset as well

automatically setting awaitingInner=false would neglect 6a

igorbernstein2:
All feedback should be addressed, PTAL

igorbernstein2:
rebased

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
pongad: Got it!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/reframing/ReframingResponseObserver.java
pongad: Right sorry. Wasn't thinking.

pongad:
LGTM. @garrettjonesgoogle Do you want to take a look? If not, we can merge.

garrettjonesgoogle:
This looks like a mostly implementation change, so I'll defer to @pongad 's LGTM.

PR #2936, opened 2018-02-22
Bigtable: 05. Surface: Bulk Mutations

igorbernstein2:
Exposes the MutateRows RPC as bulkMutations(). The actual BatchingCallable is wrapped by a BulkMutations wrapper to allow waiting for outstanding mutations and eventually for flushing the batch.

Edit: This has been reviewed by @kevinsi4508 and ready for @garrettjonesgoogle 

NOTE: this api assumes the current gax batching infrastructure. It will be extended once flushing is figured out

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/BigtableDataClientTest.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/BigtableDataClientTest.java
@@ -63,9 +63,9 @@ public class BigtableDataClientTest {
     Mockito.when(mockStub.readRowsCallable()).thenReturn(mockReadRowsCallable);
     Mockito.when(mockStub.sampleRowKeysCallable()).thenReturn(mockSampleRowKeysCallable);
     Mockito.when(mockStub.mutateRowCallable()).thenReturn(mockMutateRowCallable);
+    Mockito.when(mockStub.mutateRowsCallable()).thenReturn(mockBulkMutateRowsCallable);
     Mockito.when(mockStub.checkAndMutateRowCallable()).thenReturn(mockCheckAndMutateRowCallable);
     Mockito.when(mockStub.readModifyWriteRowCallable()).thenReturn(mockReadModifyWriteRowCallable);
-    Mockito.when(mockStub.mutateRowsCallable()).thenReturn(mockBulkMutateRowsCallable);
   }
 
   @Test
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
@@ -98,9 +98,9 @@ public class EnhancedBigtableStubSettings extends StubSettings<EnhancedBigtableS
   private final ServerStreamingCallSettings<Query, Row> readRowsSettings;
   private final UnaryCallSettings<String, List<KeyOffset>> sampleRowKeysSettings;
   private final UnaryCallSettings<RowMutation, Void> mutateRowSettings;
+  private final BatchingCallSettings<RowMutation, Void> mutateRowsSettings;
   private final UnaryCallSettings<ConditionalRowMutation, Boolean> checkAndMutateRowSettings;
   private final UnaryCallSettings<ReadModifyWriteRow, Row> readModifyWriteRowSettings;
-  private final BatchingCallSettings<RowMutation, Void> mutateRowsSettings;
 
   private EnhancedBigtableStubSettings(Builder builder) {
     super(builder);


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutations.java
```diff
@@ -0,0 +1,137 @@
...
+    Preconditions.checkState(!closed, "BulkMutations has been closed");
+    numOutstanding.incrementAndGet();
+
+    ApiFuture<Void> future = callable.futureCall(rowMutation);
+
+    ApiFutures.addCallback(
+        future,
+        new ApiFutureCallback<Void>() {
+          @Override
+          public void onFailure(Throwable throwable) {
```
kevinsi4508: I have commented something similar before. Don't we want to capture some errors here?
All users will see right now is number of failures and have no idea what the errors are.
If you are planning to add more detailed logging later, perhaps adding a TODO?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/BigtableDataClientTest.java
```diff
@@ -104,4 +115,97 @@ public void proxyMutateRowTest() {
...
+      bulkMutations.close(Duration.ofMillis(20));
+    } catch (Throwable t) {
+      error = t;
+    }
+    assertThat(error).isInstanceOf(TimeoutException.class);
+
+    // Resolve the request
+    innerResult.set(null);
+
+    // Now, close will promptly finish
```
kevinsi4508: Just "bulkMutations.close(Duration.ofMillis(20));" ?

igorbernstein2: fixed

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutations.java
igorbernstein2: I tried to address this in the javadoc for `close()`. Not all usecases need to track per item exceptions. For example beam only cares if a bundle failed as a whole. And it requires possibly unbounded resources. For example if the application is a long running stream that blindly writes to bigtable, then this class will leak memory by accumulating errors that never get checked.
Instead the user should opt-in to tracking the individual errors by tracking the futures returned from send. The coarse grained tracking done here is meant to use constant memory and is simply a safety net to remind the callers that there might data loss

igorbernstein2:
@garrettjonesgoogle, can you take a look when you have a moment?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -333,6 +334,33 @@ public void readRowsAsync(Query query, ResponseObserver<Row> observer) {
...
+   * Mutates multiple rows in a batch. Each individual row is mutated atomically as in MutateRow,
+   * but the entire batch is not executed atomically.
+   *
+   * <p>Sample code:
+   *
+   * <pre>{@code
+   * InstanceName instanceName = InstanceName.of("[PROJECT]", "[INSTANCE]");
+   * try (BigtableClient bigtableClient = BigtableClient.create(instanceName)) {
+   *   try (BulkMutations mutations = bigtableClient.newBulkMutations()) {
+   *     for(String someValue in someCollection) {
```
garrettjonesgoogle: Use `:` instead of `in`, add a space after `for`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -89,6 +90,13 @@ public String getAppProfileId() {
     return getTypedStubSettings().mutateRowSettings();
   }
 
+  /** Returns the object with the settings used for BulkMutations. */
+  public BatchingCallSettings<RowMutation, Void> bulkMutationsSettings() {
+    // NOTE: the user facing BigtableDataClient only exposes this api as newBulkMutations,
+    // so the outer name should match that. While the stub sticks to rpc names.
```
garrettjonesgoogle: s/should match that. While the stub sticks to/should match that, while the stub should stick to/ (and same down below)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/DummyBatchingDescriptor.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.api.gax.batching.PartitionKey;
+import com.google.api.gax.batching.RequestBuilder;
+import com.google.api.gax.rpc.BatchedRequestIssuer;
+import com.google.api.gax.rpc.BatchingDescriptor;
+import java.util.Collection;
+
+/**
+ * This is necessary workaround for {@link
+ * com.google.cloud.bigtable.data.v2.stub.EnhancedBigtableStub#mutateRowsCallable()}. The settings
+ * are exposed to the user using {@link com.google.cloud.bigtable.data.v2.models.RowMutation}
```
garrettjonesgoogle: s/using/using the/

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutations.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import org.threeten.bp.Duration;
+
+/**
+ * Tracker for outstanding bulk mutations. Allows for the caller to wait for all of the outstanding
+ * mutations to complete.
+ *
+ * @see com.google.cloud.bigtable.data.v2.BigtableDataClient#newBulkMutations() for example usage.
+ */
+public final class BulkMutations implements AutoCloseable {
```
garrettjonesgoogle: I get the feeling that this could be represented as a callable somehow - it seems to function like a client streaming callable.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -333,6 +334,33 @@ public void readRowsAsync(Query query, ResponseObserver<Row> observer) {
...
+   *       RowMutation mutation = RowMutation.create("[TABLE]", "[ROW KEY]")
+   *         .setCell("[FAMILY NAME]", "[QUALIFIER]", "[VALUE]");
+   *
+   *       mutations.send(mutation);
+   *     }
+   *   } catch (BulkMutationFailure failure) {
+   *     // Handle error
+   *   }
+   *   // After `mutations` are closed, all mutations have been applied
```
pongad: I'm not sure if this should be `is`. After all, `mutations` here is one variable, not actually many mutations.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutations.java
```diff
@@ -0,0 +1,137 @@
...
+    synchronized (lock) {
+      while (numOutstanding.get() > 0) {
+        long waitMs = deadlineMs - System.currentTimeMillis();
+        if (waitMs <= 0) {
+          throw new TimeoutException("Timed out waiting outstanding mutations to finish");
+        }
+        lock.wait(waitMs);
+      }
+
+      if (numFailures > 0) {
```
pongad: Is there a race for `numFailures` between this and `onMutationComplete`?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/DummyBatchingDescriptor.java
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutations.java
igorbernstein2: I don't think it fits as either:
- It's not quite a UnaryCallable because a sync `call()` doesn't make sense
- It's not quite a ClientStreamingCallable because each `send` returns a future, while client streaming callables don't make any assumptions about responses.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
igorbernstein2: I think you are right

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutations.java
igorbernstein2: I don't believe so. In `close()` `numFailures` is only checked after `numOutstanding` is zero. And in `onMutationComplete`, `numOutstanding` is only updated after updating `numFailures`.
I added some comments to clarify

igorbernstein2:
All feedback has been addressed. The only open question is if BulkMutations can be modeled in terms of callables. @garrettjonesgoogle, PTAL

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/DummyBatchingDescriptor.java
```diff
@@ -0,0 +1,72 @@
...
+ * com.google.bigtable.v2.MutateRowsRequest}s. This class is used as a placeholder for the settings
+ * and is replaced with the actual implementation of {@link
+ * com.google.cloud.bigtable.data.v2.stub.bulkmutaterows.BulkMutateRowsBatchingDescriptor} when
+ * constructing the callable chain.
+ *
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
+ */
+@InternalApi
+public class DummyBatchingDescriptor<ReqT, RespT> implements BatchingDescriptor<ReqT, RespT> {
```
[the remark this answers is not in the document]
garrettjonesgoogle: Could you spell out the parameters to match our other code? (`RequestT`, `ResponseT`)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutations.java
garrettjonesgoogle: Yes, you're correct about the mismatch with `ClientStreamingCallable`, which only has one response at the end.

Since this is making up for deficiencies in batching, let's try to match the naming of batching. I would suggest:

- `BulkMutations` -> `BulkMutationBatcher`
- `send()` -> `add()`
- `bigtableDataClient.newBulkMutations` -> `bigtableDataClient.newBulkMutationBatcher`

Since this might be replaced by a parameterized class in gax, let's mark it `@BetaApi` even though this library is still alpha. Maybe add text in the annotation like "This surface is likely to change as the batching surface evolves."

igorbernstein2:
rebased (and squashed to simplify rebasing). Will address feedback next

igorbernstein2:
Ok, all feedback has been addressed and the PR has been rebased. PTAL, if it looks good, please merge

pongad:
LGTM. @garrettjonesgoogle can +2

igorbernstein2:
Thank you!

PR #2940, opened 2018-02-22
Bigtable: 10. Implementation: integrate ReadRows

igorbernstein2:
This builds on PR #2914 by creating the ReadRows callable chain.

@kevinsi4508 can you take a look at this when you have a moment?

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
@@ -133,7 +133,8 @@ public class EnhancedBigtableStub implements AutoCloseable {
         filtering.withDefaultCallContext(clientContext.getDefaultCallContext());
 
     // NOTE: Ideally `withDefaultCallContext` should be the outer-most callable, however the
-    // ReadRowsUserCallable overrides the first() method which be lost. At some point in the future,
+    // ReadRowsUserCallable overrides the first() method. This override would be lost if
+    // ReadRowsUserCallable is wrapped by another callable.  At some point in the future,
     // gax-java should allow preserving these kind of overrides through callable chains, at which
     // point this should be re-ordered.
     return new ReadRowsUserCallable<>(withContext, requestContext);


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/FilterMarkerRowsCallable.java
```diff
@@ -0,0 +1,104 @@
...
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
+ */
+@InternalApi
+public class FilterMarkerRowsCallable<RowT> extends ServerStreamingCallable<ReadRowsRequest, RowT> {
+  private final RowAdapter<RowT> rowAdapter;
+  private final ServerStreamingCallable<ReadRowsRequest, RowT> innerCallable;
+
+  public FilterMarkerRowsCallable(
+      RowAdapter<RowT> rowAdapter, ServerStreamingCallable<ReadRowsRequest, RowT> inner) {
```
garrettjonesgoogle: We have developed a convention of having the inner callable be the first argument, so let's follow that here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsUserCallable.java
```diff
@@ -0,0 +1,57 @@
...
+  private final ServerStreamingCallable<ReadRowsRequest, RowT> inner;
+  private final RequestContext requestContext;
+  private final ReadRowsFirstCallable<RowT> firstCallable;
+
+  public ReadRowsUserCallable(
+      ServerStreamingCallable<ReadRowsRequest, RowT> inner, RequestContext requestContext) {
+    this.inner = inner;
+    this.requestContext = requestContext;
+
+    this.firstCallable = new ReadRowsFirstCallable<>(super.first());
```
garrettjonesgoogle: Could you add a note about why first() is being wrapped?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -103,15 +106,33 @@ public static EnhancedBigtableStub create(EnhancedBigtableStubSettings settings)
...
+
+    ServerStreamingCallable<ReadRowsRequest, RowT> merging =
+        new RowMergingCallable<>(stub.readRowsCallable(), rowAdapter);
+
+    FilterMarkerRowsCallable<RowT> filtering = new FilterMarkerRowsCallable<>(rowAdapter, merging);
+
+    ServerStreamingCallable<ReadRowsRequest, RowT> withContext =
+        filtering.withDefaultCallContext(clientContext.getDefaultCallContext());
+
+    return new ReadRowsUserCallable<>(withContext, requestContext);
```
garrettjonesgoogle: Actually I have one more comment - the default call context should be added to the outer-most callable.

igorbernstein2: Unfortunately I can't do that because of the special handling for `first()`.

AnonymousWithDefaultCallContextCallable#first() will return a FirstElementCallable,
whose call() will invoke AnonymousWithDefaultCallContextCallable#futureCall() which will bypass any custom FirstCallable I might set

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsUserCallable.java
igorbernstein2: will do, but to justify it in context of `withDefaultCallContext` comment below. It's mainly an optimization: to avoid unnecessary data from being buffered and to minimize the number of CANCELLED requests on the serverside

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
garrettjonesgoogle: Hmm that kind of bothers me - should we have a more first-class way of overriding first() that works in this case too? Otherwise, other implementers could get tricked because they didn't realize this fact.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsUserCallable.java
garrettjonesgoogle: Right, I gathered that from the docs in ReadRowsFirstCallable, but it's useful to know here too (this is where a code reader might care about it)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
igorbernstein2: I think this would need to be addressed in gax. Would you be ok with continuing this discussion in an issue on gax-java but merging this in the mean time?

igorbernstein2: I can also remove this optimization until we figure out a better solution. Let me know which works for you

garrettjonesgoogle: No I think retaining the optimization is good, otherwise we might forget to resolve the design issue and might never have the optimization. Could you put a note here about why the default call context can't be added to the outermost callable, and that resolving it requires design changes in gax?

igorbernstein2:
Ok, all feedback should now be addressed. Please merge when ready

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -132,6 +132,10 @@ public static EnhancedBigtableStub create(EnhancedBigtableStubSettings settings)
     ServerStreamingCallable<ReadRowsRequest, RowT> withContext =
         filtering.withDefaultCallContext(clientContext.getDefaultCallContext());
 
+    // NOTE: Ideally `withDefaultCallContext` should be the outer-most callable, however the
+    // ReadRowsUserCallable overrides the first() method which be lost. At some point in the future,
```
garrettjonesgoogle: Sorry, grammar to fix (do you want "which would be lost"?)

igorbernstein2: No, I'm sorry. My writing is terrible, thanks for catching that.

garrettjonesgoogle:
LGTM!

PR #2941, opened 2018-02-22
Bigtable: 13. Implement MutateRow

igorbernstein2:
@kevinsi4508 can you take a look at this when you have a moment?

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
@@ -22,8 +22,6 @@ import com.google.api.gax.rpc.Callables;
 import com.google.api.gax.rpc.ClientContext;
 import com.google.api.gax.rpc.ServerStreamingCallable;
 import com.google.api.gax.rpc.UnaryCallable;
-import com.google.bigtable.v2.MutateRowRequest;
-import com.google.bigtable.v2.MutateRowResponse;
 import com.google.bigtable.v2.ReadRowsRequest;
 import com.google.bigtable.v2.SampleRowKeysRequest;
 import com.google.bigtable.v2.SampleRowKeysResponse;


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -138,13 +146,18 @@ public void call(
...
-    return new UnaryCallable<RowMutation, Void>() {
-      @Override
-      public ApiFuture<Void> futureCall(RowMutation request, ApiCallContext context) {
-        throw new UnsupportedOperationException("todo");
-      }
-    };
+    UnaryCallable<MutateRowRequest, MutateRowResponse> withContext =
+        stub.mutateRowCallable().withDefaultCallContext(clientContext.getDefaultCallContext());
+
+    return new MutateRowCallable(withContext, requestContext);
```
garrettjonesgoogle: I think it's best to put the default call context on the outer-most callable.

igorbernstein2: All of the bigtable implementation PRs will have the same issue (to stay consistent with ReadRows). I'll update all of them tomorrow morning

garrettjonesgoogle: Just go ahead and add the same type of comment as you did for #2940 .

igorbernstein2: ReadRows is the only one that needs the special treatment. Since we are planning on removing that need in the future, I'd rather make this consistent with the autogen code and put withDefaultContext as the outermost callable. I just won't be able to do this until tomorrow morning

garrettjonesgoogle: Ahh got it, thanks!

garrettjonesgoogle:
No Igor still needs to make one more change, plus also rebase.

igorbernstein2:
rebased & made the default context the outer-most callable.
Please merge when ready

PR #2944, opened 2018-02-23
resourcemanager: bump dependency version

### LINKED ISSUE #2942: Upgrade google-api-services-cloudresourcemanager version to v1-rev6-1.22.0
Currently the BOM is specifying version `v1beta1-rev10-1.21.0`, which is pretty old and conflicts with current versions of libraries, such as `beam-runners-google-cloud-dataflow-java`.

See https://github.com/spring-cloud/spring-cloud-gcp/issues/442.

pongad:
Fixes #2942.

### DIFF
--- a/google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
+++ b/google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
@@ -53,7 +53,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Callable;
-import java.util.concurrent.CancellationException;
 import org.threeten.bp.Duration;
 
 public class HttpResourceManagerRpc implements ResourceManagerRpc {
@@ -65,13 +64,15 @@ public class HttpResourceManagerRpc implements ResourceManagerRpc {
   // https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/java/latest/com/google/api/services/cloudresourcemanager/CloudResourceManager.Projects.html#create(com.google.api.services.cloudresourcemanager.model.Project)
   private static final RetrySettings CREATE_RETRY_SETTINGS =
       RetrySettings.newBuilder()
-          // SLO permits 30s at 90th percentile, double it for total limit
-          .setTotalTimeout(Duration.ofMinutes(1))
+          // SLO permits 30s at 90th percentile, 4x it for total limit.
+          // Observed latency is much lower: 11s at 95th percentile.
+          .setTotalTimeout(Duration.ofMinutes(2))
+          // Linked doc recommends polling at 5th second.
           .setInitialRetryDelay(Duration.ofSeconds(5))
           .setRetryDelayMultiplier(1.5)
           // Observed P95 latency is 11s. We probably shouldn't sleep longer than this.
           .setMaxRetryDelay(Duration.ofSeconds(11))
-          .setJittered(false)
+          .setJittered(true)
           .setInitialRpcTimeout(Duration.ofSeconds(5))
           .setMaxRpcTimeout(Duration.ofSeconds(5))
           .build();
@@ -107,9 +108,12 @@ public class HttpResourceManagerRpc implements ResourceManagerRpc {
         }
 
         @Override
-        public boolean shouldRetry(Throwable prevThrowable, Operation prevOp)
-            throws CancellationException {
-          return prevThrowable != null || prevOp.getDone() == null || !prevOp.getDone();
+        public boolean shouldRetry(Throwable prevThrowable, Operation prevOp) {
+          if (prevThrowable == null) {
+            return prevOp.getDone() == null || !prevOp.getDone();
+          }
+          return prevThrowable instanceof ResourceManagerException
+              && ((ResourceManagerException) prevThrowable).isRetryable();
         }
       };
 
@@ -133,10 +137,8 @@ public class HttpResourceManagerRpc implements ResourceManagerRpc {
   }
 
   private static ResourceManagerException translate(Status status) {
-    int code;
-    if (RPC_TO_HTTP_CODES.containsKey(status.getCode())) {
-      code = RPC_TO_HTTP_CODES.get(status.getCode());
-    } else {
+    Integer code = RPC_TO_HTTP_CODES.get(status.getCode());
+    if (code == null) {
       code = BaseHttpServiceException.UNKNOWN_CODE;
     }
     return new ResourceManagerException(code, status.getMessage());
@@ -144,40 +146,44 @@ public class HttpResourceManagerRpc implements ResourceManagerRpc {
 
   @Override
   public Project create(Project project) {
+    final Operation operation;
     try {
-      final Operation operation = resourceManager.projects().create(project).execute();
+      operation = resourceManager.projects().create(project).execute();
+    } catch (IOException ex) {
+      throw translate(ex);
+    }
 
-      Operation finishedOp =
-          runWithRetries(
-              new Callable<Operation>() {
-                @Override
-                public Operation call() {
-                  try {
-                    return resourceManager.operations().get(operation.getName()).execute();
-                  } catch (IOException ex) {
-                    throw translate(ex);
-                  }
+    Operation finishedOp =
+        runWithRetries(
+            new Callable<Operation>() {
+              @Override
+              public Operation call() {
+                try {
+                  return resourceManager.operations().get(operation.getName()).execute();
+                } catch (IOException ex) {
+                  throw translate(ex);
                 }
-              },
-              CREATE_RETRY_SETTINGS,
-              OPERATION_HANDLER,
-              clock);
+              }
+            },
+            CREATE_RETRY_SETTINGS,
+            OPERATION_HANDLER,
+            clock);
       if (finishedOp.getError() != null) {
         throw translate(finishedOp.getError());
       }
 
-      // NOTE(pongad): Operation.getResponse() returns a Map<String, Object>.
-      // 1. `(Project) finishedOp.getResponse()` doesn't work,
-      // because JSON deserializer in execute() didn't know to create a Project object.
-      // 2. `new Project().putAll(finishedOp.getResponse())` doesn't work either.
-      // 64-bit integers are sent as strings in JSON,
-      // so execute(), not knowing the type, parses it as String, not Long.
+    // NOTE(pongad): Operation.getResponse() returns a Map<String, Object>.
+    // 1. `(Project) finishedOp.getResponse()` doesn't work,
+    // because JSON deserializer in execute() didn't know to create a Project object.
+    // 2. `new Project().putAll(finishedOp.getResponse())` doesn't work either.
+    // 64-bit integers are sent as strings in JSON,
+    // so execute(), not knowing the type, parses it as String, not Long.
+    try {
       String responseTxt = JSON_FACTORY.toString(finishedOp.getResponse());
       return JSON_FACTORY.fromString(responseTxt, Project.class);
     } catch (IOException ex) {
       throw translate(ex);
     }
-
   }
 
   @Override


google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,198 @@
...
   private static ResourceManagerException translate(IOException exception) {
     return new ResourceManagerException(exception);
   }
 
+  private static ResourceManagerException translate(Status status) {
+    int code;
+    // reference: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+    switch (status.getCode()) {
+      case 0:
+        code = 200;
```
garrettjonesgoogle: I know we have to have this mapping somewhere, probably in gax

pongad: We have a reverse map: https://github.com/googleapis/gax-java/blob/master/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonStatusCode.java#L60

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,198 @@
...
   }
 
   private static ResourceManagerException translate(IOException exception) {
     return new ResourceManagerException(exception);
   }
 
+  private static ResourceManagerException translate(Status status) {
+    int code;
+    // reference: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+    switch (status.getCode()) {
```
vam-google: Can this be just a map of `statusCode => code`? Having such a long and narrow switch statement hurts my eyes. Having a map at least makes it shorter:

The constant in the class:
```java
    private static final ImmutableMap<Integer, Integer> ERROR_CODES_MAP =
        ImmutableMap.builder()
            .put(0, 200)
            .put(1, 499)
            .put(2, 500)
            .put(3, 400)
            .put(4, 504)
            .put(5, 404)
            .put(6, 409)
            .put(7, 403)
            .put(16, 401)
            .put(8, 429)
            .put(9, 400)
            .put(10, 409)
            .put(11, 400)
            .put(12, 501)
            .put(13, 500)
            .put(14, 503)
            .put(15, 500)
            .build();
```

Then in this method:
```java
    code = ERROR_CODES_MAP.getOrDefault(status.getCode(), BaseHttpServiceException.UNKNOWN_CODE);
```

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,198 @@
...
+              CREATE_RETRY_SETTINGS,
+              OPERATION_HANDLER,
+              clock);
+      if (finishedOp.getError() != null) {
+        throw translate(finishedOp.getError());
+      }
+
+      // NOTE(pongad): Operation.getResponse() returns a Map<String, Object>.
+      // `(Project) finishedOp.getResponse()` doesn't work,
+      //   because JSON deserializer in execute() didn't know to create a Project object.
```
vam-google: The comment formatting is weird, is it intentional (unexpected indentations)?

pongad: It was intentional "paragraph heading". Changed to number instead to make this clear.

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
pongad: Good idea, done.

pongad:
@vam-google PTAL

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,168 @@
...
+  // https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/java/latest/com/google/api/services/cloudresourcemanager/CloudResourceManager.Projects.html#create(com.google.api.services.cloudresourcemanager.model.Project)
+  private static final RetrySettings CREATE_RETRY_SETTINGS =
+      RetrySettings.newBuilder()
+          // SLO permits 30s at 90th percentile, double it for total limit
+          .setTotalTimeout(Duration.ofMinutes(1))
+          .setInitialRetryDelay(Duration.ofSeconds(5))
+          .setRetryDelayMultiplier(1.5)
+          // Observed P95 latency is 11s. We probably shouldn't sleep longer than this.
+          .setMaxRetryDelay(Duration.ofSeconds(11))
+          .setJittered(false)
```
vam-google: Should it really be not jittered?

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,168 @@
...
-  private final Cloudresourcemanager resourceManager;
+  private static final JsonFactory JSON_FACTORY =
+      new JacksonFactory();
+
+  // See doc of create() for more details:
+  // https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/java/latest/com/google/api/services/cloudresourcemanager/CloudResourceManager.Projects.html#create(com.google.api.services.cloudresourcemanager.model.Project)
+  private static final RetrySettings CREATE_RETRY_SETTINGS =
+      RetrySettings.newBuilder()
+          // SLO permits 30s at 90th percentile, double it for total limit
+          .setTotalTimeout(Duration.ofMinutes(1))
```
vam-google: 90th percentile looks too low (basically would mean that every 10th call will fail with total timeout. Yes, this doubles 30 seconds, so in which percentile are 60 seconds (should be high, much higher than 99th percentile)?

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,168 @@
...
+          .setInitialRetryDelay(Duration.ofSeconds(5))
```
vam-google: This will practically mean that no call will complete faster than in 5seconds (looks too high). Why 5 secods?

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,168 @@
...
+
+      Operation finishedOp =
+          runWithRetries(
+              new Callable<Operation>() {
+                @Override
+                public Operation call() {
+                  try {
+                    return resourceManager.operations().get(operation.getName()).execute();
+                  } catch (IOException ex) {
+                    throw translate(ex);
```
vam-google: In case of error you this exception will be wrapped again on line 166, so wrapping/translating it here seems redundant.

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,168 @@
...
+        @Override
+        public TimedAttemptSettings createNextAttempt(
+            Throwable prevThrowable, Operation prevResponse, TimedAttemptSettings prevSettings) {
+          return null;
+        }
+
+        @Override
+        public boolean shouldRetry(Throwable prevThrowable, Operation prevOp)
+            throws CancellationException {
+          return prevThrowable != null || prevOp.getDone() == null || !prevOp.getDone();
```
vam-google: `prevThrowable != null` looks a bit suspicious, as it means "keep retrying no matter what". Usually there are some retriable and not retriable exceptions/error codes. Whould we really retry on any throwable here?

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,168 @@
...
+      }
+
+      // NOTE(pongad): Operation.getResponse() returns a Map<String, Object>.
+      // 1. `(Project) finishedOp.getResponse()` doesn't work,
+      // because JSON deserializer in execute() didn't know to create a Project object.
+      // 2. `new Project().putAll(finishedOp.getResponse())` doesn't work either.
+      // 64-bit integers are sent as strings in JSON,
+      // so execute(), not knowing the type, parses it as String, not Long.
+      String responseTxt = JSON_FACTORY.toString(finishedOp.getResponse());
+      return JSON_FACTORY.fromString(responseTxt, Project.class);
```
vam-google: I cannot verify this, so just hope that it works under all conditions...

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,58 +16,168 @@
...
+    clock = options.getClock();
   }
 
   private static ResourceManagerException translate(IOException exception) {
     return new ResourceManagerException(exception);
   }
 
+  private static ResourceManagerException translate(Status status) {
+    int code;
+    if (RPC_TO_HTTP_CODES.containsKey(status.getCode())) {
```
vam-google: This can be:

```java
Integer code = RPC_TO_HTTP_CODES.get(status.getCode());
if (code == null) {
    code = BaseHttpServiceException.UNKNOWN_CODE;
}
```
which is a typical technique, and is technically faster (one call to map instead of 2) and shorter.

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
pongad: Hmm, I suppose it retries both for errors and for operation not finishing. I'll turn jitter on.

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
pongad: The linked doc specifically recommend 5s. We're not following this completely though, as we send the first poll immediately; I think this is OK.

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
pongad: I don't think it will. This line deals with `get` throwing, but L166 deals with `get` itself succeeding but reporting that the operation wasn't successful right?

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
pongad: Done

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
pongad: Done.

pongad:
@vam-google PTAL

google-cloud-resourcemanager/src/main/java/com/google/cloud/resourcemanager/spi/v1beta1/HttpResourceManagerRpc.java
```diff
@@ -16,55 +16,171 @@
...
+  private static final ResultRetryAlgorithm<Operation> OPERATION_HANDLER =
+      new ResultRetryAlgorithm<Operation>() {
+        @Override
+        public TimedAttemptSettings createNextAttempt(
+            Throwable prevThrowable, Operation prevResponse, TimedAttemptSettings prevSettings) {
+          return null;
+        }
+
+        @Override
+        public boolean shouldRetry(Throwable prevThrowable, Operation prevOp) {
```
pongad: @vam-google Just to make sure, this gets called even if previous call doesn't throw exception right?

PR #2954, opened 2018-02-26
Allowing empty merges

schmidt-sebastian:
Port of https://github.com/googleapis/nodejs-firestore/pull/144

This PR allows the call set({}, merge:true) since it can be used to ensure that a document exist on the server.

This came up as a valid use case in an internal discussion (b/73495873 - based on customer feedback in b/73132705).

There is currently a conformance test that rejects this call. I excluded the test and will work with @jba on an update.


schmidt-sebastian:
Tests are passing.

PR #2958, opened 2018-02-26
[do not merge] add dlp v2

neozwu:
add dlp v2


jabubake:
@neozwu @pongad Ping.
I've made changes locally to get this to compile, it would be great if you could update this branch to the latest + review. I've been writing samples against this PR.

kurtisvg:
Putting together some of the samples for the library brought to light a lot of friction points in it's current state. 

One particular point is the `setParent()` function on several of the classes (which name doesn't seem clear - shouldn't it be `setParentId` like some of the other libraries?). It requires a string, but wants the string in the format "project/PROJECT_NAME". It seems the goal is to use `ProjectName.of(projectid).toString()`, but that is incredibly redundant. Ideally it should require a ProjectName and if I pass it a string, it should handle conversion to a ProjectName without all the extra steps. This seems to be a common theme - there are like 3 extra functions calls for what feels like every `.set` for some attribute.

It would also be great to get some functions for waiting until a job status is complete - currently we are using PubSub to accomplish this task in the samples, but given that there exist some endpoints for getting a job and checking it's status, it seems like something tht would be useful. 

Additionally, there is still a v1beta1 package. I don't know if those classes are there intentionally or not, but they don't seem to used anymore and are a little confusing to have around.

neozwu:
@garrettjonesgoogle Does current setParent() look like the only way code generator can produce code, or we can configure to generate code as @kurtisvg suggested.

Regarding adding 'waiting' to jobs, there was some prior discussions.  IIUC, in reality these jobs take really long time (hours to days), so API team believes the current design is sufficient (in previous version the job is an LRO, which has more wrapper functions. but it is replaced with DlpJob now)

/cc @ace-n @garrettjonesgoogle

garrettjonesgoogle:
If it is proto-generated classes you're talking about, we are unable to enhance those. It's possible to use a slightly smaller bit of code to get what you want: `ProjectName.format(projectid)`. Since it expects a string of the form `projects/{projectId}`, it's actually correct to not have `id` on the name.

neozwu:
Close this PR since DLP2 has been merged in a different PR.

PR #2959, opened 2018-02-27
batchingexperimental: add RequestAccumulator

pongad:
Instead of exposing one surface that knows how to batch end-to-end,
this PR explores ways to expose multiple surfaces
to be mixed and matched by individual clients.

RequestAccumulator accumulates requests into batches, but leaves
RPC calls, error handling, etc to the caller.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -236,7 +236,6 @@ public class Publisher {
       final PublishRequest publishRequest,
       final List<SettableApiFuture<String>> futures,
       Executor publishExecutor) {
-    // Is the create time right?
     final long createTimeMs = System.currentTimeMillis();
     publishExecutor.execute(
         new Runnable() {


pongad:
Submitting to a branch, but we should actually measure accumulator's performance before going too far.

@igorbernstein2 I think this ought to fit pubsub and logging's use case. Do you think it fits bigtable's?

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ *     // ...
+ *
+ *     accumulator.next();
+ *   }
+ * }
+ * }</pre>
+ */
+@InternalApi
+public class RequestAccumulator<E, R> {
```
igorbernstein2: Please document what E & R are, it's nonobvious

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
```diff
@@ -0,0 +1,137 @@
...
+   * with a {@code future}.
+   *
+   * <p>Given a call {@code add(e, f)}, {@code e} and {@code f} are in the same index in {@link
+   * #batch()} and {@link #futureBatch()}.
+   *
+   * @throws IllegalArgumentException if {@code bytes} is not positive.
+   * @throws IllegalStateException if {@link #hasBatch()} would return true immediately before the
+   *     call.
+   */
+  public void add(E e, long bytes, SettableApiFuture<R> future) {
```
igorbernstein2: Might be nicer to have this method create the future & return it

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ *     accumulator.next();
+ *   }
+ * }
+ * }</pre>
+ */
+@InternalApi
+public class RequestAccumulator<E, R> {
+  private final ArrayList<E> requests = new ArrayList<>();
+  private final ArrayList<SettableApiFuture<R>> futures = new ArrayList<>();
```
igorbernstein2: Might be nicer to have a value class encapsulate a request/future pair instead of managing parallel lists

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
pongad: Good idea. Done.

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
pongad: I thought about this as well. My current thinking: logging doesn't require a future for every message, so it seems wasteful to create it unconditionally. The code required to create a future is also quite simple, so I don't feel *that* bad about making PubSub and BigTable write an extra one-liner.

If this turns out to be a great inconvenience, it's also easy to change later.

Do you feel strongly about this?

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
garrettjonesgoogle: Could you spell out E & R instead of using single-letter parameters?

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
garrettjonesgoogle: I feel like the surface would be clearer if maybe we had a distinct method for the case where no future was involved.

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
pongad: Won't this move the complexity elsewhere though? If we return a `List<Pair>` then the user will have to split the pair themselves when they want to send RPCs.

Perhaps you meant we keep the current surface and create `List<E>` from `List<Pair>` in `batch()`. I'm not strongly against this idea, though it seems wasteful to create new lists and pair objects often. I'm probably prematurely optimizing. In this case, I think it's somewhat OK. The class is small enough that it isn't hard to comb through, and it's easy to change if need arises.

pongad:
@igorbernstein2 PTAL.

> Bigtable client will also need to deal with concurrent batches and flow control. Is that in scope for another shared component?

For concurrent batch, do you mean sending second batch before the first returns? Pubsub is already doing this. You can substitute the `// ...` in the class doc with an async call. You'll need to copy lists that you want to retain though.

For flow control, I think it's definitely in scope. Pubsub team requested us to not put it into pubsub client yet, and I'd like to see two use-cases before I start abstracting. My current opinion: You and I can write flow control ourselves for Logging and BigTable first. Then we can see what they have in common (could be "everything"), then we can refactor. It's more work, but I think it will help with design details. WDYT?

igorbernstein2:
I think that sounds fine. Only that's missing here is that `curBytes` needs to be exposed so that flow control can use the same numbers as the accumulator.

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
pongad: Done

google-cloud-core/src/main/java/com/google/cloud/batchingexperimental/RequestAccumulator.java
pongad: It'd still be equivalent to setting the future to null though right? Otherwise `batch()` and futureBatch()` won't be the same length anymore.

pongad:
> I'd like to see the impact to Logging too

@garrettjonesgoogle I'd like to work on this in a separate PR, since this one is already getting large. We're in a branch, so editing things should be easy.

> I think we can efficiently avoid this by storing the extra item in a separate variable.

@igorbernstein2 I think this isn't quite optimal. When working with protobuf, you have to do something like `request.addAllItems(accumulator.batch())`. `addAll` copies, so we essentially get the copy for "free" in the sense that we need to do that anyway. I'm sort of hoping that `clear()` doesn't downsize the backing array, so that we don't need to allocate temporary lists to move items.

> Only that's missing here is that curBytes needs to be exposed

I agree but I'd like to do this in a separate PR. It deserves some tests.

Unless either of you fundamentally disagree with the premise of this PR. I think we should merge this.
Things like naming and function signatures can be hashed out a little later. We're in a branch anyway. WDYT?

igorbernstein2:
SGTM

garrettjonesgoogle:
In general, I won't block any merges to the branch.

PR #2972, opened 2018-02-28
storage: delete blobs faster

### LINKED ISSUE #2962: Storage IT intermittently timing out (ITStorageTest: testListBlobRequesterPays)
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-java/1745
https://circleci.com/gh/GoogleCloudPlatform/google-cloud-java/1671

pongad:
Fixes #2962.

The stack trace from the linked issue suggests we are deleting blobs too
slowly, leading to timeouts. This PR tackles the problem in two ways.

1. Unless a test is actually testing blob deletion, the test is no
longer responsible for cleaning up after itself. Blobs left over are
batch-deleted at the end.

2. In case the batch-deletion fails, we set the bucket to auto-delete
blobs over 1 day old. Then the clean up job will delete buckets over 2 days old.
These buckets should already be empty and easy to delete.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/testing/RemoteStorageHelper.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/testing/RemoteStorageHelper.java
@@ -70,7 +70,7 @@ public class RemoteStorageHelper {
     return options;
   }
 
-  public static void cleanOldBuckets(final Storage storage, final long olderThan, long timeoutMs) {
+  public static void cleanBuckets(final Storage storage, final long olderThan, long timeoutMs) {
     Runnable task =
         new Runnable() {
           @Override
@@ -93,7 +93,7 @@ public class RemoteStorageHelper {
     try {
       thread.join(timeoutMs);
     } catch (InterruptedException e) {
-      log.info("cleanOldBuckets interrupted");
+      log.info("cleanBuckets interrupted");
     }
   }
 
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/testing/RemoteStorageHelperTest.java
```diff
@@ -126,9 +128,12 @@ public String getNextPageToken() {
   public void testForceDelete() throws InterruptedException, ExecutionException {
     Storage storageMock = EasyMock.createMock(Storage.class);
     EasyMock.expect(blob1.getBlobId()).andReturn(BLOB_ID1);
-    EasyMock.expect(storageMock.delete(BLOB_ID1)).andReturn(true);
     EasyMock.expect(blob2.getBlobId()).andReturn(BLOB_ID2);
-    EasyMock.expect(storageMock.delete(BLOB_ID2)).andReturn(true);
+
```
pongad: I just realized, can we just delete these tests? I think we're just writing code twice.

google-cloud-storage/src/main/java/com/google/cloud/storage/testing/RemoteStorageHelper.java
```diff
@@ -67,6 +70,33 @@ public StorageOptions getOptions() {
     return options;
   }
 
+  public static void cleanOldBuckets(final Storage storage, final long olderThan, long timeoutMs) {
```
garrettjonesgoogle: "old" is relative here, and is effectively determined by the `olderThan` parameter. I suggest just `cleanBuckets`.

PR #2976, opened 2018-02-28
Move bom dependencies to outer pom

yihanzhen:
- Move guava, protobuf and test dependencies from bom/pom.xml to pom.xml
- Also fix: bigquery imported truth twice

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -348,7 +348,7 @@
       <dependency>
         <groupId>com.google.apis</groupId>
         <artifactId>google-api-services-cloudresourcemanager</artifactId>
-        <version>v1beta1-rev470-1.23.0</version>
+        <version>v1-rev470-1.23.0</version>
       </dependency>
       <dependency>
         <groupId>com.google.api.grpc</groupId>


hzyi-google:
PTAL @garrettjonesgoogle @pongad

pongad:
I think we should do point 3 of https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2781#issuecomment-369372684 as well, but that can be in a separate PR.

pongad:
@garrettjonesgoogle Do you want to take a look?

pom.xml
```diff
@@ -152,8 +165,531 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-
-      <!-- pure test deps -->
+      <dependency>
+        <groupId>com.google.api.grpc</groupId>
+        <artifactId>proto-google-cloud-bigtable-v2</artifactId>
+        <version>${generated-proto-beta.version}</version>
```
garrettjonesgoogle: Actually I think it might make sense to have the proto/grpc generated classes in the BOM - let's follow up with the relevant parties to confirm.

PR #2980, opened 2018-03-01
Bigtable: 15. Implement CheckAndMutateRow

igorbernstein2:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -230,6 +242,7 @@ public static EnhancedBigtableStub create(EnhancedBigtableStubSettings settings)
     return mutateRowsCallable;
   }
 
+  /** Returns the callable chain created in {@link #createCheckAndMutateRowCallable()}. */
```
kevinsi4508: nit:
This comment is probably less useful. The reason is that createCheckAndMutateRowCallable() is private and it is called in the constructor to get checkAndMutateRowCallable initialized. By pointing user directly to createCheckAndMutateRowCallable() creates confusion. I would simple just drop that or add "created in the constructor by invoking createCheckAndMutateRowCallable()".

igorbernstein2: updated the doc

igorbernstein2:
Thanks for reviewing! Please merge when ready

PR #2983, opened 2018-03-02
pubsub: don't publish empty batches

pongad:


garrettjonesgoogle:
Is this too hard to test?

pongad:
It's quite weird to test. We can test that a request is sent by waiting until the server receives a request.

But then it's hard to determine that a request isn't going to get sent. We can wait for some duration and check but then there's no way to make sure that a request won't arrive a second later. :/

garrettjonesgoogle:
Ok, just making sure we thought about it. LGTM

PR #2987, opened 2018-03-02
Bigtable: 14. Implement Bulk Mutations

igorbernstein2:
Please note that this is only the initial implementation. There will be follow up PRs for:
1. once a new version of gax is cut, I will move retries behind th BatchingCallable
2. Once there is flushing implementation, it will be swapped in


igorbernstein2:
rebased

PR #2991, opened 2018-03-05
Fix bugs in javadoc and build

andreamlin:
- Remove javadoc link to non-existent class in bigtable
- Add gax.version 1.19.0 back into main pom.xml
- Set api-common dependency to 1.4.0

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/DummyBatchingDescriptor.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/DummyBatchingDescriptor.java
@@ -27,7 +27,10 @@ import java.util.Collection;
  * com.google.cloud.bigtable.data.v2.stub.EnhancedBigtableStub#mutateRowsCallable()}. The settings
  * are exposed to the user using the {@link com.google.cloud.bigtable.data.v2.models.RowMutation}
  * wrapper, but the actual descriptor works on the underlying {@link
- * com.google.bigtable.v2.MutateRowsRequest}s.
+ * com.google.bigtable.v2.MutateRowsRequest}s. This class is used as a placeholder for the settings
+ * and is replaced with the actual implementation of
+ * com.google.cloud.bigtable.data.v2.stub.bulkmutaterows.BulkMutateRowsBatchingDescriptor when
+ * constructing the callable chain.
  *
  * <p>This class is considered an internal implementation detail and not meant to be used by
  * applications.


igorbernstein2:
I would prefer to merge #2987 which adds the class rather then remove the doc here.

pom.xml
```diff
@@ -138,6 +138,7 @@
     <site.installationModule>google-cloud</site.installationModule>
     <bom.version>0.37.1-alpha-SNAPSHOT</bom.version><!-- {x-version-update:google-cloud-pom:current} -->
     <api-client.version>1.23.0</api-client.version>
+    <gax.version>1.19.0</gax.version>
```
garrettjonesgoogle: I see that this was broken since I created the BOM module. Ideally I'd like to avoid having the gax version specified in more than one place. Let's do this for now, knowing that we need to implement https://github.com/googleapis/gax-java/issues/489 at some point.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/DummyBatchingDescriptor.java
```diff
@@ -27,10 +27,7 @@
  * com.google.cloud.bigtable.data.v2.stub.EnhancedBigtableStub#mutateRowsCallable()}. The settings
  * are exposed to the user using the {@link com.google.cloud.bigtable.data.v2.models.RowMutation}
  * wrapper, but the actual descriptor works on the underlying {@link
- * com.google.bigtable.v2.MutateRowsRequest}s. This class is used as a placeholder for the settings
- * and is replaced with the actual implementation of {@link
- * com.google.cloud.bigtable.data.v2.stub.bulkmutaterows.BulkMutateRowsBatchingDescriptor} when
```
garrettjonesgoogle: We can just remove the `@link` here - I think the linked class might exist in a later PR by @igorbernstein2 .

andreamlin: Oh good to know.
Just removed the `@link`.

andreamlin:
PTAL

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/DummyBatchingDescriptor.java
garrettjonesgoogle: I don't see the class reference still...

igorbernstein2: Originally it was a dangling link in the bulk mutations surface PR, meant to give context until the impl PR (#2987) was merged. During the reviews, its package was renamed, but the link was never updated.
It is fixed in PR #2987
https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2987/files#diff-e26d48d3afb81b5ddff92add0217ae58

andreamlin: Ok, this time I actually pushed the commit to remove the `@link`.

PR #2994, opened 2018-03-05
Update grpc/proto packages

andreamlin:


andreamlin:
Ok. Will rerun circleci when the proto/grpc packages are up on maven.

PR #3006, opened 2018-03-07
[WIP] make smoke test a test and an app

neozwu:
Redesign autogenerated smoke test so that it has a "test entry point" and an "app entry point".  Therefore, each smoke test can be run separately as a sample app (say, for debugging purpose). Or, all smoke tests can be run using mvn. ("mvn verify -P smoke-test"). For clients that require project-id, project-id can be detected through system property or environment variable. If app entry point is used, one can also pass project id through command line.

### DIFF
--- a/google-cloud-core/src/main/java/com/google/cloud/testing/SmokeTest.java
+++ b/google-cloud-core/src/main/java/com/google/cloud/testing/SmokeTest.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2016 Google LLC
+ *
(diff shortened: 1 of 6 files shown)


google-cloud-core/src/main/java/com/google/cloud/testing/SmokeTest.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2016 Google LLC
```
garrettjonesgoogle: 2018

PR #3008, opened 2018-03-07
pubsub: make Publisher use GAPIC client

pongad:


google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/PublisherImplTest.java
```diff
@@ -384,41 +332,6 @@ public void testPublishFailureRetries_maxRetriesSetUnlimited() throws Exception
     publisher.shutdown();
   }
 
-  public void testPublishFailureRetries_exceededsRetryDuration() throws Exception {
```
pongad: This test has been disabled for a long time. It tests retries, which is already tested in GAX so we can remove this now.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminSettings.java
```diff
@@ -91,7 +90,7 @@
   }
 
   /** Returns the object with the settings used for calls to publish. */
-  /* package-private */ BatchingCallSettings<PublishRequest, PublishResponse> publishSettings() {
+  /* package-private */ UnaryCallSettings<PublishRequest, PublishResponse> publishSettings() {
```
pongad: Temporary hand-edits. If the rest of the change looks good, we can merge as-is, then update GAPIC config and regen.

garrettjonesgoogle: Could you also create the CL for updating the GAPIC config? Will stripping out batching break any other languages?

garrettjonesgoogle:
@hzyi-google could you do the first pass?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/stub/PublisherStubSettings.java
```diff
@@ -586,16 +583,21 @@ private static Builder initDefaults(Builder builder) {
...
-                  .setDelayThreshold(Duration.ofMillis(10))
-                  .setFlowControlSettings(
-                      FlowControlSettings.newBuilder()
-                          .setLimitExceededBehavior(LimitExceededBehavior.Ignore)
-                          .build())
-                  .build());
+          .setRetryableCodes(RETRYABLE_CODE_DEFINITIONS.get("non_idempotent"))
+          .setRetrySettings(RETRY_PARAM_DEFINITIONS.get("default"));
+
+      // builder
```
hzyi-google: Remove the comments?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/stub/PublisherStubSettings.java
```diff
@@ -528,9 +524,10 @@ protected Builder(ClientContext clientContext) {
 
       updateTopicSettings = UnaryCallSettings.newUnaryCallSettingsBuilder();
 
-      publishSettings =
-          BatchingCallSettings.newBuilder(PUBLISH_BATCHING_DESC)
-              .setBatchingSettings(BatchingSettings.newBuilder().build());
+      publishSettings = UnaryCallSettings.newUnaryCallSettingsBuilder();
+      // publishSettings =
```
hzyi-google: Remove the comments?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -94,21 +83,18 @@
...
 
   private final Lock messagesBatchLock;
   private List<OutstandingPublish> messagesBatch;
   private int batchedBytes;
 
   private final AtomicBoolean activeAlarm;
 
-  private final Channel channel;
-  @Nullable private final CallCredentials callCredentials;
+  private final TopicAdminClient topicClient;
```
garrettjonesgoogle: Could we directly use `PublisherStub` instead of `TopicAdminClient`? It is conceptually cleaner, and then we can actually completely remove the `publish` call from `TopicAdminClient` instead of having it package-private.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -127,40 +113,43 @@ private Publisher(Builder builder) throws IOException {
...
     if (builder.executorProvider.shouldAutoClose()) {
-      closeables.add(new ExecutorAsBackgroundResource(executor));
-    }
-    TransportChannelProvider channelProvider = builder.channelProvider;
-    if (channelProvider.needsExecutor()) {
-      channelProvider = channelProvider.withExecutor(executor);
+      closeables =
+          Collections.<AutoCloseable>singletonList(new ExecutorAsBackgroundResource(executor));
+    } else {
+      closeables = Collections.emptyList();
```
garrettjonesgoogle: Why not just pass `executorProvider` to the client and let the client close it down? (It already has to close down the channel)

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -361,37 +338,16 @@ public void onSuccess(PublishResponse result) {
...
-                for (OutstandingPublish outstandingPublish :
-                    outstandingBatch.outstandingPublishes) {
-                  outstandingPublish.publishResult.setException(gaxException);
-                }
-              } finally {
-                messagesWaiter.incrementPendingMessages(-outstandingBatch.size());
+            try {
+              ApiException gaxException =
+                  ApiExceptionFactory.createException(
+                      t, GrpcStatusCode.of(Status.fromThrowable(t).getCode()), false);
```
garrettjonesgoogle: Aren't there non-ApiException exceptions that can be thrown?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
pongad: Done.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
pongad: I'm not quite sure what you're getting at. If the `Throwable` isn't `ApiException`, won't `createException` wrap it?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
pongad: `Publisher` itself needs to use the executor to schedule publishing. AFAIK, the stub can't expose the executor it's currently using right?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
garrettjonesgoogle: The more problematic spot is `Status.fromThrowable(t)` - I doubt that will be well-behaved.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
garrettjonesgoogle: Ahh good point.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
garrettjonesgoogle: Actually, the `Throwable` itself should already be an `ApiException`, right?

pongad:
@garrettjonesgoogle PTAL

pongad:
@garrettjonesgoogle Committed the test now

pongad:
Thank you @garrettjonesgoogle . Sorry for the churn.

PR #3009, opened 2018-03-08
Bigtable: 20. fix dependencies

igorbernstein2:


google-cloud-bigtable/pom.xml
```diff
@@ -33,7 +33,6 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-cloud-bigtable-v2</artifactId>
-      <scope>test</scope>
     </dependency>
```
garrettjonesgoogle: Why do you need the grpc stubs for non-test? This is actually correct.

igorbernstein2:
You are right I don’t need these. Sorry for the noise

PR #3011, opened 2018-03-08
Fixing warn log message

pgbhagat:
Fixing warn log message

### DIFF
--- a/google-cloud-translate/src/main/java/com/google/cloud/translate/TranslateOptions.java
+++ b/google-cloud-translate/src/main/java/com/google/cloud/translate/TranslateOptions.java
@@ -175,7 +175,7 @@ public class TranslateOptions extends ServiceOptions<Translate, TranslateOptions
       if (getDefaultApiKey() != null) {
         logger.log(
             Level.WARNING,
-                "Ignoring API key set in environment variable {}: using Application Default Credentials instead.",
+                "Ignoring API key set in environment variable {0}: using Application Default Credentials instead.",
                 API_KEY_ENV_NAME);
       }
     } else {


pgbhagat:
I signed it!

google-cloud-translate/src/main/java/com/google/cloud/translate/TranslateOptions.java
```diff
@@ -169,7 +169,7 @@ private TranslateOptions(Builder builder) {
       logger.log(
           Level.WARNING,
           String.format(
-              "Ignoring Application Default Credentials: using explicit setting for API key instead.",
+              "Ignoring Application Default Credentials %s: using explicit setting for API key instead.",
```
pongad: Hmm. Can we instead do
```java
logger.log(Level.WARNING, "Ignoring Application Default Credentials {}:...", ServiceOptions.CREDENTIAL_ENV_NAME);
```
? That way we don't need to format the string if the log is turned off.

pgbhagat: Done.

google-cloud-translate/src/main/java/com/google/cloud/translate/TranslateOptions.java
```diff
@@ -155,30 +155,28 @@ private TranslateOptions(Builder builder) {
...
+                "Ignoring API key set in environment variable {}: using explicit setting for credentials instead.",
+                API_KEY_ENV_NAME);
       }
     } else if (builder.apiKey != null) {
       credentials = null;
       apiKey = builder.apiKey;
       logger.log(
           Level.WARNING,
-          String.format(
-              "Ignoring Application Default Credentials: using explicit setting for API key instead.",
```
pgbhagat: String.format was missing %s earlier for arg  ServiceOptions.CREDENTIAL_ENV_NAME.

pongad:
@pgbhagat Gah, I'm sorry. The format string should have been `{0}`, not `{}`. I was thinking in Python. LGTM otherwise. I apologize for the churn.

pgbhagat:
No problem, i checked other files and it is used like {0}. Corrected in the next commit.

pgbhagat:
Is the AppVeyor build is failing because of
'Build execution time has reached the maximum allowed time for your plan (60 minutes).' ?

PR #3018, opened 2018-03-09
pubsub: make Publisher/Subscriber accept plain strings

pongad:


google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -204,19 +202,32 @@ public void close() throws IOException {
...
+   * @param receiver an implementation of {@link MessageReceiver} used to process the received
+   *     messages
+   */
+  public static Builder newBuilder(String subscription, MessageReceiver receiver) {
     return new Builder(subscription, receiver);
   }
 
   /** Subscription which the subscriber is subscribed to. */
   public ProjectSubscriptionName getSubscriptionName() {
+    return ProjectSubscriptionName.parse(subscriptionName);
```
pongad: Should we also make this a oneof? It isn't right now.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -167,7 +167,12 @@ private Publisher(Builder builder) throws IOException {
   }
 
   /** Topic which the publisher publishes to. */
-  public ProjectTopicName getTopicName() {
+  public TopicName getTopicName() {
+    return TopicNames.parse(topicName);
+  }
+
+/** Topic which the publisher publishes to. */
```
garrettjonesgoogle: Indentation

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
garrettjonesgoogle: I struggled with that myself - I decided not to when I did the refactoring, but I think it would provide better surface consistency.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
pongad: Formatted.

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
pongad: Created https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3021 to track.

PR #3022, opened 2018-03-12
wip: spanner: implement partitionRead

pongad:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -169,6 +177,8 @@ public PartitionResponse partitionQuery(
   @Override
   public PartitionResponse partitionRead(
       PartitionReadRequest request, @Nullable Map<Option, ?> options) throws SpannerException {
-    throw new UnsupportedOperationException("Not implemented yet.");
+    // TODO(pongad): Figure out metadata
```
pongad: @garrettjonesgoogle Does GAPIC client provide a good way to put "whatever header I want"?
Currently Spanner's logic is
1. if `request.getSession()` satisfies a regular expression, set header to that
2. otherwise, set header to project ID

I think this logic is too specific for GAX to implement.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -169,6 +177,8 @@ public PartitionResponse partitionQuery(
...
+    // TODO(pongad): Figure out channel affinity
```
pongad: I'll investigate this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
garrettjonesgoogle: It allows "whatever header" in a static way (can't change once you create the client). If Spanner needs dynamic headers, then we need to add support to GAX.

vkedia: Yes spanner does need dynamic headers. I feel the specific header spanner is setting might be needed by other clients sooner or later.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
vkedia: This will basically need enhancing the channel pool in gax to take in an affinity key.

pongad: This is being done in a linked PR to GAX. Once merged and released, I'll resolve this.

PR #3023, opened 2018-03-12
storage: speed up IT

### LINKED ISSUE #3019: Storage integration tests failing
Last 8 commits have failed. https://circleci.com/gh/GoogleCloudPlatform/google-cloud-java/2684

pongad:
We create fewer blobs since the basic case is already
tested by other ITs.

Fixes #3019, hopefully.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
@@ -446,7 +446,7 @@ public class ITStorageTest {
       // Otherwise, expect one blob, since we only put in one above.
       if (!blobs.isEmpty()) {
         assertThat(blobs).hasSize(1);
-        return;
+        break;
       }
       Thread.sleep(500);
     }


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -410,45 +410,42 @@ public void testListBlobsEmptySelectedFields() throws InterruptedException {
...
           Storage.BlobListOption.fields(),
           Storage.BlobListOption.userProject("fakeBillingProjectId"));
       fail("Expected bad user project error.");
     } catch (StorageException e) {
       assertTrue(e.getMessage().contains("User project specified in the request is invalid"));
     }
-    while (Iterators.size(page.iterateAll().iterator()) != 2) {
+
+    String projectId = remoteStorageHelper.getOptions().getProjectId();
+    for (; ; ) {
```
garrettjonesgoogle: Isn't it more common to do `while(true)` in Java?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -410,45 +410,42 @@ public void testListBlobsEmptySelectedFields() throws InterruptedException {
...
+      Page<Blob> page =
+          storage.list(
+              BUCKET,
+              Storage.BlobListOption.prefix("test-list-blobs-empty-selected-fields-blob"),
+              Storage.BlobListOption.fields(),
+              Storage.BlobListOption.userProject(projectId));
+      int size = Iterators.size(page.iterateAll().iterator());
+      if (size == 1) {
```
garrettjonesgoogle: This check feels weird - what if it keeps returning `2` or something like that?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
pongad: I don't know. I like `for` because it has fewer characters, but `while` is fine with me.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
pongad: Done

pongad:
@garrettjonesgoogle PTAL. I also slightly increased the time limit. Hopefully it flakes less.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
garrettjonesgoogle: Characters, schmaracters. It increases the token count, which I feel is worse :-P

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -434,16 +434,19 @@ public void testListBlobRequesterPays() throws InterruptedException {
...
               Storage.BlobListOption.userProject(projectId));
-      int size = Iterators.size(page.iterateAll().iterator());
-      if (size == 1) {
-        break;
+      List<Blob> blobs = Lists.newArrayList(page.iterateAll());
+      // If the list is empty, maybe the blob isn't visible yet; wait and try again.
+      // Otherwise, expect one blob, since we only put in one above.
+      if (!blobs.isEmpty()) {
+        assertThat(blobs).hasSize(1);
+        return;
```
garrettjonesgoogle: Maybe `break` instead, in case we want to add something after the loop?

pongad: Done

PR #3029, opened 2018-03-13
batchingexperimental: report size of batch

pongad:
and make BatchingWriter use it.

### DIFF
--- a/google-cloud-logging/src/main/java/com/google/cloud/logging/BatchingWriter.java
+++ b/google-cloud-logging/src/main/java/com/google/cloud/logging/BatchingWriter.java
@@ -91,19 +91,13 @@ class BatchingWriter {
         final WriteLogEntriesRequest request =
             requestPrototype.toBuilder().addAllEntries(accumulator.batch()).build();
         final long size = accumulator.bytes();
-        // Whoever calls send serializes the proto; so we do it off-thread.
+        send(request, size);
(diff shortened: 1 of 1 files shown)


igorbernstein2:
Did you mean to include the logging changes in this PR?

google-cloud-logging/src/main/java/com/google/cloud/logging/BatchingWriter.java
```diff
@@ -85,43 +81,50 @@ synchronized void stopJob() {
...
         // on a many-core machine.
+        //
+        // BUG(pongad): send() puts future in the map too late. To make flush work properly,
+        // we must guarantee adding future to map before this method returns.
         executor.execute(
             new Runnable() {
               @Override
               public void run() {
-                send(request);
+                send(request, size);
```
igorbernstein2: I think it might be better to remove the off thread execution (at least for the time being). It should solve the flush race condition. I'm also not convinced that CPU utilization of serializing protos will be a bottleneck here and worth the complexity

pongad: I buy the argument that we should `send` in-thread right now to simplify things. That said, when I desktop-tested this code a while back, one thread writing trivial logs pushed my machine to 400% CPU so I think the bottleneck could be real.

I'll add a TODO for off-thread serialization instead.

PR #3033, opened 2018-03-13
Fixing order test to verify that we order by components

schmidt-sebastian:
This test now verifies that resource names like path/a, path/b and path-/a are ordered correctly.

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java
@@ -89,7 +89,7 @@ public class OrderTest {
     groups[38] =
         new Value[] {referenceValue("projects/p1/databases/dkkkkklkjnjkkk1/documents/c2/doc1")};
     groups[39] = new Value[] {referenceValue("projects/p2/databases/d2/documents/c1/doc1")};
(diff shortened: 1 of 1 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java
```diff
@@ -89,7 +89,7 @@ public void verifyOrder() {
     groups[38] =
         new Value[] {referenceValue("projects/p1/databases/dkkkkklkjnjkkk1/documents/c2/doc1")};
     groups[39] = new Value[] {referenceValue("projects/p2/databases/d2/documents/c1/doc1")};
-    groups[40] = new Value[] {referenceValue("projects/p2/databases/d2/documents/cl1-/doc1")};
+    groups[40] = new Value[] {referenceValue("projects/p2/databases/d2/documents/cl-/doc1")};
```
pongad: Commit message says "path/a, path/b and path-/a are ordered correctly". Is "path" here "c1"? If so, shouldn't this be "c1-" (see one dash) instead of "cl-" (see el dash)?

schmidt-sebastian: Ugh, that is correct. I also need to update this in the NodeJS repo.

PR #3039, opened 2018-03-14
spanner: implement LRO methods using GAPIC stub

pongad:


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -37,17 +64,85 @@
...
+    this.metadataProvider =
+        SpannerMetadataProvider.create(
+            mergedHeaderProvider.getHeaders(),
+            internalHeaderProviderBuilder.getResourceHeaderKey());
+
+    // TODO(pongad): make pool work
+
+    // TODO(pongad): make interceptors work
+
+    // Channel channel =
```
pongad: IIUC, GAX doesn't support interceptors yet, `FixedTransportChannelProvider` wants `ManagedChannel`, and `intercept` creates plain `Channel`.

My plan for the short term: Use `intercept` to create `Channel` then transform it into `ManagedChannel` by making shutdown-related operations no-op.
This way we can make progress towards making surface changes.
@garrettjonesgoogle @vkedia WDYT? If this sounds OK to you, I'll open an issue in GAX asking for interceptor support.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -37,17 +64,85 @@
...
 import java.util.Map;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import javax.annotation.Nullable;
 
 /** Implementation of Cloud Spanner remote calls using Gapic libraries. */
 public class GapicSpannerRpc implements SpannerRpc {
+  private static final PathTemplate PROJECT_NAME_TEMPLATE =
+      PathTemplate.create("projects/{project}");
```
garrettjonesgoogle: Spanner doesn't have a resource name for this?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
garrettjonesgoogle: Why do you need an interceptor?

pongad: There are 3 interceptors at play.

1. `LoggingInterceptor` that logs messages when RPCs are called. Not sure how important this is.

2. Watchdog. We can use GAX for this one.

3. `SpannerErrorInterceptor` looks at trailer to augment error description. I think GAPIC might get this feature soon for unary, but I think this interceptor also works with streaming.

garrettjonesgoogle: 1. I know we have an independent request for supporting the logging of requests - maybe worth supporting that way instead. 
3. I think we should cover streaming calls for error description augmentation at a shared level like we will for unary calls.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
pongad: Not yet. But good point, I sent you an internal PR for this.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
pongad: Documented these items in the work doc. Are you against merging this for now? I'll investigate on how to make these work.

garrettjonesgoogle: Update the TODO above also (which currently references interceptors).

PR #3043, opened 2018-03-14
Implement Paginated methods and get/drop/delete methods using GAPIC stub

yihanzhen:
Implement the following methods:
 - listInstances, listInstanceConfigs, listDatabases
 - getInstance, getInstanceConfigs, getDatabase, getDatabaseDdl
 - dropDatabase, deleteInstance

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -48,10 +48,22 @@ import com.google.longrunning.Operation;
 import com.google.protobuf.FieldMask;
 import com.google.spanner.admin.database.v1.CreateDatabaseRequest;
 import com.google.spanner.admin.database.v1.Database;
+import com.google.spanner.admin.database.v1.DropDatabaseRequest;
+import com.google.spanner.admin.database.v1.GetDatabaseDdlRequest;
+import com.google.spanner.admin.database.v1.GetDatabaseRequest;
+import com.google.spanner.admin.database.v1.ListDatabasesRequest;
+import com.google.spanner.admin.database.v1.ListDatabasesResponse;
 import com.google.spanner.admin.database.v1.UpdateDatabaseDdlRequest;
 import com.google.spanner.admin.instance.v1.CreateInstanceRequest;
+import com.google.spanner.admin.instance.v1.DeleteInstanceRequest;
+import com.google.spanner.admin.instance.v1.GetInstanceConfigRequest;
+import com.google.spanner.admin.instance.v1.GetInstanceRequest;
 import com.google.spanner.admin.instance.v1.Instance;
 import com.google.spanner.admin.instance.v1.InstanceConfig;
+import com.google.spanner.admin.instance.v1.ListInstanceConfigsRequest;
+import com.google.spanner.admin.instance.v1.ListInstanceConfigsResponse;
+import com.google.spanner.admin.instance.v1.ListInstancesRequest;
+import com.google.spanner.admin.instance.v1.ListInstancesResponse;
 import com.google.spanner.admin.instance.v1.UpdateInstanceRequest;
 import com.google.spanner.v1.BeginTransactionRequest;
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -44,27 +62,63 @@
 
 /** Implementation of Cloud Spanner remote calls using Gapic libraries. */
 public class GapicSpannerRpc implements SpannerRpc {
+
```
hzyi-google: The fields and constructor are only used to make it compile. #3039 implements this part; after that is merged I will resolve the conflicts here.

pongad: SGTM. The PR is already merged.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -44,27 +62,63 @@
...
+    ListInstanceConfigsRequest.Builder requestBuilder = 
+        ListInstanceConfigsRequest.newBuilder().setParent(projectName).setPageSize(pageSize);
+    if (pageToken != null) {
+      requestBuilder.setPageToken(pageToken);
+    }
+    ListInstanceConfigsRequest request = requestBuilder.build();
+
+    // TODO: put projectName in metadata
+    ListInstanceConfigsResponse response = 
+        instanceAdminStub.listInstanceConfigsCallable().call(request);
```
pongad: Instead of `call(request)`, I think we should use `get(instanceAdminStub.listInstanceConfigsCallable().futureCall(request)`.

`get` was implemented in #3039 and translate some exceptions into `SpannerException`.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -143,18 +156,45 @@ public GapicSpannerRpc(SpannerOptions options) throws IOException {
...
 
   @Override
   public Paginated<Instance> listInstances(
       int pageSize, @Nullable String pageToken, @Nullable String filter) throws SpannerException {
-    throw new UnsupportedOperationException("Not implemented yet.");
+    ListInstancesRequest.Builder requestBuilder =
+        ListInstancesRequest.newBuilder().setParent(projectName).setPageSize(pageSize);
+    if (pageToken != null) {
+      requestBuilder.setPageToken(pageToken);
+    }
```
garrettjonesgoogle: Why do we check for null for some Strings but not for others (e.g. `instanceName` in the `getInstance` function below)? If the proto-generated class is tolerate to nulls, then we shouldn't check anywhere.

hzyi-google: I did it the same way `GrpcSpannerRpc` implemented it but I am not super confident in it. Not sure if I understand correctly, both `ListInstancesRequest.Builder` and `GetInstanceRequest.Builder` do not allow null values (would throw a `NullPointerException`), but `pageToken` and `filter` are actually allowed to be null by API doc (the doc is for REST though): https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances/list

garrettjonesgoogle: Offline resolution: the pageToken/filter parameters are optional, so we should allow nulls.

hzyi-google:
Thank you all!

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
pongad: I could be wrong, but I don't think protobuf lets you set anything to null. IIRC, setting to null will throw `NullPointerException`.

PR #3063, opened 2018-03-20
Bigtable: copy InstanceName to data.models.

igorbernstein2:
This is a bit ugly: the autogenerate bigtable data api doesn't reference instances, only tables. So the resource name for instances is never generated. Up until this point I've been borrowing resource name from admin, which is really ugly. This PR simply copies the admin InstanceName to data's models package. At some point in the future when GAPIC implements common resource names, this can be cleaned up


igorbernstein2:
Please merge when ready

PR #3085, opened 2018-03-26
add a smoke-test mvn profile

neozwu:
This change adds a mvn profile to run smoke tests (by invoking "mvn -P smoke-test verify") once a separate change (https://github.com/googleapis/toolkit/pull/1898) is used to generate code.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -887,9 +887,6 @@
               <includes>
                 <include>**/*SmokeTest.java</include>
               </includes>
-              <excludes>
-                <exclude>**/*IT.java</exclude>
-              </excludes>
             </configuration>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -866,5 +866,42 @@
...
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <includes>
+                <include>**/*SmokeTest.java</include>
+              </includes>
+              <excludes>
+                <exclude>**/*IT.java</exclude>
+              </excludes>
```
garrettjonesgoogle: You don't need this exclusion given the inclusion pattern, right?

PR #3093, opened 2018-03-27
Expose the publishAllOutstanding() method as public

j256:
So that external systems that need to ensure that all messages have been published can clear a batch.   We need this because out system checkpoints every so often and we'd like to ensure that there aren't messages stuck in a batch buffer.

### DIFF
--- a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -290,7 +290,11 @@ public class Publisher {
     }
   }
 
-  /** publish any outstanding batches if non-empty */
+  /**
+   * Publish any outstanding batches if non-empty.   This method sends buffered messages, but does
+   * not wait for the send operations to complete.  To wait for messages to send, call {@code get}
+   * on the futures returned from {@code publish}.
+   */
   public void publishAllOutstanding() {
     messagesBatchLock.lock();
     OutstandingBatch batchToSend;


pongad:
This looks good to me. @mdietz94 do you have an opinion?

mdietz94:
Why can't users just check whether messages have been published already via the callbacks? That's meant to be the method by which users handle flow control / checkpointing.

pongad:
I might have misunderstood @j256 . My understanding is that we want "we're checkpointing, so let's just publish whatever we have right now".

@j256 Could you explain your use case a little more. Sorry I should have done this earlier.

j256:
So my use case is as follows.  We are streaming some large amount of data through the publisher so we want to have generous batch sizes for throughput reasons.  But every so often our system emits checkpoints so that the system synchronizes to a certain moment in time for recovery purposes.  This allows our various processing blocks to persist state information to disk for roll back reasons.

I could use the callback to know that messages have yet to be published but if they aren't then it's a failure in our processing and I'd need to rollback to the previous checkpoint because the publish was not acknowledged.

How do I know the difference between a message lost because of a protocol stack error versus to a message waiting for the batch time or size triggers?  I could always put small batch timeouts or sync them with our checkpoint times but exposing the ability to flush the batch (like any `stream.flush()` call) made more sense to me.

mdietz94:
What is your batch timeout? For even large batching we usually use 50ms, which does not seem to be too high to wait on, especially since the actual time to make the RPC to the server is >100ms at 99%ile.

j256:
Right now we've been using record-count or request-byte thresholds and setting the batch timeout to be higher to ensure throughput.

Unfortunately although I'm sure your bench times are correct, we are trying to build a system which survives our clients screwed up networks around the world.  It's the 1% we worry about.

Frankly I'm surprised about the pushback from effectively a flush call?  Isn't that the point of `bufferedOutputStream.flush()`?  You have a buffer with limits but every so often you need to ensure that the bytes are written to the wire?

google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -290,7 +290,8 @@ public void run() {
     }
   }
 
-  private void publishAllOutstanding() {
+  /** publish any outstanding batches if non-empty */
```
pongad: I think we should add something like

```
This method sends buffered messages, but does not wait for the send operations to complete.
To wait for messages to send, call {@code get}` on the futures returned from {@code publish}.
```

j256: Good idea.  Want me to do that @pongad?

pongad: If you could that would be great!

j256:
Done.

PR #3100, opened 2018-03-28
GAPIC Compute java library

andreamlin:
First gapic for Compute client.

Adds gax:testlib to google-cloud-bom/pom.xml.

### DIFF
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -155,7 +155,7 @@
       <dependency>
         <groupId>com.google.api</groupId>
         <artifactId>gax</artifactId>
-        <version>1.25.0</version>
+        <version>1.27.0</version>
         <classifier>testlib</classifier>
       </dependency>
       <dependency>
--- a/google-cloud-clients/google-cloud-compute/pom.xml
+++ b/google-cloud-clients/google-cloud-compute/pom.xml
@@ -64,5 +64,11 @@
       <classifier>testlib</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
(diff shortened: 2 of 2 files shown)


andreamlin:
Blocked on https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3102, which is waiting for newest gax-java packages to be released in maven.

andreamlin:
This was generated from toolkit/master. The google-java-format util was run on the output. The output was copied into google-cloud-java/google-cloud-compute.

andreamlin:
@garrettjonesgoogle Would you know how to fix the build compile errors?
For the Appveyor failure, there's apparently a missing file HttpsHealthCheck.java, even though it does exist and it's in the same package as the file where the error occurs.
For the CircleCI Java 7 failure, there's an out of memory error during the compile process that doesn't occur in the Java 8 build. I can look into the circleci config to see if there's a problem there.

andreamlin:
Upped the Java heap size for CircleCI openjdk_7 to 1024m from 256m and that fixed the out of memory error during compile.

andreamlin:
So the problem with the Appveyor build is that it runs on a Windows machine, which does *not* have a case sensitive file system. This PR includes separate HTTPHealthCheck.java and HttpHealthCheck.java, but one overwrites the other when imported into Appveyor's machine.

andreamlin:
PTAL

andreamlin:
This assumes https://github.com/googleapis/gapic-generator/pull/1956 is checked in.

andreamlin:
PTAL

garrettjonesgoogle:
🎉

PR #2986, opened 2018-03-02
BIgtable: 11. Implement ReadRows retries

igorbernstein2:

PR #3103, opened 2018-03-29
Bigtable: Implement query sharding by generalizing ReadRows resume request builder.

igorbernstein2:
This extends the work done in #2986 to allow map reduce style frameworks like beam to split queries into shards and execute them in parallel. The mechanism for chopping off part of query in a resume request is very similar to splitting a query into multiple shards. The main difference is how many splits are used.

The common functionality is extracted to an internal `RowSetUtil`  class that does all of the heavy lifting. The class is used both by`ReadRowsResumptionStrategy` for computing the resume request and the newly introduced `Query#shard` method.

Also expose the ability to get a Query's bounding range. The combination of Query#shard & Query#getBound is needed to implement a Beam source

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
@@ -172,7 +172,7 @@ public final class Query implements Serializable {
 
   /**
    * Split this query into multiple queries that can be evenly distributed across Bigtable nodes and
-   * be run im parallel. This method takes the results from {@link
+   * be run in parallel. This method takes the results from {@link
    * com.google.cloud.bigtable.data.v2.BigtableDataClient#sampleRowKeysAsync(String)} to divide this
    * query into a set of disjoint queries that logically combine into form this query.
    *
@@ -183,7 +183,7 @@ public final class Query implements Serializable {
    * List<Query> queryShards = myQuery.shard(keyOffsets);
    * List<ApiFuture<List<Row>>> futures = new ArrayList();
    * for (Query subQuery : queryShards) {
-   *   futures.add(dataClient.readRowsCallable().all().futureCall(subQuery))
+   *   futures.add(dataClient.readRowsCallable().all().futureCall(subQuery));
    * }
    * List<List<Row>> results = ApiFutures.allAsList(futures).get();
    * }</pre>
@@ -214,7 +214,7 @@ public final class Query implements Serializable {
    * List<Query> queryShards = myQuery.shard(splitPoints);
    * List<ApiFuture<List<Row>>> futures = new ArrayList();
    * for (Query subQuery : queryShards) {
-   *   futures.add(dataClient.readRowsCallable().all().futureCall(subQuery))
+   *   futures.add(dataClient.readRowsCallable().all().futureCall(subQuery));
    * }
    * List<List<Row>> results = ApiFutures.allAsList(futures).get();
    * }</pre>


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+        // Figure out the endpoint and remainder.
+        int endCmp = EndPoint.extract(rowRange).compareTo(new EndPoint(splitPoint, true));
+        if (endCmp <= 0) {
+          // The range is fully contained in the segment.
+          segment.addRowRanges(rowRange);
+
+          // Consume the range, but take care to shift partially consumed ranges.
+          // Partially consumed ranges will only exist if the original RowSet had overlapping
+          // ranges, this should be a rare occurrence.
+          System.arraycopy(
```
kevinsi4508: Pretty difficult to understand this. Maybe do some preprocessing to get rid of overlaps first? Preprocessing will make resulting ranges simpler.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+
+    @Nonnull
+    static StartPoint extract(@Nonnull RowRange rowRange) {
+      switch (rowRange.getStartKeyCase()) {
+        case STARTKEY_NOT_SET:
+          return new StartPoint(ByteString.EMPTY, true);
+        case START_KEY_CLOSED:
+          return new StartPoint(rowRange.getStartKeyClosed(), true);
+        case START_KEY_OPEN:
+          // Take care to normalize an open empty start key to be closed.
```
kevinsi4508: Very uneasy about this translation. 

For example, translating back will give us wrong start key.

if (minStartPoint.isClosed) {
       boundingRange.startClosed(minStartPoint.value);
 }

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+
+    @Nonnull
+    static EndPoint extract(@Nonnull RowRange rowRange) {
+      switch (rowRange.getEndKeyCase()) {
+        case ENDKEY_NOT_SET:
+          return new EndPoint(ByteString.EMPTY, true);
+        case END_KEY_CLOSED:
+          return new EndPoint(rowRange.getEndKeyClosed(), true);
+        case END_KEY_OPEN:
+          // Take care to normalize an open empty end key to be closed.
```
kevinsi4508: same concern here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
```diff
@@ -147,6 +154,76 @@ public Query limit(long limit) {
...
+    ImmutableSortedSet.Builder<ByteString> splitPoints =
+        ImmutableSortedSet.orderedBy(ByteStringComparator.INSTANCE);
+
+    for (KeyOffset keyOffset : sampledRowKeys) {
+      if (!keyOffset.geyKey().isEmpty()) {
+        splitPoints.add(keyOffset.geyKey());
+      }
+    }
+
+    return shard(splitPoints.build());
```
kevinsi4508: Unless we need to continue adding elements in there to be sorted, we perhaps to convert it  to a list.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+public final class RowSetUtil {
+  private RowSetUtil() {}
+
+  /**
+   * Splits the provided {@link RowSet} on the provided split and returns the second segment. If the
+   * second segment is empty, then it will return null. Note that the split point represents the
+   * last row in the first segment.
+   */
+  @Nullable
+  public static RowSet getRemaining(@Nonnull RowSet rowSet, @Nonnull ByteString split) {
```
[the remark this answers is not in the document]
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+    ImmutableSortedSet<ByteString> splitPoints =
+        ImmutableSortedSet.orderedBy(ByteStringComparator.INSTANCE).add(split).build();
```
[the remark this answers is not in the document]
igorbernstein2: No I think that will be worse. I would like to make it easy for the caller to correlate the resulting RowSets to the provided splitPoints. Doing an internal sort will prevent that. Having an an explicit ordering associated with a SortedSet is part of its api.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+  /**
+   * Splits the provided {@link RowSet} into segments partitioned by the provided {@code
+   * splitPoints}. Each split point represents the last row of the corresponding segment. The row
+   * keys contained in the provided {@link RowSet} will be distributed across the segments. Each
+   * range in the {@link RowSet} will be split up across each segment.
+   *
+   * @see #split(RowSet, SortedSet, boolean) for more details.
+   */
+  @Nonnull
+  public static List<RowSet> shard(
```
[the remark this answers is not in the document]
igorbernstein2: It's a bit subtle but the distinction is:
Shard is the verb that takes splits to create segments. A split is the dividing line between between 2 segments. Each segment is represented by a RowSet.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+   * Split a {@link RowSet} into segments.
+   *
+   * <p>Each segment is defined by a split point. The split point identifies the segment's inclusive
+   * end. This means that the first segment will start at the beginning of the table and extend to
+   * include the first split point. The last segment will start just after the last split point and
+   * extend until the end of the table. The maximum number of segments that can be returned is the
+   * number of split points + 1.
+   *
+   * <p>Each segment is represented by a RowSet in the returned List. Each of the returned RowSets
+   * will contain all of the {@link RowRange}s and keys that fall between the previous segment and
```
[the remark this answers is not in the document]
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+   * this segment's split point. If there are no {@link RowRange}s or keys that belong to a segment,
+   * then that segment will either be omitted or if {@code allowEmptySplit} is true, then it will be
```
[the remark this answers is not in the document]
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+   * represented by a null value in the returned list.
+   *
+   * <p>The segments in the returned list are guaranteed to be sorted. If {@code allowEmptySplit} is
+   * true, then it will have exactly {@code splitPoints.size() + 1} items.
```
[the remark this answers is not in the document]
igorbernstein2: It will contain a null segment at the end to represent the empty space between 3 and the end of the table.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+    int rowRangesStart = 0;
+
+    // Keys and ranges that lie before the current split point,
+    RowSet.Builder segment = RowSet.newBuilder();
+    boolean isSegmentEmpty = true;
+
+    for (ByteString splitPoint : splitPoints) {
+      Preconditions.checkState(!splitPoint.isEmpty(), "Split point can't be empty");
+
+      // Consume all of the row keys that lie to the left of the split point. Consumption is
```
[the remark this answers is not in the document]
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
igorbernstein2: I played around with implementing range merging and it's non-trivial. It creates quite a bit of noise and has very little practical benefit because most queries will not have overlapping rows. At some point in the future, if we need to implement RowSet merging, then it would make sense to premerge the ranges. But for now I updated the comments to give a better explanation of whats happening

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -0,0 +1,346 @@
...
+          segment.addRowRanges(leftSubRange);
+          // Save the remainder for the next segment. This is done by replacing the current rowRange
+          // with the remainder and not advancing rowRangesStart.
+          RowRange rightSubRange = rowRange.toBuilder().setStartKeyOpen(splitPoint).build();
+          rowRanges[i] = rightSubRange;
+        }
+      }
+
+      // Build the current segment
+      if (!isSegmentEmpty) {
```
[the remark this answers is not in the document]
igorbernstein2: I'm not sure what you mean by segment != empty. Do you mean `segment.build().equals(RowSet.defaultInstance())`? If so, I think keeping an extra bool around is cleaner and more performant.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
igorbernstein2: I'm not sure what you mean here. I updated the code to be explicit about handling the empty string case. But I don't understand what you mean about translating back giving the wrong start key.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
igorbernstein2: Please see the comment for `getRemaining`, but the short version is that I want the input be pre-sorted and SortedSet allows me to communicate that to the caller

kevinsi4508: Well, by converting to a list, you get to keep the sorted order, no?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
kevinsi4508: sgtm

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtilTest.java
```diff
@@ -0,0 +1,573 @@
...
+            .addRowKeys(ByteString.copyFromUtf8("a"))
+            .addRowKeys(ByteString.copyFromUtf8("d"))
+            .build();
+
+    ByteStringRange actual = RowSetUtil.getBound(rowSet);
+    assertThat(actual).isEqualTo(ByteStringRange.unbounded().startClosed("a").endClosed("d"));
+  }
+
+  @Test
+  public void singleRange1BoundTest() {
```
kevinsi4508: What does it mean when you say?
singleRange1
singleRange2
singleRange3

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtilTest.java
```diff
@@ -0,0 +1,573 @@
...
+    RowSet rowSet =
+        RowSet.newBuilder()
+            .addRowRanges(RowRange.newBuilder().setEndKeyClosed(ByteString.copyFromUtf8("z")))
+            .build();
+    ByteStringRange actual = RowSetUtil.getBound(rowSet);
+    assertThat(actual).isEqualTo(ByteStringRange.unbounded().endClosed("z"));
+  }
+
+  @Test
+  public void multipleRangeBoundTest() {
```
kevinsi4508: What about multipleRange overlapped and mixed of row keys and ranges?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtilTest.java
```diff
@@ -0,0 +1,573 @@
...
+import com.google.common.collect.ImmutableSortedSet;
+import com.google.protobuf.ByteString;
+import java.util.List;
+import java.util.SortedSet;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class RowSetUtilTest {
```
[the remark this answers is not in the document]
igorbernstein2: done

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtilTest.java
igorbernstein2: done

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtilTest.java
igorbernstein2: done

igorbernstein2:
@garrettjonesgoogle This is ready for review/merging. It's pretty dense bigtable specific code that Kevin reviewed. Feel free to skip the +2 review and just merge

igorbernstein2:
rebased

igorbernstein2:
This should be ready for review/merging

igorbernstein2:
@garrettjonesgoogle friendly ping

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
```diff
@@ -163,6 +170,76 @@ public Query limit(long limit) {
     return this;
   }
 
+  /**
+   * Split this query into multiple queries that can be evenly distributed across Bigtable nodes and
+   * be run im parallel. This method takes the results from {@link
```
garrettjonesgoogle: s/im/in/

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
```diff
@@ -163,6 +170,76 @@ public Query limit(long limit) {
...
+   * <p>Expected Usage:
+   *
+   * <pre>{@code
+   * List<KeyOffset> keyOffsets = dataClient.sampleRowKeysAsync("my-table").get();
+   * List<Query> queryShards = myQuery.shard(keyOffsets);
+   * List<ApiFuture<List<Row>>> futures = new ArrayList();
+   * for (Query subQuery : queryShards) {
+   *   futures.add(dataClient.readRowsCallable().all().futureCall(subQuery))
+   * }
+   * List<List<Row>> results = ApiFutures.allAsList(futures).get();
```
garrettjonesgoogle: I think the carets need to be escaped or javadoc will barf

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
igorbernstein2: done

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
igorbernstein2: Not sure what you mean. Do you mean that the `<>` need to be escaped? If so, I don't think thats true...`@code{}` renders them correctly

garrettjonesgoogle: All I know is that we escape them in our generated sample code because it caused a problem for us.

igorbernstein2: The generated code samples use `<code>` instead of `{@code}`.  See here for a good breakdown:
https://reflectoring.io/howto-format-code-snippets-in-javadoc/#pre--code

The bigtable client  uses this style throughout, for example:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java#L142
and it seems to be working ok

garrettjonesgoogle: Ahh, `{@code}` doesn't handle `@` nicely, that must be why we used `<code>`. What a pain.

PR #3106, opened 2018-03-29
Bigtable: Make models serializable.

igorbernstein2:
This along with query sharding, enables the creation of a Beam Source


igorbernstein2:
@garrettjonesgoogle This is ready for review/merging

PR #3110, opened 2018-03-29
Support for BigQuery's NUMERIC type

ElliottBrossard:
BigQuery's NUMERIC type is [currently in alpha](https://issuetracker.google.com/issues/35906014), which is restricted to whitelisted projects. I'm sending this pull request a bit early--it should not be merged until the NUMERIC type is in beta and it's possible to run integration tests--but I would like to get some initial feedback in case there are any major issues to fix before we get to that point.

Thanks!

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.io.BaseEncoding;
 import java.io.Serializable;
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nullable;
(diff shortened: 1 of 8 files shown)


pongad:
@tswast

ElliottBrossard:
Thanks, Tim. My understanding of that code is that it converts whatever Objects it receives to Strings, right? I think it should "just work" with BigDecimal, or any other numeric class for that matter. ITBigQueryTest.java should exercise the conversion logic, although I can update other tests too if I missed something.

ElliottBrossard:
I think I straightened out the commit history, and the tests pass. Any other comments?

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -269,6 +278,11 @@ public static QueryParameterValue dateTime(String value) {
           return value.toString();
         }
         break;
+      case NUMERIC:
+        if (value instanceof BigDecimal) {
+          return value.toString();
```
pongad: Sorry I just got to this. Do we need to round this value so that it only has 9 decimal points?

ElliottBrossard: In the Python code (https://github.com/GoogleCloudPlatform/google-cloud-python/pull/5331), we just convert whatever value is provided into a string, since users have control over the precision of the values. If there are two may digits after the decimal point, BigQuery will return an error when it receives the value, so users are given an explicit signal that they should round or truncate rather than having it happen silently.

tswast: +1 to explicit errors as opposed to silently dropping data.

pongad: LGTM

PR #3112, opened 2018-03-30
Spanner: migrate all unary call methods to gapic and inject headers

yihanzhen:
- Migrate all unary call methods to use `GapicSpannerRpc`
- Update `SpannerMetadataProvider` to generate headers in the format of `Map<String, List<String>>`
- Put headers into `CallContext` for each unary call method

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -147,6 +147,10 @@ public class GapicSpannerRpc implements SpannerRpc {
                 .build());
     CredentialsProvider credentialsProvider =
         GrpcTransportOptions.setUpCredentialsProvider(options);
+    
+    // Disabling retry for now because spanner handles retry in SpannerImpl.
+    // We will finally want to improve gax but for smooth transitioning we
+    // preserve the retry in SpannerImpl
     try {
       // TODO: bump the version of gax and remove this try-catch block
       // applyToAllUnaryMethods does not throw exception in the latest version


hzyi-google:
PTAL @garrettjonesgoogle @pongad @vkedia

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -806,7 +806,7 @@ public Timestamp writeAtLeastOnce(Iterable<Mutation> mutations) throws SpannerEx
                 new Callable<CommitResponse>() {
                   @Override
                   public CommitResponse call() throws Exception {
-                    return rawGrpcRpc.commit(request, options);
+                    return gapicRpc.commit(request, options);
```
vkedia: does gapicRpc do retries? We do not do any retries in GrpcSpannerRPC and instead they are done at a higher level. Is that behavior preserved here.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -349,33 +388,47 @@ public StreamingCall executeQuery(
   @Override
   public Transaction beginTransaction(
       BeginTransactionRequest request, @Nullable Map<Option, ?> options) throws SpannerException {
-    throw new UnsupportedOperationException("Not implemented yet.");
+    GrpcCallContext context = GrpcCallContext.createDefault()
+        .withChannelAffinity(Option.CHANNEL_HINT.getLong(options).intValue())
+        .withExtraHeaders(metadataProvider.newExtraHeaders(request.getSession(), projectName));
```
vkedia: Can you extract out this piece of code into a helper method. It is repeated in multiple places.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
pongad: gapicRpc does retry. I take this to mean we might retry more than we want?

hzyi-google:
@vkedia PTAL. I did not make any change to the retry logic in SpannerImpl since the only difference is we might retry more, but if you think we should get rid of retries there I can make the change.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
vkedia: So ideally we would want to get rid of the higher level retry. But last time we discussed this, I think not all the things we retry on were supported in GAPIC. The problem with retrying in two layers is that exponential backoff gets messed up. Also in spanner code we keep on retrying till the deadline specified in the grpc `Context` has elapsed. We would like to preserve that behavior.

pongad: @hzyi-google I think the best thing to do at this point is to use `SpannerStubSettings.applyToAllUnaryMethods` and make all methods not retry. Then we can use the existing retry. Then we can improve gax and get rid of the higher level one after.

vkedia: Yeah I agree.

hzyi-google:
PTAL @pongad @vkedia

google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -140,25 +147,55 @@ public GapicSpannerRpc(SpannerOptions options) throws IOException {
...
+      // applyToAllUnaryMethods does not throw exception in the latest version
+      this.stub =
         GrpcSpannerStub.create(
             SpannerStubSettings.newBuilder()
                 .setTransportChannelProvider(channelProvider)
                 .setCredentialsProvider(credentialsProvider)
+                .applyToAllUnaryMethods(
+                    new ApiFunction<UnaryCallSettings.Builder<?, ?>, Void>() {
+                      @Override
+                      public Void apply(UnaryCallSettings.Builder<?,?> builder) {
```
vkedia: can you add a comment here explaining why we are turning off retries.

hzyi-google: Sure.

PR #3113, opened 2018-03-31
Let surefire not run smoke tests

neozwu:


pom.xml
```diff
@@ -478,6 +478,7 @@
           <!-- Excludes integration tests when unit tests are run -->
           <excludes>
             <exclude>**/IT*.java</exclude>
+            <exclude>**/*SmokeTest.java</exclude>
```
garrettjonesgoogle: Update the comment above: "integration tests and smoke tests"

neozwu: Done

PR #3121, opened 2018-04-04
bigquery: add location property

pongad:
We also add JobId.Builder since the static constructors are getting
unwieldy.
The property isn't threaded through in the right places yet, so we keep
it package-private for now.


pongad:
FYI @jjolma

pongad:
@tswast I created https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3126 to track.

Issue #3065, opened 2018-03-20
[pubsub] Subscriber.stopAsync().awaitTerminated() doesn't wait for all acks to be sent


prodonjs:
In working on a gRPC emulation service for Cloud Pub/Sub, I was trying to validate correct behavior by publishing a series of messages, receiving and acknowleding the exact same set of messages, and then ensuring that no backlog existed for the subscription.

Here's the general test pattern:
```
String messagePrefix = subscriptionProperties.getName() + System.currentTimeMillis() + "-";
int messages = 5000;
Set<String> messagesSet = new TreeSet<>();
for (int i = 0; i < messages; i++) {
  messagesSet.add(messagePrefix + i);
}
CountDownLatch countDownLatch = new CountDownLatch(messages);
Set<String> publishedIds = new ConcurrentSkipListSet<>();
Map<String, Integer> receivedIds = new ConcurrentHashMap<>();
for (String data : messagesSet) {
  PubsubMessage message =
      PubsubMessage.newBuilder().setData(ByteString.copyFromUtf8(data)).build();
  ApiFutures.addCallback(
      publisher.publish(message),
      new ApiFutureCallback<String>() {
        @Override
        public void onFailure(Throwable throwable) {
          LOGGER.warning("Error on Publish " + throwable.getMessage());
        }

        @Override
        public void onSuccess(String s) {
          publishedIds.add(s);
        }
      });
}
publisher.shutdown();
assertEquals(messages, publishedIds.size());

subscriber =
    Subscriber.newBuilder(
            ProjectSubscriptionName.of(PROJECT, subscriptionProperties.getName()),
            (message, consumer) -> {
              consumer.ack();
              LOGGER.fine("Received and Acknowledging " + message.getMessageId());
              if (!receivedIds.containsKey(message.getMessageId())) {
                receivedIds.put(message.getMessageId(), 1);
              } else {
                int current = receivedIds.get(message.getMessageId());
                receivedIds.put(message.getMessageId(), Integer.valueOf(++current));
              }
              countDownLatch.countDown();
            })
        //.setChannelProvider(channelProvider)
        //.setCredentialsProvider(credentialsProvider)
        .build();

publisher.shutdown();
assertEquals(messages, publishedIds.size());

subscriber.startAsync();
LOGGER.info("Publisher complete, waiting for 5,000 messages to be received by Subscriber");
countDownLatch.await();
LOGGER.info("Shutting down Subscriber");
[...]

PR #3123, opened 2018-04-04
pubsub: use unary calls to send acks/modacks

pongad:
Previously we send ack/modack requests on the StreamingPull streams.
However, if pull responses are buffered, gRPC cannot promptly tell us
that the stream is broken, so we'll keep sending requests on the dead
stream.

This commit implements a temporary solution. We send requests by unary
RPCs instead and only use streaming for receiving messages.

Load test shows no regression. However, modacks seem to take longer to
take effect. The timeout for nack test had to be increased.

cc @mdietz94


mdietz94:
This LGTM. It is expected this would add additional latency to ack/mod ack operations, so all buffer amounts should be increased accordingly, but this looks good.

j256:
For the record, this fix may also have the side effect of properly ack-ing messages when the subscriber is shutdown with acks outstanding.  I was tracking a problem where acks were getting lost and it goes away in 0.43.  Thanks!

j256:
This also might actually be the change that fixed #3065.

pongad:
Thank you for tracking this down. This seems to be a better change than expected... :laughing:

PR #3129, opened 2018-04-05
Bigtable: Expose an API to allow users to create their own batches for MutateRows

igorbernstein2:
This has 2 objectives:
1. Expose enough primitives for users to workaround the lack of batch flushing.
This would allow me to experiment with @pongad's batching work and unblock integration with apache beam

2. Unblock a transition path from the existing [bigtable-client-core](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/tree/master/bigtable-client-core-parent/bigtable-client-core).
The current plan for a future transition from bigtable-client-core is to implement the [BigtableDataClient](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java) using this veneer. That interface sits below batching in the current client. To enable this transition path, I need to expose the manual batching primitive

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
@@ -147,7 +147,7 @@ public class EnhancedBigtableStubSettings extends StubSettings<EnhancedBigtableS
   }
 
   /**
-   * Returns the object with the settings used for to MutateRows.
+   * Returns the object with the settings used for calls to MutateRows.
    *
    * <p>Please note that these settings will affect both manually batched calls
    * (bulkMutateRowsCallable) and automatic batched calls (bulkMutateRowsBatchingCallable). The


igorbernstein2:

pongad:
The code LGTM. IIUC, we're exposing `createBulkMutateRowsCallable` so that we can batch manually is that right?

igorbernstein2:
Yep. Manual batching is needed to transition the current client to veneer and I plan to leave that feature around for advanced use cases

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/BulkMutationTest.java
```diff
@@ -0,0 +1,119 @@
...
+                        .setRowKey(ByteString.copyFromUtf8("key-b"))
+                        .addMutations(
+                            com.google.bigtable.v2.Mutation.newBuilder()
+                                .setSetCell(
+                                    SetCell.newBuilder()
+                                        .setFamilyName("fake-family3")
+                                        .setColumnQualifier(
+                                            ByteString.copyFromUtf8("fake-qualifier3"))
+                                        .setTimestampMicros(3_000)
+                                        .setValue(ByteString.copyFromUtf8("fake-value3")))))
```
garrettjonesgoogle: This is a *lot* of indentation. Could you break this up into some temporary variables?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/mutaterows/BulkMutateRowsUserFacingCallable.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.api.core.ApiFuture;
+import com.google.api.core.InternalApi;
+import com.google.api.gax.rpc.ApiCallContext;
+import com.google.api.gax.rpc.UnaryCallable;
+import com.google.bigtable.v2.MutateRowsRequest;
+import com.google.cloud.bigtable.data.v2.internal.RequestContext;
+import com.google.cloud.bigtable.data.v2.models.BulkMutation;
+
+/**
+ * A callable to convert userfacing {@link BulkMutation}s to {@link MutateRowsRequest}s.
```
garrettjonesgoogle: s/userfacing/user-facing/

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
```diff
@@ -0,0 +1,104 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * Represents a list of mutations for multiple rows. Each mutation contains multiple changes that
+ * will be atomically applied to each row. However, ordering between rows is not guaranteed.
+ *
+ * <p>This class is meant for manual batching, while {@link BulkMutationBatcher} is meant for
+ * automatic batching with flow control.
+ */
+public final class BulkMutation implements Serializable {
```
garrettjonesgoogle: Don't you need a serialization id in the class?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -146,9 +146,9 @@ public String getAppProfileId() {
     return mutateRowSettings;
   }
 
-  /** Returns the object with the settings used for calls to MutateRows. */
-  public BatchingCallSettings<RowMutation, Void> mutateRowsSettings() {
-    return mutateRowsSettings;
+  /** Returns the object with the settings used for manually batched calls to MutateRows. */
+  public BatchingCallSettings<RowMutation, Void> bulkMutateRowsSettings() {
```
garrettjonesgoogle: The `ResponseT` of the call settings doesn't match the `ResponseT` of the callable in the stub.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/BulkMutationTest.java
igorbernstein2: It's still pretty ugly with temp vars. I switched it to a text proto

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/mutaterows/BulkMutateRowsUserFacingCallable.java
igorbernstein2: done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
igorbernstein2: yep, thanks for catching that

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
igorbernstein2: The 2 callables that depend on these settings are
```
UnaryCallable<BulkMutation, Void> bulkMutateRowsCallable
UnaryCallable<RowMutation, Void> bulkMutateRowsBatchingCallable
```
Both use `Void` for `ResponseT`. However, the first one differs in `RequestT`. This is the result of them sharing the same underlying `ServerStreamingCallable<MutateRowsRequest, MutateRowsResponse> mutateRowsCallable()` from GAPIC, so their retry codes (that are passed to the underlying exception callable) must be identical. So keep them consistent, I chose to use a single settings object that is the super set of the 2 user-facing callables.

An alternative would be for me to create my own BatchingSettings class that only contains the flow & batching settings. But I think that would be more confusing

igorbernstein2:
All feedback should be addressed, PTAL

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
garrettjonesgoogle: Maybe add a comment about the mismatch.

igorbernstein2:
Updated, please merge when ready

PR #3131, opened 2018-04-05
bigquery: implement regionalization

pongad:


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/BigQueryRpc.java
```diff
@@ -29,6 +30,7 @@
 import com.google.cloud.bigquery.BigQueryException;
 import java.util.Map;
 
+@InternalExtensionOnly
```
pongad: This is technically breaking change to a public interface, but I don't think we can do anything if we don't do this :(

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
```diff
@@ -1176,4 +1176,7 @@ TableResult query(QueryJobConfiguration configuration, JobId jobId, JobOption...
    * @throws BigQueryException upon failure
    */
   TableDataWriteChannel writer(WriteChannelConfiguration writeChannelConfiguration);
+
+  // TODO(pongad): document
+  TableDataWriteChannel writer(JobId jobId, WriteChannelConfiguration writeChannelConfiguration);
```
tswast: 👍 

I assume you'll document before merging?

tswast:
We should follow this PR with another change that adds location to BigQueryOptions or something (a peer to where the default project is set).

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java
pongad: Will do this also in a follow up

PR #3132, opened 2018-04-06
internal: improve pubsub javadoc

jeanbza:
This is addressing some friction I had, as someone who hadn't used executors before. I think it's good for this code to be as copy-paste-able as possible.

Also, being new to executors, let me know if there's a better executor to be the default recommendation.


google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java
```diff
@@ -78,7 +74,7 @@ public void failed(Subscriber.State from, Throwable failure) {
             // Handle error.
           }
         },
-        executor);
+        Executors.newCachedThreadPool());
```
pongad: I think we also need to shutdown this executor.

jadekler: Done

jadekler:
@pongad Whoops, didn't try compiling (doh). Fixed the issue

PR #3143, opened 2018-04-10
Fix site (documentation) generation

### LINKED ISSUE #3084: javadoc takes a long time to generate
Sometimes it takes 2-ish hours to do a release, most of which is javadoc. We need to speed this up for agility. Currently javadoc generation only uses one core; maybe we can parallelize this somehow.

vam-google:
1) Fix #3084. Now `create_site.sh` takes `2 min 50 sec` on my workstation (compared to around `1 hr 30 min` before). Done mostly by disabling unused/incorrect reports generation.
2) Remove unused/incorrect reports (reduces size of generated site by factor of 2). The most time consuming were the cobertura test coverage, code style and dependencies reports:

    2.1) For **cobertura test coverage**, since we were generating the report with `-DdisableTests` flag, the test coverage was constantly reported as 0 with huge reports full of zeros and "red" untested code: https://googlecloudplatform.github.io/google-cloud-java/0.42.1/cobertura/index.html

    2.2) For **code style**: it was far the most time consuming report, results of which are huge and which we ignore and they seem have no real value in our workflow (if it did, we should enforce that rules on build and fix them, instead of generating reports):  https://googlecloudplatform.github.io/google-cloud-java/0.42.1/checkstyle-aggregate.html

    2.3) **Dependencies** report has marginal use, and is the only one I removed not because it was wrong, but was almos useless, but was the second most time consuming report among all (after the code style).

3) Fix `google-cloud-bom` documentation discrepancies (looked completely differently, needed extra configuration, since it is not a child of the main pom)
4) Fix site multimodule cross links (currently https://googlecloudplatform.github.io/google-cloud-java/0.42.1/modules.html is completely broken, for example try clicking on `Google Cloud Java BOM`, `Google Cloud Dataproc` and `Google Cloud Datastore` - all give different results and only dataproc kind of works).

More changes will be done in the context of moving grpc and proto artifacts under google-cloud-java (our maven plugins **need update** and currently **we use deprecated** (unsupported in newer versions) configuration.

### DIFF
--- a/google-cloud-bom/src/site/site.xml
+++ b/google-cloud-bom/src/site/site.xml
@@ -1,5 +1,5 @@
 <!--
-  ~ Copyright 2018 Google Inc. All Rights Reserved.
+  ~ Copyright 2018 Google LLC
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
   ~ you may not use this file except in compliance with the License.
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBatchReadTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITBatchReadTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc. All Rights Reserved.
+ * Copyright 2017 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


google-cloud-bom/src/site/site.xml
```diff
@@ -0,0 +1,32 @@
+<!--
+  ~ Copyright 2018 Google Inc. All Rights Reserved.
```
garrettjonesgoogle: s/Google Inc. All Rights Reserved./Google LLC/

PR #3148, opened 2018-04-12
core: add parent resource names

pongad:


google-cloud-core/src/main/java/com/google/cloud/BillingName.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+    return fieldValuesMap;
+  }
+
+  public String getFieldValue(String fieldName) {
+    return getFieldValuesMap().get(fieldName);
+  }
+
+  /** @deprecated This method is only present to satisfy the ResourceName interface. */
+  @Deprecated
```
garrettjonesgoogle: This can be removed now (we'll remove it from what the protoc plugin generates later too)

google-cloud-core/src/main/java/com/google/cloud/BillingName.java
```diff
@@ -0,0 +1,168 @@
...
+import com.google.api.pathtemplate.PathTemplate;
+import com.google.api.resourcenames.ResourceNameType;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+// AUTO-GENERATED DOCUMENTATION AND CLASS
+@javax.annotation.Generated("by GAPIC protoc plugin")
```
garrettjonesgoogle: Maybe remove this auto-generated comment/annotation now, and add a comment saying something like "Formerly generated by ...".

google-cloud-core/src/main/java/com/google/cloud/BillingName.java
```diff
@@ -0,0 +1,168 @@
...
+public class BillingName extends ParentName {
```
garrettjonesgoogle: Mark everything `@BetaApi` for now

google-cloud-core/src/main/java/com/google/cloud/BillingName.java
pongad: Done

google-cloud-core/src/main/java/com/google/cloud/BillingName.java
pongad: Done

pongad:
@garrettjonesgoogle PTAL

PR #3151, opened 2018-04-13
move google-api-services-x and datastore-v1-proto-client to google-cloud-bom

yihanzhen:
Last time we restructured google-cloud-java/pom we google-cloud-bom/pom, we moved protobuf and grpc dependencies to bom. According to discussion with @vam-google, google-api-services-x and datastore-v1-proto-client are equivalent to these libraries in the sense of dependencies so moving them to bom as well.

### DIFF
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -233,6 +233,11 @@
         <type>test-jar</type>
       </dependency>
       <dependency>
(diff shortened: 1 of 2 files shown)


hzyi-google:
PTAL

garrettjonesgoogle:
Hmm this seems slightly different... the api-* packages are generated with a completely different system and are released independently from our project.

PR #3158, opened 2018-04-16
Add support for new api features in BigQuery TimePartitioning.

mikekap:
The two features are requirePartitionFilter & field.
Also this moves to using a builder instead of using more `of` methods since the number of combinations is large.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
@@ -19,8 +19,10 @@ package com.google.cloud.bigquery;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.core.BetaApi;
+import com.google.auto.value.AutoValue;
 import com.google.common.base.MoreObjects;
 
+import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
```diff
@@ -49,9 +52,92 @@
     DAY
   }
 
-  private TimePartitioning(Type type, Long expirationMs) {
-    this.type = checkNotNull(type);
-    this.expirationMs = expirationMs;
+  /**
+   * A builder for {@code TimePartitioning} objects.
+   */
+  public abstract static class Builder {
```
pongad: I'm a little confused by two `Builder`s are necessary. Since this class is mostly a "value class" with only some conversion methods, I think we can make life easier by using `@AutoValue`?

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
```diff
@@ -72,17 +158,37 @@ public Long getExpirationMs() {
     return expirationMs;
   }
 
+
+  /**
+   * If not set, the table is partitioned by pseudo column '_PARTITIONTIME'; if set, the table is
+   * partitioned by this field.
+   */
+  public String getField() {
```
pongad: I think `getField` and `getRequirePartitionFilter` should be `@BetaApi` as well.

google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TimePartitioning.java
mikekap: I've attempted this, but I've never worked with `@AutoValue`. Let me know if this looks right?

pongad:
Thank you for the PR! LGTM but @tswast should sign off.

PR #3159, opened 2018-04-16
Expose the schema of dry run query jobs.

mikekap:
This can be very useful for running transitive type checks.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
@@ -323,6 +323,7 @@ public abstract class JobStatistics implements Serializable {
     private final Long totalBytesBilled;
     private final Long totalBytesProcessed;
     private final List<QueryStage> queryPlan;
+    private final Schema schema;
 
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
```diff
@@ -437,14 +448,24 @@ public Long getTotalBytesProcessed() {
       return queryPlan;
     }
 
+
+    /**
+     * Returns the schema for the query result. Present only for successful dry run of
+     * non-legacy SQL queries.
```
pongad: I think this will be populated for all queries, not just dry runs right?

mikekap: I copied this from the API docs - I’ll check and report back.

pongad: @mikekap You're absolutely right. I didn't read the docs closely enough.

mikekap: yep - just checked, the schema is `null` for a real query. In case of a real result `TableResult` has a schema.

PR #3163, opened 2018-04-16
Throw the creation exception if there isn't a job created

mikekap:
`getJob` returns `null` when a job wasn't actually created - e.g. when you're running a dry run query


mikekap:
@pongad My motivation was dry run queries but it might be more general. Running a dry-run query job doesn't actually create a job. Without this change, a dry run query job with an invalid query just returned `null` instead of anything helpful. With this change, it throws a proper syntax error exception.

In a larger scope, this is in the error path for `create`ing a job. Specifically this is after the job fails creation with an exception. It seemed reasonable to me that if the create fails and the job isn't found (not like a 5XX scenario) we should rethrow the creation exception. See the comment above the code I changed.

pongad:
Ah you're right. And because this is in `create`, it shouldn't change behavior of getting jobs elsewhere. Thank you for the PR and explanation.

This LGTM, but @tswast should sign off.

PR #3165, opened 2018-04-17
add javadoc test to circleci

neozwu:


.circleci/config.yml
```diff
@@ -281,3 +291,7 @@ workflows:
           filters:
             branches:
               only: master
+      - javadoc_test:
+          filters:
+            branches:
+              only: master
```
garrettjonesgoogle: Actually we should do this one against all branches/PRs.

.circleci/config.yml
```diff
@@ -281,3 +291,4 @@ workflows:
           filters:
             branches:
               only: master
+      - javadoc_test
```
garrettjonesgoogle: Move this up above the it tests so that it's easier to see what runs on all branches.

.circleci/config.yml
```diff
@@ -221,6 +221,16 @@ jobs:
...
+  javadoc_test:
+    working_directory: ~/googleapis
+    <<: *anchor_docker
+    steps:
+      - checkout
+      - run:
+          <<: *anchor_build
+      - run:
+          name: Run mvn javadoc goal
+          command: mvn javadoc:jar javadoc:test-jar
```
garrettjonesgoogle: @vam-google could you verify this is a good target to call?

vam-google: Verified, meaning the following:
- `mvn javadoc:jar` and `mvn javadoc:test-jar` are reasonably fast (`javadoc:jar` takes `1 min 54 sec` and `javadoc:test-jar` takes `1 min 13 sec` on my machine);
- the commands seem to validate javadoc correctness (for example trying to reference a non-existent parameter in `@param` block makes the command to fail);

Though to run `mvn javadoc:jar` first `mvn install` must be executed (otherwise it fails with "artifact not found" exception). Taking this into account maybe it is just fine to put the javadoc command together with normal `mvn install` command and to not create a separate config for it.

neozwu: @vam-google  line 230 `<<: *anchor_build` is running `mvn -DskipITs=true install`. Do you think this is sufficient ?

garrettjonesgoogle: @neozwu I think @vam-google might have meant that we can verify javadoc within the main unit test jobs, because the javadoc is a superset.

While true, I think it would be useful to have the task be separate, so that we can see at a glance if CI is failing because of a build failure or a javadoc failure, and we will get build failures sooner this way.

vam-google: I'm ok with things the way they are now (as long as we guarantee absence of "artifact not found" exceptions when we run mvn javadoc (I don't know in which order targets are executed in circleci)).

garrettjonesgoogle: The jobs are run independently of each other - the install is always run before javadoc within the task that Neo defined, so this should work fine.

garrettjonesgoogle:
LGTM

PR #3166, opened 2018-04-17
Spanner gapic migration lro methods

yihanzhen:
Migrate longrunning methods to GAPIC including: `createDatabase`, `updateDatabaseDdl`, `createInstance`, and `updateInstance`.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
@@ -38,11 +38,9 @@ import com.google.cloud.Date;
 import com.google.cloud.PageImpl;
 import com.google.cloud.PageImpl.NextPageFetcher;
 import com.google.cloud.Timestamp;
-import com.google.cloud.spanner.Operation.Parser;
 import com.google.cloud.spanner.Options.ListOption;
 import com.google.cloud.spanner.Options.QueryOption;
 import com.google.cloud.spanner.Options.ReadOption;
-import com.google.cloud.spanner.spi.v1.GapicSpannerRpc;
 import com.google.cloud.spanner.spi.v1.SpannerRpc;
 import com.google.cloud.spanner.spi.v1.SpannerRpc.Paginated;
 import com.google.common.annotations.VisibleForTesting;
@@ -497,11 +495,12 @@ class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
         throws SpannerException {
       final String dbName = getDatabaseName(instanceId, databaseId);
       final String opId = operationId != null ? operationId : randomOperationId();
-      // TODO(hzyi) 
+      // TODO(hzyi)
       // Spanner checks the exception and if the error code is ALREADY_EXISTS
       // it creates a new Operation instead of throwing the exception. This
       // feature is not implemented in this PR but will come later
-      OperationFuture rawOperationFuture = rpc.updateDatabaseDdl(dbName, statements, opId);
+      OperationFuture<Empty, UpdateDatabaseDdlMetadata> rawOperationFuture =
+          rpc.updateDatabaseDdl(dbName, statements, opId);
       return new OperationFutureImpl(
           rawOperationFuture.getPollingFuture(),
           rawOperationFuture.getInitialFuture(),
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/OperationFutureUtil.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/OperationFutureUtil.java
@@ -249,11 +249,6 @@ class OperationFutureUtil {
 
       @Override
       public ResponseT get() throws ExecutionException, InterruptedException {
-        // if (completedSnapshot.getErrorCode().getCode().equals(StatusCode.Code.OK)) {
-        // return (ResponseT) completedSnapshot.getResponse();
-        // }
-        // throw new ExecutionException(
-        //     new ApiException(null, completedSnapshot.getErrorCode(), false));
         return (ResponseT) pollingFuture.get().getResponse();
       }
 
--- a/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDatabaseAdminTest.java
+++ b/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDatabaseAdminTest.java
@@ -110,6 +110,7 @@ public class ITDatabaseAdminTest {
 
   @Ignore("More work needs to be done")
   @Test
+  // TODO(hzyi)
   // Changing the surface to OperationFuture breaks updateDatabaseDdl in the case
   // that there is already a longrunning operation running. Disabling this test for 
   // this PR and I will fix this in the next PR. 


hzyi-google:
PTAL @garrettjonesgoogle @vkedia @pongad

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransformingOperationFuture.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.api.core.ApiFunction;
+import com.google.api.gax.longrunning.OperationFuture;
+import com.google.api.gax.longrunning.OperationSnapshot;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/** A wrapper class to augment proto results returned from get() */
+public class TransformingOperationFuture<ResponseT, RawResponseT, MetadataT>
```
pongad: Don't make this public yet.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransformingOperationFuture.java
```diff
@@ -0,0 +1,86 @@
...
+    implements OperationFuture<ResponseT, MetadataT> {
+
+  private OperationFuture<RawResponseT, MetadataT> rawOperationFuture;
+  private ApiFunction<RawResponseT, ResponseT> transformingFunction;
+
+  public TransformingOperationFuture(
+      OperationFuture<RawResponseT, MetadataT> rawOperationFuture,
+      ApiFunction<RawResponseT, ResponseT> transformingFunction) {
```
pongad: I think we should add exception transformer here too. That way we can transform the exception from `get` into `SpannerException`, and callers don't need to change.

hzyi-google: @garrettjonesgoogle @pongad Sorry if I didn't catch the entire idea of this. Even if we can convert any of `ExecutionException`, `TimeoutException` or `InterruptedException` into a `SpannerException`, as long as we can't change the signature of `get()` and `get(long, Unit)` (since `TransformingOperationFuture` implements `OperationFuture`), callers still have to catch these exceptions, though they don't have to do the conversion themselves. Any ideas to work around this?

hzyi-google:
@vkedia My apologies for merging the previous [PR](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3139) before you approve. Michael and Garrett who are way more experienced than me have carefully reviewed the change and signed off so I merged it just to unblock myself. If you or anyone from your team think there is any issue we can definitely take a look together and see if extra work needs to be done.

google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransformingOperationFuture.java
pongad: We should get buy in from others first but I think this is fine. We are changing from `Operation` type to `Future` anyway. My thinking is that `get()` should throw `ExecutionException` but the cause of the exception should be a `SpannerException`.

hzyi-google:
An update that comes with the [PR](https://github.com/googleapis/gax-java/pull/527) in gax. This PR is partially finished (to make the PR in gax more straightforward), and after the one is gax is merged I will get this one done.

google-cloud-examples/src/main/java/com/google/cloud/examples/spanner/snippets/DatabaseAdminClientSnippets.java
```diff
@@ -53,7 +54,7 @@ public DatabaseAdminClientSnippets(DatabaseAdminClient dbAdminClient) {
   // [VARIABLE my_database_id]
   public Database createDatabase(String instanceId, String databaseId) {
     // [START createDatabase]
-    Operation<Database, CreateDatabaseMetadata> op = dbAdminClient
+    OperationFuture<Database, CreateDatabaseMetadata> op = dbAdminClient
```
vkedia: What happens when users cancel this future? Also if I just want to fire and forget and not poll for operation status, how would I do that?

google-cloud-examples/src/main/java/com/google/cloud/examples/spanner/snippets/DatabaseAdminClientSnippets.java
```diff
@@ -70,7 +71,12 @@ public Database createDatabase(String instanceId, String databaseId) {
                     + "  AlbumTitle   STRING(MAX)\n"
                     + ") PRIMARY KEY (SingerId, AlbumId),\n"
                     + "  INTERLEAVE IN PARENT Singers ON DELETE CASCADE"));
-    Database db = op.waitFor().getResult();
+    Database db;
+    try {
+      db = op.get();
+    } catch (Exception e) {
+      throw SpannerExceptionFactory.newSpannerException(e);
```
vkedia: Why would user create a SpannerException? Instead is it possible to have the ExecutionException returned by this method wrap a SpannerExceptiojn?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -440,29 +446,45 @@ private static String randomOperationId() {
...
-          });
+      try {
+        // return new TransformingOperationFuture<>(
+        // rpc.createDatabase(instanceName, createStatement, statements),
+        // new ApiFunction<com.google.spanner.admin.database.v1.Database, Database>() {
+        //   @Override
+        //   public Database apply(com.google.spanner.admin.database.v1.Database database) {
+        //     return Database.fromProto(database, DatabaseAdminClientImpl.this);
+        //   }
+        // });
```
vkedia: why is this commented?

google-cloud-examples/src/main/java/com/google/cloud/examples/spanner/snippets/DatabaseAdminClientSnippets.java
hzyi-google: @vkedia We realize that there is more work to do in gax before we can get this done. Please hold off reviewing for a bit. I'll change the title to WIP as well. Sorry that I should have let you know earlier.

google-cloud-examples/src/main/java/com/google/cloud/examples/spanner/snippets/DatabaseAdminClientSnippets.java
pongad: Isn't fire and forget a bit of an anti-pattern? You won't get notified if the operation fails. I think you can do
```java
ApiFutures.addCallback(client.createDatabase(...), new ApiCallback<Database>() {
  @Override public void onSuccess(Database db) {/*no op*/}
  @Override public void onFailure(Throwable t) {logger.log(t);}
});
```
to at least get notified.

@garrettjonesgoogle can confirm. AFAIK, cancelling the future stops the background polling job, but does not cancel the job on the server.

garrettjonesgoogle: Actually @vam-google might be a better person to confirm

vam-google: Yes, cancelling polling/retrying future does not cancel corresponding server task. This is by design, since there can be potentially 2+ polling jobs for same server task (in same or different processes/hosts), so cancelling one should not cancel the whole thing (i..e server-side and another future). Also a future on a client side can be cancelled or "broken" in any other way by programming mistake in the client-side business logic. Client side problems/bugs should not affect server-side execution. If somebody wants to cancel server side operation, a specific call must be made (which itself is retriable and uses same logic as retries/polls).

hzyi-google:
PTAL @garrettjonesgoogle @pongad @vkedia

vkedia:
Can we merge this PR directly into master? This is a breaking change so it will be good to do it sooner rather than later. Also this is the only breaking change that GAPIC would cause so once this part is done, I think rest of the GAPIC migration can happen independently of GA.

pongad:
Sorry for the delay. This LGTM. @garrettjonesgoogle Do you want to take a look?

@vkedia There are [some features](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3166/files#diff-3c8f2299f39c351aa79aba7d6a28dcadR500) missing. I also believe custom error messages are working properly yet. If we merge to master now, the next release will lose these features; are you OK with this?

google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -479,56 +489,36 @@ public Database call() throws Exception {
...
         throws SpannerException {
       final String dbName = getDatabaseName(instanceId, databaseId);
       final String opId = operationId != null ? operationId : randomOperationId();
-      Callable<Operation<Void, UpdateDatabaseDdlMetadata>> callable =
-          new Callable<Operation<Void, UpdateDatabaseDdlMetadata>>() {
+      // TODO(hzyi) 
+      // Spanner checks the exception and if the error code is ALREADY_EXISTS
+      // it creates a new Operation instead of throwing the exception. This
+      // feature is not implemented in this PR but will come later
+      OperationFuture rawOperationFuture = rpc.updateDatabaseDdl(dbName, statements, opId);
```
garrettjonesgoogle: Why doesn't this `OperationFuture` have type parameters?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/OperationFutureUtil.java
```diff
@@ -0,0 +1,276 @@
...
+        return get();
+      }
+
+      @Override
+      public ResponseT get() throws ExecutionException, InterruptedException {
+        // if (completedSnapshot.getErrorCode().getCode().equals(StatusCode.Code.OK)) {
+        // return (ResponseT) completedSnapshot.getResponse();
+        // }
+        // throw new ExecutionException(
+        //     new ApiException(null, completedSnapshot.getErrorCode(), false));
```
garrettjonesgoogle: Remove dead code?

google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDatabaseAdminTest.java
```diff
@@ -107,44 +108,44 @@ public void databaseOperations() throws Exception {
     db = dbAdminClient.getDatabase(testHelper.getInstanceId().getInstance(), dbId);
   }
 
+  @Ignore("More work needs to be done")
   @Test
+  // Changing the surface to OperationFuture breaks updateDatabaseDdl in the case
```
garrettjonesgoogle: Add "TODO" at the beginning of this comment

hzyi-google:
PTAL @garrettjonesgoogle

PR #3180, opened 2018-04-20
BigQuery: Standardizes region tags and adds query snippets from java-docs-samples

alixhami:
This PR:
+ Updates region tags to standard so they can be tracked with our sample tracking tool
+ Changes region tags that are Java-specific and only in reference docs to bigquery_java_untracked, because these do not need to be tracked in our sample tracking tool
+ Adds query samples from java-docs-samples so they can be deleted from that repo (reduces redundancy of samples)

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
@@ -50,7 +50,6 @@ import com.google.cloud.bigquery.JobStatistics.LoadStatistics;
 import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.LoadJobConfiguration;
 import com.google.cloud.bigquery.QueryJobConfiguration;
-import com.google.cloud.bigquery.QueryParameterValue;
 import com.google.cloud.bigquery.Schema;
 import com.google.cloud.bigquery.StandardTableDefinition;
 import com.google.cloud.bigquery.Table;
@@ -238,7 +237,7 @@ public class BigQuerySnippets {
     return deleted;
   }
 
- /**
+  /**
    * Example of listing the tables in a dataset, specifying the page size.
    */
   // [TARGET listTables(String, TableListOption...)]
@@ -252,7 +251,6 @@ public class BigQuerySnippets {
     // [END ]
     return tables;
   }
-
 
   /**
    * Example of listing the tables in a dataset.
@@ -337,9 +335,7 @@ public class BigQuerySnippets {
     // [START ]
     TableId tableId = TableId.of(datasetName, tableName);
     WriteChannelConfiguration writeChannelConfiguration =
-        WriteChannelConfiguration.newBuilder(tableId)
-            .setFormatOptions(FormatOptions.csv())
-            .build();
+        WriteChannelConfiguration.newBuilder(tableId).setFormatOptions(FormatOptions.csv()).build();
     TableDataWriteChannel writer = bigquery.writer(writeChannelConfiguration);
     // Write data to writer
     try {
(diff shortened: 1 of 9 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
```diff
@@ -167,14 +174,14 @@ public Table updateTable(String datasetName, String tableName, String newFriendl
   // [TARGET delete(String, DatasetDeleteOption...)]
   // [VARIABLE "my_dataset_name"]
   public Boolean deleteDataset(String datasetName) {
-    // [START deleteDataset]
+    // [START bigquery_java_untracked]
```
tswast: ~This sample is included at https://cloud.google.com/bigquery/docs/managing-datasets#delete-dataset The Python sample has region tag `bigquery_delete_dataset`.~ Nevermind, I see this is for a method overload.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
```diff
@@ -663,47 +673,268 @@ public boolean cancelJob(String jobName) {
...
+    QueryJobConfiguration queryConfig =
+        // Note that setUseLegacySql is set to false by default
+        QueryJobConfiguration.newBuilder(query)
+        // Save the results of the query to a permanent table. See:
+        // https://cloud.google.com/bigquery/docs/writing-results#permanent-table
+        .setDestinationTable(TableId.of(destinationDataset, destinationTable))
+        // Allow results larger than the maximum response size.
+        // If true, a destination table must be set. See:
+        // https://cloud.google.com/bigquery/docs/writing-results#large-results
+        .setAllowLargeResults(true)
```
tswast: `allowLargeResults` actually doesn't do anything for standard SQL queries. It is only needed for legacy SQL. We can remove it from this sample.

Perhaps you are planning to use this sample at https://cloud.google.com/bigquery/docs/writing-results#writing_large_results_using_legacy_sql ? Since that sample doesn't set to legacy SQL, it's actually incorrect. I'd be okay dropping the "large results" sample altogether, but if you don't mind making a legacySQL version of this sample, that's probably the best option for user experience.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
```diff
@@ -663,47 +673,268 @@ public boolean cancelJob(String jobName) {
...
+            "ts_value",
+            QueryParameterValue.timestamp(
+                // Timestamp takes microseconds since 1970-01-01T00:00:00 UTC
+                timestamp.getMillis() * 1000))
+        .build();
+
+    // Print the results.
+    DateTimeFormatter formatter = ISODateTimeFormat.dateTimeNoMillis().withZoneUTC();
+    for (FieldValueList row : bigquery.query(queryConfig).iterateAll()) {
+      for (FieldValue val : row) {
```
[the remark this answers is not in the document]
tswast: Since the "print" line always prints the first column, you can remove the inner "val" loop here.

tswast:
The [utilities/add_snippets_to_file.py](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/utilities/add_snippets_to_file.py) script takes two arguments:

- `snippets_path` : path to the snippets Java file. For example, [google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java).
- `class_path` : path to the corresponding Java file. For `BigQuerySnippets.java`, the corresponding source file is [google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQuery.java).

google-cloud-examples/src/main/java/com/google/cloud/examples/bigquery/snippets/BigQuerySnippets.java
```diff
@@ -174,14 +169,14 @@ public Table updateTable(String datasetName, String tableName, String newDescrip
   // [TARGET delete(String, DatasetDeleteOption...)]
   // [VARIABLE "my_dataset_name"]
   public Boolean deleteDataset(String datasetName) {
-    // [START bigquery_java_untracked]
+    // [START ]
```
tswast: 👌 Looks a little strange not to have the region tag, but makes sense to keep it this way since it's not included in cloud.google.com.

PR #3181, opened 2018-04-21
gcloud-datastore-emulator version 1.2.0 raising DatastoreException

Juroviol:
The version 1.2.0 is throwing an Exception when a key of a java.util.Map has a dot inside an entity. 

com.google.datastore.v1.client.DatastoreException: flattened property.name contains a path delimiter, and the entity contains one or more indexed entity value., code=INVALID_ARGUMENT

### DIFF
--- a/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
+++ b/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java
@@ -55,7 +55,7 @@ public class LocalDatastoreHelper extends BaseEmulatorHelper<DatastoreOptions> {
   private static final String GCLOUD_CMD_TEXT = "gcloud beta emulators datastore start";
   private static final String GCLOUD_CMD_PORT_FLAG = "--host-port=";
   private static final String VERSION_PREFIX = "cloud-datastore-emulator ";
-  private static final String MIN_VERSION = "1.2.0";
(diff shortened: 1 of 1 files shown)


Juroviol:
I signed it!

google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -295,7 +295,7 @@ public Process getProcess() {
     private boolean isGcloudInstalled() {
       Map<String, String> env = System.getenv();
       for (String envName : env.keySet()) {
-        if ("PATH".equals(envName)) {
+        if ("PATH".equalsIgnoreCase(envName)) {
```
pongad: I'm a little uneasy about this. In UNIX, these variables are case-sensitive. If `path` is set but `PATH` is not, we'll later crash because we'll try to invoke a command that's not in `PATH`.

pongad:
@wsh can you also take a look?

google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
wsh: +1.

JustinBeckwith:
Greetings folks!  Apologies for the length delay on this one.  Would you be willing to rebase and fix the merge conflicts?

PR #3189, opened 2018-04-24
Using the right compare function for FieldPaths

schmidt-sebastian:
We need to compare field paths segment per segment and not use the encoded paths.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldPath.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldPath.java
@@ -152,9 +152,4 @@ public abstract class FieldPath extends BasePath<FieldPath> implements Comparabl
   public String toString() {
     return getEncodedPath();
   }
-
-  @Override
-  public int compareTo(@Nonnull FieldPath other) {
-    return super.compareTo(other);
-  }
 }
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java
@@ -32,7 +32,7 @@ public class OrderTest {
 
   @Test
   public void verifyOrder() {
-    Value[][] groups = new Value[64][];
+    Value[][] groups = new Value[65][];
 
     groups[0] = new Value[] {nullValue()};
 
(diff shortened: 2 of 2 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java
```diff
@@ -86,10 +86,10 @@ public void verifyOrder() {
     groups[35] = new Value[] {referenceValue("projects/p1/databases/d1/documents/c1/doc2/c2/doc1")};
     groups[36] = new Value[] {referenceValue("projects/p1/databases/d1/documents/c1/doc2/c2/doc2")};
     groups[37] = new Value[] {referenceValue("projects/p1/databases/d1/documents/c10/doc1")};
-    groups[38] =
-        new Value[] {referenceValue("projects/p1/databases/dkkkkklkjnjkkk1/documents/c2/doc1")};
+    groups[38] = new Value[] {referenceValue("projects/p1/databases/d1/documents/c2/doc1")};
     groups[39] = new Value[] {referenceValue("projects/p2/databases/d2/documents/c1/doc1")};
     groups[40] = new Value[] {referenceValue("projects/p2/databases/d2/documents/c1-/doc1")};
+    groups[40] = new Value[] {referenceValue("projects/p2/databases/d3/documents/c1-/doc1")};
```
pongad: Why is 40 assigned twice?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldPath.java
```diff
@@ -155,6 +155,6 @@ public String toString() {
 
   @Override
   public int compareTo(@Nonnull FieldPath other) {
-    return getEncodedPath().compareTo(other.getEncodedPath());
+    return super.compareTo(other);
```
pongad: We can technically just not override this altogether right? If you want the delegation to superclass to be explicit, I'm OK with it too.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java
schmidt-sebastian: Thanks for noticing. Fixed.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldPath.java
schmidt-sebastian: I was debating between the two options, but I am fine with removing it.

PR #3207, opened 2018-04-27
Cleanup before protoc/grpc artifacts push

vam-google:
(this is to separate cleanup changes from the actual proto/grpc artifacts push, which is huge)
1) Cleanup `pom.xml`s (remove unused/duplicate, partially wrong configuration)
2) Update `versions.txt` to include `proto/grpc` artifacts version (will be pushed immediately after this change is pushed). Also update `bump_versions.py` and `replace_versions.py` to strip whitespaces and support empty lines.
3) Fix License to pint explicitly to `Apache 2.0` license (instead of Google Software License, which contains apache license in it (did compare the text, it is identical). Otherwise all third-party tools do not recognize the license (including github, looks like).
4) Add Andrea and Hanzhen to developers list.


google-cloud-language/pom.xml
```diff
@@ -1,11 +1,14 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <artifactId>google-cloud-language</artifactId>
   <version>1.27.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-language:current} -->
   <packaging>jar</packaging>
   <name>Google Cloud Natural Language</name>
-  <url>https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-language</url>
+  <url>https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-language
+  </url>
```
garrettjonesgoogle: This seems like an odd change...

google-cloud-notification/pom.xml
```diff
@@ -6,7 +6,7 @@
   <artifactId>google-cloud-notification</artifactId>
   <version>0.45.1-beta-SNAPSHOT</version><!-- {x-version-update:google-cloud-notification:current} -->
   <packaging>jar</packaging>
-  <name>Google Cloud Pub/Sub Notifications for Google Cloud Storage</name>
+  <name>Google Cloud Pub/Sub Notifications for GCS</name>
```
garrettjonesgoogle: Why this abbreviation?

google-cloud-util/google-cloud-compat-checker/README.md
```diff
@@ -1,41 +0,0 @@
-Google Cloud Java Compatibility Checker
```
garrettjonesgoogle: Actually I don't want to ditch the google-cloud-compat-checker, people actually use it for debugging.

versions.txt
```diff
@@ -1,48 +1,137 @@
...
+#proto-google-iam-v1:0.10.0:0.10.0
+
+#
+# google-api-pom versions
+#
+# parent
+google-api-pom:0.11.0:0.11.0
+# children
+#grpc-google-cloud-datastore-v1
+grpc-google-cloud-bigquerydatatransfer-v1:0.11.0:0.11.0
```
garrettjonesgoogle: It seems like the logical order would be to put the grpc packages after the proto ones

versions.txt
```diff
@@ -1,48 +1,137 @@
...
-google-cloud-trace:0.45.0-beta:0.45.1-beta-SNAPSHOT
-google-cloud-translate:1.27.0:1.27.1-SNAPSHOT
-google-cloud-util:0.45.0-alpha:0.45.1-alpha-SNAPSHOT
-google-cloud-video-intelligence:0.45.0-beta:0.45.1-beta-SNAPSHOT
-google-cloud-vision:1.27.0:1.27.1-SNAPSHOT
+
+#
+# common versions
+#
+#grpc-google-common-protos:1.9.0:1.9.0
```
garrettjonesgoogle: Why the commented-out lines?

versions.txt
vam-google: Well, my idea was to put them in order of "release". The more basic packages are on top

google-cloud-language/pom.xml
vam-google: It was autoformatting

google-cloud-notification/pom.xml
vam-google: Too long string makes maven output look ugly

versions.txt
vam-google: This was intended for future use, once api-common repo is created

PR #3216, opened 2018-05-01
core: add hashCode to NoCredentials

pongad:


pongad:
I am having a really hard time understanding how this is failing...

PR #3218, opened 2018-05-01
properly add equals() and compare fields

pongad:


pongad:
@garrettjonesgoogle PTAL

PR #3222, opened 2018-05-01
all: fix integration CI

pongad:
Proto- and grpc- packages have moved to this repository.
When we run integration on CI, we must build them, otherwise the
tasks will fail from not being able to find dependencies.


utilities/verify_single_it.sh
```diff
@@ -15,16 +15,8 @@ fi
...
-
-echo "----- building and installing parent pom -----"
-mvn -B -N install -DskipTests
-
-echo "----- building and installing shared modules -----"
-# storage is present because google-cloud-bigquery and google-cloud-notification depend on it
-# pubsub is present because google-cloud-notification depends on it
-mvn -B -pl google-cloud-core,google-cloud-core-http,google-cloud-core-grpc,google-cloud-storage,google-cloud-pubsub install -DskipTests
+echo "----- installing modules -----"
+mvn -B install -DskipTests
```
garrettjonesgoogle: How much longer will it take to run the integration tests with this change?

pongad: On my machine, the time increases from ~1:30 to ~3:30 for installing.

garrettjonesgoogle: @vam-google with your new structure, do you see a way of keeping the build-time optimization so that we don't have to rebuild everything for every integration test?

pongad: I'm wondering if 2min on integration test is a big deal. Currently integs run faster than unit tests, and humans don't really need to wait for them.

PR #3234, opened 2018-05-04
Added closing of the scheduled thread executor if appropriate

j256:
If we are using an `InstantiatingExecutorProvider` then we should `shutdown()` the executor so we don't leak threads.


pongad:
I think `ExecutorAsBackgroundResource` should already take care of this right?

PR #3242, opened 2018-05-08
bigquery: let users contruct TableResult for testing

### LINKED ISSUE #3228: unable do junit testing of TableResult outside of package
I am using table to result to get the result from query job and then using a FieldValueList to iterate the list of row and is working fine, However when I do the junit testing for this code , I tried the TableResultTest and getting error that TableResult() not available publicly. Is there any way I can populate the FiledValueList in TableResult and do the unit test on different package

 TableResult result = queryJob.getQueryResults();

for (FieldValueList row : result.iterateAll()) {
}

pongad:
Fixes #3228.

@tswast I'm comfortable making this public, but please let me know if you want to `@BetaApi` it or something.

### DIFF
--- a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
+++ b/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
@@ -18,6 +18,7 @@ package com.google.cloud.bigquery;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.api.core.InternalApi;
 import com.google.api.gax.paging.Page;
 import com.google.common.base.Function;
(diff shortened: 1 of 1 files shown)


tswast:
I guess [@VisibleForTesting](http://google.github.io/guava/releases/19.0/api/docs/com/google/common/annotations/VisibleForTesting.html) is a Guava-ism, which I've heard we want to be careful about exposing those types in interfaces.

garrettjonesgoogle:
Our equivalent for `@VisibleForTesting` is `@InternalApi` (add a text note in the annotation, "Exposed for testing")

pongad:
@garrettjonesgoogle PTAL

jackbarantee:
Thanks a lot a for making TableResult public

PR #3246, opened 2018-05-08
spanner: use joda LocalDate

### LINKED ISSUE #3157: com.google.cloud.Date has no interop w/ java.util.Date
It's very hard to convert c.g.c.Date to j.u.Date. Can c.g.c.Date add something similar to `toDate`, `fromDate` method?

pongad:
This also lets us nuke com.google.cloud.Date and fixes #3157.

### DIFF
--- a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
+++ b/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
@@ -21,7 +21,7 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.cloud.ByteArray;
 import com.google.cloud.Timestamp;
 import java.util.List;
-import org.joda.time.LocalDate;
+import org.threeten.bp.LocalDate;
 
 /**
  * Base class for assisting {@link StructReader} implementations.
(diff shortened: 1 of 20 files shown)


google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractStructReader.java
```diff
@@ -19,9 +19,9 @@
 import static com.google.common.base.Preconditions.checkState;
 
 import com.google.cloud.ByteArray;
-import com.google.cloud.Date;
 import com.google.cloud.Timestamp;
 import java.util.List;
+import org.joda.time.LocalDate;
```
garrettjonesgoogle: We should be using date types from threetenbp, not joda.

pongad: Oh! I saw we're using Joda in a few places, but those are internal. Will fix this.

pongad:
@garrettjonesgoogle PTAL

vkedia:
This is a breaking change and will likely break many clients. Why do we need to do this especially since we will eventually want to use java 8 LocalDate here?

pongad:
@vkedia Hm, I thought that this was something you wanted to do but didn't have time to do it. We can close this if you'd like.

vkedia:
No we want to move to java 8 LocalDate at some point but not to this.

vkedia:
cc @snehashah16

PR #3249, opened 2018-05-08
Update CreateTopicAndPublishMessages.java

michaelawyu:
Add the publisher error handler sample. As requested, the error handling part in the publishing quickstart sample is now removed.

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
@@ -83,8 +83,11 @@ public class CreateTopicAndPublishMessages {
   public static void publishMessagesWithErrorHandler() throws Exception {
     // [START pubsub_publish_error_handler]
     ProjectTopicName topicName = ProjectTopicName.of("my-project-id", "my-topic-id");
+    Publisher publisher = null;
 
-    try (Publisher publisher = Publisher.newBuilder(topicName).build()) {
+    try {
+      // Create a publisher instance with default settings bound to the topic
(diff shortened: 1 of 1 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
```diff
@@ -78,6 +80,54 @@ public static void publishMessages() throws Exception {
...
   }
 
+  public static void publishMessagesWithErrorHandler() throws Exception {
+    // [START pubsub_publish_error_handler]
+    ProjectTopicName topicName = ProjectTopicName.of("my-project-id", "my-topic-id");
+    Publisher publisher = null;
+
+    try {
+      // Create a publisher instance with default settings bound to the topic
+      publisher = Publisher.newBuilder(topicName).build();
```
pongad: Should we use the try-with-resource here?

michaelawyu:
@pongad Updated as requested, thanks!

michaelawyu:
@pongad Sorry, my bad; I didn't realize that Publisher class hasn't implemented java.lang.AutoClosable. I will revert the update now.

google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/CreateTopicAndPublishMessages.java
michaelawyu: It seems that Publisher hasn't implemented java.lang.AutoClosable :(

pongad: You're right. The shutdown is longrunning so it's not closeable in the "normal" sense.

PR #3253, opened 2018-05-10
BigQuery: add missing query statistics.

shollyman:
Plumb through statistics in the API representation for query statistics:

Statement Type: SELECT, INSERT, CREATE_TABLE, etc
DDL Target Table and Operation Performed (how DDL statements manipulate their target)
DML: number of affected rows (e.g. rows affected by a UPDATE ... WHERE ...)
Referenced Tables:  which tables were referenced by a query.
Total Partitions Processed
Total slot-milliseconds (query cost from a computation standpoint)

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
@@ -708,6 +708,7 @@ public abstract class JobStatistics implements Serializable {
       queryStatisticsPb.setBillingTier(billingTier);
       queryStatisticsPb.setCacheHit(cacheHit);
       queryStatisticsPb.setDdlOperationPerformed(ddlOperationPerformed);
+      queryStatisticsPb.setEstimatedBytesProcessed(estimatedBytesProcessed);
       queryStatisticsPb.setTotalBytesBilled(totalBytesBilled);
       queryStatisticsPb.setTotalBytesProcessed(totalBytesProcessed);
       queryStatisticsPb.setTotalPartitionsProcessed(totalPartitionsProcessed);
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
```diff
@@ -432,6 +510,47 @@ public Boolean getCacheHit() {
...
+     * INSERT
+     * UPDATE
+     * DELETE
+     * CREATE_TABLE
+     * CREATE_TABLE_AS_SELECT
+     * DROP_TABLE
+     * CREATE_VIEW
+     * DROP_VIEW
+     */
+    public String getStatementType() { return statementType; }
```
tswast: @pongad Should we make an enum for statement type? I've noticed enums for other things in the Java client.

pongad: I think enums are appropriate here.

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/JobStatistics.java
```diff
@@ -320,8 +320,16 @@ static LoadStatistics fromPb(com.google.api.services.bigquery.model.JobStatistic
 
     private final Integer billingTier;
     private final Boolean cacheHit;
+    private final String ddlOperationPerformed;
+    private final TableId ddlTargetTable;
+    private final Long estimatedBytesProcessed;
+    private final Long numDmlAffectedRows;
+    private final List<TableId> referencedTables;
```
pongad: I just realized: Do we want to make these immutable?

tswast: Since this is job stats, which are read-only properties, I think immutable makes sense.

PR #3274, opened 2018-05-16
Adjust documentation creation to new repo structure

vam-google:
1) Add `utilities/stage_sites.py` (probably temporary solution).
2) Remove `utilities/create_site.sh` and `utilities/stage_release.sh` scripts. Other `.sh` releasing scripts will be removed soon. `RELEASING.md` was not update to reflect the changes. The readme will be updated after first successfull release with new structure (the only way to have an accurate releasing reame is to make the release first and record the steps).
3) Remove `.settings` folder. It is an eclipse-specific folder which hasn't been updated for 2 years. Nobody uses eclipse in our team, plus we should keep repo IDE-independent.
4) Move documentaiton look & feel files (.css, .html and .js files) to proper location (to `google-cloud-clients` from root).
5) Remove `google-cloud` maven dependency sample from landing page (since `google-cloud` metapackage was removed)
5) Several minor documentaiton-related fixes.

### DIFF
--- a/google-cloud-clients/src/site/resources/index.html
+++ b/google-cloud-clients/src/site/resources/index.html
@@ -77,12 +77,6 @@
                         <p>Google Cloud Client Library for Java - an idiomatic, intuitive, and natural way for Java developers to integrate with Google Cloud Platform services, like Cloud Datastore and Cloud Storage.</p>
                     </div><!-- end of .col.col-left -->
                     <div class="col col-right">
-                        <h2>Quickstart with Maven: add google-cloud to your pom.xml</h2>
-                        <pre>&lt;dependency&gt;
-  &lt;groupId&gt;com.google.cloud&lt;/groupId&gt;
-  &lt;artifactId&gt;google-cloud&lt;/artifactId&gt;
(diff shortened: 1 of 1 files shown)


utilities/stage_release.sh
```diff
@@ -1,21 +0,0 @@
-#!/bin/bash
```
garrettjonesgoogle: What will we need to run instead of `stage_release.sh`?

vam-google: `stage_sites.py` and then `mvn deploy` directly or a one-line script (I haven't decided yet). Those deployment scripts (`deploy.sh`, `stage_release.sh`, `create_site.sh`, `finalize_release.sh`) are doing mostly nothing or things which are obsolete now (like figuring global version for site or deploying snapshot  to sonatype (which we have never done and I belive will never do, because this is not how we "release" snapshots for people who may be interested in them)). 

I'm also thinking about just putting everythin in one script. In general I wan tot make the whole release a one-click thing, but need to handle the version bump thing first, that is why not doing it now. I'll specify all that in RELEASING.md once release is done.

vam-google:
@garrettjonesgoogle PTAL

PR #3281, opened 2018-05-16
PubSub: Update region tags to standard

alixhami:


alixhami:
@pongad Yes, I am submitting PRs to all the languages to standardize the region tags, and the do not merge label will be on all the PRs until they are all ready and I can update the docs site at the same time.

pongad:
OK, LGTM but we're cutting a release now so let's hold off a little.

PR #3282, opened 2018-05-17
Regenerate gapic clients, add IoT and Web Security Scanner clients

vam-google:
Also remove beta packages for dlp, as they are removed from googleapis. IoT packages were moved from `cloudiot` to just `iot`.

### DIFF
--- a/google-cloud-clients/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/ImageAnnotatorClient.java
+++ b/google-cloud-clients/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/ImageAnnotatorClient.java
@@ -157,6 +157,8 @@ public class ImageAnnotatorClient implements BackgroundResource {
    * Returns the OperationsClient that can be used to query the status of a long-running operation
    * returned by another API method call.
    */
+  @BetaApi(
+      "The surface for long-running operations is not stable yet and may change in the future.")
   public final OperationsClient getOperationsClient() {
     return operationsClient;
   }
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClient.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClient.java
@@ -203,6 +203,8 @@ public class BigtableInstanceAdminClient implements BackgroundResource {
    * Returns the OperationsClient that can be used to query the status of a long-running operation
    * returned by another API method call.
    */
+  @BetaApi(
+      "The surface for long-running operations is not stable yet and may change in the future.")
   public final OperationsClient getOperationsClient() {
     return operationsClient;
   }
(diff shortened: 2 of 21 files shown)


google-cloud-clients/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/ImageAnnotatorClient.java
```diff
@@ -217,6 +232,117 @@ private final BatchAnnotateImagesResponse batchAnnotateImages(
...
+   * try (ImageAnnotatorClient imageAnnotatorClient = ImageAnnotatorClient.create()) {
+   *   List&lt;AsyncAnnotateFileRequest&gt; requests = new ArrayList&lt;&gt;();
+   *   AsyncBatchAnnotateFilesResponse response = imageAnnotatorClient.asyncBatchAnnotateFilesAsync(requests).get();
+   * }
+   * </code></pre>
+   *
+   * @param requests Individual async file annotation requests for this batch.
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   */
+  public final OperationFuture<AsyncBatchAnnotateFilesResponse, OperationMetadata>
```
garrettjonesgoogle: Well crap I was wondering when this might happen. We have a GA client which formerly only used stable features from GAX, but now the API has added an API method which uses LRO, which isn't stable yet. We need to mark this `@BetaApi` for the moment - I believe we have a visibility field in GAPIC config that can do this.

google-cloud-clients/google-cloud-texttospeech/README.md
```diff
@@ -76,7 +76,7 @@ Versioning
 
 This library follows [Semantic Versioning](http://semver.org/).
 
-It is currently in major version zero (``0.y.z``), which means that anything may change at any time and the public API should not be considered stable.
+It is currently in major version zero (``0.y.z``), which means that anything may change at any time and the public API should not be considered stable.d
```
garrettjonesgoogle: undo this change

google-cloud-clients/google-cloud-dlp/pom.xml
```diff
@@ -15,7 +15,7 @@
     <version>0.47.1-alpha-SNAPSHOT</version><!-- {x-version-update:google-cloud-clients:current} -->
   </parent>
   <properties>
-    <site.installationModule>google-cloud-dlp-v2beta1</site.installationModule>
+    <site.installationModule>google-cloud-dlp-v2</site.installationModule>
```
garrettjonesgoogle: wait this doesn't seem right - shouldn't this match the artifact? (`google-cloud-dlp`)

google-cloud-clients/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/ImageAnnotatorClient.java
vam-google: Added `@BetaApi` annotation here manually, and posted [gapic-generator PR](https://github.com/googleapis/gapic-generator/pull/2004) to add it on generation phase for all long-running operations in java.

vam-google:
@garrettjonesgoogle PTAL

vam-google:
@garrettjonesgoogle PTAL

PR #3285, opened 2018-05-17
Update pubsub sample links

stetra:
The Pub/Sub code sample links in the README were broken, this updates the links to point to their new location.


pongad:
Thank you for the fix!

PR #3297, opened 2018-05-22
[Storage] Fix broken ITStorageTests

frankyn:
Added separate test bucket to the testListBucketDefaultKmsKeyName test as it was causing testBatchRequest to fail. testBatchRequest will fail if defaultKmsKeyName is set on the main test bucket which encrypts new objects with a KMS key on insert, because PATCH requests do not respond with crc32c and md5hash and causes a mismatch with object resource returned from a GET request which does have these values.


pongad:
LGTM

PR #3301, opened 2018-05-22
Logging: set Trace in trace instead of label

andreamlin:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3280.

### DIFF
--- a/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java
+++ b/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java
@@ -35,7 +35,7 @@ public class TraceLoggingEnhancer implements LoggingEnhancer {
   /**
    * Set the Trace ID associated with any logging done by the current thread.
    *
-   * @param id The traceID
+   * @param id The traceID, in the form projects/[PROJECT_ID]/traces/[TRACE_ID]
    */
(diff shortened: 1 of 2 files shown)


andreamlin:
@pongad @joaoandremartins 
PTAL

google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java
```diff
@@ -35,7 +35,7 @@ public TraceLoggingEnhancer(String prefix) {
   /**
    * Set the Trace ID associated with any logging done by the current thread.
    *
-   * @param id The traceID
+   * @param id The traceID, in the form projects/[PROJECT_ID]/traces/[TRACE_ID]
```
joaoandremartins: Have you considered also getting the project ID as an argument and building the `projects/[PROJECT_ID]/traces/[TRACE_ID]` format from there?

andreamlin: The convention for that in this repo would be to create a `TraceName` class that implements `ResourceName` and has a Builder pattern (or static constructor) and a formatting `toString()` method.

joaoandremartins: OK, LGTM then

PR #3307, opened 2018-05-24
make MetadataConfig.getAttribute() public

urandom2:
The Cloud libraries in other languages support calling `getAttribute()` like functions directly, so that customised calls can be simplified. It would be very helpful if this was allowed in Java as well.


arnottcr:
`I signed it!`

pongad:
LGTM
@garrettjonesgoogle Do you have a strong feeling about this?

garrettjonesgoogle:
This seems fine to me.

PR #3313, opened 2018-05-25
Adding Timestamp.toDate().

schmidt-sebastian:
This will be used by consumers of the Firestore API.

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
+++ b/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
@@ -95,18 +95,6 @@ public final class Timestamp implements Comparable<Timestamp>, Serializable {
     return ofTimeMicroseconds(TimeUnit.MILLISECONDS.toMicros(date.getTime()));
   }
 
-  /**
-   * Returns a new {@code java.util.Date} corresponding to this {@code timestamp}. Any
-   * sub-millisecond precision will be stripped.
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -95,6 +95,18 @@ public static Timestamp of(Date date) {
     return ofTimeMicroseconds(TimeUnit.MILLISECONDS.toMicros(date.getTime()));
   }
 
+  /**
+   * Returns a new {@code java.util.Date} corresponding to this {@code timestamp}. Any
+   * sub-millisecond precision will be stripped.
+   *
+   * @return An approximate {@code java.util.Date} representation of this {@code timestamp}.
+   */
+  public Date toDate() {
```
garrettjonesgoogle: You kind of stuck this in the middle of the section of static methods; could you move it down, maybe near `toSqlTimestamp`?

schmidt-sebastian: Sure. Done.

This was meant to be along with its friend, `Timestamp.of(Date)`.

garrettjonesgoogle: Makes sense - unfortunately the existing pattern uses a different rule for the highest level of grouping.

PR #3317, opened 2018-05-29
Return Dates as com.google.cloud.Timestamps

schmidt-sebastian:
This PR ports the changes from cr/191495680 and enables Firestore to return com.google.cloud.Timestamps with nanosecond precision.

The PR also removes Instant from the Firestore codebase, and instead returns and uses com.google.cloud.Timestamp everwhere.

About half the lines in this PR are from updating `CustomClassMapper` to the version in Google3. Review at your own digression (it's mostly copy&paste).


google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -243,6 +245,25 @@ public void emptyQuery() throws Exception {
...
   }
 
+  @Test
+  public void queryForServerTimestamps() throws Exception {
+    DocumentReference documentReference = addDocument("serverTime", FieldValue.serverTimestamp());
+    DocumentSnapshot documentSnapshot = documentReference.get().get();
+    Timestamp serverTime = documentSnapshot.getTimestamp("serverTime");
+
+    if (TimeUnit.NANOSECONDS.toMicros(serverTime.getNanos()) > 0) {
+      // If serverTime has a microsecond component, this query produces no results as `getDate()`
```
var-const: Hmm, if `serverTime.getNanos() == 1 * 1000 * 1000` (i.e., it's one millisecond), wouldn't the check above succeed? OTOH, it seems like the query should still succeed despite the truncation. I might misunderstand the code here, but it seems like the check above should be along the lines of `if (toMicros() % 1000 > 0`.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -684,7 +697,17 @@ public boolean equals(Object o) {
...
+                        com.google.protobuf.Timestamp.newBuilder()
+                            .setSeconds(479978400)
+                            .setNanos(123000000))
+                    .build())
+            .put(
+                "timestampValue",
+                Value.newBuilder()
+                    .setTimestampValue(
+                        com.google.protobuf.Timestamp.newBuilder()
+                            .setSeconds(479978400)
```
var-const: The value of seconds is the same as for the server timestamp, but the value of nanoseconds is different. Just checking that it's deliberate.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
```diff
@@ -875,4 +969,42 @@ private static String serializedName(String methodName) {
...
+   * error messages.
+   */
+  static class ErrorPath {
+    private final int length;
+    private final ErrorPath parent;
+    private final String name;
+
+    static final ErrorPath EMPTY = new ErrorPath(null, null, 0);
+
+    ErrorPath(ErrorPath parent, String name, int length) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Sorry, I am treating you the same as the bot and will keep this as is to match the Google3 code.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
```diff
@@ -51,17 +49,17 @@
   private final FirestoreImpl firestore;
   private final DocumentReference docRef;
   @Nullable private final Map<String, Value> fields;
-  @Nullable private Instant readTime;
-  @Nullable private Instant updateTime;
-  @Nullable private Instant createTime;
+  @Nullable private Timestamp readTime;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
```diff
@@ -51,17 +49,17 @@
...
+  @Nullable private Timestamp updateTime;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
```diff
@@ -287,7 +281,17 @@ public Object get(@Nonnull FieldPath fieldPath) {
       return null;
     }
 
-    return decodeValue(value);
+    Object decodedValue = decodeValue(value);
+    return convertToDateIfNecessary(decodedValue);
+  }
+
+  private Object convertToDateIfNecessary(Object decodedValue) {
```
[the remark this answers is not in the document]
schmidt-sebastian: This is cleaner that returning in the nested if.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
```diff
@@ -287,7 +281,17 @@ public Object get(@Nonnull FieldPath fieldPath) {
...
+    if (decodedValue instanceof Timestamp) {
+      if (!this.firestore.areTimestampsInSnapshotsEnabled()) {
```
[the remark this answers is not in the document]
schmidt-sebastian: This is deliberate.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Precondition.java
```diff
@@ -28,9 +27,9 @@
   public static final Precondition NONE = new Precondition(null, null);
 
   private Boolean exists;
-  private Instant updateTime;
+  private Timestamp updateTime;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
schmidt-sebastian: That is deliberate and uses the lowest precision available for each type (note that these constant match the DATE and TIMESTAMP constants in this class). I added comments.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
schmidt-sebastian: You are correct that it was broken. With your fix, it becomes apparent that it is really broken - server timestamps only have millisecond precision.

I have removed the test and replaced it with "updateMicrosecondTimestamp", which verifies that microseconds don't get truncated.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -200,6 +200,19 @@ public void serverTimestamp() throws Exception {
     assertTrue(documentSnapshot.getDate("time").getTime() > 0);
   }
 
+  @Test
+  public void updateMicrosecondTimestamp() throws Exception {
+    DocumentReference documentReference = addDocument("time", Timestamp.ofTimeSecondsAndNanos(0, 123000));
+    DocumentSnapshot documentSnapshot = documentReference.get().get();
+
+    Timestamp timestamp = documentSnapshot.getTimestamp("time");
+    documentReference.update("time", timestamp);
```
var-const: I might be missing something, but shouldn't this update the timestamp to a different value than the original?

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -200,6 +200,19 @@ public void serverTimestamp() throws Exception {
...
+
+    documentSnapshot = documentReference.get().get();
+    timestamp = documentSnapshot.getTimestamp("time");
+    assertEquals(123000, timestamp.getNanos());
```
var-const: Optional: to test truncation, could you use a timestamp with nanosecond precision and make sure the nanosecond part is truncated when it's added to the database?

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -245,25 +258,6 @@ public void defaultQuery() throws Exception {
...
-
-    if (TimeUnit.NANOSECONDS.toMicros(serverTime.getNanos()) > 0) {
-      // If serverTime has a microsecond component, this query produces no results as `getDate()`
-      // truncates to milliseconds.
-      Query query = randomColl.whereEqualTo("foo", documentSnapshot.getDate("serverTime"));
-      QuerySnapshot querySnapshot = query.get().get();
-      assertEquals(0, querySnapshot.size());
-    }
-
-    Query query = randomColl.whereEqualTo("serverTime", serverTime);
```
var-const: Hmm, the new test doesn't involve `Query`, so it's not equivalent. This seemed like a valuable thing to check, so perhaps the new test should also have it?

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
schmidt-sebastian: I renamed the test to `timestamptDoesntGetTruncatedDuringUpdate`.  This test doesn't use server timestamps anymore, since those only have millisecond precision. Instead, the same value is written, read, and re-written.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
schmidt-sebastian: I really only want to test non-truncation. I hope the server stops truncating altogether at some point, but for now, I can only test the non-truncation using microseconds.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
schmidt-sebastian: Sounds good. I added a new test case (`queryWithMicrosecondPrecision`).

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -83,6 +85,34 @@
...
+              + "With this change, timestamps stored in Cloud Firestore will be read back as "
+              + "com.google.cloud.Timestamp objects instead of as system java.util.Date "
+              + "objects. So you will also need to update code expecting a java.util.Date to "
+              + "instead expect a Timestamp. For example:\n"
+              + "\n"
+              + "// Old:\n"
+              + "java.util.Date date = snapshot.getDate(\"created_at\");\n"
+              + "// New:\n"
+              + "Timestamp timestamp = snapshot.getTimestamp(\"created_at\");\n"
+              + "java.util.Date date = timestamp.toDate();\n"
```
pongad: IIUC the reason we nee this setting is that there are methods that "formally" return `Object` and this will change whether the object is actually a `Date` or `Timestamp` right? I think it's worth giving an example of such a method.

schmidt-sebastian: Yeah, that's a good point. I updated it to use `DocumentSnapshot.get()`.

pongad:
@garrettjonesgoogle This has surface changes :)

PR #3320, opened 2018-05-29
Exposing TransportChannelProvider & CredentialsProvider

schmidt-sebastian:
Disclaimer: I am still wrapping my head around the initialization code for Veneer clients, so please shout if this doesn't make sense.

This PR is mainly intended to allow users to specify the size of the GRPC channel pool, but it exposes a way to set all options consumed in the venner-generated FirestoreSettings. AFAIK, there was no way to overwrite the TransportChannelProvider or the CredentialsProvider.

Code snippet that sets the max pool size:

```
Builder options =  FirestoreOptions.newBuilder().build().toBuilder();
options.setChannelProvider(
   FirestoreOptions.getDefaultTransportChannelProviderBuilder().setPoolSize(2).build());
firestore = options.build().getService();
```

Without this setting, Firestore is limited to 100 concurrent listeners.


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -144,14 +220,45 @@ public FirestoreRpcFactory getDefaultRpcFactory() {
...
+    }
+  }
+
+  @Nonnull
+  public static GrpcTransportOptions.Builder getDefaultTransportOptionsBuilder() {
+    return GrpcTransportOptions.newBuilder();
   }
 
   @Nonnull
-  public static GrpcTransportOptions getDefaultGrpcTransportOptions() {
```
schmidt-sebastian: FYI: This is a breaking API change if anyone used these APIs.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -82,53 +88,123 @@ protected boolean projectIdRequired() {
...
+    /**
+     * Sets the {@link CredentialsProvider} to use with this Firestore client.
+     *
+     * @param credentialsProvider A GoogleCredentialsProvider object that defines the credential
+     * provider for this client.
+     */
+    @Nonnull
+    public Builder setCredentialsProvider(@Nonnull CredentialsProvider credentialsProvider) {
+      if (!(credentialsProvider instanceof GoogleCredentialsProvider)) {
+        throw new IllegalArgumentException(
```
pongad: Out of curiosity, why do we have this restriction?

schmidt-sebastian: I looked through the callers and it seems like it is not needed.

pongad:
@garrettjonesgoogle Do you want to take a look?

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -145,13 +217,44 @@ public FirestoreRpcFactory getDefaultRpcFactory() {
     @Nonnull
     @Override
     public TransportOptions getDefaultTransportOptions() {
-      return getDefaultGrpcTransportOptions();
+      return TRANSPORT_OPTIONS;
+    }
+
+    @Nonnull
+    TransportChannelProvider getTransportChannelProvider() {
```
garrettjonesgoogle: `getDefaultTransportChannelProvider`? (or, the previous one should be `getTransportOptions` - which could make sense since this is already in a `FirestoreDefaults` class)

schmidt-sebastian: The other methods are inherited, so I changed the name to use "Default" everywhere.

PR #3322, opened 2018-05-30
fix storage auth example in README

### LINKED ISSUE #3314: GoogleCredentials Deprecated but is still used as an example for authentication
Hi Admin,

GoogleCredentials has already been deprecated but is still being used as an example in the Authentication using OAuth2. Would like to use the more updated version. Hope there will be some update. Thanks! :)

If you already have an OAuth2 access token, you can use it to authenticate (notice that in this case, the access token will not be automatically refreshed):
```
Storage storage = StorageOptions.newBuilder()
    .setCredentials(new GoogleCredentials(new AccessToken(accessToken, expirationTime)))
    .build()
    .getService();
```

yihanzhen:
To fix #3314

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ Storage storage = StorageOptions.getDefaultInstance().getService();
   case, the access token will not be automatically refreshed):
   ```java
   Storage storage = StorageOptions.newBuilder()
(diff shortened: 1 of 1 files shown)


hzyi-google:
PTAL

PR #3324, opened 2018-05-30
Add port to storage upload url

### LINKED ISSUE #3255: File upload connections are over fixed ports and path
If one uses `StorageOptions.Builder#setHost` with a value such as `http://proxy.url.local:8888/somepath`, builds the storage client using this `StorageOptions`, *most* connections generated by the client use these values (scheme, host, port, path), eg when listing and deleting blobs, but the file upload connection does not:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/52b727aef88ae76984aa3c02b4d7067e198d34b7/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L720

I think all connections should follow the `host setting` as a hostname or as an endpoint setting.

yihanzhen:
To fix #3255


hzyi-google:
PTAL

PR #3327, opened 2018-05-31
Add TextToSpeech v1 pom.xml and java files

andreamlin:
TTS v1 was recently introduced into googleapis/googleapis.

This takes the place of the weekly batch refresh.


andreamlin:
PTAL

andreamlin:
@garrettjonesgoogle TTS v1beta1 already exists, and there is a README.md file there in google-cloud-texttospeech. Should there be another README.md for just this new version?

I did update the google-cloud-texttospeech/README.md to point to the new v1 doc instead of v1beta1.

andreamlin:
PTAL

neozwu:
@andreamlin  Not sure whether this line needs updating ? https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/pom.xml#L754

andreamlin:
@neozwu good call, just added that in.

PR #3339, opened 2018-06-04
add newBuilder() to logging.SourceLocation

### LINKED ISSUE #3331: Logging - no public API to create SourceLocation.Builder
It appears there is no public API to create a new `com.google.cloud.logging.SourceLocation.Builder` object, thus I am unable to append this metadata to the log entries. 

Expected API would be to have a public `com.google.cloud.logging.SourceLocation.newBuilder()` function, but this is not there.

yihanzhen:
To fix #3331

### DIFF
--- a/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/SourceLocation.java
+++ b/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/SourceLocation.java
@@ -37,10 +37,6 @@ public final class SourceLocation implements Serializable {
     return new Builder();
   }
 
-  public static Builder newBuilder(SourceLocation sourceLocation) {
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/SourceLocation.java
```diff
@@ -33,6 +33,14 @@
   private final Long line;
   private final String function;
 
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static Builder newBuilder(SourceLocation sourceLocation) {
```
garrettjonesgoogle: you should provide a non-static method toBuilder() instead that someone can call on `SourceLocation`.

hzyi-google:
PTAL

PR #3340, opened 2018-06-04
make storage batch request honor options.getHost()

### LINKED ISSUE #3254: Batched Requests don't honor the host setting
Batch endpoint always hits `www.googleapis.com` irrespective of the `host` setting on the `StorageOptions` that built the storage client.
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/52b727aef88ae76984aa3c02b4d7067e198d34b7/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L191

yihanzhen:
To fix #3254


hzyi-google:
@andreamlin By default yes, but users can call `ServiceOptions.Builder.setHost()` to change the default setting and storage wouldn't have been able to honor that :)

pongad:
LGTM

PR #3348, opened 2018-06-05
Fix ITComputeTest

andreamlin:
`@Ignore` failing Compute integration tests and close https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3312


andreamlin:
PTAL

PR #3354, opened 2018-06-06
regenerate clients

pongad:
@andreamlin Do you know if we updated proto dependency or something? A LOT of proto-generated files have the same changes.


garrettjonesgoogle:
This looks like an accidental reversion in a previous run: https://github.com/pongad/google-cloud-java/commit/a02ff76245067fb508d2bb10528bf25d03040c52#diff-018ee9c8f39b25865c8c1fc87966402d

@andreamlin maybe you generated with an old protoc somehow?

andreamlin:
Oh that might have been the case. My local version of protoc is 3.4.0; though I used artmanflow which I think uses the Dockerfile that installs protoc 3.5.1? Anyhow if this was generated with protoc 3.5.1 that should be correct and the last batch generation I did was using an older version of protoc.

pongad:
This is strange. The field has been around since September 2017, and we should be using the same docker image...

pongad:
@garrettjonesgoogle @andreamlin Does this look good to you? If so I'll merge and start the actual release.

PR #3387, opened 2018-06-15
make DatastoreBatchWriter public

danielnorberg:
It would be useful for `DatastoreBatchWriter` to be public for instrumentation purposes. Its sibling interfaces `DatastoreWriter`, `DatastoreReaderWriter`, `DatastoreReader` etc are all public so I assume that it not being public is not intentional.


danielnorberg:
I signed it!

pongad:
@wsh Could you take a look?

wsh:
There are a couple of adjacent weirdnesses worth pointing out: [BaseDatastoreBatchWriter](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/675ace7601e2c09474ce986d5d2223af5cb15f28/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/BaseDatastoreBatchWriter.java#L36) is public but it's an abstract class whose sole constructor has no access modifier, so in practice I think that means it can [only be overridden by another class in the same package](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html).

PR #3391, opened 2018-06-18
BigQuery: Add ORC format support for load jobs, missing bigtable support.

shollyman:
Adds ORC support for BigQuery load jobs.

Also takes the opportunity to plumb in the more extensive (and missing) bigtable options, and clarify which formats are available for federation vs explicit load.

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigtableOptionsTest.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigtableOptionsTest.java
@@ -16,8 +16,6 @@
 package com.google.cloud.bigquery;
 
 import static com.google.common.truth.Truth.assertThat;
-import static junit.framework.TestCase.assertEquals;
-import static junit.framework.TestCase.assertTrue;
 
 import com.google.common.collect.ImmutableList;
 import org.junit.Test;


google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigtableColumn.java
```diff
@@ -0,0 +1,203 @@
...
+     * Qualifier of the column.
+     *
+     * Columns in the parent column family that has this exact qualifier are exposed
+     * as . field. If the qualifier is valid UTF-8 string, it can be specified in the
+     * qualifier_string field. Otherwise, a base-64 encoded value must be set to qualifier_encoded. The column field name is the same as the
+     * column qualifier. However, if the qualifier is not a valid BigQuery field
+     * identifier, a valid identifier must be provided as field_name.
+     */
+    Builder setQualifierEncoded(String qualifierEncoded) {
+      this.qualifierEncoded = qualifierEncoded;
```
tswast: Could you have used AutoValue for this class? It's looking pretty boilerplatey to me.

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/FormatOptions.java
```diff
@@ -26,15 +26,25 @@
 /**
  * Base class for Google BigQuery format options. These class define the format of external data
  * used by BigQuery, for either federated tables or load jobs.
+ *
+ * Load jobs support the following formats:
```
tswast: Thanks. Since separate classes would be a breaking change (that's probably not worth it), I think documentation is the best way.

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigtableColumn.java
shollyman: Swapped.  I should just embrace autovalue by default, instead of the "i may need something clever down the line".

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigtableColumn.java
```diff
@@ -0,0 +1,141 @@
...
+
+
+import com.google.common.base.Function;
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class BigtableColumn {
+
+  private static final long serialVersionUID = 1L;
```
pongad: To make this class serializable, we also need to `implements Serializable` right? This probably applies to other files as well.

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigtableColumn.java
```diff
@@ -0,0 +1,141 @@
...
+  @Nullable
+  public abstract String getEncoding();
+
+  @Nullable
+  public abstract String getType();
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+
```
pongad: The whitespaces do not conform to Java style (this repo doesn't conform in general, but we're trying to make it). Could you run google-java-format on this?

google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigtableOptionsTest.java
```diff
@@ -0,0 +1,114 @@
...
+      .setReadRowkeyAsString(true)
+      .setColumnFamilies(ImmutableList.of(TESTFAMILY))
+      .build();
+
+
+  @Test
+  public void testConstructors() {
+
+    // column
+    assertEquals(COL1.getQualifierEncoded(), "aaa");
```
pongad: We have been trying to move to `assertThat(COL1.getQualifierEncoded()).isEqualTo("aaa")` could you make the change here?

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigtableColumn.java
shollyman: Yanking these manually, but my IDE formatter doesn't seem to take action on these (or its not propagating all the formatting rules through from the plugin).

PR #3394, opened 2018-06-19
test ackDeadline

dmitry-s:
This test reproduces an issue with ackDeadline. It publishes 2 messages, then it pools them and nacks one of them. The other one is not being nacked or acked. Then it tries to pull both messages again, in a loop. 

The ackDeadline is set to 10 seconds, so I would expect to get both messages after that time. But the message that wasn't nacked or acked never makes it back. 

What is interesting, if I enable debugging and put a break point inside the loop, I am able to get both messages.

I verified that ackDeadline works as expected in gcloud cli tool.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/it/SubscriberTest.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/it/SubscriberTest.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2016 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


pongad:
I think this is mostly expected. The client library will automatically extend (modack) the messages that hasn't been acked or nacked. So not getting the message again is more like a sign that automatic deadline extension is working. By default, the client will auto-extend messages for an hour. You can change this with `Subscriber.Builder.setMaxAckExtensionPeriod`.

meltsufin:
@pongad What was the reasoning behind introducing such a long default ack extension? It doesn't even seem to be documented anywhere, whereas it radically changes the expectations of the users of how the ack deadline works.

meltsufin:
@garrettjonesgoogle Friendly ping.

pongad:
Sorry I missed this earlier. I believe this was recommended by Pubsub team a long time ago. @kir-titievsky do you happen to know? I'll make a PR to document this.

meltsufin:
@pongad The main issue with this behavior is the element of surprise to the users who don't expect the ack deadline to be auto-extended for an hour by default. In the least, this needs to be documented.

pongad:
The behavior is now documented. I'll close this for now, but please let us know if there's still confusion.

meltsufin:
@pongad Link(s)?

pongad:
It should be in the [client doc](https://googlecloudplatform.github.io/google-cloud-java/google-cloud-clients/apidocs/com/google/cloud/pubsub/v1/Subscriber.html). Search for "ack management".

PR #3409, opened 2018-06-28
Regenerate all clients; add cloud tasks (WIP)

andreamlin:


andreamlin:
google-common-protos has to be released first.

PR #3413, opened 2018-06-28
Set project id from credentials

chingor13:
When explicit credentials are provided, the client library should derive the project id from the same source as the credentials. In this case, the `ServiceAccountCredentials` are the only type of credentials that provide a project id.

An alternative approach would be to add `providesProjectId()` and `getProjectId()` to the surface of `GoogleCredentials` (in case future auth mechanisms are introduced that provide project ids) but we'd still do an `instanceof` check here for `GoogleCredentials`.

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -206,7 +206,8 @@ public abstract class ServiceOptions<ServiceT extends Service<OptionsT>,
     public B setCredentials(Credentials credentials) {
       this.credentials = checkNotNull(credentials);
       // set project id if available
-      if (credentials instanceof ServiceAccountCredentials) {
+      if (this.projectId == null &&
+          credentials instanceof ServiceAccountCredentials) {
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -204,6 +205,10 @@ public B setHost(String host) {
      */
     public B setCredentials(Credentials credentials) {
       this.credentials = checkNotNull(credentials);
+      // set project id if available
+      if (credentials instanceof ServiceAccountCredentials) {
```
hzyi-google: In this case `serviceOptions = ServiceOptions.newBuilder().setCredentials(...).setProjectId(...).build()` and `serviceOptions = ServiceOptions.newBuilder().setProjectId(...).setCredentials(...).build()` might create `serviceOptions` with different `projectId`s. Probably set the `projectId` here only when `this.projectId == null`?

chingor13: Fixed and added tests.

hzyi-google:
LGTM. @pongad Do you want to take a look?

PR #3415, opened 2018-06-28
BigQuery: Add Clustering support

shollyman:
DO NOT SUBMIT

Adds Clustering support to BigQuery.  Also includes ability to add TimePartitioning configuration as part of load and query jobs, as this functionality was not yet included in the veneer.


shollyman:
I didn't split TimePartitioning out because clustering will require a partitioned table at the outset, and the timeline is short enough at this point to just push it all at once.

danoscarmike:
@shollyman is this one nearly ready to submit?

PR #3418, opened 2018-07-02
Fix repo link in pom.xml

medb:


google-cloud-clients/google-cloud-storage/pom.xml
```diff
@@ -5,7 +5,7 @@
   <version>1.35.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-storage:current} -->
   <packaging>jar</packaging>
   <name>Google Cloud Storage</name>
-  <url>https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-storage</url>
+  <url>https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-clients/google-cloud-storage</url>
```
garrettjonesgoogle: It looks like the equivalent line is incorrect for all of our pom.xml files under google-cloud-clients. @hzyi-google maybe you should fix them all up in a new PR?

hzyi-google: Ack. I'll create an issue to track it.

PR #3421, opened 2018-07-02
Adding region tags to the README file

parkjam4:
Adding region tags to include in Java Cloud Client Library docs on cloud.google.com


parkjam4:
Thanks for the quick review!

PR #3422, opened 2018-07-02
Removing the region tags

parkjam4:
The text of the tags are displaying on the page.
Need to find a different way to grab this text snippet.


pongad:
I think you can comment in Markdown (sort of) by prepending `[//]: #`. Sorry I didn't catch this earlier.

parkjam4:
That wasn't working for me. It might be that syntax doesn't work with region tags.

PR #3434, opened 2018-07-09
Fix the error when calling Timestamp.of(Date date) when date is pre epoch

### LINKED ISSUE #3429: Timestamp.of(java.util.Date) fails for pre-epoch dates not truncated to seconds
com.google.cloud.Timestamp.of(Date) fails with IllegalArgumentException if the given date is before epoch and not truncated to seconds, in the concrete example, one millisecond before epoch was used.

google-cloud-core version: 1.35.0

Stacktrace fragment:
```
java.lang.IllegalArgumentException: timestamp out of range: 0, -1000000

	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:356)
	at com.google.cloud.Timestamp.ofTimeMicroseconds(Timestamp.java:84)
	at com.google.cloud.Timestamp.of(Timestamp.java:95)
```

Unit test (the variant with seconds and nanos succeeds, the variant with Date fails for the same instant in time):

```java
package com.outfit7.gcloud;

import java.time.Instant;
import java.util.Date;

import org.assertj.core.api.Assertions;
import org.junit.jupiter.api.Test;

import com.google.cloud.Timestamp;

class TimestampTest {

   private static final long EPOCH_MILLIS = 0L;
   private static final long EPOCH_MILLIS_MINUS_ONE = EPOCH_MILLIS - 1;

   @Test
   void ofTimeSecondsAndNanos_preEpoch() {

       // given (one second backwards, 999 millis forward, effectively one millisecond before epoch)
       long seconds = -1;
       int nanos = 999 * 1000000;

       // when
       Timestamp timestamp = Timestamp.ofTimeSecondsAndNanos(seconds, nanos);

       // then
       Assertions.assertThat(timestamp.toDate().getTime()).isEqualTo(EPOCH_MILLIS_MINUS_ONE);
   }

   @Test
   void ofJavaUtilDate_preEpoch() {

       // given
       Date epochMinusMillisecond = Date.from(Instant.ofEpochMilli(EPOCH_MILLIS_MINUS_ONE));

       // when
       Timestamp timestamp = Timestamp.of(epochMinusMillisecond);

       // then
       Assertions.assertThat(timestamp.toDate().getTime()).isEqualTo(EPOCH_MILLIS_MINUS_ONE);
   }
}
```

yihanzhen:
Fix #3429

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
+++ b/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
@@ -78,12 +78,11 @@ public final class Timestamp implements Comparable<Timestamp>, Serializable {
    * @throws IllegalArgumentException if the timestamp is outside the representable range
    */
   public static Timestamp ofTimeMicroseconds(long microseconds) {
-    long seconds = TimeUnit.MICROSECONDS.toSeconds(microseconds);
-    int nanos =
-        (int) TimeUnit.MICROSECONDS.toNanos(microseconds - TimeUnit.SECONDS.toMicros(seconds));
+    long seconds = microseconds / 1_000_000;
+    int nanos = (int)(microseconds % 1_000_000 * 1000);
     if (nanos < 0) {
       seconds--;
-      nanos += TimeUnit.SECONDS.toNanos(1);
+      nanos += 1_000_000_000;
     }
     checkArgument(
         Timestamps.isValid(seconds, nanos), "timestamp out of range: %s, %s", seconds, nanos);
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -81,6 +81,10 @@ public static Timestamp ofTimeMicroseconds(long microseconds) {
     long seconds = TimeUnit.MICROSECONDS.toSeconds(microseconds);
     int nanos =
         (int) TimeUnit.MICROSECONDS.toNanos(microseconds - TimeUnit.SECONDS.toMicros(seconds));
+    if (nanos < 0) {
```
garrettjonesgoogle: Maybe a while loop instead?

pongad: I think this function is much fancier than it has to be. Due to the way Java modulus works, I think
```java
long seconds = microseconds / 1_000_000;
long nanos = microseconds % 1_000_000 * 1000;
```
should just work.

hzyi-google: @pongad `nanos` might still be negative this way. We would still need to make it positive by taking 1 second from `seconds`.
@garrettjonesgoogle  `nanos` wouldn't be less than `1_000_000_000` since it only contains the fraction less than 1 second. So an `if` should be enough.

hzyi-google:
PTAL

PR #3441, opened 2018-07-11
Add GCE and discogapic to batch gen script

andreamlin:
This is assumes https://github.com/googleapis/artman/pull/457 will be merged in.

I've run this updated script on GCE and bigquerydatatransfer and bigtable to ensure that all gapics are created, the google-cloud-clients src dirs are updated, and the proto- and grpc- libraries fully replace any existing ones.

### DIFF
--- a/utilities/generate_api.py
+++ b/utilities/generate_api.py
@@ -44,8 +44,6 @@ def run_generate_api(config_path, artifact_type, noisy=False):
 
     :param config_path: (str) Path to directory containing artman config file.
     :param artifact_type: (str) artman target, e.g "java_gapic".
-    :param root_dir: (str) Path to the directory containing the API definitions.
-    :param has_proto_libs: (bool) if this target creates proto and grpc libraries
     :param noisy: (bool) if console output should be verbose.
 
     """


andreamlin:
PTAL

utilities/batch_generate_apis.py
```diff
@@ -14,24 +14,25 @@
 
 # Instructions:
 #
-# Check out the googleapis repo somewhere locally (e.g. from
-# https://github.com/googleapis/googleapis):
+# Check out the googleapis and discovery-artifact-manager repo somewhere locally
+# (e.g. from https://github.com/googleapis/googleapis):
 #
 # $ git checkout https://github.com/googleapis/googleapis.git
+# $ git checkout https://github.com/googleapis/discovery-artifact-manager
```
garrettjonesgoogle: You need the `.git` at the end

utilities/batch_generate_apis.py
```diff
@@ -73,13 +74,27 @@ def generate(artman_yaml):
     generate('google/cloud/websecurityscanner/artman_websecurityscanner_v1alpha.yaml')
 
 
+def run_discogapic_gen(discovery_repo):
+    def generate(artman_yaml):
+        # Run java_discogapic task. No proto or grpc libraries are generated.
+        generate_api.run_generate_api(os.path.join(discovery_repo, artman_yaml),
+                                      "java_discogapic",
+                                      root_dir=discovery_repo,
+                                      has_proto_libs=False)
```
garrettjonesgoogle: I tried to make the argument list for running generation as minimal as possible. To match what I did for gapic generation, you would have to:

1. Automatically figure out the root dir from the yaml file path. For java_gapic, I search for `/google/`.
2. Automatically figure out there are no proto libs. You can probably just figure this out based on the fact that you are using "java_discogapic". 

We would probably go even further and detect whether a generation is for gapic or discogapic based on whether the full path has `/googleapis` or `/discovery` in it.

utilities/batch_generate_apis.py
```diff
@@ -73,13 +74,27 @@ def generate(artman_yaml):
...
+
+    generate('gapic/google/compute/artman_compute.yaml')
+
+
 def main():
     # TODO Make the docker image the default, add --local option
     parser = argparse.ArgumentParser(description='Batch generate all APIs.')
     parser.add_argument('googleapis', help='The path to the googleapis repo')
+    parser.add_argument('discovery', help='The path to the discovery-artifact-manager repo')
```
garrettjonesgoogle: maybe call the arg `discovery_repo`

utilities/generate_api.py
```diff
@@ -36,18 +36,33 @@
...
+    else:
+        api_dir = config_path
 
     extra_options = []
     if noisy:
         extra_options = ['-v']
 
-    subprocess.check_call(['artman', '--config', api_dir, '--local', '--root-dir', root_dir] + extra_options + ['generate', 'java_gapic'])
+    print("api_dir: "+ api_dir)
+    print("root_dir: "+ root_dir)
```
garrettjonesgoogle: Are these print outs just for you to debug? If you want to keep them, only print them if `noisy` is true.

utilities/batch_generate_apis.py
andreamlin: Done.

utilities/batch_generate_apis.py
andreamlin: Ok, removed all extra params except for artifact_type.

I'm a little wary of logic that is tied to the existence of the api repo name somewhere in full path, and I'd rather the artifact_type be explicit.

utilities/batch_generate_apis.py
andreamlin: Done.

utilities/generate_api.py
andreamlin: Removed.

andreamlin:
Addressed comments and cleaned up. PTAL

utilities/generate_api.py
```diff
@@ -35,34 +36,34 @@
...
-def run_generate_api(config_path, artifact_type, root_dir=None, has_proto_libs=True, noisy=False):
+def run_generate_api(config_path, artifact_type, noisy=False):
     """ Generate an API client library.
 
     :param config_path: (str) Path to directory containing artman config file.
     :param artifact_type: (str) artman target, e.g "java_gapic".
     :param root_dir: (str) Path to the directory containing the API definitions.
-    :param has_proto_libs: (bool) if the generated proto and grpc libraries
-              should replace the existing ones
+    :param has_proto_libs: (bool) if this target creates proto and grpc libraries
```
garrettjonesgoogle: Remove `root_dir` and `has_proto_libs`

PR #3444, opened 2018-07-11
Regenerating proto/grpc/gapic code

garrettjonesgoogle:


garrettjonesgoogle:
Tests passed on a prior iteration (before my accidental push & reset), so merging now

PR #3454, opened 2018-07-11
Fix one test to work with the other unit tests

jean-philippe-martin:
`testListFilesInRootDirectory` in `ITGcsNio.java` didn't specify explicit storage options, but it should have. Using the defaults as it did caused an issue when also running the unit tests, because the unit tests change the defaults to be the in-memory storage.

Now I set the storage options explicitly and all works as expected.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/it/ITGcsNio.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/it/ITGcsNio.java
@@ -527,6 +527,10 @@ public class ITGcsNio {
         got.add(path);
       }
       assertThat(got).containsExactlyElementsIn(goodPaths);
+      // clean up
+      for (Path path : paths) {
+        Files.delete(path);
+      }
     }
   }
 
@@ -552,10 +556,12 @@ public class ITGcsNio {
 
   @Test
   public void testListFilesInRootDirectory() throws IOException {
+    // We must explicitly set the storageOptions, because the unit tests
+    // set the fake storage as default but we want to access the real storage.
     CloudStorageFileSystem fs = CloudStorageFileSystem.forBucket(
       BUCKET, CloudStorageConfiguration.builder().permitEmptyPathComponents(true)
-      .build());
-    
+      .build(), storageOptions);
+
     // test absolute path
     Path rootPath = fs.getPath("");
     List<String> objectNames = new ArrayList<String>();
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/it/ITGcsNio.java
```diff
@@ -552,10 +556,12 @@ public void testDeleteRecursive() throws IOException {
 
   @Test
   public void testListFilesInRootDirectory() throws IOException {
+    // We must explicitly set the storageOptions, because the unit tests
+    // set the fake storage as default but we want to access the real storage.
```
pongad: If this needs to test against real storage, shouldn't it go into `ITStorageTest`?

jean-philippe-martin: This test case accesses real storage using the NIO interface, so it is correct for it to be with the others in `ITGcsNio` and not `ITStorageTest`.

pongad: Sorry I got confused. Since this is `ItGcsNio`, shouldn't other tests be using the real storage server anyway?

Pardon my ignorance, I'm still trying to page things back from my long term memory.

jean-philippe-martin: The other tests do too. They all specify `storageOptions` - this test was the only outlier, a situation that this pull request aims to resolve.

The unit tests mentioned in the comment are outside of ITGcsNio. But because they work by changing the default (something that's normally done only once per program), when running the full test suite the integration test must not rely on the default.

A different way to fix this would be to change all the unit tests to explicitly set storageOptions - however the unit tests check the various ways to create an NIO cloud provider (not all of which allow explicit options). A further alternative would be to reset the defaults when running the integration tests. However since the integration test really should be using storageOptions to set lenient retry settings (we want it to retry if GCS is transiently slow rather than erroneously reporting a failed test), I favor the solution in this PR: set storageOptions for all of the NIO integration tests.

pongad: You're right. I stand corrected.

PR #3460, opened 2018-07-12
Prefer provided credentials if they are present in FirestoreOptions

andrewparmet:
Addresses https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3458.

As of 0.52.0-beta, calls to `setCredentials()` on FirestoreOptions are effectively ignored. This is because `GrpcTransportOptions.setUpCredentialsProvider()` is no longer called in GrpcFirestoreRpc - instead the settings builder just pulls the credentials provider regardless of whether or not a fixed credential was supplied in the original options.

It used to be:
```java
settingsBuilder.setCredentialsProvider(
    GrpcTransportOptions.setUpCredentialsProvider(options));
```

Today it reads:
```java
settingsBuilder.setCredentialsProvider(options.getCredentialsProvider());
```

This was changed in #3320. ([Diff line](https://github.com/GoogleCloudPlatform/google-cloud-java/commit/9ab5280571a9c794cb54714b25eb194bc5ff2d99#diff-5fe9ae038f9c76819ca4ebcdddad2289L108))

The call can be restored in the same class with some conditional logic, or the FirestoreOptions can try to return the correct CredentialsProvider.

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
@@ -98,7 +98,11 @@ public final class FirestoreOptions extends ServiceOptions<Firestore, FirestoreO
   }
 
   public CredentialsProvider getCredentialsProvider() {
-    return credentialsProvider;
+    if (credentials != null) {
+      return GrpcTransportOptions.setUpCredentialsProvider(this);
+    } else {
+      return credentialsProvider;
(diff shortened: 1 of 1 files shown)


pongad:
@schmidt-sebastian Can you take a look?

schmidt-sebastian:
Thanks for sending this over! The problem you described exists in both the CredentialsProvider and the ChannelProvider. I have opened a PR to deal with both: https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3472

Let's leave this open until 3472 is approved.

PR #3462, opened 2018-07-12
[DO NOT MERGE] Removing doclint disabling for bigtable on Java 8

garrettjonesgoogle:
(Hopefully) Fixes https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3461

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -227,6 +227,9 @@ jobs:
     steps:
       - checkout
       - run:
+          name: print maven version
+          command: mvn --version
+      - run:
           <<: *anchor_build
       - run:
           name: Run mvn javadoc goal


andreamlin:
Has this been tested locally against a commit snapshot that we know has a failing javadoc target?

garrettjonesgoogle:
Here's the tricky thing: this will have no effect locally. The only effect it can have is on circle, where it should expose failures. I guess the only way to test is to push a commit in the PR which has a failure and make sure that it fails on Circle (where previously it passed). I'll wait for the present test run to finish before trying that out.

andreamlin:
Hmm, was javadoc_test supposed to fail?

garrettjonesgoogle:
Yes, it was supposed to fail and it didn't. Looks like this change didn't work.

JustinBeckwith:
👋 @garrettjonesgoogle what's the status on this one?

garrettjonesgoogle:
Abandoning

Issue #3466, opened 2018-07-12
Unable to set destinationDefaultAcl for copied objects


frankyn:
GCS client library doesn't support ACL properties when performing a copy/rewrite operation.

```java
Storage.CopyRequest copyRequest = Storage.CopyRequest.newBuilder()
         .setSource(sourceBucketName, sourcePath)
         .setTarget(BlobId.of(destinationBucketName, destinationPath),
                   BlobTargetOption.predefinedAcl(PredefinedAcl.PUBLIC_READ))
         .build();
         storage.copy(copyRequest).getResult();
```

Issue is predefinedAcl isn't being set in rewrite request.
Ref: [HttpStorageRpc.java#L797](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L797)

PR #3467, opened 2018-07-13
Fix Storage rewrite operation to support predefinedAcl on a copy

frankyn:
cc: @danoscarmike 

- [x] Write fix for Storage rewrite operation to support predefinedAcl on targetObject
- [x] Verify broken use case prior to applying fix. 
- [x] Write IT test to verify behavior. 
- [x] Code Review.

Note: [A unit test](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/CopyRequestTest.java#L48) for PredefinedAcl with a CopyRequest exist but a integration test didn't exercise the request.

Fixes: #3466


BrandonY:
Looks great, thanks for putting that together.

frankyn:
Thanks @pongad!

PR #3468, opened 2018-07-13
Add core NIO contributors to credits

jean-philippe-martin:
Add JP and Justine to the developers list in the pom.

Heads-up for @jart, @hzyi-google, @garrettjonesgoogle 

Thank you!


jean-philippe-martin:
Thank you for the quick review!

PR #3469, opened 2018-07-13
Avoid listing table data for destination of CREATE VIEW DDL queries.

tswast:
CREATE VIEW DDL queries set a "destination table" field, but that table
is actually a view. If you attempt to list rows on the view table, it
results in an error.

The fix is to avoid listing rows altogether if getQueryResults() says
that a query has completed but the number of rows in the result set is
undefined.

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/JobTest.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/JobTest.java
@@ -281,7 +281,6 @@ public class JobTest {
                 return Collections.emptyList();
               }
             };
-    TableResult result = new TableResult(Schema.of(), 0, emptyPage);
     QueryResponse completedQuery =
             QueryResponse.newBuilder()
                     .setCompleted(true)


tswast:
Looks like I need to update a couple unit tests.

google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/JobTest.java
```diff
@@ -224,6 +226,82 @@ public void testWaitFor() throws InterruptedException {
...
+              public Iterable<FieldValueList> iterateAll() {
+                return Collections.emptyList();
+              }
+
+              @Override
+              public Iterable<FieldValueList> getValues() {
+                return Collections.emptyList();
+              }
+            };
+    TableResult result = new TableResult(Schema.of(), 0, emptyPage);
```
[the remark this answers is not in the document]
tswast: Removed in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3469/commits/bffb0eb227863f94b9942f3618e8c33608113689

tswast:
Unit tests are now passing. Ready for review!

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Job.java
```diff
@@ -285,10 +290,24 @@ public TableResult getQueryResults(QueryResultsOption... options)
...
-      throw new JobException(getJobId(), response.getErrors());
+
+    // Get the job resource to determine if it has errored.
+    Job job = this;
+    if (job.getStatus() == null || job.getStatus().getState() != JobStatus.State.DONE) {
+      job = reload();
+    }
+    if (job.getStatus() != null && job.getStatus().getError() != null) {
+      throw new JobException(
+          getJobId(), ImmutableList.copyOf(job.getStatus().getExecutionErrors()));
```
pongad: I don't completely understand why we need to reload the job. Can we infer what we need from the response?

The reason I'm asking is that some folks has complained about high latency in the past.

tswast: No, as far as I can tell there is no way to determine that a job has failed from `getQueryResults()`. Previously this code looked for a missing schema, but that can happen for certain DDL and related queries now.

PR #3481, opened 2018-07-19
update opencensus

### LINKED ISSUE #3480: Upgrade opencensus
0.15.0 is current

elharo:
fix #3480


sduskis:
LGTM from my perspective, if that accounts for anything :)

Of note: There's a tangled set of dependencies between grpc, opencensus and the Stackdriver client used by OpenCensus's extensions (which also happens to be in google-cloud-java).

OpenCensus currently uses grpc 0.12.0.  It would be great if OpenCensus's grpc version were upgraded as well.

pongad:
LGTM, though we're in the middle of a release right now. Let's put this in the next release.

PR #3491, opened 2018-07-20
pubsub: document auto message extension

pongad:
Also tweak CircleCI doc test. It is sufficient to just generate the
docs; compressing is slow.


pongad:
Seems like a small enough change to just fold in (see motivation in the description). Would you rather have them separate?

garrettjonesgoogle:
I think separate is best because they don't seem connected.

pongad:
@garrettjonesgoogle PTAL

PR #3504, opened 2018-07-25
set userProject from static default in FileSystemProvider

### LINKED ISSUE #3475: Setting userProject from NIO FileSystemProvider
I'm happy to see that [requester-pays support was just added](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3221)!

afaict, I can't set the `userProject` in a `CloudStorageFileSystemProvider` that gets instantiated via the Java NIO `FileSystemProvider` machinery, which uses `Class.newInstance()` to invoke the nullary constructor:

```java
import java.net.URI;
import java.nio.file.Paths;
import java.nio.file.Files;
Files.newDirectoryStream(Paths.get(new URI("gs://<bucket>")))
```

throws:
```
com.google.cloud.storage.StorageException: Bucket is requester pays bucket but no user project provided.
  at com.google.cloud.storage.spi.v1.HttpStorageRpc.translate(HttpStorageRpc.java:220)
  at com.google.cloud.storage.spi.v1.HttpStorageRpc.list(HttpStorageRpc.java:347)
  at com.google.cloud.storage.StorageImpl$8.call(StorageImpl.java:299)
  …
```

Is there a way to set the user-project, other than as a ctor param?
- an environment variable would seem easiest
  - afaict there's no standard mechanism for doing that across GCS libraries in various languages
  - perhaps it's still ok to add one here
- a setter would help a bit
  - I'll be mainly using it from [hammerlab/path-utils](https://github.com/hammerlab/path-utils), where I (unfortunately) have code to [reach into the `installedProviders` and modify the providers for a given URI scheme](https://github.com/hammerlab/path-utils/blob/1.5.0/src/main/scala/org/hammerlab/paths/FileSystems.scala#L128-L145)
  - a setter would at least save me from a second reflection hack to reach

ryan-williams:
fixes #3475

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
@@ -172,7 +172,12 @@ public final class CloudStorageFileSystemProvider extends FileSystemProvider {
    * @see CloudStorageFileSystem#forBucket(String)
    */
   public CloudStorageFileSystemProvider() {
-    this("", futureStorageOptions);
+    this(
+        CloudStorageFileSystem
+            .getDefaultCloudStorageConfiguration()
+            .userProject(),
+        futureStorageOptions
+    );
   }
 
   /**


ryan-williams:
(I signed the CLA; thought I already had, weird)

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
```diff
@@ -172,7 +172,12 @@ public static void setDefaultCloudStorageConfiguration(@Nullable CloudStorageCon
    * @see CloudStorageFileSystem#forBucket(String)
    */
   public CloudStorageFileSystemProvider() {
-    this("", futureStorageOptions);
+    this(
+        CloudStorageFileSystem
+            .getDefaultCloudStorageConfiguration()
+            .userProject(),
```
hzyi-google: Thanks for contributing! I must admit I am not super familiar with NIO, so it would be best if you could clarify:
1. the default `CloudStorageConfiguration` sets `userProject` to null, would passing in the `userProject` of default `CloudStorageConfiguration` in help at all?
2. `CloudStorageConfiguration` also has a boolean value `useUserProjectOnlyForRequesterPaysBuckets`, and only when it's set to `true` would the `userProject` be used to access storage buckets. It seems your change would set the `userProject` here regardless of that flag, so would that behavior be preserved?

ryan-williams: > 1. the default `CloudStorageConfiguration` sets `userProject` to null, would passing in the userProject of default `CloudStorageConfiguration` in help at all?

do you mean that if a user doesn't change the default, now they'll get `null` instead of `""`, and does that matter? I just looked at its uses, and it seems everything is correctly handling `null`, which also seems more idiomatic than the previous `""`.

or, if your question is "how does this change anything, since we're passing through the default config's `null`?", the point is: if the user has changed the default `CloudStorageConfiguration`'s `userProject`, we should use it here.

Some misc context:

The NIO SPI machinery instantiates `FileSystemProvider`s via their nullary constructor (in the java standard lib), so the kind of static default we're using here is one way to let the user configure them. In normal usage, only one instance of a given `FileSystemProvider` is ever instantiated, and if there's more than 1, those others would be coming from user-code (not the SPI machinery), where it's easy to configure them using one of the other constructors here, if the static default is not what's desired.

Another option, that I proposed at first on #3475, would be to expose a `CloudStorageFileSystemProvider.setUserProject`, but I'm not even sure that unblocks us: NIO's default `FileSystemProvider` instances for each URI scheme are not exposed in a way that makes it easy to call that `setUserProject`; you'd have to call `FileSystemProvider.installedProviders`, find the one with the `gs` scheme, cast it to `CloudStorageFileSystemProvider`, then call the setter, and when I tried it from Scala there were multiple versions of `CloudStorageFileSystemProvider`, apparently loaded by different classloaders, so that I couldn't do the cast, and had to reflection-cheat 😱

> 2. `CloudStorageConfiguration` also has a boolean value `useUserProjectOnlyForRequesterPaysBuckets`, and only when it's set to `true` would the `userProject` be used to access storage buckets. It seems your change would set the `userProject` here regardless of that flag, so would that behavior be preserved?

I didn't verify that that flag is implemented correctly, but presumably it is, and the `userProject` value is not passed for non-requester-pays buckets when it is true 😀 it should be orthogonal to my concern here ([disabling passing `userProject` when it's set but not needed], vs [actually passing it when it is set and needed])

hzyi-google: Thanks for the response. The first question was referring to the latter. In this case LGTM

PR #3512, opened 2018-07-27
Bigtable: Decouple TableAdminClient from BigtableTableAdminSettings.

igorbernstein2:
This is in preparation for renaming the auto generated GAPIC clients &
settings to be prefixed with 'Base'. This will make it easier to
understand the layout of the code: the GAPIC generated client/settings
will be called BaseBigtableTableAdmin{Client,Settings}, while the
handwritten overlay will be called BigtableTableAdmin{Client,Settings}.


igorbernstein2:
oops, sorry about that. Should be fixed now

PR #3527, opened 2018-08-02
Move Translate snippets to test file, use new snippets.go for inclusion.

tswast:
Updates to samples which are included in the docs to meet our
cdpe-snippets-rubric (show imports, show client construction, show how
to use the resulting object). I will follow-up with a PR to
java-docs-samples to remove any redundant snippets from that repo.

Why move to a test file?

We are considering moving all snippets-style samples to test files to
avoid redundant work involved with keeping snippets separate from test
files. The important thing for snippets is that they are runnable and
copy-pastable.


google-cloud-clients/google-cloud-translate/src/main/java/com/google/cloud/translate/Translate.java
```diff
@@ -116,15 +116,35 @@ public static TranslateOption format(String format) {
    *
    * <p>Example of listing supported languages, localized according to
    * {@link TranslateOptions#getTargetLanguage()}:
-   * <pre> {@code
-   * List<Language> languages = translate.listSupportedLanguages();
-   * }</pre>
+   * <!--SNIPPET translate_list_codes-->
+   * <pre>{@code
+   *    // TODO(developer): Uncomment these lines.
+   *    // import com.google.cloud.translate.*;
```
nnegrey: Are we using `*` here to keep things shorter as this is more of an example as part of the library?

google-cloud-clients/google-cloud-translate/src/main/java/com/google/cloud/translate/Translate.java
```diff
@@ -116,15 +116,35 @@ public static TranslateOption format(String format) {
...
+   *    // Translate translate = TranslateOptions.getDefaultInstance().getService();
+   *
+   *    List<Language> languages = translate.listSupportedLanguages();
+   *
+   *    for (Language language : languages) {
+   *      System.out.printf("Name: %s, Code: %s\n", language.getName(), language.getCode());
+   *    }
+   *     }</pre>
```
nnegrey: Should this be indented back one?

google-cloud-clients/google-cloud-translate/src/main/java/com/google/cloud/translate/Translate.java
```diff
@@ -116,15 +116,35 @@ public static TranslateOption format(String format) {
...
+   * <!--SNIPPET translate_list_codes-->
+   * <pre>{@code
+   *    // TODO(developer): Uncomment these lines.
+   *    // import com.google.cloud.translate.*;
+   *    // Translate translate = TranslateOptions.getDefaultInstance().getService();
+   *
+   *    List<Language> languages = translate.listSupportedLanguages();
+   *
+   *    for (Language language : languages) {
+   *      System.out.printf("Name: %s, Code: %s\n", language.getName(), language.getCode());
```
nnegrey: I see the codacy bot is upset by printf below, maybe use `System.out.format`?

pongad:
LGTM but @nnegrey should +2

google-cloud-clients/google-cloud-translate/src/main/java/com/google/cloud/translate/Translate.java
tswast: This example appears in the JavaDoc for the library. `*` It is included here by the `snippets.go` utility.

google-cloud-clients/google-cloud-translate/src/main/java/com/google/cloud/translate/Translate.java
tswast: Yes. I've updated `snippets.go` to clean up the indentation.

google-cloud-clients/google-cloud-translate/src/main/java/com/google/cloud/translate/Translate.java
tswast: That warning is thinking that we left in a debug statement. It can be safely ignored for code samples. I think printf will be easier to understand than if we used a logger to hush codacy.

PR #3529, opened 2018-08-03
Add model table type for the new BigQuery ML models

gsabhnani:
Adding a new type=Model in TableDefinition. This type is used to identify BigQuery ML model tables in a dataset.

@tswast:  Tim, looking for early comments on this. I think location and num_bytes should be enough info to cache from PB in ModelTableDefinition. Also, I just updated the existing listTables test. Please let me know if its preferred to add a new test for listing model tables. Thanks.

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ModelTableDefinition.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ModelTableDefinition.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Google LLC
+ * Copyright 2018 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,12 +16,9 @@
 
 package com.google.cloud.bigquery;
 
+import com.google.api.core.BetaApi;
 import com.google.api.services.bigquery.model.Table;
 import com.google.auto.value.AutoValue;
-import com.google.common.base.MoreObjects;
-import java.io.Serializable;
-import java.math.BigInteger;
-import java.util.Objects;
 import javax.annotation.Nullable;
 
 /**
@@ -31,9 +28,9 @@ import javax.annotation.Nullable;
  * @see <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-create#models_in_bqml_name">BigQuery ML Model</a>
  */
 @AutoValue
+@BetaApi
 public abstract class ModelTableDefinition extends TableDefinition {
 
-  //TODO(gsabhnani): Update this id??
   private static final long serialVersionUID = 2113445776046717900L;
 
   @AutoValue.Builder


google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ModelTableDefinition.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2016 Google LLC
```
tswast: s/2016/2018/

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ModelTableDefinition.java
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * A Google BigQuery Model table definition. This definition is used to represent a BigQuery
+ * ML model.
+ *
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-create#models_in_bqml_name">BigQuery ML Model</a>
+ */
+@AutoValue
+public abstract class ModelTableDefinition extends TableDefinition {
+
+  //TODO(gsabhnani): Update this id??
```
tswast: I'm not sure where the original ID came from. I don't think it matters so long as it's consistent and we update it when we make a breaking change to the class.

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ModelTableDefinition.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * A Google BigQuery Model table definition. This definition is used to represent a BigQuery
+ * ML model.
+ *
+ * @see <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-create#models_in_bqml_name">BigQuery ML Model</a>
+ */
+@AutoValue
+public abstract class ModelTableDefinition extends TableDefinition {
```
tswast: Please also annotate with com.google.api.core.BetaApi since BQML is beta.

I recall that there will probably be a "models API" so this class may go away in the future.

gsabhnani: Done.

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ModelTableDefinition.java
gsabhnani: I copied this from StandardTableDefinition.java file. Seems like each class has a unique id and I am not sure if we need another one for this.

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/ModelTableDefinition.java
gsabhnani: Done.

tswast:
FYI: I was able to list the tables in a dataset containing a model with this change.

Code:
```
Page<Table> tables = bigquery.listTables("bqml_tutorial");
for (Table table : tables.iterateAll()) {
  System.out.println(table.getTableId());
}
```

Output:
```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.google.cloud.examples.bigquery.snippets.ITCloudSnippets2
{datasetId=bqml_tutorial, projectId=swast-scratch, tableId=sample_model}
{datasetId=bqml_tutorial, projectId=swast-scratch, tableId=sample_model_eval}
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.959 sec

Results :

Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

This code fails with `Format MODEL is not supported` on master.

PR #3547, opened 2018-08-09
Regenerate all clients

andreamlin:
Including discogapic compute

Used batch_generate.py script, and assumes https://github.com/googleapis/discovery-artifact-manager/pull/90 and https://github.com/googleapis/artman/pull/486 are checked in.


andreamlin:
PTAL

pongad:
Hmm, can you confirm that you're running protoc 3.6.0? A lot of these changes look like they're coming from different protoc version.

andreamlin:
@pongad hmm I thought I was using protoc3.6.0 but now am re-regenerating and it's reverting a lot of the proposed changes, so maybe I was using v3.5.1. batch generation is in progress...

andreamlin:
@igorbernstein2 Would you know about why there are compiling failures here? Why did `Bigtable.*Client` get renamed to `BaseBigtable.*Client`?

andreamlin:
Ok, used protoc v3.6.0 and bigtable is not a problem anymore! PTAL

PR #3552, opened 2018-08-10
generate-api: print component versions

igorbernstein2:
When generating an api, dump component versions. Example output:

```
Component versions:
artman 0.14.3
googleapis 2c07771e79489359ef48925612feb59f0cea8b27
google-cloud-java e76a9dba12842e36332b4482217af07113c17607( 4 files changed, 34 insertions(+), 5 deletions(-))
gapic_generator  21b522793f08e03ecd30db1298a05bb9af8ad7c1
```

This should ease debugging when unexpected things happen during a client regen.

### DIFF
--- a/utilities/generate_api.py
+++ b/utilities/generate_api.py
@@ -54,18 +54,23 @@ def get_git_repo_version(path):
     return ''.join([commit, suffix])
 
 
-def dump_versions(config_path):
+def dump_versions(googleapis=None, discovery_repo=None):
     print("Component versions:")
 
     print(subprocess.check_output(['artman', '--version'], stderr=subprocess.STDOUT).strip())
-
-    print("googleapis {}".format(get_git_repo_version(os.path.dirname(config_path))))
-    print("google-cloud-java {}".format(get_git_repo_version(os.path.dirname(__file__))))
 
     with io.open(os.path.expanduser("~/.artman/config.yaml"), encoding='UTF-8') as config_file:
         artman_config_data = yaml.load(config_file, Loader=yaml.Loader)
         toolkit_path = artman_config_data['local']['toolkit']
         print("gapic_generator {}".format(get_git_repo_version(toolkit_path)))
+
+    print("google-cloud-java {}".format(get_git_repo_version(os.path.dirname(__file__))))
+
+    if googleapis:
+        print("googleapis {}".format(get_git_repo_version(googleapis)))
+
+    if discovery_repo:
+        print("discovery_repo {}".format(get_git_repo_version(discovery_repo)))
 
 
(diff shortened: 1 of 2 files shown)


utilities/generate_api.py
```diff
@@ -138,6 +167,8 @@ def main():
                         help='Don\'t print informational instructions')
     args = parser.parse_args()
 
+    dump_versions(args.config_file)
```
garrettjonesgoogle: Can you put this under `if not args.quiet`? Otherwise, it will print the same thing over and over when running from `batch_generate_apis.py`.

utilities/generate_api.py
```diff
@@ -40,6 +42,33 @@
 JAVA_GAPIC="java_gapic"
 JAVA_DISCOGAPIC="java_discogapic"
 
+
+def dump_versions(config_path):
+    print("Component versions:")
+
+    print(subprocess.check_output(['artman', '--version'], stderr=subprocess.STDOUT).strip())
+
+    print("googleapis %s" % get_git_repo_version(os.path.dirname(config_path)))
```
garrettjonesgoogle: use the more modern style: `"googleapis {}".format(get...)`

utilities/generate_api.py
```diff
@@ -40,6 +42,33 @@
...
+    print("google-cloud-java %s" % get_git_repo_version(os.path.dirname(__file__)))
+
+    with io.open(os.path.expanduser("~/.artman/config.yaml"), encoding='UTF-8') as config_file:
+        artman_config_data = yaml.load(config_file, Loader=yaml.Loader)
+        toolkit_path = artman_config_data['local']['toolkit']
+        print("gapic_generator  %s" % get_git_repo_version(toolkit_path))
+
+
+def get_git_repo_version(path):
```
garrettjonesgoogle: Move this above `dump_versions` so that the ordering in the file is consistent (lowest-level functions at the top)

utilities/generate_api.py
igorbernstein2: I think something needs to be printed for batch_generate_apis.py, but I'm not sure the best way to share the functionality. I think repeatedly printing it is not ideal, but is better then nothing

garrettjonesgoogle: Call this function (`dump_versions`) directly from `batch_generate_apis.py`.

igorbernstein2: That works

utilities/generate_api.py
igorbernstein2: done

utilities/generate_api.py
igorbernstein2: done

PR #3554, opened 2018-08-10
Add Container Analysis v1beta1 for the first time

andreamlin:


andreamlin:
Java proto classes will not compile on account of one of the input protos specifying a package name that ends in ".package", which is a Java compile error.

Waiting for API owner to fix proto.

PR #3513, opened 2018-07-27
Bigtable Admin: Promote models to top level classes

igorbernstein2:
This aims be a mechanical structural change. It promotes the inner classes of `TableAdminRequests` & `TableAdminResponses` to be toplevel classes in `com.google.cloud.bigtable.admin.v2.models` and moves all of the static `convertX` methods in `TableAdminResponses` to be static `fromProto` methods in the relevant model.

PR #3570, opened 2018-08-16
Bigtable: clean up consistency token

igorbernstein2:
This is a followup cleanup work for #3513.

Consistency tokens are married to the table for which they were generated for.


igorbernstein2:
@kevinsi4508 can you take a look?

PR #3571, opened 2018-08-16
Bigtable: cleanup of futures + extras

igorbernstein2:
This started off as cleaning up deprecation warnings about futures needing a executor, but now it also has:
- re-throw errors from futures to show stack traces of the caller
- improved  client tests that also test results
- added equals & hashcode to models to ease testing
- fixed a bug in list tables that would only return the first page


pongad:
There seems to be a null pointer exception. I can't tell readily tell from the code how it's happening though :P

igorbernstein2:
It doesn't really affect the usability, so i'll just merge.

PR #3577, opened 2018-08-17
bump checkstyle version to build on Java 9

pongad:
The version of checkstyle we currently use uses tools.jar
which is removed from Java 9 and above as part of Project Jigsaw.
This commit uses a newer version of checkstyle that does not
use tools.jar.

Running `mvn checkstyle:checkstyle` succeeded.


pongad:
It looks like newer versions of checkstyle are not compat with Java 7....?

PR #3597, opened 2018-08-22
(WIP) Spanner: Block nested transactions

nithinsujir:
Cloud spanner does not support nested transactions. Use a thread-local
flag to check and throw exception.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
@@ -128,6 +128,19 @@ class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
   private static final String QUERY = "CloudSpannerOperation.ExecuteStreamingQuery";
   private static final String READ = "CloudSpannerOperation.ExecuteStreamingRead";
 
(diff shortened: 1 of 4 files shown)


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -1238,12 +1244,18 @@ void backoffSleep(Context context, long backoffMillis) {
     @Nullable
     @Override
     public <T> T run(TransactionCallable<T> callable) {
+      if (inTransaction.get() == Boolean.TRUE) {
```
snehashah16: can we avoid the call to begin a txn also ? the check here will only disallow it to be executed.

nithinsujir: Do you mean have the check in more than one place? It seems to me that we should do it in one place. If beginTransaction() is a better place, then we should move it there and get rid of the check here.
beginTransaction() seems to be at the Session level. Is a session allowed to have only one pending transaction?

nithinsujir:
Because I did a push -f earlier, circleci cannot build this branch anymore. Will close and create new PR.

PR #3628, opened 2018-08-31
Spanner: Block nested transactions

nithinsujir:
Cloud spanner does not support nested transactions. Use a thread-local
flag to check and throw exception. The flag is set to true when a transaction begins running. The checks are present when the possibly nested transactions are being prepared.

Single and batch transactions have slightly different paths, so the check is needed in two places.

+ @jcorwin


google-cloud-clients/google-cloud-spanner/pom.xml
```diff
@@ -18,6 +18,7 @@
   </parent>
   <properties>
     <site.installationModule>google-cloud-spanner</site.installationModule>
+    <skipUnitTests>${skipTests}</skipUnitTests>
```
snehashah16: lets not skip tests during install. it is helpful sanity check. u can use -DskipTests if tests are not required

google-cloud-clients/google-cloud-spanner/pom.xml
```diff
@@ -27,6 +28,7 @@
         <version>2.12.4</version>
         <configuration>
           <excludedGroups>com.google.cloud.spanner.IntegrationTest</excludedGroups>
+          <skipTests>${skipUnitTests}</skipTests>
```
snehashah16: is there a separate unit test rule too ?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -905,6 +917,8 @@ TransactionContextImpl newTransaction() {
     }
     
     <T extends SessionTransaction> T setActive(@Nullable T ctx) {
+      throwIfTransactionsPending();
```
snehashah16: what happens when they do this ?

txn1 = client.readWriteTransaction();
txn2 = client.readWriteTransaction(); // should it fail here ?

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java
```diff
@@ -349,4 +353,110 @@ public Void run(TransactionContext transaction) throws SpannerException {
...
+                          }
+                        });
+
+                return null;
+              }
+            });
+  }
+
+  @Test
+  public void nestedRwRwTransactionShouldThrowException() {
```
snehashah16: RwRw a typo ?

also have RW (i assume its for ReadWrite)

snehashah16: nit: I would prefer u shorten "TransactionShouldThrowException" over RW RD etc since those explain the test case..

google-cloud-clients/google-cloud-spanner/pom.xml
nithinsujir: This does not skip tests during install. It only sets the default value of skipUnitTests to skipTests.

google-cloud-clients/google-cloud-spanner/pom.xml
nithinsujir: I'm not following. This _is_ the unit test rule. surefire plugin is for unit tests.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
nithinsujir: No, I think it shouldn't fail here because they have not run the transaction. It's ok to have multiple prepares.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java
nithinsujir: RwRw -> Nested Read-Write transaction within a Read-Write transaction.

Didn't quite understand the shorten comment. If I shorted the name, then all the 4 methods will have the same name?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
jcorwin: We need a way to disable throwing an exception here in case the user happens to have a legitimate case for running a transaction within another transaction. This could be an option passed to SpannerOptions or settable in the TransactionContext.

nithinsujir: Thanks. Added a flag to control the behaviour.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunner.java
```diff
@@ -73,4 +73,6 @@
    * {@link #run(TransactionCallable)} has returned normally.
    */
   Timestamp getCommitTimestamp();
+
+  TransactionRunner allowNestedTransaction();
```
snehashah16: add detailed documentation for this method, including the information that this is not good practice. only for legit cases (eg: txn across DBs ).

provide examples of nesting which we would like to discourage

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java
snehashah16: had offline conversation

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -1239,11 +1259,19 @@ void backoffSleep(Context context, long backoffMillis) {
...
+          hasPendingTransaction.set(Boolean.TRUE);
+        }
+
         return runInternal(callable);
       } catch (RuntimeException e) {
         TraceUtil.endSpanWithFailure(span, e);
         throw e;
       } finally {
+        if (blockNestedTxn) {
+          hasPendingTransaction.set(Boolean.FALSE);
```
pongad: I'd use `hasPendingTransaction.remove()` instead. OpenJDK implementation (and possibly others) use `Map<Thread, Value>` underneath, `remove` will actually remove the mapping, setting this to false is a bit of a memory leak.

Is there a chance someone might change `allowNestedTransaction` while a transaction is running? If there is, We might want to remove unconditionally.

nithinsujir: Thanks for the info on the memory leak. 
Although unlikely, it _is_ possible in theory to change the flag so best to handle it I guess.

Good catch. Will make both the changes.

nithinsujir:
@pongad Has my new commit addressed your comments adequately?

pongad:
@nithinsujir Hey sorry. I think the code looks good. There's a merge conflict though. Could you take a look? I think I can merge after that.

nithinsujir:
Resolved the conflict.

PR #3633, opened 2018-09-01
Pubsub: Clean up after message extension gives up

csainty:
When a message is neither Acked nor Nacked, the state of the `MessageDispatcher` is not updated to reflect this.
This leads to consumption halting when `maxOutstandingElementCount` is reached.

I am not at all standing behind this as the correct fix, I just think it helps to illustrate the problem.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
@@ -78,7 +78,7 @@ class MessageDispatcher {
   private final MessageWaiter messagesWaiter;
 
   // Maps ID to "total expiration time". If it takes longer than this, stop extending.
-  private final ConcurrentMap<String, Instant> pendingMessages = new ConcurrentHashMap<>();
+  private final ConcurrentMap<AckHandler, Instant> pendingMessages = new ConcurrentHashMap<>();
 
   private final LinkedBlockingQueue<String> pendingAcks = new LinkedBlockingQueue<>();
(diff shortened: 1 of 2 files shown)


csainty:
I signed it!

csainty:
Relevant to the user confusion seen in #3394 as well

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -455,6 +455,10 @@ void extendDeadlines() {
         int sec = Math.max(1, (int) now.until(totalExpiration, ChronoUnit.SECONDS));
         modacks.add(new PendingModifyAckDeadline(sec, ackId));
         count++;
+      } else {
+        // TODO: How to get the byte count?
```
pongad: I think the best way to do this is to make `pendingMessages` map the ack ID to `AckHandler`. Then add the `totalExpiration` field to `AckHandler`. `AckHandler` already knows how big the message is, so we can use that here.

csainty:
I took a shot at that approach, looks like it works 👍

google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/MessageDispatcherTest.java
```diff
@@ -182,6 +189,10 @@ public void testExtension_GiveUp() throws Exception {
     clock.advance(1, TimeUnit.DAYS);
     dispatcher.extendDeadlines();
     assertThat(sentModAcks).isEmpty();
+
+    // We should be able to reserve another item in the flow controller and not block shutdown
+    flowController.reserve(1, 0);
```
csainty: A test could also be added that checked that this bytes are not longer reserved

PR #3640, opened 2018-09-04
Bigtable: table model improvements

igorbernstein2:
* Flatten cluster
* expose column families as list only
* use relative ids for tables & families


igorbernstein2:
@kevinsi4508 can you take a look?

igorbernstein2:
@pongad  this is ready for you

PR #3661, opened 2018-09-10
Logging: Allow deferred formatting for trace and spanId

bogdandrutu:


google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -234,7 +234,7 @@ public Builder setTrace(String trace) {
     /**
      * Sets the ID of the trace span associated with the log entry, if any.
      */
-    public Builder setSpanId(String spanId) {
+    public Builder setSpanId(Object spanId) {
```
sebright: I think it would be clearer to add new method overloads, for example, something similar to Flogger's lazy method (https://github.com/google/flogger/blob/36aaf311e9c9042ed938e5dd0598ef675bb81b88/api/src/main/java/com/google/common/flogger/LazyArgs.java#L51), using Java 7 types. That would make it clear that the trace and span IDs are still Strings and avoid overloading the meaning of Object.toString().

bogdandrutu: @pongad what is your opinion on this. Happy with any option you suggest.

pongad: Apologies for the late reply. There are a few threads of arguments:

- We have to be Java 7 compatible for a while longer.
- I think calling `toString` like this makes things a little too magical.
- Even if we want to use `toString`, this PR currently breaks binary compatibility, we'd still need a `setSpanID(String)` overload and forward the argument.

I think something like this should make everyone mostly happy.

- We define `LazyString` interface or something like that. We make it a functional interface, and add overload that uses it.
  - It won't work well for Java 7 users, but Java 8+ users will benefit.
- Keep the current `String` overload
  - This can be implemented by converting the String into a `LazyString` that just returns the string back.
- If you think it's valuable, we can even add an implementation of `LazyString` that calls `toString` to help Java 7 users.

Does this sound reasonable to folks?

bogdandrutu: @pongad I don't understand some comments

> We have to be Java 7 compatible for a while longer.

I don't see where the current PR is not Java 7 compatible

> Even if we want to use toString, this PR currently breaks binary compatibility, we'd still need a setSpanID(String) overload and forward the argument.

The current PR seems to be backwards compatible, `String.toString()` returns the same object.

I am happy to do what you suggest but I want to clarify the motivation. The only valid argument that I can still understand is the magic behind the `toString`.

If we do what you suggest we add an extra object allocation every time we call the old API (to create the LazyString from the String). I think the current proposal is optimized for performance in both cases calling with a String vs calling with a delayed toString.

Your call.

pongad: > We have to be Java 7 compatible for a while longer.

Sorry I didn't explain myself very well. I meant that whatever we do, it shouldn't be too Java7-unfriendly. Your proposal is fine in this aspect. I wanted to explicitly call it out as a design constraint, and called it out badly :P

> The current PR seems to be backwards compatible, String.toString() returns the same object.

The current PR is definitely source-backward-compatible: any code that currently compiles will continue to compile. It isn't binary-compatible though. Java resolves overloads at compile time. For a user who compiles against the latest version, their binary will call `setSpanID(String)`. But this PR removes `setSpanID(String)` then adds `setSpanID(Object)`. If the user updates the `trace` JAR without recompiling, their application will crash with `MethodNotFound` exception. To get around this, we have to

```java
void setSpanID(String s) {setSpanID((Object) s);}
void setSpanID(Object obj) {...}
```

so that the compiler retains the `setSpanID(String)` symbol. Am I making any sense?

> If we do what you suggest we add an extra object allocation every time we call the old API 

You're absolutely right, I didn't think about the performance implications. @garrettjonesgoogle @chingor13 @JesseLovelace Do you want to tie-break?

PR #3662, opened 2018-09-10
spanner: Add snippets for ReadContext

nithinsujir:


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/ReadContext.java
```diff
@@ -113,6 +162,20 @@ ResultSet readUsingIndex(
...
    *
+   * <!--SNIPPET read_context_analyze_query-->
+   * <pre>{@code
+   * ReadContext rc = dbClient.singleUse();
+   * ResultSet resultSet =
+   *     rc.analyzeQuery(
+   *         Statement.of("SELECT SingerId, AlbumId, MarketingBudget FROM Albums"),
+   *         ReadContext.QueryAnalyzeMode.PROFILE);
+   * while (resultSet.next()) {
+   *   resultSet.getCurrentRowAsStruct();
```
snehashah16: you may add a comment here as to why discard all the intermediate resutl rows because the Stats are only provided in the last index

nithinsujir:
Will do. Thanks!

PR #3664, opened 2018-09-11
Add Kokoro CI config

chingor13:
Adding Kokoro presubmit tests for units on Java 7, 8 (linux, ows, windows), 10, javadoc, and all integration tests.

The LoggingAppenderTest is currently broken as the default resource detected from Kokoro is a GCE instance as that's where the test is running from.

### DIFF
--- a/.kokoro/continuous/java10.cfg
+++ b/.kokoro/continuous/java10.cfg
@@ -3,5 +3,5 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
     key: "TRAMPOLINE_IMAGE"
-    value: "gcr.io/cloud-devrel-kokoro-resources/java11"
+    value: "gcr.io/cloud-devrel-kokoro-resources/java10"
 }
--- a/.kokoro/presubmit/java10.cfg
+++ b/.kokoro/presubmit/java10.cfg
@@ -3,5 +3,5 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
     key: "TRAMPOLINE_IMAGE"
-    value: "gcr.io/cloud-devrel-kokoro-resources/java11"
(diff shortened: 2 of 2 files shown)


chingor13:
I will try combining the integration tests into a single grouped job that executes sub jobs in a subsequent PR.

.kokoro/presubmit/java10.cfg
```diff
@@ -0,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Configure the docker image for kokoro-trampoline.
+env_vars: {
+    key: "TRAMPOLINE_IMAGE"
+    value: "gcr.io/cloud-devrel-kokoro-resources/java11"
```
garrettjonesgoogle: This doesn't look right - java11 for the java10 job?

.kokoro/continuous/java10.cfg
```diff
@@ -0,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Configure the docker image for kokoro-trampoline.
+env_vars: {
+    key: "TRAMPOLINE_IMAGE"
+    value: "gcr.io/cloud-devrel-kokoro-resources/java11"
```
garrettjonesgoogle: java11 for a java10 job?

.kokoro/presubmit/java10.cfg
chingor13: Whoops, you're right, I was testing the java11 image without having to make internal changes. Will fix this.

chingor13:
Kokoro tests are triggered automatically only for organization members and contributors. For external contributors, we do a quick check to make sure they're not trying to change the build to expose any secrets, then we add the `kokoro:force-run` label and the kokoro bot will pick it up.

PR #3670, opened 2018-09-12
Add redis-v1 and video-intelligence-v1p2beta1 to batch

garrettjonesgoogle:
This is https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3657 minus iamcredentials, which doesn't generate correctly


garrettjonesgoogle:
Merging because the batch file doesn't affect tests.

PR #3676, opened 2018-09-13
Verify protoc version for batch-generation

andreamlin:
Fail fast when using `utilities/batch_generate_apis.py` if the local protoc version doesn't match the protobuf-java version defined in the pom.xml.

Tested locally on the VM that I used last week with the wrong protoc version, and the output is 
```
Local version of protoc is libprotoc 3.4.0
 (see output of `which protoc`). Please use protoc version 3.6.0 to match the version of protobuf-java used in this repo.
```
Also tested on my local desktop, with protoc=3.6.0, and generation proceeds as usual.


andreamlin:
PTAL

PR #3686, opened 2018-09-14
Expand test coverage for getDatabaseClient()

nithinsujir:
This change also adds the jacoco coverage plugin in the pom.xml and sets
jacoco.skip to true to disable it by default. It can be enabled by
passing -Djacoco.skip=false to the mvn command.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerImplTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerImplTest.java
@@ -90,7 +90,7 @@ public class SpannerImplTest {
     // Get db client again
     DatabaseClient databaseClient1 = impl.getDatabaseClient(db);
 
-    assertThat(databaseClient == databaseClient1);
+    assertThat(databaseClient1).isSameAs(databaseClient);
   }
 
   @Test


nithinsujir:
Add @biswanag

google-cloud-clients/google-cloud-spanner/pom.xml
```diff
@@ -18,9 +18,29 @@
...
   <properties>
     <site.installationModule>google-cloud-spanner</site.installationModule>
+    <jacoco.skip>true</jacoco.skip>
   </properties>
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <version>0.8.2</version>
```
snehashah16: check if google-cloud-bom lists this too. We may get the latest version from there, if it exists

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerImplTest.java
```diff
@@ -70,4 +71,49 @@ public void createAndCloseSession() {
     // The same channelHint is passed for deleteSession (contained in "options").
     Mockito.verify(rpc).deleteSession(sessionName, options.getValue());
   }
+
+  @Test
+  public void getDbclientAgainGivesSame() {
```
snehashah16: Is this part of the review ?

google-cloud-clients/google-cloud-spanner/pom.xml
nithinsujir: Do you mean in the bom's pom.xml? No, there is no jacoco in the entire repo.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerImplTest.java
nithinsujir: Yes, this is the main part of the change, i.e. adding expanding test coverage.

google-cloud-clients/google-cloud-spanner/pom.xml
nithinsujir: @pongad Michael, we have a question whether this should go in the bom rather than here. And whether jacoco should be added repo-wide (skipped by default) rather than just for spanner.

What do you suggest?

pongad: I think here is fine. BOM is imported by other people, so I think test deps should go here.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerImplTest.java
```diff
@@ -70,4 +71,49 @@ public void createAndCloseSession() {
...
+        .thenReturn(GrpcTransportOptions.newBuilder().build());
+    Mockito.when(spannerOptions.getSessionPoolOptions())
+        .thenReturn(SessionPoolOptions.newBuilder().build());
+
+    DatabaseClient databaseClient = impl.getDatabaseClient(db);
+
+    // Get db client again
+    DatabaseClient databaseClient1 = impl.getDatabaseClient(db);
+
+    assertThat(databaseClient == databaseClient1);
```
pongad: I don't think this actually asserts anything, I think you have to `assertThat(databaseClient1).isSameAs(databaseClient)`

nithinsujir: Will do, thanks.

PR #3694, opened 2018-09-18
Fixes GoogleCloudPlatform/google-cloud-java#3283.

MMMarcy:


google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java
```diff
@@ -540,7 +545,9 @@ public void testCreateTable() {
     EasyMock.expect(bigqueryRpcMock.create(tableInfo.toPb(), EMPTY_RPC_OPTIONS))
         .andReturn(tableInfo.toPb());
     EasyMock.replay(bigqueryRpcMock);
-    bigquery = options.getService();
+    BigQueryOptions bigQueryOptions =
+        createBigQueryOptionsForProject(OTHER_PROJECT, rpcFactoryMock);
```
pongad: I haven't looked at this code project in a while, but why are we changing these tests?

MMMarcy: The default options did have an [hardcoded value](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/c1fb3904c648d4db17b7aa8bbcd970d0e0374bb6/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/BigQueryImplTest.java#L295) for projectId. So, any call using options mock returned TableId and Tableinfo with the hardcoded value and assertions started fail now that the update value for projectId is returned.

pongad: Ah that makes sense.

PR #3695, opened 2018-09-18
Bigtable: limit mutation sizes in the client to avoid overloading the server

igorbernstein2:


igorbernstein2:
@kevinsi4508, can you take a look when you have a moment?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -172,26 +179,36 @@ public Mutation deleteCells(
...
   }
 
+  private void addMutation(com.google.bigtable.v2.Mutation mutation) {
+    Preconditions.checkState(numMutations + 1 <= MAX_MUTATIONS,
+        "Too many mutations per row");
+    Preconditions.checkState(byteSize + mutation.getSerializedSize() <= MAX_MUTATION_SIZE,
+        "Byte size of mutations is too large");
+
+    numMutations++;
+    byteSize += mutation.getSerializedSize();
```
sduskis: you should probably `getSerializedSize()` once rather than twice.

igorbernstein2: getSerializedSize is memoized, so it shouldn't matter

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -36,11 +37,19 @@
  * encapsulate a list of mutations that will to be applied to a single row.
  */
 public final class Mutation implements MutationApi<Mutation>, Serializable {
-  private static final long serialVersionUID = 5893216644683374339L;
+  private static final long serialVersionUID = 5893216644683374340L;
+
+  @InternalApi("Visible for testing")
+  static final int MAX_MUTATIONS = 100_000;
+  @InternalApi("Visible for testing")
+  static final int MAX_MUTATION_SIZE = 200 * 1024 * 1024;
```
kevinsi4508: I assume this is max bytes size of all mutations. Sounds a bit like max bytes size per mutation. "MAX_BYTE_SIZE"? Make it long instead of int?
Do we actually enforcing this on server side? Is this the max bytes read limit?

igorbernstein2: I renamed the const, but left it as an int. I don't think we need to worry about > 4 gb request sizes.

igorbernstein2:
@pongad this is ready for you when you get a chance

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
kevinsi4508: Oh, I am not particularly worried about > 4 gb thingy. "byteSize" in long, that is why I was suggesting using long for max byte size as well.

PR #3696, opened 2018-09-18
Bigtable: be consistent and don't expose absolute table names

igorbernstein2:
The entire bigtable admin surface uses relative ids. So should listTables()


igorbernstein2:
@chingor13 Can this be merged?

igorbernstein2:
Yep, this is intentional. Thanks for checking

igorbernstein2:
There a few more breaking changes coming up

PR #3706, opened 2018-09-20
spanner: Options: Fix null dereference, expand test coverage

nithinsujir:
Options coverage now at 100%


nithinsujir:
R: @biswanag

PR #3709, opened 2018-09-20
[Storage] Bucket lock samples and bug fixes.

frankyn:
### Bug fixes from sample review
- [x]  Clean up comments for Boolean types in the Storage client library to explain possible states.
- [x]  Fix Integration tests that kept failing due to objects with retention by releasing holds during cleanup.
- [x]  Fix bug in client library not allowing to clear retentionPolicy correctly.
- [x]  Be consistent with how Booleans are handled between Bucket and Blob classes.
- [x]  Reformat BlobInfo and BucketInfo classes.

### Samples
- [x]  Update storage snippets bucket clean up to follow integration tests.
- [x]  Define a retention policy
- [x]  Remove a retention policy if not locked
- [x]  Lock a retention policy
- [x]  Set temporary hold
- [x]  Release temporary hold
- [x]  Set event-based hold
- [x]  Release event-based hold
- [x]  Set default event-based hold
- [x]  Release default event-based hold
- [x]  View default event-based hold
- [x]  View bucket retention policy metadata
- [x]  View metadata for object
- [x]  Run formatter.

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/testing/RemoteStorageHelper.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/testing/RemoteStorageHelper.java
@@ -84,10 +84,7 @@ public class RemoteStorageHelper {
                   for (Blob blob : bucket.list(BlobListOption.fields(Storage.BlobField.EVENT_BASED_HOLD,
                       Storage.BlobField.TEMPORARY_HOLD)).iterateAll()) {
                     if(blob.getEventBasedHold() == true || blob.getTemporaryHold() == true) {
-                      BlobInfo.Builder blobBuilder = BlobInfo.newBuilder(bucket.getName(), blob.getName());
-                      blobBuilder.setTemporaryHold(false);
-                      blobBuilder.setEventBasedHold(false);
-                      storage.update(blobBuilder.build());
+                      storage.update(blob.toBuilder().setTemporaryHold(false).setEventBasedHold(false).build());
                     }
                   }
                   forceDelete(storage, bucket.getName());
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
@@ -1152,23 +1152,17 @@ public class StorageSnippets {
     System.out.println("StorageClass: " + blob.getStorageClass());
     System.out.println("TimeCreated: " + new Date(blob.getCreateTime()));
     System.out.println("Last Metadata Update: " + new Date(blob.getUpdateTime()));
-    System.out.println(
-        "temporaryHold: "
-            + (blob.getTemporaryHold() != null && blob.getTemporaryHold()
-                ? "Enabled"
-                : "Disabled"));
-    System.out.println(
-        "eventBasedHold: "
-            + (blob.getTemporaryHold() != null && blob.getTemporaryHold()
-                ? "Enabled"
-                : "Disabled"));
+    Boolean temporaryHoldIsEnabled = (blob.getTemporaryHold() != null && blob.getTemporaryHold());
+    System.out.println("temporaryHold: " + (temporaryHoldIsEnabled ? "enabled" : "disabled"));
+    Boolean eventBasedHoldIsEnabled = (blob.getEventBasedHold() != null && blob.getEventBasedHold());
+    System.out.println("eventBasedHold: " + (eventBasedHoldIsEnabled ? "enabled" : "disabled"));
     if (blob.getRetentionExpirationTime() != null) {
       System.out.println("retentionExpirationTime: " + new Date(blob.getRetentionExpirationTime()));
     }
     if (blob.getMetadata() != null) {
-      System.out.println("User metadata:");
+      System.out.println("\n\n\nUser metadata:");
       for (Map.Entry<String, String> userMetadata : blob.getMetadata().entrySet()) {
-        System.out.println(userMetadata.getKey() + " " + userMetadata.getValue());
+        System.out.println(userMetadata.getKey() + "=" + userMetadata.getValue());
       }
     }
     // [END storage_get_metadata]
@@ -1184,22 +1178,21 @@ public class StorageSnippets {
     // The name of a bucket, e.g. "my-bucket"
     // String bucketName = "my-bucket";
 
-    // The retetion period for objects in bucket
-    // 1 hour in seconds
-    // Long retentionPeriod = 3600L;
+    // The retention period for objects in bucket
+    // Long retentionPeriod = 3600L; // 1 hour in seconds
 
-    Bucket bucket =
+    Bucket bucketWithRetentionPolicy =
         storage.update(
             BucketInfo.newBuilder(bucketName).setRetentionPeriod(retentionPeriod).build());
 
     System.out.println(
-        "Retention period for " + bucketName + " is now " + bucket.getRetentionPeriod());
+        "Retention period for " + bucketName + " is now " + bucketWithRetentionPolicy.getRetentionPeriod());
     // [END storage_set_retention_policy]
-    return bucket;
+    return bucketWithRetentionPolicy;
   }
 
   /** Example of removing a retention policy on a bucket */
-  public Bucket removeRetentionPolicy(String bucketName) throws StorageException {
+  public Bucket removeRetentionPolicy(String bucketName) throws StorageException, IllegalArgumentException {
     // [START storage_remove_retention_policy]
     // Instantiate a Google Cloud Storage client
     Storage storage = StorageOptions.getDefaultInstance().getService();
@@ -1209,15 +1202,14 @@ public class StorageSnippets {
 
     Bucket bucket = storage.get(bucketName, BucketGetOption.fields(BucketField.RETENTION_POLICY));
     if (bucket.retentionPolicyIsLocked() != null && bucket.retentionPolicyIsLocked()) {
-      System.out.println("Unable to remove retention period as retention policy is locked.");
-      return bucket;
+       throw new IllegalArgumentException("Unable to remove retention period as retention policy is locked.");
     }
 
-    Bucket updated_bucket = bucket.toBuilder().setRetentionPeriod(null).build().update();
+    Bucket bucketWithoutRetentionPolicy = bucket.toBuilder().setRetentionPeriod(null).build().update();
 
     System.out.println("Retention period for " + bucketName + " has been removed");
     // [END storage_remove_retention_policy]
-    return updated_bucket;
+    return bucketWithoutRetentionPolicy;
   }
 
   /** Example of removing a retention policy on a bucket */
@@ -1232,7 +1224,6 @@ public class StorageSnippets {
     Bucket bucket = storage.get(bucketName, BucketGetOption.fields(BucketField.RETENTION_POLICY));
 
     System.out.println("Retention Policy for " + bucketName);
-
     System.out.println("Retention Period: " + bucket.getRetentionPeriod());
     if (bucket.retentionPolicyIsLocked() != null && bucket.retentionPolicyIsLocked()) {
       System.out.println("Retention Policy is locked");
@@ -1255,14 +1246,14 @@ public class StorageSnippets {
 
     Bucket bucket =
         storage.get(bucketName, Storage.BucketGetOption.fields(BucketField.METAGENERATION));
-    Bucket updated_bucket =
+    Bucket lockedBucket =
         bucket.lockRetentionPolicy(Storage.BucketTargetOption.metagenerationMatch());
 
     System.out.println("Retention period for " + bucketName + " is now locked");
     System.out.println(
-        "Retention policy effective as of " + new Date(updated_bucket.getRetentionEffectiveTime()));
+        "Retention policy effective as of " + new Date(lockedBucket.getRetentionEffectiveTime()));
     // [END storage_lock_retention_policy]
-    return updated_bucket;
+    return lockedBucket;
   }
 
   /** Example of how to enable default event-based hold for a bucket */
@@ -1311,7 +1302,7 @@ public class StorageSnippets {
     Bucket bucket =
         storage.get(bucketName, BucketGetOption.fields(BucketField.DEFAULT_EVENT_BASED_HOLD));
 
-    if (bucket.getDefaultEventBasedHold() == true) {
+    if (bucket.getDefaultEventBasedHold() != null && bucket.getDefaultEventBasedHold()) {
       System.out.println("Default event-based hold is enabled for " + bucketName);
     } else {
       System.out.println("Default event-based hold is not enabled for " + bucketName);
--- a/google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
+++ b/google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
@@ -38,6 +38,7 @@ import com.google.cloud.storage.Bucket.BlobTargetOption;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.testing.RemoteStorageHelper;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Sets;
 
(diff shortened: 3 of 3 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    // Instantiate a Google Cloud Storage client
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+
+    // The name of a bucket, e.g. "my-bucket"
+    // String bucketName = "my-bucket";
+
+    Bucket bucket = storage.get(bucketName, BucketGetOption.fields(BucketField.RETENTION_POLICY));
+    if (bucket.retentionPolicyIsLocked() != null && bucket.retentionPolicyIsLocked()) {
+      System.out.println("Unable to remove retention period as retention policy is locked.");
+      return bucket;
```
kurtisvg: A returned statement wont make sense here in the context of the snippet. Would it be better to throw an error instead?

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+      throws StorageException {
+    // [START storage_set_retention_policy]
+    // Instantiate a Google Cloud Storage client
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+
+    // The name of a bucket, e.g. "my-bucket"
+    // String bucketName = "my-bucket";
+
+    // The retetion period for objects in bucket
+    // 1 hour in seconds
```
kurtisvg: nit: Move this comment to the next line (`3600L; // 1 hour`) (Otherwise it feels a little unclear what it's describing) 

Also s/retetion/retention

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    // Long retentionPeriod = 3600L;
+
+    Bucket bucket =
+        storage.update(
```
kurtisvg: This bit here is kind of confusing. Are buckets immutable? does `storage.update` return a new bucket or a reference to the original bucket?

It might be more clear if you separate the builder from the update call (and if `storage.update` returns a reference to a new bucket instead of just updating the original object, consider using a different name to suggest at that behavior)

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    // The name of a bucket, e.g. "my-bucket"
+    // String bucketName = "my-bucket";
+
+    Bucket bucket = storage.get(bucketName, BucketGetOption.fields(BucketField.RETENTION_POLICY));
+    if (bucket.retentionPolicyIsLocked() != null && bucket.retentionPolicyIsLocked()) {
+      System.out.println("Unable to remove retention period as retention policy is locked.");
+      return bucket;
+    }
+
+    Bucket updated_bucket = bucket.toBuilder().setRetentionPeriod(null).build().update();
```
kurtisvg: Are buckets immutable? Do you need to create a new builder instead of just updating the original bucket? What's the difference between `bucket.update` vs `storage.update(bucket)`?

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    // Instantiate a Google Cloud Storage client
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+
+    // The name of a bucket, e.g. "my-bucket"
+    // String bucketName = "my-bucket";
+
+    Bucket bucket = storage.get(bucketName, BucketGetOption.fields(BucketField.RETENTION_POLICY));
+
+    System.out.println("Retention Policy for " + bucketName);
+
```
kurtisvg: nit: Unnecessary empty line?

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    // [START storage_lock_retention_policy]
+    // Instantiate a Google Cloud Storage client
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+
+    // The name of a bucket, e.g. "my-bucket"
+    // String bucketName = "my-bucket";
+
+    Bucket bucket =
+        storage.get(bucketName, Storage.BucketGetOption.fields(BucketField.METAGENERATION));
+    Bucket updated_bucket =
```
kurtisvg: Do you need to call `bucket.update` here?

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    // Instantiate a Google Cloud Storage client
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+
+    // The name of a bucket, e.g. "my-bucket"
+    // String bucketName = "my-bucket";
+
+    Bucket bucket =
+        storage.get(bucketName, BucketGetOption.fields(BucketField.DEFAULT_EVENT_BASED_HOLD));
+
+    if (bucket.getDefaultEventBasedHold() == true) {
```
kurtisvg: Remove comparison

google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
```diff
@@ -435,6 +437,16 @@ public void testBlobDownload() throws Exception {
     assertArrayEquals(BLOB_BYTE_CONTENT, readBytes);
   }
 
+  @Test
+  public void testGetBlobMetadata() {
```
kurtisvg: These tests don't really test the snippet, they test the same thing the snippet is running. Ideally you should run the snippet and check the output for indicators that everything worked well.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/testing/RemoteStorageHelper.java
```diff
@@ -80,6 +81,15 @@ public void run() {
             for (Bucket bucket : buckets.iterateAll()) {
               if (bucket.getCreateTime() < olderThan) {
                 try {
+                  for (Blob blob : bucket.list(BlobListOption.fields(Storage.BlobField.EVENT_BASED_HOLD,
+                      Storage.BlobField.TEMPORARY_HOLD)).iterateAll()) {
+                    if(blob.getEventBasedHold() == true || blob.getTemporaryHold() == true) {
+                      BlobInfo.Builder blobBuilder = BlobInfo.newBuilder(bucket.getName(), blob.getName());
```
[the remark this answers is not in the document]
frankyn: Ah good point.

google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
frankyn: I wanted to keep this test simple, but I'll add that in.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -2081,7 +2083,7 @@ private void retentionPolicyLockRequesterPays(boolean requesterPays) throws Exec
     }
     Bucket remoteBucket = storage.create(bucketInfo);
     try {
-      assertFalse(remoteBucket.retentionPolicyIsLocked());
+      assertNull(remoteBucket.retentionPolicyIsLocked());
```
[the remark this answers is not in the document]
frankyn: Long story short: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/6235c62c63d00d9243f308bd34c3924fe56aa182/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L878

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    System.out.println("MediaLink: " + blob.getMediaLink());
+    System.out.println("Metageneration: " + blob.getMetageneration());
+    System.out.println("Name: " + blob.getName());
+    System.out.println("Size: " + blob.getSize());
+    System.out.println("StorageClass: " + blob.getStorageClass());
+    System.out.println("TimeCreated: " + new Date(blob.getCreateTime()));
+    System.out.println("Last Metadata Update: " + new Date(blob.getUpdateTime()));
+    System.out.println(
+        "temporaryHold: "
+            + (blob.getTemporaryHold() != null && blob.getTemporaryHold()
```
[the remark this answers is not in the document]
frankyn: ack

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1113,4 +1114,289 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+                ? "Enabled"
+                : "Disabled"));
+    if (blob.getRetentionExpirationTime() != null) {
+      System.out.println("retentionExpirationTime: " + new Date(blob.getRetentionExpirationTime()));
+    }
+    if (blob.getMetadata() != null) {
+      System.out.println("User metadata:");
+      for (Map.Entry<String, String> userMetadata : blob.getMetadata().entrySet()) {
+        System.out.println(userMetadata.getKey() + " " + userMetadata.getValue());
```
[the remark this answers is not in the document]
frankyn: ack

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: ack

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: Bucket reference didn't exist before and is being returned by `storage.update`. I'll address the naming e.g. `bucketWithRetentionPolicy`, I don't understand what you mean by updating the original object.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: Ack

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: That's correct, buckets are immutable. And `bucket.update()` calls `storage.update()` under the hood.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: ack

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: No `lockRetentionPolicy()` is a new RPC.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: whoops this is wrong. acking.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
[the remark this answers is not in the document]
kurtisvg: autogen?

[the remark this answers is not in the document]
frankyn: No autogen, still manual. The library had existing behavior of handling how fields are selected during a GET request. For example, if I don't select the field `Storage.BucketField#RETENTION_POLICY` then it's not returned in the bucket metadata resulting in `null`. If I do, then if a retention policy is not set on a bucket then it will be implicitly false or "null". If the field is selected and the field is returned then it's `true`. This method will never return false neither will existing Boolean method for versioning.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: Thanks for the feedback I'll do this instead!

frankyn: Wrong call, I need a bucket context before I can call `bucket.update` this only creates a BucketInfo. I'll make the adjustment where I can.

frankyn:
Comments addressed. @kurtisvg PTAL

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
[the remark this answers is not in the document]
kurtisvg: Sorry, I should have read the entirety of the javadoc. Could you expand on why `com.google.cloud.storage.Storage.BucketField#RETENTION_POLICY` can be set and still return null? I would expect a `False` in that scenario.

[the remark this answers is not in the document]
frankyn: The GCS API doesn't return a value if the field is either `null` or `false`.  Only field that does work as expected is requester pays which is considered a bug due to inconsistency.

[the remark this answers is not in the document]
frankyn: Additionally the client library does not handle this behavior gracefully for existing Booleans.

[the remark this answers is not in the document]
kurtisvg: Ah, that's frustrating. I think using null as false is generally considered an anti-pattern, but I'm not sure in this case since the information is sometimes unknown.  From a usability perspective I feel like I would rather have `false` than `null` so I wouldn't be forced to do a null check. But probably outside the scope of this PR.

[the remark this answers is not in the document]
frankyn: It is outside scope, but it's also making an already bad issue worse by continuing the same behavior. I'll follow-up with a few people internally on this front. There's a better solution and I can do it now, but I'd need to resolve open questions with `Bucket.versioningEnable()` behavior changes first.

What are your thoughts on return a `Long` instead of a `Date` object?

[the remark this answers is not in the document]
kurtisvg: Long instead of a Date object for what?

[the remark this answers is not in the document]
kurtisvg: getRetentionPeriod?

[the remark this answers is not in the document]
frankyn: https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3709/files#diff-2ca1569606e4f281912a4e4f5af3a408R1160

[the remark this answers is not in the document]
frankyn: DateTime from Apiary are represented as Long's

PR #3717, opened 2018-09-21
spanner: Expand test coverage for ResultSets

nithinsujir:
ResultSets coverage now at 99%.


google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ResultSetsTest.java
```diff
@@ -75,12 +179,52 @@ public void resultSetIteration() {
...
     assertThat(rs.next()).isTrue();
     assertThat(rs.getCurrentRowAsStruct()).isEqualTo(struct2);
     assertThat(rs.getString(0)).isEqualTo("y");
     assertThat(rs.getLong(1)).isEqualTo(3L);
     assertThat(rs.isNull(2)).isTrue();
     assertThat(rs.next()).isFalse();
+
+    try {
+      rs.getStats();
+      fail("Exception expected");
```
snehashah16: this will fail after my change for DML .. would u mind holding off on ur CL just so that I dont have to redo my PR for DML/PDML.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ResultSetsTest.java
```diff
@@ -34,11 +41,51 @@
 
   @Test
   public void resultSetIteration() {
+    double dval = 1.2;
+    String s = "s";
+    String byt = "101";
```
snehashah16: is this testing string or byte ?

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ResultSetsTest.java
```diff
@@ -34,11 +41,51 @@
 
   @Test
   public void resultSetIteration() {
```
snehashah16: its easier to read test case when they are as explicit as possible. ie - variable names could be more informative (even if long)

also testing edge cases or large values provide surface errors than smaller value.. e.g. - testing a large random bytes value

nithinsujir: I don't disagree if we're testing the implementation, but the purpose of this change is to expand the line coverage, i.e to ensure execution of paths. For edge case testing etc., we would need to add a lot more specific cases for each data type.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ResultSetsTest.java
nithinsujir: byte.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ResultSetsTest.java
nithinsujir: Sure, not a problem. Can you elaborate for my knowledge why it will fail?

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ResultSetsTest.java
snehashah16: i was hoping to shoot 2 birds with 1 stone, though i don't mean or need to have extensive edge case testing with this exercise either.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/ResultSetsTest.java
snehashah16: rs.getStats() returns null instead of an exception ..

nithinsujir:
Marking this WIP until DML changes are in.

nithinsujir:
@pongad Merged DML changes. Turns out there is no impact because the getStats() that DML changed is in a different class than the one in this test.

This PR is ready to go.

pongad:
We're transitioning ownership of this repo and I don't have write access anymore. @chingor13 @JesseLovelace can you help?

chingor13:
Does this need to go out with the spanner DML feature or can it wait until after the release is completed?

nithinsujir:
It can wait until after the release. It's unrelated to DML.

chingor13:
SGTM

PR #3737, opened 2018-09-26
Regenerate clients

JesseLovelace:


chingor13:
I don't think cloud assets is being generated to the right place. I just deleted these files as they were missing the pom.xml files.

PR #3743, opened 2018-09-27
pubsub: reject expired and duplicate messages

pongad:
CC @csainty 

This should let us clear through backlogs more quickly.

### DIFF
--- a/google-cloud-clients/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/NetworkClient.java
+++ b/google-cloud-clients/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/NetworkClient.java
@@ -471,8 +471,9 @@ public class NetworkClient implements BackgroundResource {
    * </code></pre>
    *
    * @param project Project ID for this request.
-   * @param networkResource Represents a Network resource. Read Networks and Firewalls for more
-   *     information. (== resource_for v1.networks ==) (== resource_for beta.networks ==)
+   * @param networkResource Represents a Network resource. Read Virtual Private Cloud (VPC) Network
+   *     Overview for more information. (== resource_for v1.networks ==) (== resource_for
+   *     beta.networks ==)
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
    */
   @BetaApi
@@ -501,8 +502,9 @@ public class NetworkClient implements BackgroundResource {
    * </code></pre>
    *
    * @param project Project ID for this request.
-   * @param networkResource Represents a Network resource. Read Networks and Firewalls for more
-   *     information. (== resource_for v1.networks ==) (== resource_for beta.networks ==)
+   * @param networkResource Represents a Network resource. Read Virtual Private Cloud (VPC) Network
+   *     Overview for more information. (== resource_for v1.networks ==) (== resource_for
+   *     beta.networks ==)
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
    */
   @BetaApi
@@ -720,8 +722,9 @@ public class NetworkClient implements BackgroundResource {
    * </code></pre>
    *
    * @param network Name of the network to update.
-   * @param networkResource Represents a Network resource. Read Networks and Firewalls for more
-   *     information. (== resource_for v1.networks ==) (== resource_for beta.networks ==)
+   * @param networkResource Represents a Network resource. Read Virtual Private Cloud (VPC) Network
+   *     Overview for more information. (== resource_for v1.networks ==) (== resource_for
+   *     beta.networks ==)
    * @param fieldMask
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
    */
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -398,6 +423,14 @@ public void nack() {
             @Override
             public void run() {
               try {
+                if (ackHandler.totalExpiration.isBefore(now())) {
+                  // Message expired while waiting. We don't extend these messages anymore,
+                  // so it was probably sent to someone else. Don't work on it.
+                  // Don't nack it either, because we'd be nacking someone else's message.
```
csainty: Is this true? Does the server hand out the same ackId each time it hands out the same message?
Without thinking too hard about how PubSub is built, that surprises me. Why isn't ackId == messageId in that case?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -433,35 +466,26 @@ void extendDeadlines() {
...
       if (totalExpiration.isAfter(extendTo)) {
         modack.ackIds.add(ackId);
-        count++;
         continue;
       }
-      it.remove();
+
+      // forget removes from pendingMessages; this is OK, concurrent maps can
+      // handle concurrent iterations and modifications.
+      entry.getValue().forget();
```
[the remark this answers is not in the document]
pongad: Concurrent maps are safe against the exception. See second paragraph on [this page](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html). I think that makes sense, because it must work when you iterate on one thread and remove on another. If it works there, it'd be pretty weird if you can't iterate and remove on the same thread!

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
pongad: The second question is easier to answer: You can have multiple subscriptions to the same topic. Acking a message in one subscription has no effect on the other subscription. If they were the same, we can't tell what subscription you're acking from.

The spec is silent on this. Generally, I don't think you can rely on the ack ID changing but you also cannot rely on it being the same. At least as of a couple months ago, the ID stays the same when you pull message multiple times. I'm sorry for being intentionally vague because I'm not sure how much info I can share. If you'd like, I can reach out to pubsub to see what they can comment publicly.

pongad:
That's right. This won't handle the case where max ack extension is zero. I'm a little concerned by this; this essentially changes the behavior from "we won't extend these" to "we'll never work on these". I'll make a small fix for this.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
csainty: Interesting, we have support contract etc with GCP so I can follow it up with our contacts 👍

csainty:
👍 
I pulled this morning and ran my sample project with the build, hits exactly-once consumption. Thanks for looking in to it.

I think this will fix https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3383 and https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2465
Also make it easier to configure for https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3152 by tweaking the max deadline so all the messages don't pull to a single subscriber and stay there, as currently reducing the max deadline to prevent this gets you in to double processing sadness.

pongad:
Don't merge this yet. Someone from pubsub should take a look.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -398,6 +423,14 @@ public void nack() {
             @Override
             public void run() {
               try {
+                if (ackHandler.totalExpiration.plusSeconds(messageDeadlineSeconds.get()).isBefore(now())) {
```
kamalaboulhosn: What happens if the following sequence occurs:

1. Message is received, but processing is delayed beyond expiration time.
2. Message is redelivered to this subscriber, but we don't add it because it's already in pendingMessages.
3. We now go to process the message, but since we are using the original ackHandler, we think it's expired, even though it got redelivered to us and we just didn't update with the new deadline.

pongad: When the processing is delayed beyond expiration time, L479 will remove the message from `pendingMessages`. I think two things can happen.

1. We remove message from `pendingMessages` before we redeliver. In this case, we'll just add the redelivered message to the queue. We'll skip processing the original message [a] but process the redelivered one.

2. We get redelivered message before we remove message from `pendingMessages`. This should be rare: if the message is in the `pendingMessages`, we should be keeping that message alive. If it happens though, we won't work on either messages. Eventually the message expires and get delivered for the 3rd time. We work on that one.

[a] We use `messageDeadlineSeconds` as a grace period in case user sets total expiration to 0. It's possible that we remove messgae from `pendingMessages`, then the message gets redelivered before the grace period is over. In that case, we still work on both messages.

PR #3760, opened 2018-10-03
Include logger as a label when mapping log events

marshallpierce:
The logger name is pretty important info for logging; I'm guessing it was just an oversight that it's not there already?

I'm not experienced with Stackdriver logging so maybe a label is the wrong place to put the logger name, but from reviewing the docs it's what made the most sense.

I tested it inside a GAE flex app and it worked fine: SLF4J logger name showed up as a label as expected.


marshallpierce:
Since this PR didn't get any attention I just made `LoggingEnhancer` implementations to do it: https://bitbucket.org/marshallpierce/gcp-logging-enhancers/src/master/

PR #3764, opened 2018-10-03
[Storage] Add launch stage annotations around bucket lock methods

frankyn:


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -260,11 +260,14 @@ static CustomerEncryption fromPb(StorageObject.CustomerEncryption customerEncryp
     abstract Builder setKmsKeyName(String kmsKeyName);
 
     /** Sets the blob's event-based hold. */
+    @GcpLaunchStage.Beta
```
chingor13: I think you're looking for the `com.google.api.core.BetaApi` annotation.

frankyn: Updated annotation. Thanks @chingor13

frankyn:
@chingor13 will merge when kokoro tests pass.

PR #3765, opened 2018-10-03
Add synthtool scripts

chingor13:
This will allow us to enable autosynth to get nightly PRs for proto changes.

For most APIs, artman is configured to generate 3 folders (grpc-google-cloud-x, proto-google-cloud-x, and google-cloud-x.

Fixes #3711

PR #3766, opened 2018-10-03
Regenerate clients

chingor13:
All service clients now expose methods that create `Operation`s as public.

Regenerated proto/grpc for Assets and Tasks


JustinBeckwith:
This is lumping together changes from *a ton* of APIs.  Is there a reasonable way for us to tell customers what has changed?  Should we consider generating a single PR for each library?

chingor13:
@JustinBeckwith this is the way the generator currently works. I added synth scripts in #3765 but it won't be ready in time for this release.

JustinBeckwith:
👍 It would be nice to track this for the future so we can provide better change logs, but this LGTM for now :)

PR #3771, opened 2018-10-04
Remove CircleCI status, coveralls, and version eye badges

chingor13:
* We migrated from Circle CI -> Kokoro
* Coveralls has not been updated since December 2017 (See #3751)
* VersionEye no longer exists as a product


chingor13:
The integration test failure (firestore test timeout) is unrelated.

PR #3775, opened 2018-10-05
Try harder to make believable directories

### LINKED ISSUE #3774: Detect directories even when not ending with "/"
The current implementation of PseudoDirectories has Files.isDirectory(path) return true only if path ends with a slash.

Users would be less surprised if instead of this behavior, we tried harder to identify things that look like directories.

So for example if `foo/bar.txt` exists, the following paths should be treated as directories:

* `foo/`
* `foo`

jean-philippe-martin:
Previous implementation of fake folders just looked for a trailing `/`. Now we also list the files with this prefix and use this to determine whether the input is a folder. This will detect `dir` in addition to just `dir/` before.

This version will still say 'yes' to anything that ends in a slash, even if they don't exist, maintaining the previous behavior.

This PR also adds support for MAX_VALUES to our fake storage provider, and adds an integration test for finding directories (passing).

Fixes #3774

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
@@ -727,7 +727,7 @@ public final class CloudStorageFileSystemProvider extends FileSystemProvider {
         try {
           BlobId blobId = cloudPath.getBlobId();
           // Null or empty name won't give us a file, so skip. But perhaps it's a pseudodir.
-          if (!Strings.isNullOrEmpty(blobId.getName())) {
+          if (!isNullOrEmpty(blobId.getName())) {
             if (isNullOrEmpty(userProject)) {
               blobInfo = storage.get(blobId);
             } else {


jean-philippe-martin:
This PR should fix issue #3774

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStoragePath.java
```diff
@@ -83,8 +86,39 @@ boolean seemsLikeADirectory() {
...
+  boolean seemsLikeADirectoryAndUsePseudoDirectories(Storage storage) {
+    if (!fileSystem.config().usePseudoDirectories()) {
+      return false;
+    }
+    if (path.seemsLikeADirectory()) {
+      return true;
+    }
+    // fancy case: the file name doesn't end in slash, but we've been asked to have pseudo dirs.
+    // Let's see if there are any files with this prefix.
+    if (null==storage) {
```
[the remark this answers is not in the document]
jean-philippe-martin: OK

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStoragePath.java
```diff
@@ -83,8 +86,39 @@ boolean seemsLikeADirectory() {
...
+    // Let's see if there are any files with this prefix.
+    if (storage == null) {
+      // we are in a context where we don't want to access the storage, so we conservatively
+      // say this isn't a directory.
+      return false;
+    }
+    Page<Blob> list = storage.list(
+        this.bucket(),
+        Storage.BlobListOption.prefix(path.removeBeginningSeparator().toString()),
+        Storage.BlobListOption.pageSize(1));
```
chingor13: What's the rationale for setting this to 1? 

Do we expect this to be resolved in a single iteration? If it's always resolved after a single request and single blob, let's comment that we expect that to be the case.

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStoragePath.java
```diff
@@ -83,8 +86,39 @@ boolean seemsLikeADirectory() {
     return path.seemsLikeADirectory();
   }
 
-  boolean seemsLikeADirectoryAndUsePseudoDirectories() {
-    return path.seemsLikeADirectory() && fileSystem.config().usePseudoDirectories();
+  boolean seemsLikeADirectoryAndUsePseudoDirectories(Storage storage) {
+    if (!fileSystem.config().usePseudoDirectories()) {
```
chingor13: If I turn off the pseudo directories option, then provide a path like `foo/bar/`, the will now return `false` when it should probably return `true`. We should probably swap the conditional order back.

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStoragePath.java
jean-philippe-martin: Good point, I added a comment. Yes, we only need to pull a single file.

If for example the user gave us: "foo/bar"

We search for anything that starts with "foo/bar/". If we find anything at all (say, "foo/bar/budget.xls") then we can say that foo/bar is a directory. If there's nothing then "foo/bar" isn't a directory.

I also took the opportunity to add the slash at the end of the search, because I realize otherwise the code might have given the wrong answer.

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStoragePath.java
jean-philippe-martin: The spec is: if you enable pseudo directories, then we give the user the illusion that GCS has folders. So if for example there's a file on GCS with name "foo/bar/hello.txt" then we pretend that folders "foo" and "foo/bar" exist, even though they're really just prefixes of a file name. If we disable them then we tell the user exactly what GCS tells us. Currently, this means that if you ask for "foo" you'll get a "file not found" error (since there's only one file and it's called "foo/bar/hello.txt").

So with pseudo directories off, this function is correct in returning false always - the function is meant to only return true if (as per the name) the given path looks like a directory and pseudo-directories are enabled. We should not swap the conditional order.

That said, I did move conditionals about because I realized it was a bad idea to have the extra list call at the start of every stat call. Now we only do this when we can't find the file.

chingor13: Ok, SGTM then.

chingor13:
@jean-philippe-martin The thing I'm most concerned about is that we're silently adding additional list requests and it's opaque and possibly unexpected. A user upgrading may see many more list requests and at scale this could cost them $$$.

jean-philippe-martin:
@chingor13 you're right, there would be more list requests: but only when the user tries to access files that do not exist (and do not end in slash). This is not a very common operation at all. And there is an easy way to make sure no list operation is sent: disable pseudo-directories (of course that has some drawbacks too), or whenever using directories, make sure to append a final "/" at the end so it's clear it's not a file name.

We could conceivably add a third intermediate mode ("pseudo-dirs that require slash", effectively what pseudo-dir mode was before this PR), but having more modes multiplies the maintenance and testing costs. I don't think it would be worth it.

What we can do is make sure the pseudodir code only triggers when the file isn't there, and I've done that already. We could search for other optimization opportunities. The other thing we can do is update the documentation. I'll be happy to do this if you think it'll be helpful.

jean-philippe-martin:
Oh you're right, it does! Let's see if I can avoid it.

jean-philippe-martin:
@chingor13 I've pushed an update that removes the call from that case and a few others.

PR #3787, opened 2018-10-08
releasetool compatibility and instructions

### LINKED ISSUE #3777: Move version management scripts to releasetool
In particular, these two scripts:

* https://github.com/googleapis/google-cloud-java/blob/master/utilities/bump_versions.py
* https://github.com/googleapis/google-cloud-java/blob/master/utilities/replace_versions.py

Key functionality that should probably be retained:

* The two-phase process of bump & replace
* The replacement of versions only at locations that are tagged
* A master version file (currently https://github.com/googleapis/google-cloud-java/blob/master/versions.txt )
* The tracking of last-released version and current version (which can be either last-released or a snapshot version)

chingor13:
Depends on https://github.com/googleapis/releasetool/pull/74

Fixes #3777

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -126,12 +126,12 @@ To push a release version
 
 13. Run `rm -rf tmp_gh-pages` to remove the generated docs directory from your local machine.
 
-14. Publish a release on Github. Run `releasetool tag`. It will list the last few merged PRs. Select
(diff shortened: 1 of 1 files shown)


chingor13:
releasetool 0.10.0 is available and includes the java functionality described here. Let's use it for tomorrow's release.

PR #3795, opened 2018-10-09
Storage: Deprecate DeleteRules and add LifecycleRules to accommodate SetStorageClass actions

JesseLovelace:
(Note: I'm holding on off Javadoc until the general class structure is approved)

This deprecates DeleteRules and establishes LifecycleRules as the standard to support object lifecycle management. The DeleteRules approach had the following problems:

- It didn't have a base "Rule" class, meaning Delete was the only supported action
- Instead of the condition being a field, all conditions were subclasses, which meant a single Delete rule could only have one condition, whereas the API specifies that a rule can have any number of conditions as long as it's at least one. 
- Following this pattern would have meant creating a new amount of subclasses equal to the number of conditions for each new action (i.e. to support SetStorageClass we would need to add AgeSetStorageClassRule, CreatedBeforeSetStorageClassRule, etc.)

Instead of following this pattern I've established a new pattern that more closely mirrors the API spec. Now rules are described by LifecycleRules, which have a LifecycleAction (currently Delete or SetStorageClass) and a LifecycleCondition(At least one of age, created before, num newer versions, etc). This way if a new action is created we only have to add one new subclass of LifecycleAction (instead of an amount equal to the number of conditions) and if a new condition is created we just add one new field to LifecycleCondition. 

In addition to the new approach, this PR adds support for SetStorageClass lifecycle actions and MatchesStorageClass lifecycle conditions.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -31,6 +31,7 @@ fi
 case $JOB_TYPE in
 test)
     mvn test -B
+    bash $KOKORO_GFILE_DIR/codecov.sh
     ;;
 javadoc)
     mvn javadoc:javadoc javadoc:test-javadoc
--- a/.kokoro/continuous/common.cfg
+++ b/.kokoro/continuous/common.cfg
@@ -28,8 +28,6 @@ before_action {
     keystore_resource {
       keystore_config_id: 73713
       keyname: "java_it_service_account"
-      # TODO(chingor): remove this after secrets have globally propagated
-      backend_type: FASTCONFIGPUSH
     }
   }
 }
@@ -39,8 +37,16 @@ before_action {
     keystore_resource {
       keystore_config_id: 73713
       keyname: "java_firestore_it_service_account"
-      # TODO(chingor): remove this after secrets have globally propagated
-      backend_type: FASTCONFIGPUSH
+    }
+  }
+}
+
+# Request the codecov master token
+before_action {
+  fetch_keystore {
+    keystore_resource {
+      keystore_config_id: 73713
+      keyname: "dpebot_codecov_token"
     }
   }
 }
--- a/.kokoro/presubmit/common.cfg
+++ b/.kokoro/presubmit/common.cfg
@@ -28,8 +28,6 @@ before_action {
     keystore_resource {
       keystore_config_id: 73713
       keyname: "java_it_service_account"
-      # TODO(chingor): remove this after secrets have globally propagated
-      backend_type: FASTCONFIGPUSH
     }
   }
 }
@@ -39,8 +37,17 @@ before_action {
     keystore_resource {
       keystore_config_id: 73713
       keyname: "java_firestore_it_service_account"
-      # TODO(chingor): remove this after secrets have globally propagated
-      backend_type: FASTCONFIGPUSH
     }
   }
 }
+
+# Request the codecov master token
+before_action {
+  fetch_keystore {
+    keystore_resource {
+      keystore_config_id: 73713
+      keyname: "dpebot_codecov_token"
+    }
+  }
+}
+
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,4 @@
+---
+codecov:
+  ci:
+    - source.cloud.google.com
--- a/google-api-grpc/pom.xml
+++ b/google-api-grpc/pom.xml
@@ -788,6 +788,25 @@
           </reportPlugins>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <version>0.8.2</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>report</id>
+            <phase>test</phase>
+            <goals>
+              <goal>report</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
   <reporting>
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -1127,6 +1127,25 @@
           </reportPlugins>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <version>0.8.2</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>report</id>
+            <phase>test</phase>
+            <goals>
+              <goal>report</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 
--- a/google-cloud-clients/google-cloud-bigtable-admin/README.md
+++ b/google-cloud-clients/google-cloud-bigtable-admin/README.md
@@ -72,22 +72,23 @@ at the top of your file:
 
 
 ```java
-import com.google.bigtable.admin.v2.ColumnFamily;
+import static com.google.cloud.bigtable.admin.v2.models.GCRules.GCRULES;
 import com.google.bigtable.admin.v2.InstanceName;
 import com.google.cloud.bigtable.admin.v2.BigtableTableAdminClient;
+import com.google.cloud.bigtable.admin.v2.models.CreateTableRequest;
+import com.google.cloud.bigtable.admin.v2.models.Table;
 ```
 
 Then, to create a table, use the following code:
 ```java
 InstanceName instanceName = InstanceName.of("my-project", "my-instance");
 
-TableAdminClient tableAdminClient = TableAdminClient.create(instanceName);
+BigtableTableAdminClient tableAdminClient = BigtableTableAdminClient.create(instanceName);
 
 try {
-    CreateTable createTableReq = TableAdminRequests.createTable("tableId")
+  Table createdTable = tableAdminClient.createTable(
+    CreateTableRequest.of("my-table")
       .addFamily("cf2", GCRULES.maxVersions(10));
-    client.createTable(createTableReq);
-
 } finally {
   tableAdminClient.close();
 }
--- a/google-cloud-clients/google-cloud-bigtable-admin/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClient.java
+++ b/google-cloud-clients/google-cloud-bigtable-admin/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClient.java
@@ -19,6 +19,7 @@ import com.google.api.core.ApiAsyncFunction;
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
+import com.google.api.gax.rpc.ApiExceptions;
 import com.google.api.resourcenames.ResourceName;
 import com.google.bigtable.admin.v2.AppProfileName;
 import com.google.bigtable.admin.v2.ClusterName;
@@ -46,9 +47,7 @@ import com.google.cloud.bigtable.admin.v2.stub.BigtableInstanceAdminStub;
 import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
-import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.MoreExecutors;
-import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.google.iam.v1.GetIamPolicyRequest;
 import com.google.iam.v1.SetIamPolicyRequest;
 import com.google.iam.v1.TestIamPermissionsRequest;
@@ -160,7 +159,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public Instance createInstance(CreateInstanceRequest request) {
-    return awaitFuture(createInstanceAsync(request));
+    return ApiExceptions.callAndTranslateApiException(createInstanceAsync(request));
   }
 
   /**
@@ -208,7 +207,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public Instance updateInstance(UpdateInstanceRequest request) {
-    return awaitFuture(updateInstanceAsync(request));
+    return ApiExceptions.callAndTranslateApiException(updateInstanceAsync(request));
   }
 
   /**
@@ -251,7 +250,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public Instance getInstance(String id) {
-    return awaitFuture(getInstanceAsync(id));
+    return ApiExceptions.callAndTranslateApiException(getInstanceAsync(id));
   }
 
   /**
@@ -302,7 +301,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public List<Instance> listInstances() {
-    return awaitFuture(listInstancesAsync());
+    return ApiExceptions.callAndTranslateApiException(listInstancesAsync());
   }
 
   /**
@@ -385,7 +384,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public void deleteInstance(String instanceId) {
-    awaitFuture(deleteInstanceAsync(instanceId));
+    ApiExceptions.callAndTranslateApiException(deleteInstanceAsync(instanceId));
   }
 
   /**
@@ -394,7 +393,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    * <p>Sample code:
    *
    * <pre>{@code
-   * ApiFuture<Void> deleteFuture = client.deleteInstance("my-instance");
+   * ApiFuture<Void> deleteFuture = client.deleteInstanceAsync("my-instance");
    * deleteFuture.get();
    * }</pre>
    */
@@ -434,7 +433,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public Cluster createCluster(CreateClusterRequest request) {
-    return awaitFuture(createClusterAsync(request));
+    return ApiExceptions.callAndTranslateApiException(createClusterAsync(request));
   }
 
   /**
@@ -478,7 +477,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public Cluster getCluster(String instanceId, String clusterId) {
-    return awaitFuture(getClusterAsync(instanceId, clusterId));
+    return ApiExceptions.callAndTranslateApiException(getClusterAsync(instanceId, clusterId));
   }
 
   /**
@@ -522,7 +521,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    *
    * <pre>{@code
    * try {
-   *   List<Cluster> clusters = cluster.listClusters("my-instance");
+   *   List<Cluster> clusters = client.listClusters("my-instance");
    * } catch (PartialListClustersException e) {
    *   System.out.println("The following zones are unavailable: " + e.getUnavailableZones());
    *   System.out.println("But the following clusters are reachable: " + e.getClusters())
@@ -531,7 +530,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public List<Cluster> listClusters(String instanceId) {
-    return awaitFuture(listClustersAsync(instanceId));
+    return ApiExceptions.callAndTranslateApiException(listClustersAsync(instanceId));
   }
 
   /**
@@ -610,12 +609,13 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    * <p>Sample code:
    *
    * <pre>{@code
-   * Cluster cluster = cluster.resizeCluster("my-instance", "my-cluster", 30);
+   * Cluster cluster = client.resizeCluster("my-instance", "my-cluster", 30);
    * }</pre>
    */
   @SuppressWarnings("WeakerAccess")
   public Cluster resizeCluster(String instanceId, String clusterId, int numServeNodes) {
-    return awaitFuture(resizeClusterAsync(instanceId, clusterId, numServeNodes));
+    return ApiExceptions.callAndTranslateApiException(
+        resizeClusterAsync(instanceId, clusterId, numServeNodes));
   }
 
   /**
@@ -623,7 +623,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    * a PRODUCTION instance can be resized.
    *
    * <pre>{@code
-   * ApiFuture<Cluster> clusterFuture = cluster.resizeCluster("my-instance", "my-cluster", 30);
+   * ApiFuture<Cluster> clusterFuture = client.resizeCluster("my-instance", "my-cluster", 30);
    * Cluster cluster = clusterFuture.get();
    * }</pre>
    */
@@ -662,7 +662,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public void deleteCluster(String instanceId, String clusterId) {
-    awaitFuture(deleteClusterAsync(instanceId, clusterId));
+    ApiExceptions.callAndTranslateApiException(deleteClusterAsync(instanceId, clusterId));
   }
 
   /**
@@ -713,7 +713,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public AppProfile createAppProfile(CreateAppProfileRequest request) {
-    return awaitFuture(createAppProfileAsync(request));
+    return ApiExceptions.callAndTranslateApiException(createAppProfileAsync(request));
   }
 
   /**
@@ -759,7 +759,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public AppProfile getAppProfile(String instanceId, String appProfileId) {
-    return awaitFuture(getAppProfileAsync(instanceId, appProfileId));
+    return ApiExceptions.callAndTranslateApiException(getAppProfileAsync(instanceId, appProfileId));
   }
 
   /**
@@ -808,7 +808,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public List<AppProfile> listAppProfiles(String instanceId) {
-    return awaitFuture(listAppProfilesAsync(instanceId));
+    return ApiExceptions.callAndTranslateApiException(listAppProfilesAsync(instanceId));
   }
 
   /**
@@ -911,7 +911,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public AppProfile updateAppProfile(UpdateAppProfileRequest request) {
-    return awaitFuture(updateAppProfileAsync(request));
+    return ApiExceptions.callAndTranslateApiException(updateAppProfileAsync(request));
   }
 
   /**
@@ -966,7 +966,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public void deleteAppProfile(String instanceId, String appProfileId) {
-    awaitFuture(deleteAppProfileAsync(instanceId, appProfileId));
+    ApiExceptions.callAndTranslateApiException(deleteAppProfileAsync(instanceId, appProfileId));
   }
 
   /**
@@ -1016,7 +1016,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public Policy getIamPolicy(String instanceId) {
-    return awaitFuture(getIamPolicyAsync(instanceId));
+    return ApiExceptions.callAndTranslateApiException(getIamPolicyAsync(instanceId));
   }
 
   /**
@@ -1083,7 +1083,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public Policy setIamPolicy(String instanceId, Policy policy) {
-    return awaitFuture(setIamPolicyAsync(instanceId, policy));
+    return ApiExceptions.callAndTranslateApiException(setIamPolicyAsync(instanceId, policy));
   }
 
   /**
@@ -1165,7 +1165,7 @@ public final class BigtableInstanceAdminClient implements AutoCloseable {
    * <p>Sample code:
    *
    * <pre>{@code
-   * ApiFuture<List<String>> grantedPermissionsFuture = client.testIamPermission("my-instance",
+   * ApiFuture<List<String>> grantedPermissionsFuture = client.testIamPermissionAsync("my-instance",
    *   "bigtable.tables.readRows", "bigtable.tables.mutateRows");
    *
    * ApiFutures.addCallback(grantedPermissionsFuture,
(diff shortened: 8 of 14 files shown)


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1105,15 +1335,24 @@ public Acl apply(ObjectAccessControl objectAccessControl) {
       builder.setNotFoundPage(website.getNotFoundPage());
     }
     if (bucketPb.getLifecycle() != null && bucketPb.getLifecycle().getRule() != null) {
-      builder.setDeleteRules(
```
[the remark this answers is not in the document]
JesseLovelace: DeleteRules is actually still set just a few lines down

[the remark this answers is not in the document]
frankyn: acking misread the diff.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1105,15 +1337,24 @@ public Acl apply(ObjectAccessControl objectAccessControl) {
...
                 @Override
-                public DeleteRule apply(Rule rule) {
-                  return DeleteRule.fromPb(rule);
+                public LifecycleRule apply(Rule rule) {
+                  return LifecycleRule.fromPb(rule);
                 }
               }));
+      builder.setDeleteRules(
+              transform(
+                      bucketPb.getLifecycle().getRule(),
```
chingor13: What is happening here when the lifecycle contains non delete rules?

JesseLovelace: The behavior for a non-delete rule is that a RawDeleteRule is created whose type is UNKNOWN, see DeleteRule's fromPb method.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,230 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+      public LifecycleCondition setAge(Integer age) {
+        this.age = age;
+        return this;
+      }
+
+      public DateTime getCreatedBefore() {
+        return createdBefore;
+      }
+
+      public LifecycleCondition setCreatedBefore(DateTime createdBefore) {
```
frankyn: Request to use bulders instead of a mutable class to follow existing classes.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -205,4 +212,21 @@ public void testDeleteRules() {
       assertEquals(delRule, DeleteRule.fromPb(delRule.toPb()));
     }
   }
+
+  @Test
+  public void testLifecycleRules() {
+    Rule deleteLifecycleRule = new LifecycleRule(new LifecycleRule.DeleteLifecycleAction(),
+            new LifecycleRule.LifecycleCondition.Builder().setAge(10).build()).toPb();
```
frankyn: Change the way a builder is created.
`newBuilder()` instead to follow along with:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L1044

Similarly if a developer wants to modify the state of the Rule, add `toBuilder()`:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L912

This will help remove the need to use `new`. Looks cleaner and consistent with the rest of the surface.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -361,17 +355,37 @@ public void testGetBucketEmptyFields() {
...
-    assertEquals(3, (int) lifecycleRule.getCondition().getNumberOfNewerVersions());
-    assertNotNull(lifecycleRule.getCondition().getCreatedBefore());
-    assertFalse(lifecycleRule.getCondition().getIsLive());
-    assertEquals(1, (int) lifecycleRule.getCondition().getAge());
-    assertEquals(1, lifecycleRule.getCondition().getMatchesStorageClass().size());
+    try {
+      assertTrue(
+          lifecycleRule.getAction()
+              instanceof BucketInfo.LifecycleRule.SetStorageClassLifecycleAction);
+      assertEquals(3, (int) lifecycleRule.getCondition().getNumberOfNewerVersions());
```
frankyn: I don't think type casting is necessary here.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -361,17 +355,37 @@ public void testGetBucketEmptyFields() {
...
-        lifecycleRule.getAction()
-            instanceof BucketInfo.LifecycleRule.SetStorageClassLifecycleAction);
-    assertEquals(3, (int) lifecycleRule.getCondition().getNumberOfNewerVersions());
-    assertNotNull(lifecycleRule.getCondition().getCreatedBefore());
-    assertFalse(lifecycleRule.getCondition().getIsLive());
-    assertEquals(1, (int) lifecycleRule.getCondition().getAge());
-    assertEquals(1, lifecycleRule.getCondition().getMatchesStorageClass().size());
+    try {
+      assertTrue(
+          lifecycleRule.getAction()
```
frankyn: This could also be reduced to:
`assertTrue(lifecycleRule.getAction().lifecycleAction().equals( BucketInfo.LifecycleRule.SetStorageClassLifecycleAction.TYPE));`

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -159,14 +158,9 @@ public static void beforeClass() throws IOException {
...
                     new BucketInfo.LifecycleRule(
-                        new BucketInfo.LifecycleRule.SetStorageClassLifecycleAction(
-                            StorageClass.NEARLINE),
-                        new BucketInfo.LifecycleRule.LifecycleCondition()
-                            .setNumberOfNewerVersions(3)
-                            .setCreatedBefore(new DateTime(System.currentTimeMillis()))
-                            .setAge(1)
-                            .setIsLive(false)
-                            .setMatchesStorageClass(ImmutableList.of(StorageClass.COLDLINE)))))
+                        new BucketInfo.LifecycleRule.DeleteLifecycleAction(),
```
frankyn: Add an import to reduce the long call for `LifecycleRule.SetStorageClassLifecycleAction` and `LifecycleRule.DeleteLifecycleAction`.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,242 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+      public static final String TYPE = "Delete";
+      private static final long serialVersionUID = -2050986302222644873L;
+
+      @Override
+      String getActionType() {
+        return TYPE;
+      }
+    }
+
+    public static class SetStorageClassLifecycleAction extends LifecycleAction {
```
frankyn: This should follow a builder pattern in case of changes made to this action. 

I don't expect changes to come to actions, but based on our offline discussion it would help here.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -205,4 +212,21 @@ public void testDeleteRules() {
...
+
+    assertEquals("Delete", deleteLifecycleRule.getAction().getType());
+    assertEquals(10, (long)deleteLifecycleRule.getCondition().getAge());
+
+    Rule setStorageClassLifecycleRule = new LifecycleRule(
+            new LifecycleRule.SetStorageClassLifecycleAction(StorageClass.COLDLINE),
+            new LifecycleRule.LifecycleCondition.Builder().setIsLive(true).setNumberOfNewerVersions(10).build()).toPb();
+
+    assertEquals("COLDLINE", setStorageClassLifecycleRule.getAction().getStorageClass());
```
frankyn: Use `StorageClass.COLDLINE`, I think the return type here should follow the same type used for `StorageClass.COLDLINE`.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -205,4 +212,21 @@ public void testDeleteRules() {
       assertEquals(delRule, DeleteRule.fromPb(delRule.toPb()));
     }
   }
+
+  @Test
+  public void testLifecycleRules() {
+    Rule deleteLifecycleRule = new LifecycleRule(new LifecycleRule.DeleteLifecycleAction(),
+            new LifecycleRule.LifecycleCondition.Builder().setAge(10).build()).toPb();
+
+    assertEquals("Delete", deleteLifecycleRule.getAction().getType());
```
frankyn: Use your enum types.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
JesseLovelace: It actually doesn't compile without that type casting, complains about ambiguous method call (object, object or long, long)

frankyn: Whoops, I had to verify. Use: `lifecycleRule.getCondition().getNumberOfNewerVersions().intValue()` to get value as an int.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -182,11 +182,11 @@ static LifecycleRule fromPb(Rule rule) {
 
       switch (action.getType()) {
         case DeleteLifecycleAction.TYPE:
-          lifecycleAction = new DeleteLifecycleAction();
+          lifecycleAction = LifecycleAction.newDeleteLifecycleAction();
```
frankyn: Naming: I think `newDeleteAction` is sufficient.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -182,11 +182,11 @@ static LifecycleRule fromPb(Rule rule) {
...
           break;
         case SetStorageClassLifecycleAction.TYPE:
           lifecycleAction =
-              new SetStorageClassLifecycleAction(StorageClass.valueOf(action.getStorageClass()));
+              LifecycleAction.newSetStorageClassLifecycleAction(StorageClass.valueOf(action.getStorageClass()));
```
frankyn: `newSetStorageClassAction`

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -215,17 +219,20 @@ public void testDeleteRules() {
...
     Rule setStorageClassLifecycleRule = new LifecycleRule(
-            new LifecycleRule.SetStorageClassLifecycleAction(StorageClass.COLDLINE),
-            new LifecycleRule.LifecycleCondition.Builder().setIsLive(true).setNumberOfNewerVersions(10).build()).toPb();
+            LifecycleAction.newSetStorageClassLifecycleAction(StorageClass.COLDLINE),
+            LifecycleCondition.newBuilder().setIsLive(true).setNumberOfNewerVersions(10).build()).toPb();
 
-    assertEquals("COLDLINE", setStorageClassLifecycleRule.getAction().getStorageClass());
+    assertEquals(StorageClass.COLDLINE, setStorageClassLifecycleRule.getAction().getStorageClass());
     assertTrue(setStorageClassLifecycleRule.getCondition().getIsLive());
     assertEquals(10, (long)setStorageClassLifecycleRule.getCondition().getNumNewerVersions());
```
frankyn: `intValue()` here as well.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -205,4 +216,24 @@ public void testDeleteRules() {
...
+  @Test
+  public void testLifecycleRules() {
+    Rule deleteLifecycleRule =
+        new LifecycleRule(
+                LifecycleAction.newDeleteLifecycleAction(),
+                LifecycleCondition.newBuilder().setAge(10).build())
+            .toPb();
+
+    assertEquals(LifecycleRule.DeleteLifecycleAction.TYPE, deleteLifecycleRule.getAction().getType());
+    assertEquals(10, (long)deleteLifecycleRule.getCondition().getAge());
```
frankyn: `intValue()`

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+                            }
+                          }));
+
+      return new LifecycleRule(lifecycleAction, conditionBuilder.build());
+    }
+
+    /**
+     * Condition for a Lifecycle rule, specifies under what criteria an Action should be executed
+     *
+     * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Object Lifecycle Management</a>
```
frankyn: Use: https://cloud.google.com/storage/docs/lifecycle#conditions

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
   private final Boolean retentionPolicyIsLocked;
   private final Long retentionPeriod;
 
+  /**
+   * Lifecycle rule for a bucket. Allows supported Actions, such as deleting and changing storage class,
+   * to be executed when certain Conditions are met
+   *
+   * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Object Lifecycle Management</a>
```
frankyn: Use: https://cloud.google.com/storage/docs/lifecycle#actions

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -177,7 +471,11 @@ static DeleteRule fromPb(Rule rule) {
    * Delete rule class that sets a Time To Live for blobs in the bucket.
    *
    * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Object Lifecycle Management</a>
+   *
+   * @deprecated Use a {@code LifecycleRule} with a {@code DeleteLifecycleAction} and use
+   * {@code LifecycleCondition.Builder.setAge} instead
```
frankyn: Can you add a small snippet in the documentation that demonstrates how to use this path? At least for this one.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -264,7 +566,7 @@ public long getTimeMillis() {
 
     @Override
     void populateCondition(Rule.Condition condition) {
-      condition.setCreatedBefore(new DateTime(timeMillis));
+      condition.setCreatedBefore(new DateTime(true, timeMillis, 0));
```
frankyn: Is this a breaking change?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -368,10 +678,19 @@ void populateCondition(Rule.Condition condition) {
...
+     * @deprecated Use {@code setLifecycleRules} instead
      */
+    @Deprecated
     public abstract Builder setDeleteRules(Iterable<? extends DeleteRule> rules);
 
+    /**
+     * Sets the bucket's lifecycle configuration as a number of lifecycle rules, consisting of an action and a
+     * condition.
+     *
+     * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Lifecycle Management</a>
```
frankyn: Object Lifecycle Management

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+
+      /** Creates a new {@code DeleteLifecycleAction}. Blobs that meet the Condition associated with this action will be
+       * deleted
+       */
+      public static DeleteLifecycleAction newDeleteAction() {
+        return new DeleteLifecycleAction();
+      }
+
+      /** Creates a new {@code SetStorageClassLifecycleAction}. Blobs that meet the Condition associated with this
+       * action will have their Storage Class changed to the one specified
```
frankyn: Change: `Blobs that meet the Condition associated with this action will have their Storage Class changed to the one specified`
To: `A Blob's storage class that meets the action's conditions will be changed to the specified storage class.`

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+      }
+    }
+
+    public static class SetStorageClassLifecycleAction extends LifecycleAction {
+      public static final String TYPE = "SetStorageClass";
+      private static final long serialVersionUID = -62615467186000899L;
+
+      private final StorageClass storageClass;
+
+      private SetStorageClassLifecycleAction(StorageClass storageClass) {
```
frankyn: Are you okay with changing this name to `SetStorageClassAction` similar with `DeleteAction` given the context?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+    }
+  }
+
   /**
    * Base class for bucket's delete rules. Allows to configure automatic deletion of blobs and blobs
    * versions.
    *
    * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Object Lifecycle Management</a>
+   *
+   * @deprecated Use a {@code LifecycleRule} with a {@code DeleteLifecycleAction} instead.
```
frankyn: Include LifecycleCondition as well because a user would need both to migrate.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -241,7 +539,11 @@ Rule toPb() {
    * Delete rule class for blobs in the bucket that have been created before a certain date.
    *
    * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Object Lifecycle Management</a>
+   *
+   * @deprecated Use a {@code LifecycleRule} with a {@code DeleteLifecycleAction} and use
+   * {@code LifecycleCondition.Builder.setCreatedBefore} instead
```
frankyn: Suggestion: "Use a {@code LifecycleRule} with an action {@code DeleteLifecycleAction} and a condition {@code LifecycleCondition.Builder.setCreateBefore} instead."
Request: Add a period to finish the sentence. **_Apply this to each condition deprecation message._**

You could include an additional snippet here or link back to one that you have already created.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -368,10 +678,19 @@ void populateCondition(Rule.Condition condition) {
     /**
      * Sets the bucket's lifecycle configuration as a number of delete rules.
      *
-     * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Lifecycle Management</a>
+     * @deprecated Use {@code setLifecycleRules} instead
```
frankyn: Could you add an example here as well on how to set a set of Rules?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -550,12 +871,20 @@ public Builder setNotFoundPage(String notFoundPage) {
       return this;
     }
 
+    /** @deprecated Use {@code setLifecycleRules} instead **/
```
frankyn: Use the {@code setLifecycleRules} method instead.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+     * Base class for the Action to take when a Lifecycle Condition is met. Specific Actions are expressed as subclasses
+     * of this class, accessed by static factory methods
+     */
+    public abstract static class LifecycleAction implements Serializable {
+      private static final long serialVersionUID = 5801228724709173284L;
+
+      public abstract String getActionType();
+
+      /** Creates a new {@code DeleteLifecycleAction}. Blobs that meet the Condition associated with this action will be
+       * deleted
```
frankyn: Add period.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+          return this;
+        }
+
+        /** Sets a list of Storage Classes ({@code StorageClass} objects) for a Blob to match for an Action to be executed **/
+        public Builder setMatchesStorageClass(List<StorageClass> matchesStorageClass) {
+          this.matchesStorageClass = matchesStorageClass;
+          return this;
+        }
+
+        /** Creates a {@code LifecycleCondition} object **/
```
frankyn: `Builds a new {@code LifecycleCondition} object.`

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+      public static class Builder {
+        private Integer age;
+        private DateTime createdBefore;
+        private Integer numberOfNewerVersions;
+        private Boolean isLive;
+        private List<StorageClass> matchesStorageClass;
+
+        private Builder() {}
+
+        /** Sets the age in days a Blob should be for an Action to be executed **/
```
frankyn: `Sets the age in days. This condition is satisfied when a Blob reaches the specified age (in days). When you specify the Age condition, you are specifying a Time to Live (TTL) for objects in a bucket with lifecycle management configured. The time when the Age condition is considered to be satisfied is calculated by adding the specified value to the object creation time.`

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+        public Builder setAge(Integer age) {
+          this.age = age;
+          return this;
+        }
+
+        /** Sets the date a Blob should be created before for an Action to be executed. Note that only the date
+         * will be considered, if the time is specified it will be truncated **/
```
frankyn: Append `This condition is satisfied when an object is created before midnight of the specified date in UTC.`

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+        public Builder setCreatedBefore(DateTime createdBefore) {
+          this.createdBefore = createdBefore;
+          return this;
+        }
+
+        /** Sets the number of newer versions a Blob should have for an Action to be executed **/
```
frankyn: Append `Relevant only when versioning is enabled on a bucket.`

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+        public Builder setNumberOfNewerVersions(Integer numberOfNewerVersions) {
+          this.numberOfNewerVersions = numberOfNewerVersions;
+          return this;
+        }
+
+        /** Sets whether a Live Blob meets the Condition for an Action to be executed. If this is set to {@code false},
+         * the condition is met by Archived Blobs.
+         */
```
frankyn: Sets a isLive Boolean condition. If the value is true, this lifecycle condition matches only live Blobs; if the value is false, it matches only archived objects. For the purposes of this condition, Blobs in non-versioned buckets are considered live.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+        public Builder setIsLive(Boolean live) {
+          this.isLive = live;
+          return this;
+        }
+
+        /** Sets a list of Storage Classes ({@code StorageClass} objects) for a Blob to match for an Action to be executed **/
```
frankyn: Sets a list of Storage Classes for a objects that satisfy the condition to execute the Action.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+        /** Creates a {@code LifecycleCondition} object **/
+        public LifecycleCondition build() {
+          return new LifecycleCondition(this);
+        }
+      }
+    }
+
+    /**
+     * Base class for the Action to take when a Lifecycle Condition is met. Specific Actions are expressed as subclasses
+     * of this class, accessed by static factory methods
```
frankyn: Add a period.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+      public static DeleteLifecycleAction newDeleteAction() {
+        return new DeleteLifecycleAction();
+      }
+
+      /** Creates a new {@code SetStorageClassLifecycleAction}. Blobs that meet the Condition associated with this
+       * action will have their Storage Class changed to the one specified
+       *
+       * @param storageclass The {@code StorageClass} to set Blobs that meet the condition to
+       */
+      public static SetStorageClassLifecycleAction newSetStorageClassAction(StorageClass storageclass) {
```
frankyn: Camel case `storageClass`

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,302 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+       * deleted
+       */
+      public static DeleteLifecycleAction newDeleteAction() {
+        return new DeleteLifecycleAction();
+      }
+
+      /** Creates a new {@code SetStorageClassLifecycleAction}. Blobs that meet the Condition associated with this
+       * action will have their Storage Class changed to the one specified
+       *
+       * @param storageclass The {@code StorageClass} to set Blobs that meet the condition to
```
frankyn: The new storage class to use when conditions are met for this action.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
JesseLovelace: No, this is a bug fix. It's actually required that it be dateOnly because the time field can't be parsed. Without this, created before never actually works.

frankyn: Thanks for clarifying! Acking.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
JesseLovelace: Given that SetStorageClassLifecycleAction is instantiated via LifecycleAction.newSetStorageClassAction(), the long name shouldn't actually have a negative impact so I prefer keeping it as descriptive as possible. But if that's something you strongly disagree with I wouldn't mind changing it.

frankyn: Let's move forward with existing design. Thanks for the feedback!

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -471,9 +481,10 @@ static DeleteRule fromPb(Rule rule) {
    * Delete rule class that sets a Time To Live for blobs in the bucket.
    *
    * @see <a href="https://cloud.google.com/storage/docs/lifecycle">Object Lifecycle Management</a>
-   *
-   * @deprecated Use a {@code LifecycleRule} with a {@code DeleteLifecycleAction} and use
-   * {@code LifecycleCondition.Builder.setAge} instead
+   * @deprecated Use a {@code LifecycleRule} with a {@code DeleteLifecycleAction} and use {@code
```
frankyn: Could you use `<p>` to space the javadoc a bit.

Example:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L704

Render:
https://googleapis.github.io/google-cloud-java/google-cloud-clients/apidocs/com/google/cloud/storage/BucketInfo.html#versioningEnabled--

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,312 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
+                            public StorageClass apply(String storageClass) {
+                              return StorageClass.valueOf(storageClass);
+                            }
+                          }));
+
+      return new LifecycleRule(lifecycleAction, conditionBuilder.build());
+    }
+
+    /**
+     * Condition for a Lifecycle rule, specifies under what criteria an Action should be executed
```
frankyn: Missing period.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -93,12 +97,312 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
   private final Boolean retentionPolicyIsLocked;
   private final Long retentionPeriod;
 
+  /**
+   * Lifecycle rule for a bucket. Allows supported Actions, such as deleting and changing storage class,
+   * to be executed when certain Conditions are met
```
frankyn: period.

PR #3800, opened 2018-10-10
Bigtable: Add createUnsafe factory method in Mutation model

ajaaym:
Add createUnsafe factory method in Mutation model.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
@@ -57,8 +57,6 @@ public final class Mutation implements MutationApi<Mutation>, Serializable {
 
   /**
    * Creates new instance of Mutation object.
-   *
-   * @return Mutation object.
    */
   public static Mutation create() {
     return new Mutation(false);
@@ -69,8 +67,6 @@ public final class Mutation implements MutationApi<Mutation>, Serializable {
    * side timestamp. This is dangerous because mutations will no longer be idempotent, which
    * might cause multiple duplicate values to be stored in Bigtable. This option should only
    * be used for advanced usecases with extreme care.
-   *
-   * @return Mutation object.
    */
   @BetaApi
   public static Mutation createUnsafe() {


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -44,17 +44,29 @@
...
   private long byteSize;
 
   public static Mutation create() {
-    return new Mutation();
+    return new Mutation(false);
   }
 
-  private Mutation() {}
+  /**
+   * Creates Mutation object which allows setCell operation to set server side timestamp.
```
igorbernstein2: Please make the docs a bit more scary:

`Creates Mutation object which allows setCell operation use set server side timestamp. This is dangerous because mutation will no longer be idempotent, which might cause multiple duplicate values to be stored in Bigtable. This option should only be used for advanced usecases with extreme care.`

Also, please add a `@BetaApi` to discourage usage

igorbernstein2: Also, please add docs for the regular create

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -102,7 +114,9 @@ public Mutation setCell(
     Validations.validateFamily(familyName);
     Preconditions.checkNotNull(qualifier, "qualifier can't be null.");
     Preconditions.checkNotNull(value, "value can't be null.");
-    Preconditions.checkArgument(timestamp != -1, "Serverside timestamps are not supported");
+    if (!allowServersideTimestamp) {
+      Preconditions.checkArgument(timestamp != -1, "Serverside timestamps are not supported");
```
igorbernstein2: Please extract -1 into a package protected constant

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/MutationTest.java
```diff
@@ -97,6 +97,19 @@ public void setCellTest() {
...
   }
 
+  @Test
+  public void setCellWithServerSideTimestamp() {
+    Mutation mutation = Mutation.createUnsafe();
+    mutation
+        .setCell(
+            "fake-family",
+            ByteString.copyFromUtf8("fake-qualifier"),
+            -1,
```
igorbernstein2: Please use a constant

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -43,18 +44,42 @@
...
   private transient ImmutableList.Builder<com.google.bigtable.v2.Mutation> mutations =
       ImmutableList.builder();
 
   private int numMutations;
   private long byteSize;
 
+  /**
+   * Creates new instance of Mutation object.
+   *
+   * @return Mutation object.
```
igorbernstein2: I've been avoiding adding `@return` & `@param` annotations unless they provide any addition info. Please remove it here and below

ajaaym: done. Thanks.

PR #3813, opened 2018-10-12
Add assumptions to storage integration tests

JesseLovelace:
Fixes two issues:

-Tests relating to signing URLs are now properly skipped when a credential is provided that is incapable of signing
-Test accessing a public bucket is skipped if ran in an environment with restricted access


google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1657,7 +1668,14 @@ public void testDownloadPublicBlobWithoutAuthentication() {
     String landsatBucket = "gcp-public-data-landsat";
     String landsatPrefix = "LC08/PRE/044/034/LC80440342016259LGN00/";
     String landsatBlob = landsatPrefix + "LC80440342016259LGN00_MTL.txt";
-    byte[] bytes = unauthorizedStorage.readAllBytes(landsatBucket, landsatBlob);
+    byte[] bytes;
+    try {
+      bytes = unauthorizedStorage.readAllBytes(landsatBucket, landsatBlob);
+    } catch (StorageException ex) {
+      assertTrue(ex.getCause() instanceof GoogleJsonResponseException);
+      assumeFalse(403 == ((GoogleJsonResponseException) ex.getCause()).getStatusCode());
```
frankyn: I think this could be reduced down to:
`assumeFalse(403 == ex.getCode());`

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1657,7 +1668,14 @@ public void testDownloadPublicBlobWithoutAuthentication() {
     String landsatBucket = "gcp-public-data-landsat";
     String landsatPrefix = "LC08/PRE/044/034/LC80440342016259LGN00/";
     String landsatBlob = landsatPrefix + "LC80440342016259LGN00_MTL.txt";
-    byte[] bytes = unauthorizedStorage.readAllBytes(landsatBucket, landsatBlob);
+    byte[] bytes;
+    try {
+      bytes = unauthorizedStorage.readAllBytes(landsatBucket, landsatBlob);
+    } catch (StorageException ex) {
+      assertTrue(ex.getCause() instanceof GoogleJsonResponseException);
```
frankyn: I don't think this is necessary if the line below is reduced.

PR #3825, opened 2018-10-16
setMaxInboundMessageSize in Pubsub example

stephenmw:
Pub/Sub may return up to 20MB of data from a single pull request. The default of 4MB is not enough for even one message of maximum size (10MB).


stephenmw:
Note: this is related to b/117792383

PR #3828, opened 2018-10-16
Make testDownloadPublicBlobWithoutAuthentication skippable with an environment variable

JesseLovelace:
testDownloadPublicBlobWithoutAuthentication will never pass when running in a restricted VPC, so this gives the option to skip to skip the test by setting the environment variable GOOGLE_CLOUD_TESTS_IN_VPCSC to "true".

This replaces the solution of skipping the test if there's any 403, since that could be a genuine error.


google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1661,20 +1661,16 @@ public void testGetBlobs() {
 
   @Test
   public void testDownloadPublicBlobWithoutAuthentication() {
+    assumeFalse(System.getenv("GOOGLE_CLOUD_TESTS_IN_VPCSC").equalsIgnoreCase("true"));
```
frankyn: I think this LGTM. My only request is that you make this available to future tests. Could you create a separate class member variable with this value which a test can check against?

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -143,6 +142,7 @@
   private static final String KMS_KEY_RING_LOCATION = "us";
   private static final String KMS_KEY_ONE_NAME = "gcs_kms_key_one";
   private static final String KMS_KEY_TWO_NAME = "gcs_kms_key_two";
+  private static final boolean isVpcTest = System.getenv("GOOGLE_CLOUD_TESTS_IN_VPCSC").equalsIgnoreCase("true");
```
frankyn: Constant casing `IS_VPC_TEST`

PR #3830, opened 2018-10-17
Add compute synth

chingor13:
Adds compute library synth file which uses the discogapic generator.

Added to synthtool in https://github.com/googleapis/synthtool/releases/tag/synthtool-0.10.0

### DIFF
--- a/google-cloud-clients/google-cloud-compute/synth.py
+++ b/google-cloud-clients/google-cloud-compute/synth.py
@@ -18,7 +18,6 @@ import synthtool as s
 import synthtool.gcp as gcp
 
 gapic = gcp.DiscoGAPICGenerator()
-common_templates = gcp.CommonTemplates()
 
 library = gapic.java_library(
     service='compute',


andreamlin:
Has this been tested?

chingor13:
These are only really testable by cd'ing to the directory and running `python3 synth.py` (assuming [synthtool](https://github.com/googleapis/synthtool) is already installed). This has been tested locally and there is a documentation change pending which will appear tomorrow assuming this is merged.

google-cloud-clients/google-cloud-compute/synth.py
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This script is used to synthesize generated parts of this library."""
+
+import synthtool as s
+import synthtool.gcp as gcp
+
+gapic = gcp.DiscoGAPICGenerator()
+common_templates = gcp.CommonTemplates()
```
andreamlin: Is this variable used somewhere?

chingor13: Good catch, it's not needed. Will remove.

PR #3831, opened 2018-10-17
Update CODEOWNERS.

chingor13:
- Remove @vam-google
- Remove @andreamlin from all but Compute
- Fix path prefix


chingor13:
cc @vam-google as they somehow don't have write access so I can't request a review.

vam-google:
@chingor13 that is weird... I created this CODEOWNERS file. Are  you sure I don't have the permissions? If I do not, then I was somehow removed by mistake. Pleaese add me back

chingor13:
@vam-google Permissions were all weird after this repo moved orgs.

PR #3840, opened 2018-10-19
bigtable emulator

igorbernstein2:
This allows customers to easily interact with the bigtable emulator. Currently the Bigtable emulator is implemented in go. It is a fairly complex bit of code that would be very difficult to replicate in java. This makes it very difficult for users to interact with it in a java project. This PR attempts to make things easier for the user by wrapping the golang emulator in a jar and providing a JUnit rule to simplify lifecycle management of the emulator.

It's implemented as 2 artifacts:
- google-cloud-gcloud-maven-plugin: a maven plugin that can download artifacts from the gcloud sdk repo and add them as resources to a jar. This avoids the need to check in binaries into google-cloud-java and makes sure that the binaries are up to date.
- google-cloud-bigtable-emulator: This uses the maven plugin to bundle the bigtable emulator in a jar and adds a java wrapper around the binary. The wrapper was extracted from the existing bigtable client integration tests: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/Emulator.java

### DIFF
--- a/google-cloud-testing/google-cloud-gcloud-maven-plugin/src/main/java/com/google/cloud/Component.java
+++ b/google-cloud-testing/google-cloud-gcloud-maven-plugin/src/main/java/com/google/cloud/Component.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.google.cloud;
 
 import com.google.gson.JsonElement;


igorbernstein2:
@sduskis can you take a peek and let me know what you think?

sduskis:
@igorbernstein2, can you please resolve the conflict?

igorbernstein2:
I'm updating the PR now

google-cloud-testing/google-cloud-bigtable-emulator/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-gcloud-maven-plugin</artifactId>
+        <version>0.67.1-alpha-SNAPSHOT</version><!-- {x-version-update:google-cloud-gcloud-maven-plugin:current} -->
+
+        <executions>
+          <execution>
+            <id>gen-sources</id>
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>download</goal>
```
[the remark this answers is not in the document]
igorbernstein2: It caches both the manifest and the binaries. The manifest is cached for 2 hours and the binaries are only downloaded if their checksums changed. So this is pretty lightweight to include in every build

igorbernstein2:
Updated:
- resolved merge conflicts
- use grpc stubs in tests
- add READMEs & javadocs
- Clean up shutdown hooks
- lazily construct channels
- add state guards
- support offline mode

PTAL

google-cloud-testing/google-cloud-bigtable-emulator/README.md
```diff
@@ -0,0 +1,181 @@
...
+        .setCredentialsProvider(NoCredentialsProvider.create())
+        .setTransportChannelProvider(
+            FixedTransportChannelProvider.create(
+                GrpcTransportChannel.create(bigtableEmulator.getAdminChannel())
+            )
+        );
+    tableAdminClient = BigtableTableAdminClient.create(tableAdminSettings.build());
+
+    BigtableDataSettings.Builder dataSettings = BigtableDataSettings.newBuilder()
+        .setInstanceName(com.google.cloud.bigtable.data.v2.models.InstanceName.of("fake-project", "fake-instance"))
```
emmmile: I wonder if this can be added as default way to get a data client to `BigtableEmulatorRule`, something like `getBigtableDataClient`. The same for the table admin client above.

emmmile:
Awesome!

google-cloud-testing/google-cloud-bigtable-emulator/README.md
igorbernstein2: I would like to use this wrapper for tests in the bigtable client, which would create a circular dependency: bigtable client depends on the emulator for tests and the emulator depends on the client to construct it.
I'm considering adding the emulator as an optional dependency to the client and then adding a helper to construct the clients for the emulator. But I need to do a bit more experimentation.

google-cloud-testing/google-cloud-gcloud-maven-plugin/src/main/java/com/google/cloud/Component.java
```diff
@@ -0,0 +1,78 @@
+package com.google.cloud;
```
sduskis: Header, please

google-cloud-testing/google-cloud-gcloud-maven-plugin/src/main/java/com/google/cloud/Component.java
```diff
@@ -0,0 +1,36 @@
...
+  static Component create(String id, String checksum, URL source, String fileType) {
+    return new AutoValue_Component(id, checksum, source, fileType);
+  }
+
+  static Component fromJson(URL baseUrl, JsonObject componentObj) throws IOException {
+    String id = componentObj.get("id").getAsString();
+
+    JsonElement data = componentObj.get("data");
+    if (data == null) {
+      throw new NullPointerException("Component " + id + " is missing a data section");
```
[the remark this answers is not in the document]
sduskis: Precondition?

google-cloud-testing/google-cloud-gcloud-maven-plugin/src/main/java/com/google/cloud/Component.java
igorbernstein2: oops, fixed

google-cloud-testing/google-cloud-gcloud-maven-plugin/src/main/java/com/google/cloud/Component.java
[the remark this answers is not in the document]
igorbernstein2: I don't want to add a guava dep just for this

igorbernstein2:
Thanks for reviewing! I'll follow up with client changes we discussed next week

PR #3851, opened 2018-10-24
Memorystore for Redis is now supported @ GA level

fewtrell:


danoscarmike:
@fewtrell while it is true that the backend service is GA -- the clients are typically versioned independently and we have some standard quality checks to make before we promote the client to a GA release.

fewtrell:
acknowledged - sorry to bother you.

PR #3858, opened 2018-10-25
Update CONTRIBUTING.md

Neeshmacpaul:


chingor13:
Thanks, but the CLA is not signed and that header is intended to be the title.

PR #3863, opened 2018-10-26
Add a traceSampled field to the google-cloud-logging LogEntry class.

sebright2:
The new field was added to the Stackdriver LogEntry protocol buffer message in
https://github.com/googleapis/googleapis/commit/7b34e5a3ed10775fe14d15ffa842870dc3b5400e#diff-a51916a680df6453b6d86586cd7d63f9R149.

______________________________________________________________________________

This change would allow [opencensus-contrib-log-correlation-stackdriver](https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/stackdriver) to set the trace_sampled field in its LoggingEnhancer.

/cc @g-easy


google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/LogEntry.java
```diff
@@ -505,6 +529,7 @@ private static Long millisFromTimestamp(Timestamp timestamp) {
     if (spanId != null) {
       builder.setSpanId(spanId);
     }
+    builder.setTraceSampled(traceSampled);
```
pongad: Do we need to null check this?

sebright: The protocol buffer LogEntry.Builder's traceSampled field is a boolean ([link](https://github.com/googleapis/google-cloud-java/blob/f69e7fafe748bc8ce49c2a8d9375b3b3a4644d5d/google-api-grpc/proto-google-cloud-logging-v2/src/main/java/com/google/logging/v2/LogEntry.java#L4141)), so I thought that it wouldn't be necessary to distinguish between false and not set.  The traceSampled field in this class is also a boolean.

pongad: Gah you're right. Sorry I have been working with Apiary for too long!

g-easy:
Consider adding wording to the description along the lines of "so that the backend can tell the difference between sampled and non-sampled traces, to e.g. link from the log view to the trace view for them"

sebright:
I tried to keep the description short, since I think the protocol buffer is the source of truth.  
The descriptions for `getTrace` and `getSpanId` are also relatively short.  Would it be better to copy the field description from the LogEntry proto?

g-easy:
Up to you. I'm fine with keeping it short.

sebright:
Thanks!

PR #3773, opened 2018-10-04
walkFiles consistently relative or absolute

jean-philippe-martin:
The expected behavior from Files.walkFileTree is that if the starting path is absolute, then all the visited paths are absolute. If the starting path is relative, then all the visited paths are relative.

This PR adds a test for this behavior, and shows that our current code
fails for an absolute path: it returns a mix of relative and absolute
paths.

This PR also adds a code change to fix the problem, so the actual
behavior matches what is expected.

This should fix issue #3772

PR #3869, opened 2018-10-26
Put new retry options in CloudStorageConfiguration

### LINKED ISSUE #3868: google-cloud-nio: implement configurable retries/reopens
google-cloud-nio currently will retry and even reopen channels on a number of conditions (if turned on in configuration). It would be nice to be able to specify which exceptions to reopen for, and which HTTP codes to retry on (in addition to the codes that already have `.isRetriable()` return true).

jean-philippe-martin:
Add `retriableHttpCodes` and `reopenableExceptions` so the user can, if they choose, customize the conditions under which we retry an access or reopen a channel after an exception closed it.

This can be handy if a cloud provider starts transiently throwing exceptions that would normally indicate a fatal condition, but which in practice will pass if retried.

The backoff logic in retries / reopens in unmodified.

Also update `testListFilesInRootDirectory` to reflect the new reality that `Files.newDirectoryStream` correctly returns absolute or relative paths depending on its input (as of #3773).

Fixes #3868


jean-philippe-martin:
Rebased to catch up with latest changes to master.

jean-philippe-martin:
Rebased again to catch up with the latest changes in master.

jean-philippe-martin:
Github's saying again that we're out-of-date with master, but I'll let this as-is until after the review, so the code is stable for someone to review.

jean-philippe-martin:
That integration test failure was with Spanner, which seems unrelated to this change. Is there a way to rerun the checks?

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -34,26 +30,30 @@
   private long totalWaitTime; // in milliseconds
   private final int maxRetries;
   private final int maxReopens;
+  private final CloudStorageConfiguration config;
 
   /**
    * Create a CloudStorageRetryHandler with the maximum retries and reopens set to the same value.
    *
-   * @param maxRetriesAndReopens value for both maxRetries and maxReopens
```
chingor13: This change looks like a breaking API change. Is this intentional?

We can add the new constructor, deprecate this one and then remote the deprecated one later?

jean-philippe-martin: You're right, this is a public method so it's part of the API (even though I don't know of anyone using this class directly). I've added the old one back as you suggested, marked as deprecated.

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -34,26 +30,45 @@
...
+    this.config = CloudStorageConfiguration.DEFAULT;
+  }
 
   /**
    * Create a CloudStorageRetryHandler with the maximum retries and reopens set to the same value.
    *
-   * @param maxRetriesAndReopens value for both maxRetries and maxReopens
+   * @param config - configuration for reopens and retryable codes.
    */
-  public CloudStorageRetryHandler(final int maxRetriesAndReopens) {
```
chingor13: We're still removing this constructor with a single int used for both retries and reopens.

chingor13: We should put this one back and also mark it as deprecated.

jean-philippe-martin: Oh you're right, thank you for catching that! I put it back, deprecated.

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -34,26 +30,60 @@
...
+  }
+
+  /**
+   * Create a CloudStorageRetryHandler with the maximum retries and reopens set to different values.
+   *
+   * @param maxRetries maximum number of retries
+   * @param maxReopens maximum number of reopens
+   * @param config http codes we'll retry on, and exceptions we'll reopen on.
+   */
+  public CloudStorageRetryHandler(final int maxRetries, final int maxReopens, final CloudStorageConfiguration config) {
```
chingor13: Do we need this one if you can set the maxRetries and maxReopens from the config?

jean-philippe-martin: We need this one because the config only holds maxReopens, not maxRetries. The maxReopen option there lets us override the setting in the config.

chingor13: Ah, I missed that. Does it make sense to add that configuration option to the CloudStorageConfiguration?

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -34,26 +30,62 @@
...
+
 
   /**
    * Create a CloudStorageRetryHandler with the maximum retries and reopens set to the same value.
    *
    * @param maxRetriesAndReopens value for both maxRetries and maxReopens
+   *
+   * @deprecated use CloudStorageRetryHandler(CloudStorageConfiguration) instead.
    */
+  @java.lang.Deprecated
```
chingor13: Import this?

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -34,26 +30,62 @@
...
-  /**
+    /**
    * Create a CloudStorageRetryHandler with the maximum retries and reopens set to different values.
    *
    * @param maxRetries maximum number of retries
    * @param maxReopens maximum number of reopens
+   *
+   * @deprecated use CloudStorageRetryHandler(CloudStorageConfiguration) instead.
    */
+  @java.lang.Deprecated
```
chingor13: Ditto, import this?

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
```diff
@@ -34,26 +30,62 @@
...
   public CloudStorageRetryHandler(final int maxRetriesAndReopens) {
     this.maxRetries = maxRetriesAndReopens;
     this.maxReopens = maxRetriesAndReopens;
+    // we're just using the retry parameters from the config, so it's OK to have a default.
+    this.config = CloudStorageConfiguration.DEFAULT;
   }
 
-  /**
+    /**
```
chingor13: nit: extra indentation

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageRetryHandler.java
jean-philippe-martin: That would make for a much bigger change than we were aiming for here. The reopen can be specified each time we open a connection (in the public API), so we'd have to createa a new config object just to carry this parameter to the retry handler.

PR #3856, opened 2018-10-25
Bigtable: add sync methods

elisheva-qlogic:
Fixes #3635 
Sync methods added in BigtableDataClient

PR #3870, opened 2018-10-29
Bigtable: Add integration test to test nonexistent row handling

igorbernstein2:


google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/ReadIT.java
```diff
@@ -111,6 +117,31 @@ public void read() throws Throwable {
     assertThat(observer.responses).containsExactlyElementsIn(expectedRows);
   }
 
+  @Test
+  public void readSingleNonexistentAsyncCallback() throws Exception {
+    ApiFuture<Row> future = testEnvRule.env().getDataClient()
+        .readRowAsync(testEnvRule.env().getTableName().getTable(), "somenonexistentkey");
+
+    final AtomicBoolean found = new AtomicBoolean();
+    final CountDownLatch latch = new CountDownLatch(1);
```
sduskis: why do you need both a `CountDownLatch` and an `AtomicBoolean`? Can you combine them or use `future.get()`?

igorbernstein2: I was trying to reproduce this issue:
https://github.com/googleapis/google-cloud-java/pull/3856#discussion_r228957914

Issue #3728, opened 2018-09-25
Don't bundle tests in jars


elharo:
I downloaded and unzipped google-cloud-dns-0.63.0-alpha-tests.jar. Here's what I found inside it. We're definitely bundling the actual tests into these jars. We either shouldn't do this (just include mocks and the like) or otherwise not put these in the BOM.

```
$ jar tvf google-cloud-dns-0.63.0-alpha-tests.jar
     0 Thu Sep 20 16:47:10 EDT 2018 META-INF/
   617 Thu Sep 20 16:47:10 EDT 2018 META-INF/MANIFEST.MF
     0 Thu Sep 20 16:47:10 EDT 2018 com/
     0 Thu Sep 20 16:47:10 EDT 2018 com/google/
     0 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/
     0 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/
     0 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/it/
 19805 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/DnsImplTest.class
  4099 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/ProjectInfoTest.class
 46519 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/it/ITDnsTest.class
  8991 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/SerializationTest.class
  4908 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/RecordSetTest.class
  3820 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/DnsExceptionTest.class
  7528 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/ChangeRequestInfoTest.class
  2639 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/OptionTest.class
   879 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/OptionTest$5.class
  3088 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/DnsBatchResultTest.class
  5755 Thu Sep 20 16:47:10 EDT 2018 com/google/cloud/dns/ZoneInfoTest.class
[...]

PR #3874, opened 2018-10-30
Fixes cleanup issues

### LINKED ISSUE #3707: test-jar in google-cloud-bom
The BOM contains many dependencies like this one:

```
      <dependency>
        <version>1.44.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-storage:current} -->
        <type>test-jar</type>
      </dependency>
```

Is this really something we want in the BOM?  I can't imagine clients should be depending on our tests. Can we take these out?

pmakani:
Fixes

#3707 test-jar in google-cloud-bom.

#3728 Don't bundle tests in jars.


sduskis:
@chingor13, do you have any issues with this?  @elharo, does this look right to you?

PR #3885, opened 2018-10-31
New Client: Cloud Security Command Center

chingor13:


chingor13:
This requires upstream proto changes and will need to be re-run. Keeping the issue open to stop autosynth from recreating the same PR.

PR #3909, opened 2018-11-06
Logging: close() should flush logs

### LINKED ISSUE #3866: com.google.cloud.logging.logback.LoggingAppender.close doesn't flush logging
It makes google-cloud-java unusable with asynchronous logging (DelayingShutdownHook doesn't work).

SergeyZadorozhniy:
https://github.com/googleapis/google-cloud-java/issues/3866
Added flash inside close stage.

Fixes #3866


google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -652,7 +652,11 @@ public void close() throws Exception {
       return;
     }
     closed = true;
-    rpc.close();
+    try {
+      flush();
```
chingor13: Won't this potentially block and prevent the rpc connection from closing. This is demonstrated by the test - it's blocking until flush succeeds.

sduskis: @chingor13, what should the behavior be for `close()` without a previous `flush()`?  In Cloud Bigtable's custom client, we always flush first, which I think meets the expectation of the user.   However, I do understand that there isn't a manual flush for Logging, which changes the dynamics; IIUC, Logging will wait for some time induced flushing.

@igorbernstein2 is working on manual flush() mechanism so that Cloud Bigtable's use of flush works as expected; hopefully, that will allow a fix for other implementations, like logging.

sduskis:
We need major changes before this can be merged.  There's no reason to keep this open at this point.

PR #3911, opened 2018-11-06
Add helpers to convert core Date class with java.util.Date class

### LINKED ISSUE #3157: com.google.cloud.Date has no interop w/ java.util.Date

pmakani:
Fixes #3157
Add methods `toDate`, `fromDate`

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
+++ b/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
@@ -71,7 +71,7 @@ public final class Date implements Comparable<Date>, Serializable {
   }
 
   /**
-   * Set the Google Date.
+   * Convert a Google Date to a Java Util Date.
    *
    * @param date the date of the Google Date.
    * @return java.util.Date
@@ -88,7 +88,7 @@ public final class Date implements Comparable<Date>, Serializable {
   }
 
   /**
-   * Set the Java Util Date.
+   * Convert a Java Util Date to a Google Date.
    *
    * @param date the date of the java.util.Date
    * @return Google Java Date


google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
```diff
@@ -68,7 +70,39 @@ public static Date parseDate(String date) {
     int dayOfMonth = Integer.parseInt(matcher.group(3));
     return new Date(year, month, dayOfMonth);
   }
+  
+  /**
+  * @param date
+  * @return LocalDate
+  */
+  public static LocalDate toDate(Date date){
+	return LocalDate.of(date.year, date.month, date.dayOfMonth);
```
elharo: Google uses two space indents

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
```diff
@@ -68,7 +70,39 @@ public static Date parseDate(String date) {
     int dayOfMonth = Integer.parseInt(matcher.group(3));
     return new Date(year, month, dayOfMonth);
   }
+  
+  /**
+  * @param date
+  * @return LocalDate
```
elharo: Is this already used somewhere? If not it exposes a new third party dependency on the API surface, which we really want to avoid: 

https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-2.md

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
```diff
@@ -68,7 +70,39 @@ public static Date parseDate(String date) {
...
+  */
+  public static LocalDate toDate(Date date){
+	return LocalDate.of(date.year, date.month, date.dayOfMonth);  
+  }
+   
+  /**
+  * @param date
```
elharo: javadoc comments should be filled in

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
chingor13: Let's remove the LocalDate helpers.

garrettjonesgoogle: threetenbp is already exposed on the surface of google-cloud-core, e.g. `Duration`: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/RetryOption.java#L23

threetenbp is specifically used throughout google-cloud-java in order to have a Java 8-like surface for date stuff while still staying on Java 7.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
```diff
@@ -68,7 +70,39 @@ public static Date parseDate(String date) {
     int dayOfMonth = Integer.parseInt(matcher.group(3));
     return new Date(year, month, dayOfMonth);
   }
+  
+  /**
+  * @param date
+  * @return LocalDate
+  */
+  public static LocalDate toDate(Date date){
```
garrettjonesgoogle: `Date::toDate` feels odd. I would suggest `toLocalDate`.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
```diff
@@ -68,7 +70,39 @@ public static Date parseDate(String date) {
...
+  */
+  public static Date toDate(LocalDate date){
+    return new Date(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
+  }
 
+  /**
+  * @param date
+  * @return LocalDate
+  */
+  public static LocalDate fromDate(Date date){
```
garrettjonesgoogle: `Date::fromDate` feels odd when it is constructing a `LocalDate`. The name should be different.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
elharo: This is a one-liner. Does it carry its weight? It can be inlined almost trivially.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
```diff
@@ -68,7 +70,39 @@ public static Date parseDate(String date) {
...
+  public static LocalDate toDate(Date date){
+	return LocalDate.of(date.year, date.month, date.dayOfMonth);  
+  }
+   
+  /**
+  * @param date
+  * @return Date
+  */
+  public static Date toDate(LocalDate date){
+    return new Date(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
```
elharo: also a one liner

PR #3923, opened 2018-11-08
Spanner: use WatchdogProvider

yihanzhen:
Switch from `WatchdogInterceptor` to `WatchdogProvider`, which is one of the shared tools implemented in gax-java. This is non-breaking change and comes with the gapic migration work, as we aim to have spanner maximize the use of shared tools after the migration.

`WatchdogInterceptor` was to monitor streaming calls every `periodSeconds` and once the client side has waited for too long or the stream has become idle for too long, the stream would be terminated. The timeouts are derived from system properties, and if not specified default values will be provided.

Change to use [`WatchdogProvider`](https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/rpc/WatchdogProvider.java) preserves the same behaviors (same `periodSeconds`, which is renamed to `checkInterval` and same `waitTimeout` and `idleTimeout`, which are put in `GrpcCallContext`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,17 +55,35 @@ This library supports the following Google Cloud Platform services with clients 
 Quickstart
 ----------
 
-To call any of the supported Google Cloud Services simply add a corresponding client library artifact as a dependency to your project. The following instructions use `google-cloud-storage` as an example (specific instructions can be found in the README of each client).
+To call any of the supported Google Cloud Services simply add a corresponding client library 
+artifact as a dependency to your project. The following instructions use `google-cloud-storage` 
+as an example (specific instructions can be found in the README of each client).
 
-[//]: # ({x-version-update-start:google-cloud-storage:released})
+[//]: # ({x-version-update-start:google-cloud-bom:released})
 If you are using Maven, add this to your pom.xml file
 ```xml
-<dependency>
-  <groupId>com.google.cloud</groupId>
-  <artifactId>google-cloud-storage</artifactId>
-  <version>1.52.0</version>
-</dependency>
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bom</artifactId>
+        <version>0.70.0-alpha</version>
+        <type>pom</type>
+        <scope>import</scope>
+       </dependency>
+     </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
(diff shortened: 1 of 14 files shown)


hzyi-google:
@nithinsujir @garrettjonesgoogle @snehashah16 PTAL

hzyi-google:
@nithinsujir updated. PTAL.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -173,6 +191,17 @@ public GapicSpannerRpc(SpannerOptions options) {
     CredentialsProvider credentialsProvider =
         GrpcTransportOptions.setUpCredentialsProvider(options);
 
+    WatchdogProvider watchdogProvider =
+        InstantiatingWatchdogProvider.create()
+            .withExecutor(
+                Executors.newSingleThreadScheduledExecutor(
+                    new ThreadFactoryBuilder()
+                        .setDaemon(true)
+                        .setNameFormat("Cloud-Spanner-WatchdogInterceptor-%d")
```
nithinsujir: Should this be WatchdogProvider?

hzyi-google: Yes! Nice catch! Thanks.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/WatchdogInterceptorTest.java
```diff
@@ -1,202 +0,0 @@
-/*
```
snehashah16: Can this test be refactored to test new functionality ?

hzyi-google: I thought about that, but these tests are all unit tests with mocks and the [test](https://github.com/googleapis/gax-java/blob/master/gax/src/test/java/com/google/api/gax/rpc/WatchdogTest.java) for watchdog in gax is mostly equivalent so it seems unnecessary.

PR #3936, opened 2018-11-13
Overloaded get methods in DocumentSnapshot to allow passing in a default value

daninge:
Overloaded get methods in Firestore DocumentSnapshot to allow passing in a default value as a second argument. If the field path passed into the get method doesn't exist, then the default value is returned rather than the current behavior of null being returned.


schmidt-sebastian:
> @schmidt-sebastian do you have objections on adding the helpers?

Sorry, I do. We would like to keep this API as similar to the Android API as possible, as well as with all other clients. We played around with the exact idea in this PR before, and decided that it wasn't worth the extra overhead of adding this API to all platforms. We are always open to reconsider, but for now I would prefer if we did not change the Server API.

PR #3941, opened 2018-11-14
Add namespace and schema to settings.xml

elharo:
@garrettjonesgoogle


garrettjonesgoogle:
How much does this matter?

elharo:
The schema probably not much. The namespace, quite a bit though Maven is notorious for ignoring this. It's sort of like if a Java compiler decided it was going to ignore package names and only used the bare classnames for everything.

PR #3946, opened 2018-11-15
Bigtable Java samples

elisheva-qlogic:
HelloWorld, TableAdmin, and InstanceAdmin examples

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
@@ -15,7 +15,7 @@
  */
 package com.google.cloud.examples.bigtable;
 
-import com.google.api.gax.rpc.ApiException;
+import com.google.api.gax.rpc.NotFoundException;
 import com.google.api.gax.rpc.ServerStream;
 import com.google.cloud.bigtable.admin.v2.BigtableTableAdminClient;
 import com.google.cloud.bigtable.admin.v2.BigtableTableAdminSettings;
@@ -34,9 +34,6 @@ public class HelloWorld {
   private static final String COLUMN_FAMILY = "cf1";
   private static final String COLUMN_QUALIFIER = "greeting";
   private static final String ROW_KEY_PREFIX = "rowKey";
-
-  private final String projectId;
-  private final String instanceId;
   private final String tableId;
   private final BigtableDataClient dataClient;
   private final BigtableTableAdminClient adminClient;
@@ -54,8 +51,6 @@ public class HelloWorld {
   }
 
   public HelloWorld(String projectId, String instanceId, String tableId) throws IOException {
-    this.projectId = projectId;
-    this.instanceId = instanceId;
     this.tableId = tableId;
 
     // [START connecting_to_bigtable]
@@ -115,7 +110,7 @@ public class HelloWorld {
         dataClient.mutateRow(rowMutation);
         System.out.println(greetings[i]);
       }
-    } catch (ApiException e) {
+    } catch (NotFoundException e) {
       System.err.println("Exception while writing to table: " + e.getMessage());
     }
     // [END writing_rows]
@@ -132,7 +127,7 @@ public class HelloWorld {
             "Family: %s    Qualifier: %s    Value: %s%n",
             cell.getFamily(), cell.getQualifier().toStringUtf8(), cell.getValue().toStringUtf8());
       }
-    } catch (ApiException e) {
+    } catch (NotFoundException e) {
       System.err.println("Exception while reading a single row: " + e.getMessage());
     }
     // [END reading_a_row]
@@ -152,7 +147,7 @@ public class HelloWorld {
               cell.getFamily(), cell.getQualifier().toStringUtf8(), cell.getValue().toStringUtf8());
         }
       }
-    } catch (ApiException e) {
+    } catch (NotFoundException e) {
       System.err.println("Exception while reading table: " + e.getMessage());
     }
     // [END scanning_all_rows]
@@ -160,11 +155,11 @@ public class HelloWorld {
 
   public void deleteTable() {
     // [START deleting_a_table]
-    System.out.println("\nDeleting table");
+    System.out.println("\nDeleting table: " + tableId);
     try {
       adminClient.deleteTable(tableId);
       System.out.printf("Table %s deleted successfully%n", tableId);
-    } catch (ApiException e) {
+    } catch (NotFoundException e) {
       System.err.println("Exception while deleting table: " + e.getMessage());
     }
     // [END deleting_a_table]
--- a/google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITHelloWorld.java
+++ b/google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITHelloWorld.java
@@ -43,8 +43,8 @@ public class ITHelloWorld {
 
   private static final String INSTANCE_PROPERTY_NAME = "bigtable.instance";
   private static final String TABLE_PREFIX = "table";
-  private static String tableId;
   private static final String COLUMN_FAMILY = "cf1";
+  private static String tableId;
   private static BigtableDataClient dataClient;
   private static BigtableTableAdminClient adminClient;
   private static HelloWorld helloWorld;


google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,101 @@
...
+public class HelloWorld {
+
+  public static void main(String[] args) throws Exception {
+
+    // the ID of the cloud bigtable project, instance and table
+    String TABLE_ID = "Hello-Bigtable";
+    String COLUMN_FAMILY_ID = "cf1";
+    String COLUMN_QUALIFIER = "greeting";
+    String INSTANCE_ID = System.getenv("INSTANCE_ID");
+    String GCLOUD_PROJECT_ID = System.getenv("GCLOUD_PROJECT_ID");
```
igorbernstein2: Do other product examples use environment variables? Spot checking other examples it seems like they use argument

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,101 @@
...
+    if (!adminClient.exists(TABLE_ID)) {
+      CreateTableRequest createTableRequest =
+          CreateTableRequest.of(TABLE_ID).addFamily(COLUMN_FAMILY_ID);
+      System.out.println("Creating table " + TABLE_ID);
+      adminClient.createTable(createTableRequest);
+    }
+
+    try {
+      System.out.println("Write some greetings to the table");
+      String[] greetings = { "Hello World!", "Hello Bigtable!", "Hello Node!" };
```
igorbernstein2: Why node?

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,101 @@
...
+      for (int i = 0; i < greetings.length; i++) {
+        RowMutation rowMutation = RowMutation.create(TABLE_ID, rowKeyPrefix + i);
+        long timestamp = new Timestamp(System.currentTimeMillis()).getTime();
```
igorbernstein2: I dont understand what you are trying to do here. If you want now, just omit the timestamp

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,101 @@
...
+        rowMutation.setCell(COLUMN_FAMILY_ID, COLUMN_QUALIFIER, timestamp, greetings[i]);
+        dataClient.mutateRow(rowMutation);
+      }
+
+      System.out.println("Reading a single row by row key");
+      Row row = dataClient.readRow(TABLE_ID, rowKeyPrefix + 1);
+      System.out.println(row.getKey().toStringUtf8());
+      for (RowCell cell : row.getCells()) {
+        System.out.println(
```
igorbernstein2: please use printf, it'll be easier to read

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,101 @@
...
+
+    } catch (Exception e) {
+      System.out.println("Exception while running HelloWorld: " + e.getMessage());
+    } finally {
+      dataClient.close();
+      adminClient.close();
+    }
+  }
+}
+// [END HelloWorld]
```
igorbernstein2: newline

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/InstanceAdmin.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class InstanceAdmin {
+
+  public static void main(String[] args) throws IOException {
+
+    String GCLOUD_PROJECT = System.getenv("GCLOUD_PROJECT");
+
+    if (GCLOUD_PROJECT.length() == 0) {
```
igorbernstein2: If its missing, won't it be null instead of empty?

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/InstanceAdmin.java
```diff
@@ -0,0 +1,213 @@
...
+
+    // Create instance if does not exists
+    if (!found) {
+      System.out.println("Creating a PRODUCTION Instance");
+      // [START bigtable_create_prod_instance]
+      // Creates a Production Instance with the ID "ssd-instance"
+      // with cluster id "ssd-cluster", 3 nodes and location us-central1-f
+      CreateInstanceRequest createInstanceRequest = CreateInstanceRequest.of(instanceID)
+          .addCluster(clusterID, "us-central1-f", 3, StorageType.SSD).setType(Type.PRODUCTION)
+          .addLabel("prod-label", "prod-label");
```
igorbernstein2: The label name/value doesn't really help me understand what a label is

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/InstanceAdmin.java
```diff
@@ -0,0 +1,213 @@
...
+      // Create production instance with given request
+      try {
+        Instance instance = adminClient.createInstance(createInstanceRequest);
+        System.out
+            .println("PRODUCTION type instance : " + instance.getId() + " created successfully");
+
+      } catch (Exception e) {
+        System.out.println("Error creating prod-instance: " + e.getMessage());
```
igorbernstein2: If it fails to create an instance, I don't think we should just continue. It will break the rest of the operations with not_found errors

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/InstanceAdmin.java
```diff
@@ -0,0 +1,213 @@
...
+    deleteInstance(adminClient, "hdd-instance");
+
+    System.out.println("Add Cluster");
+    addCluster(adminClient, "ssd-instance", "ssd-cluster");
+
+    System.out.println("Delete the Cluster");
+    deleteCluster(adminClient, "ssd-instance", "ssd-cluster");
+
+    // end operations with deleting the pro-instance created in `runInstanceOperations`
+    deleteInstance(adminClient, "ssd-instance");
```
igorbernstein2: what about deleting the dev instance? also, please spell out production

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/InstanceAdmin.java
```diff
@@ -0,0 +1,213 @@
...
+
+    BigtableInstanceAdminSettings instanceAdminSettings =
+        BigtableInstanceAdminSettings.newBuilder().setProjectName(ProjectName.of(GCLOUD_PROJECT))
+            .build();
+
+    BigtableInstanceAdminClient adminClient =
+        BigtableInstanceAdminClient.create(instanceAdminSettings);
+
+    System.out.println("Create an instance (type: PRODUCTION) and run basic instance-operations");
+    runInstanceOperations(adminClient, "ssd-instance", "ssd-cluster");
```
igorbernstein2: The grouping is a bit inconsistent:
runInstanceOperations does a bunch of actions while everything else is single actions

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/InstanceAdmin.java
```diff
@@ -0,0 +1,213 @@
...
+
+    System.out.println("Create DEVELOPMENT instance");
+    createDevInstance(adminClient, "hdd-instance", "hdd-cluster");
+
+    System.out.println("Delete the Instance");
+    deleteInstance(adminClient, "hdd-instance");
+
+    System.out.println("Add Cluster");
```
igorbernstein2: The comments should be more semantic: not just adding a cluster, but enabling replication

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/InstanceAdmin.java
```diff
@@ -0,0 +1,213 @@
...
+
+  // Delete the Instance
+  public static void deleteInstance(BigtableInstanceAdminClient adminClient, String instanceID) {
+    // [START bigtable_delete_instance]
+    System.out.println(); //for a new-line
+    System.out.println("Deleting Instance");
+    try {
+      adminClient.deleteInstance(instanceID);
+      System.out.println("Instance deleted: " + instanceID);
+    } catch (Exception e) {
```
igorbernstein2: Would be good to explicit catch a typesafe exception to demonstrate their usage. Maybe catch the NotFound exception?

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.cloud.bigtable.data.v2.BigtableDataSettings;
+import com.google.cloud.bigtable.data.v2.models.InstanceName;
+
+import com.google.cloud.bigtable.data.v2.models.Query;
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowCell;
+import com.google.cloud.bigtable.data.v2.models.RowMutation;
+import java.sql.Timestamp;
+
+// [START HelloWorld]
```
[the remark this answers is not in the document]
chingor13: These annotations mark regions that cloud.google.com and developers.google.com can extract for code snippets. This allows the snippets to be testable in their GitHub repositories.

sduskis:
@igorbernstein2, the snippets are used in https://cloud.google.com/bigtable/docs/.  Our documentation has a tool to pull in code based on the tags

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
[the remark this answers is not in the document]
igorbernstein2: Isn't this too big too be a snippet? I would imagine that the doc snippets should be a bit more targeted?

igorbernstein2:
Also, to @chingor13 comment, it would be good to add tests to make sure these snippets continue to work

sduskis:
@elisheva-qlogic, there's a build failure:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project google-cloud-examples: Compilation failure
[ERROR] /tmpfs/src/github/google-cloud-java/google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdmin.java:[56,60] incompatible types
[ERROR]   required: java.util.List<com.google.bigtable.admin.v2.TableName>
[ERROR]   found:    java.util.List<java.lang.String>
```

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,183 @@
...
+import com.google.cloud.bigtable.data.v2.BigtableDataSettings;
+import com.google.cloud.bigtable.data.v2.models.InstanceName;
+import com.google.cloud.bigtable.data.v2.models.Query;
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowCell;
+import com.google.cloud.bigtable.data.v2.models.RowMutation;
+
+public class HelloWorld {
+
+  public static void main(String... args) throws Exception {
```
igorbernstein2: I think it would be nicer to:
- add instance variables to HelloWorld for project id/instance id/etc
- make all methods non static
- remove per method parameters
- use main to simply instantiate HelloWord passing it the appropriate parameters and then call run

something like:
```java
public class HelloWorld {
  final String COLUMN_FAMILY_ID = "cf1";
  final String COLUMN_QUALIFIER = "greeting";
  final String TABLE_PREFIX = "table";
  //...

  private final String projectId;
  private final String instanceId;
  //...
  private final BigtableDataClient dataClient;
  private final BigtableTableClient tableAdminClient;

    if (args.length != 3) 
     //..
    }
   String projectId = args[0];
   String instanceId = args[1];
   String tableId = args[2];
   HelloWorld helloWorld = new HelloWorld(projectId, instanceId, tableId);
   helloWorld.run();
  }

  public HelloWorld(String projectId, String instanceId, ...) {
    this.projectId = projectId;
    //..
   dataClient = BigtableDataClient.create(settings);
  }
  public void run() {
    createTable();
    writeToTable();
    //...
  }
  void createTable() {
    CreateTableRequest request = //...
    adminClient.createTable(request);
  }
  //...
}
```

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdmin.java
```diff
@@ -0,0 +1,381 @@
...
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class TableAdmin {
+
+  public static void main(String... args) {
+
+    final String GCLOUD_PROJECT = args[0];
+    final String INSTANCE_ID = args[1];
+    final String TABLE_ID = "table";
```
igorbernstein2: Please use a dynamic table name to avoid collisions.

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/HelloWorldTest.java
```diff
@@ -0,0 +1,116 @@
...
+public class HelloWorldTest {
+
+  private static final String INSTANCE_PROPERTY_NAME = "bigtable.instance";
+  private static final String TABLE_ID = "test-table" + System.currentTimeMillis();
+  private static final String COLUMN_FAMILY_ID = "test-cf";
+  private static final String COLUMN_QUALIFIER = "test-greeting";
+  private static final String ROW_KEY_PREFIX = "test-rowKey";
+  private static BigtableDataClient dataClient;
+  private static BigtableTableAdminClient adminClient;
+
```
igorbernstein2: add a garbage collection step, something like:

I think it would be nicer to:
- add instance variables to HelloWorld for project id/instance id/etc
- make all methods non static
- remove per method parameters
- use main to simply instantiate HelloWord passing it the appropriate parameters and then call run

something like:
```java

  private static String generateTableId() {
    return String.format("%s-%016x-%d", TABLE_PREFIX, System.currentTimeMillis(), new Random().nextLong());
  }
  void garbageCollect() {
    Pattern timestampPattern = Pattern.compile(TABLE_PREFIX + "-([0-9]+");
    for(String tableId : tableAdminClient.listTables()) {
      Matcher matcher = timestampPattern.matcher(tableName.getNameAsString());
      if (!matcher.matches()) {
       continue;
      }
      String timestampStr = matcher.group(1);
      long timestamp = Long.parseLong(timestampStr);
      if (System.currentTimeMillis() - timestamp < TimeUnit.MINUTES.toMillis(15)) {
        continue;
      }
      System.err.println("Garbaging collecting orphaned table: " + tableId);
      tableAdminClient.deleteTable(tableId);
    }
  }
```

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
elisheva-qlogic: Should TableAdmin and InstanceAdmin samples be restructured like this as well?

igorbernstein2: I think so, but can you do this for HelloWorld to start with and we can see how well it works out first?

elisheva-qlogic: sure

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,183 @@
...
+      for (int i = 0; i < greetings.length; i++) {
+        RowMutation rowMutation = RowMutation.create(TABLE_ID, ROW_KEY_PREFIX + i);
+        mutation = rowMutation.setCell(COLUMN_FAMILY_ID, COLUMN_QUALIFIER, greetings[i]);
+        dataClient.mutateRow(rowMutation);
+        System.out.println(greetings[i]);
+      }
+    } catch (Exception e) {
+      System.out.println("Exception while writing to table: " + e.getMessage());
+    }
+    return mutation;
```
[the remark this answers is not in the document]
elisheva-qlogic: I did this for test purposes. Do you prefer I do it a different way? Also, as per your comment below, once the methods are made non-static, I can't call them in my test class so I won't be able to test each specific method. Do you want me to just test that main doesn't crash? If so, I can take out all the return statements.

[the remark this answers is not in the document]
igorbernstein2: Your test cases can instantiate the class and then call the individual methods like before. (just passing the parameters in the constructor instead of the method)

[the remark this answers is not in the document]
elisheva-qlogic: Is it okay to leave the return statements then?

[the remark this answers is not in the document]
igorbernstein2: I'd prefer if the test read the rows from bigtable instead

elisheva-qlogic:
The main change is that HelloWorld sample is restructured. If this looks good, I will work on restructuring TableAdmin and InstanceAdmin samples like this as well.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,204 @@
...
+      String timestampStr = matcher.group(1);
+      long timestamp = Long.parseLong(timestampStr);
+      if (System.currentTimeMillis() - timestamp < TimeUnit.MINUTES.toMillis(15)) {
+        continue;
+      }
+      System.out.println("Garbage collecting orphaned table: " + tableId);
+      adminClient.deleteTable(tableId);
+    }
+  }
+}
```
igorbernstein2: Please move generateTableId & garbageCollect to the unit test. No reason to burden the customer with this complexity. It's only needed for automated tests. Have main() hardcode the tableId and have the unit test generate a unique one and garbage collect it

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,204 @@
...
+        BigtableTableAdminSettings.newBuilder()
+            .setInstanceName(com.google.bigtable.admin.v2.InstanceName.of(projectId, instanceId))
+            .build();
+
+    // Create bigtable admin client
+    adminClient = BigtableTableAdminClient.create(adminSettings);
+    // [END connecting_to_bigtable]
+  }
+
+  public void run() throws Exception {
```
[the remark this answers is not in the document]
elisheva-qlogic: `dataClient.close()` throws an exception, that's why I needed that here.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,204 @@
...
+      System.out.println("\nWriting some greetings to the table");
+      String[] greetings = {"Hello World!", "Hello Bigtable!", "Hello Java!"};
+      for (int i = 0; i < greetings.length; i++) {
+        RowMutation rowMutation =
+            RowMutation.create(tableId, rowKeyPrefix + i)
+                .setCell(columnFamily, columnQualifier, greetings[i]);
+        dataClient.mutateRow(rowMutation);
+        System.out.println(greetings[i]);
+      }
+    } catch (ApiException e) {
```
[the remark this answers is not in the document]
igorbernstein2: It's not obvious that the client throws typed exceptions for different status codes. So I thought it would be good to show a sample. The javadocs for the data client show a couple of examples:

https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java#L450

https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java#L543

Other subclasses are: NotFoundException, DeadlineExceededException, PermissionDeniedException, AlreadyExistsException

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/HelloWorld.java
```diff
@@ -0,0 +1,167 @@
...
+      String[] greetings = {"Hello World!", "Hello Bigtable!", "Hello Java!"};
+      for (int i = 0; i < greetings.length; i++) {
+        RowMutation rowMutation =
+            RowMutation.create(tableId, ROW_KEY_PREFIX + i)
+                .setCell(COLUMN_FAMILY, COLUMN_QUALIFIER, greetings[i]);
+        dataClient.mutateRow(rowMutation);
+        System.out.println(greetings[i]);
+      }
+    } catch (NotFoundException e) {
+      System.err.println("Exception while writing to table: " + e.getMessage());
```
igorbernstein2: Please be more specific:
System.err.println("Failed to write to a non-existent table");

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITHelloWorld.java
```diff
@@ -0,0 +1,162 @@
...
+        BigtableTableAdminSettings.newBuilder()
+            .setInstanceName(com.google.bigtable.admin.v2.InstanceName.parse(targetInstance))
+            .build();
+    adminClient = BigtableTableAdminClient.create(adminSettings);
+  }
+
+  @AfterClass
+  public static void afterClass() throws Exception {
+    adminClient.deleteTable(tableId);
+    garbageCollect();
```
igorbernstein2: please move garbage collection to either the beginning or the end

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITHelloWorld.java
```diff
@@ -0,0 +1,162 @@
...
+
+  @Before
+  public void setup() throws IOException {
+    if (adminClient == null || dataClient == null) {
+      throw new AssumptionViolatedException(
+          INSTANCE_PROPERTY_NAME + " property is not set, skipping integration tests.");
+    }
+    tableId = generateTableId();
+    helloWorld = new HelloWorld(instanceName.getProject(), instanceName.getInstance(), tableId);
+    if (!adminClient.exists(tableId)) {
```
igorbernstein2: You can remove the exists check. The will not exist

igorbernstein2: Also, I think you can move the HelloWorld instantiation to beforeClass to avoid re-creating the table

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITHelloWorld.java
```diff
@@ -0,0 +1,162 @@
...
+  @Test
+  public void testWriteToAndReadFromTable() {
+    // Write to table
+    helloWorld.writeToTable();
+    Row row = dataClient.readRow(tableId, "rowKey0");
+    assertNotNull(row);
+
+    // Read a single row
+    OutputStream outputStream = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(outputStream));
```
igorbernstein2: Please modify HelloWorld to take a PrintStream instead of changing the system one

sduskis:
Closing this in favor of #4274

PR #4017, opened 2018-11-19
Firestore: Adding FieldMask support to GetAll()

schmidt-sebastian:
This PR allows users to specify the DocumentMask proto and to reduce the number of fields returned in a Firestore read operation.

This is a port of https://github.com/googleapis/nodejs-firestore/pull/443. The API surface is slightly different, as the JavaScript implementation doesn't feel very fitting for Java. It would have ended up as

```
getAll(DocRef docRef, Object ... docRefsOrFieldMask)
```

I am not 100% sold that the API surface in this PR is the best either (should we use a List instead of an Array?), so feedback would be appreciated. We are not matching a client API.


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
```diff
@@ -17,29 +17,57 @@
...
+import javax.annotation.Nonnull;
 
-/** A DocumentMask contains the field paths affected by an update. */
-final class DocumentMask {
-  static final DocumentMask EMPTY_MASK = new DocumentMask(new TreeSet<FieldPath>());
+/** A FieldMask can be used to limit the number of fields returned by a `getAll()` call. */
+public final class FieldMask {
+  static final FieldMask EMPTY_MASK = new FieldMask(new TreeSet<FieldPath>());
 
   private final SortedSet<FieldPath> fieldPaths; // Sorted for testing.
```
pongad: I don't have strong opinion but is it worth using `ImmutableSortedSet` here?

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -93,7 +94,18 @@
...
+  /**
+   * Retrieves multiple documents from Firestore, while optionally applying a field mask to reduce
+   * the amount of data transmitted.
+   *
+   * @param documentReferences Array with Document References to fetch.
+   * @param fieldMask If set, specifies the subset of fields to return.
+   */
+  @Nonnull
+  ApiFuture<List<DocumentSnapshot>> getAll(
+      @Nonnull DocumentReference[] documentReferences, @Nullable FieldMask fieldMask);
```
pongad: just curious: Do we prefer array over list?

pongad:
Generally LGTM with some questions, but Yoshi should sign off.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
schmidt-sebastian: I think we might prefer a `List` here, but the Array matches the getAll(...) varargs method. If a user passes an array, they can omit the last argument and fall back to the varargs method.

I'm super torn on this and prefer Lists in general.

pongad: Ah I see the tradeoffs. I'm OK with this now.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
schmidt-sebastian: I'll leave as is for now, mainly because I don't yet use any Guava classes. The internal object can already not be edited as is, but this is mostly enforced by convention for now.

PR #4092, opened 2018-11-21
Bigtable: readRow methods added with Query as argument

### LINKED ISSUE #4053: Bigtable: There should be versions of readRowAsync() and readRowsAsync() that takes a Query as argument

elisheva-qlogic:
Fixes #4053


sduskis:
@elisheva-qlogic, can you please resolve the conflicts?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -265,6 +341,86 @@ public Row readRow(String tableId, String rowKey) {
     return readRowsCallable().first().futureCall(Query.create(tableId).rowKey(rowKey));
```
igorbernstein2: Please call `readRowAsync(tableId, rowKey, null)` here and the other singular readRow methods that don't take a filter

igorbernstein2:
The integration test failures look unrelated

igorbernstein2:
Thanks for the PR!

PR #4106, opened 2018-11-27
Bigtable: publish emulator to maven central

igorbernstein2:
I added a bigable emulator artifact a few weeks ago (#3840), but I didn't realize that it needed some extra configuration to get published to maven central. I think this PRs allows the emulator to be published, but I don't have a way to test this.

### DIFF
--- a/google-cloud-testing/google-cloud-bigtable-emulator/pom.xml
+++ b/google-cloud-testing/google-cloud-bigtable-emulator/pom.xml
@@ -60,6 +60,7 @@
           <serverId>ossrh</serverId>
           <nexusUrl>https://oss.sonatype.org/</nexusUrl>
           <autoReleaseAfterClose>false</autoReleaseAfterClose>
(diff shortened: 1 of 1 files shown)


chingor13:
The entire google-cloud-testing module is not published to maven central. There's a section in the google-cloud-testing pom.xml that skips the nexus staging deployment.

Do we also want to publish the google-cloud-gcloud-maven-plugin?

igorbernstein2:
No, I want to keep the maven plugin as an intenal implementation detail for building the emulator. I overrode the skipping in the emulator. Does that look correct?

igorbernstein2:
@chingor13 any chance this can be merged for this week's release? A customer is waiting to use the emulator

chingor13:
Yes, we should be able to get this out this week. I need to verify that we can override the skip in the submodule if the parent is skipping (it may not even attempt it in the submodule)

chingor13:
This looks like it does correctly opt the artifact in for publishing to maven.

PR #4108, opened 2018-11-27
New Client: IAM Credentials

chingor13:
Also adds a `utilies/new_client.py` script that can easily automate adding new client libraries.


chingor13:
The IAM Credentials protos are missing configurations with the correct java namespace. Extracted the new client script into #4111

PR #4112, opened 2018-11-28
changed TestServiceFactory and TestServiceRpcFactory to public

elisheva-qlogic:
Fixes some tests on Java 9+


chingor13:
@elisheva-qlogic Can you comment on why this is necessary?

chingor13:
I assume, you mean in Java 9+?

elisheva-qlogic:
yes

PR #4114, opened 2018-11-28
Add lint test and format generated code

chingor13:
Allows us to run lint tests for presubmits, continuous, and nightly.

Synthtool will also run all generated code through the google-java-format formatter so the generated code will not break the lint tests.


chingor13:
When this PR is merged, we will want to run the formatter against all the generated code or we will get 20+ PRs when autosynth runs.

PR #4115, opened 2018-11-29
BigQuery: Fix TableId behaviour for non default BigQueryClient project

### LINKED ISSUE #3808: Change in TableId behaviour for non default BigQueryClient project 1.44 -> 1.45
When using the getTable method of the BigQuery client the project of the TableId is overwritten. TableId would previously ignore this if a project had been set. Table Id has been changed such that the project is always overwritten. There is now no way to get a table for a project other than the one the client was created for.

```
TableId tableId = TableId.of(project, schema, tableName);
Table table = client.getTable(tableId);
// table returns null if project != client.getOptions().getProjectId()
```

_this is due to the change in TableId_

In BigQueryImpl:
```
public Table getTable(TableId tableId, TableOption... options) {
    final TableId completeTableId = tableId.setProjectId(getOptions().getProjectId());
    ...
```

TableId old code:
```
TableId setProjectId(String projectId) {
  return getProject() != null ? this : TableId.of(projectId, getDataset(), getTable());
}
```

TableId new code:
```
TableId setProjectId(String projectId) {
  Preconditions
    .checkArgument(!Strings.isNullOrEmpty(projectId), "Provided projectId is null or empty");
  return TableId.of(projectId, getDataset(), getTable());
}
```

pmakani:
Fixes #3808 create, update and delete methods.

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -350,11 +350,10 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
 
   @Override
   public boolean delete(TableId tableId) {
-    final TableId completeTableId =
-        tableId.setProjectId(
+    final TableId completeTableId = tableId.setProjectId(
         Strings.isNullOrEmpty(tableId.getProject())
-        ? getOptions().getProjectId()
-        : tableId.getProject()
+            ? getOptions().getProjectId()
+            : tableId.getProject()
     );
     try {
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -345,7 +350,12 @@ public boolean delete(String datasetId, String tableId) {
 
   @Override
   public boolean delete(TableId tableId) {
-    final TableId completeTableId = tableId.setProjectId(getOptions().getProjectId());
+    final TableId completeTableId =
+        tableId.setProjectId(
+        Strings.isNullOrEmpty(tableId.getProject())
+        ? getOptions().getProjectId()
```
chingor13: nit: on spacing

PR #4116, opened 2018-11-29
Storage: Close response body after request

### LINKED ISSUE #4107: WriteChannel for GCS doesn't read response body, leaks connections
#### Environment details

- OS: linux / centos7
- Java version: 1.9
- google-cloud-java version(s): 
  - com.google.cloud:google-cloud-storage:1.52.0
  - com.google.cloud:google-cloud-core-http:1.52.0
  - org.apache.httpcomponents:httpclient:4.5.5

#### Steps to reproduce

At a high level we're trying to upload 100s of objects to GCS ~quickly. We want to do this using an Apache HTTP client because we need to use a custom resolver. That said, I think this is broken on both Apache and other http clients, it just may not manifest itself the same way.

1. Use a custom transport factory builder in order to use Apache HTTP Client
2. Allocate a PoolingHttpClientConnectionManager in your transport, set the pool size to 1
3. Instantiate the Storage service using this transport factory
4. Allocate a write channel
5. Write to it
6. Close it
7. Allocate another write channel
8. Try to write to it.

Under the hood the client will try to find an available connection from the pool and fail. It thinks the one we used earlier is still in use.

If you run ss or netstat on the machine at this point you can see an established connection to GCS with ~165 bytes sitting in the socket's receive buffer.

The reason is that we don't read the body of the response. See here:

https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L685

The response in both success and exception cases is discarded without ever reading or closing the response.

HttpRequest's execute() method pretty clearly states we need to do something with the response:

https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/HttpRequest#execute()

andrey-qlogic:
Fixes #4107 WriteChannel for GCS doesn't read response body, leaks connections

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -725,7 +725,7 @@ public class HttpStorageRpc implements StorageRpc {
         code = ex.getStatusCode();
         message = ex.getStatusMessage();
       } finally {
-        if (response != null ) {
+        if (response != null) {
           response.disconnect();
         }
       }
--- a/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
+++ b/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
@@ -39,7 +39,9 @@ import com.google.cloud.Identity;
 import com.google.cloud.Policy;
 import com.google.cloud.ReadChannel;
 import com.google.cloud.RestorableState;
+import com.google.cloud.TransportOptions;
 import com.google.cloud.WriteChannel;
+import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.kms.v1.CreateCryptoKeyRequest;
 import com.google.cloud.kms.v1.CreateKeyRingRequest;
 import com.google.cloud.kms.v1.CryptoKey;
@@ -49,8 +51,6 @@ import com.google.cloud.kms.v1.GetKeyRingRequest;
 import com.google.cloud.kms.v1.KeyManagementServiceGrpc;
 import com.google.cloud.kms.v1.KeyManagementServiceGrpc.KeyManagementServiceBlockingStub;
 import com.google.cloud.kms.v1.KeyRingName;
-import com.google.cloud.TransportOptions;
-import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.kms.v1.LocationName;
 import com.google.cloud.storage.Acl;
 import com.google.cloud.storage.Acl.Role;
@@ -1762,8 +1762,10 @@ public class ITStorageTest {
 
   @Test(timeout = 5000)
   public void testWriteChannelWithConnectionPool() throws IOException {
-    TransportOptions transportOptions = HttpTransportOptions.newBuilder()
-        .setHttpTransportFactory(new CustomHttpTransportFactory()).build();
+    TransportOptions transportOptions =
+        HttpTransportOptions.newBuilder()
+            .setHttpTransportFactory(new CustomHttpTransportFactory())
+            .build();
     Storage storageWithPool =
         StorageOptions.newBuilder().setTransportOptions(transportOptions).build().getService();
     String blobName = "test-custom-pool-management";


frankyn:
Hi @andrey-qlogic, thanks for submitting this PR! Could you add a system test for this change?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -685,6 +685,7 @@ public void write(String uploadId, byte[] toWrite, int toWriteOffset, long destO
         HttpResponse response = httpRequest.execute();
         code = response.getStatusCode();
         message = response.getStatusMessage();
+        response.getContent().close();
```
sduskis: would it make sense to put this in a finally clause, or is execute the only method that throws an exception?

frankyn: Good call, `getContent()` can raise an exception. 

Instead it could use `disconnect()`. The documentation for [`HttpResponse`](https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/HttpRequest#execute()) docs state the following

> Execute the HTTP request and returns the HTTP response.
> Note that regardless of the returned status code, the HTTP response content has not been parsed yet, and must be parsed by the calling code.
>
> Note that when calling to this method twice or more, the state of this HTTP request 
> object isn't cleared, so the request will continue where it was left. For example, the state of the HttpUnsuccessfulResponseHandler attached to this HTTP request will remain the same as it was left after last execute.
> 
> Almost all details of the request and response are logged if Level.CONFIG is loggable. The only exception is the value of the Authorization header which is only logged if Level.ALL is loggable.
> 
> Callers should call HttpResponse.disconnect() when the returned HTTP response object is no longer needed. However, HttpResponse.disconnect() does not have to be called if the response stream is properly closed. Example usage:

```java
HttpResponse response = request.execute();
     try {
       // process the HTTP response object
     } finally {
       response.disconnect();
     }
```

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -681,14 +681,19 @@ public void write(String uploadId, byte[] toWrite, int toWriteOffset, long destO
...
         code = response.getStatusCode();
         message = response.getStatusMessage();
       } catch (HttpResponseException ex) {
         exception = ex;
         code = ex.getStatusCode();
         message = ex.getStatusMessage();
+      } finally {
+        if (response != null ) {
+          response.disconnect();
+        }
```
frankyn: LGTM, I would still like to see a system test for this change. Something like the following for example in `/src/test/java/com/google/cloud/storage/it/ITStorageTest.java`

```java
private static class CustomHttpTransportFactory implements HttpTransportFactory {
    private static final HttpTransportFactory INSTANCE = new CustomHttpTransportFactory();

    @Override
    public HttpTransport create() {
      PoolingHttpClientConnectionManager manager = new PoolingHttpClientConnectionManager();
      manager.setMaxTotal(1);
      return new ApacheHttpTransport(HttpClients.createMinimal(manager));
    }
  }

  @Test(timeout = 5000)
  public void testWriteChannelWithConnectionPool() throws IOException {
    TransportOptions transportOptions = HttpTransportOptions.newBuilder().setHttpTransportFactory(new CustomHttpTransportFactory()).build();
    Storage storageWithPool = StorageOptions.newBuilder().setTransportOptions(transportOptions).build().getService();
    String blobName = "test-read-and-write-channels-blob";
    BlobInfo blob = BlobInfo.newBuilder(BUCKET, blobName).build();
    byte[] stringBytes;
    try (WriteChannel writer = storageWithPool.writer(blob)) {
      stringBytes = BLOB_STRING_CONTENT.getBytes(UTF_8);
      writer.write(ByteBuffer.wrap(BLOB_BYTE_CONTENT));
      writer.write(ByteBuffer.wrap(stringBytes));
    }
    try (WriteChannel writer = storageWithPool.writer(blob)) {
      stringBytes = BLOB_STRING_CONTENT.getBytes(UTF_8);
      writer.write(ByteBuffer.wrap(BLOB_BYTE_CONTENT));
      writer.write(ByteBuffer.wrap(stringBytes));
    }
  }

```

andrey-qlogic:
Thanks, @frankyn, I added a test as you mentioned.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -1744,6 +1760,27 @@ public void testWriteChannelExistingBlob() throws IOException {
     assertTrue(storage.delete(BUCKET, blobName));
   }
 
+  @Test(timeout = 5000)
+  public void testWriteChannelWithConnectionPool() throws IOException {
+    TransportOptions transportOptions = HttpTransportOptions.newBuilder()
+        .setHttpTransportFactory(new CustomHttpTransportFactory()).build();
+    Storage storageWithPool =
+        StorageOptions.newBuilder().setTransportOptions(transportOptions).build().getService();
+    String blobName = "test-read-and-write-channels-blob";
```
frankyn: test-custom-pool-management

andrey-qlogic:
Thanks @frankyn, blobName changed

chingor13:
Integration test failure is for downloading the bigtable emulator and is unrelated

PR #3821, opened 2018-10-16
Java 11 compatibility

chingor13:
Fixes #3729 
Fixes #3712

PR #4119, opened 2018-11-30
Add release profile to bigtable emulator

JesseLovelace:
Add release profile to bigtable emulator's pom.xml so that the release process works properly


chingor13:
Java 11 test will fail until it's configured by #3821

chingor13:
Spanner integration test was unrelated

PR #4168, opened 2018-12-04
google-cloud-bom should not need google-cloud-core test-jar

### LINKED ISSUE #3707: test-jar in google-cloud-bom

chingor13:
Fixes #3707

### DIFF
--- a/google-cloud-testing/pom.xml
+++ b/google-cloud-testing/pom.xml
@@ -37,6 +37,12 @@
         <scope>import</scope>
       </dependency>
       <dependency>
+        <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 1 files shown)


sduskis:
@chingor13, all tests are failing.  It looks like a pom.xml issue.

chingor13:
The test build is fixed, PTAL.

PR #4170, opened 2018-12-04
Storage: Give users the ability to disable gzip content encoding to increase throughput

### LINKED ISSUE #3822: Give users the ability to disable gzip content encoding to increase throughput
- OS: all
- Java version: 1.8  (probably all)
- google-cloud-java version(s): 1.23.0

#### Steps to reproduce

1. Upload to GCS using the cloud storage library.
2. Observe poor throughput (20 MiB/s)_ and that lots of time is spent compressing the content. 

On GCE I'm able to achieve 70 MiB/s if I disable gzip encoding when uploading to GCS.  With gzip encoding it's  much worse at 20 MiB/s. However  disabling gzip encoding isn't an option exposed in the interface. This makes google-cloud-java poor in comparison to the REST api it's wrapping.  

This is a request to expose the gzip encoding of the stream as a storage option.  See code in com.google.cloud.storage.spi.v1.HttpStorageRpc.create  there is no code to call StorageRequest.setDisableGZipContent  (which defaults to false) on the underling com.google.api.services.storage.StorageRequest:

      Storage.Objects.Insert insert = storage.objects()
          .insert(storageObject.getBucket(), storageObject,
              new InputStreamContent(storageObject.getContentType(), content));
      insert.getMediaHttpUploader().setDirectUploadEnabled(true);
      setEncryptionHeaders(insert.getRequestHeaders(), ENCRYPTION_KEY_PREFIX, options);
      return insert.
      ... .execute();

Can we please provide this feature to maintain comparable performance with the com.google.api.services.storage.*  client?

thanks

#### Code snippet
Uploading an object to GCS and observe performance, then do so with gzip disabled: 

        /// USING THE OLD CLIENT
        Storage client = getService();
        Storage.Objects.Insert insertRequest =
                client.objects().insert(bucketName, objectMetadata, contentStream);

        /// Currently NO ability to this in google-cloud-java, need this for much faster throughput on GCE.
        insertRequest.setDisableGZipContent(true);
        insertRequest.getMediaHttpUploader().setDisableGZipContent(true);

        insertRequest.getMediaHttpUploader().setDirectUploadEnabled(true);
        long start = System.currentTimeMillis();
        insertRequest.execute();
        long end = System.currentTimeMillis();
        System.out.println( String.format("%.02f MiB/s ",(length / (1024*1024.0) )/ ((end - start) / 1000.0)));

Also, perhaps document somewhere that in order to get any reasonable performance on Java8 it's necessary to disable Galios/Counter mode in java.security:  https://stackoverflow.com/questions/25992131/slow-aes-gcm-encryption-and-decryption-with-java-8u20

andrey-qlogic:
Fixes #3822

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
+++ b/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
@@ -165,8 +165,7 @@ public class StorageImplTest {
           StorageRpc.Option.IF_GENERATION_MATCH, 0L,
           StorageRpc.Option.PREDEFINED_ACL, BUCKET_TARGET_PREDEFINED_ACL.getValue());
   private static final Map<StorageRpc.Option, ?> BLOB_TARGET_OPTIONS_CREATE_DISABLE_GZIP_CONTENT =
-      ImmutableMap.of(
-          StorageRpc.Option.IF_DISABLE_GZIP_CONTENT, true);
+      ImmutableMap.of(StorageRpc.Option.IF_DISABLE_GZIP_CONTENT, true);
   private static final Map<StorageRpc.Option, ?> BLOB_TARGET_OPTIONS_UPDATE =
       ImmutableMap.of(
           StorageRpc.Option.IF_METAGENERATION_MATCH, BLOB_INFO1.getMetageneration(),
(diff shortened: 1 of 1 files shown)


sduskis:
@frankyn do you want to take this one?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -16,6 +16,7 @@
 
 package com.google.cloud.storage.spi.v1;
 
+import static com.google.cloud.storage.spi.v1.StorageRpc.Option.IF_DISABLE_GZIP_CONTENT;
```
frankyn: I don't think this is required because the base class has this option defined.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -346,6 +346,14 @@ public static BlobTargetOption metagenerationNotMatch() {
       return new BlobTargetOption(StorageRpc.Option.IF_METAGENERATION_NOT_MATCH);
     }
 
+    /**
+     *  Returns an option for blob's data disabledGzipContent. If this option is used,
+     *  the request will create a blob with disableGzipContent; at present, this is only for upload.
+     */
+    public static BlobTargetOption disableGzipContent() {
+      return new BlobTargetOption(StorageRpc.Option.IF_DISABLE_GZIP_CONTENT, true);
+    }
```
frankyn: Could you add a unit test for this option? 

Related unit tests: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java#L406

andrey-qlogic:
Fixed review comments and added an unit test.

PadrePadrone:
Hi,
1. In what GCP client version is this new option available?
2. Could you please provide an example of how to use it?
Thanks

andrey-qlogic:
@PadrePadrone, thanks for the comment, would you please take a look at this PR [4314](https://github.com/googleapis/google-cloud-java/pull/4314)

PadrePadrone:
@andrey-qlogic, thank you for the super quick response and handling.
When will this fix be released?

PR #4178, opened 2018-12-05
Make StandardTableDefinition$StreamingBuffer/getEstimatedRows nullable

mikekap:
They're already `Long` - currently calling the accessors will throw an exception if those are actually `null`.


mikekap:
```
Caused by: io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'logging.googleapis.com/read_requests' and limit 'ReadRequestsPerMinutePerProject' of service 'logging.googleapis.com' for consumer 'project_number:1016721519174'.
```

I don't think that's related to my change.

mikekap:
Looks like this was fixed by #4247

PR #4180, opened 2018-12-05
Bigquery: When a table is not found

### LINKED ISSUE #3930: When a table is not found, the exception is swallowed up and `null` is returned
#### Environment details

- OS: Mac (10.13.6)
- Java version: 1.8.0_181
- google-cloud-java version(s): 1.52.0

#### Steps to reproduce

1. n/a
2. n/a

#### Stacktrace

n/a

#### Code snippet

https://github.com/googleapis/google-cloud-java/blob/7a72784076308bb31786b9138efe856845a86fde/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java#L234

https://github.com/googleapis/google-cloud-java/blob/7a72784076308bb31786b9138efe856845a86fde/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java#L234

When a table is not found, `null` is returned instead of the actual HTTP code and BigQuery error message. This makes debugging much harder and is not intuitive for clients calling this method that `null` is actually returned when the table is not found. Instead, the exception should be propagated back up to the client so it can be handled accordingly.

```
@Override
  public Table getTable(String projectId, String datasetId, String tableId,
      Map<Option, ?> options) {
    try {
      return bigquery.tables()
          .get(projectId, datasetId, tableId)
          .setFields(Option.FIELDS.getString(options))
          .execute();
    } catch (IOException ex) {
      BigQueryException serviceException = translate(ex);
      if (serviceException.getCode() == HTTP_NOT_FOUND) {
        return null;
      }
      throw serviceException;
    }
  }
```

pmakani:
Fix #3930 When a table is not found.

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
@@ -246,7 +246,7 @@ public class HttpBigQueryRpc implements BigQueryRpc {
     } catch (IOException ex) {
       BigQueryException serviceException = translate(ex);
       if (serviceException.getCode() == HTTP_NOT_FOUND) {
-        return null;
+        throw new BigQueryException(HTTP_NOT_FOUND, "Table not found");
       }
       throw serviceException;
     }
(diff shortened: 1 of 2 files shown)


chingor13:
We might be able add an option parameter or option for opting in to throwing an exception on 404 to preserve backwards compatibility.

sduskis:
@pmakani, the integration tests fail.

chingor13:
The tests are failing because it's a breaking behavioral change.

Issue #4166, opened 2018-12-03
Spanner: Flaky integration tests


chingor13:
Spanner IT tests are occasionally failing with transaction aborted status

Example:
https://source.cloud.google.com/results/invocations/676bd73d-427e-48e7-b824-62b79e3288e0/targets/github%2Fgoogle-cloud-java%2Fgoogle-cloud-clients%2Fgoogle-cloud-spanner%2Ftarget%2Ffailsafe-reports/tests

Last 9 failures:
* [build 647](https://source.cloud.google.com/results/invocations/ff0d2ef9-3895-4788-b497-18ccbfab6714/targets) - timeout
* [build 643](https://source.cloud.google.com/results/invocations/baee83ce-402c-4ed8-b8d6-0a5d7792efde/targets) - timeout
* [build 642](https://source.cloud.google.com/results/invocations/e9e22a80-3fc7-4655-9f88-590ee9d7266e/targets) - Table already exists
* [build 640](https://source.cloud.google.com/results/invocations/98d65ba1-e840-4944-be10-3cd28f9a9608/targets) - ABORTED in abort and retry test
* [build 635](https://source.cloud.google.com/results/invocations/f99196f2-117f-4bd2-a3df-c83a6aab39ef/targets) - ABORTED in abort and retry test
* [build 630](https://source.cloud.google.com/results/invocations/f00b00bc-c4ee-4f4f-b290-861a39c49978/targets) - timeout
* [build 629](https://source.cloud.google.com/results/invocations/cc3ffbdb-e2db-4348-9b70-f883f966a969/targets) - timeout
* [build 628](https://source.cloud.google.com/results/invocations/36d4d872-e281-4851-b135-df5c52b3faf8/targets) - ABORTED in abort and retry test
* [build 626](https://source.cloud.google.com/results/invocations/676bd73d-427e-48e7-b824-62b79e3288e0/targets) - ABORTED w/ transaction not found in Partitioned DML test

chingor13:
We need to fix this before GA.

chingor13:
We also see `WARNING: Leaked session` during the tests as well.

chingor13:
The timeouts are being addressed in #4176

chingor13:
Spanner integration tests are looking fine after upping the timeout and handling the aborted exceptions. The only recent failures are due to maven networking issues.

PR #4181, opened 2018-12-05
Spanner: Handle AbortedExceptions that occur before commit

nithinsujir:
The current TransactionRunner code checks for aborted transactions by
using the isAborted() function. This function returns the value of
"aborted" that is only set when processing resultSets. This does not
work for DML which only uses the resultSet stats and does not actually
call ResultSet.next().

This patch adds an additional check for the aborted exception type and
retries such transactions.

Fixes one of the failures in #4166 (abort in partitionDML test)

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
@@ -88,7 +88,8 @@ public final class ITDMLTest {
             for (String stmt : stmts) {
               if (throwAbortOnce) {
                 throwAbortOnce = false;
-                throw SpannerExceptionFactory.newSpannerException(ErrorCode.ABORTED, "Abort in test");
+                throw SpannerExceptionFactory.newSpannerException(
+                    ErrorCode.ABORTED, "Abort in test");
               }
 
               rowCount += transaction.executeUpdate(Statement.of(stmt));


nithinsujir:
R: @jcorwin
R: @biswanag
R: @guangyus

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
```diff
@@ -83,6 +87,11 @@ public Long run(TransactionContext transaction) {
             long rowCount = 0;
             for (String stmt : stmts) {
               rowCount += transaction.executeUpdate(Statement.of(stmt));
+
```
guangyus: This block should probably be moved up before transaction.executeUpdate(..). Otherwise, the DML will be executed twice in the backend, second time being a replay.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
```diff
@@ -92,6 +101,16 @@ public Long run(TransactionContext transaction) {
...
   }
 
+  @Test
+  public void abortOnceShouldSucceedAfterRetry() {
+    try {
+      throwAbortOnce = true;
+      executeUpdate(DML_COUNT, INSERT_DML);
+    } catch (AbortedException e) {
+      fail("Abort Exception not caught and retried");
+    }
```
guangyus: Assert that throwAbortOnce == false;

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
nithinsujir: Fixed.

Yes, I agree about the unit test. There is no DML unit test and it's going to be a much bigger change than we want right now for GA.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
nithinsujir: Fixed.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
nithinsujir: Hm, actually that doesn't work. Looks like there are some assumptions that are affected.

com.google.cloud.spanner.SpannerException: UNKNOWN: Waited 500 milliseconds for com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture@73a2e526[status=PENDING, info=[inputFuture=[com.google.api.core.ApiFutureToListenableFuture@7d64e326], exceptionType=[class java.lang.Exception], fallback=[com.google.api.core.ApiFutures$GaxFunctionToGuavaFunction@13f95696]]]

nithinsujir: Nm, sorry for the spam. It was unrelated.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITDMLTest.java
```diff
@@ -82,6 +86,11 @@ private void executeUpdate(long expectedCount, final String... stmts) {
           public Long run(TransactionContext transaction) {
             long rowCount = 0;
             for (String stmt : stmts) {
+              if (throwAbortOnce) {
+                throwAbortOnce = false;
+                throw SpannerExceptionFactory.newSpannerException(ErrorCode.ABORTED, "Abort in test");
```
chingor13: From the code formatter:
```suggestion
                throw SpannerExceptionFactory.newSpannerException(
                    ErrorCode.ABORTED, "Abort in test");
```

PR #4185, opened 2018-12-06
Empty table results can have a schema

mikekap:
When you run a query with `LIMIT 0` it may still have a schema - make sure to return it even if there aren't any results.


sduskis:
@mikekap, this change needs unit test and a code format fix.  Are you planning to get to this?

@ajaaym, perhaps someone else can help with this?

ajaaym:
@sduskis done. can you add cla -yes?

PR #4197, opened 2018-12-07
Spanner: fix connection timeout issue.

### LINKED ISSUE #3616: How to specify Spanner connect timeout?
If the client has some kind of network issue connecting to Spanner, it seems to hang indefinitely. Is there a way to specify the timeout? I don't see anything in `SpannerOptions`.

charlesliqlogic:
Fixes #3616

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
@@ -87,9 +87,6 @@ import io.opencensus.trace.AttributeValue;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 import io.opencensus.trace.Tracing;
-import org.threeten.bp.Duration;
-import org.threeten.bp.Instant;
-
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.AbstractList;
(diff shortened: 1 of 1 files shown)


charlesliqlogic:
I signed it!

sduskis:
@chingor13, is this ready to be merged?

sqkumar:
We have ran into a similar problem with this library and it seems that it blocks on waiting for a valid session to enter the synchronized queue.

There is a waiter that tries to grab an available session from a queue, however when the connection drops it is unable to create a session and enqueue it. This means that the waiter polls indefinitely waiting on a session to take from the queue and use.  Spanner.close() actually poisons the queue by enqueuing an exception. FWIW that is my basic understanding after spending a bit of time debugging this.

DatabaseClientImpl calls getReadSession in SessionPool which "waits" for an available session:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L882

SessionPool.closeAsync which will put the poison record in the queue:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L1032

So my thinking is that even if the createSession eventually "fails", then we would still hang waiting on the session queue to be populated with a session.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -232,14 +233,22 @@ public void cancelled(Context context) {
    *
    * <p>TODO: Consider replacing with RetryHelper from gcloud-core.
    */
-  static <T> T runWithRetries(Callable<T> callable) {
+  static <T> T runWithRetries(Callable<T> callable, RetrySettings retrySettings) {
```
sqkumar: Should we also try to respect the global timeout in RetrySettings?

charlesliqlogic:
I will create new pull request from my branch.

PR #4206, opened 2018-12-10
Refresh Compute Java client.

andreamlin:
Regenerated with https://github.com/googleapis/discovery-artifact-manager/pull/100. That must be merged in first.


chingor13:
Should I not have accepted #4201? If not, we can wait for autosynth to run tomorrow morning.

andreamlin:
@chingor13 Oh I didn't see that PR. No rush. The updated GAPIC config that I just checked might have a couple more methods added.

PR #4208, opened 2018-12-11
Parse compute resource names [DO NOT MERGE]

andreamlin:
This is a sample regeneration of google-cloud-compute using artman with https://github.com/googleapis/api-common-java/pull/68 and https://github.com/googleapis/discovery-artifact-manager/pull/102.

In addition to the generated changes, [ITComputeTest](https://github.com/googleapis/google-cloud-java/pull/4208/files#diff-72d767a100f949c4606b04bb6820a8fe) has been manually changed to remove the `trimUrl()` hack that manually removed the service address from resource paths.


google-cloud-clients/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/it/ITComputeTest.java
```diff
@@ -157,9 +157,4 @@ public void testAggregatedListDiskTypes() {
       assertThat(diskType.getDefaultDiskSizeGb()).isNotNull();
     }
   }
-
-  /** For a given resource's URI, trim the path until it contains only the PathTemplate string. */
-  private static String trimUrl(String url) {
```
andreamlin: We don't have to do this manually any more, since parsing a URL resource path will be built into the ResourceName.parse() function.

Issue #3604, opened 2018-08-24
Compute API v1: xxxResourceName.parse throws ValidationException


dtretyakov:
Currently all `parse()` methods in `xxxResourceName` classes expect that input string will be formatted like `projects/{project}/...`, but data returned from APIs contains URL prefixes like `https://www.googleapis.com/compute/v1/projects/...` and as result parsing fails with `ValidationException`.

### Example
Receive instance details where:
```
instance.zone = https://www.googleapis.com/compute/v1/projects/project-123/zones/europe-west3-c
```

Then when trying to parse it with `ProjectZoneName.parse(instance.zone)` will cause the following exception:

```
ProjectZoneName.parse: formattedString not in valid format: Parameter "https://www.googleapis.com/compute/v1/projects/project-123/zones/europe-west3-c" must be in the form "projects/{project=*}/zones/{zone=*}"
```

hzyi-google:
@andreamlin I added `priotity: p2` to remove the `triage me` label. Feel free to bump to p1 if you find it necessary.

PR #4213, opened 2018-12-11
Regenerate google-cloud-compute

andreamlin:
Generation of Compute client with latest gapic-generator changes.

Created by artman, using ~https://github.com/googleapis/gapic-generator/pull/2478 and https://github.com/googleapis/discovery-artifact-manager/pull/102~ discovery-artifact-manager and toolkit from master.
This also contains one set of manual changes in [ITComputeTest](https://github.com/googleapis/google-cloud-java/pull/4213/files#diff-72d767a100f949c4606b04bb6820a8fe), where we remove the hack needed to parse a ResourceName string.

Then used java formatting via synth.py.

Fixes https://github.com/googleapis/google-cloud-java/issues/3604.

Maybe fixes https://github.com/googleapis/google-cloud-java/issues/4218?

### DIFF
--- a/google-cloud-clients/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/it/ITComputeTest.java
+++ b/google-cloud-clients/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/it/ITComputeTest.java
@@ -76,8 +76,7 @@ public class ITComputeTest {
   public void testGetDiskType() {
     DiskType diskType =
         diskTypeClient.getDiskType(ProjectZoneDiskTypeName.of(DISK_TYPE, DEFAULT_PROJECT, ZONE));
-    ProjectZoneDiskTypeName returnDiskName =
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/it/ITComputeTest.java
```diff
@@ -157,9 +155,4 @@ public void testAggregatedListDiskTypes() {
       assertThat(diskType.getDefaultDiskSizeGb()).isNotNull();
     }
   }
-
-  /** For a given resource's URI, trim the path until it contains only the PathTemplate string. */
-  private static String trimUrl(String url) {
```
andreamlin: We can remove this hack! Disregarding the URL will be a built-in to the ResourceNames now.

andreamlin:
PTAL

andreamlin:
Looks like it was a Spanner IT failure, so I assume we can ignore that.

PR #4215, opened 2018-12-12
Update gax javadoc link to point to latest.

kolea2:


chingor13:
Is this actually what we want to do? If this library depends on 1.35.0, shouldn't these links always point to the 1.35.0 docs instead of whatever the latest is?

kolea2:
True, I was initially looking at this pull request (https://github.com/googleapis/google-cloud-java/pull/2991) which references eventually switching gax.version to latest, but I see your point. Maybe we could look into having one variable for gax.version instead of updating it just for the javadoc. What do you think?

chingor13:
It may be difficult as we need to specify the gax version in the bom and google-cloud-clients and they do not currently share a parent (which would let us share properties).

If you can find a way to have us maintain the version in a single place, that would be fantastic.

chingor13:
Specifying the version in one place is not unique to gax. We should create a cleanup issue to do this across all of our dependencies where possible.

kolea2:
Created issue #4223 for this!

Issue #4223, opened 2018-12-13
Specify versions of dependencies in one central location for pom.xml files


kolea2:
Cleanup properties like `<gax.version>` in pom.xml files and have them located in one place to reference.

PR #4236, opened 2018-12-17
Update javadoc link for api-common-java.

kolea2:


sduskis:
@elharo, we have a disconnect between our `google-cloud-clients/pom.xml` use of versions for docs, and the version of the actual library that's defined in the bom.  Ideally, we only specify the version once.

kolea2:
@elharo @sduskis I can look into that as part of the javadoc cleanup issue I created last week: #4223

PR #4240, opened 2018-12-18
OpenCensus Support for Cloud Pub/Sub

dinooliva:
Adds OpenCensus context propagation to the CPS Publisher and Subscriber. This pr is initially meant for discussion between OpenCensus and CPS.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/pom.xml
+++ b/google-cloud-clients/google-cloud-pubsub/pom.xml
@@ -82,7 +82,7 @@
       <groupId>io.opencensus</groupId>
       <artifactId>opencensus-impl</artifactId>
       <version>${opencensus.version}</version>
-      <scope>runtime</scope>
+      <scope>test</scope>
     </dependency>
     <!-- Need testing utility classes for generated gRPC clients tests -->
     <dependency>
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -626,6 +626,7 @@ public class Publisher {
      * Gives the ability to set an {@link ApiFunction} that will transform the {@link PubsubMessage}
      * before it is sent
      */
+    @BetaApi
     public Builder setTransform(ApiFunction<PubsubMessage, PubsubMessage> messageTransform) {
       this.messageTransform =
           Preconditions.checkNotNull(messageTransform, "The messageTransform cannnot be null.");


sduskis:
@dinooliva, did @igorbernstein2 reach out to you yet about his work in GAX? That work should supersede this PR.

igorbernstein2:
we are in touch

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  @MustBeClosed
+  private static Scope createScopedSpan(String name) {
+    return tracer
+        .spanBuilderWithExplicitParent(name, tracer.getCurrentSpan())
+        .setRecordEvents(true)
+        // TODO(dpo): set to default.
+        .setSampler(Samplers.alwaysSample())
```
sduskis: @dinooliva, does this need to be fixed before merging?

google-cloud-clients/google-cloud-pubsub/pom.xml
```diff
@@ -16,6 +16,7 @@
   </parent>
   <properties>
     <site.installationModule>google-cloud-pubsub</site.installationModule>
+    <opencensus.version>0.15.1</opencensus.version>
```
sduskis: Please remove this variable.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,174 @@
...
+      this.receiver = receiver;
+    }
+
+    @Override
+    public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
+      String encodedTagContext = message.getAttributesOrDefault(TAG_CONTEXT_KEY, "");
+      String encodedSpanContext = message.getAttributesOrDefault(TRACE_CONTEXT_KEY, "");
+      try (
+          Scope spanScope = createScopedSpan("receiver");
+          Scope statsScope = createScopedTagContext(encodedTagContext)) {
```
igorbernstein2: Shouldn't the creation of the span be conditional on the attributes being present?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,174 @@
...
+    } catch (TagContextDeserializationException exn) {
+      logger.log(Level.INFO, "OpenCensus: Tag Context Deserialization Exception: " + exn);
+      return tagger.withTagContext(tagger.getCurrentTagContext());
+    }
+  }
+
+  @MustBeClosed
+  private static Scope createScopedSpan(String name) {
+    return tracer
+        .spanBuilderWithExplicitParent(name, tracer.getCurrentSpan())
```
igorbernstein2: What would be the current span on the receiver side? Shouldn't it start a new root span?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,174 @@
...
+    private OpenCensusMessageReceiver(MessageReceiver receiver) {
+      this.receiver = receiver;
+    }
+
+    @Override
+    public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
+      String encodedTagContext = message.getAttributesOrDefault(TAG_CONTEXT_KEY, "");
+      String encodedSpanContext = message.getAttributesOrDefault(TRACE_CONTEXT_KEY, "");
+      try (
+          Scope spanScope = createScopedSpan("receiver");
```
igorbernstein2: If I'm not mistaken, there is a good probability that this will be a top level span. Since this is a continuation of the publisher span, maybe it should be prefixed with the publisher span's name?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -187,11 +187,11 @@ public String getTopicNameString() {
    * @param message the message to publish.
    * @return the message ID wrapped in a future.
    */
-  public ApiFuture<String> publish(PubsubMessage message) {
+  public ApiFuture<String> publish(PubsubMessage originalMessage) {
     if (shutdown.get()) {
       throw new IllegalStateException("Cannot publish on a shut-down publisher.");
     }
-
+    PubsubMessage message = OpenCensusUtil.putOpenCensusAttributes(originalMessage);
```
igorbernstein2: Is there any way to hook into OpenCensus' sampling logic here and avoid increasing request size if the span will not be published?

google-cloud-clients/google-cloud-pubsub/pom.xml
```diff
@@ -46,6 +47,11 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-auth</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-api</artifactId>
+      <version>${opencensus.version}</version>
```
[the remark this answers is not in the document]
sduskis: This still needs to be changed.  Please remove `<version>${opencensus.version}</version>`

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,193 @@
+/* Copyright 2018 Google Inc.
```
sduskis: s/2018/2019

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,193 @@
...
+import io.opencensus.trace.TraceOptions;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import io.opencensus.trace.samplers.Samplers;
+
+import java.util.Base64;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+final class OpenCensusUtil {
```
sduskis: class comments are need

google-cloud-clients/pom.xml
```diff
@@ -167,6 +167,7 @@
     <!-- test dependency versions -->
     <easymock.version>3.4</easymock.version>
     <objenesis.version>2.6</objenesis.version>
+    <opencensus.version>0.16.0</opencensus.version>
```
sduskis: There is already an `<opencensus.version>` above.  Please remove this duplicate entry.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,193 @@
...
+
+  private static final Tagger tagger = Tags.getTagger();
+  private static final TagContextBinarySerializer serializer =
+      Tags.getTagPropagationComponent().getBinarySerializer();
+
+  private static final TraceOptions SAMPLED = TraceOptions.builder().setIsSampled(true).build();
+  private static final Tracer tracer = Tracing.getTracer();
+
+  // Used in Publisher.
+  // TODO: consider adding configuration support to control adding these attributes.
```
sduskis: I believe we need the configuration control before merging this.  This should be off by default, I would think.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,193 @@
...
+  // class
+  private static final class OpenCensusMessageReceiver implements MessageReceiver {
+    private final MessageReceiver receiver;
+
+    private OpenCensusMessageReceiver(MessageReceiver receiver) {
+      this.receiver = receiver;
+    }
+
+    @Override
+    public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
```
sduskis: Do we need configuration here as well, so that the user can define whether or not they want this?

google-cloud-clients/google-cloud-pubsub/pom.xml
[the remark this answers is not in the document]
dinooliva: Seems this dependency is inherited from google-cloud-clients so I can remove it altogether.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
dinooliva: Done.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
dinooliva: Done.

Are class comments necessary even though the class isn't public?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
dinooliva: I think that makes sense too.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
sduskis: I don't know offhand. I'm just concerned that this will change user behavior without user selecting to turn this on.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,197 @@
...
+  private static String encodeSpanContext(SpanContext ctxt) {
+    TraceId traceId = ctxt.getTraceId();
+    SpanId spanId = ctxt.getSpanId();
+    TraceOptions traceOpts = ctxt.getTraceOptions();
+    if (traceOpts.isSampled()) {
+      return "";
+    }
+    return "traceid=" + traceId.toLowerBase16()
+        + "&spanid=" + spanId.toLowerBase16()
+        + "&traceopt=" + (traceOpts.isSampled() ? "t&" : "f&");
```
sduskis: should this line now be `"&traceopt=f&"`?

dinooliva: I've changed the text encoding of span to use our standard TraceContext text format.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
dinooliva: The library only has a dependency on the OpenCensus API library, which means this code is a no-op. To enable propagation, an application will need to explicitly add a dependency on an OpenCensus Impl library. This is how configuration is done for other Java cloud client libraries (e.g. cloud bigtable and spanner).

sduskis:
@igorbernstein2, we need someway to configure whether or not to turn it on.  Can we piggyback your tracing configuration settings for this?

igorbernstein2:
I wasn't planning on letting end users configure tracing per service. The gapic stub settings are meant for service owners to mark their client as ready for tracing. My understanding is that if opencensus-impl is on the classpath and an exporter is configured all components should contribute traces.

I think forcing the user to opt-in per service is going to create too much friction. Why is this different from grpc auto creating spans?

JustinBeckwith:
@sduskis what is this blocked on?

sduskis:
@JustinBeckwith, we need approval from the Pub/Sub team. and we also need a way to configure whether or not this feature is active.

dinooliva:
Note: I developed the design of the OpenCensus Pub/Sub integration in conjunction with the Cloud Pub/Sub team.

Also, as previously noted, similar to Google Cloud Bigtable and Spanner (and other libraries e.g. gRPC), this feature is inactive unless a dependency on an OpenCensus implementation is added to the application.

google-cloud-clients/google-cloud-pubsub/pom.xml
```diff
@@ -78,6 +78,12 @@
       <artifactId>grpc-google-iam-v1</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
sduskis: Do we really need to add `opencensus-impl` here?

bogdandrutu: We should not, only the final app should do this.

dinooliva: These are the test dependencies rather than the library dependencies. I use the implementation to test OpenCensusUtil.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
```diff
@@ -0,0 +1,182 @@
...
+  public static final String TAG_CONTEXT_KEY = "googclient_OpenCensusTagContextKey";
+  public static final String TRACE_CONTEXT_KEY = "googclient_OpenCensusTraceContextKey";
+  @VisibleForTesting static final String MESSAGE_RECEIVER_SPAN_NAME = "OpenCensusMessageReceiver";
+  private static final String TRACEPARENT_KEY = "traceparent";
+
+  private static final Tagger tagger = Tags.getTagger();
+  private static final TagContextBinarySerializer serializer =
+      Tags.getTagPropagationComponent().getBinarySerializer();
+
+  private static final TraceOptions SAMPLED = TraceOptions.builder().setIsSampled(true).build();
```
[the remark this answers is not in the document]
dinooliva: Done.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/OpenCensusUtil.java
dinooliva: Yes, done.

dinooliva:
Thanks!

PR #4247, opened 2018-12-22
Fix NPE in BQ's StandardTableDefinition

nktaushanov:
The StandardTableDefinition.toPb() method throws a NPE if the Streamingbuffer fields are null. This seems to happen for some tables.


nktaushanov:
The failing integration test doesn't really seem related. Is there anything I can do about it or it is a known issue?

sduskis:
@nktaushanov, I'll run the tests one more time.  If they fail in a non-related way, I'll merge this.

nktaushanov:
@sduskis thanks!

PR #4250, opened 2018-12-24
Pubsub added message abandonment

### LINKED ISSUE #3854: PubSub : add the concept of message abandonment
`PusubMessage` / `AckReplyConsumer` needs the concept of "abandoning" a message.

Abandoning a message would mean that the message ack deadline would no longer be auto-extended.

```
public void receiveMessage(final PubsubMessage message, final AckReplyConsumer consumer) {
	try {

		// message processing

		consumer.ack();
	} catch (Throwable t) {
		// In an effort to prevent a tight message delivery / processing loop,
		// we want to abandon the message (stop extending the ack deadline and
		// let the message expire "naturally")
		
		consumer.abandon();
	}
}
```

charlesliqlogic:
Fixes #3854


chingor13:
Logging integration failure is unrelated

PR #4258, opened 2018-12-26
 Bigtable admin: deprecate typesafe names part 2

igorbernstein2:
This is a followup on #4257:
Unfortunately typesafe names create more issues for bigtable than they solve. A lot of the time users need to use both bigtable data & admin apis. Unfortunately, those apis define the same names in different namespaces. This forces users to use fully qualified names which is ugly. Also, absolute names don't benefit the bigtable client because the client is anchored at the instance and all names end up being relative


igorbernstein2:
The test failures are unrelated spanner tests

google-cloud-clients/google-cloud-bigtable-admin/src/main/java/com/google/cloud/bigtable/admin/v2/internal/NameUtil.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2018 Google LLC
```
sduskis: Should this be 2019?

igorbernstein2: fixed

igorbernstein2:
test failures are due to unrelated logging tests failing

PR #4264, opened 2018-12-27
Bigtable: add a separate callable for point reads

igorbernstein2:
This allows us to define separate timeouts for point reads & scans

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -436,7 +436,7 @@ public class BigtableDataClient implements AutoCloseable {
    *   String tableId = "[TABLE]";
    *
    *   Query query = Query.create(tableId)
-   *          .range("[START KEY]", "[END KEY]")
+   *          .rowKey("[KEY]")
    *          .filter(FILTERS.qualifier().regex("[COLUMN PREFIX].*"));
    *
    *   // Synchronous invocation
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -422,7 +422,88 @@ public Row readRow(String tableId, ByteString rowKey, @Nullable Filter filter) {
...
+   *
+   * <p>Sample code:
+   *
+   * <pre>{@code
+   * InstanceName instanceName = InstanceName.of("[PROJECT]", "[INSTANCE]");
+   * try (BigtableDataClient bigtableDataClient = BigtableDataClient.create(instanceName)) {
+   *   String tableId = "[TABLE]";
+   *
+   *   Query query = Query.create(tableId)
+   *          .range("[START KEY]", "[END KEY]")
```
sduskis: This seems wrong.  Shouldn't this be a point read?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -422,7 +422,88 @@ public Row readRow(String tableId, ByteString rowKey, @Nullable Filter filter) {
...
```
sduskis: Why is this a range?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
igorbernstein2: oops, copy/paste error.
Fixed now

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
igorbernstein2: oops, copy/paste error.
Fixed now

igorbernstein2:
Updated, PTAL

PR #4267, opened 2018-12-28
Add support for same-name file-and-directory

### LINKED ISSUE #3918: GCS NIO readAttributes returns CloudStorageObjectAttributes for directory like objects

Please include as much information as possible:

#### Environment details

- google-cloud-java version(s): google-cloud-nio:0.69.0-alpha

#### Steps to reproduce

1. In the GCS UI within a bucket, press create folder button and create a folder called `dir`
2. Upload a file within `dir` call this `fileA`

#### Code snippet

```java
    FileSystem fs = ....
    Files.walkFileTree(fs.getPath("dir") new SimpleFileVisitor<Path>() {
      @Override
      public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
        System.out.println(file);
        return FileVisitResult.CONTINUE;
      }
```

The GCS UI creates a 0 byte object called `dir/` This causes `readAttributes` to return https://github.com/googleapis/google-cloud-java/blob/v0.69.0/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java#L765, which prevents further traversal of objects within that directory. 

Does it make sense to do a check like https://github.com/googleapis/google-cloud-java/blob/v0.69.0/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java#L756 `usePseudoDirectories` is enabled and it is a 0 byte object?

I saw this related merged PR https://github.com/googleapis/google-cloud-java/pull/3775/files by  @jean-philippe-martin

jean-philippe-martin:
Update for the corner case of a file with a name that ends with '/', and a folder with the same name. As discussed in the issue, this can happen if a user presses "create folder" in the GCS web UI.

We now return true for both isRegularFile and isDirectory in that case if pseudodirs are enabled (as they are by default). If pseudodir is disabled then it just shows up as a file.

Fixes #3918


jean-philippe-martin:
I see one of the errors is:

```
[INFO] Processed 45 files (3 non-complying).
[ERROR] Found 3 non-complying files, failing build
[ERROR] To fix formatting errors, run "mvn fmt:format"
[ERROR] Non complying file: /tmpfs/src/github/google-cloud-java/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileAndDirectoryAttributes.java
[ERROR] Non complying file: /tmpfs/src/github/google-cloud-java/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/CloudStorageFileSystemProvider.java
[ERROR] Non complying file: /tmpfs/src/github/google-cloud-java/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/it/ITGcsNio.java
```

I tried running that but got: "[ERROR] No plugin found for prefix 'fmt' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/usr/local/google/home/jpmartin/.m2/repository), central (https://repo.maven.apache.org/maven2)] -> [Help 1]"

I already installed the files you recommend for intelliJ formatting but they don't work for me. I asked for help but never got it.

I can fix the files by hand, but I'm going to need to know what I should change.

chingor13:
Sorry, the plugin we're using reports the wrong command to run. It should be `mvn com.coveo:fmt-maven-plugin:format`

jean-philippe-martin:
Thank you very much, @chingor13! That worked!

jean-philippe-martin:
It looks like not all the tests reran? What do I need to do to have them run again?

jean-philippe-martin:
No point in continuing with this branch since the competing fix at #3918 was merged. I still think we should return a file-and-dir object but it won't be this commit as is.

PR #4270, opened 2018-12-31
3540: Filtered out com.fasterxml.jackson.core and org.threeten.bp.zon…

### LINKED ISSUE #3540: META-INF/services/ files are incorrect in shaded google-cloud-nio jar
The `google-cloud-nio-0.56.0-alpha-shaded.jar` doesn't correctly update the services files for shaded dependencies.  
Specifically the two files 
`META-INF/services/org.threeten.bp.zone.TzdbZoneRulesProvider`
and 
`META-INF/services/com.fasterxml.jackson.core.JsonFactory` should probably be updated to point to the shaded version of those classes instead.

andrey-qlogic:
…e from META-INF/services

Fixes #3540


hzyi-google:
I think the idea is to have the classes under META-INF/services point to the shaded versions instead of filtering them out. E.g., `com.fasterxml.jackson.core.JsonFactory` should be `shaded.cloud_nio.com.fasterxml.jackson.core.JsonFactory` instead.

igorbernstein2:
I think you want to use a transformer instead:
https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html#ServicesResourceTransformer

PR #4285, opened 2019-01-03
Pubsub: Add public api to publish list of messages

### LINKED ISSUE #3518: Publish batch operation for pub sub
We have an api which wraps pub sub, we accept messages one at a time, do some light processing and put them on pub sub. Are clients are slow because they need to wait  one round trip before sending the next. It would be better if they could send us a large batch and in the api we put them all on to pubsub as a batch. 

The java api hides pub subs batch write message behind a queuing / buffering layer. While this is nice for some use-cases it would be good if there was a simple way to publish a batch of messages to pub sub atomically. If we were to accept a batch from our client and pass them to the java api it could get split across two batches, and a partial write could occur

ajaaym:
Fixes #3518 Publish batch operation for pub sub


ajaaym:
@jadekler can you please review.

sduskis:
@kolea2, what's the status of this PR?

kolea2:
Apologies for the delay here. IIRC there is a workaround that was posted on the original issue - this is a change that would need to be coordinated across all other languages before merging.

sduskis:
I'm going to close this, since the Pub/Sub team has not yet prioritized this.  We can revisit this PR later.

PR #4287, opened 2019-01-03
updated links

romlogic:
The link was throwing up a "page not found" error. so, replaced it.


sduskis:
@romlogic, this PR has conflicts that need to be resloved.

PR #4297, opened 2019-01-07
Revert "Pub/Sub: Add message abandonment (#4250)"

chingor13:
This reverts commit 6e2c2dd890fe445cc896f0ddcb30309213759ee2.

This should not have been merged as it adds unsupported functionality to the pubsub client.


amohabir:
So will it not be supported at all or just not supported until a proper implementation and review is done?

nhartner:
Please bring this back. Reverting this is a step in the wrong direction. AWS doesn't make message redelivery so difficult. Why does GCP?

nhartner:
Please bring this back. This is a very useful feature. Why can't this supported functionality?

dhoard:
@chingor13 Can you provide an insight as to why this was removed?

PR #4303, opened 2019-01-08
4152: Added checking PAGE_TOKEN from options

### LINKED ISSUE #4152: Inconsistent behavior between LocalStorageHelper and real google storage

Please include as much information as possible:

#### Environment details

- OS: Mac
- Java version: 8
- google-cloud-java version(s):
  val googleStorageNew: ModuleID = "com.google.cloud" % "google-cloud-storage" % "1.53.0"
  val googleStorageLocal: ModuleID = "com.google.cloud" % "google-cloud-nio" % "0.71.0-alpha" % "test"

#### Steps to reproduce

**Code is in scala**
```
val page = db.list(bucketName.value, BlobListOption.prefix(objectNamePrefix), BlobListOption.pageSize(1), BlobListOption.currentDirectory())
```
if `db` is `LocalStorageHelper`, `page.iterateAll()` only gives me one page of object; if `db` is `com.google.cloud.storage.Storage`, then `page.iterateAll` will indeed give me multiple pages as documented.

#### Stacktrace

```
Any relevant stacktrace here.
```

#### Code snippet

```java
Any relevant code snippet to help reproduce the issue.
```

- ?

Following these steps will guarantee the quickest resolution possible.

Thanks!

andrey-qlogic:
Fixes #4152

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
@@ -123,9 +123,13 @@ class FakeStorageRpc implements StorageRpc {
       throws StorageException {
     String delimiter = null;
     String preprefix = "";
+    String pageToken = null;
     long maxResults = Long.MAX_VALUE;
     for (Map.Entry<Option, ?> e : options.entrySet()) {
       switch (e.getKey()) {
+        case PAGE_TOKEN:
+          pageToken = (String) e.getValue();
+          break;
         case PREFIX:
           preprefix = (String) e.getValue();
           if (preprefix.startsWith("/")) {
@@ -173,7 +177,7 @@ class FakeStorageRpc implements StorageRpc {
     // again).
     // The type cast seems to be necessary to help Java's typesystem remember that collections are
     // iterable.
-    return Tuple.of(null, (Iterable<StorageObject>) values);
+    return Tuple.of(pageToken, (Iterable<StorageObject>) values);
   }
 
   /** Returns the requested bucket or {@code null} if not found. */


andrey-qlogic:
@frankyn please can you take a look.

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
```diff
@@ -173,7 +177,7 @@ public StorageObject create(StorageObject object, InputStream content, Map<Optio
     // again).
     // The type cast seems to be necessary to help Java's typesystem remember that collections are
     // iterable.
-    return Tuple.of(null, (Iterable<StorageObject>) values);
+    return Tuple.of(pageToken, (Iterable<StorageObject>) values);
```
frankyn: I'm not sure if I understand replacing `null` with `pageToken` could explain this bit?

In addition, please add a unit test to exercise this change.

andrey-qlogic: We have the issue [#4152](https://github.com/googleapis/google-cloud-java/issues/4152) saying that there an inconsistency of LocalStorageHelper and real storage. 

The real implementation of StorageRpc list method in HttpStorageRpc class https://github.com/googleapis/google-cloud-java/blob/06b0ed50e31e9bef72ffac1788e0ec9ff9b9b6dd/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L329 returns 

return Tuple.<String, Iterable<Bucket>>of(buckets.getNextPageToken(), buckets.getItems());

that allows have multiple pages, 
but the the list method in FakeStorageRpc  https://github.com/googleapis/google-cloud-java/blob/06b0ed50e31e9bef72ffac1788e0ec9ff9b9b6dd/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java#L176
 always returns null, that means only one page.

The PR above allows removing the inconsistance. 
Regarding a unit test, I'm not sure that a new unit test may cover that case, I would appreciate your input.

frankyn: Thanks for clarifying and ignore my request for a unit test.

PR #4305, opened 2019-01-08
3540: Used transformer for shaded plugin

### LINKED ISSUE #3540: META-INF/services/ files are incorrect in shaded google-cloud-nio jar

andrey-qlogic:
Fixes #3540


andrey-qlogic:
@hzyi-google, please take a look, instead of #4270

PR #4313, opened 2019-01-09
Cloud Bigtable: tableAdmin sample

elisheva-qlogic:


google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
```diff
@@ -0,0 +1,373 @@
...
+    // [END bigtable_create_table]
+  }
+
+  /** Demonstrates how to list all tables within an instance. */
+  public void listAllTables() {
+    System.out.println("\nListing tables in current instance");
+    // [START bigtable_list_tables]
+    // Lists tables in the current instance.
+    try {
+      List<String> listTables = adminClient.listTables();
```
igorbernstein2: s/listTables/tableIds

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
```diff
@@ -0,0 +1,373 @@
...
+  public void getTableMeta() {
+    System.out.println("\nPrinting table metadata");
+    // [START bigtable_get_table_metadata]
+    // Gets table metadata, and applies a view to the table fields.
+    try {
+      Table table = adminClient.getTable(tableId);
+      System.out.println("Table: " + table.getId());
+      Collection<ColumnFamily> columnFamilies = table.getColumnFamilies();
+      for (ColumnFamily columnFamily : columnFamilies) {
+        printColumnFamily(columnFamily);
```
igorbernstein2: Although it will lead to a bit of duplication, I think it would be better to inline this function. I think the snippets should be mostly self contained

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
```diff
@@ -0,0 +1,373 @@
...
+
+  /** Demonstrates how to create a new instance of the VersionRule. */
+  public void maxVersionsRule() {
+    System.out.printf("%nCreating column family %s with max versions GC rule%n", COLUMN_FAMILY_2);
+    // [START bigtable_create_family_gc_max_versions]
+    // Creates a column family with GC policy : most recent N versions
+    // where 1 = most recent version
+
+    // Defines the GC policy to retain only the most recent 2 versions.
+    GCRules.VersionRule versionRule1 = GCRules.GCRULES.maxVersions(2);
```
igorbernstein2: The variable names are confusing, why is versionRule1 used for columnFamiliesRequest2?
If I understand correctly, the docs will only contain the snippet between `[START bigtable_create_family_gc_max_versions]` and `[END bigtable_create_family_gc_max_versions]`, so the reader will have very little context about the overall naming scheme. I would try to keep the names meaningful within the snippet.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
```diff
@@ -0,0 +1,373 @@
...
+      }
+    } catch (NotFoundException e) {
+      System.err.println(
+          "Failed to retrieve table metadata for a non-existent table: " + e.getMessage());
+    }
+    // [END bigtable_get_table_metadata]
+  }
+
+  /** Demonstrates how to create a new instance of the DurationRule. */
+  public void maxAgeRule() {
```
igorbernstein2: The method naming is inconsistent for the rules. All other methods are `verbNoun`.  These methods add families, so it should be something like addFamilyWithMaxAgeRule

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
igorbernstein2: Also, please use a static import for `GCRULES`

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
```diff
@@ -0,0 +1,373 @@
...
+  public void unionRule() {
+    System.out.printf("%nCreating column family %s with union GC rule%n", COLUMN_FAMILY_3);
+    // [START bigtable_create_family_gc_union]
+    // Creates a column family with GC policy to drop data that matches at least one condition.
+
+    // Defines a GC rule to drop cells older than 5 days OR not the most recent version.
+    GCRules.DurationRule maxAgeRule2 = GCRules.GCRULES.maxAge(5, TimeUnit.DAYS);
+    GCRules.VersionRule versionRule2 = GCRules.GCRULES.maxVersions(1);
+    // Add rules to union rule list
+    GCRules.UnionRule unionRule1 = GCRules.GCRULES.union().rule(maxAgeRule2).rule(versionRule2);
```
igorbernstein2: Please import UnionRule directly

Also please inline the creation of the child rules:

```
UnionRule unionRUle = GCRULES.union()
  .rule(GCRULES.maxAge(5, TimeUnit.DAYS))
  .rule(GCRULES.maxVersions(1))
```

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
```diff
@@ -0,0 +1,373 @@
...
+    GCRules.IntersectionRule intersectionRule1 =
+        GCRules.GCRULES.intersection().rule(maxAgeRule3).rule(versionRule3);
+
+    // Creates column family with given GC rule.
+    try {
+      ModifyColumnFamiliesRequest columnFamiliesRequest4 =
+          ModifyColumnFamiliesRequest.of(tableId).addFamily(COLUMN_FAMILY_4, intersectionRule1);
+      adminClient.modifyFamilies(columnFamiliesRequest4);
+      System.out.println("Created column family: " + COLUMN_FAMILY_4);
+    } catch (AlreadyExistsException e) {
```
igorbernstein2: A modify shouldnt throw an AlreadyExistsException

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
```diff
@@ -0,0 +1,373 @@
...
+      }
+    } catch (NotFoundException e) {
+      System.err.println(
+          "Failed to list column families from a non-existent table: " + e.getMessage());
+    }
+    // [END bigtable_list_column_families]
+  }
+
+  /** Demonstrates how to modify a column family's rule. */
+  public void modifyColumnFamilyRule() {
```
igorbernstein2: How is this different then the examples above?

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITTableAdminExample.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  private static void garbageCollect() {
+    Pattern timestampPattern = Pattern.compile(TABLE_PREFIX + "-([0-9a-f]+)-([0-9a-f]+)");
+    for (String tableId : adminClient.listTables()) {
+      Matcher matcher = timestampPattern.matcher(tableId);
+      if (!matcher.matches()) {
+        continue;
+      }
+      System.out.println("\nGarbage collecting orphaned table: " + tableId);
```
igorbernstein2: make sure to only garbage collect tables older than 10 minutes

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITTableAdminExample.java
```diff
@@ -0,0 +1,216 @@
...
+  @Test
+  public void testRunDoesNotFail() {
+    tableAdmin.run();
+  }
+
+  // TODO: add test for tableAdmin.listAllTables()
+  // TODO: add test for tableAdmin.getTableMeta()
+  // TODO: add test for tableAdmin.listColumnFamilies()
+
+  private boolean ruleCheck(Object condition) {
```
igorbernstein2: Why Object?

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
elisheva-qlogic: `ModifyColumnFamiliesRequest` is also used for **adding** a new columnFamily after the table is already created, not just modifying. So, if someone tries to add a columnFamily with the same name as an existing one, `AlreadyExistsException` will be thrown.

google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable/TableAdminExample.java
elisheva-qlogic: This method is modifying a columnFamily's rule, as opposed to the above methods are actually adding a columnFamily with a rule to the table. 
Above it does: ModifyColumnFamiliesRequest.of(tableId).**addFamily**(COLUMN_FAMILY, rule). 
Here is does: ModifyColumnFamiliesRequest.of(tableId).**updateFamily**(COLUMN_FAMILY, rule)

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITTableAdminExample.java
```diff
@@ -0,0 +1,224 @@
...
+
+  private static void garbageCollect() {
+    Pattern timestampPattern = Pattern.compile(TABLE_PREFIX + "-([0-9a-f]+)-([0-9a-f]+)");
+    for (String tableId : adminClient.listTables()) {
+      Matcher matcher = timestampPattern.matcher(tableId);
+      if (!matcher.matches()) {
+        continue;
+      }
+      String timestampStr = matcher.group(1);
+      long timestamp = Long.parseLong(timestampStr);
```
igorbernstein2: The time stamps are base 16

billyjacobson:
This is looking pretty good, will just need the TODO tests

google-cloud-examples/src/test/java/com/google/cloud/examples/bigtable/ITTableAdminExample.java
```diff
@@ -0,0 +1,224 @@
...
+
+  @BeforeClass
+  public static void beforeClass() throws IOException {
+    String targetInstance = System.getProperty(INSTANCE_PROPERTY_NAME);
+    if (targetInstance == null) {
+      adminClient = null;
+      return;
+    }
+    instanceName = InstanceName.parse(targetInstance).getInstance();
+    projectName = InstanceName.parse(targetInstance).getProject();
```
igorbernstein2: Can you update this to use ids like in #4339

PR #4314, opened 2019-01-09
3822: Fixed setDisableGZipContent call

### LINKED ISSUE #3822: Give users the ability to disable gzip content encoding to increase throughput

andrey-qlogic:
Fixes #3822

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -288,7 +288,9 @@ public class HttpStorageRpc implements StorageRpc {
                   new InputStreamContent(storageObject.getContentType(), content));
       insert.getMediaHttpUploader().setDirectUploadEnabled(true);
       Boolean disableGzipContent = Option.IF_DISABLE_GZIP_CONTENT.getBoolean(options);
-      if (disableGzipContent != null) insert.setDisableGZipContent(disableGzipContent);
+      if (disableGzipContent != null) {
+        insert.setDisableGZipContent(disableGzipContent);
+      }
       setEncryptionHeaders(insert.getRequestHeaders(), ENCRYPTION_KEY_PREFIX, options);
       return insert
           .setProjection(DEFAULT_PROJECTION)


andrey-qlogic:
@frankyn , would you please take a look

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -288,8 +288,7 @@ public StorageObject create(
                   new InputStreamContent(storageObject.getContentType(), content));
       insert.getMediaHttpUploader().setDirectUploadEnabled(true);
       Boolean disableGzipContent = Option.IF_DISABLE_GZIP_CONTENT.getBoolean(options);
-      if (disableGzipContent != null)
-        insert.getMediaHttpUploader().setDisableGZipContent(disableGzipContent);
+      if (disableGzipContent != null) insert.setDisableGZipContent(disableGzipContent);
```
sduskis: Please add curly braces around the body of the if statement

andrey-qlogic: Fixed

Issue #4241, opened 2018-12-19
 "Maven Central" badge shows old version


garrettjonesgoogle:
The "Maven Central" badge in the main README shows version `v0.47.0-alpha` as the latest version of google-cloud-java because it is based on the `google-cloud` artifact which was deprecated and no longer exists. The badge should be changed to look up a different version (since the latest is now v0.74.0).

chingor13:
Perhaps a single maven badge no longer makes sense and we should instead put a badge next to each API listed below (corresponding to the google-cloud-X version).

garrettjonesgoogle:
If this is going to be retained, I would suggest using the version of `google-cloud-bom`.

sduskis:
This was fixed by #4322

PR #4322, opened 2019-01-10
Update README version to use bom version

sduskis:
Improves the Readme towards issue #4241.

Later, we should add badges per library


sduskis:
It looks like `datastore-it` and `bigtableadmin-it` are flaky.

PR #4330, opened 2019-01-11
4117

### LINKED ISSUE #4117: blob.getMd5() doesn't work as expected
When I get any blob MD5 checksum using java API, the result is a Base64 string, that, when decoded, is not a MD5 string.

Expected result: the file MD5 checksum string.

Can anyone explain me why getMd5() returns a base64 of a MD5 instead of the MD5 itself?

#### Environment details

- OS: Linux
- Java version: 1.8.0_172
- google-cloud-java version(s): google-cloud-storage:1.53.0

#### Steps to reproduce

1. Save a blob using bucket.create(completeFilePath, fileByteArray)
2. Get the file metadata using storage.get(blobId)
3. Get the file MD5 using blob.getMd5()

andrey-qlogic:
Fixes #4117


andrey-qlogic:
@frankyn would you please take a look.

PR #4337, opened 2019-01-14
3929 Slow download performance for Storage API. Added new downloadToPathWithMediaHttpDownloader method with better performance.

### LINKED ISSUE #3929: Slow download performance for Storage API
The new version of the storage client (com.google.cloud:google-cloud-storage:1.52.0) appears download storage content at a MUCH slower rate than the legacy client (com.google.apis:google-api-services-storage:v1-rev141-1.25.0).

Legacy client (~40MB/s):

```
@Test
public void download() throws Exception {
    Storage storage = buildStorage();
    Storage.Objects.Get get = storage.objects().get(BUCKET_NAME, BUCKET_PATH);
    StorageObject storageObject = get.execute();

    File tempFile = createTempFile();
    try (OutputStream out = new FileOutputStream(tempFile)) {
        Stopwatch stopwatch = Stopwatch.createStarted();
        get.getMediaHttpDownloader().setDirectDownloadEnabled(true);
        get.executeMediaAndDownloadTo(out);
        long elapsedSeconds = stopwatch.stop().elapsed(TimeUnit.SECONDS);
        double fileSizeMb = storageObject.getSize().doubleValue() / 1024 / 1024;
        double throughput = fileSizeMb / elapsedSeconds;
        log.info("Completed download: elapsed={}s,fileSize={}MB,throughput={}MB/s",
                elapsedSeconds, fileSizeMb, throughput);
        assertTrue("Expected at least 20MB/s", throughput > 20);
    }
}
```

New client (~10MB/s):

```
@Test
   public void download() throws Exception {
       Storage storage = StorageOptions.getDefaultInstance().getService();
       Blob blob = storage.get(BLOB_ID);
       Path tempFile = createTempFile();

       Stopwatch stopwatch = Stopwatch.createStarted();
       blob.downloadTo(tempFile);
       long elapsedSeconds = stopwatch.stop().elapsed(TimeUnit.SECONDS);
       double fileSizeMb = blob.getSize().doubleValue() / 1024 / 1024;
       double throughput = fileSizeMb / elapsedSeconds;
       log.info("Completed download: elapsed={}s,fileSize={}MB,throughput={}MB/s",
               elapsedSeconds, fileSizeMb, throughput);
       assertTrue("Expected at least 20MB/s", throughput > 20);
   }
```

I'm attaching a couple of test cases that I've ran from a GCE instance (Ubuntu 16.04 with Java 1.8.0_191):

[storage-performance-legacy.zip](https://github.com/googleapis/google-cloud-java/files/2567343/storage-performance-legacy.zip)
[storage-performance-new.zip](https://github.com/googleapis/google-cloud-java/files/2567344/storage-performance-new.zip)

andrey-qlogic:
Fixes #3929 Added new **downloadToPathWithMediaHttpDownloader** method with better performance.

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -236,17 +236,19 @@ public class Blob extends BlobInfo {
    * @param options blob read options
    * @throws StorageException upon failure
    */
-  public void downloadTo(OutputStream outputStream, final BlobSourceOption... options) {
+  public void downloadTo(OutputStream outputStream, final BlobSourceOption... options)
+      throws StorageException {
     try (CountingOutputStream countingOutputStream = new CountingOutputStream(outputStream)) {
       final StorageRpc storageRpc = this.options.getStorageRpcV1();
-      final Map<StorageRpc.Option, ?> requestOptions = StorageImpl.optionMap(getBlobId(), options);
       runWithRetries(
           callable(
               new Runnable() {
                 @Override
                 public void run() {
                   storageRpc.readToOutputStream(
-                      getBlobId().toPb(), countingOutputStream, requestOptions);
+                      getBlobId().toPb(),
+                      countingOutputStream,
+                      StorageImpl.optionMap(getBlobId(), options));
                 }
               }),
           this.options.getRetrySettings(),
--- a/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
+++ b/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
@@ -32,6 +32,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import com.google.api.core.ApiClock;
 import com.google.api.gax.retrying.RetrySettings;
@@ -50,6 +51,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.CountingOutputStream;
 import java.io.File;
+import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URL;
 import java.nio.ByteBuffer;
@@ -626,23 +628,16 @@ public class BlobTest {
   @Test
   public void testDownloadToOutputStream() throws Exception {
     final byte[] expected = {1, 2};
-
     File file = File.createTempFile("blob", ".tmp");
     StorageRpc mockStorageRpc = createNiceMock(StorageRpc.class);
-
     expect(storage.getOptions()).andReturn(mockOptions).times(2);
     replay(storage);
-
     expect(mockOptions.getStorageRpcV1()).andReturn(mockStorageRpc);
-
     expect(mockOptions.getRetrySettings()).andReturn(RETRY_SETTINGS);
     expect(mockOptions.getClock()).andReturn(API_CLOCK);
-
     replay(mockOptions);
-
     storage.getOptions();
     blob = new Blob(storage, new BlobInfo.BuilderImpl(BLOB_INFO));
-
     expect(
             mockStorageRpc.readToOutputStream(
                 anyObject(StorageObject.class),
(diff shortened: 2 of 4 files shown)


andrey-qlogic:
@frankyn 
The approach is tied to an obsolete client, but the performance of the new method downloadToPathWithMediaHttpDownloader
looks better than the other downloadTo method. I would suggest adding this approach to the user separately from another method. If it looks reasonable, then I will add also unit tests.

sduskis:
@andrey-qlogic, can you please add a descriptive title?

mcantrell:
I would suggest creating a method that is decoupled from java.nio.file.Path. I would assume that a lot of users (myself included) would not have a file to write to. For example, you may want to stream the data from the storage API to a browser via HTTP response

```
    downloadWithMediaHttpDownloaderTo(OutputStream out)
```

and/or 
   
```
    downloadWithMediaHttpDownloaderTo(WritableChannel out)
```

andrey-qlogic:
@mcantrell 
Changing to downloadWithMediaHttpDownloaderTo(OutputStream out, ...) is more appreciate way than using WritableChannel.
But the definition of the next parameter becomes complicated for User we have either build [com.google.api.services.storage.Storage](https://github.com/googleapis/google-api-java-client-services/blob/d581badc85a1a5d119c185a872aacdfefb55b72e/clients/1.26.0/google-api-services-storage/v1beta2/com/google/api/services/storage/Storage.java#L17) with HttpStorageOptions that may not available in StorageOptions for getting MediaHttpDownloader.We have to make the second parameter com.google.api.services.storage.Storage and make the user responsible for building it the right way. I would appreciate your input here.

`public void downloadToPathWithMediaHttpDownloader(Path path, com.google.api.services.storage.Storage storage)`

mcantrell:
Sorry, I don't know the internals in detail but wouldn't the blob's storage object contain the transport options? Building the client like this?

```
  public void downloadWithMediaHttpDownloader(OutputStream targetOut) {
    com.google.api.services.storage.Storage storage = buildStorage(this.storage.getOptions());
```

mcantrell:
I suppose that another alternative to using the media downloader is to do something similar to the BlobReaderChannel. You can use this.storage.getOptions().getStorageRpcV1().read() to fetch the required bytes using the CountingOutputStream's offset and bytes remaining.

andrey-qlogic:
> Sorry, I don't know the internals in detail but wouldn't the blob's storage object contain the transport options? Building the client like this?
> 
> ```
>   public void downloadWithMediaHttpDownloader(OutputStream targetOut) {
>     com.google.api.services.storage.Storage storage = buildStorage(this.storage.getOptions());
> ```

the change to OutputStream is reasonable, but  building com.google.api.services.storage.Storage object from Blob may not be a good idea.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -225,6 +234,62 @@ public void downloadTo(Path path, BlobSourceOption... options) {
...
+          },
+          options.getRetrySettings(),
+          BaseService.EXCEPTION_HANDLER,
+          options.getClock());
+    } catch (RetryHelper.RetryHelperException e) {
+      throw StorageException.translateAndThrow(e);
+    } catch (IOException e) {
+      throw new StorageException(e);
+    }
+  }
```
frankyn: I appreciate your patience.

I'd recommend passing through an additional option (USE_DIRECT_DOWNLOAD) to `downloadTo()` if using `getMediaHttpDownloader`  is considered a breaking change:

Then in the underlying RPC class handle the request in
[HttpStorageRpc.getCall()](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L414) which called by [HttpStorageRpc.get()](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L430). 

WDYT?

andrey-qlogic: Added USE_DIRECT_DOWNLOAD to set directDownloadEnabled for MediaHttpDownloader

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -405,6 +414,12 @@ public Builder setEventBasedHold(Boolean eventBasedHold) {
       return this;
     }
 
+    @Override
+    public Builder setUseDirectDownload(Boolean useDirectDownload) {
+      infoBuilder.setUseDirectDownload(useDirectDownload);
+      return this;
+    }
+
```
frankyn: I don't think this is required.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -175,6 +177,13 @@ public static BlobSourceOption decryptionKey(String key) {
       return new BlobSourceOption(StorageRpc.Option.CUSTOMER_SUPPLIED_KEY, key);
     }
 
+    /**
+     * Returns an option to download blobs directly when you use the MediaHttpDownloader
```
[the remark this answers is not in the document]
andrey-qlogic: Fixed

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
```diff
@@ -123,6 +123,8 @@
           .build();
   private static final BlobInfo BLOB_INFO =
       BlobInfo.newBuilder("b", "n").setMetageneration(42L).build();
+  private static final BlobInfo BLOB_INFO_USE_DIRECT_DOWNLOAD =
+      BlobInfo.newBuilder("b", "n").setMetageneration(42L).setUseDirectDownload(true).build();
```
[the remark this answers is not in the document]
andrey-qlogic: Fixed

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
andrey-qlogic: Fixed.

andrey-qlogic:
I still need more benchmarking

andrey-qlogic:
@frankyn, it turned out that adding the new option USE_DIRECT_DOWNLOAD and setting MediaHttpDownloader directDownloadEnable to true/false depends on the option does not change the performance. The throughput still ~10Mb/s when the legacy client showed 40Mb/s. At the same time, if I increase buffer allocating size at the 'downloadTo' method https://github.com/googleapis/google-cloud-java/blob/9c47e1b4ff84f5ac07b0b38848e662c0e7eb3dad/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java#L217, it improves the performance up to 3x time on files of size 1Gb. What if instead, we add the next parameter to the method 'downloadTo' to give the user a chance to set the buffer size on his own?

frankyn:
Thanks @andrey-qlogic, given the new helper method `downloadTo(OutputStream)` WDYT about removing the code around `useDirectDownload`? The issue I see is that there are two overloads with different download expectations.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/StorageRpc.java
```diff
@@ -282,6 +284,14 @@ StorageObject compose(
    */
   Tuple<String, byte[]> read(StorageObject from, Map<Option, ?> options, long position, int bytes);
 
+  /**
+   * Reads from a storage object at the given position directly to outputstream
```
romlogic: "Reads ...to the output stream. "If you want to use outputstream as one word, do it consistently across. in line 293, you have used it as outputStream, here it is outputstream. However, I feel the meaning still stands when you use output stream instead.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -250,6 +289,16 @@ public void downloadTo(Path path) {
     downloadTo(path, new BlobSourceOption[0]);
   }
 
+  /**
+   * Downloads this blob to the given outputStream.
```
romlogic: If the word output stream conveys the literal meaning, please leave it expanded as "Downloads this blob to the given output stream." like line 245.

mcantrell:
> This fell through my email. Apologies for the delay. @mcantrell this is a trade-off between the two.
> 
> IIUC what you were asking for was removing the need to make multiple GET requests to a single GET request to mitigate the overhead. Is this correct?

Not exactly. I wanted reasonable performance compared to the deprecated client. I think it's a mistake to trade download performance for memory performance. This is a huge downgrade for us.

The multiple requests appears to be the cause. I would suggest again that resumable downloads is a better pattern for retries than what is currently implemented.

mcantrell:
Let's step back and talk about the practical implication of the proposed fix. To achieve reasonable download performance, I need 10x the memory. That means that for every 10 concurrent downloads I could handle before, I can now only handle 1. I'll need 10x the compute engines to handle out peak traffic.

mcantrell:
Sorry, I missed the comment regarding [commit](https://github.com/googleapis/google-cloud-java/pull/4337/commits/7b6f476e31bcdde82f23a941e0b19a5e41dd3d45). Just to be clear (there are a lot of threads going on here), we're talking about using a resume instead of increased buffer size to fix the issue?

sduskis:
@andrey-qlogic can you please answer this comment: https://github.com/googleapis/google-cloud-java/pull/4337#issuecomment-461072584

JustinBeckwith:
*gentle ping*

andrey-qlogic:
@JustinBeckwith, @sduskis, @mcantrell . That is correct, the PR is about using a resumable download with Http MediaDownloader instead of increase buffer size to fix the issue.

andrey-qlogic:
@sduskis , PTAL

sduskis:
@mcantrell, can you PTAL?  Your question was answered. @JesseLovelace or @frankyn: would you be able to review this?

sduskis:
@JesseLovelace or @frankyn: would you be able to review this?

sduskis:
@frankyn, this PR is stale.  I'm closing it.  Please reach out offline if you would like to restore this PR.

franDiazBitmover:
Hey there! Any news? I see this was closed like one month ago and I'm wondering whether this is addressed on another PR or if there are plans to fix this at all...

ajaaym:
@franDiazBitmover we closed this pr without merging it. We will have a new pr early next week addressing this issue.

PR #4338, opened 2019-01-14
BigQuery: Fix NPE for null table schema fields

chingor13:
Fixes internal bug 122674607


yazgoo:
Any ETA for this to be released ?

Thanks

PR #4336, opened 2019-01-14
Bump gax, grpc (1.17.1) & opencensus version

igorbernstein2:
Depends on https://github.com/googleapis/gax-java/pull/644 being merged & released

This updates gax to a new version that has the newest grpc version, updates opencensus to match and removes spanner's exclusion now that the versions align.

NOTE: ALthough grpc 1.18 has been released, I think we should still merge this before starting the next bump

PR #4343, opened 2019-01-15
WIP: debugging junit fork error

igorbernstein2:
As found in #4336:
Error: Could not find or load main class org.apache.maven.surefire.booter.ForkedBooter

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/pom.xml
+++ b/google-cloud-clients/google-cloud-bigtable/pom.xml
@@ -116,6 +116,7 @@
           <parallel>classes</parallel>
           <perCoreThreadCount>true</perCoreThreadCount>
           <threadCount>2</threadCount>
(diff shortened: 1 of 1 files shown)


igorbernstein2:
Reproduced the issue at head:
https://source.cloud.google.com/results/invocations/df22f10d-b06b-4658-9da3-f9597ede97fe/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fgoogle-cloud-java%2Fpresubmit%2Fjava8/log


Seems to be caused by:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911925
https://stackoverflow.com/questions/53010200/maven-surefire-could-not-find-forkedbooter-class

Trying to fix with `<useSystemClassLoader>false</useSystemClassLoader>`

PR #4344, opened 2019-01-15
Update maven-surefire-plugin to 3.0.0-M3 to fix Java 8 classloader

chingor13:
Fixes Java 8 builds which started failing with:

`Error: Could not find or load main class org.apache.maven.surefire.booter.ForkedBooter`


chingor13:
@sduskis This should fix the classloader issues. I tested against the "bad" java test image in #4345

PR #4392, opened 2019-01-18
Skip integration tests if no changes

chingor13:
For presubmit tests, skip integration tests unless code has changed in that branch.

Note that for nightly and continuous builds, we will run all the integration tests.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -17,9 +17,39 @@ set -eo pipefail
 
 cd github/google-cloud-java/
 
+function client_has_changes() {
+  CLIENT_NAME=$1
+  if [[ ! -z $(git diff master google-cloud-clients/google-cloud-core*) ]]; then
+    echo "true"
+    return
+  fi
+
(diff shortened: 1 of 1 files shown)


igorbernstein2:
Have you ever tried something like this?
https://github.com/lesfurets/partial-build-plugin

chingor13:
> Have you ever tried something like this?
> https://github.com/lesfurets/partial-build-plugin

I haven't, but I'll take a look.

chingor13:
In theory, that plugin will work well, but not as well as we'd like in our case.
1. It appears we need to run it at the root level of the project. Running maven from a google-cloud-X client can't see the generated code artifacts as they don't share a parent project. If run from the root and there are changes to a core library, it will run all the tests sequentially.
2. We still need to install all the artifacts before running the tests anyways (which takes > 8 minutes) before we know if we need to run tests.

chingor13:
The 2 sample PRs correctly limited the ITs run.

PR #4397, opened 2019-01-21
4117: Added Md5String() method and updated tests.

### LINKED ISSUE #4117: blob.getMd5() doesn't work as expected

andrey-qlogic:
Fixes #4117

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
@@ -35,13 +35,13 @@ import com.google.common.io.BaseEncoding;
 import java.io.Serializable;
 import java.math.BigInteger;
 import java.util.AbstractMap;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import javax.xml.bind.DatatypeConverter;
 
 /**
  * Google Storage object metadata.
@@ -443,11 +443,14 @@ public class BlobInfo implements Serializable {
     }
 
     public Builder setMd5FromHexString(String md5HexString) {
-      System.out.println("crc32cHexString: " + md5HexString);
       if (md5HexString == null) {
         return this;
       }
-      byte[] bytes = DatatypeConverter.parseHexBinary(md5HexString);
+      byte[] bytes = new BigInteger(md5HexString, 16).toByteArray();
+      int leadingEmptyBytes = bytes.length - md5HexString.length() / 2;
+      if (leadingEmptyBytes > 0) {
+        bytes = Arrays.copyOfRange(bytes, leadingEmptyBytes, bytes.length);
+      }
       this.md5 = BaseEncoding.base64().encode(bytes);
       return this;
     }
@@ -463,7 +466,11 @@ public class BlobInfo implements Serializable {
       if (crc32cHexString == null) {
         return this;
       }
-      byte[] bytes = DatatypeConverter.parseHexBinary(crc32cHexString);
+      byte[] bytes = new BigInteger(crc32cHexString, 16).toByteArray();
+      int leadingEmptyBytes = bytes.length - crc32cHexString.length() / 2;
+      if (leadingEmptyBytes > 0) {
+        bytes = Arrays.copyOfRange(bytes, leadingEmptyBytes, bytes.length);
+      }
       this.crc32c = BaseEncoding.base64().encode(bytes);
       return this;
     }


andrey-qlogic:
@frankyn, would you please take a look at this PR.

andrey-qlogic:
Added Crc32cString() method and updated tests. Added to snippets.

frankyn:
Hi @andrey-qlogic, Thanks for the update. I chatted with the GCS team about this design. It sounds good to them but suggested adding a Hex Hash String to GCS base64 encoded string for setCrc32c() method to the `Blob class`.

The naming convention would help: @googleapis/yoshi-java do you have input on Java community method names? 

WDYT about the following names?
```java
getMd5ToHexString()
setMd5FromHexString()

getCrc32cToHexString()
setCrc32cFromHexString()
```

andrey-qlogic:
Renamed new methods using naming convention getMd5ToHexString(), getCrc32cToHexString().
I believe that setter methods are not required as those new methods return a string representation md5 and cr32 that set in another method.

frankyn:
The other method [setMd5](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java#L422) for example expects the base64 encoded string of an Md5 hash. It would be nice to have a similar helper method to getting the md5 hash as a base64 decoded md5 hex string.

andrey-qlogic:
@frankyn, that is unclear, are you suggesting to add a new setter method to class BuilderImpl 
 like 
`public Builder setMd5FromHexString(String md5HexString)`
that will do something like 
`byte[] decoded = BaseEncoding.base64().decode(md5HexString);`

and assign it to md5 ?

Otherwise the method getMd5ToHexString is implepented for BlobInfo class in https://github.com/googleapis/google-cloud-java/pull/4397/commits/85588a3c990f2a080ed38ee4555711ca33f1522d

I would appreciate for your input

andrey-qlogic:
Added methods to set Md2 and Crc32s from hex strings.

andrey-qlogic:
there are code format issues that will be fixed with #4437

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -423,12 +441,37 @@ public Builder setMd5(String md5) {
...
 
+    public Builder setMd5FromHexString(String md5HexString) {
+      if (md5HexString == null) {
+        return this;
+      }
+      if (md5HexString.startsWith("0x")) {
+        md5HexString = md5HexString.replaceFirst("0x", "");
+      }
+      byte[] bytes = new BigInteger(md5HexString, 16).toByteArray();
+      this.md5 = "0x" + BaseEncoding.base64().omitPadding().encode(bytes);
```
frankyn: Why does this prepend `0x`? 

Could you add an integration that exercises the `setMd5/Crc32cFromHexString` methods?

andrey-qlogic:
@frankyn, I realized that Md5 sometimes used with 0x prefix that can be easily removed at BlobInfo class.

frankyn:
For this PR, @andrey-qlogic could you skip supporting the removal of "0x" for now? 
I'm leaning towards supporting that incrementally instead of in the first iteration.

Thank you, apologies for constant delays on these reviews.

andrey-qlogic:
@frankyn, do you want me to drop the whole last commit [https://github.com/googleapis/google-cloud-java/pull/4397/commits/7f19e33c3516c119f1a2775d0eb0fd07533a06c7](https://github.com/googleapis/google-cloud-java/pull/4397/commits/7f19e33c3516c119f1a2775d0eb0fd07533a06c7) ?

frankyn:
Yes except the integration test.

andrey-qlogic:
Reverted back the commit [https://github.com/googleapis/google-cloud-java/commit/7f19e33c3516c119f1a2775d0eb0fd07533a06c7](https://github.com/googleapis/google-cloud-java/commit/7f19e33c3516c119f1a2775d0eb0fd07533a06c7) except integration tests.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
frankyn: Please remove prepending "0x".

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -423,12 +441,37 @@ public Builder setMd5(String md5) {
       return this;
     }
 
+    public Builder setMd5FromHexString(String md5HexString) {
+      if (md5HexString == null) {
+        return this;
+      }
+      if (md5HexString.startsWith("0x")) {
+        md5HexString = md5HexString.replaceFirst("0x", "");
+      }
```
frankyn: Remove code that removes "0x" from the hex string.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
andrey-qlogic: @frankyn, do you mean the change for both Md5 and Crc32c, correct 
What about unit tests values? https://github.com/googleapis/google-cloud-java/blob/7de46b4ce194f4c7d7094dd1d7dec4f4a85ad363/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobInfoTest.java#L55

We allow user to enter the value with 0x and without.

andrey-qlogic:
@frankyn, do you mean the change for both Md5 and Crc32c, correct
What about unit tests values?

https://github.com/googleapis/google-cloud-java/blob/7de46b4ce194f4c7d7094dd1d7dec4f4a85ad363/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobInfoTest.java#L55

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
andrey-qlogic: Removed code that removes 0x from hex string and code that adds prepending 0x. OmitPadding removed

andrey-qlogic:
Removed code that removes 0x from hex string and code that adds prepending 0x. OmitPadding removed

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
```diff
@@ -71,11 +71,13 @@
   private static final String CONTENT_ENCODING = "UTF-8";
   private static final String CONTENT_LANGUAGE = "En";
   private static final String CRC32 = "0xFF00";
```
frankyn: I understand the confusion now. Sorry I didn't see this before, thanks for linking to it.

The hex values used for `CRC32` and `MD5` are not valid and only used for unit tests. I recommend replacing these values out with the base64 encoded crc32c and md5 values.

andrey-qlogic:
@frankyn , I realized that new BigInteger(md5HexString, 16).toByteArray() is not safe with leading zeros. I changed to DatatypeConverter.parseHexBinary(md5HexString). It works fine. If needed we can make an utility method instead.

andrey-qlogic:
thanks for the approval

PR #4408, opened 2019-01-25
[Storage] Bucket Policy Only samples

frankyn:
Samples for the GCS Bucket Policy Only feature. Includes some reformatting.

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
@@ -1538,9 +1538,8 @@ public class StorageSnippets {
     // The name of a bucket, e.g. "my-bucket"
     // String bucketName = "my-bucket";
 
-    BucketInfo.BucketPolicyOnly bucketPolicyOnly = BucketInfo.BucketPolicyOnly.newBuilder().setEnabled(true).build();
-    BucketInfo.IamConfiguration iamConfiguration = BucketInfo.IamConfiguration.newBuilder().setBucketPolicyOnly(
-        bucketPolicyOnly).build();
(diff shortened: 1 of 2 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1393,4 +1526,74 @@ public Blob releaseTemporaryHold(String bucketName, String blobName) throws Stor
...
+  /**
+   * Example of how to enable Bucket Policy Only for a bucket
+   */
+  public Bucket enableBucketPolicyOnly(String bucketName) throws StorageException {
+    // [START storage_enable_bucket_policy_only]
+    // Instantiate a Google Cloud Storage client
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+
+    // The name of a bucket, e.g. "my-bucket"
+    // String bucketName = "my-bucket";
```
kolea2: I think this should be uncommented?

kolea2: Oh nvm, I see it's being passed through as a parameter. Can we remove the passed in parameter and leave the code above for clarity?

frankyn: @kolea2, thanks for the feedback.

The passed in parameter is used by the integration test and the commented variable could be copy/pasted by a developer into their code. I'm using existing sample design followed by this file. 

Could you add an example of what you'd like to see?

kolea2: Ah, ok makes sense - didn't realize the parameters were used in that way. That seems like the best approach then!

frankyn:
Thanks @kolea2! I'm running kokoro which didn't run on this PR yet. Is that a bug?

PR #4413, opened 2019-01-26
Set connect/read timeout when calling the metadata service

### LINKED ISSUE #4398: calls to metadataConfig getAttribute can hang
https://github.com/googleapis/google-cloud-java/pull/4278 changed the hostname of the metadata service to be the FQDN. For whatever reason my companies internal dns resolvers actually return the the IP address of the metadata service.

From what I can gather the default timeout for `HttpURLConnection`[1] is forever which would explain why my service fails to startup when running locally. 

This happens to me when running the latest google-cloud-core which is 1.6.0.

It seems like it would be a sane thing to add a timeout to the HTTP request. I can submit a PR to do this if others agree.

[1] https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java#L73

dmichel1:
Otherwise callers of these methods can be waiting indefinitely.

Fixes #4398

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
+++ b/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/MetadataConfig.java
@@ -34,6 +34,7 @@ import java.net.URL;
 public class MetadataConfig {
 
   private static final String METADATA_URL = "http://metadata.google.internal/computeMetadata/v1/";
+  private static final int TIMEOUT_MS = 5000;
 
   private MetadataConfig() {}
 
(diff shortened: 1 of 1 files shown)


dmichel1:
@ajaaym can you take a look at this pls?

sduskis:
@dmichel1, can you please run `mvn com.coveo:fmt-maven-plugin:format` so that your code complies with our code formatting rule?

dmichel1:
@sduskis i rebased and pushed my code again - build should be passing.

dmichel1:
Unrelated IT tests are failing for this PR now.

sduskis:
I'll try to run the tests again, just to be sure.

dmichel1:
@sduskis yaya they are passing now. thanks

PR #4421, opened 2019-01-31
Bigtable: make row & cell ordering more consistent.

### LINKED ISSUE #4055: Bigtable: add helper methods to Rows to efficiently fetch qualifier values
A Row is currently a flat list of cells. The cells have a particular order: all cells for a column family are clustered (NOTE: column families are not sorted, only clustered.) and then sorted lexicographically by qualifier. To expose this behavior to the user we should have 3 methods that on the Row class:

```java
List<RowCell> getCells(String family)
List<RowCell> getCells(String family, ByteString qualifier)
List<RowCell> getCells(String family, String qualifier)
```

These methods should efficiently return a list of matching cells. To implement this efficiently, the `Row` class should include `Map` of index ranges for each family cluster, which should populated by the `RowMerger`. Then the 3 methods could use a combination of index lookups (for the family) and binary searches (for the qualifier) to return the matching cells.

igorbernstein2:
* RowCells should always be ordered lexicographically by family, then qualifier and finally by reverse chronological order
* Although rows will always be ordered lexicographically by row key, they should not implement Comparable to avoid confusion when compareTo() == 0 and equals() is false. Instead that ordering was moved to a separate comparator.

Note: this is an api breaking change due to Row no longer implementing comparable

Fixes #4055

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Row.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Row.java
@@ -31,8 +31,8 @@ import javax.annotation.Nullable;
 /**
  * Default representation of a logical row.
  *
- * <p>The cells contained within, will be sorted by the native order. Please see
- * {@link RowCell#compareByNative()} for details.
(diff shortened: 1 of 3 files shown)


sduskis:
Please fix the code format.

igorbernstein2:
The remaining formatting issues are unrelated

Issue #4440, opened 2019-02-04
Google Cloud components should not depend on `-android` versions


stevenschlansker:
Currently, the `google-cloud-logging-logback` component depends on Guava version `26.0-andriod`.  This is not appropriate for Google Cloud projects which are running on a full JDK in e.g. GKE.  The two different versions are not considered compatible by Maven and version checker plugins.

It looks like this issue was introduced in https://github.com/googleapis/google-cloud-java/pull/3980

#### Environment details

- OS: Fedora Linux 29
- Java version: JDK 11
- google-cloud-java version(s): `0.79.0-alpha`

#### Steps to reproduce

1. Use Guava
2. Import `google-cloud-logging-logback`
3. Run a build checker that ensures dependency convergence

#### Stacktrace

```
[INFO] --- maven-dependency-versions-check-plugin:2.0.2:check (default) @ mycorp-server ---
[INFO] Checking dependency versions
[ERROR] Found a problem with the direct dependency com.google.guava:guava of the current project
  Expected version is 26.0-jre
  Resolved version is 26.0-jre
  Version 26.0-android was expected by artifact: com.google.cloud:google-cloud-logging-logback
  Version 26.0-jre was expected by artifacts: com.mycorp.components:mycorp-jaxrs-client, com.mycorp.components:mycorp-spring
```

stevenschlansker:
Additionally, the docs specify that Google Cloud components aren't even compatible with Android:
> Note: Cloud Java client libraries do not currently support Android.
(https://cloud.google.com/logging/docs/setup/java)
so it's even stranger to depend on `26.0-android` version.

sduskis:
We base our guava version based on [grpc](https://github.com/grpc/grpc-java/), which currently [uses](https://github.com/grpc/grpc-java/blob/master/build.gradle#L112) 26.0-android.  

@ejona86, do you have any comments about the history of the selection of the guava version?

sduskis:
Would it make sense for us to exclude grpc's dependency on `-android` and use `-jre` instead?

sduskis:
Oh... I see.  @andreamlin [commented](https://github.com/googleapis/google-cloud-java/pull/3980#issuecomment-460449373):

> We chose the android flavor because it supports JDK 7. (See guava [README](https://github.com/google/guava)).

Our libraries need to support JDK 7, but the `-jre` version supports 8+ only.

ejona86:
[...]

PR #4441, opened 2019-02-04
Comments to explain guava:android version

andreamlin:
In response to https://github.com/googleapis/google-cloud-java/issues/4440, to explain why we chose the *-android flavor of Guava.


andreamlin:
PTAL

PR #4442, opened 2019-02-05
Regen vision with string format functions

andreamlin:


sduskis:
It looks like there was a code generation that happened later and was merged first.  I'm going to close this.

PR #4457, opened 2019-02-06
[bigtable] Addition of Row#fromProto 

rahulKQL:
Adding `Row#fromProto` to generate `models.Row` from protobuf `Row`.
Would need this feature for translation of model in `BigtableDataClientWrapper`, related in [GBT-2061](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/2061).

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Row.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Row.java
@@ -172,7 +172,7 @@ public abstract class Row implements Serializable {
     return index;
   }
 
-  /** wraps the protobuf {@link com.google.bigtable.v2.Row} */
+  /** Wraps the protobuf {@link com.google.bigtable.v2.Row} */
   public static Row fromProto(@Nonnull com.google.bigtable.v2.Row rowProto) {
     Preconditions.checkArgument(rowProto != null, "Row must not be null");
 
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Row.java
```diff
@@ -167,4 +171,30 @@ private int getLast(@Nonnull String family, @Nullable ByteString qualifier, int
...
+
+    for (Family family : rowProto.getFamiliesList()) {
+      for (Column column : family.getColumnsList()) {
+        for (Cell cell : column.getCellsList()) {
+          rowBuilder.startCell(
+              family.getName(),
+              column.getQualifier(),
+              cell.getTimestampMicros(),
+              cell.getLabelsList(),
+              column.getCellsCount());
```
sduskis: @igorbernstein2, what's the right value for "size" here?  I can't imagine that `column.getCellsCount()` is right.

rahulKQL: My apologies for adding `cellsCount` instead of `cell#getValue`'s `size`.

updated the same.

igorbernstein2:
I recently merged:
https://github.com/googleapis/google-cloud-java/pull/4458

Any particular reason you'd want to have this on the model?

sduskis:
We can use #4458 instead.

PR #4461, opened 2019-02-07
Adding Cloud Bigtable Mutation fromProto

sduskis:


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -62,6 +62,13 @@ public static Mutation create() {
     return new Mutation(false);
   }
 
+  /** Wraps the List of protobuf {@link com.google.bigtable.v2.Mutation}. */
+  public static Mutation fromProto(List<com.google.bigtable.v2.Mutation> protos) {
+    Mutation mutation = new Mutation(false);
```
igorbernstein2: i'm concerned about this assuming that all of the mutations contain serverside timestamps

sduskis: What would you suggest?  Should we add a boolean, and maybe a check for server-side timestamps?

sduskis: (`toProto` is required for migrating cloud-bigtable-client's BulkMutation infrastructure)

igorbernstein2: I think easiest would be call this `fromProtoUnsafe` and mark is as `InternalOnly` with similar warning as `createUnsafe`? Also, can you move the method to after createUnsafe()?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -73,6 +73,15 @@ public static Mutation createUnsafe() {
     return new Mutation(true);
   }
 
+
+  /** Wraps the List of protobuf {@link com.google.bigtable.v2.Mutation}. */
```
igorbernstein2: Please add a warning similar to createUnsafe:

It is the callers responsibility to make sure that all of the mutations have a deterministric timestamp. If a mutation has a serverside timestamp, then the mutation will no longer be idempotent, which might cause multiple duplicate values to be stored in Bigtable. This option should only be used for advanced usecases with extreme care.

sduskis: @igorbernstein2: done.

PR #4472, opened 2019-02-08
updating google-auth-library-java to 0.13.0

lbergelson:
updating the google-auth-library-java dependency from 0.12.0 to 0.13.0

It includes some changes to logging output that I would like in my downstream project.

### DIFF
--- a/google-cloud-clients/pom.xml
+++ b/google-cloud-clients/pom.xml
@@ -161,7 +161,7 @@
     <grpc.version>1.17.1</grpc.version>
     <!-- We are currently using the *-android version to support JDK7. -->
     <guava.version>26.0-android</guava.version>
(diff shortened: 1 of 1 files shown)


sduskis:
Please also upgrade com.google.http-client:google-http-client, as per the failing tests.

sduskis:
Upgrading seems to be more involved than I expected.  I'll move this forward in a different PR.

lbergelson:
@sduskis Ah, sorry about that.  The project doesn't build for me locally under normal circumstances due to dependency resolution issues,  so I didn't notice that there additional dependency issues due to my changes.  

I pushed one more with what I think are all the necessary version bumps, but as you said it's more complicated than I thought it would be and I don't know if there are any other issues.

PR #4474, opened 2019-02-08
Add ordered publishing support to Cloud Pub/Sub

kamalaboulhosn:
This is being added to the pubsub-ordering-keys branch only for now.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -45,11 +45,13 @@ import com.google.pubsub.v1.TopicName;
 import com.google.pubsub.v1.TopicNames;
 import java.io.IOException;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
@@ -131,9 +133,29 @@ public class Publisher {
     // We post-process this here to keep backward-compatibility.
     // Also, if "message ordering" is enabled, the publisher should retry sending the failed
     // message infinitely rather than sending the next one.
-    RetrySettings retrySettings = builder.retrySettings;
-    if (retrySettings.getMaxAttempts() == 0 || builder.enableMessageOrdering) {
-      retrySettings = retrySettings.toBuilder().setMaxAttempts(Integer.MAX_VALUE).build();
+    RetrySettings.Builder retrySettingsBuilder = builder.retrySettings.toBuilder();
+    if (retrySettingsBuilder.getMaxAttempts() == 0) {
+      retrySettingsBuilder.setMaxAttempts(Integer.MAX_VALUE);
+    }
+    if (enableMessageOrdering) {
+      retrySettingsBuilder
+          .setMaxAttempts(Integer.MAX_VALUE)
+          .setTotalTimeout(Duration.ofNanos(Long.MAX_VALUE));
+    }
+
+    Set<StatusCode.Code> retryCodes;
+    if (enableMessageOrdering) {
+      retryCodes = EnumSet.allOf(StatusCode.Code.class);
+    } else {
+      retryCodes =
+          EnumSet.of(
+              StatusCode.Code.ABORTED,
+              StatusCode.Code.CANCELLED,
+              StatusCode.Code.DEADLINE_EXCEEDED,
+              StatusCode.Code.INTERNAL,
+              StatusCode.Code.RESOURCE_EXHAUSTED,
+              StatusCode.Code.UNKNOWN,
+              StatusCode.Code.UNAVAILABLE);
     }
 
     PublisherStubSettings.Builder stubSettings =
@@ -143,15 +165,8 @@ public class Publisher {
             .setTransportChannelProvider(builder.channelProvider);
     stubSettings
         .publishSettings()
-        .setRetryableCodes(
-            StatusCode.Code.ABORTED,
-            StatusCode.Code.CANCELLED,
-            StatusCode.Code.DEADLINE_EXCEEDED,
-            StatusCode.Code.INTERNAL,
-            StatusCode.Code.RESOURCE_EXHAUSTED,
-            StatusCode.Code.UNKNOWN,
-            StatusCode.Code.UNAVAILABLE)
-        .setRetrySettings(retrySettings)
+        .setRetryableCodes(retryCodes)
+        .setRetrySettings(retrySettingsBuilder.build())
         .setBatchingSettings(BatchingSettings.newBuilder().setIsEnabled(false).build());
     this.publisherStub = GrpcPublisherStub.create(stubSettings.build());
 
@@ -200,7 +215,7 @@ public class Publisher {
     }
 
     final String orderingKey = message.getOrderingKey();
-    if (!orderingKey.isEmpty() && !enableMessageOrdering) {
+    if (orderingKey != null && !orderingKey.isEmpty() && !enableMessageOrdering) {
       throw new IllegalStateException(
           "Cannot publish a message with an ordering key when message ordeirng is not enabled.");
     }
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -124,8 +129,10 @@ private Publisher(Builder builder) throws IOException {
 
     // Publisher used to take maxAttempt == 0 to mean infinity, but to GAX it means don't retry.
     // We post-process this here to keep backward-compatibility.
+    // Also, if "message ordering" is enabled, the publisher should retry sending the failed
+    // message infinitely rather than sending the next one.
     RetrySettings retrySettings = builder.retrySettings;
-    if (retrySettings.getMaxAttempts() == 0) {
+    if (retrySettings.getMaxAttempts() == 0 || builder.enableMessageOrdering) {
```
kolea2: Here we're using builder.enableMessageOrdering and on line 203 we're using enableMessageOrdering. I would guess this one should change to enableMessageOrdering?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -238,29 +254,18 @@ public String getTopicNameString() {
...
       logger.log(Level.FINER, "Scheduling a batch for immediate sending.");
-      final OutstandingBatch finalBatchToSend = batchToSend;
-      executor.execute(
-          new Runnable() {
-            @Override
-            public void run() {
-              publishOutstandingBatch(finalBatchToSend);
-            }
-          });
+      publishOutstandingBatch(batchToSend);
```
kolea2: Apologies if I've already asked this, but can you please explain the change here and why we're calling publishOutstandingBatch and publishAllOutstanding?

google-api-grpc/proto-google-cloud-pubsub-v1/src/main/proto/google/pubsub/v1/pubsub.proto
```diff
@@ -399,6 +399,8 @@ message PubsubMessage {
   // it receives the `Publish` call. It must not be populated by the
   // publisher in a `Publish` call.
   google.protobuf.Timestamp publish_time = 4;
+
+  string ordering_key = 5;
```
ajaaym: missing description functionality and usage for this field.

google-api-grpc/proto-google-cloud-pubsub-v1/src/main/proto/google/pubsub/v1/pubsub.proto
```diff
@@ -599,6 +601,7 @@ message Subscription {
   // managing labels</a>.
   map<string, string> labels = 9;
 
+  bool enable_message_ordering = 10;
```
ajaaym: missing description functionality and usage for this field.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -124,8 +129,10 @@ private Publisher(Builder builder) throws IOException {
 
     // Publisher used to take maxAttempt == 0 to mean infinity, but to GAX it means don't retry.
     // We post-process this here to keep backward-compatibility.
+    // Also, if "message ordering" is enabled, the publisher should retry sending the failed
+    // message infinitely rather than sending the next one.
     RetrySettings retrySettings = builder.retrySettings;
-    if (retrySettings.getMaxAttempts() == 0) {
+    if (retrySettings.getMaxAttempts() == 0 || builder.enableMessageOrdering) {
       retrySettings = retrySettings.toBuilder().setMaxAttempts(Integer.MAX_VALUE).build();
```
ajaaym: There is a total rpc timeout of 10 sec in addition to max attempts which will cause the rpc to stop retrying. RPC will also fail on non retryable error codes how that should be handled? Also doesnt isnt it to fail all pending batch for this ordering key rather than retrying indefinitely and hanging application thread if there are more concurrent rpcs in progress than number of gax thread.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -192,42 +199,51 @@ public String getTopicNameString() {
       throw new IllegalStateException("Cannot publish on a shut-down publisher.");
     }
 
+    final String orderingKey = message.getOrderingKey();
+    if (!orderingKey.isEmpty() && !enableMessageOrdering) {
```
ajaaym: what if ordering key is null and ordering is not enabled?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -192,42 +199,51 @@ public String getTopicNameString() {
...
+      throw new IllegalStateException(
+          "Cannot publish a message with an ordering key when message ordeirng is not enabled.");
+    }
+
     final int messageSize = message.getSerializedSize();
     OutstandingBatch batchToSend = null;
     SettableApiFuture<String> publishResult = SettableApiFuture.<String>create();
     final OutstandingPublish outstandingPublish = new OutstandingPublish(publishResult, message);
     messagesBatchLock.lock();
```
ajaaym: better to use lock per ordering key.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -353,20 +362,44 @@ public void onFailure(Throwable t) {
               messagesWaiter.incrementPendingMessages(-outstandingBatch.size());
             }
           }
-        });
+        };
+
+    if (outstandingBatch.orderingKey.isEmpty()) {
```
ajaaym: check for null

google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/PublisherImplTest.java
```diff
@@ -241,6 +243,177 @@ public void testPublishMixedSizeAndDuration() throws Exception {
...
+    assertFalse(publishFuture3.isDone());
+    assertFalse(publishFuture4.isDone());
+
+    // One of the batches reaches the limit.
+    ApiFuture<String> publishFuture5 = sendTestMessageWithOrderingKey(publisher, "m5", "OrderA");
+    // Verify that they were delivered in order per ordering key.
+    assertTrue(Integer.parseInt(publishFuture1.get()) < Integer.parseInt(publishFuture3.get()));
+    assertTrue(Integer.parseInt(publishFuture2.get()) < Integer.parseInt(publishFuture4.get()));
+    assertTrue(Integer.parseInt(publishFuture3.get()) < Integer.parseInt(publishFuture5.get()));
+
```
ajaaym: I think correct check should be Integer.parseInt(publishFuture5.get()) < Integer.parseInt(publishFuture2.get()) if batch with ordering key "OrderA" is sent first or Integer.parseInt(publishFuture2.get()) < Integer.parseInt(publishFuture1.get()) if batch with orderking key "OrderB" is sent first?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
kamalaboulhosn: The fact that they are in this order is the bug that @kimkyung-goog mentioned. We actually need to reverse the order (and have a bug to do so that he will take care of). The idea is that we'd need to flush all outstanding messages and also flush this message if it is too big to fit within the confines of a single batch.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
kamalaboulhosn: There is a comment in `publishAllOutstanding` that says unlocking messagesBatchLock there could result in out-of-order messaging. Even if we have a per-ordering-key lock, we'll still need the global one as well in order to insert into `messageBatches`. I think the move to more fine-grained locking is a performance enhancement we can look into as we run more tests and after a change to avoid premature batching, which will come soon. For now, I want to make sure we get the correctness right and understand the scope of the work being undertaken so we can start it in client libraries in the other languages.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
ajaaym: I think some kind of interface providing feedback to application about the failure and let application take the decision whether to continue or to stop publishing for that ordering key would be better, rather than keep retrying for all error code indefinitely.

kolea2:
LGTM. CC @jadekler to see if he has any additional comments. In the meantime looking into running the precommit checks for this PR.

kolea2:
Spanner build failure unrelated

PR #4477, opened 2019-02-11
Add MDC support in Logback appender

dietervdw-spotify:
Logback MDC values are a perfect fit to propagate as labels on log messages.
Let's do this by default?

Alternative is to use a `LoggingEventEnhancer`, but this seems like such a common and expected use-case I'd argue making it default behaviour seems sensible.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
@@ -36,6 +36,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -231,6 +232,10 @@ public class LoggingAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
     builder
         .addLabel(LEVEL_NAME_KEY, level.toString())
         .addLabel(LEVEL_VALUE_KEY, String.valueOf(level.toInt()));
+
+    for (Map.Entry<String, String> entry : e.getMDCPropertyMap().entrySet()) {
+      builder.addLabel(entry.getKey(), entry.getValue());
+    }
 
     if (loggingEnhancers != null) {
       for (LoggingEnhancer enhancer : loggingEnhancers) {
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/LoggingAppenderTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/LoggingAppenderTest.java
@@ -181,6 +181,36 @@ public class LoggingAppenderTest {
     // assertThat(resourceArg.getValue()).isEqualTo(defaultWriteOptions[1]);
   }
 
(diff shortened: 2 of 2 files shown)


sduskis:
@dietervdw-spotify, can you please sign the CLA?

dietervdw-spotify:
@sduskis : Working on it!

dietervdw-spotify:
I signed it! -> Apparently there is a Spotify CLA contributors google group I had to join.

sduskis:
@dietervdw-spotify, it looks like we need some unit tests to pass code coverage tests.

ajaaym:
@dietervdw-spotify why not do this using logging enhancer?

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
```diff
@@ -232,6 +233,10 @@ private LogEntry logEntryFor(ILoggingEvent e) {
         .addLabel(LEVEL_NAME_KEY, level.toString())
         .addLabel(LEVEL_VALUE_KEY, String.valueOf(level.toInt()));
 
+    for (Map.Entry<String, String> entry : e.getMDCPropertyMap().entrySet()) {
+      builder.addLabel(entry.getKey(), entry.getValue());
```
ajaaym: I think you should check that its not overriding ```LEVEL_NAME_KEY``` and ```LEVEL_VALUE_KEY```.

ajaaym:
@dietervdw-spotify making it default might not seems sensible here, bcz different user might want to map to different part of log entry for eg httpRequest or source location or any other stuff. This change is kind of breaking change in that sense as there is no way for user to disable this and may starting seeing their labels at different location. So its better to map this labels using logging enhancer rather than making it default.

dietervdw-spotify:
@ajaaym That's up to your product people to decide I guess.
I personally was a bit surprised I didn't get this 'for free' and it turned me off a little bit.

I know it's not that hard to add a `LoggingEventEnhancer` that does this, but it's one extra hurdle that lessens the usability of the product.
Also the `LoggingEventEnhancer` is mentioned nowhere in the stackdriver logback appender documentation (https://cloud.google.com/logging/docs/setup/java#logback_appender_for) so one could easily get the impression that the `LoggingEnhancer` is the only available extension, and it doesn't support what we want to do.

Suggestion: what if I wrap the above functionality behind a flag (similar to what Logstash does).
Users can then enable/disable inclusion of MDC values by setting a flag on the appender?

ajaaym:
@dietervdw-spotify example of enhancer is already mentioned in that link in configuration section.

@sduskis WDYT about adding this as optional and disabled by default?

dietervdw-spotify:
There is an example of a `LoggerEnhancer` indeed that takes only one parameter, the builder. Looking at that documentation, I would think it was not possible.

Only when looking at the code did I see the `LoggingEventEnhancer` also exists, and that is what you can use to enrich with MDC values.
On top of that you need to be pretty familiar with Logback to know that appender child tag names will get automatically mapped to setters on the appender class (since there is no example of how to configure a loggingEventEnhancer) .

There's a few pretty high hurdles here to climb if you want to use MDC.
I think the average developer will just look at this, find it very limited and move on.
Depends on who you want your target audience to be I guess.

sduskis:
@dietervdw-spotify, this PR is failing.  Can you please run `mvn com.coveo:fmt-maven-plugin:format` to fix the formatting problems?

dietervdw-spotify:
@sduskis Rebased and formatted!

PR #4480, opened 2019-02-11
Removing some deprecation warnings

sduskis:
Most of these changes were based on "inline" refactoring of deprecated methods.


igorbernstein2:
A lot of the files touched are autogenerated, is that ok?

sduskis:
@igorbernstein2, thanks for the catch.  I did a blind refactor.  We'll have to change the generator.

PR #4486, opened 2019-02-12
WIP: Upgrade proto-google-common-protos to 1.14.0.

kolea2:


kolea2:
@andreamlin do you know if this needs to be upgraded in gax as well?

andreamlin:
@kolea2 gax 1.39.0 has this new common protos v1.14.0 now!

andreamlin:
Hey @kolea2 as part of https://github.com/googleapis/google-cloud-java/pull/4496 it was suggested to also bump the proto-google-common-protos there so we could have the same versions of deps across the board at once.

PR #4489, opened 2019-02-13
BigQuery: Fix support standard java date in QueryParameterValue classToType

### LINKED ISSUE #3589: com.google.cloud.bigquery.QueryParameterValue classToType should support standard Java date/time classes.
Please consider adding standard java date, datetime and time classes to the classToType method so that they can be used with "of" generic:
  /** Creates a {@code QueryParameterValue} object with the given value and type. */
  public static <T> QueryParameterValue of(T value, Class<T> type) {
    return of(value, classToType(type));
  }

pmakani:
Fixes #3589 com.google.cloud.bigquery.QueryParameterValue classToType should support standard Java

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/QueryParameterValueTest.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/QueryParameterValueTest.java
@@ -21,14 +21,11 @@ import static com.google.common.truth.Truth.assertThat;
 import com.google.api.services.bigquery.model.QueryParameterType;
 import java.math.BigDecimal;
 import java.text.ParseException;
-import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.List;
 import org.junit.Test;
 
 public class QueryParameterValueTest {
-
-  private static final SimpleDateFormat SIMPLE_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
 
   @Test
   public void testBool() {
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -67,6 +70,7 @@
       DateTimeFormatter.ofPattern("HH:mm:ss.SSSSSS");
   private static final DateTimeFormatter datetimeFormatter =
       DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS");
+  private static final DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd");
```
sduskis: Please use `DateTimeFormatter` instead of the java standard format.

pmakani: DateTimeFormatter is only available after java 8.

sduskis: Please see line 72.

pmakani: Updated to remove formatter and replaced with utility method from google cloud date. Sorry I misread your comment earlier and though of jdk8s datetimeformatter.

PR #4510, opened 2019-02-15
BigQuery: Fix timestamp parsing regression

### LINKED ISSUE #4500: BigQuery timestamp parsing regression (are restrictive timestamp parsing changes intentional?)
#### Environment details

- OS: Linux (Ubuntu Bionic and Alpine 3.8)
- Java version: 11.0.2 and 8.191.12
- google-cloud-java version(s): 1.62.0

#### Overview of issue

In commit fc8438b6a6ad8b1877b086b8fa1481994bf0a3bf the BigQuery date parsing format string was changed from  
`yyyy-MM-dd HH:mm:ss.SSSSSSZZ` to  
`yyyy-MM-dd HH:mm:ss.SSSSSSxxx`

This has broken backwards compatibility in two ways.

- with strings that were using `Z` or `+0000` for the timezone component.
- with strings that were using less than six decimal places of sub-second accuracy

Was this change desirable? I don't understand why you would intentionally make the timestamp parsing as restrictive as this.

In the code sample below, all the example date strings will parse fine in the older BigQuery SDK versions, but fail in the current version.

#### Code sample

(code is in Scala but that isn't relevant to the issue)

```scala
val ourTime = "2019-02-14 12:34:45.938993Z"
val ourTime2 = "2019-02-14 12:34:45.938993+0000"
val ourTime3 = "2019-02-14 12:34:45.102+00:00"

val jodaFormat =
  org.joda.time.format.DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss.SSSSSSZZ")

val threetenFormat =
  org.threeten.bp.format.DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSSxxx")


jodaFormat.parseDateTime(ourTime)
jodaFormat.parseDateTime(ourTime2)
jodaFormat.parseDateTime(ourTime3)
// success!

threetenFormat.parse(ourTime)
// fails!
threetenFormat.parse(ourTime2)
// fails!
threetenFormat.parse(ourTime3)
// fails!
```

pmakani:
Fixes #4500

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
@@ -16,6 +16,11 @@
 
 package com.google.cloud.bigquery;
 
+import static org.threeten.bp.temporal.ChronoField.HOUR_OF_DAY;
+import static org.threeten.bp.temporal.ChronoField.MINUTE_OF_HOUR;
+import static org.threeten.bp.temporal.ChronoField.NANO_OF_SECOND;
+import static org.threeten.bp.temporal.ChronoField.SECOND_OF_MINUTE;
+
 import com.google.api.services.bigquery.model.QueryParameterType;
 import com.google.auto.value.AutoValue;
 import com.google.common.base.Function;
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryParameterValue.java
```diff
@@ -60,6 +62,21 @@
 @AutoValue
 public abstract class QueryParameterValue implements Serializable {
 
+  private static final DateTimeFormatter timestampValidator =
```
sduskis: Can we replace `timestampFormatter` with your configuration for `timestampValidator`?

pmakani: It can't be done with single formatter.

Issue #4485, opened 2019-02-12
Java subscriber throughput limited to 130,000 qps


dpcollins-google:
#### Environment details

- OS: Ubuntu
- Java version: 8
- google-cloud-java version(s): 1.62.0

#### Steps to reproduce

1. Run a publisher publishing with > 130,000 qps of 1kb sized messages
2. Run a subscriber doing minimal work in the callback other than incrementing an atomic counter

#### Code snippet

```java
Integer BYTES_PER_WORKER = 100000000;
Integer workerCount = 5* Runtime.getRuntime().availableProcessors();
Subscriber.newBuilder(subscription, MinimalMessageReceiver)
                            .setParallelPullCount(workerCount)
                            .setFlowControlSettings(FlowControlSettings.newBuilder()
                                    .setMaxOutstandingElementCount(Long.MAX_VALUE)
                                    .setMaxOutstandingRequestBytes(BYTES_PER_WORKER * workerCount)
                                    .build())
                            .build();
```

#### Any additional information below

I've been working to improve the Cloud Pub/Sub loadtest framework.  I've been unable to achieve the same throughput in java as I have in golang under the same conditions.  I'm unable to increase subscriber throughput on a 16 core machine beyond 130MB/s with 1KB messages, but have been seeing 300+MB/s rates with golang under the same conditions.  I've also tried setting the channel provider and executors to different values to no avail.  Since the cpu load is stuck at about 35%, perhaps this is a contention or artificial limitation somewhere else in the client library?

In progress branch for updates here https://github.com/dpcollins-google/pubsub/tree/loadtest_updates

sduskis:
@jadekler, @kolea2 FYI.

Let's discuss this offline.  Please include Kir in the conversation.

jadekler:
:+1:

dpcollins-google:
sure.  dpcollins@.  Put something on my calendar.

sduskis:
@dpcollins-google, can we close this?

dpcollins-google:
Yes, I thought it would be automatically closed by putting resolves in the
last pr description.  You can close or I can when I get home.

PR #4514, opened 2019-02-18
PubSub: Reduce synchronization overhead in MessageDispatcher

dpcollins-google:
Replace the Deque in MessageDispatcher with a LinkedBlockingDeque, and have less code within a synchronized block to reduce contention.

Contributes to fixing #4485 by reducing subscriber contention.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
@@ -381,48 +381,52 @@ class MessageDispatcher {
         } catch (FlowControlException unexpectedException) {
           throw new IllegalStateException("Flow control unexpected exception", unexpectedException);
         }
-
-        final PubsubMessage message = nextMessage.receivedMessage().getMessage();
-        final AckHandler ackHandler = nextMessage.ackHandler();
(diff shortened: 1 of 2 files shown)


sduskis:
I believe that this change would benefit from baby steps.  Can you please do the various changes in different PRs?  The following in separate PRs are one approach:

1. refactoring of `processMessage`
2. the change to outstandingMessageBatches
3. switching `while` to `for`

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -37,13 +37,7 @@
...
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.Executor;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
sduskis: Please change your editor settings to not combine imports from the same package.  i.e. please keep individual imports and don't use `*`

dpcollins-google:
Going to revert this and submit it in pieces, as well as some other changes I've found that make a substantial impact on throughput.

PR #4515, opened 2019-02-19
Pubsub ordering keys subscriber

kamalaboulhosn:
Add support for ordered delivery in subscribers (on ordering-keys branch)


kolea2:
LGTM - will approve after builds pass.

kolea2:
KMS build failure can be ignored; @kamalaboulhosn - we added a new IT presubmit check for KMS, next time you'll need to update the ordering keys branch to pick up the new kokoro config. Thanks!

PR #4516, opened 2019-02-19
Datastore: Fix `more_results` value to query results

### LINKED ISSUE #4504: google-cloud-datastore Add `more_results` value to query results
Please return the `MoreResults` query state value in query results. The end cursor to the next set of results will often return an empty result set, so this value could be used to determine whether or not the next page should be fetched and checked for pagination (and thus confidently return an empty next cursor if the next page is indeed empty as well). The QueryResultImpl only checks if the state value is not "NOT_FINISHED", not the other values as well.

See https://github.com/googleapis/google-cloud-java/issues/3279

pmakani:
Fixes #4504


pmakani:
Done.

mtso:
Thank you!

PR #4526, opened 2019-02-21
[cloud spanner] Add dependency for surefire-junit47

qian-long:
Fixes junit dependency issue when running tests in a VPC.


qian-long:
@silverdev, @snehashah16, @nithinsujir  for FYI

sduskis:
Is this something we should apply for all projects?

qian-long:
I think this was only missing for spanner. For example, bigtable has a clause for this: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/pom.xml#L134

PR #4534, opened 2019-02-21
Upgrade gax dependency to 1.40.0.

kolea2:
Building on @andreamlin's initial PR.

### DIFF
--- a/google-cloud-clients/google-cloud-core/pom.xml
+++ b/google-cloud-clients/google-cloud-core/pom.xml
@@ -37,12 +37,6 @@
         </exclusion>
       </exclusions>
     </dependency>
-
-    <dependency>
-      <groupId>com.google.http-client</groupId>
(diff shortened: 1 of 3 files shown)


kolea2:
@sduskis in response to your comment, this was because of this breaking change in the http client: https://github.com/googleapis/google-http-java-client/releases/tag/v1.28.0

google-cloud-clients/google-cloud-core/pom.xml
```diff
@@ -38,6 +38,11 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client-apache</artifactId>
```
sduskis: Can this be moved to specific projects?  I'd rather reduce dependencies, if possible.

PR #4545, opened 2019-02-22
Change subscriber system executor to scale with numPullers

dpcollins-google:
System executor is the executor that runs the initial callback, this needs to scale with the number of grpc channels open.

Also remove unused variables.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
@@ -162,9 +162,12 @@ public class Subscriber extends AbstractApiService {
 
     ExecutorProvider systemExecutorProvider = builder.systemExecutorProvider;
     if (systemExecutorProvider == null) {
-      systemExecutorProvider = FixedExecutorProvider.create(Executors.newScheduledThreadPool(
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -150,28 +147,40 @@ private Subscriber(Builder builder) {
...
+      channelProvider = channelProvider.withPoolSize(numPullers);
+    }
+
+    ExecutorProvider systemExecutorProvider = builder.systemExecutorProvider;
+    if (systemExecutorProvider == null) {
+      systemExecutorProvider =
+          FixedExecutorProvider.create(
+              Executors.newScheduledThreadPool(
+                  Math.max(
+                      SYSTEM_EXECUTOR_MIN_THREAD_COUNT,
```
saturnism: what is the reason for the default min thread of 6? in container environments, a single process may be assigned few cores shares.

dpcollins-google: This is to maintain parity with the current status quo, that the system executor has 6 cores always, by making it have at least 6 cores.  If you have more insight into the 'correct' minimum number of threads, I would appreciate it, but this makes this a noop in the default case where numPullers is 1 for all systems.

dpcollins-google:

PR #4546, opened 2019-02-22
Change MessageDispatcher to have ownership of the batch queue.

dpcollins-google:
Change MessageDispatcher to have ownership of the batch queue.

Parent is #4545


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -341,10 +336,9 @@ private void startStreamingConnections() throws IOException {
                 ackExpirationPadding,
                 maxAckExtensionPeriod,
                 ackLatencyDistribution,
-                subStub,
+                GrpcSubscriberStub.create(subStubSettings),
```
kolea2: is it okay to lose the IllegalStateException?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -430,7 +421,7 @@ private void stopConnections(List<? extends ApiService> connections) {
         FlowControlSettings.newBuilder().setMaxOutstandingElementCount(1000L).build();
 
     ExecutorProvider executorProvider = DEFAULT_EXECUTOR_PROVIDER;
-    ExecutorProvider systemExecutorProvider = FixedExecutorProvider.create(SHARED_SYSTEM_EXECUTOR);
+    ExecutorProvider systemExecutorProvider = null;
```
kolea2: why are we setting this to null? Is it used anywhere else?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
dpcollins-google: That only existed because the AbstractApiService doStart can't throw a non-RuntimeException.  startStreamingConnections already throws an IOException.

dpcollins-google:

PR #4548, opened 2019-02-22
Decrease message dispatcher synchronization

dpcollins-google:
Decrease the synchronization of the message dispatcher.

Parent is #4546


dpcollins-google:

PR #4549, opened 2019-02-22
Nack when flow controlled

dpcollins-google:
Immediately nack messages when flow controlled. This mirrors the current behavior of the go client libraries and informs cloud pubsub that the messages are unable to be handled.

https://github.com/googleapis/google-cloud-go/blob/709cb6afc0a1b39250cf86d47d84ed44925dbcb9/pubsub/subscription.go#L560

Parent is #4548


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -365,86 +357,77 @@ public void processReceivedMessages(List<ReceivedMessage> messages, Runnable don
...
         }
-        nextBatch.messages.poll(); // We got a hold to the message already.
-        batchDone = nextBatch.messages.isEmpty();
-        if (batchDone) {
-          outstandingMessageBatches.poll();
-          batchCallback = nextBatch.doneCallback;
-        }
+        processOutstandingMessage(nextMessage);
       }
+      nextBatch.doneCallback.run();
```
ajaaym: This is really not good idea. You shouldn't call this if msg is flow controlled. Since you are nacking all msg thats flow controlled it will go in to loop of receiving msg , nacking and requesting more msg, utilizing unnecessary bandwidth and will not allow grpc to auto tune the flow control as well. This should only be called when application has actually completed processing all pending received msgs. previous code of calling this on batchDone with correct.

PR #4559, opened 2019-02-25
Change modack extension logic to not send tens of thousands of rpcs as messages start expiring.

dpcollins-google:
Change modack logic to prevent performance degradation.

Performance decays significantly as messages start expiring, as the current logic sends one modack request per message which is nearly expired. This change makes it instead overextend the deadline on these messages.

This is the same behavior as the go client library https://github.com/googleapis/google-cloud-go/blob/e2c125ceac8b663cfcf4610477d4d67827377cb7/pubsub/iterator.go#L359

Parent is #4548

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
@@ -54,7 +54,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.threeten.bp.Duration;
 import org.threeten.bp.Instant;
-import org.threeten.bp.temporal.ChronoUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


saturnism:
> This change makes it instead overextend the deadline on these messages.

Could I trouble you to clarify what is the new behavior? It sounds like the PR will drop messages that are expired (which makes sense).

dpcollins-google:
> > This change makes it instead overextend the deadline on these messages.
> 
> Could I trouble you to clarify what is the new behavior? It sounds like the PR will drop messages that are expired (which makes sense).

Yes, this pr will (still) drop messages that are expired.  It will not drop them for up to 100ms after they expire however, whereas the current behavior extends them exactly to their deadline.  The reason behind this is every extension of a set of messages to a given time in the future is an rpc call, which can have up to 1000 messages as part of it.  When messages get close to expiring however, they are currently sent alone in an rpc call with only one message and deadline extension to their final deadline, resulting in many outstanding rpcs as the expiry deadline approaches.

dpcollins-google:

PR #4560, opened 2019-02-25
Use a separate executor for each stream's work

dpcollins-google:
Ensure each pull stream has its own executor to reduce contention on the executor queue lock.

Also set channel affinity on acks to ensure responses are sent on the same channel as the StreamingPullMessage came from.

Resolves #4485

Parent is #4559

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java
@@ -298,7 +298,8 @@ final class StreamingSubscriberConnection extends AbstractApiService implements 
                         .setSubscription(subscription)
                         .addAllAckIds(idChunk)
                         .setAckDeadlineSeconds(modack.deadlineExtensionSeconds)
-                        .build());
+                        .build(),
+                    GrpcCallContext.createDefault().withChannelAffinity(channelAffinity));
         ApiFutures.addCallback(future, loggingCallback);
       }
     }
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -468,23 +457,18 @@ void extendDeadlines() {
     List<PendingModifyAckDeadline> modacks = new ArrayList<>();
     PendingModifyAckDeadline modack = new PendingModifyAckDeadline(extendSeconds);
     Instant now = now();
-    Instant extendTo = now.plusSeconds(extendSeconds);
 
     for (Map.Entry<String, AckHandler> entry : pendingMessages.entrySet()) {
       String ackId = entry.getKey();
       Instant totalExpiration = entry.getValue().totalExpiration;
-      if (totalExpiration.isAfter(extendTo)) {
+      if (totalExpiration.isAfter(now)) {
```
kamalaboulhosn: What was the reason for this change? This is changing the semantics of extending ack deadlines for messages that are near the total expiration. With this change, messages could be extended beyond their total expiration.

sduskis:
@dpcollins-google, your PRs have too many changes in any given PR.  Can you please break down these PRs into smaller targeted changes?  I don't feel comfortable merging things as is.

dpcollins-google:
@sduskis I do not believe this is accurate.  Each of these prs builds on the parent (referenced in the comment) forming a chain, but the full diff of the chain shows up in the git diff.  The first pr is #4545, I can resubmit them one at a time if that makes it easier?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
dpcollins-google: This is part of #4559, and only exists in this pr due to the diffbase.  I believe the comment there explains my reasoning behind this change- let me know if you disagree and think I should revert that there.

sduskis:
@dpcollins-google, I hear what you're saying, but I respectfully can't take responsibility to merge these PRs as is.

dpcollins-google:

PR #4581, opened 2019-02-27
Prepare for KMS GA release - upgrade versions to 1.0.0.

kolea2:
When we next release, the version increments for KMS will need to be reverted so we can start at 1.0.0.

### DIFF
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -467,7 +467,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-kms</artifactId>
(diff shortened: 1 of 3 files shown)


sduskis:
I'm not completely sure how the `versions.txt` gets processed, but I don't think that the current proposal will work.  Maybe you'd need `1.0.0-SNAPSHOT`?

kolea2:
@sduskis see here for an example of how versions get updated: https://github.com/googleapis/google-cloud-java/pull/4536/files#diff-9fa081a8634fe9e631cd85762bb9ceb9R4. We'll need to manually keep it at 1.0.0 when we release this first time (no snapshot). This is more a placeholder commit to at least change the files to be 1.x.x with no 'beta' version marking.

PR #4590, opened 2019-02-28
Change MessageDispatcher to own its own queue.

dpcollins-google:


sduskis:
@dpcollins-google: Do you know why there was a shared list to begin with?  What behavior is changed as part of this PR?

dpcollins-google:
> @dpcollins-google: Do you know why there was a shared list to begin with? What behavior is changed as part of this PR?

I don't have the background on why there was a shared list.  Under default configurations, there is no change as there is only StreamingSubscriberConnection.

As you increase this number, having a shared list could hypothetically decrease the latency to handle a given batch, but it will also massively increase contention as every ack action and every new batch received by every connection has to synchronize on that one object.

dpcollins-google:
I should clarify that, because pulls occur on every new batch or ack/nack/expiration, the realistic latency is unlikely to be high anyway.

sduskis:
@kamalaboulhosn or @kimkyung-goog, can you please confirm that there isn't a conflict between this and the ordered key issues?

kamalaboulhosn:
@sduskis There will be some things to work out with ordering keys, but @dpcollins-google and @kimkyung-goog will work to ensure that everything is resolved on the ordering keys branch before we integrate any of those changes into master.

PR #4594, opened 2019-02-28
Using the google-cloud-http-bom

### LINKED ISSUE #4580: Depend on google-http-java-client BOM
Instead of versions directly.

sduskis:
Fixes #4580


sduskis:
@elharo, FYI

PR #4596, opened 2019-02-28
Have each StreamingSubscriberConnection use its own stub.

dpcollins-google:
This has the effect of, if the user passes an instantiatingExecutorProvider as a systemExecutorProvider, instantiating a new executor for every connection.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
@@ -308,6 +308,8 @@ public class Subscriber extends AbstractApiService {
   private void startStreamingConnections() throws IOException {
     synchronized (streamingSubscriberConnections) {
       for (int i = 0; i < numPullers; i++) {
+        // Each connection gets its own stub so it gets its own system executor and channel
+        // instantiation.
         streamingSubscriberConnections.add(
             new StreamingSubscriberConnection(
                 subscriptionName,


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java
```diff
@@ -329,7 +320,7 @@ private void startStreamingConnections() {
                 ACK_EXPIRATION_PADDING,
                 maxAckExtensionPeriod,
                 ackLatencyDistribution,
-                subStub,
+                GrpcSubscriberStub.create(subStubSettings),
```
sduskis: Please add a comment to this line indicating why each connection needs its own stub, so that this isn't removed in the future.

ajaaym: If new stub is created for each puller, whats point of keeping channel pooling and affinity?

sduskis: @ajaaym, does pub/sub have a channel pool with affinity, or is that only Cloud Spanner?

ajaaym: @sduskis here is the [pooling](https://github.com/googleapis/google-cloud-java/blob/9a8ef2ec51c425e6451a3a95f6b3df9650bbc418/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java#L164) and [afinity](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java#L196) for subcriber. If using new client for each puller then I dont think we need this. Also whats reason for using new client vs channel pooling if each client uses same settings?

sduskis: @ajaaym, do you see a need for affinity here?  I know why we need it for Cloud Spanner, but I don't understand why we'd need it here.

dpcollins-google: The purpose here of using a new stub was to respect the user's passed settings.  If the user passes an instantiatingExecutorProvider for the SystemExecutorProvider, they probably expect us to instantiate a new executor for each grpc channel, not instantiate one and share.

You're correct that manually overriding the pooling settings of the channel provider doesn't really make sense, for the same reason: If the user passes an InstantiatingTransportChannelProvider, they probably expect the transport channel to be instantiated once for every puller.  I planned to remove the lines below (although I've now removed them here):

```
if (channelProvider.acceptsPoolSize()) {
  channelProvider = channelProvider.withPoolSize(numPullers);
}
```

Which non-optionally override the users settings.  This will mean that yes, we no longer need affinity with the default options.  However, because the user can still pass a FixedTransportChannelProvider that has pooling enabled, we still need to handle this case.  I also plan on handling this better by also setting affinity on the ack/nack/extension callables in StreamingSubscriberConnection, although this will no longer make a difference in the default case.

It should be noted that the other place we do not respect the user's indicated preferences, the regular executorProvider, Is another change that I intend to make; such that if they pass an instantiatingExecutorProvider, it will be instantiated once for each connection.

sduskis: I'm not sure how setting an explicit affinity to `i` is going to help in any case.  If it's a single channel, then affinity is meaningless.  If the user passes in a `
FixedTransportChannelProvide`, will affinity of `i` be meaningful?

dpcollins-google:
This seems somewhat contentious, but is also mostly a qol fix not a performance one.  I'm going to shelve this and I'll reopen and come back to it if after making the other performance changes that should be more straightforward I'm still not seeing the increased performance I am with the fully patched version.

PR #4597, opened 2019-02-28
Using the latest version of truth

sduskis:
In order to use a newer version of truth, I had to update other dependencies as well, and change some spanner tests.


sduskis:
The storage integration test failed: 

> [16:03:21][ERROR] Cannot check out non-mergeable pull request as if merged

That's weird, but unrelated to this PR.

PR #4599, opened 2019-02-28
Adding a pubsub publisher message transformer

sduskis:
This will allow generic transformations of messages before a message is published.

This should help towards #4240

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.pubsub.v1;
 
+import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
@@ -95,6 +96,7 @@ public class Publisher {
   private final List<AutoCloseable> closeables;
   private final MessageWaiter messagesWaiter;
   private ScheduledFuture<?> currentAlarmFuture;
+  private final ApiFunction<PubsubMessage, PubsubMessage> messageTransform;
 
   /** The maximum number of messages in one request. Defined by the API. */
   public static long getApiMaxRequestElementCount() {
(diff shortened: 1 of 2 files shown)


dpcollins-google:
To me, this looks like it would likely be abused by users to implement a pipeline construct, which doesn't seem to be the intent.  This would mean the order of application would have to be guaranteed and never changed to prevent breaking user code.  What would you think about instead of having this be intrinsic, making Publisher an Abstract Class/Interface, and having the user wrap their publisher, similar to guava's ForwardingQueue?  Something like

```java
class CensusTrackedPublisher extends ForwardingPublisher {
  // Could be implicit, here for clarity.
  CensusTrackedPublisher(AbstractPublisher delegate) {
     super(delegate);
  }

  @Override
  public ApiFuture<String> publish(PubsubMessage message) {
    return delegate().publish(censusTransform(message));
  }
}
```

sduskis:
@dpcollins-google, it would seem to me that your suggestion would make the client backwards incompatible, which I don't think we can do.

dpcollins-google:
@sduskis in what way?  Publisher only has a private constructor and can only be instantiated through the Builder.  If Publisher were instead a public abstract class with one abstract method (Publish) and Builder.build() instead returned a {package-private} `class PublisherImpl extends Publisher`, I don't think any user code would have to be changed.  There would be 0 api changes relative to the current state.  Other languages (python, javascript, go) are dynamic enough that this would already be possible without the explicit interface, and you could maintain parity with no changes to other languages.

sduskis:
@dpcollins-google, there are public static methods on `Publisher` which we would not be able to support in Java 7.

dhoard:
@sduskis This type of functionality seems like it should be above the Pub/Sub client since it's specific to the application.

```
class CensusMessagePublisher {

  private Publisher publisher;

  public CensusMessagePublisher(Publisher publisher) {
     this.publisher = publisher;  
  }

  @Override
  public ApiFuture<String> publish(CensusMessage message) {
    return this.publisher.publish(transform(message));
  }

  private PubsubMessage transform(CensusMessage message) {
      // transform the CensusMessage to a PubsubMessage
      return transformedMessage;
   }
}
```

sduskis:
@dhoard, thanks for the guidance.  FWIW, this change was in context of #4240 which is OpenCensus / PubSub.  A review of #4240 would also be greatly appreciated.

PR #4616, opened 2019-03-01
Set the isDirectory flag appropriately on the test storage rpc

clarkperkins:
Set the isDirectory flag appropriately on the test storage rpc when using the currentDirectory() option


clarkperkins:
I signed it!

PR #4617, opened 2019-03-03
Change MessageDispatcher to not synchronize on the global queue.

dpcollins-google:
This is no longer necessary, as it is owned by this class.  It also prevents every ack, nack and batch receipt from happening only one at a time.


sduskis:
Can you please break this up into straightforward refactoring(s) and then the crux of the change?

PR #4622, opened 2019-03-04
Change each StreamingSubscriberConnection to have its own executor by default.

dpcollins-google:
This increases throughput by reducing contention on the executor queue mutex and makes the Subscriber implementation more accurately reflect the users intent when an InstantiatingExecutorProvider is passed.

Fixes the majority of the issues with #4485, although there may be other changes if they show to increase performance.


sduskis:
@dpcollins-google, can you add some comments on `executor` and `alarmsExecutor`?

dpcollins-google:
@sduskis Sorry about this, I didn't realize this was still blocked.  I've added a comment to both of these.

PR #4656, opened 2019-03-07
Fix Timestamp.parseTimestamp.

### LINKED ISSUE #4583: Documentation is incorrect/confusing for Timestamp.parseTimestamp
The code currently in master is:

```java
/**
   * Creates a Timestamp instance from the given string. String is in the RFC 3339 format without
   * the timezone offset (always ends in "Z").
   */
  public static Timestamp parseTimestamp(String timestamp) {
    Instant instant = Instant.parse(timestamp);
    return ofTimeSecondsAndNanos(instant.getEpochSecond(), instant.getNano());
  }
```

But if we look at the code for `org.threeten.bp.Instant.parse`:

```java
/**
     * Obtains an instance of {@code Instant} from a text string such as
     * {@code 2007-12-03T10:15:30.000Z}.
     * <p>
     * The string must represent a valid instant in UTC and is parsed using
     * {@link DateTimeFormatter#ISO_INSTANT}.
     *
     * @param text  the text to parse, not null
     * @return the parsed instant, not null
     * @throws DateTimeParseException if the text cannot be parsed
     */
    public static Instant parse(final CharSequence text) {
        return DateTimeFormatter.ISO_INSTANT.parse(text, Instant.FROM);
    }
```

This implies to me that per the Google Cloud `Timestamp` documentation the following should work:

`Timestamp.parseTimestamp("2019-01-23T12:30:00")`

But it generates:

```
org.threeten.bp.format.DateTimeParseException: Text '2019-01-23T12:30:00' could not be parsed at index 19
```

It also would imply to me that the following _shouldn't_ work:

`Timestamp.parseTimestamp("2019-01-23T12:30:00Z")`

pmakani:
Fixes #4583

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
+++ b/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
@@ -26,7 +26,6 @@ import java.util.concurrent.TimeUnit;
 import org.threeten.bp.Instant;
 import org.threeten.bp.LocalDateTime;
 import org.threeten.bp.ZoneOffset;
-import org.threeten.bp.ZonedDateTime;
 import org.threeten.bp.format.DateTimeFormatter;
 import org.threeten.bp.format.DateTimeFormatterBuilder;
 import org.threeten.bp.temporal.TemporalAccessor;
@@ -50,13 +49,14 @@ public final class Timestamp implements Comparable<Timestamp>, Serializable {
 
   private static final DateTimeFormatter format = DateTimeFormatter.ISO_LOCAL_DATE_TIME;
 
-  private static final DateTimeFormatter timestampFormatter =
+  private static final DateTimeFormatter timestampParser =
       new DateTimeFormatterBuilder()
           .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
           .optionalStart()
           .appendOffsetId()
           .optionalEnd()
-          .toFormatter();
+          .toFormatter()
+          .withZone(ZoneOffset.UTC);
 
   private final long seconds;
   private final int nanos;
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -47,6 +50,14 @@
 
   private static final DateTimeFormatter format = DateTimeFormatter.ISO_LOCAL_DATE_TIME;
 
+  private static final DateTimeFormatter timestampFormatter =
```
sduskis: Please call this `timestampParser`

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -170,7 +181,10 @@ public static Timestamp fromProto(com.google.protobuf.Timestamp proto) {
    * the timezone offset (always ends in "Z").
    */
   public static Timestamp parseTimestamp(String timestamp) {
-    Instant instant = Instant.parse(timestamp);
+    TemporalAccessor temporalAccessor = timestampFormatter.parse(timestamp);
+    LocalDateTime localDateTime = LocalDateTime.from(temporalAccessor);
+    ZonedDateTime zonedDateTime = ZonedDateTime.of(localDateTime, ZoneOffset.UTC);
```
sduskis: Should our parser have settings to automaticcaly be in `ZoneOffset.UTC`?

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
pmakani: Done.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
pmakani: Done.

PR #4660, opened 2019-03-08
add public interface of paritionToken for google internal systest

gyang-google:
This is required for cloud spanner systest framework to test batch transaction part.


sduskis:
@gyang-google, this PR is failing.  Please reopen a new PR, if this is still necessary.  There is no annotation like the one you added in this repo.

PR #4663, opened 2019-03-08
Spanner: Throw exception when SSLHandshakeException occurs instead of infinite retry loop

### LINKED ISSUE #3889: Google-cloud-spanner:Error in connection java.util.concurrent.ExecutionException: io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
I am trying to run my application which connects to google cloud spanner in a yarn container and the application gets blocked. Upon further investigation I found that the following error was being thrown 

`java.util.concurrent.ExecutionException: io.grpc.StatusRuntimeException: UNAVAILABLE: io exception`

 Environment details

- OS: linux
- Java version: 1.8
- google-cloud-java version(s): google-spanner-0.52-beta

The code I am trying to execute:

```java
  Spanner spannerClient=SpannerOptions.newBuilder().build().getService();
InstanceAdminClient instanceAdminClient = spannerClient.getInstanceAdminClient();
            Page<Database> databasePage = instanceAdminClient.getInstance(instanceID).listDatabases();
            for (Database database : databasePage.getValues()) {
                String databaseName = database.getId().getDatabase();
                log.info("Database-->"+databaseName);
            }
```
This code gets blocked while running in yarn container. Upon inspecting I got the following thread dump:

```
sun.misc.Unsafe.park(boolean, long) @bci=0 (Compiled frame; information may be imprecise)
 - java.util.concurrent.locks.LockSupport.parkNanos(java.lang.Object, long) @bci=20, line=215 (Compiled frame)
 - java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(int, long) @bci=139, line=1037 (Interpreted frame)
 - java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(int, long) @bci=25, line=1328 (Interpreted frame)
 - java.util.concurrent.CountDownLatch.await(long, java.util.concurrent.TimeUnit) @bci=10, line=277 (Interpreted frame)
 - com.google.cloud.spanner.SpannerImpl.backoffSleep(io.grpc.Context, long) @bci=67, line=200 (Interpreted frame)
 - com.google.cloud.spanner.SpannerImpl.backoffSleep(io.grpc.Context, com.shaded.scanner.google.api.client.util.BackOff) @bci=5, line=170 (Interpreted frame)
 - com.google.cloud.spanner.SpannerImpl.runWithRetries(java.util.concurrent.Callable) @bci=105, line=238 (Interpreted frame)
 - com.google.cloud.spanner.SpannerImpl$InstanceAdminClientImpl.getInstance(java.lang.String) @bci=25, line=660 (Interpreted frame)
 - com.test.spannerscanner.utils.GoogleSpannerManager.execute() @bci=289, line=93 (Interpreted frame)
```

Upon further debugging I found the runWithRetries method in SpannerImpl is throwing SpannerException due to which the code is getting blocked. I examined the SpannerException and in the detailed message I found 

`java.util.concurrent.ExecutionException: io.grpc.StatusRuntimeException: UNAVAILABLE: io exception`

The code works fine in my local environment but in yarn container it gets blocked due to the SpannerException. Any guidance in how to resolve the issue is greatly appreciated.

olavloite:
Fixes #3889

SSLHandshakeExceptions are not retryable, as it is most probably an indication that the client does not accept the server certificate. The Spanner client now detects this exception and throws it instead of repeatedly retrying without success.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java
@@ -16,10 +16,6 @@
 
 package com.google.cloud.spanner;
 
-import java.util.concurrent.CancellationException;
-import java.util.concurrent.TimeoutException;
-import javax.annotation.Nullable;
-import javax.net.ssl.SSLHandshakeException;
 import com.google.api.gax.grpc.GrpcStatusCode;
 import com.google.api.gax.rpc.ApiException;
 import com.google.cloud.spanner.SpannerException.DoNotConstructDirectly;
@@ -28,6 +24,10 @@ import com.google.common.base.Predicate;
 import io.grpc.Context;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.TimeoutException;
+import javax.annotation.Nullable;
+import javax.net.ssl.SSLHandshakeException;
 
 /**
  * A factory for creating instances of {@link SpannerException} and its subtypes. All creation of
@@ -213,11 +213,12 @@ public final class SpannerExceptionFactory {
             return false;
           }
         };
-    static final Predicate<Throwable> isSSLHandshakeException = new Predicate<Throwable>() {
-      @Override
-      public boolean apply(Throwable input) {
-        return input instanceof SSLHandshakeException;
-      }
-    };
+    static final Predicate<Throwable> isSSLHandshakeException =
+        new Predicate<Throwable>() {
+          @Override
+          public boolean apply(Throwable input) {
+            return input instanceof SSLHandshakeException;
+          }
+        };
   }
 }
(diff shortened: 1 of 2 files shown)


ajaaym:
@olavloite I think this will not fix the problem as the exception raised [here](https://github.com/grpc/grpc-java/blob/master/alts/src/main/java/io/grpc/alts/GoogleDefaultChannelBuilder.java#L120) doesn't propagated properly with cause. Did you get chance to test with actual missing certificate or invalid server side certificate?

ajaaym:
There is also test related to this [here](https://github.com/grpc/grpc-java/blob/master/netty/src/test/java/io/grpc/netty/TlsTest.java#L263) which talks about this problem

olavloite:
@ajaaym Yes, I have a local emulator on my machine running with a self-signed certificate. When I open a Spanner client without this fix with a JVM that does not have this self-signed certificate in the trust store, I get the exact same behavior as @ash1292, i.e. an endless retry loop, as the SSLHandshakeException is translated into a StatusRuntimeException with code UNAVAILABLE. With this fix, the specific case where an SSLHandshakeException is in the cause is recognized and the exception is re-thrown.
Running both versions (with and without fix) with a JVM that has the self-signed certificate in the trust store also works.

ajaaym:
@olavloite I think when its uses TCNATIVE implementation then it throws SSLHandshakeException properly. Most of the case would be using TCNATIVE so it would be fine.

kolea2:
Hi @olavloite! Can you please rename this PR to something like "Spanner: [description of change]"? It helps us with release notes later on. Thanks!

kolea2:
@snehashah16 FYI

PR #4674, opened 2019-03-13
Added framework for simulating aborted transactions in test code

### LINKED ISSUE #4673: Enable injecting transaction aborts in test code
One of the challenges developers encounter when using Cloud Spanner is that Cloud Spanner aborts a relatively high percentage of all read/write transactions. Transactions must therefore be included in retry-loops or other safe-guarding techniques. As a transaction can be aborted at any time, it is hard to write tests that demonstrate the working of the safe-guarding techniques in use. It would be nice to be able to inject a transaction abort in a specific place in test code, so that I can write reliable tests for my retry safeguarding.

olavloite:
Fixes #4673 

One of the challenges developers encounter when using Cloud Spanner is that Cloud Spanner aborts a relatively high percentage of all read/write transactions. Transactions must therefore be included in retry-loops or other safe-guarding techniques. As a transaction can be aborted at any time, it is hard to write tests that demonstrate the working of the safe-guarding techniques in use. This change enables developers to inject transaction aborts in test code.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
@@ -157,7 +157,6 @@ class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
   private final Random random = new Random();
   private final SpannerRpc gapicRpc;
   private final int defaultPrefetchChunks;
-  private final AbortedTransactionInjector abortedInjector;
 
   @GuardedBy("this")
   private final Map<DatabaseId, DatabaseClientImpl> dbClients = new HashMap<>();
@@ -171,7 +170,6 @@ class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
   SpannerImpl(SpannerRpc gapicRpc, int defaultPrefetchChunks, SpannerOptions options) {
     super(options);
     this.gapicRpc = gapicRpc;
-    this.abortedInjector = options.getAbortedInjector();
     this.defaultPrefetchChunks = defaultPrefetchChunks;
     this.dbAdminClient = new DatabaseAdminClientImpl(options.getProjectId(), gapicRpc);
     this.instanceClient =
@@ -186,8 +184,7 @@ class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
     return new ExponentialBackOff.Builder()
         .setInitialIntervalMillis(MIN_BACKOFF_MS)
         .setMaxIntervalMillis(MAX_BACKOFF_MS)
-        // Prevent Backoff.STOP from getting returned.
-        .setMaxElapsedTimeMillis(Integer.MAX_VALUE)
+        .setMaxElapsedTimeMillis(Integer.MAX_VALUE) // Prevent Backoff.STOP from getting returned.
         .build();
   }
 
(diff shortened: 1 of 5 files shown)


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbortedTransactionInjector.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.google.cloud.spanner;
+
+/**
+ * Interface for simulating an aborted read/write transaction in test code. You should only use this
+ * class in test cases to verify that your code handles aborted exceptions correctly.
+ *
+ * <p>DO NOT USE IN PRODUCTION CODE.
+ */
+public interface AbortedTransactionInjector {
```
sduskis: This probably should be package private.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -928,8 +931,14 @@ public Transaction call() throws Exception {
     }
 
     TransactionContextImpl newTransaction() {
-      TransactionContextImpl txn =
-          new TransactionContextImpl(this, readyTransactionId, gapicRpc, defaultPrefetchChunks);
+      TransactionContextImpl txn;
```
sduskis: Can this be accomplished with a subclass used in a test rather than changing `SpannerImpl`?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -1056,7 +1065,7 @@ private ResultSet executeQueryInternal(
         QueryOption... options) {
       Options readOptions = Options.fromQueryOptions(options);
       return executeQueryInternalWithOptions(
-          statement, queryMode, readOptions, null /*partitionToken*/);
+          statement, queryMode, readOptions, null /* partitionToken */);
```
sduskis: Please raise a separate PR for formatting changes.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbortedTransactionInjector.java
olavloite: No, for two reasons:
1. It is referenced in integration tests, which are in a different package.
2. The intention is that other (framework) developers should be able to use this to test any custom aborted transaction handling.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
olavloite: Good idea, I changed it to an implementation using a subclass. I still needed a couple of small additions to `SpannerImpl`, but those are limited to a couple of package-private getter methods.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
olavloite: Removed all formatting changes from the PR.

sduskis:
I spoke with @kolea2 about this PR.  We don't want any test code in the public interface.  Is there a way to test the same thing in another way?  We can chat offline about different strategies.

olavloite:
This should be done through a proxy/mock server instead of code in the main client library.

PR #4680, opened 2019-03-14
Fix ordering problem and reduce synchronization.

krka:
Fix:
pendingWrites.add() is not guaranteed to be called before pendingWrites.remove()

Optimization:
Remove synchronization on pendingWrites by using a concurrent data structure instead.
Since there's no concurrent identity based hash map, we can use a regular
ConcurrentHashMap together with a key that implements equals and hashcode based on identity
(a simple Object)


sduskis:
@krka, please sign the CLA/

krka:
Oops, I used my personal account - will resubmit later with my employee account and fix the CLA issue

sduskis:
@krka, I'm going to close this PR.  Please reopen the PR from your other account.

PR #4683, opened 2019-03-15
#3616 use retry settings when executing runWithRetries(Callable)

### LINKED ISSUE #3616: How to specify Spanner connect timeout?

olavloite:
The method SpannerImpl#runWithRetries(Callable) did not consider the retry settings that had been set on the SpannerOptions that created it. This could cause certain calls to wait indefinitely without ever
breaking off with an exception.
This fix changes the SpannerImpl#runWithRetries(Callable) to consider the retry settings set on SpannerOptions.

I also considered changing the implementation into using RetryHelper#runWithRetries, but that API is (according to itself) still beta.

Fixes #3616

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
@@ -114,6 +114,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
+import org.threeten.bp.Duration;
 
 /** Default implementation of the Cloud Spanner interface. */
 class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
@@ -171,9 +172,9 @@ class SpannerImpl extends BaseService<SpannerOptions> implements Spanner {
     super(options);
     this.gapicRpc = gapicRpc;
     this.defaultPrefetchChunks = defaultPrefetchChunks;
-    this.dbAdminClient = new DatabaseAdminClientImpl(options.getProjectId(), gapicRpc);
+    this.dbAdminClient = new DatabaseAdminClientImpl(options.getProjectId(), gapicRpc, options);
     this.instanceClient =
-        new InstanceAdminClientImpl(options.getProjectId(), gapicRpc, dbAdminClient);
+        new InstanceAdminClientImpl(options.getProjectId(), gapicRpc, dbAdminClient, options);
   }
 
   SpannerImpl(SpannerOptions options) {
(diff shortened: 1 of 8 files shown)


kolea2:
CC @snehashah16

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -263,6 +265,19 @@ public void cancelled(Context context) {
         } else {
           backoffSleep(context, backOff);
         }
+        if (options.getRetrySettings() != null) {
```
snehashah16: i would suspect the gax library being able to take care of these.

We should sync with gax|gapic team to help here and make this generic across all APIs and client libraries.

@hzyi-google - any thoughts on how to implement this?

hzyi-google: It does not seem this implementation is using any retry helpers provided by gax. what it's doing is extracting retry configurations set in `SpannerOptions` and does manual retries here.

olavloite: The Spanner library has its own implementation of runWithRetries(...), which also uses io.grpc.Context to allow the user to cancel calls or set timeouts. Re-writing this method, and thereby all the Spanner library, to use the standard helper methods from gax is probably a good idea. But as far as I can see, it will probably also have some side-effects to the current behavior.

Is that something we would want as part of solving this issue, or should that be a separate project?

hzyi-google: I agree with you that it will be ideal to use retry helpers in gax, and I also think it should be a separate project as well.

sduskis: @hzyi-google, should we close this PR and work on using the GAX retry helpers in another PR?

hzyi-google:
Sorry for the delay. This PR LGTM. But I think @snehashah16 should approve as well.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
snehashah16: @sduskis - we can have knut (or whoever) write up a design on migrating to gax, and its side-effects (or gaps in gax library) that Cloud Spanner may have a change in behavior.

I would start a new project and close this PR.
Also, there is a parallel thread for setting timeouts (and rpc cancellations) that should be included in this project (if possible) to make sure we dont drop any existing feature from the clients.

@crwilcox for timeouts & cancellations issue.

snehashah16:
@olavloite 
"I also considered changing the implementation into using RetryHelper#runWithRetries, but that API is (according to itself) still beta."

Can you provide the link/reference to this library RetryHelper#runWithRetries

olavloite:
@snehashah16 
> Can you provide the link/reference to this library RetryHelper#runWithRetries

com.google.cloud.RetryHelper from google-cloud-core: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/RetryHelper.java

olavloite:
Replaced by #4940

PR #4684, opened 2019-03-15
Partition token

gyang-google:
add public interface of partition Token for google internal systest.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Partition.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Partition.java
@@ -96,7 +96,6 @@ public class Partition implements Serializable {
     return partitionToken;
   }
 
-  @GoogleInternal
   public ByteString getPartitionTokenForTest() {
     return partitionToken;
   }


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Partition.java
```diff
@@ -96,6 +96,10 @@ ByteString getPartitionToken() {
     return partitionToken;
   }
 
+  public ByteString getPartitionTokenForTest() {
```
sduskis: can this be package private?

gyang-google: We intend to make it public to get them out of this class for test. Actually that's what we want to do.

sduskis: We should be super careful to not have test classes / methods in the public GA API.  You can create a test utility in the `com.google.cloud.spanner` project, so that we don't have a public test method in the class we expose.

gyang-google: What do you mean by creating a test utility? Basically we want something that can get the partitionToken variable out of the partition class.

sduskis: I mean creating a `com/google/cloud/spanner/PartitionTestUtil.java` class that would be able to access the package-private variables, and implementing `PartitionTestUtil.extractPartitionToken(partition)` which will call `Partition.getPartitionToken()`

gyang-google: Sure, will do that.

PR #4692, opened 2019-03-18
(storage) Add V4 signing support

JesseLovelace:
Adds support for V4 signing. Continues to use V2 signing as default.


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/CanonicalExtensionHeadersSerializer.java
```diff
@@ -81,4 +67,55 @@ public StringBuilder serialize(Map<String, String> canonicalizedExtensionHeaders
...
+    // Make all custom header names lowercase.
+    Map<String, String> lowercaseHeaders = new HashMap<>();
+    for (String headerName : new ArrayList<>(canonicalizedExtensionHeaders.keySet())) {
+
+      String lowercaseHeaderName = headerName.toLowerCase();
+
+      // If present, remove the x-goog-encryption-key and x-goog-encryption-key-sha256 headers.
+      if ("x-goog-encryption-key".equals(lowercaseHeaderName)
+          || "x-goog-encryption-key-sha256".equals(lowercaseHeaderName)
+          || (isV4 && "x-goog-encryption-algorithm".equals(lowercaseHeaderName))) {
```
frankyn: At the moment `x-goog-encryption-key` and `x-goog-encryption-key-sha256` should be removed for both v2 and v4. There's an open question the GCS team if this is the case moving forward. (no-op for now).

How does knowing v4 help here?

JesseLovelace: "x-goog-encryption-key" and "x-goog-encryption-key-sha256" are removed for both v2 and v4 here. "x-goog-encyrption-algorithm" is only removed for v4 here (C# does this as well) which is why there's a check

frankyn: Linking you to an internal bug: 128647687. This should be clarified when we get a response there.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -162,6 +162,7 @@
   public static void beforeClass() throws IOException {
     remoteStorageHelper = RemoteStorageHelper.create();
     storage = remoteStorageHelper.getOptions().getService();
+
```
frankyn: I'm assuming you're adding integration tests (I think you mentioned it), but I want to make sure they are still added before this is merged.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/CanonicalExtensionHeadersSerializerTest.java
```diff
@@ -70,8 +74,24 @@ public void givenHeadersWhenSerializeThenSuccess() {
...
+    StringBuilder sb = v2Serializer.serialize(encryptionHeaders);
 
     assertEquals(sb.toString(), "x-goog-acl:public-read\nx-goog-meta-owner:myself and others\n");
   }
+
+  @Test
+  public void testV4Serialization() {
+    Map<String, String> encryptionHeaders = new HashMap<>();
+    encryptionHeaders.put("x-goog-encryption-key", "key");
+    encryptionHeaders.put("x-GOOg-acl", " \n public-read ");
```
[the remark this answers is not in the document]
JesseLovelace: yes, it's just an additional check to make sure it gets converted to lowercase

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/CanonicalExtensionHeadersSerializer.java
```diff
@@ -32,31 +32,29 @@
...
+  private static final char HEADER_NAME_SEPARATOR = ';';
+
+  private final Storage.SignUrlOption.SignatureVersion signatureVersion;
+
+  public CanonicalExtensionHeadersSerializer(
+      Storage.SignUrlOption.SignatureVersion signatureVersion) {
+    this.signatureVersion = signatureVersion;
+  }
+
+  public CanonicalExtensionHeadersSerializer() {
```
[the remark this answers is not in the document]
JesseLovelace: It's not, I just didn't want to invalidate the constructor

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/V4SigningTest.java
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright 2015 Google LLC
```
frankyn: 2019

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -609,6 +612,11 @@ public URL signUrl(BlobInfo blobInfo, long duration, TimeUnit unit, SignUrlOptio
     for (SignUrlOption option : options) {
       optionMap.put(option.getOption(), option.getValue());
     }
+
```
frankyn: Could you add an example of using v4 in the javadoc (https://github.com/googleapis/google-cloud-java/blob/836d7d5860578ab5ac2c05a26e4386707dda7eb6/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L2121)?

frankyn:
@JesseLovelace LGTM, please don't merge until one more change to conformance tests and languages are aligned for release.

frankyn:
@JesseLovelace please update the conformance tests once more. The trailing slash for list objects signed URLs was removed. Then it's waiting for the rest of the languages to be ready.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -619,42 +628,63 @@ public URL signUrl(BlobInfo blobInfo, long duration, TimeUnit unit, SignUrlOptio
...
+        stBuilder.append("&X-Goog-Signature=").append(signature);
+      } else {
+        BaseEncoding encoding = BaseEncoding.base64();
+        String signature = URLEncoder.encode(encoding.encode(signatureBytes), UTF_8.name());
+        stBuilder.append("?");
+        stBuilder.append("GoogleAccessId=").append(credentials.getAccount());
+        stBuilder.append("&Expires=").append(expiration);
+        stBuilder.append("&Signature=").append(signature);
+
+        log.warning(
```
frankyn: One last comment: Jesse is it possible to only log this when they use signedUrl? Possibly using a third enum to determine if it's using v2 by default or if the user wants to use v2.

PR #4699, opened 2019-03-19
Add Cloud Talent Solution API.

kolea2:
Generated with new_client.py.

Please double check pom.xml's and versions.txt. Any miscellaneous changes to versions.txt were a re-alphabetizing as a result of running new_client.py.


google-api-grpc/pom.xml
```diff
@@ -792,7 +803,7 @@
     <module>proto-google-cloud-speech-v1</module>
     <module>proto-google-cloud-speech-v1beta1</module>
     <module>proto-google-cloud-speech-v1p1beta1</module>
-    <module>proto-google-cloud-tasks-v2beta2</module>
+    <module>proto-google-cloud-talent-v4beta1</module><module>proto-google-cloud-tasks-v2beta2</module>
```
sduskis: Please add a new line between modules

PR #4709, opened 2019-03-20
Added feature to accept Firestore Value.

### LINKED ISSUE #3315: Firestore should accept Protobuf and Firestore Value(s)
Currently, the Firestore SDK provides the following interfaces for `set(`/`update(`/etc, w.r.t. write batches and writing in general (parameters irrelevant to this ticket omitted):

- Add via Java: `set(..., Map<String, Object> fields, ...)`
- Add via POJO: `set(..., Object pojo, ...)`

These are great for a lot of use cases. But, those underlying methods immediately serialize everything to `com.google.firestore.v1beta1.Value` objects.

If your application would like more control over the `Value`(s) selected during serialization, why can't you provide your object via `set(..., Map<String, com.google.firestore.v1beta1.Value>)`?

I'd like to request that such an interface be added. Or, another way to do it would be to accept Protobuf's regular `Value` well-known-type. However, it seems as though it reflects a subset of Firestore's types (i.e. `setNumberValue` where Firestore has `setDoubleValue`, `setIntegerValue`, etc) so I understand why that particular type wrapper was not used.

Thank you in advance

pmakani:
Fixes #3315

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
+++ b/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
@@ -150,6 +150,14 @@ public class ITSystemTest {
   }
 
   @Test
+  public void createDocumentWithValue() throws Exception {
+    assertEquals(20, randomDoc.getId().length());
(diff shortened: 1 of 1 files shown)


sduskis:
Would an integration test be possible?

pmakani:
> Would an integration test be possible?

Done.

PR #4721, opened 2019-03-22
Spanner: Added extra documentation to the DatabaseClient.singleUse methods

### LINKED ISSUE #4212: Spanner: Better document the difference between singleUse() and singleUseReadOnlyTransaction().
Hi,
I think the documentation of the difference between `DatabaseClient.singleUse()` and `DatabaseClient.singleUseReadOnlyTransaction()` needs to be more verbose.

This is the documentation for both of these:
> Returns a context in which a single read can be performed using TimestampBound.strong() concurrency.

> Returns a read-only transaction context in which a single read or query can be performed using TimestampBound.strong() concurrency.

By looking at the code, the only difference I can really find is that `ReadOnlyTransaction` (returned by `singleUseReadOnlyTransaction()`) holds the `read_timestamp` selected internally by Spanner once the `Read` or query finishes. If I'm reading it correctly, that means that `singleUse()` is not less transactional than `singleUseReadOnlyTransaction()` is. Is there any reason (beyond compatibility) not to remove `singleUse()` from the `DatabaseClient`?

olavloite:
Fixes #4212

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
@@ -91,16 +91,13 @@ public interface DatabaseClient {
    *
    * <p>Example of single use.
    *
-   * <pre>
-   * {
-   *   &#64;code
-   *   long singerId = my_singer_id;
-   *   String column = "FirstName";
-   *   Struct row =
-   *       dbClient.singleUse().readRow("Singers", Key.of(singerId), Collections.singleton(column));
-   *   String firstName = row.getString(column);
-   * }
-   * </pre>
+   * <pre>{@code
+   * long singerId = my_singer_id;
+   * String column = "FirstName";
+   * Struct row =
+   *     dbClient.singleUse().readRow("Singers", Key.of(singerId), Collections.singleton(column));
+   * String firstName = row.getString(column);
+   * }</pre>
    */
   ReadContext singleUse();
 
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -85,34 +85,43 @@
...
   /**
    * Returns a context in which a single read can be performed using {@link TimestampBound#strong()}
-   * concurrency.
+   * concurrency. This method will return a {@link ReadContext} that will not return the read
+   * timestamp that was used by Cloud Spanner. If you want to be able to access the read timestamp,
+   * you should use the method {@link #singleUseReadOnlyTransaction()}.
    *
    * <p>Example of single use.
    *
-   * <pre>{@code
```
kolea2: Do we need to remove `@code`?

olavloite: It's no removed, but changed by the formatter into '&#64;code'.

kolea2: That's odd - I tried running it myself and didn't see the same output (stayed the same). Can you try resubmitting as `@code` and see if the linter passes?

olavloite: The wrong formatting was first introduced by my IDE (Eclipse), and the mvn plugin did not automatically change it back to the original value. It's now changed back to `@code`.

PR #4726, opened 2019-03-22
Partition token

gyang-google:
Add PartitionTestUtil.java to extract PartitionToken out of Partition class for internal test.


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionTestUtil.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2017 Google LLC
```
sduskis: s/2017/2019

PR #4728, opened 2019-03-23
Spanner: Fix default value of maxSessions in document of setMaxSessions

rock619:
Document of setMaxSessions says it defaults to 2000, but it is wrong.
https://github.com/googleapis/google-cloud-java/blob/877f9c9bc1f8f04713396b8585cef3f3751a0439/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java#L24
Default value of maxSessions is 400, so I fixed it.


kolea2:
Thanks for the fix!

PR #4729, opened 2019-03-23
google-cloud-logging-logback: Allow user to specify custom LoggingOptions

### LINKED ISSUE #3215: Injecting google cloud credentials instead GOOGLE_APPLICATION_CREDENTIALS environment
Please provide a way to inject google credentials information for the stackdriver logger by means of public method/property instead of placing a private key file on machine and refer it from the GOOGLE_APPLICATION_CREDENTIALS environment variable.

Thanks.

olavloite:
Allow users to subclass com.google.cloud.logging.logback.LoggingAppender to specify custom LoggingOptions, such as custom credentials.
Fixes #3215

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/README.md
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/README.md
@@ -50,6 +50,8 @@ See [Logback filters](https://logback.qos.ch/manual/filters.html#thresholdFilter
       <level>INFO</level>
     </filter>
     <log>application.log</log> <!-- Optional : default java.log -->
+    <!-- Optional : will use the default credentials of the environment if this property is not set -->
+    <credentialsFile>/path/to/credentials.json</credentialsFile>
     <enhancer>com.example.enhancers.TestLoggingEnhancer</enhancer> <!-- Optional -->
     <enhancer>com.example.enhancers.AnotherEnhancer</enhancer> <!-- Optional -->
     <flushLevel>WARN</flushLevel> <!-- Optional : default ERROR -->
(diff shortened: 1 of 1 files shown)


ajaaym:
WDYT about adding setter method for credential file so user can configure from xml rather than writing code to subclass?

olavloite:
> WDYT about adding setter method for credential file so user can configure from xml rather than writing code to subclass?

Yes, you are absolutely right. I'll change it to support that.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
```diff
@@ -212,13 +230,32 @@ Logging getLogging() {
...
-          logging = LoggingOptions.getDefaultInstance().getService();
+          logging = createLoggingOptions().getService();
         }
       }
     }
     return logging;
   }
 
+  /** Creates the {@link LoggingOptions} to use for this {@link LoggingAppender}. */
+  LoggingOptions createLoggingOptions() {
```
kolea2: Can we do a similar approach as LoggingHandler (saving loggingOptions as a variable)?

olavloite: Changed to store the LoggingOptions as a variable.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/README.md
```diff
@@ -88,6 +88,23 @@ Authentication
 See the [Authentication](https://github.com/googleapis/google-cloud-java#authentication)
 section in the base directory's README.
 
+You can also specify custom credentials and other options by creating a subclass of `com.google.cloud.logging.logback.LoggingAppender` and override the method `createLoggingOptions()`. Your logback.xml configuration file must reference your subclass instead of `com.google.cloud.logging.logback.LoggingAppender`.
+
```
kolea2: do we still need these instructions for overriding createLoggingOptions?

olavloite: You're right, I changed these instructions into using the new custom property fileCredentials.

PR #4734, opened 2019-03-24
Spanner: Re-create sessions that have been invalidated by the server

olavloite:
Sessions in the session pool that have been invalidated by the server may cause `SessionNotFoundException`s to bubble up to the user, even though the user has no control over the session management. When this occurs, the session should be removed from the SessionPool and a new session should be fetched from the SessionPool without the user having to intervene.
In case of a `SessionNotFoundException` being thrown while using a `TransactionManager`, the client library will replace the session with a new one from the pool, and trigger a restart of the transaction by throwing an `AbortedException` (it is not possible to replace the session without also restarting the transaction).

This change has been initiated on request from the Spanner team.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
@@ -36,7 +36,7 @@ class DatabaseClientImpl implements DatabaseClient {
     TraceUtil.exportSpans(READ_WRITE_TRANSACTION, READ_ONLY_TRANSACTION, PARTITION_DML_TRANSACTION);
   }
 
-  private final SessionPool pool;
+  @VisibleForTesting final SessionPool pool;
 
   DatabaseClientImpl(SessionPool pool) {
     this.pool = pool;
@@ -50,11 +50,6 @@ class DatabaseClientImpl implements DatabaseClient {
   @VisibleForTesting
   PooledSession getReadWriteSession() {
     return pool.getReadWriteSession();
-  }
-
-  @VisibleForTesting
-  int getNumberOfAvailableWritePreparedSessions() {
-    return pool.getNumberOfAvailableWritePreparedSessions();
   }
 
   @Override
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryOnInvalidatedSessionTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryOnInvalidatedSessionTest.java
@@ -245,7 +245,7 @@ public class RetryOnInvalidatedSessionTest {
     service.shutdown();
     service.awaitTermination(10L, TimeUnit.SECONDS);
     Stopwatch watch = Stopwatch.createStarted();
-    while (((DatabaseClientImpl) client).getNumberOfAvailableWritePreparedSessions() == 0) {
+    while (((DatabaseClientImpl) client).pool.getNumberOfAvailableWritePreparedSessions() == 0) {
       if (watch.elapsed(TimeUnit.MILLISECONDS) > 1000L) {
         fail("No read/write sessions prepared");
       }


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -75,68 +76,119 @@ Instant instant() {
...
+                session.markUsed();
+                beforeFirst = false;
+              }
+              if (!ret && isSingleUse) {
+                close();
+              }
+              return ret;
+            } catch (SessionNotFoundException e) {
+              recreateSessionIfPossible(e);
+              replaceDelegate(resultSetSupplier.get());
```
sduskis: can we just do a `return next()` here instead of adding `while(true)`?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -206,7 +268,16 @@ public Timestamp getReadTimestamp() {
...
     public TransactionContext begin() {
-      return delegate.begin();
+      while (true) {
+        try {
+          TransactionContext res = delegate.begin();
+          session.markUsed();
+          return res;
+        } catch (SessionNotFoundException e) {
+          session.maybeRecreateUnderlyingSession(e);
+          delegate = session.delegate.transactionManager();
```
sduskis: can you just use `return begin()` instead of adding the `while` loop?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -75,68 +76,119 @@ Instant instant() {
...
+      session.maybeRecreateUnderlyingSession(e);
+      session.markUsed();
+      readContextDelegate = readContextDelegateSupplier.get();
+    }
+
     @Override
     public ResultSet read(
-        String table, KeySet keys, Iterable<String> columns, ReadOption... options) {
-      return wrap(delegate.read(table, keys, columns, options));
+        final String table,
```
sduskis: Can you please make style changes in a separate PR?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -378,17 +496,33 @@ public ReadOnlyTransaction singleUseReadOnlyTransaction(TimestampBound bound) {
...
         throw e;
       }
     }
 
     @Override
-    public ReadOnlyTransaction readOnlyTransaction(TimestampBound bound) {
+    public ReadOnlyTransaction readOnlyTransaction(final TimestampBound bound) {
       try {
-        return new AutoClosingReadTransaction(delegate.readOnlyTransaction(bound), this, false);
+        return new AutoClosingReadTransaction(
```
sduskis: This snippet is used 4 times.  Please create a method for this snippet.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -317,13 +317,17 @@ public DatabaseClient getDatabaseClient(DatabaseId db) {
...
         SessionPool pool = SessionPool.createPool(getOptions(), db, SpannerImpl.this);
-        DatabaseClientImpl dbClient = new DatabaseClientImpl(pool);
+        DatabaseClientImpl dbClient = createDatabaseClient(pool);
         dbClients.put(db, dbClient);
         return dbClient;
       }
     }
   }
 
+  DatabaseClientImpl createDatabaseClient(SessionPool pool) {
```
sduskis: Please make these types of style changes in a different PR.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: I replaced the while(true) loop with an internal method that is can be re-used in the SessionNotFound handler.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: I replaced the while(true) loop with an internal method that is can be re-used in the SessionNotFound handler.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: This change is made automatically by the code formatter, as the method parameters are changed to final, which is needed in order to be able to reference them in the anonymous inner class.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Added method to reduce code duplication.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
olavloite: I'll keep that in mind for future changes.
For now, I added a missing `@VisibleForTesting` annotation.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -75,35 +76,66 @@ Instant instant() {
...
+      } catch (SessionNotFoundException e) {
+        recreateSessionIfPossible(e);
+        res = resultSetSupplier.get();
       }
-      return new ForwardingResultSet(resultSet) {
+      return new ForwardingResultSet(res) {
+        private boolean beforeFirst = true;
+
         @Override
         public boolean next() throws SpannerException {
```
kolea2: can we change to isNext() and isInternalNext()?

kolea2: nevermind - didn't realize this wasn't a new method

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -75,35 +76,66 @@ Instant instant() {
...
 
     private AutoClosingReadContext(
-        ReadContext delegate, PooledSession session, boolean isSingleUse) {
-      this.delegate = delegate;
+        Supplier<T> delegateSupplier, PooledSession session, boolean isSingleUse) {
+      this.readContextDelegateSupplier = delegateSupplier;
       this.session = session;
       this.isSingleUse = isSingleUse;
+      try {
+        this.readContextDelegate = delegateSupplier.get();
```
kolea2: I would use this.readContextDelegateSupplier

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -114,29 +146,56 @@ public boolean next() throws SpannerException {
...
+            AutoClosingReadContext.this.close();
+          }
         }
       };
     }
 
+    private void recreateSessionIfPossible(SessionNotFoundException e) {
+      session.maybeRecreateUnderlyingSession(e);
+      session.markUsed();
+      readContextDelegate = readContextDelegateSupplier.get();
```
kolea2: can we just call getReadContextDelegate()?

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
```diff
@@ -53,6 +58,13 @@ public GceTestEnvConfig() {
     if (!serverUrl.isEmpty()) {
       builder.setHost(serverUrl);
     }
+    if (!credentialsFile.isEmpty()) {
```
kolea2: Can you please clarify what this change is?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: I'm not sure I understand what you mean. The variable `this.readContextDelegate` is of type `T extends ReadContext`, so I don't think it should be named `this.readContextDelegateSupplier`. Or do you mean something else?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: No, it's important that `Supplier.get()` is called here, as the `Supplier` should (have the chance to) re-create the `ReadContext`. If the session was re-created at line 157, the `Supplier` must also re-create the `ReadContext` for it to be associated with the new session.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
olavloite: This variable works in the same way as the serverUrl variable. It is read from a custom environment variable named `spanner.gce.config.credentials_file` (defined in `GceTestEnvConfig#GCE_CREDENTIALS_FILE`) and these credentials are used if this custom environment variable has been set. If it has not been set, the integration test will use the default credentials.
The reason that I have added this option is that when I run the integration tests locally, I use a number of different environments that need different credentials, and changing the default credentials of my local environment is a lot more cumbersome than just passing in a custom system property when running the tests.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
kolea2: Apologies for not being clear - what I meant was I would have expected this line to be
`this.readContextDelegate = this.readContextDelegateSupplier.get();`
since we just assigned it earlier. Let me know if I'm missing something.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
kolea2: Gotcha - since this is already a larger PR, can we move this out into a separate change?

olavloite: I moved the change to a new PR.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -289,32 +372,48 @@ private PooledSession(Session delegate) {
...
     }
 
+    @VisibleForTesting
+    void setAllowDelegateRecreation(boolean allow) {
+      allowDelegateRecreation = allow;
+    }
+
+    private void maybeRecreateUnderlyingSession(SessionNotFoundException e) {
+      if (beforeFirstServerCall && allowDelegateRecreation) {
+        delegate = spanner.createSession(db);
```
snehashah16: could we get another session from the pool instead of creating a new one ?

I am afraid that the latency on this request will be too high when we internally retry.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -289,32 +372,48 @@ private PooledSession(Session delegate) {
...
+        lastException = null;
+        return internalRunWithSessionRetry(method);
       }
     }
 
-    @Override
-    public long executePartitionedUpdate(Statement stmt) throws SpannerException {
+    private <T> T internalRunWithSessionRetry(Supplier<T> method) {
       try {
+        T res = method.get();
```
snehashah16: should this be after markUsed() ?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -289,32 +372,48 @@ private PooledSession(Session delegate) {
...
-        return delegate.write(mutations);
-      } catch (SpannerException e) {
-        throw lastException = e;
-      } finally {
-        close();
+        return internalRunWithSessionRetry(method);
+      } catch (SessionNotFoundException e) {
+        maybeRecreateUnderlyingSession(e);
+        lastException = null;
+        return internalRunWithSessionRetry(method);
```
snehashah16: there is a chance we got sessionNotFound here as well.. 

have u considered putting this in a loop ? 
Pros: customers never see a SessionNotFound
Cons: high latency, but our timeouts should still take effect.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -404,8 +557,15 @@ public TransactionRunner readWriteTransaction() {
         @Nullable
         public <T> T run(TransactionCallable<T> callable) {
           try {
+            T result;
+            try {
+              result = runner.run(callable);
+            } catch (SessionNotFoundException e) {
+              maybeRecreateUnderlyingSession(e);
+              TransactionRunner newRunner = delegate.readWriteTransaction();
+              result = newRunner.run(callable);
```
snehashah16: same as above, this could potentially still see a sessionNotFound (SNF)

sduskis:
@olavloite, it looks like there are conflicts.

olavloite:
@sduskis Yeah, I know. I'll start to rebase it once #4895 has been merged, as that will otherwise also cause further merge conflicts with this change.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
kolea2: Should the while(true) still be here?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -919,6 +1245,24 @@ Session getReadWriteSession() {
...
+  private PooledSession replaceReadSession(SessionNotFoundException e, PooledSession session) {
+    if (!options.isFailIfSessionNotFound() && session.allowReplacing) {
+      closeSessionAsync(session);
+      return getReadSession();
+    } else {
+      throw e;
+    }
+  }
+
+  PooledSession replaceReadWriteSession(SessionNotFoundException e, PooledSession session) {
```
kolea2: should these have the same visibility?

olavloite: `replaceReadWriteSession` needs at least package visibility, as it is called from `DatabaseClientImpl`. `replaceReadSession` does not need more than private visibility, as it is only called from withing `SessionPool.java`.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -146,24 +176,48 @@ public TransactionRunner readWriteTransaction() {
...
+  }
+
+  private <T> T runWithSessionRetry(SessionMode sessionMode, Function<Session, T> callable) {
+    PooledSession session =
+        sessionMode == SessionMode.READ ? getReadSession() : getReadWriteSession();
+    while (true) {
+      try {
+        return callable.apply(session);
+      } catch (SessionNotFoundException e) {
+        session = pool.replaceReadWriteSession(e, session);
```
kolea2: Following up from my last comment, if the session is READ, is this correct, as replaceReadWriteSession eventually returns 'getReadWriteSession()'?

olavloite: Good point, it wasn't right, even though the method was never called with READ. But that was not correct either, as `executePartitionedUpdate` does not need a read/write session, as it always starts a new transaction (as it needs a special type of transaction).

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -188,7 +188,7 @@ public long executePartitionedUpdate(final Statement stmt) {
     Span span = tracer.spanBuilder(PARTITION_DML_TRANSACTION).startSpan();
     try (Scope s = tracer.withSpan(span)) {
       return runWithSessionRetry(
-          SessionMode.READ_WRITE,
+          SessionMode.READ, // PartitionedUpdate does not need a prepared tx.
```
kolea2: Can we do this in a separate PR? Because this PR is quite large, I'd like to separate out the recreating sessions implementation from the refactorings/fixes.

olavloite: I'll move it to a separate PR later this afternoon.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -39,11 +42,32 @@
...
+    return pool.getReadSession();
+  }
+
+  @VisibleForTesting
+  PooledSession getReadWriteSession() {
+    return pool.getReadWriteSession();
+  }
+
+  @VisibleForTesting
+  int getNumberOfAvailableWritePreparedSessions() {
```
snehashah16: this is not DBClient's responsibility. I wouldn't expose it here

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -69,7 +99,7 @@ public Timestamp writeAtLeastOnce(Iterable<Mutation> mutations) throws SpannerEx
   public ReadContext singleUse() {
     Span span = tracer.spanBuilder(READ_ONLY_TRANSACTION).startSpan();
     try (Scope s = tracer.withSpan(span)) {
-      return pool.getReadSession().singleUse();
+      return getReadSession().singleUse();
```
snehashah16: curious whats the behavior with read sessions seeing SNF errors ?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -146,24 +176,41 @@ public TransactionRunner readWriteTransaction() {
...
+          });
     } catch (RuntimeException e) {
       TraceUtil.endSpanWithFailure(span, e);
       throw e;
     }
   }
 
+  private <T> T runWithSessionRetry(Function<Session, T> callable) {
+    PooledSession session = getReadWriteSession();
+    while (true) {
```
snehashah16: Will this respect client side timeout values set ?
If our backend is down, and does not return any sessions -> whats the outcome of the customer application ?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
olavloite: Method is now removed and the visibility of the session pool changed to package-private to make it visible for test cases.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
olavloite: The SNF is detected at the first actual database request. For read-only transactions (both single use and actual read transactions), that is at the first query. The query is sent to the server at the first call to ResultSet#next(), so the SNF for a read transaction is handled here:
https://github.com/googleapis/google-cloud-java/blob/1b61e287e97cc9cc17333caf50413346edfd00b6/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L129

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
olavloite: The short answer: It will behave the same as now, and the behavior will depend on whether the pool has a session ready to be returned or not. The `runWithSessionRetry` will only retry on the specific condition of a `SessionNotFoundException` being returned.

There are two possible scenarios:
1. The application requests a session and the session pool has a session that it thinks is valid. This session is handed to the application and the application tries to execute the transaction. This will cause the query or begin transaction statement to time out.
2. The pool does not have a session available that can be returned and starts the creation of a new session asynchronously. The requesting thread is placed in a waiter. The create session request will fail and notify the waiter of the error. The waiting thread will return with the create session error.

Both of the above will not be considered a `SessionNotFoundException` by the `runWithSessionRetry` method, and the method will return with an error.

PR #4741, opened 2019-03-26
Regenerate Firestore clients

andreamlin:
The firestore GAPIC configs were updated to use Resource Name classes.

This also removes the string formatting functions, but there is no semantic loss in functionality because the existing resource name classes are superior. Firestore is a beta client, so it is more acceptable to remove these extraneous functions.


andreamlin:
PTAL

PR #4742, opened 2019-03-26
Regen Monitoring with STATIC_TYPES

andreamlin:
Also enable the string formatting functions so as to not break this GA client.


andreamlin:
whoops sorry i'll move this over to my fork

PR #4745, opened 2019-03-27
Refresh Iamcredentials client

andreamlin:
This diff was generated by `python3 synth.py` in the relevant API directory.


andreamlin:
PTAL

sduskis:
@andreamlin, it looks like the path methods were removed.  The library is a `0.*` version, but the API is `v1`.  Would it make sense to restore the path methods?

andreamlin:
@sduskis The path methods were indeed removed. I was hoping this wouldn't be too bad because they are GAPIC-specific methods that could be removed because the client itself is not GA, even though the underlying API is. 

Though this is a breaking change, there is no practical loss of functionality because the pre-existing generated Resource Name classes are superior in usability.

PR #4749, opened 2019-03-27
Spanner: Refactor SpannerImpl.java

olavloite:
SpannerImpl has become quite long and difficult to read. This PR splits it into a number of logical parts.

The following new files have been added as a result of splitting SpannerImpl (all classes defined at top-level of these files are package-private):

- `AbstractReadContext.java`: Contains the abstract base class of all `com.google.cloud.spanner.ReadContext` implementations, as well as all concrete implementations of  **read-only** `ReadContext`s. Read-write transactions, that also implement `ReadContext`, are not included in this file.
- `AbstractResultSet.java`: Contains the abstract base class for `ResultSet` implementations and the concrete implementation of `GrpcResultSet` (the `ResultSet` returned by database queries on Spanner).
- `DatabaseAdminClientImpl.java`: The concrete implementation of `DatabaseAdminClient`. Used to be an inner class of `SpannerImpl`.
- `InstanceAdminClientImpl.java`: The concrete implementation of `InstanceAdminClient`. Used to be an inner class of `SpannerImpl`.
- `PartitionedDmlTransaction.java`: Implementation of partitioned DML transactions for Spanner. Used to be an inner class of `SpannerImpl`.
- `SessionImpl.java`: Implementation of `Session`. Used to be an inner class of `SpannerImpl`.
- `TransactionRunnerImpl.java`: Implemenation of `TransactionRunner` and `TransactionContext`. `TransactionContext` is also an implementation of `ReadContext` and a subclass of `AbstractReadContext`, but defined in this file as it is only used by 'TransactionRunnerImpl`. Used to be an inner class of `SpannerImpl`.

As a result of the above, SpannerImpl is reduced from more than 3,000 lines of code to approx 360 lines of code.

Changes to other files are a result of updated references (i.e. no longer referencing `SpannerImpl.SessionImpl` but `SessionImpl`), the need to pass in a reference to SpannerImpl or SpannerGrpc, the need to change private inner classes into package-private inner classes or the need to change private methods and final fields in SpannerImpl into package-private methods and final fields.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
@@ -20,22 +20,7 @@ import static com.google.cloud.spanner.SpannerExceptionFactory.newSpannerExcepti
 import static com.google.cloud.spanner.SpannerExceptionFactory.newSpannerExceptionForCancellation;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executor;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
-import java.util.logging.Logger;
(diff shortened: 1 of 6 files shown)


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -401,18 +326,9 @@ Object value() {
     return ImmutableMap.copyOf(tmp);
   }
 
-  private static <T extends Message> T unpack(Any response, Class<T> clazz)
```
olavloite: This method was removed as it turned out that it was never used.

olavloite:
Replaced by separate PR's per change set.

PR #4755, opened 2019-03-27
Refresh websecurity client

andreamlin:
websecurityscanner_gapic.yaml was just updated with resource name configurations.

This does remove the path formatting methods. Those methods are soon-to-be deprecated, and their functionality is already covered by the better generated resource name classes. This is a non-GA client.


andreamlin:
PTAL

PR #4756, opened 2019-03-28
LoggingAppender - removed DCL

saturnism:
Avoid using deprecated newInstance() method.
Removed double checked locking pattern - it doesn't work, and unnecessary, since start() is already synchronized.


sduskis:
@saturnism, all of the builds fail.

sduskis:
Can we remove the `getLogging()` method?

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
```diff
@@ -228,13 +239,6 @@ public synchronized void stop() {
   }
 
   Logging getLogging() {
-    if (logging == null) {
-      synchronized (this) {
-        if (logging == null) {
-          logging = getLoggingOptions().getService();
-        }
-      }
-    }
```
ajaaym: where logging service would be created? Public constructor dont have logging parameter and other constructor with logging parameter is package private.

sduskis: The change is to create the logging service in start.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java
```diff
@@ -190,18 +198,21 @@ public synchronized void start() {
     if (isStarted()) {
       return;
     }
+    super.start();
+
+    if (logging == null) {
+      logging = LoggingOptions.getDefaultInstance().getService();
```
ajaaym: can you please use getLoggingOptions as it sets the credential if configured.

sduskis:
@saturnism, can you please address the [issue](https://github.com/googleapis/google-cloud-java/pull/4756#discussion_r271746551) that @ajaaym raised?

PR #4781, opened 2019-03-28
[WIP] Spanner - Use gax for runWithRetries()

olavloite:
Instead of using its own retry logic, SpannerImpl should use the retry helper and options from gax.
This PR is based on #4749 Spanner refactor spannerimpl, which should be processed first.

Issues / discussion:

SpannerImpl already uses `io.grpc.Context` to enable setting timeouts and cancelling database operations. This will interfere with using the functionality from gax. The implementation in this PR supports both, but the behavior is different from the previous version if the user for example sets a short timeout on `Context` in combination with a longer timeout on `RetrySettings`. If the timeout of `RetrySettings` is longer than the timeout set on `Context`, the context timeout will only occur after the timeout of `RetrySettings` has occurred (i.e. the time before the timeout exception is thrown is longer than the user would have expected).

Possible solutions as I see them at this moment:

1. Remove support for using `io.grpc.Context` for setting timeouts and cancelling database operations, and only rely on the functionality from gax (which use the `RetrySettings`). This would change the behavior for clients who currently use this to set a timeout or to cancel calls.
1. Use the deadline from `io.grpc.Context` if one is set (or if the current context is a CancellableContext), and use gax in all other cases. This basically means 'use the current functionality if the user has set a deadline on the Context or indicated that it should be cancellable, and use `RetrySettings` in all other cases'.
1. Do not migrate this functionality to use gax.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
@@ -25,7 +25,6 @@ import com.google.api.gax.paging.Page;
 import com.google.api.pathtemplate.PathTemplate;
 import com.google.cloud.spanner.Options.ListOption;
 import com.google.cloud.spanner.SpannerImpl.PageFetcher;
-import com.google.cloud.spanner.spi.v1.SpannerRpc;
 import com.google.cloud.spanner.spi.v1.SpannerRpc.Paginated;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.FieldMask;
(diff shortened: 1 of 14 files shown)


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/InstanceAdminClientImpl.java
```diff
@@ -0,0 +1,213 @@
+/*
```
kolea2: are these chances part of the overall refactoring?

olavloite: Yes, this class used to be an inner class of SpannerImpl.

ajaaym:
@olavloite ApiFuture returned by callable do support cancellation of calls. Also with every call to callable, we can provide call context with deadline. I think whole of retrying logic can be migrated to gax and configured in handwritten client.

olavloite:
> @olavloite ApiFuture returned by callable do support cancellation of calls. Also with every call to callable, we can provide call context with deadline. I think whole of retrying logic can be migrated to gax and configured in handwritten client.

@ajaaym Yes, I agree with you on that, but as far as I can see we will be changing the behavior for users that currently set a deadline or cancellation themselves. Consider the code in `SpannerImplRetryTest.contextDeadlineExceeded()` and assume that this was not a test case, but code written by a user. Moving the retry logic of Spanner from the current implementation to gax will lead to the question: Which deadline should we apply to a database operation. The one specified in the `RetrySettings` of the `SpannerOptions` that was used to create the `Spanner` instance, or the one the user has set on the call context? Gax will by default use the one specified in `RetrySettings`.

Or am I missing something here?

olavloite:
Replaced by #4940

PR #4782, opened 2019-03-28
Spanner: Allow specifying a custom credentials file for integration tests

olavloite:
The integration test configuration allowed users to specify a custom project id and host for the integration test, but always used the default credentials of the environment. This change adds support for specifying a specific credentials file to use for a test run. If no credentials file is specified, the tests will continue to use the default credentials of the environment.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java
@@ -18,6 +18,7 @@ package com.google.cloud.spanner;
 
 import static com.google.common.base.Preconditions.checkState;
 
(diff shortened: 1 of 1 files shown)


igorbernstein2:
Just out of curiosity, why couldn't this be overriden using the existing `GOOGLE_APPLICATION_CREDENTIALS` environment variable? https://cloud.google.com/docs/authentication/production

olavloite:
@igorbernstein2 This is more convenient when running integration tests on different servers that require different credentials directly from my IDE (Eclipse). I can't easily set different environment variables in test run configurations in Eclipse, but I can easily set different JVM parameters for different test run configurations.
It is of course also possible to set an environment variable at runtime in Java, but this setup is equal to the existing setup used to specify the project ID and custom server hostname in GceTestEnvConfig, keeping the configuration options of integration tests consistent.

igorbernstein2:
Thanks, I was just curious if this is a new pattern that I should follow for bigtable

PR #4784, opened 2019-03-28
Logging: call flush before close

### LINKED ISSUE #4783: Logging pending writes not flushed on close causing hanging
Context: https://github.com/spring-cloud/spring-cloud-gcp/issues/1608.

meltsufin:
Also prevent more logs to be written after close.

Fixes #4783.

### DIFF
--- a/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
+++ b/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
@@ -582,8 +582,7 @@ class LoggingImpl extends BaseService<LoggingOptions> implements Logging {
 
   /* Write logs synchronously or asynchronously based on writeSynchronicity setting. */
   private void writeLogEntries(Iterable<LogEntry> logEntries, WriteOption... writeOptions) {
-    if (closed)
-      return;
+    if (closed) return;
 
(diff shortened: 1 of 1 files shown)


sduskis:
There was a similar PR in #3909.  Right now, `flush()` ought to be called, and ideally we force the flushing to occur.  Unfortunately, `flush()` just waits for a timer to complete.  @chingor13 didn't want to merge the change for the first attempt at this bug fix.  We need to figure out how to fix this for the long haul, once @chingor13 is back.

meltsufin:
@sduskis The critical problem is that everything gets closed while there are pending API calls. Calling `flush()` before `close()` actually fixes the issue.
This is a critical issue for us in Spring Cloud GCP (https://github.com/googleapis/google-cloud-java/pull/4784) as it currently hangs apps for 5 minutes on startup as GRPC is unable to complete those pending requests after the logging service was shut down.
Unless you think there is harm in this change, can we merge this, and @chingor13 can redo the fix when he comes back?

sduskis:
@meltsufin, I understand why this is critical.  Merging this is @chingor13's call.

sduskis:
@meltsufin, can you please run `mvn com.coveo:fmt-maven-plugin:format` to appease the linter?

meltsufin:
@sduskis Done.

dzou:
@chingor13 Hey, any updates regarding your thoughts on this PR?

PR #4819, opened 2019-03-29
Upgrade protobuf version to 3.7.0.

kolea2:
This will fix the currently failing regen PRs.


andreamlin:
Could we move a bit higher to 3.7.1?
 I think artman was just updated to use 3.7.1

kolea2:
Oops, there is a 3.7.1 version, but looks like not for protobuf-java: https://search.maven.org/artifact/com.google.protobuf/protobuf-java/3.7.0/bundle

andreamlin:
oh good call

PR #4831, opened 2019-03-31
google-cloud-core: Optimize Timestamp.parseTimestamp(String)

olavloite:
Stress testing the Spanner client library showed that more than 50% of the total **cpu** time consumed by the client library while executing queries and consuming result sets is spent on parsing dates and timestamps. This stress test uses a result set containing one column of each possible data type of Spanner (both single value column and array column). Spanner automatically translates the string values that are sent by the server into Date and Timestamp objects before the user requests these from the ResultSet.

In a real-life scenario, the total percentage of total time spent on parsing these values will however be a lot lower, as most of the time the client library will be waiting on network traffic.

The total execution time of the stress tests added for Spanner on my own laptop with and without this optimization is:
- Without optimization (3 test runs): 71.1s, 69.5s, 67.0s
- With optimization (3 test runs): 34.1s, 34.7s, 35.1s

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
+++ b/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
@@ -200,6 +200,7 @@ public final class Timestamp implements Comparable<Timestamp>, Serializable {
       int fraction = 0;
       if (timestamp.length() > 20) {
         if (timestamp.charAt(19) == '.') {
+          // The timestamp contains a fraction.
           int endIndex;
           if (timestamp.charAt(timestamp.length() - 1) == 'Z'
               || timestamp.charAt(timestamp.length() - 1) == 'z') {
@@ -211,11 +212,9 @@ public final class Timestamp implements Comparable<Timestamp>, Serializable {
             throw new IllegalArgumentException(invalidTimestamp);
           }
           fraction = Integer.parseInt(timestamp.substring(20, endIndex));
-          // Adjust the result to nanoseconds if the input length is less than 9 digits (9 -
-          // (endIndex
-          // - 20)).
+          // Adjust the result to nanoseconds if the input length is less than 9 digits
+          // (9 - (endIndex - 20)).
           fraction *= POWERS_OF_10[9 - (endIndex - 20)];
-        } else {
         }
       }
       LocalDateTime ldt = LocalDateTime.of(year, month, day, hour, minute, second, fraction);


google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/IntParser.java
```diff
@@ -0,0 +1,82 @@
...
+    Preconditions.checkArgument(end <= input.length(), "End must be <= input.length()");
+    int res = 0;
+    for (int index = begin; index < end; index++) {
+      res += parseDigit(input.charAt(index), input) * POWERS_OF_10[end - index - 1];
+    }
+    return res * POWERS_OF_10[exponent];
+  }
+
+  private static int parseDigit(char c, String input) {
+    if (c == '0') {
```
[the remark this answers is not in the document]
olavloite: I changed it to use a switch statement instead of the block of if-statements.

I also tried using `Character.getNumericValue`, but that is slower than the switch-statement, as it also checks a for number of additional unicode characters, such as `'\u216C'`, that are not relevant for timestamp parsing.

sduskis:
@olavloite, is this ready for review?

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/IntParser.java
```diff
@@ -0,0 +1,82 @@
...
+      case '6':
+        return 6;
+      case '7':
+        return 7;
+      case '8':
+        return 8;
+      case '9':
+        return 9;
+    }
+    throw new NumberFormatException("Not a decimal digit: " + c);
```
sduskis: Why not `Character. getNumericValue()`, and check for -1?

sduskis:
Date parsing is non-trivial.  I'm not sure we ought to improve times this way.  Is there any way to improve performance while still using a standard date parser?

olavloite:
Standard date/time parsers are based on regular expressions in order to support different date/time formats, which make them slower than an implementation that is specifically written specifically for the only format that we support. I understand the reluctance on changing this. One option could be to add this specific parse method to the Spanner library (or only the `GrpcResultSet` class) and only use it there.

I did some tests to get an idea of the actual difference in performance by parsing 1,000,000 random timestamps a number of times.

- Current implementation of `Timestamp.parseTimestamp(String)`: 2300ms
- Using `org.threeten.bp.Instant.parse(CharSequence)`: 1600ms
- The parse method in this PR: 330ms

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Date.java
```diff
@@ -57,13 +53,14 @@ public static Date fromYearMonthDay(int year, int month, int dayOfMonth) {
...
   /** @param date Data in RFC 3339 date format (yyyy-mm-dd). */
   public static Date parseDate(String date) {
-    Matcher matcher = FORMAT_REGEXP.matcher(date);
-    if (!matcher.matches()) {
-      throw new IllegalArgumentException("Invalid date: " + date);
-    }
-    int year = Integer.parseInt(matcher.group(1));
-    int month = Integer.parseInt(matcher.group(2));
-    int dayOfMonth = Integer.parseInt(matcher.group(3));
+    Preconditions.checkNotNull(date);
```
sduskis: This change seems straightforward.  Can you please break it out into a separate PR?

olavloite: Removed from this PR and submitted as separate PR.

google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/DateTest.java
```diff
@@ -38,6 +39,20 @@ public void parseDate() {
     assertThat(date.getYear()).isEqualTo(2016);
     assertThat(date.getMonth()).isEqualTo(9);
     assertThat(date.getDayOfMonth()).isEqualTo(18);
+    parseInvalidDate("2016/09/18");
+    parseInvalidDate("2016 09 18");
+    parseInvalidDate("2016-9-18");
+    parseInvalidDate("2016-09-18T10:00");
+    parseInvalidDate("");
```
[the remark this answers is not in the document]
olavloite: Added (see separate PR).
`com.google.cloud.Date` does however support invalid dates. A date like `2001/02/29` is accepted, both when you submit it through `Date.parseDate` as well as by constructing it using the `Date.fromYearMonthDay`. I have not changed that part of the behavior.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -186,12 +190,12 @@ public static Timestamp parseTimestamp(String timestamp) {
                 && timestamp.length() > 20
                 && (timestamp.charAt(20) != 'Z' || timestamp.charAt(20) == 'z')),
         invalidTimestamp);
-    int year = IntParser.parseInt(timestamp, 0, 4);
-    int month = IntParser.parseInt(timestamp, 5, 7);
-    int day = IntParser.parseInt(timestamp, 8, 10);
-    int hour = IntParser.parseInt(timestamp, 11, 13);
-    int minute = IntParser.parseInt(timestamp, 14, 16);
-    int second = IntParser.parseInt(timestamp, 17, 19);
+    int year = Integer.parseInt(timestamp.substring(0, 4));
```
sduskis: I think you need to catch `NumberFormatException`.

olavloite: I don't think that would add very much. The method already throws `IllegalArgumentException` if the input is not valid. `NumberFormatException` is a subclass of `IllegalArgumentException`, so it should be caught by any exception handling the client already has.
The only reason to do so would be to add a custom message. Is that what you mean?

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -176,14 +165,64 @@ public static Timestamp fromProto(com.google.protobuf.Timestamp proto) {
...
+          } else {
+            endIndex = timestamp.length();
+          }
+          if (endIndex - 20 > 9) {
+            throw new IllegalArgumentException(invalidTimestamp);
+          }
+          fraction = Integer.parseInt(timestamp.substring(20, endIndex));
+          // Adjust the result to nanoseconds if the input length is less than 9 digits (9 -
+          // (endIndex
+          // - 20)).
```
sduskis: Please have the mathematical formula on a single line, if possible.  `(9 - (endIndex - 20))`.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -176,14 +165,64 @@ public static Timestamp fromProto(com.google.protobuf.Timestamp proto) {
...
+          fraction *= POWERS_OF_10[9 - (endIndex - 20)];
+        } else {
```
sduskis: Does this `else` need to be here?

chingor13:
I'm reluctant to have us maintain our own date parsing code within this repo. Can we not tell provide the expected date format for one of the existing libraries so it can parse more performantly?

PR #4832, opened 2019-03-31
Spanner timeout wait for session

olavloite:
If the SessionPool has been exhausted and the (default) wait setting of BLOCK is used, the SessionPool would wait indefinitely for a session to become available. If the user had created a session leak situation, for example by leaving a lot of ResultSets open, this would mean that the application at a given moment would hang without any error message, making it harder for the user to debug what is actually happening. By allowing the session acquisition to timeout, the user is presented with an error message that will help him debug and fix the problem.


olavloite:
The default behavior will probably be changed from BLOCK to FAIL, making this change obsolete.

olavloite:
@snehashah16 Any update on this?

snehashah16:
we discussed to not change the current behavior.

We will continue to BLOCK for session creation. we are hoping the session improvements will help. we can revisit once all session improvements are completed and we get some  customer feedback.

PR #4878, opened 2019-04-02
Auto close input stream for MetadataConfig

saturnism:


sduskis:
Thanks for the contribution!

PR #4890, opened 2019-04-03
Spanner: Refactor SpannerImpl - Move AbstractReadContext to separate file

olavloite:
SpannerImpl is more than 3,000 lines of code and should be split into a number of logical separate files. This step moves AbstractReadContext and its concrete read-only sub classes to a separate file.


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -0,0 +1,558 @@
...
+          .setSingleUse(TransactionOptions.newBuilder().setReadOnly(bound.toProto()))
+          .build();
+    }
+  }
+
+  private static void assertTimestampAvailable(boolean available) {
+    checkState(available, "Method can only be called after read has returned data or finished");
+  }
+
+  static class SingleUseReadOnlyTransaction extends SingleReadContext
```
kolea2: can this be private?

olavloite: No, this class is referenced from the class SessionImpl.

PR #4901, opened 2019-04-03
Add web risk to list of clients.

kolea2:


sduskis:
Is the module in the bom?

kolea2:
Yes, this was the only place it was missing.

Issue #2452, opened 2017-09-19
Setting the Flow Control MaxOutstandingElementCount cause stuck messages


robertsaxby:
The behaviour was observed in the 0.21.1-beta version of the Java client library
It seems resolved in the 0.23.1-beta version, however this newer version implements the bi-directional streaming pull and the previous doesn't.

**Test Setup:**
- 100k messages published to a Topic
- Single Subscription
- _n_ Subscribers (k8s running on GKE, each pod running in separate node) {_n_ ∈ Z |  1 ≤ _n_ ≤ 4}
- Each node had 2 vCPU and each Subscriber provided 10 user threads to the client library
- All Subscribers running the same code (Parse message, build various classes, callback processing takes ~20ms)
- Flow control settings: MaxOutstandingElementCount: 1000
**Results:**
- _n_ Messages remain stuck on Topic _n_  ∈ [223, 6652]
**Observations:**
- When messages are stuck: mod_ack_deadline_request_count  > 0, Pull request count > 0
- When messages are stuck Subscriber callback threads are Waiting
- No stuck messages when retesting with default Flow Control Settings (i.e. MaxOutstandingElementCount = null)
- Experiments with MaxOutstandingElementCount ∈ [1000, 500, 250, 125] resulted in stuck message count ∈ [~6K, ~3K, ~1.7K, ~6K]

**Interpretation:**
Message ACK deadlines are being continuously extended and the callback threads are not being invoked

kir-titievsky:
@robertsaxby It sounds like the subscriber application code executes the listener MessageReceiver (parsing the message, etc), but never calls AckReplyConsumer.ack for some messages.  Can you confirm that:
* The MessageReceiver implementation actually calls AckReplyConsumer.ack explicitly in the user code.
* That for stuck messages the listener code executes successfully, including a call to .ack?

kir-titievsky:
Got answers offline: The call to .ack happens successfully in the application code, but the acks are never actually sent.

pongad:
I'll look into this, but @robertsaxby is there any reason you cannot move to streaming pull?

pongad:
Good news is I can reproduce this. I'm working to determine the root cause.

pongad:
I found the problem. @davidtorres @mdietz94

[...]

PR #4916, opened 2019-04-05
Change MessageDispatcher to be synchronous instead of asynchronous.

dpcollins-google:
This removes the failure mode described in #2452 that can occur when MaxOutstandingElementCount is low and there is more than one connection.  In this case, it is possible for an individual MessageDispatcher to have no outstanding in-flight messages, but also be blocked by flow control with a whole new batch outstanding. In this case, it will never make progress on that batch since it will never receive another batch and the queue was made to not be shared in  #4590, so the batch will never be pulled off by another MessageDispatcher.

By changing this to use a blocking flow controller, this will never happen, as each batch will synchronously wait until it is allowed by flow control before being processed.


kimkyung-goog:
I confirm that there are no conceptual conflicts between this change and ordering keys implementation.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -355,42 +316,25 @@ public void processReceivedMessages(List<ReceivedMessage> messages, Runnable don
...
+      outstandingBatch.add(new OutstandingMessage(message, ackHandler));
       pendingReceipts.add(message.getAckId());
     }
 
-    if (outstandingBatch.messages.isEmpty()) {
-      doneCallback.run();
-      return;
-    }
-
-    messagesWaiter.incrementPendingMessages(outstandingBatch.messages.size());
```
sduskis: can `messagesWaiter` be removed?

dpcollins-google: No, it is needed to block shutdown on outstanding messages in the stop method.

public void stop() {
    messagesWaiter.waitNoMessages();
    jobLock.lock();
    try {
      if (backgroundJob != null) {
        backgroundJob.cancel(false);
        backgroundJob = null;
      }
    } finally {
      jobLock.unlock();
    }
    processOutstandingAckOperations();
  }

PR #4920, opened 2019-04-06
google-cloud-core: Optimize Date.parseDate(String)

olavloite:
Date parsing using a regular expression is slower than a specific implementation for the only format that is supported.

`com.google.cloud.Date` supports invalid dates. A date like `2001-02-29` is accepted, both when you submit it through `Date.parseDate` as well as by constructing it using the `Date.fromYearMonthDay`. This behavior is not changed by this PR.

### DIFF
--- a/google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/DateTest.java
+++ b/google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/DateTest.java
@@ -64,13 +64,15 @@ public class DateTest {
     parseInvalidDate("aaaa-01-01");
     parseInvalidDate("2019-bb-01");
     parseInvalidDate("2019-01-cc");
-    parseInvalidDate("2000-13-01");
-    parseInvalidDate("2000-12-32");
+    parseInvalidMonth("2000-13-01");
+    parseInvalidMonth("2000-00-01");
+    parseInvalidDay("2000-12-32");
+    parseInvalidDay("2000-12-00");
     parseInvalidDate("10000-01-01");
-    parseInvalidDate("0000-01-01");
-    parseInvalidDate("-001-01-01");
-    parseInvalidDate("0001--1-01");
-    parseInvalidDate("0001-01--1");
+    parseInvalidYear("0000-01-01");
+    parseInvalidYear("-001-01-01");
+    parseInvalidMonth("0001--1-01");
+    parseInvalidDay("0001-01--1");
   }
 
   private void parseInvalidDate(String input) {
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/IntParser.java
```diff
@@ -0,0 +1,82 @@
...
+  static int parseInt(String input, int begin, int end, int exponent) {
+    Preconditions.checkNotNull(input);
+    Preconditions.checkArgument(
+        exponent >= 0 && exponent < POWERS_OF_10.length, "Exponent out of range");
+    Preconditions.checkArgument(end - begin <= 10, "Max input length is 10");
+    Preconditions.checkArgument(end >= begin, "End must be greater or equal to begin");
+    Preconditions.checkArgument(begin >= 0, "Begin must be >= 0");
+    Preconditions.checkArgument(end <= input.length(), "End must be <= input.length()");
+    int res = 0;
+    for (int index = begin; index < end; index++) {
```
sduskis: Is there a reason not to call `Integer.parseInt(input.substring(begin, end)` here?

google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/DateTest.java
```diff
@@ -34,10 +35,44 @@
...
+    parseInvalidDate("2016/09/18");
+    parseInvalidDate("2016 09 18");
+    parseInvalidDate("2016-9-18");
+    parseInvalidDate("2016-09-18T10:00");
+    parseInvalidDate("");
+    parseInvalidDate("test");
+    parseInvalidDate("2000-13-01");
+    parseInvalidDate("2000-12-32");
+    parseInvalidDate("10000-01-01");
+    parseInvalidDate("0000-01-01");
```
sduskis: please add `aaaa-bb-cc`

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/IntParser.java
olavloite: No, good point. Changed.

google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/DateTest.java
olavloite: Added.

google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/DateTest.java
```diff
@@ -34,10 +35,51 @@
...
+    parseInvalidDate("0001--1-01");
+    parseInvalidDate("0001-01--1");
+  }
+
+  private void parseInvalidDate(String input) {
+    try {
+      Date.parseDate(input);
+      fail("Expected exception");
+    } catch (IllegalArgumentException e) {
+      // Ignore, this is the expected exception.
```
sduskis: The message should contain `"Invalid date"`

olavloite: Added try-catch for `NumberFormatException` and check for `"Invalid date"` in the exception message.

PR #4923, opened 2019-04-08
BigQuery: Added listPartitions.

### LINKED ISSUE #3097: [BigQuery] Add possibility to list partitions
It would be great if java version of bq library would have the similar method for listing partitions as it's in Python version:
https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/client.py#L1336-L1357

pmakani:
Fixes #3097

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
@@ -644,21 +644,20 @@ public class ITBigQueryTest {
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


pmakani:
> Thanks for this!
> 
> One minor nit: I suspect the integration test may end up flaky due to timing issues with tabledata.insertAll buffer attachment. For a safer test, consider building a table using DDL, which will commit the results immediately and avoid dealing with the streaming buffer.
> 
> Ex:
> CREATE OR REPLACE TABLE myproject.mydataset.mytable
> PARTITION BY DATE(tscol)
> AS SELECT
> CURRENT_TIMESTAMP() as tscol,
> "foo" as othercol

Done. PTAL.

PR #4935, opened 2019-04-10
#4933 Adding support to custom ports on LocalDatastoreHelper

### LINKED ISSUE #4933: LocalDatastoreHelper should accept a custom port to start the emulator.
Hello.

I'm using Spring libraries to operate with the GCP products and I found a issue with the datastore emulator helper.

Unfortunately the class `LocalDatastoreHelper` has a hardcoded way to find the port to start the emulator and doesn't accepts a custom port.

To ensure datastore testing with all spring libraries I will need to set the port based on my own configuration and this will also make easly the usage of the helper itself.

I was wondering about a new constructor adding the support for a custom port.

See spring discussion [here](https://github.com/spring-cloud/spring-cloud-gcp/issues/1430#issuecomment-481429999).

lucasoares:
Fixes #4933

I want the reviewer to validate if the create method with only the integer port can create any kind of misunderstanding since there was another create method with a single double value. If not, I will maintain it.

Thanks.


lucasoares:
I commited with my work email address. Now I have the CLA in the work email address too.

PR #4940, opened 2019-04-13
Spanner: Use gax for retries

olavloite:
Use GAX instead of the custom `SpannerImpl.runWithRetries` method for retrying server calls.

This migration involves three changes:

- Remove the override of the generated default settings in `GapicSpannerRpc` that set the retryable error codes to an empty set. This will automatically enable GAX retries for all server calls using the default values that have already been defined. The defaults are the same as the defaults that are used by `SpannerClient`.
- Add configuration options to `SpannerOptions` to allow the user to change the default `RetrySettings`. As `SpannerOptions` is used to create a `Spanner` instance, which again contains both a `DatabaseClient`, `DatabaseAdminClient` and `InstanceAdminClient`, `SpannerOptions` must contain configuration options for all three underlying clients.
- The `runWithRetries` method in `SpannerImpl` is removed and all references to it changed into direct calls that will automatically be retried by GAX.

To avoid a lot of unnecessary configuration options on `SpannerOptions`, the retry settings of the three underlying clients are exposed by exposing the `StubSettings.Builders` that are used to create the underlying clients. This only adds three extra methods to `SpannerOptions`, while still granting (advanced) users access to all configuration options.

The client library will use the defaults that are globally defined for retries in the configuration yaml files of Spanner (i.e. in the files spanner_gapic.yaml, spanner_admin_database_gapic.yaml and spanner_admin_instance_gapic.yaml)

The following defaults are defined:

- Spanner default settings:
  - Initial retry delay: 1 second
  - Retry multiplier: 1.3
  - Max retry delay: 32 seconds
  - Initial RPC timeout: 60 seconds
  - RPC timeout multiplier: 1.0
  - Max RPC timeout: 60 seconds
  - Total timeout: 600 seconds
- Spanner streaming settings (used for executeStreamingSql and streamingRead):
  - Initial retry delay: 1 second
  - Retry multiplier: 1.3
  - Max retry delay: 32 seconds
  - Initial RPC timeout: 120 seconds
  - RPC timeout multiplier: 1.0
  - Max RPC timeout: 120 seconds
  - Total timeout: 1200 seconds
- Spanner long-running settings (used for commit):
  - Initial retry delay: 1 second
  - Retry multiplier: 1.3
  - Max retry delay: 32 seconds
  - Initial RPC timeout: 3600 seconds
  - RPC timeout multiplier: 1.0
  - Max RPC timeout: 3600 seconds
  - Total timeout: 3600 seconds
- Database admin default settings (used for all methods except methods that return an operation, i.e. createDatabase and updateDatabaseDdl, these have a total timeout of 24 hours):
  - Initial retry delay: 1 second
  - Retry multiplier: 1.3
  - Max retry delay: 32 seconds
  - Initial RPC timeout: 60 seconds
  - RPC timeout multiplier: 1.0
  - Max RPC timeout: 60 seconds
  - Total timeout: 600 seconds
- Instance admin default settings (used for all methods except methods that return an operation, i.e. createInstance and updateInstance, these have a total timeout of 24 hours):
  - Initial retry delay: 1 second
  - Retry multiplier: 1.3
  - Max retry delay: 32 seconds
  - Initial RPC timeout: 60 seconds
  - RPC timeout multiplier: 1.0
  - Max RPC timeout: 60 seconds
  - Total timeout: 600 seconds


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -179,6 +206,99 @@ public Builder setSessionLabels(Map<String, String> sessionLabels) {
...
+     *     .applyToAllUnaryMethods(
+     *         new ApiFunction&lt;UnaryCallSettings.Builder&lt;?, ?&gt;, Void&gt;() {
+     *           public Void apply(Builder&lt;?, ?&gt; input) {
+     *             input.setRetrySettings(retrySettings);
+     *             return null;
+     *           }
+     *         });
+     * </code></pre>
+     */
+    public SpannerStubSettings.Builder spannerStubSettingsBuilder() {
```
kolea2: I think I would prefix 'new' to all of the builder names to be consistent with other areas in the library.

kolea2: ex. `newSpannerStubSettingsBuilder()`

kolea2: sorry - looking at this again and realize it's returning a variable. Should we just return a new builder here instead, or is there a reason why we are storing it?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -104,6 +122,12 @@ private SpannerOptions(Builder builder) {
     private int prefetchChunks = DEFAULT_PREFETCH_CHUNKS;
     private SessionPoolOptions sessionPoolOptions;
     private ImmutableMap<String, String> sessionLabels;
+    private SpannerStubSettings.Builder spannerStubSettingsBuilder =
```
kolea2: Can we remove the initialization here, as they are getting set to the options' builders in the below private constructor?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -179,6 +206,99 @@ public Builder setSessionLabels(Map<String, String> sessionLabels) {
...
+     *     .applyToAllUnaryMethods(
+     *         new ApiFunction&lt;UnaryCallSettings.Builder&lt;?, ?&gt;, Void&gt;() {
+     *           public Void apply(Builder&lt;?, ?&gt; input) {
+     *             input.setRetrySettings(retrySettings);
+     *             return null;
+     *           }
+     *         });
+     * </code></pre>
+     */
+    public InstanceAdminStubSettings.Builder instanceAdminStubSettingsBuilder() {
```
kolea2: Taking a step back, I find it confusing that there are getters on a builder (IMO). What do you think about adding setting APIs as part of the SpannerOptions Builder, and just setting defaults if they are not set by the user? CC @sduskis for your thoughts on how the API should look here.

olavloite: I based this idea on how the (generated) `SpannerSettings` for `SpannerClient` does this with 'recursive builders': https://github.com/googleapis/google-cloud-java/blob/d931fac64e331bbcffea7ff9dbd271e31bac5e05/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/v1/SpannerSettings.java#L78

We will need some sort of defaults, as gax requires it (even if the default is 'do not retry' or 'retry without ever giving up').

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
olavloite: No, it is needed, because the no-arg constructor on line 132 is called by `SpannerOptions.newBuilder()`, and does not contain any initialization code. The other fields that require initialization also do the initialization at  declaration instead of in the no-arg constructor. One option could be to move all these initializations from the declarations to the no-arg constructor, but my initial feeling is that it wouldn't make it clearer.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
kolea2: Ah I see - in that case I would either add 'get' or 'create' to these to be a little clearer

olavloite: Changed to get...SettingsBuilder(). 'create' would not be appropriate here, as the settings builders have already been created with their default values, and these methods get these builders.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -179,6 +206,99 @@ public Builder setSessionLabels(Map<String, String> sessionLabels) {
...
+      throw new UnsupportedOperationException(
+          "SpannerOptions does not support setting global retry settings. "
+              + "Call spannerStubSettingsBuilder().<method-name>Settings().setRetrySettings(RetrySettings) instead.");
+    }
+
+    /**
+     * Returns the {@link SpannerStubSettings.Builder} that will be used to build the {@link
+     * SpannerRpc}. Use this to set custom {@link RetrySettings} for individual gRPC methods.
+     *
+     * <p>The library will automatically use sensible defaults if no custom settings are set. The
```
kolea2: small comment - can we replace "sensible defaults" to a link to where the defaults are set?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -179,6 +206,99 @@ public Builder setSessionLabels(Map<String, String> sessionLabels) {
...
+     */
+    public SpannerStubSettings.Builder getSpannerStubSettingsBuilder() {
+      return spannerStubSettingsBuilder;
+    }
+
+    /**
+     * Returns the {@link InstanceAdminStubSettings.Builder} that will be used to build the {@link
+     * SpannerRpc}. Use this to set custom {@link RetrySettings} for individual gRPC methods.
+     *
+     * <p>The library will automatically use sensible defaults if no custom settings are set. The
```
kolea2: same here

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -179,6 +206,99 @@ public Builder setSessionLabels(Map<String, String> sessionLabels) {
...
+     */
+    public InstanceAdminStubSettings.Builder getInstanceAdminStubSettingsBuilder() {
+      return instanceAdminStubSettingsBuilder;
+    }
+
+    /**
+     * Returns the {@link DatabaseAdminStubSettings.Builder} that will be used to build the {@link
+     * SpannerRpc}. Use this to set custom {@link RetrySettings} for individual gRPC methods.
+     *
+     * <p>The library will automatically use sensible defaults if no custom settings are set. The
```
kolea2: same

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
olavloite: I've added a reference to the defaults for all three.

olavloite:
This PR will also solve the original issue mentioned in #3616, as it will enable users to configure timeouts and custom retry settings for a SpannerOptions instance.

PR #4948, opened 2019-04-15
More Pubsub cleanup

sduskis:
- Removing PublisherStats, since it's package private and not used anywhere.
- StatusUtil and SubscriberStats are package private.  Remove the `public` modifier on methods in those classes

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
@@ -376,7 +376,8 @@ class MessageDispatcher {
   /** Compute the ideal deadline, set subsequent modacks to this deadline, and return it. */
   @InternalApi
   int computeDeadlineSeconds() {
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/SubscriberStats.java
```diff
@@ -27,52 +27,52 @@
 
   @AutoValue
   // TODO: Finish implementation.
-  public abstract static class Stats {}
+  abstract static class Stats {}
```
kolea2: To confirm, everything in this class is only package private? Want to make sure we won't break any existing implementations.

sduskis: correct.  The top level class `SubscriberStats` is package private, so no one can access the `Stats` inner class from outside the package.

kolea2:
This LGTM, but I had a question regarding the MessageDispatcher change in https://github.com/googleapis/google-cloud-java/pull/4949. It looks like this change made it into both PRs, can you remove it from this once, I'll approve, and we can follow up on the other PR separately?

sduskis:
@kolea2, done.

PR #4949, opened 2019-04-15
Cleaning up deprecated pubsub code.

sduskis:


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/MessageDispatcher.java
```diff
@@ -380,8 +380,7 @@ public void run() {
   /** Compute the ideal deadline, set subsequent modacks to this deadline, and return it. */
   @InternalApi
   int computeDeadlineSeconds() {
-    long secLong = ackLatencyDistribution.getNthPercentile(PERCENTILE_FOR_ACK_DEADLINE_UPDATES);
-    int sec = Ints.saturatedCast(secLong);
+    int sec = ackLatencyDistribution.getPercentile(PERCENTILE_FOR_ACK_DEADLINE_UPDATES);
```
kolea2: shouldn't this still have the Ints.saturatedCast?

sduskis: `Distribution.getNthPercentile()` returns a `long`, but that `long` is just cast from an `int` (see [here](https://github.com/googleapis/gax-java/blob/8d45d186e36ae97b789a6f89d80ae5213a773b65/gax/src/main/java/com/google/api/gax/core/Distribution.java#L57))

`Ints.saturatedCast` isn't actually need if a `long` value can be outside of the MAX/MIN range of an `int`, which in this case it can't be due to the original cast from an `int` to a `long`.

PR #4952, opened 2019-04-15
Use splitindex for google-api-grpc javadoc.

kolea2:


kolea2:
@chingor13 PTAL

chingor13:
Should we do it for `google-cloud-clients` as well?

PR #4953, opened 2019-04-16
First generation of Data Catalog

sduskis:


google-cloud-bom/pom.xml
```diff
@@ -354,6 +354,21 @@
         <artifactId>google-cloud-core-http</artifactId>
         <version>1.69.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-core-http:current} -->
       </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-datacatalog</artifactId>
+        <version>0.87.1-alpha-SNAPSHOT</version><!-- {x-version-update:google-cloud-datacatalog:current} -->
```
kolea2: shouldn't the version be 0.0.1-alpha-SNAPSHOT?

google-cloud-clients/google-cloud-datacatalog/pom.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>google-cloud-datacatalog</artifactId>
+  <version>0.87.1-beta-SNAPSHOT</version><!-- {x-version-update:google-cloud-datacatalog:current} -->
```
kolea2: also, inconsistency between alpha and beta here

sduskis: I'll change to alpha.

google-cloud-bom/pom.xml
sduskis: We have been using a consistent version across all artifacts that are pre-0.  I'm not sure what we ought to do going forward.

PR #4955, opened 2019-04-16
Spanner: Aborted batch dml should be retried

olavloite:
When executing batchDml statements, it is possible that the transaction is aborted by Spanner for one of the statements. This used to throw a SpannerBatchUpdateException, which would cause the transaction to fail, while the appropriate exception is AbortedException to trigger a retry of the transaction.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionManagerTest.java
@@ -172,7 +172,7 @@ public class ITTransactionManagerTest {
           Thread.sleep(e.getRetryDelayInMillis() / 1000);
           // It is possible that it was txn2 that aborted.
           // In that case we should just retry without resetting anything.
-          if(manager1.getState() == TransactionState.ABORTED) {
+          if (manager1.getState() == TransactionState.ABORTED) {
             txn1 = manager1.resetForRetry();
           }
         }


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -270,7 +271,12 @@ public long executeUpdate(Statement statement) {
         results[i] = response.getResultSets(i).getStats().getRowCountExact();
       }
 
-      if (response.getStatus().getCode() != 0) {
+      // If one of the DML statements was aborted, we should throw an aborted exception.
```
kolea2: is it possible to also add a test to TransactionRunnerImplTest?

olavloite: Added.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/TransactionRunnerImplTest.java
```diff
@@ -141,6 +154,82 @@ public void runResourceExhaustedNoRetry() throws Exception {
...
+                public long[] run(TransactionContext transaction) throws Exception {
+                  numCalls.incrementAndGet();
+                  return transaction.batchUpdate(Arrays.asList(statement, statement));
+                }
+              });
+    } catch (SpannerBatchUpdateException e) {
+      exception = e;
+      updateCount = e.getUpdateCounts();
+    }
+    if (status == Code.ABORTED_VALUE) {
```
sduskis: Is there anyway to move the asserts into `batchDmlAborted` and `batchDmlFailedPrecondition`?

kolea2:
@olavloite LGTM after running code formatting

PR #4956, opened 2019-04-16
PubSub : Fix Publisher.shutdown should return promptly

### LINKED ISSUE #3687: Publisher.shutdown should return promptly
Current implementation waits for all messages to either succeed or fail before returning.

This will change the behavior of `shutdown`. So we should wait for `awaitTermination` to bake in for a while before doing this.

abhinav-qlogic:
Fixes #3687

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.pubsub.v1;
 
-import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
@@ -47,7 +46,6 @@ import com.google.pubsub.v1.TopicName;
 import com.google.pubsub.v1.TopicNames;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -430,7 +431,21 @@ public void shutdown() throws Exception {
...
    */
   public boolean awaitTermination(long duration, TimeUnit unit) throws InterruptedException {
-    return publisherStub.awaitTermination(duration, unit);
+    messagesWaiter.waitNoMessages();
+    boolean isAwaited = publisherStub.awaitTermination(duration, unit);
+    if (isAwaited) {
+      for (AutoCloseable closeable : closeables) {
+        ExecutorAsBackgroundResource executorAsBackgroundResource =
+            (ExecutorAsBackgroundResource) closeable;
+        isAwaited = executorAsBackgroundResource.awaitTermination(duration, unit);
```
sduskis: The user is asking for something like: wait for 8 seconds total for termination.

`publisherStub.awaitTermination(duration, unit);` could take 2.1 seconds.  The next await termination should wait 8-2.1 seconds.

Please keep track of the start time, and the number of millis that have gone by since the method was invoked, and then wait the remainder of the users requested timeout.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -408,17 +408,18 @@ private long getMaxBatchBytes() {
...
       throw new IllegalStateException("Cannot shut down a publisher already shut-down.");
     }
     if (currentAlarmFuture != null && activeAlarm.getAndSet(false)) {
       currentAlarmFuture.cancel(false);
     }
     publishAllOutstanding();
-    messagesWaiter.waitNoMessages();
     for (AutoCloseable closeable : closeables) {
-      closeable.close();
+      ExecutorAsBackgroundResource executorAsBackgroundResource =
```
sduskis: `closable.close()` calls `shutdown()` under the covers. There's no need to cast to a `ExecutorAsBackgroundResource`

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
abhinav-qlogic: @sduskis changes done

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
abhinav-qlogic: @sduskis changes done

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -430,7 +429,32 @@ public void shutdown() throws Exception {
    * <p>Call this method to make sure all resources are freed properly.
    */
   public boolean awaitTermination(long duration, TimeUnit unit) throws InterruptedException {
-    return publisherStub.awaitTermination(duration, unit);
+    long startDuration = System.currentTimeMillis();
+    long remainingDuration = TimeUnit.MILLISECONDS.convert(duration, unit);
+    messagesWaiter.waitNoMessages();
+    remainingDuration = getRemainingDuration(remainingDuration, startDuration);
+    startDuration = System.currentTimeMillis();
```
sduskis: Please don't update `startDuration`

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -430,7 +429,32 @@ public void shutdown() throws Exception {
    * <p>Call this method to make sure all resources are freed properly.
    */
   public boolean awaitTermination(long duration, TimeUnit unit) throws InterruptedException {
-    return publisherStub.awaitTermination(duration, unit);
+    long startDuration = System.currentTimeMillis();
```
sduskis: please make `startDuration` final.  There's never a need to update it, since startDuration represents the start time when the method was called.

abhinav-qlogic: @sduskis startDuration needs to be update because it is used(counting) to find elapsed time which occur to execute waitNoMessages method only, so we have to minus this time from actual time pass to  duration parameter of awaitTermination method and so on needs to update every time to find elapsed time so we can remove it from total remaining time.

sduskis: I think I understand what you mean.  Let's have a static `startTime` and `totalDurationMs` variables that are final.  We can use those plus current time to calculate `remainingDurationMs`

abhinav-qlogic: @sduskis chnages done

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -430,7 +429,37 @@ public void shutdown() throws Exception {
    * <p>Call this method to make sure all resources are freed properly.
    */
   public boolean awaitTermination(long duration, TimeUnit unit) throws InterruptedException {
-    return publisherStub.awaitTermination(duration, unit);
+    final long startDuration = System.currentTimeMillis();
```
sduskis: Please call this variable `startTimeMs`

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -430,7 +429,37 @@ public void shutdown() throws Exception {
...
+    boolean isAwaited =
+        remainingDuration < totalDurationMs
+            ? publisherStub.awaitTermination(remainingDuration, TimeUnit.MILLISECONDS)
+            : false;
+    if (isAwaited) {
+      for (AutoCloseable closeable : closeables) {
+        ExecutorAsBackgroundResource executorAsBackgroundResource =
+            (ExecutorAsBackgroundResource) closeable;
+        remainingDuration = getRemainingDuration(startDuration, totalDurationMs);
+        System.out.println(remainingDuration);
```
sduskis: please remove this println.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -430,7 +429,37 @@ public void shutdown() throws Exception {
...
+        isAwaited =
+            remainingDuration < totalDurationMs
+                ? executorAsBackgroundResource.awaitTermination(
+                    getRemainingDuration(startDuration, totalDurationMs), TimeUnit.MILLISECONDS)
+                : false;
+        if (!isAwaited) {
+          return false;
+        }
+      }
+    } else {
```
sduskis: please remove the else.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -430,7 +429,37 @@ public void shutdown() throws Exception {
...
+      return false;
+    }
+    return true;
```
sduskis: please use `return isAwaited`

abhinav-qlogic: @sduskis changes done

ajaaym:
@abhinav-qlogic closing this PR due to lot of unintended merge/conflict. Please create new with just your change.

PR #4970, opened 2019-04-17
Prep version numbers for Speech GA.

### LINKED ISSUE #4915: google-cloud-speech should be promoted to GA
google-cloud-speech is Beta, and has been for a while, so it should be promoted to GA. It was Beta because the surfaces for LRO and Bidi streaming weresn't stable yet, but those are now stable.

kolea2:
Fixes #4915


kolea2:
@crwilcox FYI

PR #4972, opened 2019-04-17
More refactoring of SequentialExecutor

sduskis:
@kimkyung-goog, FYI


sduskis:
@olavloite, I don't think that these changes made that happen.  I can't reproduce the flakiness.

This PR is a series of relatively small refactorings, which shouldn't lead to additional instability.  I'll close this PR, and raise other small PRs.

PR #4979, opened 2019-04-18
Refactoring SequentialExecutorService.java

sduskis:
- Marking methods as private
- renaming variables from `finalTasks` to `tasks` for clarity
- reducing code duplication by calling `execute`


sduskis:
@olavloite, we need to do more refactoring to make a clear distinction between publicly available methods and implementation details.  Thanks for the review.

PR #4983, opened 2019-04-18
Exposing AutoExecutor and CallbackExecutor directly

sduskis:
@kimkyung-goog, I did a bunch of refactoring which was reviewed by a contractor.  I basically finished splitting up the `SequentialExecutor` into `CallbackExecutor` which is used by the `Publisher` and an `AutoExecutor` that's used by `MessageDispatcher`.

I think that there's more refactoring I wanted to do, but I think that we're at a good stopping point for review.  PTAL at your convenience.


kimkyung-goog:
Thank you for doing the refactoring. I see that they are much readable now. LGTM.

PR #4990, opened 2019-04-22
Bigtable: Added unit tests for `GCRule#fromProto`

### LINKED ISSUE #4057: Bigtable: investigate GCRule.fromProto being incorrect
From a user:

> GCRule and GcRule classes are confusing and .fromProto() conversion does not work properly in some cases. Not 100% sure which scenario it was, but I think we were setting maxversions and maxage, and only one of them was in the final result after calling .fromProto().

rahulKQL:
Fixes #4057 

### Observation:
If used multiple `setXXX()` on protobuf `GcRule#Builder` then the last most added value would be persisted after the `build()`. This is also valid for model `GCRule#fromProto`.

Added few unit test cases around the `fromProto` operation.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/GCRulesTest.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/GCRulesTest.java
@@ -281,19 +281,8 @@ public class GCRulesTest {
   }
 
   @Test
-  public void multipleSetters() {
-    com.google.protobuf.Duration seconds =
-        com.google.protobuf.Duration.newBuilder().setSeconds(1).build();
-    // Only the last setXXX() value persists after build().
-    GcRule proto_gcRule = GcRule.newBuilder().setMaxNumVersions(5).setMaxAge(seconds).build();
-
-    GCRules.GCRule model_gcRule = GCRULES.maxAge(Duration.ofSeconds(1));
-    assertThat(GCRULES.fromProto(proto_gcRule)).isEqualTo(model_gcRule);
-  }
-
-  @Test
   public void unionOfRules() {
-    GcRule proto_gcRule =
+    GcRule protoGCRule =
         GcRule.newBuilder()
             .setUnion(
                 Union.newBuilder()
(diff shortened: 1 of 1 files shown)


rahulKQL:
@sduskis / @igorbernstein2 
Please have a look at the observations, If valid then we can close the mentioned issue.

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/GCRulesTest.java
```diff
@@ -280,6 +280,97 @@ public void intersectionOfUnions() {
...
+  @Test
+  public void multipleSetters() {
+    com.google.protobuf.Duration seconds =
+        com.google.protobuf.Duration.newBuilder().setSeconds(1).build();
+    // Only the last setXXX() value persists after build().
+    GcRule proto_gcRule = GcRule.newBuilder().setMaxNumVersions(5).setMaxAge(seconds).build();
+
+    GCRules.GCRule model_gcRule = GCRULES.maxAge(Duration.ofSeconds(1));
+    assertThat(GCRULES.fromProto(proto_gcRule)).isEqualTo(model_gcRule);
+  }
```
igorbernstein2: I'm having a hard time understanding what this is testing.  The protobuf behavior for one-of fields?

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/GCRulesTest.java
```diff
@@ -280,6 +280,97 @@ public void intersectionOfUnions() {
...
+
+  @Test
+  public void unionOfRules() {
+    GcRule proto_gcRule =
```
igorbernstein2: codestyle is a bit weird with underscores.

igorbernstein2:
The proto uses a one-of for max age & max versions, so you can't set both. This confuses a lot of people, which is one of the reasons we created this wrapper

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/GCRulesTest.java
rahulKQL: Tests that the multiple setters override existing value... It was to confirm my understanding of protobuf behavior. I think this could be removed.

igorbernstein2: Yea, I don't think testing underlying protobuf behavior should happen in this repo

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/GCRulesTest.java
rahulKQL: Updated code style to camel case. Please have a look.

PR #4992, opened 2019-04-22
SequentialExecutorService.callNextTaskAsync only uses key

sduskis:
1. Removing the `Deque` parameter from `SequentialExecutorService.callNextTaskAsync()` and looking it up the `Deque` in the method
2. Simplify `AutoExecutor` and `CallbackExecutor` after the refactoring in 1).


sduskis:
FYI, @kimkyung-goog

PR #4993, opened 2019-04-22
Using generics for Runnable type in SequentialExecutorService.

sduskis:
This reduces complexity in `CallbackExecutor`, since we can be assured that all entries in the Queue are `CancellableRunnable`s.


sduskis:
FYI, @kimkyung-goog

PR #4994, opened 2019-04-22
Using a Queue instead of Deque in P/S OK

sduskis:
The executor needs to have a FIFO ordering (first in, first out).  Deques allow for pulling entries from either the head or tail, which is not something we want to support in the executor.


sduskis:
FYI, @kimkyung-goog

PR #4995, opened 2019-04-22
Fix javadoc linking

chingor13:
Fixes maven configs to correctly link javadoc from google-cloud-clients to google-api-grpc classes.


kolea2:
Should we add something to new_client.py for when we generate to add this to the pom.xmls?

chingor13:
> Should we add something to new_client.py for when we generate to add this to the pom.xmls?

Yeah, I'll add it to the template.

PR #5003, opened 2019-04-23
Update maven-checkstyle-plugin and puppycrawl checkstyle

chingor13:
- Update dependencies for: [CVE-2019-9658](https://nvd.nist.gov/vuln/detail/CVE-2019-9658)
- Only run the checkstyle plugin on Java 8+ as the fixed puppycrawl checkstyle artifact only works on Java 8+.


chingor13:
hmm, looks like the new puppycrawl version is not compatible with Java 7

chingor13:
cc @JustinBeckwith

PR #5018, opened 2019-04-25
P/S refactoring Publisher.publish

sduskis:
`publish()` has a special case where the message is larger the the max byte count.  That special case logic can be merged with the general scenario by using a list of `batchesToSend` rather than having a single one.


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -197,7 +197,7 @@ public String getTopicNameString() {
     }
 
     message = messageTransform.apply(message);
-    OutstandingBatch batchToSend = null;
+    List<OutstandingBatch> batchesToSend = new ArrayList<>();
```
kolea2: IIRC, this used to be a LinkedList. Is changing to an ArrayList ok?

sduskis: This was never a `LinkedList`, AFAIK.  Historically, there was only `OutstandingBatch batchToSend = null`

kolea2: sorry - was confusing it was MessagesBatch.

PR #5019, opened 2019-04-25
Split javadocs

chingor13:
This PR splits the javadoc generation into per-artifact minisites. The publish_javadoc kokoro job will now publish each artifact's javadocs separately (and fix linking between all the javadoc sites).

This is not compatible with our current format of aggregating the javadocs into google-api-grpc and google-cloud-clients javadocs. The resulting docs should be better targeted and more usable.


sduskis:
@chingor13, this PR is a bit stale.

chingor13:
I think reorganizing the clients will supersede this PR.

PR #5026, opened 2019-04-26
Add Phishing Protection API Client

chingor13:

PR #5027, opened 2019-04-26
Add reCAPTCHA Enterprise client

chingor13:


google-cloud-clients/google-cloud-recaptchaenterprise/synth.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2018 Google LLC
```
kolea2: 2019

chingor13: Fixed here. Template fixed in #5026

PR #5028, opened 2019-04-26
Update new client script

### LINKED ISSUE #4958: Update new_client.py to add new clients to the list of modules in google-cloud-clients
New clients should be added to the list of modules in the pom.xml file.

chingor13:
- `-c` option for the path to the artman config is now optional
- add `-n` option for the pretty name used in the README and pom.xml
- add `-d` option for the description used in the README and pom.xml
- add `--docs-url` and `--overview-url` used in the README
- fix detecting the parent artifact version when generating the pom.xml files
- removes legacy scripts which are no longer used for generation/release
- new client versions are no longer versioned with "alpha"

Fixes #4958

### DIFF
--- a/utilities/new_client.py
+++ b/utilities/new_client.py
@@ -33,6 +33,8 @@ class Context:
     artman_config: str = None
     google_cloud_artifact: str = None
     google_cloud_version: releasetool.Version = None
+    google_cloud_parent_version: releasetool.Version = None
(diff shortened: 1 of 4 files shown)


kolea2:
This will close https://github.com/googleapis/google-cloud-java/issues/4958 as well yes?

chingor13:
Yes this will fix #4958

PR #5029, opened 2019-04-26
Allow colons in Identity values

JesseLovelace:
This is to address internal bug 118884564. Identity is written for colons not to be allowed in the value, however colons are actually allowed in some legacy cases, like domain-scoped project IDs.


google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/IdentityTest.java
```diff
@@ -141,11 +141,6 @@ public void testValueOfEmpty() {
     Identity.valueOf("");
   }
 
-  @Test(expected = IllegalArgumentException.class)
-  public void testValueOfThreePart() {
-    Identity.valueOf("a:b:c");
```
chingor13: Can you change the test to make sure it takes "a:b" as the identity?

JesseLovelace: > Should we still throw an IllegalArgumentException if there's more parts than we expect?

The output of the split can now no longer be longer than 2, so that's impossible. Now the exception is only thrown if the output is of size zero, which is still correct

JesseLovelace: > Can you change the test to make sure it takes "a:b" as the identity?

For the input string "a:b:c," this test doesn't cause "a" to be the type and "b" to be the value, it causes "a" to be the type and "b:c" to be the value. If you think it's necessary I can add a test for that

chingor13: Let's still have a test to make sure it's split the way we expect

google-cloud-clients/google-cloud-core/src/test/java/com/google/cloud/IdentityTest.java
```diff
@@ -141,16 +141,16 @@ public void testValueOfEmpty() {
...
-  }
-
   @Test
   public void testUnrecognizedToString() {
     assertEquals("a:b", Identity.valueOf("a:b").strValue());
   }
 
+  @Test
+  public void testValueOfThreePart() {
+    assertEquals("a:b:c", Identity.valueOf("a:b:c").strValue());
```
chingor13: I meant to make sure that the the type and value are split the way we expect - type is `a` and value is `b:c`

PR #5034, opened 2019-04-29
Bigtable: clean up the public surface

igorbernstein2:
* Exclude autogenerated files that aren't used
* ~~Internal implementation details are now package private~~
* Internal implementation details are now @InternalApi
* package-info.java documents the public surface instead of the autogen one
* EnhancedBigtableStubSetting is now public api
* synth.py is responsible for the transformations

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BaseBigtableInstanceAdminClient.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BaseBigtableInstanceAdminClient.java
@@ -73,9 +73,11 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Generated;
 
 // AUTO-GENERATED DOCUMENTATION AND SERVICE
 /** For internal use only. */
+@Generated("by gapic-generator")
 @InternalApi
 public class BaseBigtableInstanceAdminClient implements BackgroundResource {
   private final BaseBigtableInstanceAdminSettings settings;
--- a/google-cloud-clients/google-cloud-bigtable/synth.py
+++ b/google-cloud-clients/google-cloud-bigtable/synth.py
@@ -119,7 +119,7 @@ def make_internal_only(sources):
   s.replace(sources,
 
             before=r'/\*\*.+?\*/\n(?:^@[^\n]*\n)*(?=public [a-zA-B ]*class)',
-            after='/** For internal use only. */\n@InternalApi\n',
+            after='/** For internal use only. */\n@Generated("by gapic-generator")\n@InternalApi\n',
             flags=re.MULTILINE | re.DOTALL)
 
 main()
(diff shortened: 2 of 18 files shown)


kolea2:
Can/should this be done on the proto level instead?

igorbernstein:
I don’t know of a way. Do you have a recommendation?

igorbernstein2:
Updated PR to only mark things as `InternalApi` instead of marking things as package private, which broke tests using Mockito to mock the stubs

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BaseBigtableInstanceAdminClient.java
```diff
@@ -72,79 +73,10 @@
...
- * <pre>
- * <code>
- * BaseBigtableInstanceAdminSettings baseBigtableInstanceAdminSettings =
- *     BaseBigtableInstanceAdminSettings.newBuilder().setEndpoint(myEndpoint).build();
- * BaseBigtableInstanceAdminClient baseBigtableInstanceAdminClient =
- *     BaseBigtableInstanceAdminClient.create(baseBigtableInstanceAdminSettings);
- * </code>
- * </pre>
- */
-@Generated("by gapic-generator")
```
chingor13: Should we preserve this annotation so we mark the class as generated.

igorbernstein2: I was debating this myself. As far as I know nothing uses that annotation and I decided to keep it simple until its needed. If you feel strongly, I can try to retain it

igorbernstein2:
As discussed offline:
- created a gapic-generator feature request: https://github.com/googleapis/gapic-generator/issues/2742
- inserted Autogenerated annotations
- added unit tests to make sure that the InternalApi annotations stick

google-cloud-clients/google-cloud-bigtable/synth.py
```diff
@@ -14,39 +14,112 @@
...
+  s.replace(sources,
+            before=r'^(import (?!static).*?\n)\n',
+            after=r'\1import com.google.api.core.InternalApi;\n\n',
+            flags=re.MULTILINE | re.DOTALL)
 
-s.copy(admin_library / 'gapic-google-cloud-bigtable-admin-v2/src', 'src')
-s.copy(admin_library / 'grpc-google-cloud-bigtable-admin-v2/src', '../../google-api-grpc/grpc-google-cloud-bigtable-admin-v2/src')
-s.copy(admin_library / 'proto-google-cloud-bigtable-admin-v2/src', '../../google-api-grpc/proto-google-cloud-bigtable-admin-v2/src')
+  # Replace javadoc and annotations before every public class with InternalApi
+  # javadoc and annotation
```
chingor13: Can you add the link to the feature request issue for the gapic generator?

igorbernstein2: It's at the top of the file

PR #5042, opened 2019-05-01
Migrate from containsAllIn to containsAtLeastElementsIn.

cpovirk:
The two behave identically, and containsAllIn is being removed.

The new method was added in Truth 0.44, so update to that.


cpovirk:
Thanks. I'm hoping that it's fixed now -- `mvn install -DskipITs` passes -- but I notice that Kokoro is still running 16 hours after my last push.

sduskis:
@cpovirk, kokoro kicks in either with a google-cloud-java team member's PR or an explicit label on the PR.  I'll kick off the build.

cpovirk:
Ah, thanks! Looks like Kokoro is happy now. Code coverage is allegedly down slightly, but that seems unlikely to mean anything real here.

PR #5044, opened 2019-05-01
Pub/Sub: more refactoring

sduskis:
- Using `Preconditions`
- Adding `onSuccess` and `onFailure` to `OutstandingPublish`


chingor13:
Looks like you need to run the code formatter again.

PR #5055, opened 2019-05-02
Pubsub ordering keys

sduskis:
Pubsub has a new upcoming feature which will allow users to send and receive messages in a guaranteed order.

@kamalaboulhosn and @kimkyung-goog, I believe that all of the features are ready to go.  Can you please review this PR?  Also, guidance would be appreciated on whether or not you feel like this can be merged to master.


kamalaboulhosn:
@kimkyung-goog can you just verify that setting ordering keys in messages results in the messages being rejected by the Pub/Sub server for now? If that is true, then it is is safe to merge to master.

kimkyung-goog:
> @kimkyung-goog can you just verify that setting ordering keys in messages results in the messages being rejected by the Pub/Sub server for now? If that is true, then it is is safe to merge to master.

Answered by email (since it contains some internal links)

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -209,6 +240,10 @@ public String getTopicNameString() {
     messagesWaiter.incrementPendingMessages(1);
 
     if (!batchesToSend.isEmpty()) {
+      // TODO: if this is not an ordering keys scenario, will this do anything?
+      publishAllWithoutInflight();
+
+      // TODO: if this is an ordering keys scenario, is this safe without messagesBatchLock?
       for (final OutstandingBatch batch : batchesToSend) {
         logger.log(Level.FINER, "Scheduling a batch for immediate sending.");
         executor.execute(
```
kimkyung-goog: It seems that this execute() call was removed in the original implementation of ordering keys, but was added back when master is merged to this branch. Please see https://github.com/googleapis/google-cloud-java/commit/3bdbf7701bdb29a3a70a667d2a624d7fae0e5442#diff-c482c19af8c46aea82546fa17236dfd2

publishOutstandingBatch() should not be called by the executor here because it can reorder sequence of invoking the function. Inside publishOutstandingBatch(), we run an executor, a regular one for non-ordered messages and the sequential executor for ordered messages.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -194,13 +211,27 @@ public String getTopicNameString() {
...
+    Preconditions.checkState(
+        orderingKey.isEmpty() || enableMessageOrdering,
+        "Cannot publish a message with an ordering key when message ordering is not enabled.");
+
     final OutstandingPublish outstandingPublish =
         new OutstandingPublish(messageTransform.apply(message));
     List<OutstandingBatch> batchesToSend;
     messagesBatchLock.lock();
     try {
+      // Check if the next message makes the current batch exceed the max batch byte size.
```
kimkyung-goog: I think this comment can be removed now.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -209,6 +240,10 @@ public String getTopicNameString() {
     messagesWaiter.incrementPendingMessages(1);
 
     if (!batchesToSend.isEmpty()) {
+      // TODO: if this is not an ordering keys scenario, will this do anything?
+      publishAllWithoutInflight();
+
+      // TODO: if this is an ordering keys scenario, is this safe without messagesBatchLock?
```
kimkyung-goog: > is this safe without messagesBatchLock?
Since we are assuming publish() for the same ordering key is not called by multiple threads, I think it would be safe without the lock. If users use multiple threads for publishing an ordering key, we cannot guarantee the order anyway. However, it is always possible that I am missing something. WDYT? (But, the executor below should still be removed. Please see my another comment below)

kamalaboulhosn: I don't think we are assuming this. People could call publish for the same ordering key from multiple threads. It just means that the order among those two messages can not be known by the publisher. It will still be established (and will in fact be established in the order in which they are sent from the publisher client to Cloud Pub/Sub.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -257,16 +306,51 @@ public void run() {
...
+    try {
+      Iterator<Map.Entry<String, MessagesBatch>> it = messagesBatches.entrySet().iterator();
+      while (it.hasNext()) {
+        Map.Entry<String, MessagesBatch> entry = it.next();
+        MessagesBatch batch = entry.getValue();
+        String key = entry.getKey();
+        if (batch.isEmpty()) {
+          it.remove();
+        } else if (key.isEmpty() || !sequentialExecutor.hasTasksInflight(key)) {
+          // TODO(kimkyung-goog): Do not release `messagesBatchLock` when publishing a batch. If
```
kimkyung-goog: Sorry, I think this comment is confusing. Can you remove it? I think you already addressed the potential performance issue in the TODO comment below.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -257,16 +306,51 @@ public void run() {
...
+          // it's released, the order of publishing cannot be guaranteed if `publish()` is called
+          // while this function is running. This locking mechanism needs to be improved if it
+          // causes any performance degradation.
+
+          // TODO: Will this cause a performance problem for non-ordering keys scenarios?
```
kimkyung-goog: Ah, your comment is for the non-ordering scenarios. Can you also add the potential performance issue related to locking mechanism here? We grab a lock for the entire batch before the for loop, which means no messages can be added when this function is running. One of the typical solutions might be getting a reader lock for the message batches before the for loop and getting writer lock for each batch before calling publishOutstandingBatch(batch.popOutstandingBatch()). 
I am not sure this will make it better or even worse. It will probably depends on how many ordering keys are involved. Any ideas?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
kimkyung-goog: Right, users can definitely use threads for the same ordering key if they want, and we should support it. I think what I wanted to say was that if a user calls publish() for the same ordering key concurrently using threads without a proper synchronization, then there is no ordering in that behavior anyway.
I actually looked into the code again, and I think I was wrong. We should get messagesBatchLock. Here is the scenario:
(1) By messagesBatch.add(outstandingPublish) at line #231, `outstandingPublish` is added to `messages`, and an old batch is returned.
(2) Before the returned batch is published, publishAllWithoutInflight() is called by the timer.
(3) publishAllWithoutInflight() publishes the message added to the current batch at (1). Now, the new message is published before the old batch is published.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
sduskis: Woops.  I'll change it back.  What do you think about moving the `publish*` methods inside the lock?

PR #5060, opened 2019-05-04
Spanner: fix possible thread leak in Spanner.close()

### LINKED ISSUE #5059: Spanner: Spanner.close() does not clean up all resources
Opening and closing several Spanner instances during the lifetime of an application can cause a thread leak, as the method Spanner.close() does not clean up all threads that are created.

A Spanner instance contains three underlying gRPC stubs: `SpannerStub`, `InstanceAdminStub` and `DatabaseAdminStub`. In order to prevent these from using the same thread pool for gRPC calls, the stubs are handed a custom `ExecutorProvider`. This happens here: [https://github.com/googleapis/google-cloud-java/blob/cc04990d8e57f5804f6963b2a3ce8c1e2e18ab18/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L193](https://github.com/googleapis/google-cloud-java/blob/cc04990d8e57f5804f6963b2a3ce8c1e2e18ab18/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L193)

The `InstantiatingGrpcChannelProvider` will create the underlying transport channels and assign these executors that are gotten from the above `ExecutorProvider` here: [https://github.com/googleapis/gax-java/blob/13564bbe38e5496821e4686ab629181c5bb4ac66/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L178](https://github.com/googleapis/gax-java/blob/13564bbe38e5496821e4686ab629181c5bb4ac66/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L178)

This executor is then passed on to a `ManagedChannelBuilder` (or actually a `NettyChannelBuilder` which extends `AbstractManagedChannelImplBuilder`). This latter builder wraps the executor in a `FixedObjectPool` which always returns this executor when one is requested from the 'pool', and does nothing when it is returned to the pool. This means that the underlying executor is never shutdown when the transport channel is shutdown.

Removing the `setExecutorProvider` call on line 193 of `GapicSpannerRpc` solves the resource leak, but causes the underlying gRPC stubs to use a shared executor for their transport channels, possibly causing race conditions.

olavloite:
A Spanner instance holds a reference to three underlying gRPC stubs: spannerStub, instanceAdminStub and databaseAdminStub. These should all have their own thread pool for gRPC calls. This was achieved by setting an executor provider on the channel provider for the stubs. The worker threads of the thread pools created by this executor provider were however not shutdown when the stubs were closed, causing thread leaks if an application would open and close multiple Spanner instances during its lifetime.

Fixes #5059

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -114,8 +114,6 @@ import org.threeten.bp.Duration;
 
 /** Implementation of Cloud Spanner remote calls using Gapic libraries. */
 public class GapicSpannerRpc implements SpannerRpc {
-  // Thread factory to use to create our worker threads
-
   private static final PathTemplate PROJECT_NAME_TEMPLATE =
       PathTemplate.create("projects/{project}");
   private static final PathTemplate OPERATION_NAME_TEMPLATE =
@@ -581,6 +579,7 @@ public class GapicSpannerRpc implements SpannerRpc {
     return context.withStreamWaitTimeout(waitTimeout).withStreamIdleTimeout(idleTimeout);
   }
 
+  @Override
   public void shutdown() {
     this.spannerStub.close();
     this.instanceAdminStub.close();


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -115,6 +114,7 @@
 
 /** Implementation of Cloud Spanner remote calls using Gapic libraries. */
 public class GapicSpannerRpc implements SpannerRpc {
+  // Thread factory to use to create our worker threads
```
kolea2: minor - think you can remove this comment

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -190,7 +190,6 @@ public GapicSpannerRpc(SpannerOptions options) {
                         options.getInterceptorProvider(),
                         SpannerInterceptorProvider.createDefault()))
                 .setHeaderProvider(mergedHeaderProvider)
-                .setExecutorProvider(InstantiatingExecutorProvider.newBuilder().build())
```
kolea2: Since each underlying grpc stub gets set this executor provider by default, is it correct that each stub will not use a shared executor for its transport channel (the one potential problem listed in the issue with this change)?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
olavloite: You're right, this is no longer valid.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
olavloite: Correct.

PR #5071, opened 2019-05-06
Fix broken javadoc link for Bigtable.

kolea2:


igorbernstein2:
I think we should add package-info in com.google.cloud.bigtable instead of updating docs to data client

kolea2:
Hmm, weirdly there is a package-info.java file there, but no corresponding generated page. I'll keep looking into this.

igorbernstein2:
I created #5177 to track the missing javadocs

PR #5120, opened 2019-05-08
Storage : Fix blob update metadata.

### LINKED ISSUE #4942: Storage Client doesn't offer method to update blob metadata
It's possible to update object metadata (like a custom fields) using `gsutil` or `REST API`. However, it's not possible directly with the Java client.

suraj-qlogic:
Fixes #4942


sduskis:
@JesseLovelace, can you please review this?

JesseLovelace:
I discussed with Frank and we've decided not to merge this in.

The fact that setting metadata does an upsert augmentation of the map instead of a full replace is already an exceptional case when compared to how other "set" operations work in storage. We've decided to update the documentation to explain this, and add new samples.

Adding this additional method would mean adding even more documentation to explain it and samples around an already exceptional case and that would cause unnecessary clutter and confusion. We can readress this if we get more feedback on it from customers but I'm closing this for now.

PR #5125, opened 2019-05-08
[Bigtable] Added StreamController cancel in readRowsAsync docs

### LINKED ISSUE #4229: bigtable: add docs demonstrating stream cancellation
docs for readRowsAsync should give an example of how to cancel the stream using the StreamController

rahulKQL:
Fixes #4229

This change contains a simple example of how to cancel a stream received from `BigtableDataClient#readRowsAsync`.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -513,9 +513,16 @@ public class BigtableDataClient implements AutoCloseable {
    *          .range("[START KEY]", "[END KEY]")
    *          .filter(FILTERS.qualifier().regex("[COLUMN PREFIX].*"));
    *
-   *   // Iterator style
    *   try {
-   *     for(Row row : bigtableDataClient.readRows(query)) {
+   *     ServerStream<Row> stream = bigtableDataClient.readRows(query);
+   *     int count = 0;
+   *
+   *     // Iterator style
+   *     for (Row row : stream) {
+   *       if (++count > 10) {
+   *         stream.cancel();
+   *         break;
+   *       }
    *       // Do something with row
    *     }
    *   } catch (NotFoundException e) {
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -540,29 +540,45 @@ public Row readRow(String tableId, ByteString rowKey, @Nullable Filter filter) {
    * <p>Sample code:
    *
    * <pre>{@code
+   * class ClientObserver implements ResponseObserver<Row> {
```
igorbernstein2: Please keep this as an anonymous inner class, something like:

```java

   *   bigtableDataClient.readRowsAsync(query, new ResponseObserver<Row>() {
   *     StreamController controller;
   *     int count = 0;
   *
   *     public void onStart(StreamController controller) {
   *         this.controller = controller;
   *     }
   *     public void onResponse(Row row) {
   *        if (++count > 10) {
  *            controller.cancel();
   *           return;
  *        }
   *       // Do something with Row
   *     }
   *     public void onError(Throwable t) {
   *       if (t instanceof NotFoundException) {
   *         System.out.println("Tried to read a non-existent table");
   *       } else {
   *         t.printStackTrace();
   *       }
   *     }
   *     public void onComplete() {
   *       // Handle stream completion
   *     }
   *   });

```

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -540,29 +540,45 @@ public Row readRow(String tableId, ByteString rowKey, @Nullable Filter filter) {
    * <p>Sample code:
```
igorbernstein2: Please add an example above as well:

```java
   ServerStream<Row> stream = ...
   int count = 0;

  for(Row row : ...) {
     if (++count > 10) {
       stream.cancel();
       break;
    }
   // Do something with row
...
  }

```

rahulKQL:
@igorbernstein2 PTAL, Also please let me know if this sample code needs to be more concrete/realistic.

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
rahulKQL: Have updated the sample code to include the described approach. Please have a fresh look

rahulKQL:
@igorbernstein2
I have addressed both feedback comments. PTAL and merge if it looks good to you.

PR #5128, opened 2019-05-08
Update apiary version dependencies

chingor13:


chingor13:
cc @shollyman

google-cloud-clients/google-cloud-storage/pom.xml
```diff
@@ -32,9 +32,9 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>com.google.http-client</groupId>
-      <artifactId>google-http-client-apache</artifactId>
-      <version>2.1.0</version>
+      <groupId>org.apache.httpcomponents</groupId>
```
chingor13: The dependency in this package is actually on a constant from the apache httpclient library which we were pulling in as a transitive dependency here. We should instead be explicit in the dependency.

PR #5167, opened 2019-05-09
Spanner use custom thread factory

olavloite:
Use a custom ThreadFactory for the underlying gRPC clients to make it easier to determine and debug which service has created the threads.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpcTest.java
@@ -169,8 +169,15 @@ public class GapicSpannerRpcTest {
       }
       // Now close the Spanner instance and check whether the threads are shutdown or not.
       spanner.close();
-      // Wait a little to allow the threads to actually shutdown.
-      Thread.sleep(500L);
+      // Wait for up to two seconds to allow the threads to actually shutdown.
+      int totalWaits = 0;
+      while (true) {
+        Thread.sleep(100L);
+        if (getNumberOfThreadsWithName(SPANNER_THREAD_NAME) == 0 || totalWaits > 20) {
+          break;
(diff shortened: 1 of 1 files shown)


kolea2:
This looks good, do you think we can add an IT test as well for this?

Separately, anything we can test on the transport channel and its threads?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -102,18 +103,57 @@
...
+    private final List<ScheduledExecutorService> executors = new LinkedList<>();
+    private final ThreadFactory threadFactory;
+
+    private ManagedInstantiatingExecutorProvider(ThreadFactory threadFactory) {
+      this.threadFactory = threadFactory;
+    }
+
+    @Override
+    public boolean shouldAutoClose() {
+      return false;
```
kolea2: Quick question - should this be configurable (the InstantiatingExecutorProvider returns true)? Or no, since we are now managing the closing ourselves? If this is true, it gets added to background resources in gax: https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L136. But not sure if this is necessary since we are now handling all shutdowns.

olavloite: It should not be configurable, but return false to ensure that we handle the shutdown ourselves. As you mentioned, setting this to true will cause the executors to be added to the background resources in gax, but will (currently?) not be shutdown by gax. To prevent unexpected regression if gax behavior should change in the future, I would say we should keep this returning false.

kolea2: Nevermind, sorry - this is only for the transport provider I realized :)

PR #5172, opened 2019-05-10
Migrate from isSameAs to isSameInstanceAs (and isNotSameAs to isNotSameInstanceAs).

cpovirk:
The new methods behave identically to the old, which are being removed.

The new methods were added to Truth in version 0.44.


cpovirk:
OS X failures are probably unrelated.

PR #5178, opened 2019-05-13
Bigtable: Update javadoc

cshaff0524:
Text-only edits to comments and error messages.


cshaff0524:
I added a new commit after building this locally.

PR #5185, opened 2019-05-14
Fix FakeStorageRpc that doesn't set the blob size when returning blobs from the list method

maxtomassi:
`FakeStorageRpc` implements a list method that returns a list of the blobs that satisfy some given conditions. The current implementation doesn't set the size of the blobs that are returned. This would cause a problem if the client calling that method expects the size to be set in the blobs (pretty likely within a unit test). This PR fixes that issue by setting the size of the blobs being returned.


sduskis:
@JesseLovelace can you please review this PR?

PR #5188, opened 2019-05-15
Spanner: PartitionedDML transaction does not need a prepared session

olavloite:
Partitioned DML requests do not need to get a write-prepared session, as the request will start a new transaction anyway. Taking a read session from the pool is more efficient than 'wasting' a prepared session.


kolea2:
I think this will have merge conflicts once you merge in the gax retries PR, so holding off on approving this until that one is in.

PR #5194, opened 2019-05-17
Core: Default executor used non-daemon threads

### LINKED ISSUE #2315: JVM Hangs on gRPC if you don't close Spanner Object
I ran across a path with the Java Spanner client that if you create several Spanner instances and do not explicitly close them the JVM will hang (it will not terminate it's process after code execution).

To reproduce the behavior create an object that wraps Spanner creation. Create the object several times in a loop. Then execute some additional code.  

You can see the hanging wait on KQueueArrayWrapper.

Thanks.

screen shot 2017-08-07 at 9 05 21 pm

screen shot 2017-08-07 at 9 05 36 pm

olavloite:
The default executor for GrpcTransportOptions created non-daemon threads, possibly keeping applications alive after the main method has finished.
Note: While this fixes #2315, I do not know whether this could cause unwanted side effects for other clients.


sduskis:
This is a great find.

PR #5195, opened 2019-05-17
DataStore:Fix countEntities method implemented

### LINKED ISSUE #5061: StructuredQuery.Builder - Entity Count Query - Allow limit of 0
According to the group, the best solution for counting entities is:

> What Alfred describes is how countEntities is implemented. it simpy(?) does a normal query with a limit of 0 and a max_int offset. The count is then available via the response.batch.skipped_results (where response is the RunQueryResponse returned by RunQuery), however it may be necessary to run the query multiple times starting from reponse.batch.skipped_cursor if response.batch.more_results is not NO_MORE_RESULTS/MORE_RESULTS_AFTER_CURSOR.

https://groups.google.com/forum/#!topic/gcd-discuss/wH8lVOA-a8Y

This was made possible by this ticket.
https://github.com/googleapis/google-cloud-java/issues/3279

However, currently the setLimit enforces a value > 0.

```
    @Override
    public B setLimit(Integer limit) {
      Preconditions.checkArgument(limit == null || limit > 0, "limit must be positive");
      this.limit = limit;
      return self();
    }
```

https://googleapis.github.io/google-cloud-java/google-cloud-clients/apidocs/com/google/cloud/datastore/StructuredQuery.Builder.html#setLimit-java.lang.Integer-

StructuredQuery.java needs to be updated to allow a value >= 0.

https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQuery.java#L760

This will allow the following code to work:
```
        int count = 0;
        Query query = Query.newKeyQueryBuilder().setKind(kind)
                .setOffset(Integer.MAX_VALUE).setLimit(1).build();
        results = datastore.run(query);
        for (count = results.getSkippedResults(); results.getMoreResults() != QueryResultBatch.MoreResultsType.NO_MORE_RESULTS; i++) {
            query = query.toBuilder().setStartCursor(results.cursorAfter).build();
            results = gdatastore.run(query);
            count += results.getSkippedResults();
        }
```

abhinav-qlogic:
Fixes #5061

### DIFF
--- a/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/QueryResultsImpl.java
+++ b/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/QueryResultsImpl.java
@@ -127,18 +127,19 @@ class QueryResultsImpl<T> extends AbstractIterator<T> implements QueryResults<T>
   }
 
   @Override
-  public int getCountEntities() {
-    return getEntitiesCount();
+  public int countEntities() {
+    return count();
   }
 
-  private int getEntitiesCount() {
+  private int count() {
     int count = 0;
     Integer limit = ((KeyQuery) this.query).getLimit();
     if (null != limit && limit == 0) {
-      count = runQueryResponsePb.getBatch().getSkippedResults();
-      while (runQueryResponsePb.getBatch().getMoreResults()
+      QueryResultBatch queryResultBatch = runQueryResponsePb.getBatch();
+      count = queryResultBatch.getSkippedResults();
+      while (queryResultBatch.getMoreResults()
           != QueryResultBatch.MoreResultsType.NO_MORE_RESULTS) {
-        count += runQueryResponsePb.getBatch().getSkippedResults();
+        count += queryResultBatch.getSkippedResults();
       }
     }
     return count;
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/QueryResultsImpl.java
```diff
@@ -124,4 +125,23 @@ public int getSkippedResults() {
   public MoreResultsType getMoreResults() {
     return moreResults;
   }
+
+  @Override
+  public int countEntities() {
+    return count();
```
sduskis: There's no need for a separate `count()` method.  Please inline it.

google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/StructuredQuery.java
```diff
@@ -757,7 +757,7 @@ public B setOffset(int offset) {
 
     @Override
     public B setLimit(Integer limit) {
-      Preconditions.checkArgument(limit == null || limit > 0, "limit must be positive");
+      Preconditions.checkArgument(limit == null || limit >= 0, "limit must be positive");
```
sduskis: limit should still be positive please revert this.

ajaaym:
closing this for now, waiting for user to reply.

PR #5200, opened 2019-05-18
Spanner: Create new instance if existing Spanner is closed

olavloite:
SpannerOptions caches any Spanner instance that has been created, and hands this cached instance out to all subsequent calls to SpannerOptions.getService(). This also included closed Spanner instances. The getService() method now creates a new instance if the Spanner instance has already been closed.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
@@ -374,7 +374,8 @@ public class SpannerOptionsTest {
     Spanner service3 = options.getService();
     assertThat(service3 == service1, is(false));
     assertThat(service1.isClosed()).isTrue();
-    assertThat(service3.isClosed()).isFalse();;
+    assertThat(service3.isClosed()).isFalse();
+    ;
     // Getting another service from the SpannerOptions should return the new cached instance.
     Spanner service4 = options.getService();
     assertThat(service3 == service4, is(true));


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -298,6 +298,36 @@ protected SpannerRpc getSpannerRpcV1() {
...
+  /**
+   * Returns a {@link Spanner} service object. This method will create a {@link Spanner} instance on
+   * the first call and subsequent calls to this method will return the same {@link Spanner}
+   * instance. If the instance is closed, the following call to this method will create a new {@link
+   * Spanner} instance.
+   */
+  @Override
+  public Spanner getService() {
+    Spanner spanner = super.getService();
+    if (spanner.isClosed()) {
```
sduskis: I think that the new object needs to be cached.  Is there a way to make sure that `super.getService()`'s object is refreshed?  The same comment applies to `getRpc()`

olavloite: Yes, that is done by the method call `ServiceOptions.createNewService()` on L311 (and on L326 for `getRpc()`).

ajaaym:
I think the design followed here is one instance of ServiceOption refer to one instance of Service, rather than treating ServiceOption as factory for Service. If service returned from serviceOptions is used after close it would throw proper error message which would direct user to use new ServiceOptions. This change would be inconsistent with the other client in this library.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
sduskis: Where is the object cached?

olavloite: It's cached in the super class. The `super.getService()`, which is called on first line of the method, gets the cached object from the super class. If it is closed, the `createNewService()` method (also defined in the super class `ServiceOptions`) will refresh it.

The implementation of `ServiceOptions.getService()` looks like this:

```java
  ...
  private transient ServiceT service;
  ...
  public ServiceT getService() {
    if (service == null) {
      service = serviceFactory.create((OptionsT) this);
    }
    return service;
  }
```

sduskis: Won't the super class will cache the closed service, hence after super.service is closed, each new call to `getService()` will get a new instance?

olavloite: The easiest way to show what happens is by showing the code as if it was in one class (in reality it is split between the super class and sub class, the `createNewService()` method is in the super class). This is what is going on:

```java
  ...
  private transient ServiceT service;
  ...
  @Override
  public Spanner getService() {
    Spanner spanner = super.getService();
    if (spanner.isClosed()) {
      spanner = createNewService();
    }
    return spanner;
  }

  protected ServiceT createNewService() {
    service = serviceFactory.create((OptionsT) this);
    return service;
  }
```

olavloite: I've added an additional test case to ensure that the correct caching behavior is actually happening here: https://github.com/googleapis/google-cloud-java/blob/75d68e7ee646065f0794923520c0332a41d278d9/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java#L89

sduskis: ugh... I reviewed the code a little closer, and debugged it.  I see what you did now.  it was a bit confusing.

Perhaps change this:

```
  public ServiceT getService() {
    if (service == null) {
      service = serviceFactory.create((OptionsT) this);
    }
    return service;
  }

  protected ServiceT createNewService() {
    service = serviceFactory.create((OptionsT) this);
    return service;
  }
```

to 
```
  public ServiceT getService() {
    if (shouldRefreshService()) {
      service = serviceFactory.create((OptionsT) this);
    }
    return service;
  }

  protected boolean shouldRefreshService() {
    return service == null;
  }
```

That would make the logic a bit less spread out and readable.

olavloite: That is a cleaner solution. The subclass does need access to the cached object, so I've included the cached service object to the signature of the `shouldRefreshService(ServiceT)` method.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -433,7 +433,41 @@ public TransportOptions getDefaultTransportOptions() {
...
+
+  /**
+   * Returns a {@link Spanner} service object. This method will create a {@link Spanner} instance on
+   * the first call and subsequent calls to this method will return the same {@link Spanner}
+   * instance. If the instance is closed, the following call to this method will create a new {@link
+   * Spanner} instance.
+   */
+  @Override
+  public Spanner getService() {
+    // Method is only overridden in order to supply additional documentation.
```
kolea2: My preference here would be to instead add documentation to getService in ServiceOptions so we don't have to override this (same with getRpc).

olavloite: That's not possible as there are Service classes other than Spanner that are not closeable and do not have an `isClosed()` method, which means that the behavior that is described for Spanner is only applicable to Spanner. That is probably also the reason that the default implementation is to cache the Service object indefinitely without checking its validity.

One possible generic solution could be to change the `ServiceOptions` class to implement the behavior of this PR for all Service classes that implement a custom interface `CloseableService`. All other Service classes would keep the current default behavior.

kolea2: Oh, sorry, didn't realize that was part of google-cloud-core. Can we perhaps add documentation to shouldRefreshService?

olavloite: I've added documentation to the shouldRefresh... methods.
Note that we cannot/should not rely on the documentation of these methods for our users, as these methods are protected.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -494,24 +494,40 @@ static String getServiceAccountProjectId(String credentialsPath) {
...
   public ServiceRpc getRpc() {
-    if (rpc == null) {
+    if (shouldRefreshRpc(rpc)) {
       rpc = serviceRpcFactory.create((OptionsT) this);
     }
     return rpc;
   }
 
+  /**
+   * @param cachedService The currently cached service object
```
kolea2: minor - rename to cachedRpc

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -353,4 +356,26 @@ public void testNullSessionLabels() {
...
+
+  @Test
+  public void testDoNotCacheClosedSpannerInstance() {
+    SpannerOptions options =
+        SpannerOptions.newBuilder()
+            .setProjectId("[PROJECT]")
+            .setCredentials(NoCredentials.getInstance())
+            .build();
+    // Getting a service twice should give the same instance.
+    Spanner service1 = options.getService();
```
kolea2: for completeness and coverage, can you add in a few assertions like:
`assertThat(spanner1.isClosed()).isFalse();`

kolea2:
LGTM besides a few minor comments

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
olavloite: Changed.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
olavloite: Added assertions.

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -494,24 +494,40 @@ static String getServiceAccountProjectId(String credentialsPath) {
...
       service = serviceFactory.create((OptionsT) this);
     }
     return service;
   }
 
+  /**
+   * @param cachedService The currently cached service object
+   * @return true if the currently cached service object should be refreshed.
+   */
+  protected boolean shouldRefreshService(ServiceT cachedService) {
```
chingor13: Should this and `shouldRefreshRpc` be static? 

It looks like we're not using anything from the instance but we're allowing for that possibility in the future. If so, should we instead provide a protected getter for the cached service/rpc?

olavloite: The idea is to let these two methods be overridable. The base `Service` and `ServiceRpc` classes are not closeable and therefore only does a `cachedService == null` check.

The `Spanner` class (a subclass of `Service`) implements `AutoCloseable` and `SpannerOptions` therefore implements a custom check in the `shouldRefreshService()` method that calls `Spanner#isClosed()` instead of the `cachedService == null` check.

PR #5214, opened 2019-05-23
BigQuery: Augment BQ Model metadata.

shollyman:
This change introduces extended ML model metadata into the existing CRUD
implementation for BigQuery ML models.  This implementation exposes the
underlying discovery types directly without an abstraction for the ML metadata.

Methods that provide this metadata are all marked @BetaApi to indicate the getters may change.


shollyman:
> LGTM. This is so much nicer than if we had to duplicate all those properties manually.

The combination of the annotation, that the representations are effectively read only, and the fact that we pin the discovery artifacts in the maven dependencies makes me much more comfortable here.  Assuming no other comments come up, I'll merge it when I'm back in the office tomorrow my time.

PR #5242, opened 2019-05-30
Add integration tests for snapshot read and resuming.

mmladenovski:
The change adds the following integration tests:

* Resuming by row offset when calling ReadRows.
* Reading rows at different snapshots in time.

@tswast or @shollyman, please review.

CC: @kmjung , @aryann

### DIFF
--- a/google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
+++ b/google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
@@ -67,7 +67,6 @@ import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.threeten.bp.Duration;
-import org.threeten.bp.Instant;
 
 /** Integration tests for BigQuery Storage API. */
 public class ITBigQueryStorageTest {
@@ -175,7 +174,7 @@ public class ITBigQueryStorageTest {
     // We have to read some number of rows in order to be able to resume. More details:
     // https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1beta1#google.cloud.bigquery.storage.v1beta1.ReadRowsRequest
 
-    long avroRowCount = SkipNumberOfRows(session.getStreams(0), /* rowsToSkip = */ 34_846);
+    long avroRowCount = ReadStreamToOffset(session.getStreams(0), /* rowOffset = */ 34_846);
 
     StreamPosition readPosition =
         StreamPosition.newBuilder()
@@ -373,18 +372,18 @@ public class ITBigQueryStorageTest {
             .setProjectId(ServiceOptions.getDefaultProjectId())
             .build();
 
-    RunQueryJobAndExpectSuccess(
-        /* destinationTableId = */ testTableId, /* query = */ "SELECT 1 AS col");
-    Instant firstSnapshot = Instant.now();
+    Job firstJob =
+        RunQueryJobAndExpectSuccess(
+            /* destinationTableId = */ testTableId, /* query = */ "SELECT 1 AS col");
 
-    RunQueryJobAndExpectSuccess(
-        /* destinationTableId = */ testTableId, /* query = */ "SELECT 2 AS col");
-    Instant secondSnapshot = Instant.now();
+    Job secondJob =
+        RunQueryJobAndExpectSuccess(
+            /* destinationTableId = */ testTableId, /* query = */ "SELECT 2 AS col");
 
     final List<Long> rowsAfterFirstSnapshot = new ArrayList<>();
     ProcessRowsAtSnapshot(
         tableReference,
-        firstSnapshot,
+        firstJob.getStatistics().getEndTime(),
         new AvroRowConsumer() {
           @Override
           public void accept(GenericRecord record) {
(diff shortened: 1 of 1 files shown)


kmjung:
cc: @JesseLovelace

google-cloud-clients/google-cloud-bigquerystorage/pom.xml
```diff
@@ -41,6 +41,10 @@
             <artifactId>gax-grpc</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-bigquery</artifactId>
+        </dependency>
```
sduskis: Is this dependency only added for testing?  If so, can you please add `<scope>test</scope>`?

mmladenovski: It is only needed for test. Forgot to adjust scope when I added it.

Done with commit: https://github.com/googleapis/google-cloud-java/pull/5242/commits/8c728a6e16a249baef728f80c384cdd5df97d9a5

google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
```diff
@@ -254,4 +352,174 @@ public void accept(GenericRecord record) {
...
+    bigquery.create(TableInfo.of(testTableId, StandardTableDefinition.of(tableSchema)));
+
+    TableReference tableReference =
+        TableReference.newBuilder()
+            .setTableId(testTableId.getTable())
+            .setDatasetId(DATASET)
+            .setProjectId(ServiceOptions.getDefaultProjectId())
+            .build();
+
+    RunQueryJobAndExpectSuccess(
```
shollyman: Consider using the snapshot times from the job's metadata rather than Instant.now().  Using the local system time for snapshots will cause flakiness due to clock drift between the testing instance and the backend, which we've seen in other tests related to snapshot times.

google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
```diff
@@ -104,6 +151,57 @@ public void testSimpleRead() {
...
+        String.format(
+            "Did not receive expected number of streams for table reference '%s' CreateReadSession response:%n%s",
+            TextFormat.shortDebugString(tableReference), session.toString()),
+        1,
+        session.getStreamsCount());
+
+    // We have to read some number of rows in order to be able to resume. More details:
+    // https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1beta1#google.cloud.bigquery.storage.v1beta1.ReadRowsRequest
+
+    long avroRowCount = SkipNumberOfRows(session.getStreams(0), /* rowsToSkip = */ 34_846);
```
shollyman: Naming of SkipNumberOfRows implied to me you're merely forwarding the offset, but it turns out it also performs reads to that point due to the need to allocated data to the stream.  Would this be better named something like ReadStreamToOffset?

google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
```diff
@@ -104,6 +151,57 @@ public void testSimpleRead() {
     assertEquals(164_656, avroRowCount);
   }
 
+  @Test
+  public void testSimpleReadAndResume() {
```
kmjung: What is the end user use case for resume? Shouldn't this be handled transparently by the client?

mmladenovski: @kmjung , it is handled transparently by the client. The goal of this test is to explicitly verify it.

google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
mmladenovski: @shollyman , feedback applied in commit:
https://github.com/googleapis/google-cloud-java/pull/5242/commits/83a7a95a3e54b85cf83e7d7b10b2b293f81dd5fc

google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
mmladenovski: @shollyman , feedback applied in commit:
https://github.com/googleapis/google-cloud-java/pull/5242/commits/83a7a95a3e54b85cf83e7d7b10b2b293f81dd5fc

google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
mmladenovski: Resuming by offset can also be used explicitly by clients in scenarios:

1. The row progress per stream is stored in customers systems
1. Customer pauses the stream reading for applying changes on their system.
1. Customer resumes processing.

I thought it would be good to have an explicit tests for this functionality.

shollyman:
Kokoro - Test: Windows Java 8 failure is unrelated to changes (spanner flakes).

Issue #4384, opened 2019-01-17
Remove warnings


sduskis:
There are a bunch of deprecation warnings.  We should clean up a bunch of them in hand-crafted libraries, and work on fixing the generators for the generated libraries.

ajaaym:
@sduskis yeah she will start on this.

PR #5247, opened 2019-05-31
FireStore:deprecated warnings removed

abhinav-qlogic:
Works towards #4384


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -339,9 +339,11 @@ public void onSuccess(List<WriteResult> writeResults) {
                             span.end();
                             resultFuture.set(userResult);
                           }
-                        });
```
sduskis: This should be `MoreExecutors.directExecutor()`

sduskis: All callbacks that didn't have an explicit executor should use `MoreExecutors.directExecutor()`

PR #5284, opened 2019-06-03
Storage: HMAC service account support

### LINKED ISSUE #5054: Storage: HMAC Service Account key support
assign: @JesseLovelace

JesseLovelace:
Fixes #5054

Adds support for HMAC service accounts. Holding off on adding documentation until the general implementation is approved


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/HmacKey.java
```diff
@@ -0,0 +1,287 @@
...
+    this.secretKey = builder.secretKey;
+    this.metadata = builder.metadata;
+  }
+
+  public static Builder newBuilder(String secretKey) {
+    return new Builder(secretKey);
+  }
+
+  public static class Builder {
+    private String secretKey;
```
jkwlui: Not super familiar with documenting Java code, but can we add a note saying the `secret` is only available on creation?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/HmacKey.java
```diff
@@ -0,0 +1,287 @@
...
+    return secretKey;
+  }
+
+  public HmacKeyMetadata getMetadata() {
+    return metadata;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(secretKey,  metadata);
```
jkwlui: extra space here?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/HmacKey.java
JesseLovelace: yeah--i'm planning to add the documentation once the code is otherwise approved

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -63,6 +63,7 @@
 import com.google.common.io.BaseEncoding;
 import com.google.common.net.UrlEscapers;
 import com.google.common.primitives.Ints;
+import com.google.cloud.storage.HmacKey.HmacKeyMetadata;
```
frankyn: I don't think this is necessary.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2709,19 @@ Blob create(
...
+  Page<HmacKeyMetadata> listHmacKeys(
+      ServiceAccount serviceAccount, String pageToken, Long maxResults);
+
+  Page<HmacKeyMetadata> listHmacKeys(ServiceAccount serviceAccount);
+
+  HmacKeyMetadata getHmacKey(String accessId);
+
+  void deleteHmacKey(String accessId);
+
+  HmacKeyMetadata updateHmacKeyState(
```
frankyn: Nice helper method! (no action required).

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -35,6 +35,7 @@
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.io.BaseEncoding;
+import com.google.cloud.storage.HmacKey.HmacKeyMetadata;
```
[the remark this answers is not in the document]
JesseLovelace: It is, would need "HmacKey.HmacKeyMetadata" without it

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
JesseLovelace: It is, would need "HmacKey.HmacKeyMetadata" without it

frankyn: ah, gotcha. Thanks for clarifying!

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2709,17 +2709,117 @@ Blob create(
...
+   * @param maxResults the maximum amount of results that can be returned by this request
+   * @param showDeletedKeys whether to show keys with the DELETED state in the result
+   * @throws StorageException upon failure
+   */
   Page<HmacKeyMetadata> listHmacKeys(
-      ServiceAccount serviceAccount, String pageToken, Long maxResults);
+      ServiceAccount serviceAccount, String pageToken, Long maxResults, boolean showDeletedKeys);
 
+  /**
+   * Lists HMAC keys for a given service account. Note this returns {@code HmacKeyMetadata} objects,
```
frankyn: Lists active HMAC keys for a given service account.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2709,17 +2709,117 @@ Blob create(
...
+   * which do not contain secret keys.
+   *
+   * <p>Example of listing HMAC keys, specifying max results.
+   *
+   * <pre>{@code
+   * ServiceAccount serviceAccount = ServiceAccount.of("my-service-account@google.com");
+   *
+   * Page<HmacKey.HmacKeyMetadata> metadataPage = storage.listHmacKeys(serviceAccount, null, 10);
```
frankyn: Could you add the boolean value of true for Deleted keys?

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2709,17 +2709,117 @@ Blob create(
...
+   *
+   * @param serviceAccount the service account whose HMAC keys to list
+   * @param pageToken the page from which to start results
+   * @param maxResults the maximum amount of results that can be returned by this request
+   * @param showDeletedKeys whether to show keys with the DELETED state in the result
+   * @throws StorageException upon failure
+   */
   Page<HmacKeyMetadata> listHmacKeys(
-      ServiceAccount serviceAccount, String pageToken, Long maxResults);
+      ServiceAccount serviceAccount, String pageToken, Long maxResults, boolean showDeletedKeys);
```
frankyn: Out of curiosity, why the overload for showDeletedKeys?

JesseLovelace: My reasoning here is that the most common use case is going to be listHmacKeys(serviceAccount), so offering that along with one other signature to be more precise would meet most needs

frankyn: Gotcha that makes sense. What about defaulting to not showing deleted keys?

JesseLovelace: That is already the default I believe

jkwlui: API defaults to `false` if not given.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2709,119 @@ Blob create(
...
+   *
+   * Page<HmacKey.HmacKeyMetadata> metadataPage = storage.listHmacKeys(serviceAccount);
+   * for (HmacKey.HmacKeyMetadata hmacKeyMetadata : metadataPage.getValues()) {
+   *     //do something with the metadata
+   * }
+   * }</pre>
+   *
+   * @throws StorageException upon failure
+   */
+  Page<HmacKeyMetadata> listHmacKeys(ServiceAccount serviceAccount);
```
jkwlui: `listHmacKeys` should also be able to accept no `serviceAccount`, which will show HMAC keys from all accounts. Can your implementation accept no `serviceAccount`?

JesseLovelace: Listing keys from all accounts can be accomplished by calling this with serviceAccount=null. I'll add a note about it in the documentation

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
frankyn: Thanks @jkwlui and @JesseLovelace!

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
jkwlui: Got it. That would really help, thank you!

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2709,120 @@ Blob create(
...
+   * }</pre>
+   *
+   * @param serviceAccount the service account whose HMAC keys to list
+   * @param pageToken the page from which to start results
+   * @param maxResults the maximum amount of results that can be returned by this request
+   * @param showDeletedKeys whether to show keys with the DELETED state in the result
+   * @throws StorageException upon failure
+   */
+  Page<HmacKeyMetadata> listHmacKeys(
+      ServiceAccount serviceAccount, String pageToken, Long maxResults, boolean showDeletedKeys);
```
frankyn: Based on other RPCs the Storage library has parameters: `serviceAccountEmail`, `pageToken`, `showDeletedKeys`, and `maxResults` should be listed under an Options optional parameter because they are optional. The only required value is `project_id` which is provided by client credentials. 

Other HmacKey RPCs should have a similar Options parameter for the value userProject. At the moment it's ignored by the service but it's expected to be available. Given it would change the signature of the method I thought it'd be best to do it now rather than wait until later. (I'm open to opinions here).

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -253,6 +254,40 @@ public static BucketSourceOption userProject(String userProject) {
...
+    /** Returns an option to specify the page token from which to start listing HMAC keys. */
+    public static ListHmacKeysOption pageToken(String pageToken) {
+      return new ListHmacKeysOption(StorageRpc.Option.PAGE_TOKEN, pageToken);
+    }
+
+    /**
+     * Returns an option to specify whether to show deleted keys in the result. This option is false
+     * by default.
+     */
+    public static ListHmacKeysOption showDeletedKeys(boolean showDeletedKeys) {
```
frankyn: `userProject` should be available based on the discovery document. https://www.googleapis.com/discovery/v1/apis/storage/v1/rest

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2743,98 @@ Blob create(
...
+   *
+   * HmacKey hmacKey = storage.createHmacKey(serviceAccount);
+   *
+   * String secretKey = hmacKey.getSecretKey();
+   * HmacKey.HmacKeyMetadata metadata = hmacKey.getMetadata();
+   * }</pre>
+   *
+   * @throws StorageException upon failure
+   */
+  HmacKey createHmacKey(ServiceAccount serviceAccount);
```
frankyn: Include options for `userProject`.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2743,98 @@ Blob create(
...
+   * <p>Example of getting an HMAC key.
+   *
+   * <pre>{@code
+   * String hmacKeyAccessId = "my-access-id";
+   * HmacKey.HmackeyMetadata hmacKeyMetadata = storage.getHmacKey(hmacKeyAccessId);
+   * }</pre>
+   *
+   * @throws StorageException upon failure
+   */
+  HmacKeyMetadata getHmacKey(String accessId);
```
frankyn: Include options for `userProject`.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2743,98 @@ Blob create(
...
+   * String hmacKeyAccessId = "my-access-id";
+   * HmacKey.HmacKeyMetadata hmacKeyMetadata = storage.getHmacKey(hmacKeyAccessId);
+   *
+   * storage.updateHmacKeyState(hmacKeyMetadata, HmacKey.HmacKeyState.INACTIVE);
+   * storage.deleteHmacKey(hmacKeyMetadata.getAccessId());
+   * }</pre>
+   *
+   * @throws StorageException upon failure
+   */
+  void deleteHmacKey(String accessId);
```
frankyn: Include options for `userProject`.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2743,98 @@ Blob create(
...
+   * ServiceAccount serviceAccount = ServiceAccount.of("my-service-account@google.com");
+   * HmacKey key = storage.createHmacKey(serviceAccount);
+   *
+   * storage.updateHmacKeyState(hmacKey.getMetadata(), HmacKey.HmacKeyState.INACTIVE);
+   * }</pre>
+   *
+   * @throws StorageException upon failure
+   */
+  HmacKeyMetadata updateHmacKeyState(
+      final HmacKeyMetadata hmacKeyMetadata, final HmacKey.HmacKeyState state);
```
frankyn: Include options for `userProject`.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2835,100 @@ Blob create(
...
+   * <pre>{@code
+   * ServiceAccount serviceAccount = ServiceAccount.of("my-service-account@google.com");
+   *
+   * Page<HmacKey.HmacKeyMetadata> metadataPage = storage.listHmacKeys(
+   *     Storage.ListHmacKeysOption.serviceAccount(serviceAccount),
+   *     Storage.ListHmacKeysOption.maxResults(10L),
+   *     Storage.ListHmacKeysOption.showDeletedKeys(true));
+   * for (HmacKey.HmacKeyMetadata hmacKeyMetadata : metadataPage.getValues()) {
+   *     //do something with the metadata
+   * }
```
frankyn: Add example of listing keys using only the projectId as well which is the simplified example.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2835,100 @@ Blob create(
...
+
+  /**
+   * Gets an HMAC key given its access id. Note that this returns a {@code HmacKeyMetadata} object,
+   * which does not contain the secret key.
+   *
+   * <p>Example of getting an HMAC key.
+   *
+   * <pre>{@code
+   * String hmacKeyAccessId = "my-access-id";
+   * HmacKey.HmackeyMetadata hmacKeyMetadata = storage.getHmacKey(hmacKeyAccessId);
```
frankyn: Add comment that it uses the same project as the storage client instance.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -1231,6 +1234,132 @@ public ObjectAccessControl patchAcl(ObjectAccessControl acl) {
...
+      span.setStatus(Status.UNKNOWN.withDescription(ex.getMessage()));
+      throw translate(ex);
+    } finally {
+      scope.close();
+      span.end();
+    }
+  }
+
+  @Override
+  public HmacKeyMetadata getHmacKey(String accessId, Map<Option, ?> options) {
```
frankyn: Silly question, but what if getHmacKey used HMACMetadata as well? Throwing it out there but okay if you're okay with moving forward as-is as well.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2708,6 +2835,100 @@ Blob create(
...
+  void deleteHmacKey(HmacKeyMetadata hmacKeyMetadata, DeleteHmacKeyOption... options);
+
+  /**
+   * Updates the state of an HMAC key and returns the updated metadata.
+   *
+   * <p>Example of updating the state of a newly created HMAC key.
+   *
+   * <pre>{@code
+   * ServiceAccount serviceAccount = ServiceAccount.of("my-service-account@google.com");
+   * HmacKey key = storage.createHmacKey(serviceAccount);
```
frankyn: Perform a Get request instead of a create.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
JesseLovelace: I don't really see the benefit of that, it just seems confusing to me. I'd rather move forward with accessId

PR #5309, opened 2019-06-04
Spanner: Properly end span for ResumableStreamIterator.

kolea2:
Most changes in the test are because of variable renaming, new test is closedSpan().


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractResultSet.java
```diff
@@ -847,6 +847,7 @@ protected ResumableStreamIterator(int maxBufferSize, String streamName, Span par
     public void close(@Nullable String message) {
       if (stream != null) {
         stream.close(message);
+        span.end();
```
chingor13: Is it possible that `close` is called more than once?

chingor13: Looks like on the OpenCensus side, calling `end()` on an already ended span is a no-op with logging.

PR #5363, opened 2019-06-05
Cloud Scheduler GA

chingor13:


chingor13:
Releasetool should handle replacing the versions although we'll have to fix it as the first release will end up 1.1.0

PR #5405, opened 2019-06-05
Logging:deprecated warnings removed

abhinav-qlogic:
work towards #4384


google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java
```diff
@@ -629,7 +632,7 @@ public void onFailure(Throwable t) {
   static ListLogEntriesRequest listLogEntriesRequest(
       String projectId, Map<Option.OptionType, ?> options) {
     ListLogEntriesRequest.Builder builder = ListLogEntriesRequest.newBuilder();
-    builder.addProjectIds(projectId);
+    builder.addResourceNames(projectId);
```
sduskis: I dont think this is the same. Calling `addProjectIds(projectId)` should be the same as `.addResourceNames("project/" + projectId)`

PR #5412, opened 2019-06-05
Spanner: Retry aborted partitioned dml transactions

### LINKED ISSUE #5410: Retry on Transaction Aborted Errors for Partitioned DML

olavloite:
Fixes #5410

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
@@ -28,6 +28,7 @@ import com.google.spanner.v1.Transaction;
 import com.google.spanner.v1.TransactionOptions;
 import com.google.spanner.v1.TransactionSelector;
 import java.util.Map;
+import java.util.concurrent.Callable;
 
 /** Partitioned DML transaction for bulk updates and deletes. */
 class PartitionedDMLTransaction implements SessionTransaction {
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDMLTransaction.java
```diff
@@ -59,24 +57,40 @@ private ByteString initTransaction() {
...
-            .setSession(session.getName())
-            .setTransaction(TransactionSelector.newBuilder().setId(transactionId).build());
-    Map<String, Value> stmtParameters = statement.getParameters();
-    if (!stmtParameters.isEmpty()) {
-      com.google.protobuf.Struct.Builder paramsBuilder = builder.getParamsBuilder();
-      for (Map.Entry<String, Value> param : stmtParameters.entrySet()) {
-        paramsBuilder.putFields(param.getKey(), param.getValue().toProto());
-        builder.putParamTypes(param.getKey(), param.getValue().getType().toProto());
+    com.google.spanner.v1.ResultSet resultSet;
+    while (true) {
```
snehashah16: can we avoid these local retry loops and build the txn aborted retry generically into our retry logic ?

there are several other retry codes that probably got deleted with a refactor/change ?

 com.google.spanner.v1.ResultSet resultSet =
          runWithRetries(            //      lost this ?
              new Callable<com.google.spanner.v1.ResultSet>() {
                @Override
                public com.google.spanner.v1.ResultSet call() throws Exception {
                  return rpc.executeQuery(builder.build(), session.options);
                }
              });
These are all the codes we should be retrying on:
private static boolean isRetryable(ErrorCode code, @Nullable Throwable cause) {
    switch (code) {
      case INTERNAL:
        return hasCauseMatching(cause, Matchers.isRetryableInternalError);
      case UNAVAILABLE:
        return true;
      case RESOURCE_EXHAUSTED:
        return SpannerException.extractRetryDelay(cause) > 0;
      default:
        return false;
    }
  }

olavloite: The above mentioned method was removed during the refactor to use Gax for retries. Retries of single gRPC calls are now automatically handled and retried by Gax (i.e. an rpc call that temporarily returns UNAVAILABLE will automatically be retried by Gax without the need for any custom code in the Spanner library).

Transaction retries are still handled by custom loops in the Spanner library, as these need to retry application logic and possibly multiple gRPC calls. We could refactor these to use the standard Gax `com.google.cloud.RetryHelper` instead of our own custom retry loops. I've changed this PR for PDML to use the `RetryHelper`.

I'll submit a separate PR for using `RetryHelper` for normal transaction retries.

PR #5413, opened 2019-06-05
Update google-auth-library to 0.16.1 and import using new bom

chingor13:
* Upgrades google-auth-library to 0.16.1
* Upgrades google-http-client to 0.30.1
* Upgrades google-oauth-client to 0.30.1
* Import google-auth-library-bom
* Drops dependency on Apache HttpClient which was only used to get a HTTP status code constant.


chingor13:
Dependency convergence issues. Will need to update http-client in this PR too

chingor13:
Looks like we found a regression in http-client with the `Data.nullOf` for a Map type

chingor13:
@sduskis This is ready for review. We fixed the regression in google-http-client 1.30.1.

PR #5433, opened 2019-06-06
Spanner: Use generic retry helper for transaction retries

olavloite:
As discussed in PR #5412 we should move away from custom retry loops and use a generic retry helper. This PR changes the retry loop of read/write transactions from a custom loop to using the Gax `RetryHelper`.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerRetryHelperTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerRetryHelperTest.java
@@ -71,7 +71,7 @@ public class SpannerRetryHelperTest {
         fail(
             String.format(
                 "unexpected error %s, expected %s",
-                e.getErrorCode().name(), ErrorCode.DEADLINE_EXCEEDED.name()));
+                e.getErrorCode().name(), ErrorCode.CANCELLED.name()));
       }
     }
   }


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -321,96 +302,98 @@ public TransactionRunner allowNestedTransaction() {
...
-        span.addAnnotation(
-            "Transaction Attempt Failed in Commit",
-            ImmutableMap.<String, AttributeValue>builder()
-                .putAll(TraceUtil.getExceptionAnnotations(e))
-                .put("Attempt", AttributeValue.longAttributeValue(attempt))
-                .build());
-        throw e;
-      }
-    }
+  private <T> T runInternal(final TransactionCallable<T> txCallable) {
```
kolea2: Would you mind just briefly explaining what this change is? Hard to see from the diff. Is it mainly the return statement at the bottom?

olavloite: The web diff view is indeed not very helpful in this case...

In general: Instead of using a while(true) {...} loop for retrying the transaction, the transaction is placed in a `Callable` that can be passed to the generic retry helper. The retry helper will handle the retries if the transaction aborts.

More specifically, the change is as follows:
1. Everything inside the `while(true)` loop is moved into the `call()` method of the `Callable` retryCallable.
1. The `session.newTransaction()` call is moved from the constructor and the `backoff(...)` method to inside the `call()` method to give each attempt a fresh transaction.
1. The `continue` statement on line 354 is replaced with a `throw` statement in order to leave the method exceptionally.
1. The `Callable` is passed in to the new method `SpannerRetryHelper.runTxWithRetriesOnAborted(Callable)` that handles the retries if the transaction is aborted.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -321,96 +302,98 @@ public TransactionRunner allowNestedTransaction() {
...
 
-  private <T> T runInternal(TransactionCallable<T> callable) {
-    BackOff backoff = SpannerImpl.newBackOff();
-    final Context context = Context.current();
-    int attempt = 0;
-    // TODO: Change this to use TransactionManager.
-    while (true) {
-      checkState(
-          isValid, "TransactionRunner has been invalidated by a new operation on the session");
-      SpannerImpl.checkContext(context);
```
kolea2: What happens if the context is cancelled now? Are the same error messages thrown for deadline exceeded and cancelled?

olavloite: In the initial version of this PR, the context was ignored. I've added a check for context cancellation to the generic retry helper, and it will now throw the same exceptions as previous retry loops.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRetryHelper.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Util class for retrying aborted transactions. This class is a wrapper around {@link RetryHelper}
+ * that uses specific settings to only retry on aborted transactions, without a timeout and without
+ * a cap on the number of retries.
+ */
+class SpannerRetryHelper {
+  private static final RetrySettings txRetrySettings =
+      RetrySettings.newBuilder()
+          .setInitialRetryDelay(Duration.ofMillis(1000L))
```
kolea2: Minor - I would take these numbers and make them constants.

PR #5439, opened 2019-06-06
Add integration tests for partitioning.

mmladenovski:
The tests in this change cover:

* Reading from a column partitioned table.
* Reading from a table that uses ingestion time partitioning.

@tswast or @shollyman, please review.

CC: @kmjung , @aryann

### DIFF
--- a/google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
+++ b/google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
@@ -412,52 +412,28 @@ public class ITBigQueryStorageTest {
 
   @Test
   public void testColumnPartitionedTableByDateField() throws InterruptedException, IOException {
-    Field intFieldSchema =
-        Field.newBuilder("num_field", LegacySQLTypeName.INTEGER)
-            .setMode(Mode.REQUIRED)
-            .setDescription("IntegerDescription")
-            .build();
-    Field dateFieldSchema =
-        Field.newBuilder("date_field", LegacySQLTypeName.DATE)
-            .setMode(Mode.REQUIRED)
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta1/it/ITBigQueryStorageTest.java
```diff
@@ -373,39 +375,157 @@ public void testReadAtSnapshot() throws InterruptedException, IOException {
...
+            .setDescription("DateDescription")
+            .build();
+    com.google.cloud.bigquery.Schema tableSchema =
+        com.google.cloud.bigquery.Schema.of(intFieldSchema, dateFieldSchema);
+
+    TableId testTableId =
+        TableId.of(/* dataset = */ DATASET, /* table = */ "test_column_partition_temp_data");
+    bigquery.create(TableInfo.of(testTableId, StandardTableDefinition.of(tableSchema)));
+
+    RunQueryAppendJobAndExpectSuccess(
```
shollyman: It doesn't look like you're doing anything with the incremental jobs, so why not just do everything at once with a single CREATE TABLE AS SELECT... DDL query job here and in the other test?  It also avoids a lot of builder spam.

mmladenovski: Change applied in a separate commit: https://github.com/googleapis/google-cloud-java/pull/5439/commits/693f9be08244dd8cd8aa9c22f02378e1e058c8ce

The testIngestionTimePartitionedTable() could not be changed to use one DDL statement due to:
* Needs a table to be created with ingestion time partitioning
* Needs one row insert into two different partitions by using $YYYYMMDD as separate statements.

Left the test as is.

mmladenovski:
@shollyman, rebased from latest master and updated the PR.

PR #5473, opened 2019-06-13
Updated `UnsafeByteOperations.unsafeWrap` to `ByteString.copyFrom`

### LINKED ISSUE #5030: google-cloud-bigtable uses unstable protobuf APIs
In looking at grpc/grpc-java#5644 I noticed Range referenced UnsafeByteOperations. That seemed weird, because [UnsafeByteOperations is ExperimentalApi](https://github.com/protocolbuffers/protobuf/blob/v3.7.1/java/core/src/main/java/com/google/protobuf/UnsafeByteOperations.java#L65).

It looks like the reference is at:
https://github.com/googleapis/google-cloud-java/blob/675ace7601e2c09474ce986d5d2223af5cb15f28/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Range.java#L363

Cloud libraries can only use stable APIs from other libraries, as otherwise it causes diamond dependency problems.

rahulKQL:
Fixes #5030
Updated `UnsafeByteOperations.unsafeWrap` as this is still a `@ExperimentalApi`.

To confirm the impact I ran `Range.ByteRangeString#prefix` 1 million times with a bounded random string, I observed that with `UnsafeByteOperations.unsafeWrap` their avg is around **~200ms** and after update with `ByteString.copyFrom` the new avg is **~220ms**.(10% decrement in performance).

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Range.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Range.java
@@ -19,7 +19,6 @@ import com.google.api.core.InternalExtensionOnly;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.ByteString;
-import com.google.protobuf.UnsafeByteOperations;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


igorbernstein2:
I think the performance hit is acceptable here

PR #5482, opened 2019-06-13
BigQuery snippets: Use getValues() instead of iterateAll() when pageSize is specified

JesseLovelace:
This addresses internal bug 124834696

When pageSize is specified, it's confusing/counterintuitive to use iterateAll (Which lists every element including past the current page), instead of getValues (which only includes elements from the current page, i.e. only up to pageSize elements)


tswast:
Wouldn't we want to recommend `iterateAll()` in general, though? I could see `getValues()` being useful for a document that explicitly talks about pagination, such as https://cloud.google.com/bigquery/docs/paging-results

meredithslota:
I'm assuming based on the above discussion that this PR will not be merged and that this is ok to close. @shollyman can you work with @stephaniewang526 to ensure that the sample changes you want to make are clear?

stephaniewang526:
Acknowledged. I will keep `iterateAll()` in most common cases and use `getValues()` only when applicable.

PR #5510, opened 2019-06-18
Preparing Easymock for a future upgrade.

sduskis:
Easymock 4.0 changed `EasyMock.createMock`'s signature in a backwards incompatible way from `.createMock(Class<T> clazz)` to `.createMock(Class<> clazz)`.  This requires adding an explicit generics type to be set.

Easymock 4.0+ doesn't support Java 7, so we can't immediately upgrade.  This PR future proofs Easymock upgrades once we no longer have to support Java 7.


chingor13:
Looks like easymock 4.0 no longer supports Java 7

sduskis:
@chingor13, I reverted the `pom.xml` changes, but kept the generics changes in the code.  Can you PTAL?

PR #5515, opened 2019-06-19
Storage: remove reflection to get raw input stream

ajaaym:
remove reflection to get raw input stream, work towards #3929


ajaaym:
merging this as codecov failure are unrelated to this change.

PR #5525, opened 2019-06-20
Storage: add service api to read storage object in to output stream using direct download

ajaaym:
towards #3929 

Note: This pr just adds the service api. There would be follow up pr to this one which would add public api to download blob as a file or in to output stream using this service api.


ajaaym:
@frankyn @JesseLovelace can you please review?

astroyang:
Is there an estimated time for the next step of adding a public api to download blob using this? @ajaaym

ajaaym:
@astroyang #5791 is merged yesterday.

Issue #5550, opened 2019-06-23
Unable to use Bigtable Query object as cache key 


rculbertson:
Hello!

We'd like to cache Bigtable query results with the [Caffeine library][1], using [Query][2] as the cache key. That's difficult to do right now because the Query object is modified when the request is executed. During the call to `toProto`, the table name and profile ID are set on the `ReadRowsRequest` builder. This means the Query object has a different hash code after the request executes. So a subsequent request with the exact same query parameters will be a cache miss.

One solution would be to make a copy of the Query object before executing the request, and use that as the cache key. But I don't think it's currently possible to copy a Query object. I've submitted PR #5551 which allows users to copy Query objects. 

[1]: https://github.com/ben-manes/caffeine
[2]: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java

igorbernstein2:
Hey man, long time no speak :)

Sorry for the limitations of the current surface. The api was designed to be a dsl rather then a value object, but I can see the issues that this is causing you.
Although the approach in your PR fixes the issue, I would prefer to take a different approach: I think it would be better to remove the side effects of toProto() and make the Query cloneable instead.

I'll try to address this next week

rculbertson:
Hi! Agreed, that sounds like a better fix. Thanks!

PR #5551, opened 2019-06-23
Allow Bigtable Query to be copied

rculbertson:
Added a new create method to the Bigtable Query object that behaves like a
copy constructor. It takes an existing Query object and performs a deep copy.

This is to address #5550, but this could be useful in other situations as well.


sduskis:
As per this comment (https://github.com/googleapis/google-cloud-java/issues/5550#issuecomment-505604768), I'm going to close this PR.  @igorbernstein2 and crew will work on another approach.

PR #5553, opened 2019-06-24
Adding `grpc-alts` in google clients pom

rahulKQL:
Adding this dependency as `InstantiatingGrpcChannelProvider` needs `io.grpc.alts.ComputeEngineChannelBuilder`, which comes from `grpc-alts`.

### DIFF
--- a/google-cloud-clients/pom.xml
+++ b/google-cloud-clients/pom.xml
@@ -313,6 +313,11 @@
       </dependency>
       <dependency>
         <groupId>io.grpc</groupId>
+        <artifactId>grpc-alts</artifactId>
+        <version>${grpc.version}</version>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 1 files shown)


rahulKQL:
@igorbernstein2  Please have a look & Please merge if looks good to you.

chingor13:
Seems like this should be declared by gax-grpc's dependencies. Looks like gax-grpc only declares grpc-alts as a compile dependency.

rahulKQL:
@chingor13 Thanks a lot!! for the background. 
My intent was to get this dependency directly from `google-cloud-clients/pom.xml`, Mostly to skip conflict in the version between `google-cloud-clients` & `gax-grpc` in `cloud-bigtable-client`. 
But now these versions are consistent across our dependencies so closing this PR.

PR #5568, opened 2019-06-26
[Bigtable] minor Javadoc clean up for ModifyColumnFamiliesRequest

rahulKQL:
Removed some undefined tags from `ModifyColumnFamiliesRequest.java`

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ModifyColumnFamiliesRequest.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ModifyColumnFamiliesRequest.java
@@ -48,12 +48,12 @@ public final class ModifyColumnFamiliesRequest {
     this.tableId = tableId;
   }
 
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ModifyColumnFamiliesRequest.java
```diff
@@ -42,33 +42,18 @@ public static ModifyColumnFamiliesRequest of(String tableId) {
...
     this.tableId = tableId;
   }
 
-  /**
-   * Configures the name of the new ColumnFamily to be created
-   *
-   * @param familyId
-   * @return
-   */
+  /** Configures the name of the new ColumnFamily to be created */
```
sduskis: Would it make sense to add `{@link ColumnFamily}`? (and the same applies for all of the other comments)

rahulKQL: Yes, It makes sense to me as well. I have added `{@link ...}`.

PR #5590, opened 2019-06-28
Update gax -> 1.47.0, guava -> 28.0-android, gson -> 2.8.5

### LINKED ISSUE #5555: New Cloud Tasks Library has breaking changes
#### Code example
[Sample](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/tasks/src/main/java/com/example/task/CreateHttpTask.java)

[Test](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/tasks/src/test/java/com/example/task/CreateHttpTaskIT.java)

This sample imports `com.google.cloud.tasks.v2beta3.CloudTasksClient` and is working with version 1.3.0 and breaking with version 1.6.0

#### Stack trace
```
com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: The request deadline is 2019-06-18T20:59:52.360292137-07:00. The deadline cannot be more than 30s in the future.
Caused by: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: The request deadline is 2019-06-18T20:59:52.360292137-07:00. The deadline cannot be more than 30s in the future.
```

chingor13:
Fixes #5555

### DIFF
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -168,7 +168,7 @@
       <dependency>
         <groupId>com.google.api</groupId>
         <artifactId>gax-bom</artifactId>
-        <version>1.46.1</version>
+        <version>1.47.0</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
--- a/google-cloud-clients/pom.xml
+++ b/google-cloud-clients/pom.xml
@@ -161,7 +161,7 @@
     <google.auth.version>0.16.2</google.auth.version>
     <grpc.version>1.21.0</grpc.version>
     <!-- We are currently using the *-android version to support JDK7. -->
-    <guava.version>27.1-android</guava.version>
+    <guava.version>28.0-android</guava.version>
     <http-client-bom.version>1.30.1</http-client-bom.version>
     <oauth-client.version>1.30.1</oauth-client.version>
     <protobuf.version>3.7.1</protobuf.version>
@@ -252,6 +252,13 @@
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>animal-sniffer-annotations</artifactId>
         <version>1.18</version>
+      </dependency>
+
+      <!-- TODO(chingor): Remove this when protobuf-java-util and gax-httpjson converge -->
+      <dependency>
+        <groupId>com.google.code.gson</groupId>
+        <artifactId>gson</artifactId>
+        <version>2.8.5</version>
       </dependency>
 
       <dependency>


chingor13:
Spanner tests that disable gax retries are now failing. For non-retryable methods, we started setting a timeout from the retry config but the default setting is 10 minutes for API calls so it allowed these tests to pass. We "fixed" this for Tasks by preferring the initialRpcTimeout, then maxRpcTimeout, then totalTimeout for the default timeout, but this breaks the spanner tests.

@kolea2 @snehashah16 I'm not sure what the spanner tests that disable gax retries is attempting to do - if we're trying to check that disabling retries also disables timeouts, that's in direct conflict with what gax implemented in https://github.com/googleapis/gax-java/pull/712.

kolea2:
@olavloite can you weigh in here?

olavloite:
@kolea2 @chingor13 @snehashah16 
The Spanner client library used to disable both gax retries and timeouts by setting all retryable codes on the gapic rpc client to an empty set, and then handle all retries and timeouts itself. I added these test cases during the transition from the custom retry handling to using the default gax retry logic in order to verify the behavior in both the old and new situation. I guess we don't need the test cases for the old situation anymore, as the transition has been completed.

olavloite:
@chingor13 I'm changing the test cases to only test for the new situation, i.e. with Gax handling of retries and timeouts enabled. One other thing that I noticed while changing these test cases is that the retry handling of streaming calls have also changed. Can you confirm that that is correct?

The following code sets a timeout for the `executeStreamingSql` method of Spanner. In gax 1.46.1 and earlier, setting this did not cause the method to throw a `DEADLINE_EXCEEDED` exception. In Gax 1.47 it does (when the execution time exceeds the timeout settings).

```java
final RetrySettings retrySettings =
    RetrySettings.newBuilder()
        .setInitialRpcTimeout(Duration.ofMillis(500L))
        .setMaxRpcTimeout(Duration.ofMillis(500L))
        .setMaxAttempts(3)
        .setTotalTimeout(Duration.ofMillis(1500L))
        .build();
SpannerOptions.Builder builder =
    SpannerOptions.newBuilder()
        .setProjectId("[PROJECT]")
        .setChannelProvider(channelProvider)
        .setCredentials(NoCredentials.getInstance());
builder
    .getSpannerStubSettingsBuilder()
    .executeStreamingSqlSettings()
    .setRetrySettings(retrySettings);

...

// In gax 1.46.1 and earlier, this will succeed.
// In gax 1.47, this will throw a DEADLINE_EXCEEDED.
mockSpanner.setExecuteStreamingSqlExecutionTime(ONE_SECOND);
try (ResultSet rs = client.singleUse().executeQuery(SELECT1AND2)) {
  while (rs.next()) {}
}


```

olavloite:
The enclosed patch should fix the broken test cases by removing all test cases with gax disabled. Also, the test case for executeStreamingSql has been updated to reflect the fact that it now does throw a DEADLINE_EXCEEDED when a timeout has been set.

[fix-spanner-test-cases.zip](https://github.com/googleapis/google-cloud-java/files/3345638/fix-spanner-test-cases.zip)

sduskis:
@olavloite, can you please submit a separate PR for those changes?

olavloite:
@sduskis PR for this change can be found here: https://github.com/googleapis/google-cloud-java/pull/5625

I've (temporarily) disabled the test case for `executeStreamingSql` as that test case would fail in gax 1.46.

PR #5607, opened 2019-06-29
Fix publisher throughput when there are no ordering keys.

kamalaboulhosn:


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -327,15 +339,20 @@ private void publishAllWithoutInflight() {
         String key = entry.getKey();
         if (batch.isEmpty()) {
           it.remove();
-        } else if (key.isEmpty() || !sequentialExecutor.hasTasksInflight(key)) {
-          // TODO: Will this cause a performance problem for non-ordering keys scenarios?
+        } else if (key.isEmpty()) {
+          // We will publish the batch with no ordering key outside messagesBatchLock.
+          unorderedOutstandingBatch = batch.popOutstandingBatch();
```
sduskis: You'd need to add `it.remove()` in order to make sure that this batch is only sent once.

Issue #4383, opened 2019-01-17
Remove unused dependencies


sduskis:
Running `mvn dependency:analyze` shows that there are a bunch of "Used undeclared dependencies found" and "Unused declared dependencies found".  This issue is to clean up the "Unused declared dependencies found"

chingor13:
In the split repo world, we will be adding a CI check for `mvn dependency:analyze`

PR #5618, opened 2019-07-01
google-cloud-core-grpc : cleanup dependency 

athakor:
towards #4383


sduskis:
Firestore fails due to read quota issues.

PR #5621, opened 2019-07-01
[Bigtable] Fixes Query#toProto state change effect

### LINKED ISSUE #5550: Unable to use Bigtable Query object as cache key

rahulKQL:
Fixes #5550

This change address the hashcode change problem of Query after calling `Query#toProto`.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
@@ -291,12 +291,15 @@ public final class Query implements Serializable {
     }
     Query query = (Query) o;
     return Objects.equal(tableId, query.tableId)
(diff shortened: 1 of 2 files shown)


rahulKQL:
@igorbernstein2, @sduskis Please have a look.

PR #5624, opened 2019-07-01
Pubsub ordering keys

kamalaboulhosn:
Copy ordering keys code to master branch. The performance regression for the non-ordering case has been fixed.

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -315,7 +315,6 @@ public class Publisher {
         if (!batch.isEmpty()) {
           if (!batch.orderingKey.isEmpty()) {
             // For messages with an ordering key, we need to publish with messagesBatchLock held in
-            // order
             // order to ensure another publish doesn't slip in and send a batch before these batches
             // we already want to send.
             publishOutstandingBatch(batch.popOutstandingBatch());


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -85,15 +88,17 @@
   private final String topicName;
 
   private final BatchingSettings batchingSettings;
+  private final boolean enableMessageOrdering;
 
   private final Lock messagesBatchLock;
-  private MessagesBatch messagesBatch;
+  final Map<String, MessagesBatch> messagesBatches;
```
sduskis: This variable should be private.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -85,15 +88,17 @@
...
 
   private final AtomicBoolean activeAlarm;
 
   private final PublisherStub publisherStub;
 
   private final ScheduledExecutorService executor;
+  final SequentialExecutorService.CallbackExecutor sequentialExecutor;
```
sduskis: This variable should be privte

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -194,38 +210,72 @@ public String getTopicNameString() {
...
     messagesBatchLock.lock();
     try {
+      MessagesBatch messagesBatch = messagesBatches.get(orderingKey);
+      if (messagesBatch == null) {
+        messagesBatch = new MessagesBatch(batchingSettings, orderingKey);
+        messagesBatches.put(orderingKey, messagesBatch);
+      }
+
       batchesToSend = messagesBatch.add(outstandingPublish);
-      // Setup the next duration based delivery alarm if there are messages batched.
```
sduskis: This comment should be restored by `setupAlarm()`

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -194,38 +210,72 @@ public String getTopicNameString() {
...
     } finally {
       messagesBatchLock.unlock();
     }
 
     messagesWaiter.incrementPendingMessages(1);
 
-    if (!batchesToSend.isEmpty()) {
+    // For messages without ordering keys, it is okay to send batches without holding
+    // messagesBatchLock.
+    if (!batchesToSend.isEmpty() && orderingKey.isEmpty()) {
```
sduskis: If ordering keys is turned on, should `publishAllWithoutInflight()` still happen?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -257,16 +307,53 @@ public void run() {
...
     try {
-      if (messagesBatch.isEmpty()) {
-        return;
+      for (MessagesBatch batch : messagesBatches.values()) {
+        if (!batch.isEmpty()) {
+          // TODO(kimkyung-goog): Do not release `messagesBatchLock` when publishing a batch. If
+          // it's released, the order of publishing cannot be guaranteed if `publish()` is called
+          // while this function is running. This locking mechanism needs to be improved if it
+          // causes any performance degradation.
+          publishOutstandingBatch(batch.popOutstandingBatch());
```
sduskis: Should this happen for non ordering keys scenarios?

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
kamalaboulhosn: Looks like we shouldn't call it, no.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
kamalaboulhosn: I moved it out from under the lock. We should do the work, but it doesn't need to be done while the lock is held.

kamalaboulhosn:
@sduskis I have removed the test that was problematic. It was the one that had been using the now-private elements of the class and so I think it is too dependent on the internal details to be able to consistently test something meaningful. Do you want to complete this review or do you want me to hand it off to someone else? Thanks!

sduskis:
@kamalaboulhosn, please find someone else.

google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -256,17 +308,62 @@ public void run() {
...
    */
   public void publishAllOutstanding() {
+    OutstandingBatch unorderedOutstandingBatch = null;
+    messagesBatchLock.lock();
+    try {
+      for (MessagesBatch batch : messagesBatches.values()) {
+        if (!batch.isEmpty()) {
+          if (!batch.orderingKey.isEmpty()) {
+            // For messages with an ordering key, we need to publish with messagesBatchLock held in
+            // order
```
manuelmenzella-google: Nit: I think this break is unintentional.

PR #5628, opened 2019-07-01
Add Tasks smoke test

chingor13:
This test is adapted from java-docs-samples.


google-cloud-clients/google-cloud-tasks/src/test/java/com/google/cloud/tasks/v2beta3/CloudTasksSmokeTest.java
```diff
@@ -0,0 +1,112 @@
...
+        queue = client.getQueue(queueName);
+      }
+    } catch (IOException e) {
+      fail("Failed to create queue: " + e.getMessage());
+    }
+  }
+
+  @Test
+  public void run() {
+    main(null);
```
kolea2: minor - want to inline this?

chingor13: I followed the existing format for the generated SmokeTests ([example](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-trace/src/test/java/com/google/cloud/trace/v2/TraceServiceSmokeTest.java))

PR #5645, opened 2019-07-02
Logging : Cleanup dependency

athakor:
towards #4383


google-cloud-clients/google-cloud-logging/pom.xml
```diff
@@ -59,21 +39,11 @@
...
-      <groupId>com.google.api.grpc</groupId>
-      <artifactId>proto-google-common-protos</artifactId>
-      <scope>compile</scope>
-    </dependency>
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-cloud-logging-v2</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
elharo: This one is very surprising. Do we really not have any unit tests for this code? Maybe they're somewhere else?

elharo: defintiely something weird: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-logging/src/test/java/com/google/cloud/logging/BaseSystemTest.java is a junit test that seems to be controlled by this pom.xml.

Can someone explain what's going on here? Are the tests not run? Is the dependency pulled in from a parent?

PR #5647, opened 2019-07-02
Spanner: Min backoff should be 250ms

### LINKED ISSUE #5544: Spanner: Adjust timeouts for CreateDatabase and retries for other methods
As part of a recent change to GAPIC Configuration, timeouts and retries need to be updated. 

Ensure that the timeouts and retires specified in https://github.com/googleapis/googleapis/commit/cc233544aa39b8947c9b929819aeb08e2cb71feb#diff-4501db3e3507bcce8496c9aea2017d96 are reflected in this client library.

olavloite:
Fixes #5544

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionNotFoundException.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionNotFoundException.java
@@ -26,6 +26,11 @@ import javax.annotation.Nullable;
 public class SessionNotFoundException extends SpannerException {
   private static final long serialVersionUID = -6395746612598975751L;
 
+  static boolean isSessionNotFoundMessage(String message) {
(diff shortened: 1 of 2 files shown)


kolea2:
Hmm.. I thought the 250 ms was picked up by this PR: https://github.com/googleapis/google-cloud-java/pull/5523. Is this still needed?

olavloite:
The 250ms is picked up for all timeouts for normal gRPC calls. It is not automatically picked up for the custom retry logic in the Spanner library that is still used for:
* Retrying aborted transactions
* Retrying broken streaming result sets

kolea2:
Apologies, just read through the issue. I think this should be picked up automatically and we shouldn't rely on hardcoded constants - let's go back to the issue and figure out where that should go to work as expected.

Separately, can we move the session not found commit to a new PR?

olavloite:
I'll move the `SessionNotFound` issue into a separate PR. Closing this to continue the discussion in the original issue.

PR #5655, opened 2019-07-02
Add Kokoro test config for running dependency:analyze

chingor13:
We won't turn on the CI until we fix the dependencies.


chingor13:
We'll get to this after the repo split

PR #5658, opened 2019-07-03
Update resumption strategy to use format-independent row count.

kmjung:
This change modifies the ReadRowsResumptionStrategy helper class in the BigQuery
storage client to use the new format-independent row count value in the
ReadRowsResponse message in order to track stream position. It also modifies
various test files to use the new row count value.


kmjung:
cc: @tswast

PR #5692, opened 2019-07-04
Spanner: PDML should have its own timeout settings with a default of 2 hours

olavloite:
Partitioned DML statements could be long running and should have its own timeout settings with a higher default value.


olavloite:
This will be solved by creating an rpc call with a custom client context.

PR #5698, opened 2019-07-08
Spanner: PDML should have its own timeout settings with a default of 2 hours and share resources with the base stub

olavloite:


olavloite:
This will be solved by creating an rpc call with a custom client context.

PR #5703, opened 2019-07-08
Spanner: Add trace+timeout for waiting for sessions

olavloite:
Adds a trace for waiting for a session from the session pool. The maximum wait time is default 30 seconds, and this time can be changed through SessionPoolOptions#setBlockForSessionTimeoutMillis(long).

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
@@ -612,7 +612,7 @@ public class SessionPoolTest extends BaseSessionPoolTest {
         SessionPoolOptions.newBuilder()
             .setMinSessions(minSessions)
             .setMaxSessions(1)
-            .setInitialWaitForSessionTimeoutMillis(20L)
+            .setInitialWaitForSessionTimeoutMillis(40L)
             .build();
     SessionImpl mockSession = mockSession();
     when(client.createSession(db)).thenReturn(mockSession);
@@ -642,7 +642,7 @@ public class SessionPoolTest extends BaseSessionPoolTest {
                 }
               });
       try {
-        fut.get(40L, TimeUnit.MILLISECONDS);
+        fut.get(80L, TimeUnit.MILLISECONDS);
         fail("missing expected timeout exception");
       } catch (TimeoutException e) {
         // This is the expected exception, just ignore it.


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -822,12 +830,44 @@ private void put(SpannerException e) {
     }
 
     private PooledSession take() throws SpannerException {
-      SessionOrError s = Uninterruptibles.takeUninterruptibly(waiter);
+      SessionOrError s = takeUninterruptiblyWithTimeout();
```
kolea2: how come this needs to change?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
```diff
@@ -39,6 +40,7 @@ private SessionPoolOptions(Builder builder) {
     this.writeSessionsFraction = builder.writeSessionsFraction;
     this.actionOnExhaustion = builder.actionOnExhaustion;
     this.actionOnSessionNotFound = builder.actionOnSessionNotFound;
+    this.waitForSessionTimeoutMillis = builder.blockForSessionTimeoutMillis;
```
kolea2: can we make these variables all 'wait...' or all 'block..'

olavloite: Changed to 'wait...'.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: The `Uninterruptibles.takeUninterruptibly(waiter)` will block indefinitely if the resource pool is exhausted because of a session leak. In order to be able to log that a session request is blocked, we need a trigger that will end the trace span and do the actual logging. That is what this timeout does.

This timeout is currently also propagated to the user as a DEADLINE_EXCEEDED `SpannerException`. We could choose to handle this internally and retry the session request after logging the timeout. That would retain the current behavior where the application will potentially block indefinitely in case of a session leak. My opinion is however that it would be better to propagate this error to the user in order to give the user more information on what is going on.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -822,11 +832,49 @@ private void put(SpannerException e) {
     }
 
     private PooledSession take() throws SpannerException {
-      SessionOrError s = Uninterruptibles.takeUninterruptibly(waiter);
-      if (s.e != null) {
-        throw newSpannerException(s.e);
+      long currentTimeout = options.getInitialWaitForSessionTimeoutMillis();
+      while (true) {
+        Tracer tracer = Tracing.getTracer();
```
songy23: Nit: `tracer` can be made a private static field.

google-cloud-clients/google-cloud-spanner/pom.xml
```diff
@@ -145,6 +145,12 @@
       <classifier>testlib</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-impl</artifactId>
+      <version>${opencensus.version}</version>
+      <scope>test</scope>
```
songy23: This should be runtime only.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -822,11 +832,49 @@ private void put(SpannerException e) {
...
+        Scope waitScope = tracer.spanBuilder(WAIT_FOR_SESSION).startScopedSpan();
+        try {
+          SessionOrError s = pollUninterruptiblyWithTimeout(currentTimeout);
+          if (s == null) {
+            // End the span with status DEADLINE_EXCEEDED and retry.
+            tracer
+                .getCurrentSpan()
+                .end(EndSpanOptions.builder().setStatus(Status.DEADLINE_EXCEEDED).build());
```
songy23:
```suggestion
                .setStatus(Status.DEADLINE_EXCEEDED);
```

You don't need to end the span here. When scope is closed, span will be ended automatically. See https://github.com/census-instrumentation/opencensus-java/blob/916adcbaba580cb223cb37ca34cf1f902903ede9/api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java#L97-L103.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -822,11 +832,49 @@ private void put(SpannerException e) {
...
+            currentTimeout = Math.min(currentTimeout * 2, MAX_SESSION_WAIT_TIMEOUT);
+          } else {
+            if (s.e != null) {
+              throw newSpannerException(s.e);
+            }
+            tracer.getCurrentSpan().end();
```
songy23: Same here. You don't need to call this explicitly.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -822,11 +832,49 @@ private void put(SpannerException e) {
     }
 
     private PooledSession take() throws SpannerException {
-      SessionOrError s = Uninterruptibles.takeUninterruptibly(waiter);
-      if (s.e != null) {
-        throw newSpannerException(s.e);
+      long currentTimeout = options.getInitialWaitForSessionTimeoutMillis();
+      while (true) {
+        Tracer tracer = Tracing.getTracer();
+        Scope waitScope = tracer.spanBuilder(WAIT_FOR_SESSION).startScopedSpan();
```
skuruppu: The standard way to create a scoped span is in a try block and then you don't explicitly have to close the scope [[example](https://opencensus.io/quickstart/java/tracing/#create-a-span)].

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -822,11 +832,49 @@ private void put(SpannerException e) {
...
+        try {
+          SessionOrError s = pollUninterruptiblyWithTimeout(currentTimeout);
+          if (s == null) {
+            // End the span with status DEADLINE_EXCEEDED and retry.
+            tracer
```
skuruppu: Instead of creating a span each time pollUninterruptiblyWithTimeout() is called, an alternative is to create a single span in take() then each time around the loop, add a trace annotation to indicate that the operation timed out. You can also add attributes to the annotation and which gives you the option to record the current timeout at each failure [[example](https://opencensus.io/quickstart/java/tracing/#create-a-child-span)].

There might be a good reason why you chose to do multiple spans and I'd like to understand it.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -822,11 +832,49 @@ private void put(SpannerException e) {
...
+                .getCurrentSpan()
```
skuruppu: [nit] Here and below, instead of calling  tracer.getCurrentSpan() repeatedly, you can save it in a variable above all the if/else logic.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -564,6 +579,89 @@ public void keepAlive() throws Exception {
...
+      return;
+    }
+    // Setup a dummy trace handler.
+    final AtomicInteger deadlineExceededCount = new AtomicInteger();
+    Handler handler =
+        new Handler() {
+          @Override
+          public void export(Collection<SpanData> spanDataList) {
+            for (SpanData sd : spanDataList) {
+              if (sd.getStatus() == Status.DEADLINE_EXCEEDED) {
```
skuruppu: Should this counter only be incremented for spans that match WAIT_FOR_SESSION?

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -564,6 +579,89 @@ public void keepAlive() throws Exception {
...
+              });
+      try {
+        fut.get(40L, TimeUnit.MILLISECONDS);
+        fail("missing expected timeout exception");
+      } catch (TimeoutException e) {
+        // Return the checked out session to the pool so the async request will get a session and
+        // finish.
+        checkedOutSession.close();
+        fut.get();
+        executor.shutdown();
```
skuruppu: close() and shutdown() should be done in a finally as we should be cleaning up regardless of whether the timeout happened or not.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Changed.

google-cloud-clients/google-cloud-spanner/pom.xml
olavloite: Changed to runtime.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Changed.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Changed

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Changed to try-with-resources block.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: The reason that I chose to create multiple spans, is that the `take()` method may block indefinitely, which would mean that the span would never be ended. Does a span that never ends, but that does add a trace annotation every 30 seconds, get flushed and sent to the server each time an annotation is added? If so, then I agree that having a single span in the `take()` would be the better solution. But to my (limited) knowledge of OpenCensus, the span needs to be closed before it is flushed.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: After the other changes after the review, the number of calls to `tracer.getCurrentSpan()` is now reduced to two, and both are only called if an exception occurs, so I would prefer to keep these two calls. But I agree with you that in the previous version of the code, that would have been better.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
olavloite: Agree, that makes the test case less prone to false failures.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
olavloite: Changed.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
skuruppu: I checked with @g-easy and you're right. The span will never be flushed if take() can block indefinitely so your implementation is the only way to get the trace information in this case.

olavloite:
@kolea2 Apparently, GitHub requires your explicit approval of this PR to allow merging :-)

PR #5709, opened 2019-07-09
Spanner: Add specific timeout for Partitioned DML with default of 2 hours

olavloite:
Partitioned DML statements should use a different timeout value than the default timeout value used for normal queries and update statements, as these could be long running. This PR adds a default timeout of 2 hours for Partitioned DML statements and an extra configuration option for setting the timeout.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -43,6 +43,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Map;
 import java.util.Set;
+import org.threeten.bp.Duration;
 
 /** Options for the Cloud Spanner service. */
 public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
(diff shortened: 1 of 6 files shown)


google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/DatabaseClientImplTest.java
```diff
@@ -158,4 +167,89 @@ public void testExecutePartitionedDmlWithException() {
...
+            .setMaxAttempts(1)
+            .setTotalTimeout(Duration.ofMillis(1L))
+            .build();
+    SpannerOptions.Builder builder =
+        SpannerOptions.newBuilder()
+            .setProjectId("[PROJECT]")
+            .setChannelProvider(channelProvider)
+            .setCredentials(NoCredentials.getInstance());
+    // Set normal DML timeout value.
+    builder.getSpannerStubSettingsBuilder().executeSqlSettings().setRetrySettings(retrySettings);
```
kolea2: Is it possible to add an assertion that the PDML timeout is 2 hrs?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -114,6 +116,7 @@ private SpannerOptions(Builder builder) {
     } catch (IOException e) {
       throw SpannerExceptionFactory.newSpannerException(e);
     }
+    partitionedDmlTimeout = builder.partitionedDmlTimeout;
```
kolea2: very minor - I would move this above the try-catch, after sessionLabels.

PR #5725, opened 2019-07-11
Storage: Fix locationType field is added in bucket

### LINKED ISSUE #5442: Storage: Support bucket.locationType
Location type is a new read-only field for buckets. This corresponds to internal bug 130187927.

Location type will be a new field added for Bucket resource. It will be populated on-the-fly for output path, and is a read-only field so patch/update/create will not have any effect on bucket resource.
Example

Google Cloud Storage Python client library feature request: Add location type on bucket
```
// Get bucket
storage_client = storage.Client()
bucket = storage_client.get_bucket("bucket-name")

// Get location type
location_type = bucket.location_type

// Request:
// GET https://www.googleapis.com/storage/v1/b/bucket-name?fields=locationType
// Response Body
// {
//   "locationType": "multi-region"
// }
```

abhinav-qlogic:
Fixes #5442 

Due to `LocationType` is a read-only field, So Have not added test case specific for `dual-region` or `region` locationType.

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
@@ -662,7 +662,7 @@ public class Bucket extends BucketInfo {
     }
 
     @Override
-    public Builder setLocationType(String locationType) {
+    Builder setLocationType(String locationType) {
       infoBuilder.setLocationType(locationType);
       return this;
     }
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1127,6 +1132,12 @@ public Builder setIamConfiguration(IamConfiguration iamConfiguration) {
       return this;
     }
 
+    @Override
+    public Builder setLocationType(String locationType) {
```
frankyn: I think this should be a `private setLocationType` due to read-only.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -661,6 +661,12 @@ public Builder setIamConfiguration(IamConfiguration iamConfiguration) {
       return this;
     }
 
+    @Override
+    public Builder setLocationType(String locationType) {
```
frankyn: Private here as well.

frankyn:
I

PR #5732, opened 2019-07-12
BigQuery: update signature for ML model access to label/feature columns.

shollyman:
This is a breaking change to a @BetaApi within BigQuery.

It swaps in a new veneer StandardSQLField type in place of the underlying
discovery type for getting information about the feature and label columns
from ML models, which was functionality introduced in:
https://github.com/googleapis/google-cloud-java/pull/5214

### DIFF
--- a/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Model.java
+++ b/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Model.java
@@ -18,7 +18,6 @@ package com.google.cloud.bigquery;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.api.services.bigquery.model.StandardSqlField;
 import com.google.api.services.bigquery.model.TrainingRun;
 import com.google.cloud.bigquery.BigQuery.ModelOption;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


shollyman:
> LGTM. No tests to update? I guess I didn't really check for all properties on the Models API stats, either.

The tests accessed the data transitively without declaring the intermediate types.  The types changed but the getter names didn't.  Ex:

```
 assertEquals(model.getFeatureColumns().get(0).getName(), "f1");
 assertEquals(model.getLabelColumns().get(0).getName(), "predicted_label");
```

google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/Model.java
```diff
@@ -116,13 +115,13 @@ Builder setTrainingRuns(List<TrainingRun> trainingRunList) {
     }
 
     @Override
-    Builder setLabelColumns(List<StandardSqlField> labelColumnList) {
+    Builder setLabelColumns(List<StandardSQLField> labelColumnList) {
```
kolea2: Where is this annotated with `@BetaApi`? Just want to make sure we don't introduce a breaking change to something that wasn't labeled as such.

shollyman: The feature represents readonly properties provided by the BQ backend, so I marked it as BetaApi on the public getters in ModelInfo.  The internal builder was not annotated this way.

kolea2: Ah, sorry didn't realize this wasn't a public method. SGTM.

PR #5746, opened 2019-07-15
Add debugging information for Firestore integration tests

BenWhitehead:
Add an assertion to verify the expected number of events are delivered
to the queryWatch test. Awaiting the semaphore now has a 60sec timeout.
If not satisfied within that timespan the test will fail, and provide
an assertion message of the progress.

Updates to allow running builds locally
* Update build script to cd to checkout directory relative to itself
  instead of using a hard-coded path.
* Update build script to be tolerant of absolute paths specified for
  GOOGLE_APPLICATION_CREDENTIALS

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -60,7 +60,7 @@ mvn install -B -V \
   -Dgcloud.download.skip=true \
   -T 1C
 
-# prepend Kokoro root directory onto GOOGLE_APPLICATION_CREDENTIALS path
+# if GOOGLE_APPLICATION_CREDIENTIALS is specified as a relative path prepend Kokoro root directory onto it
 if [[ ! -z "${GOOGLE_APPLICATION_CREDENTIALS}" && "${GOOGLE_APPLICATION_CREDENTIALS}" != /* ]]; then
     export GOOGLE_APPLICATION_CREDENTIALS=$(realpath ${KOKORO_ROOT}/src/${GOOGLE_APPLICATION_CREDENTIALS})
 fi


BenWhitehead:
An example of the new diagnostic assertion:
```
java.lang.AssertionError:
did not receive all expected messages within the deadline.
expectedOperations = [case 0, case 1, case 2, case 3, case 4, case 5]
  actualOperations = [case 0]
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.google.cloud.firestore.it.ITSystemTest.queryWatch(ITSystemTest.java:968)
```

kolea2:
So does this mean it failed on case 0, or finished case 0 and failed in case 1? I would add clarifying language in the error message to demonstrate this.

BenWhitehead:
> So does this mean it failed on case 0, or finished case 0 and failed in case 1? I would add clarifying language in the error message to demonstrate this.

It's saying that it expected 6 cases to be performed, but only 1 of them was performed. The cases reported in actual are what actually happened. The main thing is that the number of expected async operations didn't happen during the test run. What would you expect as the reported message?

BenWhitehead:
@kolea2 Updated message per our offline conversation.

BenWhitehead:
@chingor13 Can you take a look at my changes to `.kokoro/build.sh` and give it a +1. All of the builds are running in this PR so I think everything should be working well.

.kokoro/build.sh
```diff
@@ -15,7 +15,10 @@
 
 set -eo pipefail
 
-cd github/google-cloud-java/
+## Get the directory of the build script
+scriptDir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
chingor13: Can this be:

```suggestion
scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
```

.kokoro/build.sh
```diff
@@ -58,7 +61,7 @@ mvn install -B -V \
   -T 1C
 
 # prepend Kokoro root directory onto GOOGLE_APPLICATION_CREDENTIALS path
-if [[ ! -z "${GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
+if [[ ! -z "${GOOGLE_APPLICATION_CREDENTIALS}" && "${GOOGLE_APPLICATION_CREDENTIALS}" != /* ]]; then
```
chingor13: I don't know what this is trying to do?

BenWhitehead: When running my test locally the credentials file isn't in the kokoro directory, and I pass a fully qualified path as the env var. This additional condition ensures the value for `GOOGLE_APPLICATION_CREDIENTIALS` doesn't have a leading `/` before it prepends the kokoro dir to it.

chingor13: Ok, let's add to the comment above something like "if set and is a relative path"

BenWhitehead: Done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -898,39 +909,46 @@ public void queryWatch() throws Exception {
                     @Override
                     public void onEvent(
                         @Nullable QuerySnapshot value, @Nullable FirestoreException error) {
+                      System.out.printf("onEvent(value : %s, error : %s)%n", value, error);
```
chingor13: nit: is this intentionally left here?

BenWhitehead: Yes, I'm wanting to see the events flowing in vs the switch branches below being hit. I haven't yet ruled out if there is some sort of race happening on the events being delivered.

PR #5759, opened 2019-07-17
Expose FullEntity::fromPb and Value::fromPb

### LINKED ISSUE #1431: BaseEntity.toPb method is not public
There are cases where users want to convert com.google.cloud.datastore.Entity objects to native protobuf Entity objects (com.google.datastore.v1.Entity). For ex: while using Cloud Dataflow DatastoreIO, which only works with protobuf Entities. It would be nice to expose the toPb method by making it public.

topecongiro:
Fixes #1431 (though the issue is currently closed).

### DIFF
--- a/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/FullEntity.java
+++ b/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/FullEntity.java
@@ -62,7 +62,7 @@ public class FullEntity<K extends IncompleteKey> extends BaseEntity<K> {
     return new Builder<>(copyFrom);
   }
 
-  static FullEntity<?> fromPb(com.google.datastore.v1.Entity entityPb) {
+  public static FullEntity<?> fromPb(com.google.datastore.v1.Entity entityPb) {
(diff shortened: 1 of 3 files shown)


google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/FullEntity.java
```diff
@@ -62,7 +62,7 @@
     return new Builder<>(copyFrom);
   }
 
-  static FullEntity<?> fromPb(com.google.datastore.v1.Entity entityPb) {
+  public static FullEntity<?> fromPb(com.google.datastore.v1.Entity entityPb) {
```
chingor13: This isn't compiling. `Entity` extends `FullEntity` and it's `fromPb` method is still package private.

topecongiro: Thank you for the review. Fixed it and rebased in https://github.com/googleapis/google-cloud-java/pull/5759/commits/e7146087c4715368caff098bc6f309c574fd2b7e.

PR #5789, opened 2019-07-19
Storage-nio: Only com.google.cloud.storage.** should be excluded from shading

### LINKED ISSUE #5127: Incomplete shading in google-cloud-nio leads to downstream breakage
The shaded release of google-cloud-nio (we are currently on release `0.81.0-alpha:shaded`) does not shade all of its dependencies. In particular, classes in `google-cloud-core-*` (such as `google-cloud-core-http`) are not being shaded. 

As a result, we run into problems such as this one: `com.google.cloud.http.HttpTransportOptions` is not itself shaded, but its method `getHttpTransportFactory()` returns an instance of a shaded class `shaded.cloud_nio.com.google.auth.http.HttpTransportFactory`, which calling code from outside the library cannot handle. This results in the error below:

```
Exception in thread "main" java.lang.NoSuchMethodError: com.google.cloud.http.HttpTransportOptions.getHttpTransportFactory()Lcom/google/auth/http/HttpTransportFactory;
	at com.google.cloud.bigquery.spi.v2.HttpBigQueryRpc.<init>(HttpBigQueryRpc.java:89)
	at com.google.cloud.bigquery.BigQueryOptions$DefaultBigQueryRpcFactory.create(BigQueryOptions.java:56)
	at com.google.cloud.bigquery.BigQueryOptions$DefaultBigQueryRpcFactory.create(BigQueryOptions.java:50)
	at com.google.cloud.ServiceOptions.getRpc(ServiceOptions.java:510)
	at com.google.cloud.bigquery.BigQueryOptions.getBigQueryRpcV2(BigQueryOptions.java:123)
	at com.google.cloud.bigquery.BigQueryImpl.<init>(BigQueryImpl.java:150)
	at com.google.cloud.bigquery.BigQueryOptions$DefaultBigQueryFactory.create(BigQueryOptions.java:46)
	at com.google.cloud.bigquery.BigQueryOptions$DefaultBigQueryFactory.create(BigQueryOptions.java:40)
	at com.google.cloud.ServiceOptions.getService(ServiceOptions.java:498)
	at org.broadinstitute.hellbender.tools.examples.ExampleBigQueryReader.traverse(ExampleBigQueryReader.java:108)
	at org.broadinstitute.hellbender.engine.GATKTool.doWork(GATKTool.java:984)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:139)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:191)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:210)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:162)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:205)
```

Would it be possible to shade these classes from `google-cloud-core-*` in the shaded `google-cloud-nio` jar?

olavloite:
There's no automatic test for this, but I did the following manual testing:
1. Before the change: Include the shaded nio library as a dependency along with the google-cloud-spanner library. It is then possible to use the nio library, but any call to Spanner fails with a `NoSuchMethodError`.
1. After the change: Include the shaded nio library as a dependency along with the google-cloud-spanner library. It is now possible to use both nio as Spanner without any conflicts.

Fixes #5127

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-nio/pom.xml
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-nio/pom.xml
@@ -96,7 +96,7 @@
                 <include>com.fasterxml.**</include>
               </includes>
               <excludes>
-                <exclude>com.google.cloud.**</exclude>
+                <exclude>com.google.cloud.storage.**</exclude>
                 <exclude>com.google.auto.**</exclude>
               </excludes>
             </relocation>


chingor13:
cc @jean-philippe-martin

jean-philippe-martin:
I need to look at this in more detail, but to recap just to make sure I got it right:

* The original bug is that NIO code passes an `com.google.cloud.http.HttpTransportOptions` around (in ways unspecified), and this object has public getter that returns a shaded `shaded.cloud_nio.com.google.auth.http.HttpTransportFactory`. Other code that looks at the `HttpTransportOptions` don't expect that type and crash
* The original bug description suggests that shading more may be a solution
* This fix does shade more (by narrowing the exception list), so the `HttpTransportOptions` would be shaded too.

This fix also points out a new failure mode, where including both Spanner and NIO causes Spanner to break.

olavloite:
@jean-philippe-martin 
This problem is triggered when you use the shaded version of google-cloud-storage-nio in combination with other Google Cloud libraries. The original bug originates from a user that uses google-cloud-storage-nio in combination with Big Query (see the included stack trace). I tested it with Spanner, but it will probably break with most (all?) Google Cloud libraries, as the shaded nio library does not shade its core com.google.cloud.* dependencies, but does include them in the shaded .jar. So the result is that you end up with the core Google Cloud libraries twice on the classpath.

jean-philippe-martin:
This looks good to me. With this change in, I was able to load BigQuery and Spanner along with the NIO library and initialize them successfully.

:+1:

olavloite:
@JesseLovelace Friendly ping. Do you have time to have a look at this?

PR #5794, opened 2019-07-19
Bigtable: update retry & timeout settings

igorbernstein2:
Regenerate the client using the updated gapic settings:
https://github.com/googleapis/googleapis/commit/184ab77f4cee62332f8f9a689c70c9bea441f836

And the updated gapic-generator:
https://github.com/googleapis/gapic-generator/pull/2878

Update the javadocs for the new settings

Re-enable the dropRowRange integration test now that the settings are correct


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
```diff
@@ -503,9 +503,7 @@ public String extractNextToken(ListAppProfilesResponse payload) {
           ImmutableSet.copyOf(
               Lists.<StatusCode.Code>newArrayList(
                   StatusCode.Code.DEADLINE_EXCEEDED, StatusCode.Code.UNAVAILABLE)));
-      definitions.put(
-          "non_idempotent",
-          ImmutableSet.copyOf(Lists.<StatusCode.Code>newArrayList(StatusCode.Code.UNAVAILABLE)));
+      definitions.put("non_idempotent", ImmutableSet.copyOf(Lists.<StatusCode.Code>newArrayList()));
```
elharo: Lists.newArrayList is no longer needed. Just use new ArrayList<>(). Also easier to read if there's an extra local variable for this.

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
```diff
@@ -516,15 +514,37 @@ public String extractNextToken(ListAppProfilesResponse payload) {
...
               .setRetryDelayMultiplier(2.0)
               .setMaxRetryDelay(Duration.ofMillis(60000L))
               .setInitialRpcTimeout(Duration.ofMillis(60000L))
               .setRpcTimeoutMultiplier(1.0)
               .setMaxRpcTimeout(Duration.ofMillis(60000L))
               .setTotalTimeout(Duration.ofMillis(600000L))
               .build();
-      definitions.put("default", settings);
+      definitions.put("idempotent_params", settings);
+      settings =
```
elharo: can we avoid reusing local variables here as it's error prone?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
```diff
@@ -516,15 +514,37 @@ public String extractNextToken(ListAppProfilesResponse payload) {
       RetrySettings settings = null;
       settings =
           RetrySettings.newBuilder()
-              .setInitialRetryDelay(Duration.ofMillis(5L))
+              .setInitialRetryDelay(Duration.ofMillis(1000L))
```
elharo: Duration.ofSeconds(1) is easier to read

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
```diff
@@ -516,15 +514,37 @@ public String extractNextToken(ListAppProfilesResponse payload) {
...
               .setTotalTimeout(Duration.ofMillis(600000L))
               .build();
-      definitions.put("default", settings);
+      definitions.put("idempotent_params", settings);
+      settings =
+          RetrySettings.newBuilder()
+              .setInitialRetryDelay(Duration.ofMillis(0L))
+              .setRetryDelayMultiplier(1.0)
+              .setMaxRetryDelay(Duration.ofMillis(0L))
+              .setInitialRpcTimeout(Duration.ofMillis(60000L))
```
elharo: Rather than very long values where I have to count zeroes to see how long it is, how about Duration.ofMinutes(1) and the like?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
```diff
@@ -749,15 +769,15 @@ private static Builder initDefaults(Builder builder) {
                       .setInitialRpcTimeout(Duration.ZERO) // ignored
                       .setRpcTimeoutMultiplier(1.0) // ignored
                       .setMaxRpcTimeout(Duration.ZERO) // ignored
-                      .setTotalTimeout(Duration.ofMillis(300000L))
+                      .setTotalTimeout(Duration.ofMillis(600000L))
```
elharo: Duration.ofMinutes(10)

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableStubSettings.java
```diff
@@ -226,26 +226,48 @@ protected BigtableStubSettings(Builder settingsBuilder) throws IOException {
...
+      definitions.put("non_idempotent_params", settings);
+      settings =
+          RetrySettings.newBuilder()
+              .setInitialRetryDelay(Duration.ofMillis(10L))
+              .setRetryDelayMultiplier(2.0)
+              .setMaxRetryDelay(Duration.ofMillis(60000L))
+              .setInitialRpcTimeout(Duration.ofMillis(300000L))
+              .setRpcTimeoutMultiplier(1.0)
+              .setMaxRpcTimeout(Duration.ofMillis(300000L))
+              .setTotalTimeout(Duration.ofMillis(43200000L))
```
elharo: This is a really magic number that could use a comment explaining why 432 and not 431 or 433?

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
igorbernstein2: This code is autogenerated by gapic-generator. So it would have to change in its templates

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
igorbernstein2: This code is autogenerated by gapic-generator. So it would have to change in its templates

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
igorbernstein2: This code is autogenerated by gapic-generator. So it would have to change in its templates

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
igorbernstein2: This code is autogenerated by gapic-generator. So it would have to change in its templates

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableStubSettings.java
igorbernstein2: This code is autogenerated by gapic-generator. So I don't have to ability to add comments

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableInstanceAdminStubSettings.java
igorbernstein2: This code is autogenerated by gapic-generator. So it would have to change in its templates

igorbernstein2:
@elharo  thanks for the review, unfortunately I can't address any of the feedback because all of the java is autogenerated and I don't have control over it. PTAL

igorbernstein2:
So I think you disagree with the entire approach that google has taken with client strategy. Across all of our compiled languages we are checking in generated source files, including generated protos:

* [java generated protos](https://github.com/googleapis/google-cloud-java/blob/master/google-api-grpc/grpc-google-cloud-bigtable-v2/src/main/java/com/google/bigtable/v2/BigtableGrpc.java)

* [gapic generated code](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableStub.java)

* [golang genrated protos](https://github.com/googleapis/go-genproto/blob/master/googleapis/bigtable/v2/bigtable.pb.go)

* [dotnet generated protos](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableGrpc.cs)

* [dotnet generated client](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClient.cs)

The same goes for almost every cloud client that google has. All of the code is generated by [gapic](https://github.com/googleapis/gapic-generator).

While that approach does have drawbacks, discussing them here is completely out of scope for this PR or really any PR in this repository. This is an approach that was adopted by the clients team, which if you feel strongly, I encourage you reach out to them. Telling me (a Bigtable developer) that checking in generated code is wrong, is roughly the same as you telling me that my use of generics is incorrect because you disagree with type erasure. If you dislike the general approach, start a discussion at the policy level, not on a PR fixing a bug within that ecosystem.

This PR needs to be merged ASAP because it solves a very real problem: customers can't current invoke one of the api methods (dropRowRange) due to a dependency update that activated a new code path: https://github.com/googleapis/gax-java/pull/745, which is now using a previously dormant setting

PR #5800, opened 2019-07-22
Storage: Bucket metadata sample

JesseLovelace:
Bucket metadata sample and test


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1126,6 +1126,39 @@ public Bucket setDefaultKmsKey(String bucketName, String kmsKeyName) throws Stor
...
+    // Fields can be selected individually e.g. Storage.BucketField.NAME
+    Bucket bucket = storage.get(bucketName, BucketGetOption.fields(Storage.BucketField.values()));
+
+    // Print bucket metadata
+    System.out.println("BucketName: " + bucket.getName());
+    System.out.println("DefaultEventBasedHold: " + bucket.getDefaultEventBasedHold());
+    System.out.println("DefaultKmsKeyName: " + bucket.getDefaultKmsKeyName());
+    System.out.println("Id: " + bucket.getGeneratedId());
+    System.out.println("IndexPage: " + bucket.getIndexPage());
+    System.out.println("Labels: " + bucket.getLabels());
```
frankyn: This is a Map. How will this be handled?

JesseLovelace: It's a Map<String, String>, so the toString function works just fine here, would look like "{k1=v1, k2=v2}"

frankyn: Gotcha, thanks for clarifying. Please handle the map through an iterator so the data type is surfaced.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1157,6 +1152,21 @@ public void getBucketMetadata(String bucketName) throws StorageException {
...
+        Map.Entry<String, String> next = iterator.next();
+        labels.append(next.getKey() + "=" + next.getValue());
+        if (iterator.hasNext()) {
+          labels.append(",");
+        }
+      }
+      labels.append("}");
+      labelsString = labels.toString();
+    }
+    System.out.println("Labels: " + labelsString);
```
frankyn: Could you follow this pattern instead? https://github.com/googleapis/google-cloud-java/blob/0731387f0edd9991dca3483aa2aa5a82ed96e0ac/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java#L1219-L1224

I misspoke, I should have pointed to this example first.

PR #5802, opened 2019-07-22
Fix dependency declarations

chingor13:
Fixes the rest of failures when running `mvn dependency:analyze -DfailOnWarning=true`


chingor13:
We'll get to this after the repo split

PR #5803, opened 2019-07-23
Firestore: Add support for IN Queries

schmidt-sebastian:
This adds support for Firestore's new IN and ARRAY_CONTAINS_ANY queries.

This can not yet be merged:
- The underlying Proto change is not public yet (hence the PR won't compile)
- The backend does not yet support these types of queries (the test won't pass)

This PR is meant to be a reference for other languages. 

Port of https://github.com/googleapis/nodejs-firestore/pull/715

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
@@ -50,7 +50,6 @@ import com.google.protobuf.Int32Value;
 import io.opencensus.trace.AttributeValue;
 import io.opencensus.trace.Tracing;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


schmidt-sebastian:
cc @thebrianchen

schmidt-sebastian:
> Just double checking, do we need any serialization/deserialization tests for the java sdk?
> otherwise, LGTM

We create the Proto directly via the somewhat hidden `whereHelper`. The unit tests then confirm that we created the expected request.

BenWhitehead:
@schmidt-sebastian Can we rebase this and get it fixed up to merge now that the protos are available as of #6479?

schmidt-sebastian:
We can probably launch this as soon as next week. I will rebase the PR on Monday.

BenWhitehead:
@schmidt-sebastian The two new integration tests are failing with assertion errors. Does the integration test project need to be given explicit access for these to pass?

schmidt-sebastian:
> @schmidt-sebastian The two new integration tests are failing with assertion errors. Does the integration test project need to be given explicit access for these to pass?

No, they just have never been run and are buggy. I am updating them right now, and will also fix one incompatibility with Java 7 that we just noticed on Android.

schmidt-sebastian:
I updated the PR, it should now be in a good state. At first, we wanted to add a varargs method, but we noticed that it break when a user calls Query.whereIn(field, null) since Java 7 picks the wrong overload (`@NunNull List<>` is preferred). We decided to punt on this for now and may add the overloads later if requested.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -608,6 +611,89 @@ public Query whereArrayContains(@Nonnull FieldPath fieldPath, @Nonnull Object va
...
+   * @param values A list that contains the values to match.
+   * @return The created Query.
+   */
+  @Nonnull
+  public Query whereArrayContainsAny(@Nonnull String field, @Nonnull List<? extends Object> values) {
+    Preconditions.checkState(
+            options.getStartCursor() == null && options.getEndCursor() == null,
+            "Cannot call whereArrayContainsAny() after defining a boundary with startAt(), "
+                    + "startAfter(), endBefore() or endAt().");
+    return whereHelper(FieldPath.fromDotSeparatedString(field), ARRAY_CONTAINS_ANY, value);
```
BenWhitehead:
```suggestion
    return whereHelper(FieldPath.fromDotSeparatedString(field), ARRAY_CONTAINS_ANY, values);
```

schmidt-sebastian: Embarrassing... Fix applied.

PR #5806, opened 2019-07-23
spanner-jdbc: Step 02 - Client side statement interfaces

olavloite:
The Cloud Spanner jdbc driver supports several client side statements that are interpreted by the driver instead of being sent to Cloud Spanner. These statements are defined in the ClientSideStatements.json file. Whenever a statement is executed by the driver, it checks whether it is a client side statement, and if so, interprets it and passes it into a `ConnectionStatementExecutor` that will call the appropriate method on a `Connection`.


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/resources/com/google/cloud/spanner/jdbc/ClientSideStatements.json
```diff
@@ -0,0 +1,246 @@
...
+	  "prerequisiteStatements": ["set readonly = true", "SELECT 1 AS TEST"]
+	},
+	{
+	  "name": "SHOW VARIABLE COMMIT_TIMESTAMP",
+	  "executorName": "ClientSideStatementNoParamExecutor",
+	  "resultType": "RESULT_SET",
+	  "regex": "(?is)\\A\\s*show\\s+variable\\s+commit_timestamp\\s*\\z",
+	  "method": "statementShowCommitTimestamp",
+	  "exampleStatements": ["show variable commit_timestamp"],
+	  "prerequisiteStatements": ["update foo set bar=1"]
```
kolea2: Just to confirm my understanding, these are returned on `getPrerequisiteStatements` as examples of what needs to be called beforehand? If yes, can we call these `examplePrerequisiteStatements`?

olavloite: Your assumption is correct, and these example prerequisite statements are used by some automated test cases that execute the example statements. I've changed the name of the field and the getter.

PR #5813, opened 2019-07-24
spanner-jdbc: Step 05 - Util for parsing read-only staleness values

olavloite:
A user can specify a read-only staleness value to use for a read-only transaction by executing a statement that is parsed by the client. This util parses the possible staleness values and can translate these back to text values.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadOnlyStalenessUtil.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadOnlyStalenessUtil.java
@@ -35,8 +35,9 @@ import java.util.concurrent.TimeUnit;
 class ReadOnlyStalenessUtil {
   /**
    * Parses an RFC3339 date/time value with millisecond precision and returns this as a {@link
-   * Timestamp}. Although {@link Timestamp} has nanosecond precision, this method is only used for
-   * parsing read-only staleness values, and more than millisecond precision is not needed for that.
+   * Timestamp}.
+   *
+   * @TODO: add support for nanosecond precision.
    */
   static Timestamp parseRfc3339(String str) throws SpannerException {
     try {
@@ -118,10 +119,10 @@ class ReadOnlyStalenessUtil {
     }
   }
 
-  static final class GetMaxStaleness implements DurationValueGetter {
+  static final class MaxStalenessGetter implements DurationValueGetter {
     private final TimestampBound staleness;
 
-    public GetMaxStaleness(TimestampBound staleness) {
+    public MaxStalenessGetter(TimestampBound staleness) {
       this.staleness = staleness;
     }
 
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadOnlyStalenessUtil.java
```diff
@@ -0,0 +1,324 @@
...
+
+/**
+ * Util class for parsing and converting ReadOnlyStaleness values to/from strings. This util is used
+ * to parse client side statements and values for read only staleness for read-only transactions on
+ * Cloud Spanner.
+ */
+class ReadOnlyStalenessUtil {
+  private static final TimeZone GMT = TimeZone.getTimeZone("GMT");
+  /** Regular expression for parsing RFC3339 times with nanosecond precision. */
+  private static final String RFC3339_REGEX =
```
kolea2: There is an RFC parser in the http client. Can we please switch to that? https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/util/DateTime.java

olavloite: The RFC parser in the http client has millisecond precision, both for parsing and converting to string values. Switching to that one is possible, but would mean that we would not be able to support microsecond or nanosecond precision.

kolea2: In that case, let's add the enhanced precision to the http client repository, and use the current http client implementation with a todo for now since this is still in a branch.

olavloite: I've changed the JDBC implementation to use the parser in the http client, and I've added a PR in the http client for adding nanosecond support.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadOnlyStalenessUtil.java
```diff
@@ -0,0 +1,260 @@
...
+      return staleness.getExactStaleness(unit);
+    }
+
+    @Override
+    public boolean hasDuration() {
+      return staleness.getMode() == Mode.EXACT_STALENESS;
+    }
+  }
+
+  static final class GetMaxStaleness implements DurationValueGetter {
```
kolea2: minor - can address now or in future commit - I would rename to "MaxStalenessGetter"

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadOnlyStalenessUtil.java
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * Util class for parsing and converting ReadOnlyStaleness values to/from strings. This util is used
+ * to parse client side statements and values for read only staleness for read-only transactions on
+ * Cloud Spanner.
+ */
+class ReadOnlyStalenessUtil {
+  /**
+   * Parses an RFC3339 date/time value with millisecond precision and returns this as a {@link
+   * Timestamp}. Although {@link Timestamp} has nanosecond precision, this method is only used for
+   * parsing read-only staleness values, and more than millisecond precision is not needed for that.
```
kolea2: minor can address in upcoming - I would just replace this with a todo for nanosecond precision

PR #5814, opened 2019-07-24
spanner-jdbc: Step 06 - Client side statement parser

olavloite:
Adds the client side statement parser for statements that are interpreted and executed by the client instead of being sent to Cloud Spanner. Note:
* The changes to `ClientSideStatements.json` are not directly related to this change, but the `ClientSideStatementValueConverter` implementations were moved from one file to another by one of the previous changes, and I forgot to change that in the .json file.
* There are more tests for parsing and executing the client side statements, but these rely on the class `ConnectionImpl` that has not yet been added.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/StatementParser.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/StatementParser.java
@@ -22,8 +22,9 @@ import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.jdbc.ClientSideStatementImpl.CompileException;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
-import java.util.Arrays;
-import java.util.List;
+import com.google.common.collect.ImmutableSet;
+import java.util.Collections;
+import java.util.Set;
 
 /**
  * Internal class for the Spanner Connection API.
@@ -159,15 +160,16 @@ class StatementParser {
     }
   }
 
-  private final List<String> ddlStatements = Arrays.asList("CREATE", "DROP", "ALTER");
-  private final List<String> selectStatements = Arrays.asList("SELECT");
-  private final List<String> dmlStatements = Arrays.asList("INSERT", "UPDATE", "DELETE");
-  private final List<ClientSideStatementImpl> statements;
+  private final Set<String> ddlStatements = ImmutableSet.of("CREATE", "DROP", "ALTER");
+  private final Set<String> selectStatements = ImmutableSet.of("SELECT");
+  private final Set<String> dmlStatements = ImmutableSet.of("INSERT", "UPDATE", "DELETE");
+  private final Set<ClientSideStatementImpl> statements;
 
   /** Private constructor for singleton instance. */
   private StatementParser() {
     try {
-      statements = ClientSideStatements.INSTANCE.getCompiledStatements();
+      statements =
+          Collections.unmodifiableSet(ClientSideStatements.INSTANCE.getCompiledStatements());
     } catch (CompileException e) {
       throw new RuntimeException(e);
     }
@@ -253,7 +255,7 @@ class StatementParser {
     return statementStartsWith(sql, dmlStatements);
   }
 
-  private boolean statementStartsWith(String sql, List<String> checkStatements) {
+  private boolean statementStartsWith(String sql, Iterable<String> checkStatements) {
     Preconditions.checkNotNull(sql);
     String[] tokens = sql.split("\\s+", 2);
     if (tokens.length > 0) {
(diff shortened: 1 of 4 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ClientSideStatementValueConverters.java
```diff
@@ -0,0 +1,210 @@
...
+            }
+          default:
+            // fall through to allow the calling method to handle this
+        }
+      }
+      return null;
+    }
+
+    private Mode parseMode(String name) {
+      for (Mode mode : Mode.values()) {
```
kolea2: can we make this a Set lookup instead?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ClientSideStatementValueConverters.java
```diff
@@ -0,0 +1,210 @@
...
+    public AutocommitDmlModeConverter(String allowedValues) {}
+
+    @Override
+    public Class<AutocommitDmlMode> getParameterClass() {
+      return AutocommitDmlMode.class;
+    }
+
+    @Override
+    public AutocommitDmlMode convert(String value) {
+      for (AutocommitDmlMode mode : AutocommitDmlMode.values()) {
```
kolea2: same question for this and the below for loops

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/StatementParser.java
```diff
@@ -0,0 +1,379 @@
...
+  boolean isUpdateStatement(String sql) {
+    return statementStartsWith(sql, dmlStatements);
+  }
+
+  private boolean statementStartsWith(String sql, List<String> checkStatements) {
+    Preconditions.checkNotNull(sql);
+    String[] tokens = sql.split("\\s+", 2);
+    if (tokens.length > 0) {
+      for (String check : checkStatements) {
+        if (tokens[0].equalsIgnoreCase(check)) {
```
kolea2: can this be simplified to `sql.startsWith(check)` (handling upper and lower case)?

olavloite: Not really (at least not with the goal of simplifying). The problems with that is that we need a regular expression check for all types of spaces (space, tab, enter, ...), as any type of space character is allowed as a separator between for example CREATE and TABLE. The nice thing with the `String#split(String, int)` method is that it stops searching after finding the requested matches. `String.startsWith(String)` does not support regular expressions.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ClientSideStatementValueConverters.java
olavloite: Changed to a case-insensitive map.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ClientSideStatementValueConverters.java
olavloite: Changed to a case-insensitive map.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ClientSideStatementImpl.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Override
+    public String getMessage() {
+      return "Could not execute statement " + this.statement.name + " (" + sql + ")";
+    }
+  }
+
+  /** The name of this statement. Used in error and info messages. */
+  private String name;
```
kolea2: minor - can address later - we should clean up the whitespace here between variable declarations and javadoc

olavloite:
> Please rebase

I think your request crossed with my actual rebase :-)

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/StatementParser.java
```diff
@@ -0,0 +1,379 @@
...
+
+    ClientSideStatement getClientSideStatement() {
+      Preconditions.checkState(
+          clientSideStatement != null,
+          "This ParsedStatement does not contain a ClientSideStatement");
+      return clientSideStatement;
+    }
+  }
+
+  private final List<String> ddlStatements = Arrays.asList("CREATE", "DROP", "ALTER");
```
kolea2: minor, can address later - I would change these to sets

PR #5815, opened 2019-07-24
spanner-jdbc: Step 07 - DirectExecuteResultSet

olavloite:
The default Cloud Spanner `ResultSet` implementation defers the execution of a query to the first call to `ResultSet#next()`. The JDBC Driver requires that queries are executed directly when a statement is passed to the driver, and that any query errors are returned directly. The `DirectExecuteResultSet` is a wrapper around a normal result set that ensures that the query is executed directly, and that any query errors are propagated directly when the result set is created, instead of deferring this until the first call to `ResultSet#next()`.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/DirectExecuteResultSetTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/DirectExecuteResultSetTest.java
@@ -18,6 +18,7 @@ package com.google.cloud.spanner.jdbc;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/DirectExecuteResultSetTest.java
```diff
@@ -0,0 +1,242 @@
...
+      List<String> excludedMethods,
+      Class<? extends Exception> expectedException)
+      throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
+    for (Method method : DirectExecuteResultSet.class.getDeclaredMethods()) {
+      if (Modifier.isPublic(method.getModifiers()) && !excludedMethods.contains(method.getName())) {
+        boolean exception = false;
+        try {
+          if (method.getParameterTypes().length == 0) {
+            method.invoke(subject);
+          } else if (method.getParameterTypes().length == 1
```
kolea2: minor - for readability, can we extract some of these out into descriptive variables?

olavloite: Number of parameters and parameter type have been extracted to variables and the if-statement has been partially re-written as a switch statement.

PR #5816, opened 2019-07-24
spanner-jdbc: Step 08 - ReplaceableForwardingResultSet

olavloite:
The JDBC Driver contains an internal retry algorithm for aborted read/write transactions that rely on keeping track of a SHA-256 checksum for all data that is returned by Cloud Spanner. When a transaction is successfully retried, all open result sets must be replaced with a fresh one that is linked with the new transaction. This is achieved by returning result sets to the user that contain a delegate that may be replaced by the JDBC Driver at a later moment.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReplaceableForwardingResultSetTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReplaceableForwardingResultSetTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.spanner.jdbc;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReplaceableForwardingResultSetTest.java
```diff
@@ -0,0 +1,294 @@
+/*
+ * Copyright 2018 Google LLC
```
kolea2: 2019

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReplaceableForwardingResultSetTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+  @Test
+  public void testValidMethodCall()
+      throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
+    ResultSet delegate = mock(ResultSet.class);
+    when(delegate.next()).thenReturn(true, true, false);
+    try (ReplaceableForwardingResultSet subject = new ReplaceableForwardingResultSet(delegate)) {
+      subject.next();
+
+      subject.getBoolean(0);
```
kolea2: I'd make a note somewhere in the above implementation that column indexes start at 0

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReplaceableForwardingResultSetTest.java
olavloite: Changed.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReplaceableForwardingResultSetTest.java
olavloite: I added a note about this in the test case, as I assume that you mean that. In the class itself (`ReplaceableForwardingResultSet`), it is clear from the interface it is implementing that column indices are zero-based.

PR #5847, opened 2019-07-24
spanner-jdbc: Step 09 - StatementResultImpl

olavloite:
Adds implementation for `StatementResult`. `StatementResult` is a wrapper around the result returned by a generic statement that is executed by the JDBC Driver. It can contain:
* A `ResultSet` returned by a query on Cloud Spanner, or a `ResultSet` returned by a client side statement (e.g. `SHOW VARIABLE AUTOCOMMIT`).
* An update count returned by a DML statement on Cloud Spanner. There are no client side statements that return an update count.
* No result: DDL statements on Cloud Spanner do not return any results. `SET <VARIABLE_NAME> = <VALUE>` type client side statements also do not return any results.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/StatementResultImplTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/StatementResultImplTest.java
@@ -22,8 +22,10 @@ import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.mock;
 
+import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.ResultSet;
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/StatementResultImpl.java
```diff
@@ -0,0 +1,187 @@
...
+            Arrays.asList(Struct.newBuilder().set(name).to(value).build())),
+        clientSideStatementType);
+  }
+
+  /**
+   * Convenience method for creating a {@link StatementResult} containing a {@link ResultSet} with
+   * one STRING column containing an {@link Enum} value and one row that is created by a {@link
+   * ClientSideStatement}.
+   */
+  static StatementResult resultSet(
```
kolea2: do we need more tests here for all the different types?

olavloite: I've added test cases for all the missing convenience methods.

PR #5875, opened 2019-07-25
spanner-jdbc: Step 11 - CredentialService

olavloite:
Adds the `CredentialsService` helper class for loading credentials that have been specified in the connection URL. Credential files can be loaded from both a local file as well as Cloud Storage.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
@@ -16,9 +16,6 @@
 
 package com.google.cloud.spanner.jdbc;
 
-import com.google.api.client.http.HttpTransport;
-import com.google.api.client.http.javanet.NetHttpTransport;
-import com.google.auth.http.HttpTransportFactory;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.SpannerException;
@@ -37,24 +34,10 @@ import java.nio.channels.Channels;
 /** Service class for getting credentials from key files stored locally or on cloud storage. */
 class CredentialsService {
   private static final String GOOGLE_CLOUD_STORAGE_PREFIX = "gs://";
-  private static final String INVALID_GCS_PREFIX_MSG = String.format("Storage URL must start with %s", GOOGLE_CLOUD_STORAGE_PREFIX);
+  private static final String INVALID_GCS_PREFIX_MSG =
+      String.format("Storage URL must start with %s", GOOGLE_CLOUD_STORAGE_PREFIX);
 
   static final CredentialsService INSTANCE = new CredentialsService();
-
-  /** Util for creating a default http transport */
-  static class CloudSpannerOAuthUtil {
-    static final HttpTransport HTTP_TRANSPORT = new NetHttpTransport();
-    static final HttpTransportFactory HTTP_TRANSPORT_FACTORY = new DefaultHttpTransportFactory();
-
-    static class DefaultHttpTransportFactory implements HttpTransportFactory {
-      @Override
-      public HttpTransport create() {
-        return HTTP_TRANSPORT;
-      }
-    }
-
-    private CloudSpannerOAuthUtil() {}
-  }
 
   CredentialsService() {}
 
@@ -106,8 +89,7 @@ class CredentialsService {
           String.format("Error reading credential file %s: File does not exist", filePath));
     }
     try (InputStream credentialsStream = new FileInputStream(credentialsFile)) {
-      return GoogleCredentials.fromStream(
-          credentialsStream, CloudSpannerOAuthUtil.HTTP_TRANSPORT_FACTORY);
+      return GoogleCredentials.fromStream(credentialsStream);
     }
   }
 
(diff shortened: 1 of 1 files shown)


kolea2:
@chingor13 can we please get your input here? Can we make use of existing stuff in the auth library?

chingor13:
This seems to be adding convenience methods for creating credentials. It's pretty opinionated on the options available for obtaining the credentials so it probably doesn't make sense to put it into the auth library.

I'm not familiar with the configuration options available to jdbc - i.e. are you only allowed to configure via a configuration file (thus only have a String uri as configuration options)?

We normally prefer users to instantiate their `GoogleCredentials` themselves and provide that instance to our clients - this allows users to use all credential methods, rather than a curated subset of them.

olavloite:
@chingor13 JDBC only allows you to configure the connection through the connection URL or through a `Properties` object containing only String values. So yes, only String configuration is possible.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
```diff
@@ -0,0 +1,168 @@
...
+import java.io.InputStream;
+import java.nio.channels.Channels;
+
+/**
+ * Internal class for the Spanner Connection API. This class and its methods have public visibility
+ * only for allowing integration tests to access it. This may change in future releases.
+ *
+ * <p>Service class for getting credentials from key files stored locally or on cloud storage.
+ */
+@VisibleForTesting
```
chingor13: We also have an `@InternalApi` annotation we could use. We could also put this in a `com.google.cloud.spanner.jdbc.internal` package if necessary.

olavloite: I've looked into why this class was given public visibility in the first place, and it is no longer needed. It used to be referenced by one or more integration tests in a different package, but that is no more, so I've changed the visibility to package-private.

olavloite:
@chingor13 Fetching app default credentials is now mocked in the test case.

kolea2:
LGTM, will merge later today unless I hear any objections from @chingor13

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
```diff
@@ -0,0 +1,163 @@
...
+  @VisibleForTesting
+  InputStream internalCreateInputStream(Blob blob) {
+    return Channels.newInputStream(blob.reader());
+  }
+
+  @VisibleForTesting
+  String internalGetBucket(String storageUrl) {
+    Preconditions.checkArgument(
+        storageUrl.startsWith(GOOGLE_CLOUD_STORAGE_PREFIX),
+        String.format("Storage URL must start with %s", GOOGLE_CLOUD_STORAGE_PREFIX));
```
chingor13: nit: not sure how often this gets called (likely just once) but we could make this a constant for use here and in `internalGetBlob`

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
```diff
@@ -0,0 +1,163 @@
...
+                  + "Credentials on Google Cloud Storage can only be used when the application is running "
+                  + "in an environment that already has default credentials to access Google Cloud Storage",
+              credentialsUrl),
+          e);
+    }
+  }
+
+  @VisibleForTesting
+  Storage internalCreateStorage() {
+    return StorageOptions.newBuilder().build().getService();
```
chingor13: If we're defining our own `HttpTransport`/`HttpTransportFactory` then we should probably be using them here as well. 

But, since we're not actually defining anything custom (at least for now), we could probably use the defaults across the board (i.e. `GoogleCredentials.fromStream(InputStream)` instead of `GoogleCredentials.fromStream(InputStream, HttpTransportFactory)` and drop the `CloudSpannerOAuthUtil` inner class.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
olavloite: Moved to a constant.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
olavloite: There was a reason why it was made that way, but I cannot remember why (it's 9 months ago). I've just tested without the custom HttpTransport and it seems to work, so I've removed it.

chingor13: Eventually, it could be possible that a user needs a custom HttpTransport (proxy or other rules).

Issue #5764, opened 2019-07-18
CollectionReference listDocuments return ApiFuture of DocRefs 


ChengyuanZhao:
Hi, I'm from the java frameworks team (I'm cccz@google.com) and we're working on a Java Spring Data Reactive implementation for Firestore. (an ORM sort of thing)

We came across CollectionReference.listDocuments which is exactly what we needed but we see that the method blocks and waits for the list of docrefs to return. We also see that the package-private Firestore.sendRequest method within actually returns an ApiFuture so it seems like a small amount of work to make an alternate function, perhaps `futureListDocuments`, that doesn't block and returns an ApiFuture.

The ApiFuture return values are very important to us since we're building a reactive implementation.

We would be happy to contribute a PR for this feature if you guys would accept that. 

my teammates: @elefeint @dmitry-s @meltsufin @dzou

BenWhitehead:
@ChengyuanZhao I believe you decided to go in a different direction from this issue, is that correct? Can we close this?

ChengyuanZhao:
Yes @BenWhitehead , thanks much this issue is now very out of date and obsolete. Closing.

PR #5878, opened 2019-07-25
Firestore: add CollectionReference#listDocumentAsync

athakor:
Towards #5764

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
@@ -167,8 +167,13 @@ public class CollectionReference extends Query {
     };
   }
 
+  /**
+   * Retrieves the list of documents in this collection
+   *
+   * @return An ApiFuture
+   */
   @Nonnull
-  public ApiFuture<?> futureListDocuments() {
+  public ApiFuture<?> listDocumentAsync() {
     ListDocumentsRequest.Builder request = ListDocumentsRequest.newBuilder();
     request.setParent(options.getParentPath().toString());
     request.setCollectionId(options.getCollectionId());
--- a/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
+++ b/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
@@ -54,10 +54,12 @@ import com.google.cloud.firestore.Transaction;
 import com.google.cloud.firestore.Transaction.Function;
 import com.google.cloud.firestore.WriteBatch;
 import com.google.cloud.firestore.WriteResult;
+import com.google.cloud.firestore.v1.FirestoreClient;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
+import com.google.firestore.v1.Document;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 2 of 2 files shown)


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
```diff
@@ -167,6 +167,27 @@ public void remove() {
...
+    request.setShowMissing(true);
+
+    try {
+      return firestore.sendRequest(
+          request.build(), firestore.getClient().listDocumentsPagedCallable());
+    } catch (ApiException exception) {
+      throw FirestoreException.apiException(exception);
+    }
+  }
+
```
schmidt-sebastian: The original API is already somewhat async. It returns an Iterator right away, but it does block on the "hasNext/next" call.  If you want to move away from the Iterator, then we have to expose `UnaryCallable<ListCollectionIdsRequest, ListCollectionIdsResponse>` (the type that the question mark refers to) and you can no longer use `iterateAll()`.

I understand that there are some usages for which this high-level API is not the right fit. There is however a middle-ground between using this SDK and the RPC layer - have you looked into using the auto-generated client that is exposed by https://github.com/googleapis/google-cloud-java/blob/10070bd8c646780363da6c21fe3d0f358be0b724/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/v1/FirestoreClient.java

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -662,6 +664,35 @@ public void listDocuments() throws Exception {
...
+      batch.create(randomColl.document(document), SINGLE_FIELD_OBJECT);
+    }
+    batch.commit().get();
+
+    ApiFuture<FirestoreClient.ListDocumentsPagedResponse> collectionRefs =
+        (ApiFuture<FirestoreClient.ListDocumentsPagedResponse>) randomColl.listDocumentAsync();
+    Iterator<Document> iterator = collectionRefs.get().iterateAll().iterator();
+
+    int count = 0;
+    for (Iterator<Document> documentIterator = iterator; documentIterator.hasNext(); ) {
```
schmidt-sebastian: Unfortunately, this still blocks. iterateAll() uses an auto-paging iterator that receives the next page synchronously: https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/paging/AbstractPage.java#L118

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
athakor: @schmidt-sebastian yes right but i have question regarding `UnaryCallable<ListCollectionIdsRequest,ListCollectionIdsResponse>` if we expose `UnaryCallable<ListCollectionIdsRequest,ListCollectionIdsResponse>` then it will only returns collectionIdsList  there is no way to get list of document so how we can move ahead with this callable if we want to get list of document?

schmidt-sebastian: You have to run it in two stages and use ListDocuments in the second stage:

https://github.com/googleapis/google-cloud-java/blob/10070bd8c646780363da6c21fe3d0f358be0b724/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/v1/FirestoreClient.java#L262

If you want to iterate a Firestore database you have to switch recursively between these two calls.

athakor: @schmidt-sebastian could please verify below code snippest and please suggest me if any changes required.
`
public ApiFuture<ListDocumentsPagedResponse> listDocumentAsync() throws  Exception {

      ListCollectionIdsRequest.Builder collectionIdsRequest = ListCollectionIdsRequest.newBuilder().setParent(options.getParentPath().toString());
      ListCollectionIdsResponse response = firestore.getClient().listCollectionIdsCallable().call(collectionIdsRequest.build());
      ApiFuture<ListDocumentsPagedResponse> listDocPagedResponseApiFuture = null;
      for(String id : response.getCollectionIdsList()){
          ListDocumentsRequest docRequest =
                  ListDocumentsRequest.newBuilder()
                          .setParent(options.getParentPath().toString())
                          .setCollectionId(id).build();
          listDocPagedResponseApiFuture= firestore.sendRequest(docRequest, firestore.getClient().listDocumentsPagedCallable());
      }
      return listDocPagedResponseApiFuture;
  }`

PR #5879, opened 2019-07-25
Firestore:Fix make change in CustomClassMapper serialize and deserialize methods to public

### LINKED ISSUE #5809: Make CustomClassMapper's serialization and deserialization methods public
Hi,

we are working on Spring Data for Firestore. 

It would be useful for us if we could reuse existing object conversion logic so we don't have to re-implement it for our project.
It looks like `com.google.cloud.firestore.CustomClassMapper` have all the necessary logic, but it is package-private.

teammates: @meltsufin @elefeint @ChengyuanZhao @dzou

abhinav-qlogic:
Fixes #5809


BenWhitehead:
Superseded by #6085

PR #5882, opened 2019-07-25
Add CSEK decryption for storage.get(BlobId,..)

frankyn:
Hi @JesseLovelace,

Could you take a look at this fix for an internal customer request: b/79339975.

Currently storage.get() doesn't support decrypting object crc32c and md5 using CSEK. When objects are encrypted using CSEK, md5 and crc32c are also encrypted. Without supporting the option to decrypt it blocks this specific use-case. 

For more information: https://cloud.google.com/storage/docs/encryption/customer-supplied-keys#encrypted-elements


frankyn:
Thanks @jesselovelace!

PR #5888, opened 2019-07-25
feat: add shared config maven project

chingor13:
The new parent pom contains the build config shared between all our releasable artifacts.

### DIFF
--- a/google-api-grpc/proto-google-cloud-os-login-v1/pom.xml
+++ b/google-api-grpc/proto-google-cloud-os-login-v1/pom.xml
@@ -11,6 +11,11 @@
     <artifactId>google-api-grpc</artifactId>
     <version>0.67.1-SNAPSHOT</version><!-- {x-version-update:google-api-grpc:current} -->
   </parent>
+
+  <properties>
+    <checkstyle.excludes>com/google/cloud/oslogin/**</checkstyle.excludes>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>com.google.protobuf</groupId>
@@ -28,18 +33,4 @@
       <scope>compile</scope>
     </dependency>
   </dependencies>
-
-  <build>
-    <pluginManagement>
-      <plugins>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-checkstyle-plugin</artifactId>
-          <configuration>
-            <excludes>com/google/cloud/oslogin/**</excludes>
-          </configuration>
-        </plugin>
-      </plugins>
-    </pluginManagement>
-  </build>
 </project>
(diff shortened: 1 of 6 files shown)


google-api-grpc/proto-google-cloud-monitoring-v3/src/main/java/google/monitoring/v3/DroppedLabelsOuterClass.java
```diff
@@ -1,914 +0,0 @@
-// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
BenWhitehead: What this java file supposed to be deleted?

chingor13: That file was erroneously generated a long time ago. It's missing `com.` in the package. The real version is https://github.com/googleapis/google-cloud-java/blob/master/google-api-grpc/proto-google-cloud-monitoring-v3/src/main/java/com/google/monitoring/v3/DroppedLabelsOuterClass.java

google-api-grpc/proto-google-cloud-os-login-v1/pom.xml
```diff
@@ -28,4 +28,18 @@
...
   </dependencies>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-checkstyle-plugin</artifactId>
+          <configuration>
+            <excludes>com/google/cloud/oslogin/**</excludes>
```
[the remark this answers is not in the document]
chingor13: Good idea, there's already a `checkstyle.excludes` property to use.

versions.txt
```diff
@@ -55,6 +55,7 @@ google-cloud-redis:0.102.0-alpha:0.102.1-alpha-SNAPSHOT
 google-cloud-resourcemanager:0.102.0-alpha:0.102.1-alpha-SNAPSHOT
 google-cloud-scheduler:1.7.0:1.7.1-SNAPSHOT
 google-cloud-securitycenter:0.102.0:0.102.1-SNAPSHOT
+google-cloud-shared-config:0.0.0:0.0.1-SNAPSHOT
```
kolea2: we're releasing this?

chingor13: I *think* we have to release it. It's like how we release google-api-grpc and google-cloud-client's as poms to maven.

We could also wait until the split repo to actually do this. Doing it now will mean it gets versioned everytime we release the monorepo. When we have a split repo available to move it to, we would stop versioning it here.

BenWhitehead: It does have to be released. Anything that is referenced as a parent is also loaded a use time.

chingor13:
Let's hold off on this until the repo split.

igorbernstein2:
Should google-cloud-testing/google-cloud-bigtable-emulator's parent also change?

chingor13:
We will introduce this artifact in the new split repo.

PR #5890, opened 2019-07-26
spanner-jdbc: Step 15 - ConnectionOptions and SpannerPool

olavloite:
A `ConnectionImpl` instance, which is the internal Spanner connection API, can only be created from a `ConnectionOptions` instance. All connections that are created by `ConnectionOptions` are registered with a `SpannerPool` that keeps track of the connections that have been opened and which `Spanner` instances from the client library are associated with which connection. The `Spanner` instances are kept in a pool and reused for connections because these are heavyweight objects that are expensive to create and maintain, and are intended to live as long as the application interacts with Cloud Spanner.

This change also adds a client lib token property to `SpannerOptions` in the client library. This token will be used to track the usage of the JDBC driver. The actual usage of this token will be added in a separate PR.

This PR contains a couple of stub classes that will be replaced by the actual classes in subsequent PRs. This is done to keep the PR as small as possible. The stubs are:
* ConnectionImpl
* CredentialsService
* JdbcDriver
* UnitOfWork

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
@@ -230,7 +230,7 @@ public class SpannerPool {
       if (!initialized) {
         initialize();
       }
-      if (spanners.get(key) == null) {
+      if (spanners.get(key) != null) {
         spanner = spanners.get(key);
       } else {
         spanner = createSpanner(key);


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionImpl.java
```diff
@@ -0,0 +1,318 @@
...
+
+  @Override
+  public void close() {
+    leakedException = null;
+    SpannerPool.INSTANCE.removeConnection(options, this);
+  }
+
+  @Override
+  public boolean isClosed() {
+    // TODO Auto-generated method stub
```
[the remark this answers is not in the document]
olavloite: done

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
```diff
@@ -0,0 +1,407 @@
...
+    }
+    initialized = true;
+  }
+
+  @SuppressWarnings("rawtypes")
+  @VisibleForTesting
+  Spanner createSpanner(SpannerPoolKey key) {
+    SpannerOptions.Builder builder = SpannerOptions.newBuilder();
+    builder
+        .setClientLibToken(MoreObjects.firstNonNull(key.userAgent, JdbcDriver.getClientLibToken()))
```
kolea2: Here is how spring-cloud-gcp does it: https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java#L126. I'd prefer this approach so we don't need to add the setter in the builder.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
```diff
@@ -0,0 +1,407 @@
...
+        if (!keysStillInUse.isEmpty()) {
+          logLeakedConnections(keysStillInUse);
+          logger.log(
+              Level.WARNING,
+              "There is/are "
+                  + keysStillInUse.size()
+                  + " connection(s) still open."
+                  + " Close all connections before stopping the application");
+        }
+        closeUnusedSpanners(Long.MIN_VALUE);
```
kolea2: to confirm my understanding, passing through Long.MIN_VALUE will force the spanner sessions to close?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
```diff
@@ -0,0 +1,407 @@
...
+  /**
+   * Closes Spanner objects that are no longer in use by connections, and where the last connection
+   * that used it was closed more than <code>closeSpannerAfterMillisecondsUnused</code> seconds ago.
+   * The delay ensures that Spanner objects are not closed unless there's a good reason for it.
+   *
+   * @param closeSpannerAfterMillisecondsUnused The number of milliseconds a {@link Spanner} object
+   *     should not have been used for a {@link Connection} before it is closed by this method.
+   */
+  @VisibleForTesting
+  void closeUnusedSpanners(long closeSpannerAfterMillisecondsUnused) {
```
kolea2: where does this parameter get its value?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
olavloite: That's something that we would need to discuss with the Spanner team in that case. The header used by spring-cloud-gcp sets a 'User-Agent' header, the client lib token of the JDBC driver sets the ApiClientHeader ('x-goog-api-client'). My initial solution was also what spring-cloud-gcp did, but it didn't show up in the monitoring that the Spanner team had. The ApiClientHeader did, which is why this solution was chosen, and the monitoring is now built around that.

That being said, I looked a little bit more into the spring-cloud-gcp solution and compared that with the original solution for the JDBC driver, and I noticed one important difference: spring-cloud-gcp sets a 'User-Agent' header, the JDBC driver used a 'user-agent' header, so that could be the reason it didn't show up.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
```diff
@@ -0,0 +1,407 @@
...
+  Spanner getSpanner(ConnectionOptions options, ConnectionImpl connection) {
+    Preconditions.checkNotNull(options);
+    Preconditions.checkNotNull(connection);
+    SpannerPoolKey key = SpannerPoolKey.of(options);
+    Spanner spanner;
+    synchronized (this) {
+      if (!initialized) {
+        initialize();
+      }
+      if (spanners.containsKey(key)) {
```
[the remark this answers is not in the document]
olavloite: Agree, in this case that is more appropriate, as the return value of the `getSpanner(...)` method should never be null (even if the map somehow contained a null value for the key).

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
```diff
@@ -0,0 +1,407 @@
...
+      } else {
+        spanner = createSpanner(key);
+        spanners.put(key, spanner);
+      }
+      List<ConnectionImpl> registeredConnectionsForSpanner = connections.get(key);
+      if (registeredConnectionsForSpanner == null) {
+        registeredConnectionsForSpanner = new ArrayList<>();
+        connections.put(key, registeredConnectionsForSpanner);
+      }
+      registeredConnectionsForSpanner.add(connection);
```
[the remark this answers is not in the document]
olavloite: I'm not 100% sure what you mean with 'duplicate' connections in this case, but:
* Several `Connection` instances can share the same `Spanner` instance, and a `Spanner` instance should only be closed once all the related `Connection` instances have been closed. The reason behind this is that `Spanner` instances are heavy-weight objects that should live throughout (most of) the lifetime of an application. A `Spanner` instance also maintains a session pool. `Connection` instances are relatively light-weight objects that an application may create and close many times during its lifetime. In Cloud Spanner terms, a `Connection` is comparable to a `Session`. But as the JDBC driver is built on top of the client library, and the client library does not expose sessions and session management, this solution is the closest we can get having `Connection` == `Session`.
* `Connection` instances are not thread safe and should not be used by several processes at the same time. An application will normally either maintain its own connection pool and reuse connections through this pool, or create and close connections when needed for a process. `ConnectionImpl` does not override the default `equals(Object)` method, so different connection instances will never be considered equal.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
```diff
@@ -0,0 +1,407 @@
...
+  private ScheduledExecutorService closerService;
+
+  @GuardedBy("this")
+  private final Map<SpannerPoolKey, Spanner> spanners = new HashMap<>();
+
+  @GuardedBy("this")
+  private final Map<SpannerPoolKey, List<ConnectionImpl>> connections = new HashMap<>();
+
+  @GuardedBy("this")
+  private final Map<SpannerPoolKey, Long> lastConnectionClosedAt = new HashMap<>();
```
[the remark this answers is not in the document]
olavloite: Done.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
olavloite: Correct. I've added a comment to the method invocation to make that clearer.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
olavloite: The default value that is used when nothing else is configured is found here: https://github.com/googleapis/google-cloud-java/blob/a6d9f88a154c223bb079e6e4d0d5a9619fd6b3d6/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java#L75

This value is currently only configurable for testing. A client application cannot change this value and will always use the default.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SpannerPool.java
```diff
@@ -211,7 +211,7 @@ Spanner getSpanner(ConnectionOptions options, ConnectionImpl connection) {
       if (!initialized) {
         initialize();
       }
-      if (spanners.containsKey(key)) {
+      if (spanners.get(key) == null) {
```
kolea2: I think this is backwards now?

kolea2: can we add some more testing around this?

olavloite: Oops, that's indeed backward. It's already covered by the test case [here](https://github.com/googleapis/google-cloud-java/blob/a070c22624a73246059e3b694af6d4595de62fc4/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/SpannerPoolTest.java#L93), but because this PR has build errors, I didn't pick up on that one. I added the same error to the private repo and ran the test cases, and there were multiple test failures.

PR #5891, opened 2019-07-26
spanner-jdbc: Step 16 - JdbcResultSet

olavloite:
Adds `java.sql.ResultSet` implementation for Cloud Spanner. This is basically a wrapper around a `com.google.cloud.spanner.ResultSet`.

(The branch of this PR is based on step 13. Step 13 has been merged and this PR has been rebased.)

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
@@ -25,7 +25,11 @@ import com.google.common.base.Preconditions;
 
 /** Implementation of {@link ResultSetMetaData} for Cloud Spanner */
 class JdbcResultSetMetaData extends AbstractJdbcWrapper implements ResultSetMetaData {
-
+  /**
+   * The default column display size for columns with a data type of variable size that is used
+   * when the actual column size is not known.
+   */
+  private static final int DEFAULT_COL_DISPLAY_SIZE_FOR_VARIABLE_LENGTH_COLS = 50;
   private final ResultSet spannerResultSet;
   private final Statement statement;
 
@@ -82,12 +86,12 @@ class JdbcResultSetMetaData extends AbstractJdbcWrapper implements ResultSetMeta
     int colType = getColumnType(column);
     switch (colType) {
       case Types.ARRAY:
-        return 50;
+        return DEFAULT_COL_DISPLAY_SIZE_FOR_VARIABLE_LENGTH_COLS;
       case Types.BOOLEAN:
         return 5;
       case Types.BINARY:
         int binaryLength = getPrecision(column);
-        return binaryLength == 0 ? 50 : binaryLength;
+        return binaryLength == 0 ? DEFAULT_COL_DISPLAY_SIZE_FOR_VARIABLE_LENGTH_COLS : binaryLength;
       case Types.DATE:
         return 10;
       case Types.DOUBLE:
@@ -96,7 +100,7 @@ class JdbcResultSetMetaData extends AbstractJdbcWrapper implements ResultSetMeta
         return 10;
       case Types.NVARCHAR:
         int length = getPrecision(column);
-        return length == 0 ? 50 : length;
+        return length == 0 ? DEFAULT_COL_DISPLAY_SIZE_FOR_VARIABLE_LENGTH_COLS : length;
       case Types.TIMESTAMP:
         return 16;
       default:
@@ -134,8 +138,11 @@ class JdbcResultSetMetaData extends AbstractJdbcWrapper implements ResultSetMeta
       case Types.TIMESTAMP:
         return 24;
       default:
-        // Not fixed size, we could try to get it from INFORMATION_SCHEMA
-        return 50;
+        // For column types with variable size, such as text columns, we should return the length
+        // in characters. We could try to fetch it from INFORMATION_SCHEMA, but that would mean
+        // parsing the SQL statement client side in order to figure out which column it actually
+        // is. For now we just return the default column display size.
+        return DEFAULT_COL_DISPLAY_SIZE_FOR_VARIABLE_LENGTH_COLS;
     }
   }
 


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/AbstractJdbcResultSet.java
```diff
@@ -0,0 +1,627 @@
...
+  }
+
+  @Override
+  public int getConcurrency() throws SQLException {
+    return CONCUR_READ_ONLY;
+  }
+
+  @Override
+  public boolean rowUpdated() throws SQLException {
+    return false;
```
kolea2: do any of these have meaningful values? If no, should we just throw unsupported for all of them?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/AbstractJdbcWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+import com.google.cloud.spanner.Type;
+import com.google.cloud.spanner.Type.Code;
+
+/** Base class for all Cloud Spanner JDBC classes that implement the {@link Wrapper} interface. */
+abstract class AbstractJdbcWrapper implements Wrapper {
+  static final String OTHER_NAME = "OTHER";
+
+  /** Extract {@link java.sql.Types} code from Spanner {@link Type}. */
+  static int extractColumnType(Type type) {
+    if (type.equals(Type.bool()))
```
kolea2: I would either flip the equality here, or check for null and throw first.

kolea2: Same applies to further if statements in this file.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcArray.java
```diff
@@ -0,0 +1,204 @@
...
+    if (data != null) {
+      Object res = java.lang.reflect.Array.newInstance(type.getJavaClass(), count);
+      System.arraycopy(data, (int) index - 1, res, 0, count);
+      return res;
+    }
+    return null;
+  }
+
+  private static final String RESULTSET_NOT_SUPPORTED =
+      "Getting a resultset from an array is not supported";
```
kolea2: 'ResultSet'

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcClob.java
```diff
@@ -0,0 +1,219 @@
...
+  public InputStream getAsciiStream() throws SQLException {
+    return new ByteArrayInputStream(StandardCharsets.US_ASCII.encode(value.toString()).array());
+  }
+
+  @Override
+  public long position(String searchstr, long start) throws SQLException {
+    JdbcPreconditions.checkArgument(start >= 1, "Start position must be >= 1");
+    JdbcPreconditions.checkArgument(searchstr != null, "searchstr may not be null");
+    checkPosition(start);
+    int res = value.indexOf(searchstr, (int) start - 1);
```
kolea2: 'searchStr'

kolea2: same applies to below usage of the parameter

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDataType.java
```diff
@@ -0,0 +1,267 @@
+/*
```
kolea2: I think this was already done in step 13. Does it need a rebase?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
```diff
@@ -0,0 +1,205 @@
...
+  public int getColumnDisplaySize(int column) throws SQLException {
+    int colType = getColumnType(column);
+    switch (colType) {
+      case Types.ARRAY:
+        return 50;
+      case Types.BOOLEAN:
+        return 5;
+      case Types.BINARY:
+        int binaryLength = getPrecision(column);
+        return binaryLength == 0 ? 50 : binaryLength;
```
kolea2: let's extract 50 out to a variable - not clear to me what it represents

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
```diff
@@ -0,0 +1,205 @@
...
+        return 10;
+      case Types.DOUBLE:
+        return 14;
+      case Types.BIGINT:
+        return 10;
+      case Types.TIMESTAMP:
+        return 24;
+      default:
+        // Not fixed size, we could try to get it from INFORMATION_SCHEMA
+        return 50;
```
kolea2: from here, guessing it's default precision?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
```diff
@@ -0,0 +1,205 @@
...
+  public int getScale(int column) throws SQLException {
+    int colType = getColumnType(column);
+    if (colType == Types.DOUBLE)
+      return 15;
+    return 0;
+  }
+
+  @Override
+  public String getTableName(int column) throws SQLException {
+    return "";
```
kolea2: why empty string?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcTypeConverter.java
```diff
@@ -0,0 +1,274 @@
+/*
```
kolea2: think this was again already added

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaDataTest.java
```diff
@@ -0,0 +1,512 @@
...
+    if (type == Type.float64())
+      return Value.float64(123.45D);
+    if (type == Type.int64())
+      return Value.int64(12345L);
+    if (type == Type.string())
+      return Value.string("test value " + row);
+    if (type == Type.timestamp())
+      return Value.timestamp(com.google.cloud.Timestamp.now());
+
+    if (type.getCode() == Code.ARRAY && type.getArrayElementType() == Type.bool())
```
kolea2: let's refactor this to a larger if statement checking if the code == ARRAY

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaDataTest.java
```diff
@@ -0,0 +1,512 @@
...
+        return String[].class.getName();
+      if (type.getArrayElementType() == Type.timestamp())
+        return Timestamp[].class.getName();
+    }
+    return null;
+  }
+
+  @Test
+  public void testToString() {
+    assertNotNull(subject.toString());
```
kolea2: let's also assert what the string value is

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetTest.java
```diff
@@ -0,0 +1,1101 @@
...
+  static final int URL_COLINDEX_NULL = 19;
+  static final int URL_COLINDEX_NOTNULL = 20;
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  private JdbcResultSet subject;
+
+  static ResultSet getMockResultSet() {
+    //@formatter:off
```
kolea2: delete?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetTest.java
```diff
@@ -0,0 +1,1101 @@
...
+          .set(TIMESTAMP_COL_NOT_NULL).to(TIMESTAMP_VALUE)
+          .set(TIME_COL_NULL).to((Timestamp) null)
+          .set(TIME_COL_NOT_NULL).to(TIME_VALUE)
+          .set(ARRAY_COL_NULL).toInt64Array((long[]) null)
+          .set(ARRAY_COL_NOT_NULL).toInt64Array(ARRAY_VALUE)
+          .set(URL_COL_NULL).to((String) null)
+          .set(URL_COL_NOT_NULL).to(URL_VALUE)
+        .build()
+      ));
+    //@formatter:on
```
kolea2: delete?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/AbstractJdbcResultSet.java
olavloite: There's something to say for both throwing `SQLFeatureNotSupportedException` and for returning `false` in these cases. The JDBC specification does not specify exactly what a driver should do with these methods when the `updateXYZ` are not supported, only the other way around: When `updateXYZ` is supported, then these methods must also be implemented. In general, I've chosen the following generic approach for all methods in the JDBC driver where the implementation is optional:
* Return a sensible (constant) value when the driver can determine one, even when the other methods that would normally be used to change the state that is returned by the method are not supported.
* Throw a `SQLFeatureNotSupportedException` when no reasonable (constant) value can be returned, or when the method should alter the state of the object and it is not supported by the driver or Cloud Spanner.

In this specific case we know that the row has not been updated, because we do not support the update methods. This is also reflected in the return value of the `DatabaseMetaData#updatesAreDetected(int)` method, so a user should be able to know beforehand that calling these methods will not have any effect.

The reasoning behind the above generic approach is that the driver can also be used with generic third-party tools like for example DBeaver and with all sorts of generic Java frameworks. As both we and the end user have no influence on how these tools use the driver, it will give a better user experience when we are little bit lenient when possible, rather than throwing exceptions for everything that is not supported, even when the tool in question could have determined that calling the methods is not supported.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDataType.java
olavloite: I based this PR on step 13 to avoid build errors. Once 13 is merged, this one should be rebased.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcTypeConverter.java
olavloite: I based this PR on step 13 to avoid build errors. Once 13 is merged, this one should be rebased.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaDataTest.java
olavloite: done

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaDataTest.java
olavloite: done

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetTest.java
olavloite: Removed.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcResultSetTest.java
olavloite: Removed.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDataType.java
olavloite: It has now been rebased.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
olavloite: Extracted to a variable and added javadoc to explain its purpose.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
olavloite: For text columns, this method should return the size of the column in number of characters. Cloud Spanner does not return that information, meaning that we could only get it by parsing the SQL statement that has been used to create the result set in order to figure out which column it is, and then fetch it from INFORMATION_SCHEMA. I've added some more documentation to make that clearer.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java
olavloite: Cloud Spanner does not return the table name in its metadata. We could try to figure it out by parsing the SQL statement that was used to get the result set, but I don't think that's something we want to do for the first version. The javadoc of this method on the interface states `@return table name or "" if not applicable`.

PR #5894, opened 2019-07-26
spanner-jdbc: Step 17 - Base class for transactions and batches

olavloite:
This adds the abstract base implementation for all transactions and batches that can be executed on a connection, and the executor that is used for executing statements on a connection.

(This PR will have build errors until 15 has been merged)

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/AbstractBaseUnitOfWork.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/AbstractBaseUnitOfWork.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2019 Google LLC
+ *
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/AbstractBaseUnitOfWork.java
```diff
@@ -0,0 +1,159 @@
+/*
```
kolea2: Any tests we should be adding here? Or will be in future commits?

olavloite: The tests for both this abstract base class as well as much of the concrete subclasses are handled by the SQL based unit tests. I will add these through separate PRs, as they need the whole stack of connection and transactions to be present to work.

PR #5895, opened 2019-07-26
spanner-jdbc: Step 18 - ConnectionImpl

olavloite:
Adds the implementation class of the internal Spanner connection API. JDBC connections act as wrappers around this class. A ConnectionImpl instance is used to interact with Cloud Spanner through several different types of transactions or batches. There can be at most one transaction or batch active at any one time on a connection.

This PR will have build errors as it depends on step 15 (ConnectionOptions), but also on the different types of transaction classes that will be brought in through separate PRs.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionImpl.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionImpl.java
@@ -38,6 +38,7 @@ import com.google.cloud.spanner.SpannerExceptionFactory;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.TimestampBound;
 import com.google.cloud.spanner.TimestampBound.Mode;
+import com.google.cloud.spanner.jdbc.StatementExecutor.StatementTimeout;
 import com.google.cloud.spanner.jdbc.StatementParser.ParsedStatement;
 import com.google.cloud.spanner.jdbc.StatementParser.StatementType;
 import com.google.cloud.spanner.jdbc.UnitOfWork.UnitOfWorkState;
@@ -362,7 +363,7 @@ class ConnectionImpl implements Connection {
   @Override
   public void setStatementTimeout(long timeout, TimeUnit unit) {
     Preconditions.checkArgument(timeout > 0L, "Zero or negative timeout values are not allowed");
-    Preconditions.checkArgument(StatementExecutor.isValidTimeoutUnit(unit),
+    Preconditions.checkArgument(StatementTimeout.isValidTimeoutUnit(unit),
         "Time unit must be one of NANOSECONDS, MICROSECONDS, MILLISECONDS or SECONDS");
     ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
     this.statementTimeout.setTimeoutValue(timeout, unit);
@@ -377,7 +378,7 @@ class ConnectionImpl implements Connection {
   @Override
   public long getStatementTimeout(TimeUnit unit) {
     ConnectionPreconditions.checkState(!isClosed(), CLOSED_ERROR_MSG);
-    Preconditions.checkArgument(StatementExecutor.isValidTimeoutUnit(unit),
+    Preconditions.checkArgument(StatementTimeout.isValidTimeoutUnit(unit),
         "Time unit must be one of NANOSECONDS, MICROSECONDS, MILLISECONDS or SECONDS");
     return this.statementTimeout.getTimeoutValue(unit);
   }


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionImpl.java
```diff
@@ -0,0 +1,922 @@
...
+   */
+  private final ConnectionStatementExecutor connectionStatementExecutor =
+      new ConnectionStatementExecutorImpl(this);
+
+  /** Simple thread factory that is used for fire-and-forget rollbacks. */
+  static final class DaemonThreadFactory implements ThreadFactory {
+    @Override
+    public Thread newThread(Runnable r) {
+      Thread t = new Thread(r);
+      t.setDaemon(true);
```
kolea2: Should we also add a qualifying name?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionImpl.java
```diff
@@ -0,0 +1,922 @@
...
+  private TimestampBound readOnlyStaleness = TimestampBound.strong();
+
+  /** Create a connection and register it in the SpannerPool. */
+  ConnectionImpl(ConnectionOptions options) {
+    Preconditions.checkNotNull(options);
+    this.statementExecutor = new StatementExecutor(options.getStatementExecutionInterceptors());
+    this.spannerPool = SpannerPool.INSTANCE;
+    this.options = options;
+    this.spanner = spannerPool.getSpanner(options, this);
+    this.ddlClient = DdlClient.newBuilder().setDatabaseAdminClient(spanner.getDatabaseAdminClient())
```
kolea2: I'd move this logic to a private method, `createDdlClient`

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionStatementExecutorImpl.java
```diff
@@ -0,0 +1,227 @@
...
+  ConnectionStatementExecutorImpl(ConnectionImpl connection) {
+    this.connection = connection;
+  }
+
+  ConnectionImpl getConnection() {
+    return connection;
+  }
+
+  @Override
+  public StatementResult statementSetAutocommit(Boolean autocommit) {
```
kolea2: should autocommit be autoCommit? WDYT?

kolea2: (not sure if this is an already defined naming convention elsewhere)

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionImpl.java
olavloite: Added.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionImpl.java
olavloite: Moved to a separate method, and moved the assignment to the end of the constructor to ensure that all fields have been initialized.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionStatementExecutorImpl.java
olavloite: Both are used. I have a slight preference for `autocommit`, but I don't feel very strongly about this.

JDBC uses `autoCommit`. [SQL Server](https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/auto-commit-mode?view=sql-server-2017) uses `auto-commit`. [PostgreSQL](https://www.postgresql.org/docs/11/ecpg-sql-set-autocommit.html) and [MySQL](https://dev.mysql.com/doc/refman/8.0/en/innodb-autocommit-commit-rollback.html) use `autocommit` in their documentation.

olavloite: One reason to keep this as `autocommit` is that we have defined the variable `AUTOCOMMIT_DML_MODE`, which indicates that the driver assumes autocommit to be one word.

kolea2: SGTM

PR #5897, opened 2019-07-26
spanner-jdbc: Step 20 - ReadOnlyTransaction

olavloite:
Adds `ReadOnlyTransaction` for the internal Spanner connection API. These transactions can only be used to execute queries. ReadOnlyTransactions support using a read-only staleness mode and timestamp or staleness.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReadOnlyTransactionTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReadOnlyTransactionTest.java
@@ -1,0 +1,403 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ReadOnlyTransactionTest.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    @Override
+    public Struct readRowUsingIndex(String table, String index, Key key, Iterable<String> columns) {
+      return null;
+    }
+
+    @Override
+    public ResultSet executeQuery(Statement statement, QueryOption... options) {
+      if (readTimestamp == null) {
```
kolea2: How come we are overriding these methods for the test?

olavloite: I guess the naming is a little bit confusing in this case, as we have two different `ReadOnlyTransaction`s in this case. The interface that is being implemented is `com.google.cloud.spanner.ReadOnlyTransaction`, i.e. the read-only transaction of the normal client library, and not the read-only transaction of the connection API. That is done in this test in order to create a simple mock object with some logic to use in the test cases.

The logic in the `executeQuery(...)` method is intended to make sure the connection API actually uses the correct staleness mode for the underlying read-only transaction (and to return a mock result set).

PR #5898, opened 2019-07-26
spanner-jdbc: Step 21 - ReadWriteTransaction

olavloite:
Adds ReadWriteTransaction including internal retries if the transaction is aborted. All statements that are executed on a ReadWriteTransaction are recorded internally in the transaction, and a running SHA-256 checksum is kept based on all the query results that are returned to the user. If the transaction is aborted by Cloud Spanner, a retry is automatically started and deemed successful if:
1. All DML statements return the same update count or the same error as during the original transaction.
2. All queries that returned an error during the original transaction return the same error during the retry.
3. The calculated checksums of all the rows that have been consumed of each result set returned by the transaction are equal.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
@@ -610,21 +610,7 @@ class ReadWriteTransaction extends AbstractMultiUseTransaction {
               ErrorCode.CANCELLED, "The statement was cancelled");
         }
         try {
-          try {
-            txContext = txManager.resetForRetry();
-          } catch (IllegalStateException e) {
-            if (txManager.getState()
-                == com.google.cloud.spanner.TransactionManager.TransactionState.STARTED) {
-              // In case the transaction manager didn't pick up the abort correctly, we'll rollback
-              // manually and do another try. This seems to be the case if the transaction was
-              // aborted and txContext.executeUpdate(Statement) is called.
-              // TODO: Check whether TransactionContextImpl.executeUpdate(Statement) handles
-              // AbortedException correctly.
-              txManager.rollback();
-              txManager = dbClient.transactionManager();
-              txContext = txManager.begin();
-            }
-          }
+          txContext = txManager.resetForRetry();
           // Inform listeners about the transaction retry that is about to start.
           invokeTransactionRetryListenersOnStart();
           // Then retry all transaction statements.


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
```diff
@@ -0,0 +1,1283 @@
...
+
+  @Override
+  public boolean isReadOnly() {
+    return false;
+  }
+
+  @Override
+  void checkValidTransaction() {
+    ConnectionPreconditions.checkState(
+        state == UnitOfWorkState.STARTED, "This transaction has status " + state.name());
```
kolea2: Would slightly change the precondition message to add acceptable status

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
```diff
@@ -0,0 +1,1283 @@
+/*
```
kolea2: This is a really big class. Is it possible to break out some of the package private classes into their own files?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
```diff
@@ -0,0 +1,1283 @@
...
+  public void write(Iterable<Mutation> mutations) {
+    Preconditions.checkNotNull(mutations);
+    checkValidTransaction();
+    for (Mutation mutation : mutations) {
+      this.mutations.add(checkNotNull(mutation));
+    }
+  }
+
+  /** Create a pseudo COMMIT statement to allow it to be cancelled or time out. */
+  private final ParsedStatement commitStatement =
```
kolea2: sorry, could you explain what this does?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
olavloite: I've moved the retriable queries and updates to separate class files. That significantly reduces the size of the `ReadWriteTransaction` file.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
olavloite: done

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
olavloite: I've added the following javadoc to the field:

```java
  /**
   * Create a COMMIT statement to use with the {@link #commit()} method to allow it to be
   * cancelled, time out or retried.
   *
   * {@link ReadWriteTransaction} uses the generic methods
   * {@link #executeAsync(ParsedStatement, Callable)} and {@link #runWithRetry(Callable)} to
   * allow statements to be cancelled, to timeout and to be retried. These methods require a
   * {@link ParsedStatement} as input. When the {@link #commit()} method is called directly, we do
   * not have a {@link ParsedStatement}, and the method uses this statement instead in order to use
   * the same logic as the other statements.
   */
```

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
```diff
@@ -0,0 +1,777 @@
...
+        try {
+          try {
+            txContext = txManager.resetForRetry();
+          } catch (IllegalStateException e) {
+            if (txManager.getState()
+                == com.google.cloud.spanner.TransactionManager.TransactionState.STARTED) {
+              // In case the transaction manager didn't pick up the abort correctly, we'll rollback
+              // manually and do another try. This seems to be the case if the transaction was
+              // aborted and txContext.executeUpdate(Statement) is called.
+              // TODO: Check whether TransactionContextImpl.executeUpdate(Statement) handles
```
kolea2: do we still need this TODO?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
```diff
@@ -0,0 +1,777 @@
...
+            }
+          }
+          // Inform listeners about the transaction retry that is about to start.
+          invokeTransactionRetryListenersOnStart();
+          // Then retry all transaction statements.
+          transactionRetryAttempts++;
+          for (RetriableStatement statement : statements) {
+            statement.retry(aborted);
+          }
+          successfulRetries++;
```
kolea2: should this be in the loop, or is it per batch of statements?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
olavloite: No, this related to a bug that was in the Spanner client library when the initial copy to the private repo was made. It has long been fixed (and if not, it will be picked up by a test case, as was how it was found in the first place).
It's removed.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
olavloite: It is per successful retry of an entire transaction, i.e. per batch of statements, so it should be outside of the loop.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ReadWriteTransaction.java
```diff
@@ -0,0 +1,763 @@
...
+      final ParsedStatement statement,
+      final AnalyzeMode analyzeMode,
+      final QueryOption... options) {
+    Preconditions.checkArgument(statement.isQuery(), "Statement is not a query");
+    checkValidTransaction();
+    try {
+      if (retryAbortsInternally) {
+        return asyncExecuteStatement(
+            statement,
+            new Callable<ResultSet>() {
```
kolea2: minor - can address in a future PR: let's extract out the Callable into a descriptive variable

PR #5901, opened 2019-07-26
Storage: Update conformance tests

BenWhitehead:
* Refactor V4SigningTest to use JUnits Parameterized Test runner.
  This allows running each of the tests as an individual test thereby
  providing success/ignore/failure information per test.
* Refactor V4SigningTest to use the generated Java Protobuf classes
  from the google-cloud-conformance-test module. The format read from is
  still json.
* Update to use the latest revision of the conformance-tests data.
  * One of the updated tests `Headers should be trimmed` does not
    currently pass and has been marked as ignore until there is a
    resolution to b/136171758 relating to the specification of how
    whitespace is to be handled when signing URLs.
  * I choose to move forward with the update, to reduce the drift
    between the test definitions in conformance-tests where other
    automation is being introduced to hopefully reduce the maintenance
    burden of conformance tests.


google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/V4SigningTest.java
```diff
@@ -54,77 +72,89 @@ public long millisTime() {
...
+  /**
+   * @param testData The serialized test data representing the test case.
+   * @param serviceAccountCredentials The credentials to use in this test.
+   * @param description Not used by the test, but used by the parameterized test runner as the name
+   *     of the test.
+   */
+  public V4SigningTest(
+      SigningV4Test testData,
+      ServiceAccountCredentials serviceAccountCredentials,
+      @SuppressWarnings("unused") String description) {
```
frankyn: It would be helpful to log out the name of the test being ran. Otherwise remove description given it's Test code and shouldn't impact the library.

BenWhitehead: The test name is written into the test results for the test suite, and is referenced by the `@Parameters(name = "{2}")` here: https://github.com/googleapis/google-cloud-java/pull/5901/files#diff-265f00ab3c7dd6cdaf7e8a7392158e52R132

In the JUnit.xml picked up by the build server it is rendered like this:
```xml
  <testcase name="test[Simple GET]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.04"/>
  <testcase name="test[Simple PUT]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.015"/>
  <testcase name="test[POST for resumable uploads]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.006"/>
  <testcase name="test[Vary expiration and timestamp]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.005"/>
  <testcase name="test[Vary bucket and object]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.005"/>
  <testcase name="test[Simple headers]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.006"/>
  <testcase name="test[Headers should be trimmed]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0">
    <skipped message="Test skipped until b/136171758 is resolved: got: &quot;test[Headers should be trimmed]&quot;, expected: is not &quot;test[Headers should be trimmed]&quot;"/>
  </testcase>
  <testcase name="test[Header value with multiple inline values]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.004"/>
  <testcase name="test[Customer-supplied encryption key]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.005"/>
  <testcase name="test[List Objects]" classname="com.google.cloud.storage.V4SigningTest(sponge_log)" time="0.004"/>
```

In IntelliJ it renders like this:
image

Unfortunately, when using the Parameterized runner the number of properties passed to the constructor has to match exactly with the number of values provided. And in the `@Parameters` only allows a value that will have `.toString` called on it, it doesn't allow us to do something like `@Parameters(name = "{1}.getDescription())")`.

frankyn: Thanks for clarifying.

BenWhitehead:
Thanks for the review @frankyn!

PR #5902, opened 2019-07-26
Refactor com.google.cloud.firestore.it.ITSystemTest.queryWatch to make it more reliable.

BenWhitehead:
### Changes
1. Separate document update logic from event listener code
2. Move assertions out of listener code
3. Add timeouts to all await conditions with associated assertions
   to clearly communicate the test fails due to an await, and which one.
4. Switch from using a semaphore to using multiple CountDownLatches

### Motivation
The queryWatch test has been flaky for some time, resulting in failure
most of the time. Due to these persistent failures the trust of the
firestore integration test suite has been low, and the results often
ignored.

After being able to run the test in a profileable environment, I ran the
old implementation repeatedly for 25 minutes resulting in a total of 136
runs, 100 of which failed -- each with the same error. The new
implementation was ran for 25 minutes resulting in a total of 2500 runs,
none of which failed.

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
+++ b/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
@@ -907,7 +907,7 @@ public class ITSystemTest {
     //      series of events and will then be asserted on.
     //
     // The mechanics of how the test is executed are as follows
-    //   A. All events delivered to the listener in 1 are added to a list of event we receive.
+    //   A. All events delivered to the listener in 1 are added to a list of events we receive.
     //   B. A separate ExecutorService is created and has tasks submitted to it to perform the
     //      series of document updates mentioned in 2. (Each document gets its own task).
     //      * Each of these update tasks will wait for the snapshot listener to be active before
@@ -965,7 +965,7 @@ public class ITSystemTest {
                 snapshotListerActive.await(5, TimeUnit.SECONDS);
                 // create the first document
                 doc1.set(map("baz", "foo")).get();
-                // update a field in the document
+                // update the document
                 doc1.set(map("foo", "bar")).get();
                 // add a field to the document
                 doc1.set(map("foo", "bar", "bar", "foo")).get();
@@ -989,7 +989,7 @@ public class ITSystemTest {
                 snapshotListerActive.await(5, TimeUnit.SECONDS);
                 // create a second document
                 doc2.set(map("foo", "bar")).get();
-                // update a field in the document
+                // update the document
                 doc2.set(map("foo", "foo")).get();
               } catch (InterruptedException | ExecutionException e) {
                 fail(String.format("Error while processing doc2: %s", e.getMessage()));


google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -889,95 +893,184 @@ public void onEvent(
...
-                  + "expectedEvents = [%s]%n"
-                  + "  actualEvents = [%s]%n",
-              expectedString, actualString),
-          tryAcquire);
+          query.addSnapshotListener(
+              new EventListener<QuerySnapshot>() {
+                @Override
+                public void onEvent(
+                    @Nullable QuerySnapshot value, @Nullable FirestoreException error) {
+                  receivedEvents.add(new ListenerEvent(value, error));
```
schmidt-sebastian: We should never see an `error` here. To simplify this test, I would add an assert here that the error is null (and maybe that `value` is not) and replace ListenerEvent with QuerySnapshot.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -889,95 +893,184 @@ public void onEvent(
...
+      updatesExecutor.submit(
+          new Runnable() {
+            DocumentReference doc1 = randomColl.document("doc1");
+
+            @Override
+            public void run() {
+              try {
+                // create the first document
+                doc1.set(map("baz", "foo")).get();
+                // update a field in the document
```
schmidt-sebastian: s/a field in//

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -889,95 +893,184 @@ public void onEvent(
...
+              } catch (InterruptedException | ExecutionException e) {
+                fail(String.format("Error while processing doc2: %s", e.getMessage()));
+              } finally {
+                doc2CDL.countDown();
+              }
+            }
+          });
+
+      // Wait for the document update operations to be performed
+      final boolean doc1CDLAwait = doc1CDL.await(10, TimeUnit.SECONDS);
```
schmidt-sebastian: If I am not mistaken, doc1CDLAwait and doc2CDLAwait could be a single count down latch initialized with `new CountDownLatch(2)`.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -889,95 +893,184 @@ public void onEvent(
...
-                    public void onEvent(
-                        @Nullable QuerySnapshot value, @Nullable FirestoreException error) {
-                      System.out.printf("onEvent(value : %s, error : %s)%n", value, error);
-                      try {
-                        switch (semaphore.availablePermits()) {
-                          case 0:
-                            actualEvents.add("event 0");
-                            assertTrue(value.isEmpty());
-                            ref1 = randomColl.add(map("foo", "foo")).get();
-                            ref2 = randomColl.add(map("foo", "bar")).get();
```
schmidt-sebastian: I think the problem with this test is that these two adds are not guaranteed to generate two different events. One way of fixing this would be either split these up into two tasks or use a WriteBatch for both writes, which is guaranteed to only raise one event of size 2.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
BenWhitehead: Asserts inside the snapshotListener unfortunately do not automatically cascade outside the listener and as such can't fail the test, that I why I pulled it up to the top level and added the assertion after all the operations.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
BenWhitehead: It is true we could do this, but then we would lose the info about which of the document operations timed out for the assertion. I'd like to keep the extra detail in case the test ever fails due to this.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
BenWhitehead: This scenario feels a bit like a different test case to me. I can look at adding another test if that sounds good to you?

schmidt-sebastian: It's different for sure - but this test is just meant to show that basic Watch functionality works. The actual steps that we verify are of minor importance.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
schmidt-sebastian: What happens if we don't see the right number of snapshots (we get an error but we ignore it and don't add a receivedEvent)? My only reservation with this PR is that the generic nature of it adds a lot of code that makes the event flow somewhat hard to parse. I'm totally biased here, but the original test seems significantly easier to follow.

BenWhitehead: When I started investigating why this tests would fail ~65% of the time in CI I narrowed it down to the fact that the semaphore wasn't ever being satisfied. In #5746 I added a timeout as well as some diagnostic information to the test and found that the failure always occurred in the `case 0` brach and resulted in the other branches never being executed. To ensure that performing new document updates wouldn't ever block the processing of received I separated the update and the lister as presented in this PR. I have ran this new implementation continuously for 2,500 iterations and all 2,500 runs passed.

I agree that it is more code -- a fair amount of it would be reduced if we were able to use java 1.8 constructs (which will hopefully happen sooner rather than later) -- and not super straightforward from a cursory glance, but I don't think the original test was either. One advantage with this new implementation is that it's more explicit about what the interrelationships are between the listener and document updates. In addition, there are explicit timeouts defined for each type of operation so we don't ever hang the entire build, and having the assertions outside the listeners means if there is a failure it will actually fail the test.

If it isn't important for `doc1` and `doc2` to be updated independent from one another I can merge those workers. If we want to avoid the anonymous inner classes from `FluentIterable` I can rewrite that with for loops.

PR #5907, opened 2019-07-29
spanner-jdbc: Step 22 - DdlBatch

olavloite:
Adds DdlBatch. A DdlBatch can be used to execute a set of DDL statements as one batch. A DDL batch is not atomic, and it could be that only some of the statements will succeed. DdlBatch is implemented as a `UnitOfWork` on a connection, and there can only be one `UnitOfWork` active at any time on a connection. This means that when a connection is executing a DDL batch, it will not be executing any other transaction, and only DDL statements are allowed.

A DDL batch can be executed both through calling specific methods on a connection instance, as well as by using the auxiliary SQL statements `START BATCH DDL` and  `RUN BATCH` / `ABORT BATCH`.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DdlBatch.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DdlBatch.java
@@ -31,6 +31,7 @@ import com.google.cloud.spanner.SpannerExceptionFactory;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.jdbc.StatementParser.ParsedStatement;
 import com.google.cloud.spanner.jdbc.StatementParser.StatementType;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.spanner.admin.database.v1.UpdateDatabaseDdlMetadata;
 
@@ -184,15 +185,7 @@ class DdlBatch extends AbstractBaseUnitOfWork {
               // Return metadata.
               return operation.getMetadata().get();
             } catch (Exception e) {
-              UpdateDatabaseDdlMetadata metadata = operation.getMetadata().get();
-              long[] updateCounts = new long[metadata.getStatementsCount()];
-              for(int i = 0; i < updateCounts.length; i++) {
-                if(metadata.getCommitTimestampsCount() > i && metadata.getCommitTimestamps(i) != null) {
-                  updateCounts[i] = 1L;
-                } else {
-                  updateCounts[i] = 0L;
-                }
-              }
+              long[] updateCounts = extractUpdateCounts(operation.getMetadata().get());
               throw SpannerExceptionFactory.newSpannerBatchUpdateException(
                   ErrorCode.INVALID_ARGUMENT,
                   e.getMessage(),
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DdlBatch.java
```diff
@@ -0,0 +1,211 @@
...
+          public Void call() throws Exception {
+            OperationFuture<Void, UpdateDatabaseDdlMetadata> operation =
+                ddlClient.executeDdl(statements);
+            return operation.get();
+          }
+        };
+        asyncExecuteStatement(RUN_BATCH, callable);
+      }
+      this.state = UnitOfWorkState.RAN;
+      return new long[0];
```
kolea2: Shouldn't this return the number of successful updates?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/DdlBatchTest.java
```diff
@@ -0,0 +1,372 @@
...
+    public boolean matches(Object list) {
+      return ((List<String>) list).size() == size;
+    }
+  }
+
+  @Test
+  public void testRunBatch() {
+    DdlClient client = createDefaultMockDdlClient();
+    DdlBatch batch = createSubject(client);
+    batch.runBatch();
```
kolea2: following my previous comment, should we add assertions of the return type for `runBatch()`?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DdlBatch.java
olavloite: Changed to return an array containing an indication per statement whether the statement was successfully executed (1) or not (0).

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/DdlBatchTest.java
olavloite: Added two additional test cases to check for successful and failed batches and the actual return values.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DdlBatch.java
```diff
@@ -0,0 +1,233 @@
...
+          public UpdateDatabaseDdlMetadata call() throws Exception {
+            OperationFuture<Void, UpdateDatabaseDdlMetadata> operation =
+                ddlClient.executeDdl(statements);
+            try {
+              // Wait until the operation has finished.
+              operation.get();
+              // Return metadata.
+              return operation.getMetadata().get();
+            } catch (Exception e) {
+              UpdateDatabaseDdlMetadata metadata = operation.getMetadata().get();
```
kolea2: let's extract into a private method - `computeUpdateCounts`?

PR #5908, opened 2019-07-29
spanner-jdbc: Step 23 - DmlBatch

olavloite:
Adds DmlBatch. DML batches are similar to DDL batches, except they execute DML instead of DDL statements. DML batches are executed like this:

```sql
START BATCH DML;
UPDATE FOO SET BAR=1 WHERE ID=2;
RUN BATCH;
```

An active batch can be aborted by executing the `ABORT BATCH` statement.

DML batches can be executed while the transaction is in autocommit mode, which means that the effects of the batch will be visible to all other users directly after the `RUN BATCH` statement has finished executing. DML batches can also be executed as part of a read/write transaction. As a connection only allows one `UnitOfWork` to be active at any time, and both a read/write transaction and a DML batch is a `UnitOfWork`, the DML batch is executed as an nested batch of the active transaction. When `RUN BATCH` is executed, the batch of DML statements are executed on the active read/write transaction, and its effects will only be visible to other users once the transaction has been committed.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DmlBatch.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DmlBatch.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google LLC
+ * Copyright 2019 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/DmlBatch.java
```diff
@@ -0,0 +1,190 @@
...
+    Preconditions.checkArgument(update.getType() == StatementType.UPDATE,
+        "Only DML statements are allowed. \"" + update.getSqlWithoutComments()
+            + "\" is not a DML-statement.");
+    statements.add(update);
+    return -1L;
+  }
+
+  @Override
+  public long[] executeBatchUpdate(Iterable<ParsedStatement> updates) {
+    throw SpannerExceptionFactory.newSpannerException(ErrorCode.FAILED_PRECONDITION,
```
kolea2: This is surprising to me from an API perspective. We can call executeUpdate, then runBatch(), but I would have expected this to be somewhat of a convenience method?

olavloite: I understand the confusion, but `DmlBatch` should be seen as a kind of wrapper around the fact that both standard SQL and JDBC do not have any `executeBatchUpdate(Iterable)` equivalent. Instead, both work by declaring the start of a batch, then add statements to the batch, and then execute the batch.

In case of SQL:

```sql
START BATCH DML;
UPDATE FOO SET BAR=1 WHERE ID=1;
UPDATE FOO SET BAR=2 WHERE ID=2;
RUN BATCH;
```

The following (or similar) is not possible in standard SQL:

```sql
EXECUTE('UPDATE FOO SET BAR=1 WHERE ID=1', 'UPDATE FOO SET BAR=2 WHERE ID=2')
```

Much the same applies to JDBC:

```java
Statement statement = createStatemet(...);
statement.addBatch("UPDATE FOO SET BAR=1 WHERE ID=1");
statement.addBatch("UPDATE FOO SET BAR=2 WHERE ID=2");
statement.executeBatch();
```

Allowing it as a convenience method could maybe have been an option if it was a public interface, but as this is an internal interface my opinion is that we should be strict when it comes to correct usage.

kolea2: Thanks, that makes sense! Also didn't realize this wasn't a public API 😄

PR #5912, opened 2019-07-29
spanner-jdbc: Step 27 - JdbcDriver

olavloite:
Adds the actual JDBC Driver, including the service registration file for the driver.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcDriverTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcDriverTest.java
@@ -43,7 +43,7 @@ public class JdbcDriverTest {
   @Test
   public void testConnect() throws SQLException {
     try (Connection connection = DriverManager.getConnection(
-        "jdbc:cloudspanner://localhost:8443/projects/test-project/instances/static-test-instance/databases/test-database;credentials=/home/loite/CloudSpannerKeys/cloudspanner-emulator-key.json")) {
+        "jdbc:cloudspanner:/projects/test-project/instances/static-test-instance/databases/test-database;credentials=/path/to/key.json")) {
       assertThat(connection.isClosed(), is(false));
     }
   }
@@ -51,7 +51,7 @@ public class JdbcDriverTest {
   @Test(expected = SQLException.class)
   public void testInvalidConnect() throws SQLException {
     try (Connection connection = DriverManager.getConnection(
-        "jdbc:cloudspanner://localhost:8443/projects/test-project/instances/static-test-instance/databases/test-database;credentialsUrl=/home/loite/CloudSpannerKeys/cloudspanner-emulator-key.json")) {
+        "jdbc:cloudspanner:/projects/test-project/instances/static-test-instance/databases/test-database;credentialsUrl=/path/to/key.json")) {
       assertThat(connection.isClosed(), is(false));
     }
   }


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcDriverTest.java
```diff
@@ -0,0 +1,59 @@
...
+    } catch (ClassNotFoundException e) {
+      throw new IllegalStateException(
+          "JdbcDataSource failed to load com.google.cloud.spanner.jdbc.JdbcDriver", e);
+    }
+  }
+
+  @Test
+  public void testConnect() throws SQLException {
+    try (Connection connection = DriverManager.getConnection(
+        "jdbc:cloudspanner://localhost:8443/projects/test-project/instances/static-test-instance/databases/test-database;credentials=/home/loite/CloudSpannerKeys/cloudspanner-emulator-key.json")) {
```
kolea2: Think this needs to be changed

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/resources/META-INF/services/java.sql.Driver
```diff
@@ -0,0 +1 @@
+com.google.cloud.spanner.jdbc.JdbcDriver
```
kolea2: where is this used?

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcDriverTest.java
olavloite: Changed test connection URL's.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/resources/META-INF/services/java.sql.Driver
olavloite: This is used by the Java service loader. See the answer to this question: https://stackoverflow.com/questions/18921352/using-meta-inf-services-for-internal-plumbing-of-driver

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/JdbcDriverTest.java
```diff
@@ -0,0 +1,59 @@
...
+      throw new IllegalStateException(
+          "JdbcDataSource failed to load com.google.cloud.spanner.jdbc.JdbcDriver", e);
+    }
+  }
+
+  @Test
+  public void testConnect() throws SQLException {
+    try (Connection connection = DriverManager.getConnection(
+        "jdbc:cloudspanner:/projects/test-project/instances/static-test-instance/databases/test-database;credentials=/path/to/key.json")) {
+      assertThat(connection.isClosed(), is(false));
```
kolea2: any teardown we have to do here?

olavloite: No, not other than the auto close of the connection that is handled by the try-with-resources block..

Normally `SpannerPool.closeSpannerPool()` should be called at the end of an application after all connections have been closed, but the end of this test case is not necessarily the end of the application. As there are other tests in the entire test suite that deliberately leak connections, calling `SpannerPool.closeSpannerPool()` here could interfere with those test cases when all test cases are executed in one go.

olavloite: There is also no need to call `DriverManager.deregisterDriver(Driver)`. The `JdbcDriver` class is not holding on to any resources that need to be closed, and calling `DriverManager.deregisterDriver(Driver)` is something that should normally only be done by the driver itself.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDriver.java
```diff
@@ -16,12 +16,225 @@
...
+  static {
+    try {
+      register();
+    } catch (SQLException e) {
+      java.sql.DriverManager.println("Registering driver failed: " + e.getMessage());
+    }
+  }
+  private static JdbcDriver registeredDriver;
+
+  static void register() throws SQLException {
```
kolea2: Sorry, one last question. How come this is different than the class loading done in JdbcDataSource?

olavloite: The JdbcDataSource is not responsible for registering the JDBC driver. That is the responsibility of the driver itself, and is done automatically when the driver class (JdbcDriver) is loaded. In most newer environments this is done automatically thanks to the service registration in META-INF/services/java.sqlDriver. The JdbcDataSource does a manual load of the driver class to be 100% certain that the driver is loaded and registered, although in most cases it will be redundant.

PR #5913, opened 2019-07-29
spanner-jdbc: Step 28 - JdbcDatabaseMetadata

olavloite:
Adds JDBC database metadata.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDatabaseMetaData.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDatabaseMetaData.java
@@ -29,6 +29,9 @@ import java.sql.Types;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Scanner;
+import com.google.auth.Credentials;
+import com.google.auth.ServiceAccountSigner;
+import com.google.auth.oauth2.UserCredentials;
 import com.google.cloud.spanner.ResultSets;
 import com.google.cloud.spanner.Struct;
 import com.google.cloud.spanner.Type;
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDatabaseMetaData.java
```diff
@@ -0,0 +1,1390 @@
...
+  }
+
+  @Override
+  public String getURL() throws SQLException {
+    return connection.getConnectionUrl();
+  }
+
+  @Override
+  public String getUserName() throws SQLException {
+    // TODO: extract from credentials
```
kolea2: still have a todo here?

olavloite: Added implementation for `getUserName()` for known `Credentials` types.

olavloite: I've left the `getFunctions(String, String, String)` method as is (i.e. returning an empty `ResultSet`), as there is no server side equivalent for this in Cloud Spanner, and this is hardly ever used

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDatabaseMetaData.java
```diff
@@ -0,0 +1,1400 @@
...
+    return false;
+  }
+
+  @Override
+  public RowIdLifetime getRowIdLifetime() throws SQLException {
+    return RowIdLifetime.ROWID_UNSUPPORTED;
+  }
+
+  @Override
+  public ResultSet getSchemas(String catalog, String schemaPattern) throws SQLException {
```
kolea2: do we have tests for this (and the other sql query methods)?

olavloite: Yes, but they are created as integration tests (`ITJdbcDatabaseMetaDataTest` in the private repo), so they will be added in a later PR.

PR #5915, opened 2019-07-29
feat: allow endpoint override on Publisher/Subscriber

### LINKED ISSUE #5792: PubSub: support `setEndpoint` on Publisher/Subscriber builders
Currently, the underlying clients can customize the gRPC endpoint but that option is not surfaced to the higher-level handwritten layer.

chingor13:
Fixes #5792

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
@@ -155,7 +155,8 @@ public class Publisher {
         PublisherStubSettings.newBuilder()
             .setCredentialsProvider(builder.credentialsProvider)
             .setExecutorProvider(FixedExecutorProvider.create(executor))
-            .setTransportChannelProvider(builder.channelProvider);
+            .setTransportChannelProvider(builder.channelProvider)
+            .setEndpoint(builder.endpoint);
     stubSettings
         .publishSettings()
         .setRetryableCodes(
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java
```diff
@@ -155,7 +155,8 @@ private Publisher(Builder builder) throws IOException {
         PublisherStubSettings.newBuilder()
             .setCredentialsProvider(builder.credentialsProvider)
             .setExecutorProvider(FixedExecutorProvider.create(executor))
-            .setTransportChannelProvider(builder.channelProvider);
+            .setTransportChannelProvider(builder.channelProvider)
```
kamalaboulhosn: Before the existence of setEndpoint, one would create a custom TransportChannelProvider and set the endpoint in there. If one does that and calls setEndpoint, which one will win?

chingor13: If the `TransportChannelProvider#needsEndpoint()`, then the value configured with `setEndpoint()` will take effect via `TransportChannelProvider#withEndpoint(String)`. So a custom `TransportChannelProvider` that specifies an endpoint would not be overwritten.

See https://github.com/googleapis/gax-java/blob/fd5740a2726d0ab3f3124351ed0384911fa8f5c2/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L154-L156

PR #5916, opened 2019-07-29
Firestore: Add @Deprecated annotation to getCollections

schmidt-sebastian:
These methods are deprecated as per the JavaDoc but we should probably use an Annotation as well.

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
@@ -403,7 +403,7 @@ public class DocumentReference {
   /**
    * Fetches the subcollections that are direct children of this document.
    *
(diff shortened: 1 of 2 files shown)


BenWhitehead:
I'm okay with the additional detail, but the javadocs don't seem to provide an alternative of how the users is supposed to be able to do this sort of operation. Do you know what the replacement is that we could refer to?

schmidt-sebastian:
The JavaDoc's deprecated comments do say to use `listCollections()`. GitHub decided that is not worth showing :/ Is there something else we should be doing?

BenWhitehead:
Ah, so it does my mistake. LGTM.

BenWhitehead:
Actually, can we change the deprecation javadocs to use this format:
```
   * @deprecated Use {@link #listCollections()}.
```

Instead of the back-ticks. This will make the link clickable in the javadocs if we do it.

schmidt-sebastian:
Updated the JavaDoc to use the links as suggested.

PR #5924, opened 2019-07-30
Firestore: add emulator support

thebrianchen:
Context b/134525351

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
@@ -35,7 +35,11 @@ import com.google.common.collect.ImmutableSet;
 import io.grpc.ManagedChannelBuilder;
 import java.io.IOException;
 import java.net.URI;
-import java.util.*;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
@@ -225,10 +229,9 @@ public final class FirestoreOptions extends ServiceOptions<Firestore, FirestoreO
       // Override credentials and channel provider if we are using the emulator.
       String emulatorHost = System.getenv(FIRESTORE_EMULATOR_SYSTEM_VARIABLE);
       if (emulatorHost != null) {
-        String hostUrlString = "http://" + emulatorHost;
-
         // Try creating a host in order to validate that the host name is valid.
         try {
+          String hostUrlString = "http://" + emulatorHost;
           URI.create(hostUrlString);
         } catch (Exception e) {
           throw new IllegalArgumentException(


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -208,6 +216,32 @@ public Builder setTimestampsInSnapshotsEnabled(boolean value) {
...
+                        public ManagedChannelBuilder apply(ManagedChannelBuilder input) {
+                          input.usePlaintext();
+                          return input;
+                        }
+                      })
+                  .build());
+          setCredentialsProvider(FixedCredentialsProvider.create(new FakeCredentials()));
+        } catch (MalformedURLException e) {
+          throw new IllegalArgumentException(
+              "Value for property FIRESTORE_EMULATOR_HOST is not a valid host", e);
```
BenWhitehead: Can we make this error message more informative?

> The environment variable FIRESTORE_EMULATOR_HOST is unable to be resolved.

Lets also use the constant here instead of duplicating the value.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -208,6 +216,32 @@ public Builder setTimestampsInSnapshotsEnabled(boolean value) {
     @Override
     @Nonnull
     public FirestoreOptions build() {
+      if (System.getProperty(FIRESTORE_EMULATOR_SYSTEM_PROPERTY) != null) {
```
schmidt-sebastian: I wasted the last 10 minutes on this, and I couldn't come to a conclusion. If it is possible to change environment variables after the fact (it looks like it might be on windows) then checking this twice might be racy. But it's probably not.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -208,6 +216,32 @@ public Builder setTimestampsInSnapshotsEnabled(boolean value) {
...
+                        @Override
+                        public ManagedChannelBuilder apply(ManagedChannelBuilder input) {
+                          input.usePlaintext();
+                          return input;
+                        }
+                      })
+                  .build());
+          setCredentialsProvider(FixedCredentialsProvider.create(new FakeCredentials()));
+        } catch (MalformedURLException e) {
+          throw new IllegalArgumentException(
```
schmidt-sebastian: Can we move this closer to where the exception happens and maybe even include the value? Thanks!

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -208,6 +216,32 @@ public Builder setTimestampsInSnapshotsEnabled(boolean value) {
...
     @Nonnull
     public FirestoreOptions build() {
+      if (System.getProperty(FIRESTORE_EMULATOR_SYSTEM_PROPERTY) != null) {
+        String emulatorHost = System.getProperty(FIRESTORE_EMULATOR_SYSTEM_PROPERTY);
+        String hostUrlString = "http://" + emulatorHost;
+        // Try creating a host in order to validate that the host name is valid.
+        try {
+          new URL(hostUrlString);
+          setHost(emulatorHost);
+          setChannelProvider(
```
schmidt-sebastian: Nit: add a comment that says that this turns off SSL

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -208,6 +216,32 @@ public Builder setTimestampsInSnapshotsEnabled(boolean value) {
     @Override
     @Nonnull
     public FirestoreOptions build() {
+      if (System.getProperty(FIRESTORE_EMULATOR_SYSTEM_PROPERTY) != null) {
+        String emulatorHost = System.getProperty(FIRESTORE_EMULATOR_SYSTEM_PROPERTY);
+        String hostUrlString = "http://" + emulatorHost;
+        // Try creating a host in order to validate that the host name is valid.
+        try {
+          new URL(hostUrlString);
```
BenWhitehead: Is the intention here to force a DNS probe, or just to validate that the string given to us is well formatted? If we aren't trying to force a DNS probe, please use `URI.create` here instead of `new URL`.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -48,6 +54,8 @@
 
   private static final long serialVersionUID = -5853552236134770090L;
 
+  private static final String FIRESTORE_EMULATOR_SYSTEM_PROPERTY = "FIRESTORE_EMULATOR_HOST";
```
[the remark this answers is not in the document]
thebrianchen: @schmidt-sebastian thoughts on this?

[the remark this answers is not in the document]
ryanpbrewster: Are you suggesting changing the name of the environment variable? Of the name of this Java `private static final String` constant?

The environment variable is pretty much stuck at this point, there are too many clients relying on it to change it now.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -216,8 +221,74 @@ public FirestoreOptions build() {
...
+          URI.create(hostUrlString);
+        } catch (Exception e) {
+          throw new IllegalArgumentException(
+              "Value: '"
+                  + emulatorHost
+                  + "' for property "
+                  + FIRESTORE_EMULATOR_SYSTEM_PROPERTY
+                  + " is not a valid host",
+              e);
+        }
```
ryanpbrewster: can we extract this into a helper method `validateUrl`?

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -216,8 +221,74 @@ public FirestoreOptions build() {
...
+        // Setting the channel provider to use plaintext turns off SSL.
+        this.setChannelProvider(
+            InstantiatingGrpcChannelProvider.newBuilder()
+                .setEndpoint(emulatorHost)
+                .setChannelConfigurator(
+                    new ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder>() {
+                      @Override
+                      public ManagedChannelBuilder apply(ManagedChannelBuilder input) {
+                        input.usePlaintext();
+                        return input;
```
ryanpbrewster: can we `return input.usePlaintext()` here?

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -216,8 +221,74 @@ public FirestoreOptions build() {
...
+        }
+
+        // Setting the channel provider to use plaintext turns off SSL.
+        this.setChannelProvider(
+            InstantiatingGrpcChannelProvider.newBuilder()
+                .setEndpoint(emulatorHost)
+                .setChannelConfigurator(
+                    new ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder>() {
+                      @Override
+                      public ManagedChannelBuilder apply(ManagedChannelBuilder input) {
```
ryanpbrewster: any chance this repo supports java8 lambdas? this is much cleaner with them

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -216,8 +221,74 @@ public FirestoreOptions build() {
...
+                    })
+                .build());
+        this.setCredentials(new FakeCredentials());
+      }
+
       return new FirestoreOptions(this);
     }
+
+    public class FakeCredentials extends Credentials {
+      private Map<String, List<String>> headers;
```
ryanpbrewster:
```
private static final Map<String, List<String>> HEADERS = ImmutableMap.of("Authorization", ImmutableList.of("Bearer owner"))
```

we have Guava around, might as well use it

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -216,8 +221,74 @@ public FirestoreOptions build() {
...
+
+      public FakeCredentials() {
+        headers = new HashMap<>();
+        headers.put("Authorization", Arrays.asList("Bearer owner"));
+      }
+
+      @Override
+      public String getAuthenticationType() {
+        return "";
```
ryanpbrewster: throw an IllegalStateException here, no one should ever call this method

we're returning a nonsense value here, IMO better to immediately fail than let this propagate around

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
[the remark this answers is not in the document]
BenWhitehead: Changing the name of the java constant. Sorry for not clarifying.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
BenWhitehead: Unfortunately, google-cloud-java is still at java 1.7 compatibility and we can't use lambdas.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
thebrianchen: We're not using this anywhere else, and the error code is pretty specific in that it's checking for the host.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
thebrianchen: I wanted to use lambdas, but it's locked on java 7 here.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
thebrianchen: I initially tried that when I started, but ran into the error
```
Error:(263, 72) java: incompatible types: com.google.common.collect.ImmutableMap<java.lang.String,com.google.common.collect.ImmutableList<java.lang.String>> cannot be converted to java.util.Map<java.lang.String,java.util.List<java.lang.String>>
```

ryanpbrewster: how about
```
ImmutableMap.of("Authorization", Arrays.asList("Bearer owner"))
```
?

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -216,8 +222,69 @@ public FirestoreOptions build() {
         }
       }
 
+      // Override credentials and channel provider if we are using the emulator.
+      String emulatorHost = System.getenv(FIRESTORE_EMULATOR_SYSTEM_VARIABLE);
+      if (emulatorHost != null) {
+        String hostUrlString = "http://" + emulatorHost;
```
schmidt-sebastian: Nit: Move this into the try/catch block to make sure this is never used anywhere else, especially not in `setEndpoint()`.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java
```diff
@@ -16,22 +16,26 @@
...
 import com.google.cloud.firestore.v1.FirestoreSettings;
 import com.google.cloud.grpc.GrpcTransportOptions;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import io.grpc.ManagedChannelBuilder;
 import java.io.IOException;
-import java.util.Objects;
-import java.util.Set;
+import java.net.URI;
+import java.util.*;
```
schmidt-sebastian: Can you expand the import?

SylvainAssemat:
Hi all
I dont know if i could post my trouble here about the **PERMISSION_DENIED: Metadata operations require admin authentication** when i try to use the firestore emulator

```
[firestore]    export FIRESTORE_EMULATOR_HOST=127.0.0.1:8772
[firestore]
[firestore] Dev App Server is now running.
```

I can not find a working solution for my spring boot project.
I can not find on the web any clue or snippet.

Can you help me ?

```
Firestore db = FirestoreOptions.newBuilder()
		.setHost("localhost:8772")
		.setProjectId("xxxxx")
		//.setCredentials(GoogleCredentials.getApplicationDefault())
		//.setHeaderProvider(FixedHeaderProvider.create("Authorization", "Bearer owner"))
		.setCredentials(credentials)
		//.setDatabaseId("(default)")
		.setCredentialsProvider(FixedCredentialsProvider.create(credentials))
		.build().getService();
```

The CURL is OK (http 200)
```
curl -X POST \
  http://localhost:8772/v1/projects/xxxxx/databases/(default)/documents:listCollectionIds \
  -H 'Authorization: Bearer owner' \
```

_google-cloud-core-grpc-1.90.0
google-cloud-firestore-1.20.0
google-cloud-bom 0.108.0-alpha_

schmidt-sebastian:
You need to provide the `Authorization: Bearer owner` header for ListDocuments and ListCollectionIDs (which you have commented out). Your best bet is to just use the emulator environment variable, which the SDK uses to configure itself. In you case this seems to be `FIRESTORE_EMULATOR_HOST=localhost:8772`.

It's usually best to open a GitHub issue for questions like these since it is pretty easy to miss a comment on a merged GitHub PR :)

PR #5954, opened 2019-07-31
spanner-jdbc: Set clientLibToken on RPC instance

olavloite:
The client lib token must be set as a header on for the RPC instance.


google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -382,4 +383,27 @@ public void testDoNotCacheClosedSpannerInstance() {
     assertThat(service3.isClosed()).isFalse();
     service3.close();
   }
+
+  @Test
+  public void testSetClientLibToken() {
```
kolea2: let's also add an assertion for not setting the client lib and the value returned

olavloite: That assertion is on line 402.

PR #5959, opened 2019-08-01
Move setClientLibToken to ServiceOptions.

kolea2:
After discussion with @chingor13, moving this API to ServiceOptions to use more broadly.


chingor13:
cc @dazuma

google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -253,6 +258,18 @@ public B setHeaderProvider(HeaderProvider headerProvider) {
...
+    @InternalApi
+    public B setClientLibToken(String clientLibToken) {
+      Preconditions.checkArgument(
+          getAllowedClientLibTokens().contains(clientLibToken), "Illegal client lib token");
+      this.clientLibToken = clientLibToken;
+      return self();
+    }
+
+    protected Set<String> getAllowedClientLibTokens() {
+      return ImmutableSet.of(ServiceOptions.getGoogApiClientLibName());
```
chingor13: this could probably be a constant

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -178,6 +172,14 @@ public Builder setTransportOptions(TransportOptions transportOptions) {
       return super.setTransportOptions(transportOptions);
     }
 
+    @Override
+    protected Set<String> getAllowedClientLibTokens() {
+      return ImmutableSet.of(
```
chingor13: This could also probably remain a constant

PR #5996, opened 2019-08-02
spanner-jdbc-cleanup - Step 07 - Increase timeout when creating a new test instance

olavloite:
When working on a local development environment in Europe, the test environment will automatically try to create an instance in a nearby configuration (europe/west). A timeout of 500 milliseconds is (way!) too short for that. This change ensures the integration tests do not fail with a timeout when executed from such an environment.


kolea2:
LGTM just waiting for builds to pass

PR #5997, opened 2019-08-02
spanner-jdbc-cleanup - Step 08 -  Use Spanner IT test env

olavloite:
Change the integration tests for the JDBC driver and generic connection API to use standard Spanner IT test environment. This PR also adds two simple IT's to show the working:
* ITBulkConnectTest: Create a large number of generic Spanner connections and close these, and show that the SpannerPool can be closed down after closing all connections.
* ITJdbcConnectTest: Create a JDBC connection in different ways.

This PR also fixes a small bug in the connection options: The regex for connection URL's did not allow database names with an underscore. This is now fixed.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ITAbstractJdbcTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ITAbstractJdbcTest.java
@@ -16,8 +16,6 @@
 
 package com.google.cloud.spanner.jdbc;
 
-import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/test/java/com/google/cloud/spanner/jdbc/ITAbstractJdbcTest.java
```diff
@@ -0,0 +1,205 @@
...
+        "SELECT INDEX_NAME FROM INFORMATION_SCHEMA.INDEXES WHERE UPPER(TABLE_NAME)=? AND UPPER(INDEX_NAME)=?")) {
+      ps.setString(1, table);
+      ps.setString(2, index);
+      try (ResultSet rs = ps.executeQuery()) {
+        return rs.next();
+      }
+    }
+  }
+
+  List<String> readStatementsFromFile(String filename) {
```
kolea2: I think this is somewhere else in the tests. Can we use a Util class?

olavloite: You're right, it was already in the `AbstractSqlScriptVerifier` class. I've changed the tests to use that method.

PR #6002, opened 2019-08-02
spanner-jdbc-cleanup - Step 12 - Add integration tests for JDBC database metadata

olavloite:
Adds integration tests for the Spanner implementation for JDBC `DatabaseMetadata`.


olavloite:
Close in favor of new test case based on master.

PR #6038, opened 2019-08-06
Update protobuf to 3.9.1

elharo:


BenWhitehead:
Doesn't this change require a corresponding bump to synthtool so that we have protoc 3.9.1 there?

chingor13:
In theory, the generated code from an older version of protoc should be compatible with the newest protobuf version at runtime. The inverse is not the case.

I will run this through all the integration tests (they are skipped in the presubmit here)

chingor13:
Spanner IT and Spanner JDBC IT tests failed.

Spanner test failed with "INVALID_ARGUMENT: com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: projects/{project ID} is missing from uri."
and the JDBC IT test failed in `ITTransactionRetryTest.testAbortWithConcurrentInsert`. I rerunning to see if it was flakey

chingor13:
gprc 1.23.0 has updated to 3.9

PR #6047, opened 2019-08-07
spanner-jdbc: Add beta note to README and cleanup pom.xml

kolea2:


kolea2:
@chingor13 please take another quick look. Thanks!

Issue #3263, opened 2018-05-14
Unit test firestore: Mocking DocumentReference and CollectionReference


einevea:
We are trying to test classes with dependencies on firestore and we found we cannot mock these classes because their constructors are internal. Is it possible if the following methods return interfaces instead of classes: 

val collectionReference:CollectionReference = db.collection(collectionName)
val document: DocumentReference = collectionReference.document(docId)

schmidt-sebastian:
Thanks for sending this over. I filed an internal issue and will update you as we make progress.

mustafaakin:
We need them too for unit testing purposes

AbhishekChugh:
Looks like the constructor of CollectionReference was missed in the fix

schmidt-sebastian:
Follow up PR: https://github.com/googleapis/google-cloud-java/pull/6054

PR #6053, opened 2019-08-08
Elevate access level to support CollectionReference mocking

abchugh:
In the fix to #3263 CollectionReference was missed


AbhishekChugh:
@googlebot I signed it!

PR #6062, opened 2019-08-11
storage: add benchwrapper

jeanbza:
This adds a small gRPC server around the google-cloud-storage client library,
which allows g3 benchmarking code to prod at the library and record
measurements without needing to be written in Java / be in the same repository.

Since g3 requires TLS on all connections, the server is TLS-ed with a dummy
cert. These certs aren't intended to be private: the kokoro environment
is the only place this gets run. (never in prod, etc)

ps: I'm not sure what the right place to put this code is. Minimally it requires the local version of google-cloud-storage, so I stuck it alongside it. Perhaps there's a better place, with a way to access google-cloud-storage without going to maven though.

### DIFF
--- a/google-cloud-testing/storage-benchwrapper/README.md
+++ b/google-cloud-testing/storage-benchwrapper/README.md
@@ -1,0 +1,54 @@
+# storage-benchwrapper
+
+storage-benchwrapper is a gRPC wrapper around the storage library for benchmarking purposes.
+
+## Running
+
+```
+mvn clean install -DskipTests=true (one time)
+cd google-cloud-testing/storage-benchwrapper
+export STORAGE_EMULATOR_HOST=localhost:8080
+mvn clean install exec:java -DskipTests=true -Dport=8081
+```
+
+## Generating .proto sources
+
+Sources are generated as part of the protobuf-maven-plugin plugin, but if you'd
+like to generate them yourself to see the output you can run:
+
+```
+cd google-cloud-testing/storage-benchwrapper
+protoc \
+    --plugin=protoc-gen-grpc-java=build/exe/java_plugin/protoc-gen-grpc-java \
+    --java_out=src/main/java \
+    --grpc-java_out=src/main/java \
+    --proto_path=src/main/proto \
+    src/main/proto/*.proto
+```
+
+Note that you'll need to delete these, or uncomment the plugin, since multiple
+definitions of the same class are not allowed.
+
+Note that these should not be committed into git.
+
+Note that you can also `mvn compile -DskipTests=true` and see sources in
+`target/generated-sources/`.
+
+## Debugging HTTP requests
+
+To debug HTTP requests, place a `logging.properties` at
+`google-cloud-testing/storage-benchwrapper/logging.properties` with the
+following contents:
+
+```
+# Properties file which configures the operation of the JDK logging facility.
+# The system will look for this config file to be specified as a system property:
+# -Djava.util.logging.config.file=${project_loc:googleplus-simple-cmdline-sample}/logging.properties
+
+# Set up the console handler (uncomment "level" to show more fine-grained messages)
+handlers = java.util.logging.ConsoleHandler
+java.util.logging.ConsoleHandler.level = CONFIG
+
+# Set up logging of HTTP requests and responses (uncomment "level" to show)
+com.google.api.client.http.level = CONFIG
+```
--- a/google-cloud-testing/storage-benchwrapper/pom.xml
+++ b/google-cloud-testing/storage-benchwrapper/pom.xml
@@ -1,0 +1,116 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>benchwrapper</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <groupId>com.google.cloud</groupId>
+
+  <properties>
+    <grpcVersion>1.23.0</grpcVersion>
+    <protobufVersion>3.9.1</protobufVersion>
+  </properties>
+  
+  <build>
+    <extensions>
+      <extension>
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>1.5.0.Final</version>
+      </extension>
+    </extensions>
+    
+    <plugins>
+      <!-- Auto-generate classes from .protos -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.0</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.xolstice.maven.plugins</groupId>
+        <artifactId>protobuf-maven-plugin</artifactId>
+        <version>0.6.1</version>
+        <configuration>
+          <protocArtifact>com.google.protobuf:protoc:${protobufVersion}:exe:${os.detected.classifier}</protocArtifact>
+          <pluginId>grpc-java</pluginId>
+          <pluginArtifact>io.grpc:protoc-gen-grpc-java:${grpcVersion}:exe:${os.detected.classifier}</pluginArtifact>
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>compile</goal>
+              <goal>compile-custom</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+
+      <!-- Protobuf runtime -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>1.6.0</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <mainClass>com.google.cloud.benchwrapper.Main</mainClass>
+          <executable>maven</executable>
+          <skip>false</skip>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <!-- storage -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
+      <version>1.85.0</version>
+    </dependency>
+
+    <!-- protobuf  -->
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>${protobufVersion}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-protobuf</artifactId>
+      <version>${grpcVersion}</version>
+    </dependency>
+
+    <!-- grpc -->
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
+      <version>${grpcVersion}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-stub</artifactId>
+      <version>${grpcVersion}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty</artifactId>
+      <version>${grpcVersion}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative-boringssl-static</artifactId>
+      <version>2.0.25.Final</version>
+    </dependency>
+  </dependencies>
+</project>
--- a/google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
+++ b/google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.benchwrapper;
+
+import java.util.Properties;
+import io.grpc.Server;
+import io.grpc.netty.NettyServerBuilder;
+
+class Main {
+  public static void main(String[] args) throws Exception {
+    Properties properties = new Properties(System.getProperties());
+    String port = properties.getProperty("port");
+    if (port == null || port.equals("")) {
+      System.err.println("Usage: mvn clean install exec:java -DskipTests=true -Dport=8081");
+      System.exit(1);
+    }
+
+    String storageEmulatorHost = System.getenv("STORAGE_EMULATOR_HOST");
+    if (storageEmulatorHost == null || storageEmulatorHost.equals("")) {
+      // We could use system properties here too, but every other language uses
+      // an environment variable called STORAGE_EMULATOR_HOST, so the
+      // consistency is nice to maintain.
+      System.err.println("Please set STORAGE_EMULATOR_HOST=localhost:8080");
+      System.exit(1);
+    }
+
+    System.out.println("Server starting up...");
+
+    int portInt = Integer.parseInt(port);
+    final Server server = NettyServerBuilder
+      .forPort(portInt)
+      .addService(new StorageBenchWrapperImpl(storageEmulatorHost))
+      .build()
+      .start();
+
+    System.out.println("Server starting up... done. Listening on " + port);
+    Runtime.getRuntime().addShutdownHook(new Thread() {
+      @Override
+      public void run() {
+        // Use stderr here since the logger may have been reset by its JVM shutdown hook.
+        System.err.println("Shutting down gRPC server since JVM is shutting down");
+        server.shutdown();
+      }
+    });
+
+    server.awaitTermination();
+  }
+}
+
--- a/google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
+++ b/google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.benchwrapper;
+
+import io.grpc.stub.StreamObserver;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+import com.google.cloud.benchwrapper.StorageBenchWrapperGrpc.StorageBenchWrapperImplBase;
+import com.google.cloud.ReadChannel;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobId;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
+import com.google.cloud.storage.StorageOptions;
+
+class StorageBenchWrapperImpl extends StorageBenchWrapperImplBase {
+  private Storage client;
+
+  public StorageBenchWrapperImpl(String storageEmulatorHost) {
+    client = StorageOptions.newBuilder()
+        .setHost("http://" + storageEmulatorHost)
+        .build()
+        .getService();
+  }
+
+  public void write(ObjectWrite request, StreamObserver<EmptyResponse> responseObserver) {
+    System.out.println("write has been called");
+    EmptyResponse reply = EmptyResponse.newBuilder().build();
+    responseObserver.onNext(reply);
+    responseObserver.onCompleted();
+  }
+
+  public void read(ObjectRead request, StreamObserver<EmptyResponse> responseObserver) {
+    System.out.println("read has been called");
+
+    Blob blob = client.get(BlobId.of(request.getBucketName(), request.getObjectName()));
+
+    try (ReadChannel reader = blob.reader()) {
+      ByteBuffer bytes = ByteBuffer.allocate(64 * 1024);
+      while (reader.read(bytes) > 0) {
+        bytes.flip();
+        // do nothing with bytes
+        bytes.clear();
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+      System.exit(1);
+    }
+
+    EmptyResponse reply = EmptyResponse.newBuilder().build();
+    responseObserver.onNext(reply);
+    responseObserver.onCompleted();
+  }
+}
--- a/google-cloud-testing/storage-benchwrapper/src/main/proto/storage.proto
+++ b/google-cloud-testing/storage-benchwrapper/src/main/proto/storage.proto
@@ -1,0 +1,45 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package storage_bench;
+option java_multiple_files = true;
+option java_package = "com.google.cloud.benchwrapper";
+
+message ObjectRead{
+  // The bucket string identifier.
+  string bucketName = 1;
+  // The object/blob string identifier.
+  string objectName = 2;
+}
+
+message ObjectWrite{
+   // The bucket string identifier.
+  string bucketName = 1;
+  // The object/blob string identifiers.
+  string objectName = 2;
+  // The string containing the upload file path.
+  string destination = 3;
+}
+
+message EmptyResponse{
+}
+
+service StorageBenchWrapper{
+  // Performs an upload from a specific object.
+  rpc Write(ObjectWrite) returns (EmptyResponse) {}
+  // Read a specific object.
+  rpc Read(ObjectRead) returns (EmptyResponse){}
+}


chingor13:
I don't think this is the right place for this code - it will be shipped with the client library and adds additional dependencies for the consumers of google-cloud-java. It probably belongs in its own artifact (probably under google-cloud-testing/) which can depend upon the snapshot version of google-cloud-storage. In a test environment, we install all artifacts to the local maven repository and so maven will find the snapshot version before trying to fetch from Maven Central.

jadekler:
> I don't think this is the right place for this code - it will be shipped with the client library and adds additional dependencies for the consumers of google-cloud-java. It probably belongs in its own artifact (probably under google-cloud-testing/)

Done

> In a test environment, we install all artifacts to the local maven repository and so maven will find the snapshot version before trying to fetch from Maven Central.

What are the commands that need to be run to do that?

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,119 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>storage-benchwrapper</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <groupId>com.benchwrapper</groupId>
```
kolea2: I don't think you need a `groupId` here, I'd add a `<parent>` reference and make the artifactId 'google-cloud-benchmarking' or something similar. Take a look at https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-testing/google-cloud-conformance-tests/pom.xml#L6-L14 for an example

chingor13: The `groupId` here should match the Java package and be a domain that we can publish to. We should probably make it `com.google.cloud`.

The java package name should probably live under `com.google.cloud.benchwrapper` unless this is supposed to be a general tool outside of the scope of the cloud libraries.

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.xolstice.maven.plugins</groupId>
+        <artifactId>protobuf-maven-plugin</artifactId>
+        <version>0.6.1</version>
+        <configuration>
+          <protocArtifact>com.google.protobuf:protoc:3.7.1:exe:${os.detected.classifier}</protocArtifact>
```
chingor13: nit: can you make the protobuf version here and the grpc version into properties for use here and in the dependencies section?

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,63 @@
+# storage-benchwrapper
+
+storage-benchwraper is a gRPC wrapper around the storage library for benchmarking purposes.
```
chingor13: nit:
```suggestion
storage-benchwrapper is a gRPC wrapper around the storage library for benchmarking purposes.
```

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+mvn clean install -DskipTests=true (one time)
+cd google-cloud-testing/storage-benchwrapper
+mvn clean install exec:java -DskipTests=true
+```
+
+## Generating .proto sources
+
+Sources are generated automatically as part of the protobuf-maven-plugin plugin,
+but if you'd like to generate them yourself to see the output you can run:
```
chingor13: Another option is if you run `mvn compile` should should be able to see the generated sources in `target/generated-sources/`

google-cloud-testing/storage-benchwrapper/src/main/java/com/benchwrapper/StorageBenchWrapperImpl.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
+import com.google.cloud.storage.StorageOptions;
+
+class StorageBenchWrapperImpl extends StorageBenchWrapperImplBase {
+  private Storage client;
+
+  public StorageBenchWrapperImpl() {
+    client = StorageOptions.newBuilder()
+        .setHost("http://localhost:8080")
```
chingor13: What's running locally on port 8080? A proxy to GCS?

jadekler: Yes. I'd actually like to change the port to a flag: in other languages we have `--port` defined. To do so, what's the best approach? Something like http://commons.apache.org/proper/commons-cli/?

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Done

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Done

google-cloud-testing/storage-benchwrapper/README.md
jadekler: Done

google-cloud-testing/storage-benchwrapper/README.md
jadekler: Done

jadekler:
Thanks for the review. All comments addressed.

google-cloud-testing/storage-benchwrapper/src/main/java/com/benchwrapper/StorageBenchWrapperImpl.java
chingor13: System properties or environment variables would probably make sense. With System properties, you would add `-Dpropname=propvalue` to your Maven command or configure through your IDE.

google-cloud-testing/storage-benchwrapper/src/main/proto/storage.proto
```diff
@@ -0,0 +1,45 @@
+// Copyright 2016 Google LLC
```
kolea2: sorry, missed this earlier - year should be 2019

jadekler: Done

google-cloud-testing/storage-benchwrapper/src/main/java/com/benchwrapper/StorageBenchWrapperImpl.java
jadekler: Done

jadekler:
Thanks for the third review. All comments addressed.

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>benchwrapper</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <groupId>com.google.cloud</groupId>
+
+  <properties>
+    <grpcVersion>1.21.0</grpcVersion>
+    <protobufVersion>3.7.1</protobufVersion>
```
elharo: 3.9.1 is current and recommended.

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>benchwrapper</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <groupId>com.google.cloud</groupId>
+
+  <properties>
+    <grpcVersion>1.21.0</grpcVersion>
```
elharo: 1.23.0

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.xolstice.maven.plugins</groupId>
+        <artifactId>protobuf-maven-plugin</artifactId>
+        <version>0.6.1</version>
+        <configuration>
+          <protocArtifact>com.google.protobuf:protoc:3.7.1:exe:${os.detected.classifier}</protocArtifact>
```
elharo: use ${protobufVersion}

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+          <pluginId>grpc-java</pluginId>
+          <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.21.0:exe:${os.detected.classifier}</pluginArtifact>
```
elharo: use ${grpcversion}

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
```diff
@@ -0,0 +1,63 @@
...
+
+package com.google.cloud.benchwrapper;
+
+import java.util.Properties;
+import io.grpc.Server;
+import io.grpc.netty.NettyServerBuilder;
+
+class Main {
+  public static void main(String[] args) throws Exception {
+    Properties p = new Properties(System.getProperties());
```
elharo: p -->properties per google style rules

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.cloud.benchwrapper.StorageBenchWrapperGrpc.StorageBenchWrapperImplBase;
+import com.google.cloud.ReadChannel;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobId;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
+import com.google.cloud.storage.StorageOptions;
+
+class StorageBenchWrapperImpl extends StorageBenchWrapperImplBase {
+  private Storage client;
```
elharo: final?

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
```diff
@@ -0,0 +1,69 @@
...
+    System.out.println("write has been called");
+    EmptyResponse reply = EmptyResponse.newBuilder().build();
+    responseObserver.onNext(reply);
+    responseObserver.onCompleted();
+  }
+
+  public void read(ObjectRead request, StreamObserver<EmptyResponse> responseObserver) {
+    System.out.println("read has been called");
+
+    Blob b = client.get(BlobId.of(request.getBucketName(), request.getObjectName()));
```
elharo: b --> blob

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Done

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Done

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Done

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Done

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
jadekler: Could you point me at the relevant part of go/java-style? AFAICT no part of the variables section makes any mention of local variables (5.1, 5.2.*) needing to be words.

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
jadekler: Is there any impetus? AFAICT no point since this is application code, and a pretty small application.

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
jadekler: As above: Could you point me at the relevant part of go/java-style? AFAICT no part of the variables section makes any mention of local variables (5.1, 5.2.*) needing to be words.

jadekler:
Thanks for the fourth review. All comments addressed.

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
elharo: general good habit. It helps a lot with thread safety.

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
elharo: go/java-practices/naming

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+storage-benchwrapper is a gRPC wrapper around the storage library for benchmarking purposes.
+
+## Running
+
+```
+mvn clean install -DskipTests=true (one time)
+cd google-cloud-testing/storage-benchwrapper
+export STORAGE_EMULATOR_HOST=localhost:8080
+mvn clean install exec:java -DskipTests=true -Dport=8081
```
BenWhitehead: `exec:java` doesn't actually fork to a subprocess and as such doesn't properly handle send JVM level shutdown signals. If you are depending on JVM level shutdown signals (such as shutting down daemon thread pools) you'll want to switch to using `exec:exec`.

Here is a PR where I did just that if you want to use it as a starting place: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1523

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Sources are generated automatically as part of the protobuf-maven-plugin plugin,
+but if you'd like to generate them yourself to see the output you can run:
+
+```
+cd google-cloud-testing/storage-benchwrapper
+protoc \
+    --plugin=protoc-gen-grpc-java=build/exe/java_plugin/protoc-gen-grpc-java \
+    --java_out=src/main/java \
+    --grpc-java_out=src/main/java \
```
BenWhitehead: If you change this to `target/generated-sources/protobuf/java/` it will use the same location the maven plugin does and will remove the possible confusion of there being duplicates that would need to be manually cleaned up.

Alternatively, if you want to provide the command that will run just the protobuf generation in maven you could use: `mvn generate-sources`

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>benchwrapper</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <groupId>com.google.cloud</groupId>
+
```
BenWhitehead: Without this module being added to https://github.com/googleapis/google-cloud-java/blob/82ab4cb829b1121052689b7f4de738e54e1ad68d/google-cloud-testing/pom.xml#L21-30 it won't actually be loaded and ran as part of the build. Is that intentional?

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
+import com.google.cloud.storage.StorageOptions;
+
+class StorageBenchWrapperImpl extends StorageBenchWrapperImplBase {
+  private Storage client;
+
+  public StorageBenchWrapperImpl(String storageEmulatorHost) {
+    client = StorageOptions.newBuilder()
+        .setHost("http://" + storageEmulatorHost)
```
BenWhitehead: From the readme I'd expect this to need to be `https` is this correct with `http`?

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
```diff
@@ -0,0 +1,69 @@
...
+    try (ReadChannel reader = b.reader()) {
+      ByteBuffer bytes = ByteBuffer.allocate(64 * 1024);
+      while (reader.read(bytes) > 0) {
+        bytes.flip();
+        // do nothing with bytes
+        bytes.clear();
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+      System.exit(1);
```
BenWhitehead: Possibly use a different exit code than `1` here, since `1` above appears to be used for argument validation.

google-cloud-testing/storage-benchwrapper/README.md
elharo: Also, why clean and install? exec:exec or exec:java should do it as needed.

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+mvn clean install -DskipTests=true (one time)
+cd google-cloud-testing/storage-benchwrapper
+export STORAGE_EMULATOR_HOST=localhost:8080
+mvn clean install exec:java -DskipTests=true -Dport=8081
+```
+
+## Generating .proto sources
+
+Sources are generated automatically as part of the protobuf-maven-plugin plugin,
```
elharo: Try to avoid passive voice.
"automatically" is unneeded since it's implied.

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,67 @@
...
+protoc \
+    --plugin=protoc-gen-grpc-java=build/exe/java_plugin/protoc-gen-grpc-java \
+    --java_out=src/main/java \
+    --grpc-java_out=src/main/java \
+    --proto_path=src/main/proto \
+    src/main/proto/*.proto
+```
+
+Note that you'll need to delete these, or uncomment the plugin, since multiple
+definitions of the same class is not allowed.
```
elharo: is not --> are not

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Note that these should not be committed into git.
```
elharo: You might be able to delete some or all of these "Note that"s.

google-cloud-testing/storage-benchwrapper/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Note that you can also `mvn compile -DskipTests=true` and see sources in
+`target/generated-sources/`.
+
+## Generating keys
+
+Note: SslContextBuilder.forServer appears to only accept pkcs8 keys, as opposed
```
elharo: can you make this a definite statement?

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>benchwrapper</artifactId>
+  <version>0.0.0</version>
```
elharo: At least 0.1.0 and perhaps 1.0.0

jadekler: 0.0.0 is intentional: this is never going to be released, so the version should look like something fake. If version weren't a required field, I would delete it.

google-cloud-testing/storage-benchwrapper/README.md
jadekler: Actually can be deleted. (woohoo!)

google-cloud-testing/storage-benchwrapper/README.md
jadekler: Ack

google-cloud-testing/storage-benchwrapper/README.md
jadekler: Done

google-cloud-testing/storage-benchwrapper/README.md
jadekler: Done

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
jadekler: 1 is the unix catchall for generic errors.

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
jadekler: All emulators run on http. Is there a part of the readme I could clarify?

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Yes.

google-cloud-testing/storage-benchwrapper/README.md
jadekler: I included this here because vscode doesn't seem to know how to interpret the target/ directory well enough to understand import paths of imported things. So, these commands are included if you want to wack away at some temp things that live in src/main (which vscode will happily understand).

google-cloud-testing/storage-benchwrapper/README.md
jadekler: > exec:java doesn't actually fork to a subprocess and as such doesn't properly handle send JVM level shutdown signals.

Thanks. I'm not, so I'll save myself `pom.xml` frustration and skip that suggestion. :)

> Also, why clean and install? exec:exec or exec:java should do it as needed.

Doesn't seem to: https://gist.github.com/jadekler/d88583bc1c221ef9f1adb44ee88d6d0d

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.0</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.xolstice.maven.plugins</groupId>
```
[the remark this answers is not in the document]
jadekler: I have no idea. This one is the one recommended by grpc-java: https://github.com/grpc/grpc-java#generated-code

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
jadekler: Thanks. Done

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
jadekler: Ack. I prefer not - I strongly believe in writing code that serves a purpose. Let me know if you have strong feelings about this.

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
jadekler: Thanks. Done

jadekler:
Thanks for the fifth review. All comments addressed.

google-cloud-testing/storage-benchwrapper/README.md
BenWhitehead: Ah, fair enough.

Do we want to add a `.gitignore` to this module with the paths of the generated files under `src/main/java` set to ignore so they aren't accidentally committed? If not, the resulting build failure might be sufficient. Up to you.

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/StorageBenchWrapperImpl.java
BenWhitehead: I think I was looking at an old revision of the readme that was talking about generating certs, but I see that's no longer in the readme.

google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+      <artifactId>grpc-netty</artifactId>
+      <version>${grpcVersion}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative-boringssl-static</artifactId>
+      <version>2.0.25.Final</version>
+    </dependency>
+
+    <!-- ??? -->
```
elharo: replace or remove the ???

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.Properties;
+import io.grpc.Server;
+import io.grpc.netty.NettyServerBuilder;
+
+class Main {
+  public static void main(String[] args) throws Exception {
+    Properties properties = new Properties(System.getProperties());
+    String port = properties.getProperty("port");
+    if (port == null || port == "") {
```
elharo: "".equals(port)

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
```diff
@@ -0,0 +1,63 @@
...
+      System.err.println("Usage: mvn clean install exec:java -DskipTests=true -Dport=8081");
+      System.exit(1);
+    }
+
+    String storageEmulatorHost = System.getenv("STORAGE_EMULATOR_HOST");
+    if (storageEmulatorHost == null || storageEmulatorHost == "") {
```
elharo: "".equals(storageEmulatorHost)

jadekler: Done (inverted)

google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
jadekler: Done (inverted)

google-cloud-testing/storage-benchwrapper/pom.xml
jadekler: Done

jadekler:
Thanks for the sixth review. All comments addressed.

google-cloud-testing/storage-benchwrapper/README.md
jadekler: Build failure is OK for me. (less magic) Thank you!

PR #6067, opened 2019-08-12
connecting to Bigtable emulator using environment variable

### LINKED ISSUE #6057: Implement BIGTABLE_EMULATOR_HOST environment variable support as in golang
The documentation for the [Bigtable emulator](https://cloud.google.com/bigtable/docs/emulator) states:
```
In the environment for your application, set the BIGTABLE_EMULATOR_HOST
environment variable to the host and port where the Cloud Bigtable emulator
is running (for example, myhost.example.com:8010).

If you are running the emulator on the same machine as your application,
you can use the following command to set this environment variable automatically:

  $(gcloud beta emulators bigtable env-init)

Setting this environment variable causes the application to run in the emulator,
with no further action required.  The client automatically uses the variable
if it is set, instead of connecting to the Cloud Bigtable service.
```
This is not yet implemented for this new Java client.

Take a look at [this stackoverflow golang example](https://stackoverflow.com/questions/50897883/how-to-connect-to-bigtable-emulator-from-a-golang-application-how-to-use-it) that shows how the client is used no matter what is requested for project or instance.

It would be a great addition if this Java client would implement this behavior.

Note also that the [newBuilderForEmulator](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java#L70) code uses an integer port argument meaning that a client that wants to use the emulator needs to crack that integer from the `BIGTABLE_EMULATOR_HOST` environment variable.

rahulKQL:
Fixes #6057

User can now connect to Bigtable emulator in three ways:
 - By settings hostName:portNum in `BIGTABLE_EMULATOR_HOST` environment variable.
 - By providing only the port number within the local workstation.
 - By providing hostName & port Number.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
@@ -124,7 +124,7 @@ public final class BigtableTableAdminSettings {
    * number.
    */
   public static Builder newBuilderForEmulator(String hostname, int port) {
-    Builder builder = new Builder().setProjectId("fake-project").setInstanceId("fake-instance");
+    Builder builder = new Builder();
 
     builder
         .stubSettings()
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
@@ -101,7 +101,7 @@ public final class BigtableDataSettings {
    * port number.
    */
   public static Builder newBuilderForEmulator(String hostname, int port) {
-    Builder builder = new Builder().setProjectId("fake-project").setInstanceId("fake-instance");
+    Builder builder = new Builder();
 
     builder
         .stubSettings()
--- a/google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/TestEnvRule.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/TestEnvRule.java
@@ -15,7 +15,8 @@
  */
 package com.google.cloud.bigtable.data.v2.it.env;
 
-import com.google.common.truth.TruthJUnit;
+import static com.google.common.truth.TruthJUnit.assume;
+
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.junit.rules.ExternalResource;
@@ -38,18 +39,19 @@ import org.junit.rules.ExternalResource;
  * <p>By default, {@code emulator} will be used
  */
 public class TestEnvRule extends ExternalResource {
+
   private static final Logger LOGGER = Logger.getLogger(TestEnvRule.class.getName());
   private static final String BIGTABLE_EMULATOR_HOST_ENV_VAR = "BIGTABLE_EMULATOR_HOST";
-
   private static final String ENV_PROPERTY = "bigtable.env";
 
   private TestEnv testEnv;
 
   @Override
   protected void before() throws Throwable {
-    TruthJUnit.assume()
+    assume()
         .withMessage(
-            "Integration tests can't run with the BIGTABLE_EMULATOR_HOST_ENV_VAR environment variable set. Please use the emulator-it maven profile instead")
+            "Integration tests can't run with the BIGTABLE_EMULATOR_HOST environment variable set"
+                + ". Please use the emulator-it maven profile instead")
         .that(System.getenv())
         .doesNotContainKey(BIGTABLE_EMULATOR_HOST_ENV_VAR);
     String env = System.getProperty(ENV_PROPERTY, "emulator");
--- a/google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/EmulatorEnv.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/EmulatorEnv.java
@@ -39,10 +39,16 @@ public class EmulatorEnv implements TestEnv {
 
     tableAdminClient =
         BigtableTableAdminClient.create(
-            BigtableTableAdminSettings.newBuilderForEmulator(emulator.getPort()).build());
+            BigtableTableAdminSettings.newBuilderForEmulator(emulator.getPort())
+                .setProjectId("fake-project")
+                .setInstanceId("fake-instance")
+                .build());
     dataClient =
         BigtableDataClient.create(
-            BigtableDataSettings.newBuilderForEmulator(emulator.getPort()).build());
+            BigtableDataSettings.newBuilderForEmulator(emulator.getPort())
+                .setProjectId("fake-project")
+                .setInstanceId("fake-instance")
+                .build());
 
     tableAdminClient.createTable(CreateTableRequest.of(TABLE_ID).addFamily(FAMILY_ID));
   }


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -66,16 +70,55 @@ public static Builder newBuilder() {
     return new Builder();
   }
 
-  /** Create a new builder preconfigured to connect to the Bigtable emulator. */
+  /**
+   * Create a new builder preconfigured to connect to Bigtable emulator when hostname and port
+   * number are provided through BIGTABLE_EMULATOR_HOST environment variable.
+   */
+  public static Builder newBuilderForEmulator() {
```
rahulKQL: @igorbernstein2 
Q: What are your thoughts on making user set VM args, to be able to use `Builder#newBuilderForEmulator()`?

Q: Should admin & data client can share utility class? because till this point, it's logically separate clients within one module.

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
```diff
@@ -89,14 +92,53 @@ public static Builder newBuilder() {
...
+            + "newBuilderForEmulator(hostname, port)");
+
+    String[] hostPort = emulatorHostAndPort.split(":");
+    Preconditions.checkArgument(
+        hostPort.length == 2,
+        "Malformed "
+            + BIGTABLE_EMULATOR_HOST_ENV_VAR
+            + " environment variable: "
+            + emulatorHostAndPort
+            + ". Expecting host:port.");
```
igorbernstein2: I think this will fail for ipv6 addresses. It would be better to do use last index of

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
igorbernstein2: Q1. We should keep it as an environment variable. The main reason we are adding this feature is to be consistent with the emulator docs.

Q2. Let's keep it separate for now and discuss what other things need to be merge

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
```diff
@@ -89,14 +92,53 @@ public static Builder newBuilder() {
     return new Builder();
   }
 
-  /** Create a new builder preconfigured to connect to the Bigtable emulator. */
+  /**
+   * Create a new builder preconfigured to connect to Bigtable emulator when hostname and port
+   * number are provided through BIGTABLE_EMULATOR_HOST environment variable.
+   */
+  public static Builder newBuilderForEmulator() {
+    String emulatorHostAndPort = System.getProperty(BIGTABLE_EMULATOR_HOST_ENV_VAR);
```
igorbernstein2: This should be an environment variable not a system property

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
```diff
@@ -89,14 +92,53 @@ public static Builder newBuilder() {
     return new Builder();
   }
 
-  /** Create a new builder preconfigured to connect to the Bigtable emulator. */
+  /**
+   * Create a new builder preconfigured to connect to Bigtable emulator when hostname and port
+   * number are provided through BIGTABLE_EMULATOR_HOST environment variable.
+   */
+  public static Builder newBuilderForEmulator() {
```
igorbernstein2: To keep consistent with the emulator docs, I think this logic needs to be on by default so I think it needs to be implemented in newBuilder():

Something along the lines of:

```java
public static Builder newBuilder() {
  String emulatorEndpoint = System.getenv(BIGTABLE_EMULATOR_HOST);
  if (emulatorEndpoint != null) {
    new Builder().configureForEmulator(emulatorEndpoint);
  }
  return new Builder();
}
public static newBuilderForEmulator(int port) {
  return newBuilderForEmulator("localhost", port);
}
public static newBuilderForEmulator(String hostname, int port) {
  return new Builder().configureForEmulator(hostname, port);
}

private Builder configureForEmulator(String endpoint) {
   // parse hostname:port
   return configureForEmulator(host, port);
}

private Builder configureForEmulator(host, port) {
   ...
}

```

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
rahulKQL: Thanks for the feedback, I have just updated this change. Please have a look.

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -55,27 +56,57 @@
...
+  /**
+   * Create a new builder preconfigured to connect to the Bigtable emulator with o host & port name.
+   */
+  public static Builder newBuilderForEmulator(String hostname, int port) {
+    Builder builder = new Builder();
 
     builder
         .stubSettings()
         .setProjectId("fake-project")
         .setInstanceId("fake-instance")
```
igorbernstein2: This diverges from the other clients. I don't think we should set project & instance ids any more

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -55,27 +56,57 @@
...
 
-  /** Create a new builder preconfigured to connect to the Bigtable emulator. */
+  /** Create a new builder preconfigured to connect to the Bigtable emulator with port number. */
   public static Builder newBuilderForEmulator(int port) {
-    Builder builder = newBuilder();
+    return newBuilderForEmulator("localhost", port);
+  }
+
+  /**
+   * Create a new builder preconfigured to connect to the Bigtable emulator with o host & port name.
```
igorbernstein2: `with __a__ hostname and port.`

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
```diff
@@ -84,19 +89,45 @@ public Builder toBuilder() {
...
   public static Builder newBuilderForEmulator(int port) {
-    Builder builder = newBuilder().setProjectId("fake-project").setInstanceId("fake-instance");
+    return newBuilderForEmulator("localhost", port);
+  }
+
+  /**
+   * Create a new builder preconfigured to connect to the Bigtable emulator with host & port number.
+   */
+  public static Builder newBuilderForEmulator(String hostname, int port) {
+    Builder builder = new Builder().setProjectId("fake-project").setInstanceId("fake-instance");
```
igorbernstein2: I don't think we should be setting the project instance ids to stay consistent with other languages

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -55,27 +56,57 @@
...
-  /** Create a new builder. */
+  /**
+   * Create a new builder.
+   *
+   * <p>If emulator configuration provided in BIGTABLE_EMULATOR_HOST environment variable then it
+   * creates a builder preconfigured to connect to Bigtable using emulator hostname and port number.
+   */
   public static Builder newBuilder() {
+    String hostAndPort = System.getenv(BIGTABLE_EMULATOR_HOST_ENV_VAR);
+    if (hostAndPort != null && !hostAndPort.isEmpty()) {
```
igorbernstein2: `Strings.isNullOrEmpty(hostAndPort)`

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -55,27 +56,57 @@
...
+      try {
+        int lastIndexOfCol = hostAndPort.lastIndexOf(":");
+        int port = Integer.parseInt(hostAndPort.substring(lastIndexOfCol + 1));
+        return newBuilderForEmulator(hostAndPort.substring(0, lastIndexOfCol), port);
+      } catch (NumberFormatException ex) {
```
igorbernstein2: Please handle index out of bounds in case the port is missing

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/TestEnvRule.java
```diff
@@ -17,6 +17,7 @@
 
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.junit.Assume;
```
igorbernstein2: Please use Truth:

```
import static com.google.common.truth.TruthJUnit.assume;

TruthJUnit.assume()
        .withMessage("Integration tests can't run with the BIGTABLE_EMULATOR_HOST_ENV_VAR environment variable set. Please use the emulator-it maven profile instead")
        .that(System.getenv())
        .doesNotContainKey(BIGTABLE_EMULATOR_HOST_ENV_VAR);
```

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/TestEnvRule.java
```diff
@@ -38,13 +39,15 @@
...
   private static final Logger LOGGER = Logger.getLogger(TestEnvRule.class.getName());
+  private static final String BIGTABLE_EMULATOR_HOST_ENV_VAR = "BIGTABLE_EMULATOR_HOST";
 
   private static final String ENV_PROPERTY = "bigtable.env";
 
   private TestEnv testEnv;
 
   @Override
   protected void before() throws Throwable {
+    Assume.assumeTrue(System.getenv(BIGTABLE_EMULATOR_HOST_ENV_VAR) == null);
```
igorbernstein2: Please use Truth (see above)

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
rahulKQL: For Java, we have projectId & instanceId required. Shall I remove the `Preconditions` checks for projectId & InstanceId?
 I would vote to keep these checks and continue with "fake-project" & "fake-instance".

igorbernstein2: I'm not sure what you mean by that. I'm picturing something along the lines of:

```
BigtableTableAdminSettings.newBuilderForEmulator("localhost", 1234).setProjectId("blah").setInstanceId("blah2").build()
```

I think the other client languages require something similar

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/TestEnvRule.java
```diff
@@ -47,7 +47,11 @@
 
   @Override
   protected void before() throws Throwable {
-    Assume.assumeTrue(System.getenv(BIGTABLE_EMULATOR_HOST_ENV_VAR) == null);
+    TruthJUnit.assume()
```
igorbernstein2: Please use a static import

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/env/TestEnvRule.java
```diff
@@ -47,7 +47,11 @@
...
+        .withMessage(
+            "Integration tests can't run with the BIGTABLE_EMULATOR_HOST_ENV_VAR environment variable set. Please use the emulator-it maven profile instead")
```
igorbernstein2: Please use better line wrapping

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminSettings.java
rahulKQL: Thanks for the explanation, I mistook it as emulator without any projectId & instanceId. Have updated with the suggestion. please have a look.

PR #6074, opened 2019-08-14
Logging: Fix resourceNames fields when fetching log entries.

### LINKED ISSUE #5753: Logging: Cannot specify resources when fetching log entries
In LoggingImpl, the ListLogEntriesRequest is built and only specifies the projectId as the resource names: https://github.com/googleapis/google-cloud-java/blob/64189021c41c0f96b8a96beefb737cbf76be22c8/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingImpl.java#L632-L653

The end-user should be able to somehow set the resource names fields.

[RPC Specification](https://cloud.google.com/logging/docs/reference/v2/rpc/google.logging.v2#listlogentriesrequest)

Note: this will require some substantial public interface changes.

pmakani:
Fixes #5753


pmakani:
@chingor13 PTAL.

PR #6087, opened 2019-08-15
auto-value 1.6.5

sullis:


chingor13:
Thanks for submitting this. To update auto-value, we'll also need to update some other dependencies. Closing in favor of #6115

Issue #6075, opened 2019-08-14
Spanner JDBC: Fix flaky test


kolea2:
On java 11 build:

`[ERROR]   SingleUseTransactionTest.testExecuteUpdateWithTimeout:726 » IllegalArgument`

kolea2:
Also:

```
[ERROR] testTimeoutExceptionReadWriteTransactionalSlowRollback(com.google.cloud.spanner.jdbc.StatementTimeoutTest)  Time elapsed: 0.13 s  <<< ERROR!
com.google.cloud.spanner.SpannerException: DEADLINE_EXCEEDED: Statement execution timeout occurred for SELECT fast_column FROM fast_table
	at com.google.cloud.spanner.jdbc.StatementTimeoutTest.testTimeoutExceptionReadWriteTransactionalSlowRollback(StatementTimeoutTest.java:565)
Caused by: java.util.concurrent.TimeoutException
	at com.google.cloud.spanner.jdbc.StatementTimeoutTest.testTimeoutExceptionReadWriteTransactionalSlowRollback(StatementTimeoutTest.java:565)
```

olavloite:
```
[ERROR] Tests run: 20, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.05 s <<< FAILURE! - in com.google.cloud.spanner.jdbc.SingleUseTransactionTest
[ERROR] testExecuteUpdateWithTimeout(com.google.cloud.spanner.jdbc.SingleUseTransactionTest)  Time elapsed: 0.014 s  <<< ERROR!
java.lang.IllegalArgumentException
	at com.google.cloud.spanner.jdbc.SingleUseTransactionTest.testExecuteUpdateWithTimeout(SingleUseTransactionTest.java:726)
```

olavloite:
Also:

```
[ERROR] Tests run: 20, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.122 s <<< FAILURE! - in com.google.cloud.spanner.jdbc.SingleUseTransactionTest
[ERROR] testExecuteUpdateWithTimeout(com.google.cloud.spanner.jdbc.SingleUseTransactionTest)  Time elapsed: 0.016 s  <<< ERROR!
java.lang.IllegalArgumentException
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:128)
[...]

PR #6092, opened 2019-08-16
spanner-jdbc: Allow empty or invalid env credentials

olavloite:
On environments with no or invalid application credentials, the JDBC driver would throw an `IllegalArgumentException` when a connection was being created without an explicit credentials URL included in the connection string.

Updates #6075

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionOptions.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionOptions.java
@@ -225,6 +225,7 @@ class ConnectionOptions {
     private Credentials credentials;
     private List<StatementExecutionInterceptor> statementExecutionInterceptors =
         Collections.emptyList();
+    private CredentialsService credentialsService = CredentialsService.INSTANCE;
 
     private Builder() {}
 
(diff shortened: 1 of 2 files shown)


olavloite:
The build failure on Java 7 is related to the second issue described in #6075 and that is solved by #6094.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/ConnectionOptions.java
```diff
@@ -370,10 +370,23 @@ private ConnectionOptions(Builder builder) {
...
-            ? getCredentialsService().createCredentials(this.credentialsUrl)
-            : builder.credentials;
+    Credentials credentials = null;
+    if (builder.credentials == null) {
+      try {
+        credentials = getCredentialsService().createCredentials(this.credentialsUrl);
+      } catch (Exception e) {
+        // Ignore the error if the user did not set a credentials URL.
+        // The error is then caused by an invalid environment configuration
+        // and should not be thrown here.
```
kolea2: can we add a test that asserts the correct exception is thrown?

olavloite: I've added a couple of test cases to cover this.

olavloite:
Abandoned. This should be fixed in the test cases rather than deferring the exception to a later moment.

PR #6109, opened 2019-08-19
Pub/Sub: subscriber async pull example update

anguillanneuf:
The old example exits immediately without receiving or printing messages, giving some users the impression that the sample doesn't work.
The updated one keeps the subscriber alive to listen to messages.


anguillanneuf:
Thanks Alex!

PR #4022, opened 2018-11-20
Add an enhanced layer for BigQuery Storage client.

kmjung:
This pull request creates an enhanced shim layer in the BigQuery Storage API client and adds additional settings and configuration to the shim layer.

PR #6112, opened 2019-08-20
Fixes an issue with splitReadStream

aryann:
The method requires a fraction, which is not provided in this client.

Additionally, I'm removing some comments that don't make sense.

### DIFF
--- a/google-cloud-clients/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta1/BigQueryStorageClient.java
+++ b/google-cloud-clients/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta1/BigQueryStorageClient.java
@@ -491,17 +491,21 @@ public class BigQueryStorageClient implements BackgroundResource {
    *
    * <pre><code>
    * try (BigQueryStorageClient bigQueryStorageClient = BigQueryStorageClient.create()) {
-   *   Stream originalStream = Stream.newBuilder().build();
+   *   Stream originalStream = getStream();
    *   SplitReadStreamResponse response = bigQueryStorageClient.splitReadStream(originalStream);
    * }
    * </code></pre>
    *
    * @param originalStream Stream to split.
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   * @deprecated Use {@link BigQueryStorageClient#splitReadStream(SplitReadStreamRequest)}.
    */
   public final SplitReadStreamResponse splitReadStream(Stream originalStream) {
     SplitReadStreamRequest request =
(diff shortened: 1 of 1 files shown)


chingor13:
Thanks for submitting this. The comments mean that this entire file is managed by a code generator, so any hand-written updates will be overwritten. We will need to address this upstream in our generator and/or service definitions.

Can you open an issue here to reference and we will look into how to best make the changes upstream?

aryann:
I was told my @mmladenovski that this file was written by hand, which is why I removed the comments.

How is this file generated?

mmladenovski:
> I was told my @mmladenovski that this file was written by hand, which is why I removed the comments.
> 
> How is this file generated?

@aryann , the file was added with https://github.com/googleapis/google-cloud-java/pull/4022 . @kmjung , might provide more details. 

I think it was copied from the auto-generated BaseBigQueryStorageClient and then modified to use the Enhanced stub along with the ResumptionStrategy.

The comments related to auto-generation should be removed.

google-cloud-clients/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta1/BigQueryStorageClient.java
```diff
@@ -501,21 +491,24 @@ public final void finalizeStream(FinalizeStreamRequest request) {
...
    * @param originalStream Stream to split.
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   * @deprecated Use {@link BigQueryStorageClient#splitReadStream(SplitReadStreamRequest)}.
    */
   public final SplitReadStreamResponse splitReadStream(Stream originalStream) {
     SplitReadStreamRequest request =
-        SplitReadStreamRequest.newBuilder().setOriginalStream(originalStream).build();
+        SplitReadStreamRequest.newBuilder()
+            .setOriginalStream(originalStream)
+            .setFraction(0.5f)
```
ehabqadah: I would suggest to move the `0.5f`  value to a constant `STREAM_SPLIT_FRACTION` and use it in all places!

meredithslota:
The BigQuery Storage client for Java now lives here: https://github.com/googleapis/java-bigquerystorage and has recently been re-generated. The comments re: auto-generation are still present because this is an autogenerated library, but if you see other issues (the file referenced from v1beta1 is here: https://github.com/googleapis/java-bigquerystorage/blob/master/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta1/BigQueryStorageClient.java) please file an issue in that repo. I'm going to go ahead and close this. 

cc: @stephaniewang526 as an FYI and in case you want to do something different with this issue. :)

PR #6115, opened 2019-08-21
Update core dependencies

chingor13:
* update gax to 1.48.0
* update google-auth-library to 0.17.0
* update grpc to 1.23.0
* ensure guava versions at 28.0-android
* update protobuf to 3.9.1
* update opencensus to 0.23.0
   * in opencensus 0.22.0, the runningSpanStore defaulted to not storing spans, so we need to configure the store to actually keeps some spans for testing
* update auto-value to 1.6.6


chingor13:
I cannot figure out why this won't compile on Java 7. It's failing to find another package installed from this repo.

chingor13:
autovalue >= 1.5 must use Java 8 for compilation but will work on Java 7

PR #6148, opened 2019-08-23
Firestore:Fix Observable in calls instead of wrapping in ApiFuture

### LINKED ISSUE #5887: Firestore expose Observable in calls instead of wrapping in ApiFuture
Hi there, this is another issue related to the GCP Java Frameworks team's reactive Spring Data project. 

We noticed that the main document retrieval method `getAll()` (and the singular `get()` is implemented using `getAll()`) actually creates an observer in-line for the request ,but then wraps the entire return value in an ApiFuture. (FirestoreImpl.java line 154 or thereabouts)

The problem for us is that the ApiFuture requires a whole new thread from the user to actually use, while letting us access the observer (either by accepting a callback that is then called inside your in-line observer or otherwise) lets us utilize the thread GRPC manages, which would be really ideal for us. 

So our request is: would it be possible to expose an alternative method `getAll` and `get`  in the Firestore client that lets us tie into the observer without the extra wrapping layer of ApiFuture? 

cc my teammates: @meltsufin, @dzou, @elefeint, @dmitry-s

abhinav-qlogic:
Fixes #5887

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
@@ -19,7 +19,6 @@ package com.google.cloud.firestore;
 import com.google.api.core.ApiFuture;
 import com.google.api.gax.rpc.ApiStreamObserver;
 import com.google.cloud.Service;
-import com.google.protobuf.ByteString;
 import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -121,18 +120,16 @@ public interface Firestore extends Service<FirestoreOptions>, AutoCloseable {
 
   /**
    * Retrieves multiple documents from Firestore while optionally applying a field mask to reduce
-   * the amount of data transmitted. Returned documents will be out of order and missing documents
-   * will not be returned.
+   * the amount of data transmitted. Returned documents will be out of order.
    *
    * @param documentReferences Array with Document References to fetch.
    * @param fieldMask If not null, specifies the subset of fields to return.
-   * @param transactionId Id of {@link Transaction}
-   * @param responseObserver ApiStreamObserver of DocumentSnapshot
+   * @param responseObserver The observer to be notified when {@link DocumentSnapshot} details
+   *     arrive.
    */
   void getAll(
       @Nonnull DocumentReference[] documentReferences,
       @Nullable FieldMask fieldMask,
-      @Nullable ByteString transactionId,
       final ApiStreamObserver<DocumentSnapshot> responseObserver);
 
   /**
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -140,19 +140,26 @@ class FirestoreImpl implements Firestore {
   @Override
   public ApiFuture<List<DocumentSnapshot>> getAll(
       @Nonnull DocumentReference... documentReferences) {
-    return this.getAll(documentReferences, null, null);
+    return this.getAll(documentReferences, null, (ByteString) null);
   }
 
   @Nonnull
   @Override
   public ApiFuture<List<DocumentSnapshot>> getAll(
       @Nonnull DocumentReference[] documentReferences, @Nullable FieldMask fieldMask) {
-    return this.getAll(documentReferences, fieldMask, null);
+    return this.getAll(documentReferences, fieldMask, (ByteString) null);
   }
 
   @Nonnull
   @Override
   public void getAll(
+      final @Nonnull DocumentReference[] documentReferences,
+      @Nullable FieldMask fieldMask,
+      @Nonnull final ApiStreamObserver<DocumentSnapshot> apiStreamObserver) {
+    this.getAll(documentReferences, fieldMask, null, apiStreamObserver);
+  }
+
+  void getAll(
       final @Nonnull DocumentReference[] documentReferences,
       @Nullable FieldMask fieldMask,
       @Nullable ByteString transactionId,
--- a/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
+++ b/google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
@@ -1348,7 +1348,6 @@ public class ITSystemTest {
     firestore.getAll(
         documentReferences,
         FieldMask.of("foo"),
-        null,
         new ApiStreamObserver<DocumentSnapshot>() {
 
           @Override


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
```diff
@@ -354,6 +355,15 @@ public T apply(List<T> results) {
     return extractFirst(firestore.getAll(this));
   }
 
+  /** Fetches the document reference in the form of the stream inside apiStreamObserver */
+  @Nonnull
+  public void get(
```
schmidt-sebastian: I'm still lukewarm on adding additional API surface. I think we can live without DocumentReference.get() since you should be able to call getAll().

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -117,6 +118,19 @@
...
 
+  /**
+   * Retrieves multiple documents from Firestore, while optionally applying a field mask to reduce
+   * the amount of data transmitted.
+   *
+   * @param documentReferences Array with Document References to fetch.
+   * @param responseObserver ApiStreamObserver of DocumentSnapshot
+   * @param fieldMask If set, specifies the subset of fields to return.
+   */
+  void getAll(
```
schmidt-sebastian: Please match the signature above and change the order to:

```getAll(DocumentReference[] documentReferences, @Nullable FieldMask fieldMask,  final ApiStreamObserver<DocumentSnapshot> responseObserver)```

Putting the observer last also makes for much more readable code when a Lambda is used.

File: google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java [hunk 3]
```diff
@@ -117,6 +118,19 @@
   ApiFuture<List<DocumentSnapshot>> getAll(
       @Nonnull DocumentReference[] documentReferences, @Nullable FieldMask fieldMask);
 
+  /**
+   * Retrieves multiple documents from Firestore, while optionally applying a field mask to reduce
+   * the amount of data transmitted.
```
schmidt-sebastian: Please mention that the documents will be returned out of order and that missing documents will not be returned at all.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -248,6 +248,99 @@ public void onCompleted() {
     return futureList;
   }
 
+  void getAll(
```
schmidt-sebastian: Please merge this with the internal method above. You can likely rewrite the existing `getAll()` to use your new method.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,59 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+          @Override
+          public void onError(Throwable throwable) {
+            Assert.fail("Error occurred while streaming");
+          }
+
+          @Override
+          public void onCompleted() {}
+        },
+        documentReferences);
+    Thread.sleep(1000);
```
schmidt-sebastian: Please rewrite this tests to not sleep.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,59 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+          @Override
+          public void onError(Throwable throwable) {
+            Assert.fail("Error occurred while streaming");
+          }
+
+          @Override
+          public void onCompleted() {}
+        };
+    ref.get(FieldMask.of("foo"), responseObserver, ref);
+    Thread.sleep(1000);
```
schmidt-sebastian: Please rewrite this tests to not sleep.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,59 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+          public void onError(Throwable throwable) {
+            Assert.fail("Error occurred while streaming");
+          }
+
+          @Override
+          public void onCompleted() {}
+        },
+        documentReferences);
+    Thread.sleep(1000);
+    assertEquals(map("foo", "bar"), documentSnapshots.get(0).getData());
```
schmidt-sebastian: Please assert response size and fetch multiple documents in this test, including non-existing documents.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,59 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+
+          @Override
+          public void onError(Throwable throwable) {
+            Assert.fail("Error occurred while streaming");
+          }
+
+          @Override
+          public void onCompleted() {}
+        };
+    ref.get(FieldMask.of("foo"), responseObserver, ref);
```
schmidt-sebastian: I think this test will likely go away as per my previous suggestion of removing this API.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -117,6 +118,20 @@
   ApiFuture<List<DocumentSnapshot>> getAll(
       @Nonnull DocumentReference[] documentReferences, @Nullable FieldMask fieldMask);
 
+  /**
+   * Retrieves multiple documents from Firestore while optionally applying a field mask to reduce
+   * the amount of data transmitted. Returned documents will be out of order and missing documents
+   * will not be returned.
+   *
+   * @param documentReferences Array with Document References to fetch.
+   * @param fieldMask If set, specifies the subset of fields to return.
```
schmidt-sebastian: s/set/not null/

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -140,21 +140,31 @@ public DocumentReference document(@Nonnull String documentPath) {
   @Override
   public ApiFuture<List<DocumentSnapshot>> getAll(
       @Nonnull DocumentReference... documentReferences) {
-    return this.getAll(documentReferences, null, null);
+    return this.getAll(documentReferences, null, null, null);
```
schmidt-sebastian: For cases like this, it might make sense to add inline comments to describe the arguments:

```
   return this.getAll(documentReferences, /* foo= */ null, /* bar= */ null, /* baz= */ null);
```

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -140,21 +140,31 @@ public DocumentReference document(@Nonnull String documentPath) {
...
+    this.getAll(documentReferences, fieldMask, null, responseObserver);
   }
 
   /** Internal getAll() method that accepts an optional transaction id. */
   ApiFuture<List<DocumentSnapshot>> getAll(
       final DocumentReference[] documentReferences,
       @Nullable FieldMask fieldMask,
-      @Nullable ByteString transactionId) {
+      @Nullable ByteString transactionId,
+      @Nullable final ApiStreamObserver apiStreamObserver) {
```
schmidt-sebastian: Do you think we could have one method that adds the ApiStreamObserver, and one that turns it into an ApiFuture?

```
void getAll(
      final DocumentReference[] documentReferences,
      @Nullable FieldMask fieldMask,
      @Nullable ByteString transactionId,
      @Nullable final ApiStreamObserver apiStreamObserver) { ... }

ApiFuture<List<DocumentSnapshot>> getAll(
      final DocumentReference[] documentReferences,
      @Nullable FieldMask fieldMask,
      @Nullable ByteString transactionId) { 
   List results....
   getAll(..., new ApiStreamObserver() -> {
      result.add();
    });
    // Sort and return task
 }
```

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
       this.error = error;
     }
   }
+
+  @Test
+  public void getAllWithObserver() throws Exception {
+    DocumentReference ref = randomColl.document("doc1");
```
schmidt-sebastian: s/ref/ref1/

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+    ref.set(ALL_SUPPORTED_TYPES_MAP).get();
+
+    DocumentReference ref1 = randomColl.document("doc2");
```
schmidt-sebastian: s/ref1/ref2

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+    ref1.set(ALL_SUPPORTED_TYPES_MAP).get();
+
+    DocumentReference ref2 = randomColl.document("doc3");
```
schmidt-sebastian: s/ref2/ref3/

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+    ref2.set(ALL_SUPPORTED_TYPES_MAP).get();
+    ref2.delete();
```
schmidt-sebastian: You don't need to set and delete a document in order for it to show up as "missing". You can just query a non-existing document that you never created.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+
+    if (null != documentSnapshots && documentSnapshots.size() > 0) {
+      for (DocumentSnapshot documentSnapshot : documentSnapshots) {
+        if (null != documentSnapshot.getData()) {
+          assertEquals(map("foo", "bar"), documentSnapshot.getData());
+        } else {
+          assertFalse(documentSnapshot.exists());
+        }
+      }
+    }
```
schmidt-sebastian: I'd slightly prefer if you checked the document references. Do note that they are returned alphabetically sorted by the backend and you could just verify get them by index.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+        });
+
+    future.get();
+
+    if (null != documentSnapshots && documentSnapshots.size() > 0) {
+      for (DocumentSnapshot documentSnapshot : documentSnapshots) {
+        if (null != documentSnapshot.getData()) {
+          assertEquals(map("foo", "bar"), documentSnapshot.getData());
+        } else {
+          assertFalse(documentSnapshot.exists());
```
schmidt-sebastian: This check (getData() == null && ! docSnapshot.exists()) is always true and doesn't quite verify that one of your documents is not returned.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+            future.setException(throwable);
+          }
+
+          @Override
+          public void onCompleted() {
+            future.set(null);
+          }
+        });
+
+    future.get();
```
schmidt-sebastian: Thank you for updating this!

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+
+    if (null != documentSnapshots && documentSnapshots.size() > 0) {
+      for (DocumentSnapshot documentSnapshot : documentSnapshots) {
+        if (null != documentSnapshot.getData()) {
+          assertEquals(map("foo", "bar"), documentSnapshot.getData());
```
schmidt-sebastian: I'm surprised this work because ALL_SUPPORTED_TYPES_MAP has a lot more fields.

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+    ref.set(ALL_SUPPORTED_TYPES_MAP).get();
+
+    DocumentReference ref1 = randomColl.document("doc2");
+    ref1.set(ALL_SUPPORTED_TYPES_MAP).get();
+
+    DocumentReference ref2 = randomColl.document("doc3");
+    ref2.set(ALL_SUPPORTED_TYPES_MAP).get();
+    ref2.delete();
+
+    final List<DocumentSnapshot> documentSnapshots = new ArrayList<>();
```
BenWhitehead: This should be wrapped in a synchronized list since it is being written and read from different threads.

```suggestion
    final List<DocumentSnapshot> documentSnapshots = Collections.synchronizedList(new ArrayList<DocumentSnapshot>());
```

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1327,4 +1329,54 @@ public void floatIncrement() throws ExecutionException, InterruptedException {
...
+
+          @Override
+          public void onCompleted() {
+            future.set(null);
+          }
+        });
+
+    future.get();
+
+    if (null != documentSnapshots && documentSnapshots.size() > 0) {
```
BenWhitehead: This if statement doesn't actually do anything. `documentSnapshots` is never null due to the final initializer above, and the size check before iterating is not needed due to the for-each not executing if the list is empty.

abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -117,6 +119,22 @@
...
 
+  /**
+   * Retrieves multiple documents from Firestore while optionally applying a field mask to reduce
+   * the amount of data transmitted. Returned documents will be out of order and missing documents
+   * will not be returned.
+   *
+   * @param documentReferences Array with Document References to fetch.
+   * @param fieldMask If not null, specifies the subset of fields to return.
+   * @param transactionId Id of {@link Transaction}
+   * @param responseObserver ApiStreamObserver of DocumentSnapshot
```
schmidt-sebastian: Since this is the external Firestore API, do you mind coming up with a more descriptive way to describe `responseObserver`?

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -117,6 +119,22 @@
...
       @Nonnull DocumentReference[] documentReferences, @Nullable FieldMask fieldMask);
 
+  /**
+   * Retrieves multiple documents from Firestore while optionally applying a field mask to reduce
+   * the amount of data transmitted. Returned documents will be out of order and missing documents
+   * will not be returned.
+   *
+   * @param documentReferences Array with Document References to fetch.
+   * @param fieldMask If not null, specifies the subset of fields to return.
+   * @param transactionId Id of {@link Transaction}
```
schmidt-sebastian: Ideally, the transaction ID handling would not be exposed in the public API. We currently don't use the transaction ID in any of our customer facing APIs, but instead have helper methods that use them internally.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -117,6 +119,22 @@
   ApiFuture<List<DocumentSnapshot>> getAll(
       @Nonnull DocumentReference[] documentReferences, @Nullable FieldMask fieldMask);
 
+  /**
+   * Retrieves multiple documents from Firestore while optionally applying a field mask to reduce
+   * the amount of data transmitted. Returned documents will be out of order and missing documents
+   * will not be returned.
```
schmidt-sebastian: Hm, it looks like I was wrong and missing documents do get returned. Can you remove the last part of your API doc?

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
abhinav-qlogic: done

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
abhinav-qlogic: done

schmidt-sebastian:
@BenWhitehead This is good to be merged.

BenWhitehead:
Thanks @schmidt-sebastian, I'll hold off until the builds have completed (there appears to be some slowness with the build queue).

PR #6157, opened 2019-08-25
spanner-jdbc: Add tests for JdbcConnection

olavloite:
Adds additional tests for `JdbcConnection`. Also fixes a small issue in `JdbcConnection` that caused the method `isValid(int)` to throw an exception when called on a closed connection, instead of returning `false` as it should have.


kolea2:
LGTM just needs a rebase

PR #6161, opened 2019-08-26
Firestore: Make DocumentSnapshot.fromDocument(...) public

### LINKED ISSUE #5937: Firestore: Make DocumentSnapshot.fromDocument(...) public
Hi, this is one more issue related to the GCP Java Frameworks team's reactive Spring Data project.

In order to convert documents to java objects we would like to be able to create a `DocumentSnapshot` from a `Document` so we can then call `DocumentSnapshot.toObject(...)`

Note, that currently it requires FirestoreImpl which is package-private.

If possible, a static method that converts a `Document` to java object would be preferable.

athakor:
Fixes #5937


schmidt-sebastian:
This is a publicly usable from here: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Internal.java#L54

Let me know if there is anything else we need, but to me this looks like it should solve your use case.

PR #6170, opened 2019-08-27
spanner-jdbc: Don't call getExitingExecutorService in StatementExecutor

nevinheintze:
Remove call to getExitingExecutorService when we create a new
ThreadPoolExecutor in spanner-jdbc's StatementExecutor.

Calling getExitingExecutorService converts the ThreadPoolExecutor into an
ExecutorService that exits when the application is complete i.e. we keep
around the ThreadPoolExecutor and won't GC it. Given that we allocate a new
ThreadPoolExecutors for every StatementExecutor, this is effectively a
memory leak since the ThreadPoolExecutors is kept around after
StatementExecutor is GC'd and the ThreadPoolExecutor has been shutdown.  It
can lead to substantial memory usage if an app uses many
StatementExecutors.


olavloite:
Good spot. The method was called to ensure that the executor would use a daemon `ThreadFactory`, but that is not necessary as it is already passed a `ThreadFactory` that produces daemon threads. But the `MoreExecutors.getExitingExecutorService` would also add a shutdown hook which kept the executor in memory.

olavloite:
The failed build on Windows Java 8 is unrelated to this PR, and is fixed by https://github.com/googleapis/google-cloud-java/pull/6174.

olavloite:
@kolea2 Could you have a quick look as well?

PR #6175, opened 2019-08-27
spanner-jdbc: Fix possible IllegalArgumentException in SingleUseTransaction

olavloite:
The fireAndForgetRollbackAndCloseTxManager method could cause an IllegalArgumentException if it happened to execute after a different method had already closed the TransactionManager. The fireAndForgetRollback method has therefore been removed and replaced with a synchronous rollback call. This will make update statements in autocommit mode that return an error slightly slower than they were, as they will not return until the rollback method has been executed.

The fireAndForgetRollback method is something that should rather be added to the TransactionManager itself.

Fixes this flake: https://source.cloud.google.com/results/invocations/df3b7c34-a981-4dfa-9978-7fa34e290142/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fgoogle-cloud-java%2Fpresubmit%2Fjava7/log

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
@@ -378,6 +378,8 @@ class SingleUseTransaction extends AbstractBaseUnitOfWork {
         }
       } finally {
         if (txManager != null) {
+          // Calling txManager.close() will rollback the transaction if it is still active, i.e. if
+          // an error occurred before the commit() call returned successfully.
           txManager.close();
         }
       }


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
```diff
@@ -365,23 +363,23 @@ public Long call() throws Exception {
 
     @Override
     public T call() throws Exception {
-      txManager = dbClient.transactionManager();
-      // Check the interrupted state after each (possible) round-trip to the db to allow the
-      // statement to be cancelled.
-      checkInterrupted();
-      try (TransactionContext txContext =
-          txManager.getState()
-                  == com.google.cloud.spanner.TransactionManager.TransactionState.ABORTED
```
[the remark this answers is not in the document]
olavloite: The `TransactionManager` is always (re-)created at the beginning of this method and closed at the end of the method, so it can never be aborted at this point, so the whole check was redundant anyways.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
```diff
@@ -460,33 +448,6 @@ private void checkInterrupted() throws InterruptedException {
...
-        txManager.getState()
-            == com.google.cloud.spanner.TransactionManager.TransactionState.STARTED);
-    ExecutorService rollbackExecutor =
-        Executors.newSingleThreadExecutor(new ConnectionImpl.DaemonThreadFactory());
-    rollbackExecutor.submit(
-        new Callable<Void>() {
-          @Override
-          public Void call() throws Exception {
-            try {
-              txManager.rollback();
```
skuruppu: [nit] If you could update the PR description to state that now we rely on the backend to abort and rollback the transaction rather than explicitly triggering a rollback, that would be great.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
```diff
@@ -424,32 +422,22 @@ private TransactionalBatchUpdateCallable(Iterable<ParsedStatement> updates) {
...
+          Thread.sleep(e.getRetryDelayInMillis() / 1000);
+        } catch (InterruptedException e1) {
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.CANCELLED, "Statement execution was interrupted", e1);
         }
-      } catch (Exception e) {
-        if (txManager != null) {
-          if (txManager.getState()
-              == com.google.cloud.spanner.TransactionManager.TransactionState.STARTED) {
-            fireAndForgetRollbackAndCloseTxManager(txManager);
```
skuruppu: Can you comment on what the performance implication for the client is if the rollback is not triggered immediately? We may be locking the rows for longer than is necessary while we wait for the backend to abort the transaction which might prevent the user from doing a manual retry.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
olavloite: I've updated the description to better describe the entire change. The change does however not rely on the backend to rollback/abort the transaction. Instead, it executes a rollback statement synchronously. I.e. an update statement that is executed in autocommit mode and that returns an error, will now wait until the rollback call has finished before returning. Previously, the rollback call would be executed using a separate worker thread.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
olavloite: The rollback is triggered immediately, but I understand why that is not obvious from the code. The `TransactionManager.close()` method that is called on line 381 will execute a rollback if the transaction is still active. I've added a comment there to make that clearer.

google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/SingleUseTransaction.java
skuruppu: Thanks Knut for the explanation and for updating the description.

olavloite:
@skuruppu I had the same issue with a different PR yesterday and talked to @kolea2 about it. She confirmed that it is unrelated to this

kolea2:
@skuruppu @olavloite yep, that is a temporary break while we wait for the linkage monitor build to be updated. Fine to merge.

Issue #6169, opened 2019-08-26
Spanner: BatchCreateSessions


kolea2:

olavloite:
The Spanner client library should use the new BatchCreateSessions API to create and maintain the session pool. In addition to using the new API, the following changes will be applied to the session pool to improve performance:
* Change the order of the session pool from FIFO to LIFO. LIFO has been determined to show better and more stable query performance.
* Use BatchCreateSessions to create and maintain the session pool.
* Change the default session pool options to MinSessions > 0.

PR #6178, opened 2019-08-27
Spanner: Session pool should return sessions in LIFO order

olavloite:
Makes the session pool return sessions in LIFO order, as this decreases the chance that a client is handed a session that has had its server side caches and connections closed. By returning sessions in LIFO order, the P99 latency for queries, especially in typical 1QPS load scenarios, is reduced.

Updates #6169

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
@@ -157,12 +157,12 @@ public class SessionPoolTest extends BaseSessionPoolTest {
     session3.close();
     session4.close();
 
-    Session rw1 = pool.getReadWriteSession();
-    Session rw2 = pool.getReadWriteSession();
-    assertThat(rw1).isEqualTo(session4);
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolTest.java
```diff
@@ -140,6 +140,31 @@ public void sessionCreation() {
...
+    session1.close();
+    Session session3 = pool.getReadSession();
+    Session session4 = pool.getReadSession();
+    assertThat(session3).isEqualTo(session1);
+    assertThat(session4).isEqualTo(session2);
+    session3.close();
+    session4.close();
+
+    Session rw1 = pool.getReadWriteSession();
+    Session rw2 = pool.getReadWriteSession();
```
skuruppu: [nit] Naming wise, I would call these session5 and session6 to be consistent with the variable naming elsewhere in this test and file, and since the type is not relevant.

olavloite: Renamed to session5 and session6.

PR #6185, opened 2019-08-28
Spanner: Add BatchCreateSessions rpc and use rotating channel number

olavloite:
Adds the BatchCreateSessions rpc to the Spanner client and uses this for session creation. A new `SessionClient` is added that encapsulates the details of how the session creation is evenly distributed over all available channels and that multiple calls might be necessary before all sessions have been created.
The sessions are returned by the `SessionClient` as an `Enumeration` that will yield results as soon as sessions are available. This way, the session pool does not have to wait until all the requested sessions have been created, but can start to serve sessions as soon as some have come available.

Updates #6169


olavloite:
@snehashah16 The previous version of this PR only added the rpc definition for BatchCreateSessions (my intention was to break the change into somewhat smaller pieces), but I can see how that introduces more questions than convenience. I have now updated PR to include the changes to the session pool as well.

I also refactored the code quite extensively based on your suggestions. The details for spreading the calls over all channels is now encapsulated in the `SessionClient` class, along with making additional calls if the server does not return the number of sessions that was requested.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1568,53 +1600,80 @@ public void run() {
...
-              }
-              if (closeSession) {
-                closeSession(pooledSession);
-              }
-            }
-          });
+      numSessionsBeingCreated += sessionCount;
+      try {
+        // Create a batch of sessions. The actual session creation can be split into
+        // multiple gRPC calls and the sessions are returned by the enumerator as they come
```
skuruppu: "the sessions are returned by the enumerator" should be "the session consumer consumes the returned sessions as they become available".

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1568,53 +1600,80 @@ public void run() {
...
+        pooledSession = new PooledSession(session);
+        numSessionsBeingCreated--;
+        if (closureFuture != null) {
+          closeSession = true;
+        } else {
+          Preconditions.checkState(totalSessions() <= options.getMaxSessions() - 1);
+          allSessions.add(pooledSession);
+          // Release the session to a random position in the pool to prevent that a
+          // batch of sessions that are affiliated with the same channel are all placed
+          // sequentially in the pool.
```
skuruppu: "to prevent the case that a batch of sessions that are affiliated ..."

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -0,0 +1,287 @@
...
+    public void run() {
+      List<SessionImpl> sessions = null;
+      int remainingSessionsToCreate = sessionCount;
+      while (remainingSessionsToCreate > 0) {
+        try {
+          sessions = internalBatchCreateSessions(remainingSessionsToCreate, channelHint);
+        } catch (Throwable t) {
+          consumer.onSessionCreateFailure(t, remainingSessionsToCreate);
+          break;
+        }
```
snehashah16: would we not retry  ? 
If retries are  already handled by gax layer, could u verify what error codes are retried ?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -0,0 +1,287 @@
...
+        int numActuallyCreated = sessions.size();
+        if (numActuallyCreated == 0) {
+          // No sessions returned by the server. Give up creation to avoid an infinite loop.
+          consumer.onSessionCreateFailure(
+              newSpannerException(ErrorCode.UNKNOWN, "Server did not return any sessions"),
+              remainingSessionsToCreate);
+          break;
+        }
```
snehashah16: when would this happen ?

Background context: we removed session limits and therefore this seems highly unlikely.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -0,0 +1,287 @@
...
+
+  /** Create a single session. */
+  SessionImpl createSession() {
+    // The sessionChannelCounter could overflow, but that will just flip it to Integer.MIN_VALUE,
+    // which is also a valid channel hint.
+    final Map<SpannerRpc.Option, ?> options;
+    synchronized (this) {
+      options = optionMap(SessionOption.channelHint(sessionChannelCounter++));
+    }
+    Span span = SpannerImpl.tracer.spanBuilder(SpannerImpl.CREATE_SESSION).startSpan();
```
snehashah16: should we also add tracing for everythig around BatchCreateSession..
If the traces do have any logging, please add numRequested & numReceived session counts which will greatly help with debugging.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
```diff
@@ -0,0 +1,287 @@
...
+   */
+  void asyncBatchCreateSessions(final int sessionCount, SessionConsumer consumer) {
+    // We spread the session creation evenly over all available channels.
+    int sessionCountPerChannel = sessionCount / spanner.getOptions().getNumChannels();
+    int remainder = sessionCount % spanner.getOptions().getNumChannels();
+    int numBeingCreated = 0;
+    synchronized (this) {
+      for (int i = 0; i < spanner.getOptions().getNumChannels(); i++) {
+        // Create one more session for the first X calls to fill up the remainder of the division.
+        int createCountForChannel = sessionCountPerChannel + (i < remainder ? 1 : 0);
```
snehashah16: i would simplify this

sessionCountPerChannel = sessionCount / spanner.getOptions().getNumChannels();
remainder = sessionCount % spanner.getOptions().getNumChannels();
// just add remainder to the requested count for last or first channel. 
for ( channel i : channels) {
 if ( i == 0 OR i == (numChannels - 1)) {
   sessionCountRequested =  sessionCountPerChannel + remainder
 }
}

Evenly distributing is great but wont buy us much optimization. Simple code will definitely improve readability and avoid bugs.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -437,6 +449,13 @@ public TransactionContext begin() {
         try {
           return internalBegin();
         } catch (SessionNotFoundException e) {
+          sessionReplacementCount++;
+          if (sessionReplacementCount > MAX_SESSION_REPLACES) {
+            throw SpannerExceptionFactory.newSpannerException(
+                e.getErrorCode(),
+                "Could not replace invalidated session because the maximum number of times the session may be replaced was exceeded",
```
snehashah16: This is not a good experience for our customers since this degrades their read / writes. 

When would this happen ?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -450,6 +469,13 @@ private TransactionContext internalBegin() {
     }
 
     private SpannerException handleSessionNotFound(SessionNotFoundException e) {
+      sessionReplacementCount++;
+      if (sessionReplacementCount > MAX_SESSION_REPLACES) {
+        throw SpannerExceptionFactory.newSpannerException(
+            e.getErrorCode(),
+            "Could not replace invalidated session because the maximum number of times the session may be replaced was exceeded",
+            e);
```
snehashah16: same as above.. just leave this comment open until there is a better strategy defined for prev instance

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -490,6 +516,13 @@ public TransactionContext resetForRetry() {
             return new SessionPoolTransactionContext(delegate.resetForRetry());
           }
         } catch (SessionNotFoundException e) {
+          sessionReplacementCount++;
+          if (sessionReplacementCount > MAX_SESSION_REPLACES) {
```
snehashah16: same as above.. just leave this comment open until there is a better strategy defined for prev instance

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -550,6 +584,13 @@ private SessionPoolTransactionRunner(SessionPool sessionPool, PooledSession sess
             result = runner.run(callable);
             break;
           } catch (SessionNotFoundException e) {
+            sessionReplacementCount++;
+            if (sessionReplacementCount > MAX_SESSION_REPLACES) {
```
snehashah16: same as above.. just leave this comment open until there is a better strategy defined for prev instance

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -772,7 +818,7 @@ public void close() {
         if (state != SessionState.CLOSING) {
           state = SessionState.AVAILABLE;
         }
-        releaseSession(this);
+        releaseSession(this, Position.FIRST);
```
snehashah16: could u tell me when this action will be performed ?

Should we release the (least recently used) LRU index ?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1009,7 +1055,7 @@ private void keepAliveSessions(Instant currTime) {
           logger.log(Level.FINE, "Keeping alive session " + sessionToKeepAlive.getName());
           numSessionsToKeepAlive--;
           sessionToKeepAlive.keepAlive();
-          releaseSession(sessionToKeepAlive);
+          releaseSession(sessionToKeepAlive, Position.FIRST);
```
snehashah16: if we re-balancing the pool, we should definitely release the LRU sessions.. Hope this first position does not point to (most recently used) MRU session

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1568,53 +1648,88 @@ public void run() {
         });
   }
 
+  /**
+   * Returns the minimum of the wanted number of sessions that the caller wants to create and the
+   * actual max number that may be created at this moment.
+   */
+  private int getAllowedCreateSessions(int wantedSessions) {
+    synchronized (lock) {
```
snehashah16: isn't this sort of diverging from the initial design.. we should always enqueue waiters to get session.. the waiters can decide to get more session or not based on the max value.. this seemed to have percolated outside that creation logic.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
olavloite: Gax handles the retries, and will do so based on the settings here: https://github.com/googleapis/googleapis/blob/6b2ba2ae3124c22ecb56af7102c78110b8576671/google/spanner/v1/spanner_gapic.yaml#L93

That means that the method will retry on:
* Server returns error code Unavailable
* Deadline_Exceeded (client side) after 60 seconds, and it will give up retrying after 600 seconds.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
olavloite: This happened once while stress testing the functionality on the staging environment, so I added this as a safeguard against an infinite loop. I have not seen this happen on the production environment.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
olavloite: I've added tracing for BatchCreateSessions on two levels:
1. The `BatchCreateSessionsRunnable` creates a scoped span for the entire session creation, including an annotation indicating the number of sessions being requested.
1. The `BatchCreateSessions` rpc call has its own child span that also includes the number of requested and actually returned sessions.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
olavloite: Done.
This change also has the advantage that requests for numSessions < numChannels will be executed in one rpc call instead of numSessions rpc calls.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: Sorry, this shouldn't have been in this PR and has been removed.

It could happen if a large number of sessions in the pool have been invalidated by the server without the client knowing about it, and this would safeguard against a large number of retries that all fail.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: This is called when a pooled session is returned to the session pool (`PooledSession#close()` does not close the session, but checks a session back into the pool). This is therefore the most recently used session, assuming that the client actually used the session while it was checked out, and hence is added to the head of the pool.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: I'm not sure I understand this comment correctly.

The reasoning behind what is happening here is the following:
1. A session is checked out of the pool by the maintainer, as it has determined that this is a session that needs a keepAlive query.
1. The maintainer executes a `SELECT 1` query on the session, effectively using the session.
1. The maintainer then releases the session back into the pool at the position of the MRU session, as the `SELECT 1` query should be regarded as any other usage of a session.

Should we not consider a `SELECT 1` query executed by the maintainer as a an action that marks the session as most recently used?

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
olavloite: No, this is still the same logic as the original design, only it has been moved to a separate method as it has become a little bit more complex. It is only called before createSessions(..) is called to determine the maximum number of sessions we are allowed to create. Previously, calling `canCreateSession()` was enough to ensure that it was ok to create 1 session, now it additionally checks that it is not trying to create more than the allowed number of sessions as well.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
snehashah16: i would still retry the request (same logic as numRequested > 0)

Exposing an unknown error is degradation/error to their subsequent read / write requests..

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
snehashah16: SGTM !

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
snehashah16: Thanks !!

it has a be a huge config mistake if customers have  "numSessions < numChannels"
Though this change guards against using all those number of channels.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
snehashah16: sessions in the pool are kept alive using "GetSession() rpc" or "SELECT 1" query.

Hopefully we shouldnt have this catastrophic event

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
snehashah16: sgtm!

snehashah16:
@skuruppu  - i have reviewed the crux of the CL (mostly SessionClient, SpannerImpl and SessionPool.java) 
Feel free to LGTM when u are done with the review.

snehashah16:
@olavloite - LGTM! feel free to check-in when @skuruppu is done with the review.

QQ: When the LRU session is returned to the pool, when does the BeginTxn rpc get called ?
I am curious if session is only returned to the pool after a transaction is started OR does the BeginTxn rpc occur in the write request path ?

olavloite:
@snehashah16 
> QQ: When the LRU session is returned to the pool, when does the BeginTxn rpc get called ?
> I am curious if session is only returned to the pool after a transaction is started OR does the BeginTxn rpc occur in the write request path ?

When a session is released into the pool (be it a session that is returned from a client, a session that has just been created or a session that has just executed a keep-alive query), the session pool will check whether it needs to prepare a transaction or not. So what happens is:
1. A session is returned to the pool.
1. The session pool checks whether a read/write transaction should be prepared by checking this: `currentWriteFraction < wantWriteFraction || numReadWriteWaiters > numSessionsBeingPrepared`
1. If false --> Session is added to the read queue. If true --> the following steps are executed.
1. The session will not be added to the queue of ready read- or write sessions, but instead an asynchronous BeginTransaction rpc will be started using the same executor that is also used for creating sessions.
1. When the BeginTransaction rpc has finished, the session will be returned to the first read/write waiter (if any), and otherwise it will be added to the head of the read/write sessions queue.

So the BeginTransaction rpc is not executed as part of the write request path.

However, if the write fraction setting of the session pool is too low for the number of read/write transactions that is executed, the BeginTransaction rpc will be a bottleneck, as any read/write transaction that wants a session will be placed in the wait queue and will wait for a BeginTransaction rpc to finish.

The BeginTransaction rpc can also be a bottleneck in a very high load read/write transaction situation for another reason:
1. Preparing read/write transactions in the session pool is (by default) executed using a shared grpc-transport executor which is fixed to 8 worker threads.
1. If the client application is executing (short) read/write transactions using a thread pool with more than 8 worker threads, there is a chance that the executor that is used for preparing transactions will not be able to keep up with the client application.

google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionClient.java
olavloite: Changed to retry also when the number of returned sessions is 0.

PR #6190, opened 2019-08-28
Change submodule to use https style of URL.

### LINKED ISSUE #6189: googleapis/conformance-tests submodule should use https style of URL
https://github.com/googleapis/google-cloud-java/blob/172f7a898779ef2aa6805e1a5fecedf1cefb146c/.gitmodules#L3

The current URL for that submodule uses the git-style scheme, which requires users to authenticate to GitHub using an SSH key.  This should instead use the `https://` style of URL, which doesn't require any auth since the `googleapis/conformance-tests` repo is public.

houglum:
Fixes #6189.

Relevant GitHub doc: https://help.github.com/en/articles/which-remote-url-should-i-use


houglum:
@frankyn @JesseLovelace not sure if this will break any existing workflows, but I think it's worth doing to make the new-contributor flow smoother.

PR #6211, opened 2019-09-04
Bigtable: RowMutationEntry#create for unsafe mutation creation

rahulKQL:
Introducing `RowMutationEntry#create(key, mutation)` so that mutation can be passed directly to `RowMutationEntry`.

This shall be used for HBase batching with the server-side timestamp([here](https://github.com/googleapis/java-bigtable-hbase/blob/master/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/HBaseRequestAdapter.java#L332-L335)).

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutationEntry.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutationEntry.java
@@ -51,26 +51,19 @@ public class RowMutationEntry implements MutationApi<RowMutationEntry>, Serializ
 
   /** Creates a new instance of the mutation builder. */
   public static RowMutationEntry create(@Nonnull ByteString key) {
-    return create(key, Mutation.create());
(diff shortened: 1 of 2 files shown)


rahulKQL:
@igorbernstein2 Thanks for the suggestion, I have updated method the same name.

Please have a fresh look.

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutationEntry.java
```diff
@@ -54,6 +54,18 @@ public static RowMutationEntry create(@Nonnull ByteString key) {
...
   }
 
+  /**
+   * Creates new instance of mutation builder which allows server timestamp for setCell operations.
+   *
+   * <p>NOTE: This functionality is intended for advanced usage.
+   *
+   * @see Mutation#createUnsafe() for more explanation.
+   */
+  @InternalApi("For internal usage only")
```
rahulKQL: I believe this annotation should be fine.

PR #6214, opened 2019-09-04
Bigtable: Adding ABORTED as default retry code

### LINKED ISSUE #6205: Bigtable client should retry ABORTED status codes similar to the hbase bigtable client
**Is your feature request related to a problem? Please describe.**

We've been seeing sporadic timeouts when reading from Bigtable
```
Exception in thread "main" java.lang.RuntimeException: com.google.api.gax.rpc.AbortedException: io.grpc.StatusRuntimeException: ABORTED: Error while reading table 'projects/etsy-tanuki-prod/instances/tanuki/tables/search-denormalized-listings' : Response was not consumed in time; terminating connection.(Possible causes: slow client data read or network problems)
```
In our testing, we're attempting 20 concurrent bigtable reads of a large range of rows. At least one of these reads fail within ~3 minutes with the above error. As far as we can tell, there aren't any networking issues and we're processing the rows fairly quickly.

We noticed that the ABORTED status code isn't part of the default retryable codes. After adding it to the retryable codes, the issue went away.

I think adding the ABORTED status code to the default retryable codes would be nice. Currently the hbase client has it as one of the default retryable codes. https://github.com/googleapis/java-bigtable-hbase/pull/280/files#diff-b3f2d46de9814d2abefed2339765ba01R100

 **Describe the solution you'd like**

I would like the default retryable codes to include the ABORTED code. (Or an understanding of why this is undesirable)

 **Describe alternatives you've considered**

At the moment, I'm manually adding in the ABORTED code as a retryable code.

rahulKQL:
Fixes #6205 

- Added `StatusCode.ABORTED` as default retry code to bigtable.
- Updated test case to verify the same.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
@@ -418,6 +418,11 @@ public class EnhancedBigtableStubSettings extends StubSettings<EnhancedBigtableS
 
       // Per-method settings using baseSettings for defaults.
       readRowsSettings = ServerStreamingCallSettings.newBuilder();
+
+      // Allow retrying ABORTED statuses. These will be returned by the server when the client is
+      // too slow to read the rows. This makes sense for the java client because retries happen
+      // after the row merging logic. Which means that the retry will not be invoked until the
+      // current buffered chunks are consumed.
       readRowsSettings
           .setRetryableCodes(
               ImmutableSet.<Code>builder()
@@ -430,11 +435,7 @@ public class EnhancedBigtableStubSettings extends StubSettings<EnhancedBigtableS
       // Point reads should use same defaults as streaming reads, but with a shorter timeout
       readRowSettings = UnaryCallSettings.newUnaryCallSettingsBuilder();
       readRowSettings
-          .setRetryableCodes(
-              ImmutableSet.<Code>builder()
-                  .addAll(baseDefaults.readRowsSettings().getRetryableCodes())
-                  .add(Code.ABORTED)
-                  .build())
+          .setRetryableCodes(readRowsSettings.getRetryableCodes())
           .setRetrySettings(
               baseDefaults
                   .readRowsSettings()


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -420,14 +422,14 @@ private Builder() {
       // Per-method settings using baseSettings for defaults.
       readRowsSettings = ServerStreamingCallSettings.newBuilder();
       readRowsSettings
-          .setRetryableCodes(baseDefaults.readRowsSettings().getRetryableCodes())
```
igorbernstein2: We should just add ABORTED to gapic codes:

```java
// Allow retrying ABORTED statuses. These will be returned by the server when the client is too slow to read the rows.
// This makes sense for the java client because retries happen after the row merging logic. Which means that the retry
// will not be invoked until the current buffered chunks are consumed. 
readRowsSettings.setRetryableCodes(
  ImmutableSet.builder()
   .addAll(baseDefaults.readRowsSettings().getRetryableCodes())
   .add(baseDefaults.readRowsSettings().getRetryableCodes())
   .build());
```

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -81,6 +81,9 @@
   private static final Set<Code> IDEMPOTENT_RETRY_CODES =
       ImmutableSet.of(Code.DEADLINE_EXCEEDED, Code.UNAVAILABLE);
 
+  private static final Set<Code> READ_ROWS_RETRY_CODES =
+      ImmutableSet.of(Code.DEADLINE_EXCEEDED, Code.UNAVAILABLE, Code.ABORTED);
+
```
igorbernstein2: Please remove and see the comment below

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
rahulKQL: By any chance do you mean something like this:
```java
readRowsSettings.setRetryableCodes(
  ImmutableSet.builder()
   .addAll(baseDefaults.readRowsSettings().getRetryableCodes())
   .add(Code.ABORTED)
   .build());
```

igorbernstein2: Oops , sorry about that. Yes that’s what I meant

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -420,14 +419,22 @@ private Builder() {
       // Per-method settings using baseSettings for defaults.
       readRowsSettings = ServerStreamingCallSettings.newBuilder();
       readRowsSettings
-          .setRetryableCodes(baseDefaults.readRowsSettings().getRetryableCodes())
+          .setRetryableCodes(
+              ImmutableSet.<Code>builder()
```
igorbernstein2: Please add a comment to explain why we are diverging from the standard gapic settings:

```
// Allow retrying ABORTED statuses. These will be returned by the server when the client is too slow to read the rows.
// This makes sense for the java client because retries happen after the row merging logic. Which means that the retry
// will not be invoked until the current buffered chunks are consumed. 
```

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -420,14 +419,22 @@ private Builder() {
...
+                  .add(Code.ABORTED)
+                  .build())
           .setRetrySettings(baseDefaults.readRowsSettings().getRetrySettings())
           .setIdleTimeout(Duration.ofMinutes(5));
 
       // Point reads should use same defaults as streaming reads, but with a shorter timeout
       readRowSettings = UnaryCallSettings.newUnaryCallSettingsBuilder();
       readRowSettings
-          .setRetryableCodes(baseDefaults.readRowsSettings().getRetryableCodes())
+          .setRetryableCodes(
```
igorbernstein2: ```.setRetryableCodes(readRowsSettings.getRetryableCodes())```

PR #6232, opened 2019-09-06
Extract core modules to https://github.com/googleapis/java-core

chingor13:


chingor13:
The datastore failure happens occasionally locally as well. Filed #6234

PR #6235, opened 2019-09-06
Extract google-cloud-trace to https://github.com/googleapis/java-trace

chingor13:


chingor13:
This test failure is expected because the client is gone from this repo. We will disable that check after its merged.

Issue #5227, opened 2019-05-28
PubSub: subscriber.stopAsync().awaitTerminated() does not destroy respective threads


menulis:
#### Environment details

OS type and version: macOS 10.14.5 / Debian 4.9.168-1 (App Engine Flex)
Java version: Oracle 1.8.0_181 / openjdk version "1.8.0_212"
google-cloud-java version: 1.75.0

#### Steps to reproduce

  1. Create a subscriber, call startAsync() followed by stopAsync() as per documentation
  2. Observe the number of threads before and after calling stopAsync()
  3. The number of threads is growing

#### Code example

```kotlin
import com.google.api.gax.core.ExecutorProvider
import com.google.api.gax.core.FixedExecutorProvider
import com.google.cloud.ServiceOptions
import com.google.cloud.pubsub.v1.Subscriber
import com.google.pubsub.v1.ProjectSubscriptionName
import java.util.concurrent.Executors
import java.util.concurrent.TimeUnit

fun main() {
    val subscriptionName = ProjectSubscriptionName.of(
        ServiceOptions.getDefaultProjectId(), "test"
    )

    for (x in 1..2) {
        var i = 0
        val subscriber = Subscriber.newBuilder(subscriptionName)
            { message, consumer -> consumer.ack() }
            .build()
        println("$x/${++i}: ${Thread.activeCount()}: ${Thread.getAllStackTraces().entries.map { it.key.name }}")
        subscriber.startAsync().awaitRunning()
        println("$x/${++i}: ${Thread.activeCount()}: ${Thread.getAllStackTraces().entries.map { it.key.name }}")
        subscriber.stopAsync().awaitTerminated()
        println("$x/${++i}: ${Thread.activeCount()}: ${Thread.getAllStackTraces().entries.map { it.key.name }}")
    }

    for (x in 3..8) {
        var i = 0
        val executorProvider: ExecutorProvider =
            FixedExecutorProvider.create(Executors.newSingleThreadScheduledExecutor())
        val subscriber = Subscriber.newBuilder(subscriptionName)
        { message, consumer -> consumer.ack() }
            .setExecutorProvider(executorProvider)
            .setSystemExecutorProvider(executorProvider)
            .build()
        subscriber.startAsync().awaitRunning()
        println("$x/${++i}: ${Thread.activeCount()}: ${Thread.getAllStackTraces().entries.map { it.key.name }}")
        subscriber.stopAsync().awaitTerminated()
[...]

PR #6238, opened 2019-09-09
Make sure that background resources are closed when the PubSub subscriber fails

mdirkse:
Completes fix of #5227.

We [found](https://github.com/spinnaker/spinnaker/issues/4835) that the PubSub Subscriber can go into a thread creation loop when it fails on startup, eventually causing system-wide thread starvation. This PR fixes that problem by making sure that all background resources are also shut down on failure (instead of just when stopped).

### DIFF
--- a/google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
+++ b/google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
@@ -176,6 +176,38 @@ public class SubscriberTest {
     }
   }
 
+  @Test(expected = IllegalStateException.class)
+  public void testFailedChannel_shutdownBackgroundResources() throws Exception {
+    ExecutorProvider provider =
+        new ExecutorProvider() {
+          @Override
+          public boolean shouldAutoClose() {
+            return true;
+          }
+
+          @Override
+          public ScheduledExecutorService getExecutor() {
+            return fakeExecutor;
+          }
+        };
+
+    Subscriber subscriber =
+        startSubscriber(getTestSubscriberBuilder(testReceiver).setExecutorProvider(provider));
+
+    // Fatal error
+    fakeSubscriberServiceImpl.sendError(new StatusException(Status.INVALID_ARGUMENT));
+
+    try {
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/FakeFixedExecutorProvider.java
```diff
@@ -0,0 +1,26 @@
+package com.google.cloud.pubsub.v1;
```
hannahrogers-google: We should be able to delete this class

google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
```diff
@@ -176,6 +176,29 @@ public void testFailedChannel_fatalError_subscriberFails() throws Exception {
     }
   }
 
+  @Test(expected = IllegalStateException.class)
+  public void testFailedChannel_shutdownBackgroundResources() throws Exception {
+    Subscriber subscriber =
```
hannahrogers-google: We shouldn't have to create a new FakeFixedExecutorProvider class.  We can just create our own custom provider that will autoClose.  For example:

```
ExecutorProvider provider =
        new ExecutorProvider() {
          @Override
          public boolean shouldAutoClose() {
            return true;
          }

          @Override
          public ScheduledExecutorService getExecutor() {
            return fakeExecutor;
          }
        };
```

We can then just pass in the provider when we .setExecutorProvider

hannahrogers-google:
@mdirkse, did you want to take a look at the requested changes so that we can merge this PR?

mdirkse:
Yes! Sorry, will do so today.

google-cloud-clients/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/SubscriberTest.java
mdirkse: Done! Sorry for the delay.

mdirkse:
Fixed the formatting error.

PR #6242, opened 2019-09-09
Extract google-cloud-speech to https://github.com/googleapis/java-speech

chingor13:


elharo:
Looks like a kokoro job needs to be deleted.

PR #6250, opened 2019-09-10
storage: add javax.annotation dependency

jeanbza:
It turns out that generated sources use a @javax.annotation.Generated
annotation. This PR adds it as a dependency, without which things don't
compile.

### DIFF
--- a/google-cloud-testing/storage-benchwrapper/pom.xml
+++ b/google-cloud-testing/storage-benchwrapper/pom.xml
@@ -112,5 +112,13 @@
       <artifactId>netty-tcnative-boringssl-static</artifactId>
       <version>2.0.25.Final</version>
     </dependency>
+
+    <!-- Generated classes are generated with a @javax.annotation.Generated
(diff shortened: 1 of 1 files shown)


google-cloud-testing/storage-benchwrapper/pom.xml
```diff
@@ -112,5 +112,13 @@
       <artifactId>netty-tcnative-boringssl-static</artifactId>
       <version>2.0.25.Final</version>
     </dependency>
+
+    <!-- Generated classes are generated with a @javax.annotation.Generated
+      annotation, so we need this dependency or else they don't compile. -->
+    <dependency>
+      <groupId>javax.annotation</groupId>
+      <artifactId>javax.annotation-api</artifactId>
+      <version>1.2</version>
```
chingor13: Please make this 1.3.2

jadekler: Done

PR #6251, opened 2019-09-11
storage benchwrapper: update google-cloud-storage version and keep in sync

chingor13:


kolea2:
@jadekler yep! This version number will now get bumped every time we do a release or snapshot bump with this annotation (part of releasetool)

jadekler:
Sounds good thanks!

PR #6255, opened 2019-09-11
Add repo metadata files for all the existing clients

chingor13:
This file helps us programmatically track the release level of our clients across languages/repositories


google-cloud-clients/google-cloud-contrib/.repo-metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "cloudasset",
```
BenWhitehead: How come the name here is `cloudasset` whereas the directory is `contrib`?

chingor13: oops, that file should not have been included

PR #6269, opened 2019-09-13
Pub/Sub: max_rpc_timeout should be short like 10s

anguillanneuf:
If we set max PRC timeout to be 10 minutes, we are also not retrying.


hongalex:
Same as https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2398, I think the current sample is ok.

PR #6287, opened 2019-09-17
Fix flaky test in google-cloud-translate

pkumusic:
Fixes b/141087597.


chingor13:
Sorry, we missed this, but we did switch the translate test to numbers which are translated more consistently.

PR #6293, opened 2019-09-18
Implement Table-level IAM Policy controls.

alexoneill:


kolea2:
LGTM, will merge unless @igorbernstein2 has any additional comments

rahulKQL:
@Kolea2, @igorbernstein2 Sorry for pointing this now, But I think these methods are available in BigtableInstanceAdminClient already

https://github.com/googleapis/google-cloud-java/blob/8177abd29230443ccd765df65ad87b4c05786453/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClient.java#L1077-L1304

I might have missed checking the open PRs till now.

igorbernstein2:
They need to be duplicated because one set is for instances and  the other for tables

PR #6303, opened 2019-09-19
Bigquery: Expose CMEK functionality for ML models.

### LINKED ISSUE #6300: BigQuery: expose CMEK functionality for ML models
This request is to expose customer managed encryption key (CMEK) for BigQuery ML models.

It uses the same message (encryptionConfig) as is present on the rest interface
for managed tables, e.g. the expression of a related KMSKeyName.

ML models aren't instantiated via the API directly, but via CREATE MODEL sql query.
The CMEK functionality should be present in the API for other relevant mechanisms, 
such as models.get and models.update.

For more information, contact @shollyman.

pmakani:
Fixes #6300


pmakani:
> Thanks! Please add some minimal testing e.g. ModelInfoTest to ensure we propagate through builders, etc.

Done, PTAL.

PR #6328, opened 2019-09-24
Updated JavaDoc to mention connection pooling over ManagedChannel

### LINKED ISSUE #5190: Bigtable: Update docs to make it obvious that a connection pool is being used

rahulKQL:
Fixes #5190

Updated JavaDoc at `EnhancedBigtableStubSetting#defaultTransportChannelProvider` to mention about connection pooling over `ManagedChannel`.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
@@ -36,6 +36,8 @@ import javax.annotation.Nonnull;
  *
  * <ul>
  *   <li>The default service address (bigtable.googleapis.com) and default port (443) are used.
+ *   <li>The transport provider is configured with a channel pool that contains twice as many
+ *       connections as CPUs.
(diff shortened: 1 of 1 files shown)


rahulKQL:
Thanks! for the direction. I have added that change, PTAL.

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -54,6 +54,10 @@
  * BigtableDataSettings settings = builder.build();
  * }</pre>
  *
+ * <p>{@link com.google.api.gax.rpc.TransportChannelProvider} provides thread-safe {@link
+ * io.grpc.ManagedChannel channels} for this service. Each of these channel manage connection
+ * pooling; means it automatically creates and destroys connection as they needed.
+ *
```
igorbernstein2: I think this can should be a bullet point on line 39:
`The transport provider is configured with a channel pool that contains twice as many connections as CPUs`

RahulKesharwani: Thanks, Have updated with the same.

PR #6331, opened 2019-09-24
Spanner: Close ResultSet from session keep-alive request

### LINKED ISSUE #6330: Spanner: ResultSet from session keep-alive request is not closed.
When I ran an application using google-cloud-spanner 1.30.0 with opencensus-impl 0.23.0, I noticed this log message:

```
io.opencensus.implcore.trace.RecordEventsSpanImpl finalize
SEVERE: Span CloudSpannerOperation.ExecuteStreamingQuery is GC'ed without being ended.
```

It looks like this is caused by the session keep-alive request: https://github.com/googleapis/google-cloud-java/blob/ca141fcb4628a85bf775e8550eedb1e8833c52ea/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L790-L796

If I understand correctly, the ResultSet contains a ResumableStreamIterator, which contains the span representing the request.  The issue is that keepAlive doesn't call close on the ResultSet, so the span is never ended.  Calling close on the ResultSet seems to fix the issue.

sebright2:
Fixes #6330

Closing the ResultSet also closes the span associated with the keep-alive
request, which fixes a tracing issue.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -789,10 +789,12 @@ final class SessionPool {
 
     private void keepAlive() {
       markUsed();
-      delegate
-          .singleUse(TimestampBound.ofMaxStaleness(60, TimeUnit.SECONDS))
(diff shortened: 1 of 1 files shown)


sebright:
Thanks for the review!

PR #6345, opened 2019-09-25
Spanner: Allow BeginTransaction to be inlined with first statement

olavloite:
The Spanner client library will normally prepare a fraction of the sessions in the session pool with a read/write transaction. These sessions are used when an application requests a read/write transaction. This eliminates the need for a BeginTransaction RPC to be executed as part of a transaction, as the call has already been executed in advance by the session pool.

This can however be less efficient in some specific cases:
* When the number of read/write transactions that are needed is higher than the number of transactions the session pool is able to prepare (preparing sessions is done using the same executor that is also used to create sessions, and is by default limited to 8 worker threads).
* When the fraction of read/write transactions is higher than what has been configured by the user (the default is 0.2).

Cloud Spanner also supports starting a new transaction as part of a query or DML statement. This eliminates the need for a separate round-trip to the server for starting a transaction. This PR adds the possibility to instruct the Spanner client to start read/write transactions by including this option in the first query/update statement of a read/write transaction, instead of preparing the read/write transactions in the session pool. This can improve the overall read/write transaction execution performance.


elharo:
now in a separate repo

PR #6360, opened 2019-09-26
Datastore : add GqlQuery#setBinding method which accept null value

### LINKED ISSUE #6357: com.google.cloud.datastore.GqlQuery.Builder does not accept null values and results in an exception
Query.newGqlQueryBuilder(type, gql).setBinding('testparam', null)
results in an exception because there is no way to set a null value.

Looking at the source code, I supposed a temporary workaround might be to bind to an empty List as there is a check in the binding code to set that to NullValue.

Currently, I am doing a workaround by putting null in the query String and using setAllowLiteral(true), but this is subOptimal as setAllowLiteral defaults to false.

What needs to be added is a setBinding that accepts NullValue, similar to how 
StructuredQuery.PropertyFilter.eq allows 

This is a simple addition as the MARSHALLER code for NullValue already exist.  The implementation can be 

    public Builder<V> setNullBinding(String name) {
      namedBindings.put(name, toBinding(NullValue.MARSHALLER, new ArrayList()));
      return this;
    }

toBinding checks for an empty list, and will bind NullValue

athakor:
Fixes #6357

### DIFF
--- a/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
+++ b/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
@@ -296,7 +296,7 @@ public final class GqlQuery<V> extends Query<V> {
      *
      * @param name of the binding
      */
-    public Builder<V> setBinding(String name) {
+    public Builder<V> setNullBinding(String name) {
       namedBindings.put(name, toBinding(NullValue.MARSHALLER, new ArrayList()));
       return this;
     }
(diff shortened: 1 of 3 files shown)


google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/GqlQuery.java
```diff
@@ -291,6 +291,16 @@ static Binding fromPb(com.google.datastore.v1.GqlQueryParameter argPb) {
       return this;
     }
 
+    /**
+     * Sets a new named binding with null value.
+     *
+     * @param name of the binding
+     */
+    public Builder<V> setBinding(String name) {
```
BenWhitehead: Please rename this method to `setNullBinding` so that the name of the method matches the behavior that is implemented.

athakor: Done. PTAL

PR #6375, opened 2019-09-30
Redis: Added system test 

athakor:
towards #5947

### DIFF
--- a/google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
+++ b/google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
@@ -43,7 +43,7 @@ public class ITSystemTest {
       INSTANCE_NAME_PREFIX + "-" + UUID.randomUUID().toString().substring(0, 8);
   private static final String LOCATION = "us-central1";
   private static final int MEMORY_SIZE_GB = 1;
-  private static final String AUTHORIZED_NETWORK = System.getProperty("redis.network");
+  private static final String AUTHORIZED_NETWORK = System.getProperty("redis.network", "default");
 
   @BeforeClass
   public static void beforeClass() throws Exception {
@@ -52,10 +52,7 @@ public class ITSystemTest {
 
     /** Creates a Redis instance based on the specified tier and memory size. */
     LocationName parent = LocationName.of(projectId, LOCATION);
-    String authorizedNetwork =
-        AUTHORIZED_NETWORK == null || AUTHORIZED_NETWORK.isEmpty()
-            ? "projects/" + projectId + "/global/networks/default"
-            : "projects/" + projectId + "/global/networks/" + AUTHORIZED_NETWORK;
+    String authorizedNetwork = "projects/" + projectId + "/global/networks/" + AUTHORIZED_NETWORK;
     Instance instance =
         Instance.newBuilder()
             .setTier(TIER)
@@ -72,6 +69,7 @@ public class ITSystemTest {
     /** Deletes a specific Redis instance. Instance stops serving and data is deleted. */
     InstanceName name = InstanceName.of(projectId, LOCATION, INSTANCE);
     client.deleteInstanceAsync(name);
+    log.info("redis instance deleted successfully.");
     client.close();
   }
 


athakor:
@chingor13 Build failed due to Redis service is not enable or activate for this project.

google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
```diff
@@ -0,0 +1,92 @@
...
+  @After
+  public void tearDown() {
+    client.close();
+  }
+
+  @Test
+  public void testCreateInstance() throws Exception {
+    LocationName parent = LocationName.of(projectId, LOCATION);
+    Instance instance = Instance.newBuilder().setTier(TIER).setMemorySizeGb(MEMORY_SIZE_GB).build();
+    Instance response = client.createInstanceAsync(parent, INSTANCE, instance).get();
```
chingor13: When we create the instance, we'll need to make this some kind of unique id so that the test can possibly run in parallel on the same test project at the same time.

chingor13:
@athakor I enabled the Redis API on the test project and restarted the test. It's now failing with a validation error about an invalid network parameter.

google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
athakor: Done

google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    client = CloudRedisClient.create();
+    projectId = ServiceOptions.getDefaultProjectId();
+
+    /** Creates a Redis instance based on the specified tier and memory size. */
+    LocationName parent = LocationName.of(projectId, LOCATION);
+    Instance instance = Instance.newBuilder().setTier(TIER).setMemorySizeGb(MEMORY_SIZE_GB).build();
+    client.createInstanceAsync(parent, INSTANCE, instance).get();
```
chingor13: This is failing with an invalid config: `Invalid authorized network 'default'. Legacy networks are not supported.`

chingor13: This is likely because our test project is very old and has some older grandfathered in settings (like default networking) still configured. I'll investigate some more.

chingor13: I'm hesitant to change the default network for the test project, so I created a new `redis-vpc` network in our test project. Let's make the test configurable (via a maven param like `-Dredis.network=redis-vpc` which we can add [here](https://github.com/googleapis/google-cloud-java/blob/cd38245528822af3a321fcd0a2edac4352f1c3e9/.kokoro/presubmit/redis-it.cfg#L11) (and in the continuous integration config). We can default that value to `default` for most other users who run the tests.

ajaaym: sounds good. Thanks @chingor13

athakor:
@chingor13 PTAL

google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
```diff
@@ -0,0 +1,98 @@
...
+  private static String projectId;
+
+  private static final Logger log = Logger.getLogger(ITSystemTest.class.getName());
+  private static final Instance.Tier TIER = Instance.Tier.BASIC;
+  private static final String INSTANCE_NAME_PREFIX = "test-instance";
+  private static final String INSTANCE =
+      INSTANCE_NAME_PREFIX + "-" + UUID.randomUUID().toString().substring(0, 8);
+  private static final String LOCATION = "us-central1";
+  private static final int MEMORY_SIZE_GB = 1;
+  private static final String AUTHORIZED_NETWORK = System.getProperty("redis.network");
```
chingor13:
```suggestion
  private static final String AUTHORIZED_NETWORK = System.getProperty("redis.network", "default");
```

google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
```diff
@@ -0,0 +1,98 @@
...
+  public static void beforeClass() throws Exception {
+    client = CloudRedisClient.create();
+    projectId = ServiceOptions.getDefaultProjectId();
+
+    /** Creates a Redis instance based on the specified tier and memory size. */
+    LocationName parent = LocationName.of(projectId, LOCATION);
+    String authorizedNetwork =
+        AUTHORIZED_NETWORK == null || AUTHORIZED_NETWORK.isEmpty()
+            ? "projects/" + projectId + "/global/networks/default"
+            : "projects/" + projectId + "/global/networks/" + AUTHORIZED_NETWORK;
```
chingor13: We can simplify this if we set the default above:
```java
String authorizedNetwork = "projects/" + projectId + "/global/networks/" + AUTHORIZED_NETWORK;
```

google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
athakor: Done

google-cloud-clients/google-cloud-redis/src/test/java/com/google/cloud/redis/it/ITSystemTest.java
athakor: Done. @chingor13  PTAL

PR #6379, opened 2019-09-30
fix: bump grpc-google-iam to 0.13.0 to introduce GetPolicyOptions

jkwlui:
This is blocking new regeneration of libraries with updated IAM methods.


elharo:
Is this ready to merge?

PR #6399, opened 2019-10-01
Recommend linkage checker enforcer rule instead of dependency convergence

elharo:
@chingor13 @kolea2 @saturnism


saturnism:
hiya! it looks like the changes are cleaning up the verbiages. where does it recommend linkage checker? thanks!

elharo:
Damn it, looks like I may have put those changes in the wrong client. Let me find them.

PR #6426, opened 2019-10-03
Spanner: Support the use of io.grpc.CallCredentials

### LINKED ISSUE #6373: How do I apply a `io.grpc.CallCredentials` to `ServiceOptions`?
I have a private class that generates `io.grpc.CallCredentials` to authorize to Cloud Spanner. Apparently the `ServiceOptions.Builder.setCredentials` only knows about `com.google.auth.Credentials`. I know that the Cloud Spanner communication is done over gRPC. Is there a way to apply my `CallCredentials`?

olavloite:
Allow the user to supply io.grpc.CallCredentials instead of only com.google.auth.Credentials. Any CallCredentials supplied will take precedence above the Credentials set on SpannerOptions.

Fixes #6373

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
@@ -47,6 +47,7 @@ import com.google.cloud.spanner.admin.instance.v1.stub.GrpcInstanceAdminStub;
 import com.google.cloud.spanner.admin.instance.v1.stub.InstanceAdminStub;
 import com.google.cloud.spanner.v1.stub.GrpcSpannerStub;
 import com.google.cloud.spanner.v1.stub.SpannerStub;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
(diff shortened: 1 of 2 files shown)


Capstan:
My one concern is that this seems more widely applicable to `ServiceOptions` rather than `SpannerOptions`, especially for any server backed by gRPC. I don't know if it's desirable to spill implementation details into these classes.

My other thought is on the other side of the problem, and whether we should come up with a sane way to have a private class that extends `com.google.auth.Credentials` such that it'd obviate this request.

/cc @tsalomie

olavloite:
> My one concern is that this seems more widely applicable to `ServiceOptions` rather than `SpannerOptions`, especially for any server backed by gRPC. I don't know if it's desirable to spill implementation details into these classes.
> 
I can see that one too, and adding this option to `ServiceOptions` is obviously a trivial change. However, the way that the different client libraries setup and invoke the gRPC calls can differ quite a lot. Spanner happens to already have one method `newCallContext` that is called _for every gRPC_ call to setup the call context to use, which makes it easy to respect the CallCredentials setting. I'm not sure whether that part of the change would be equally trivial for other client libraries.

olavloite:
@Capstan Any update on whether this change is still something you would need, or did you find some other way to use `CallCredentials` instead of `Credentials`?

Capstan:
@olavloite I'm discussing this with Google-internal colleagues on the team that needs it. I don't believe we have an alternate solution at the moment, but I've not checked in recently and need to sync with them.

Capstan:
I am going to pursue – after the weekend – extending the internal support so that also can handle com.google.auth.Credentials. I'm fairly positive that will work out of the box and will obviate the need for this PR.

chingor13:
@olavloite This PR will need to be reopened in the https://github.com/googleapis/java-spanner repo as we've migrated the google-cloud-spanner code

olavloite:
@Capstan Is this PR still needed? If so, we need to migrate it to the new repository.

Capstan:
Yes, we still need this functionality.

olavloite:
Moved to the new repo and merged there.

PR #6429, opened 2019-10-03
Spanner: Change default MinSessions to 100

### LINKED ISSUE #6169: Spanner: BatchCreateSessions

olavloite:
Changes the default MinSessions configuration for Cloud Spanner to 100 sessions. The sessions are created using the new BatchCreateSessions API of Spanner, meaning that initializing the
pool will finish quickly.

Fixes #6169

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPoolOptions.java
@@ -35,7 +35,10 @@ public class SessionPoolOptions {
   private final long initialWaitForSessionTimeoutMillis;
 
   private SessionPoolOptions(Builder builder) {
-    this.minSessions = builder.minSessions;
+    // minSessions > maxSessions is only possible if the user has only set a value for maxSessions.
+    // We allow that to prevent code that only sets a value for maxSessions to break if the
+    // maxSessions value is less than the default for minSessions.
+    this.minSessions = Math.min(builder.minSessions, builder.maxSessions);
     this.maxSessions = builder.maxSessions;
     this.maxIdleSessions = builder.maxIdleSessions;
     this.writeSessionsFraction = builder.writeSessionsFraction;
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/RetryOnInvalidatedSessionTest.java
```diff
@@ -140,7 +134,7 @@
   private static final Statement UPDATE_STATEMENT =
       Statement.of("UPDATE FOO SET BAR=1 WHERE BAZ=2");
   private static final long UPDATE_COUNT = 1L;
-  private static final int MAX_SESSIONS = 10;
+  //  private static final int MAX_SESSIONS = 10;
```
skuruppu: Why is this commented out?

olavloite: Good spot. The line needed to be removed, as it would be invalid to configure a session pool with `MaxSessions=10` now that `MinSessions=100` is the default. I was testing a couple of things during testing, and therefore I had only commented it out, instead of actually removing it.

This does trigger another thought, though: Changing the default of MinSessions to 100, means that anyone that currently has a session pool configuration with MaxSessions set to anything between 1 and 99 (inclusive) will get an error after this change, as the session pool configuration will be invalid (`MaxSessions >= MinSessions` is enforced by the session pool). Do we want to:
1. Accept this, and instruct users to change their session pool configuration in these cases?
1. Change the behavior of `SessionPoolConfiguration.Builder.setMaxSessions(int)` into automatically also change `MinSessions` if the new value for `MaxSessions` is less than the current value for `MinSession`?
1. Something else?

skuruppu: I prefer option 2. If the user has bother to explicitly set a max session size, then I feel that we should respect it.
1. If user has set both min sessions and max sessions, use the user settings.
2. If user has only set max sessions but not min sessions, then set min sessions = max sessions.
3. If user hasn't set either min or max sessions, then set min sessions to 100.

Hope that helps.

PR #6430, opened 2019-10-03
fix(logging): add ability to filter MDC labels

### LINKED ISSUE #6320: Add possibility to control MDC values for adding to log message
### The issue:
All values from MDC are written as labels to each log message, even if there is no need in some of them.
Usually MDC is populated by a built-in components (like MDCInsertingServletFilter) with all possible values, but only several of them are used in log message.

```
google-cloud-java/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java

    for (Map.Entry<String, String> entry : e.getMDCPropertyMap().entrySet()) {
      if (null != entry.getKey() && null != entry.getValue()) {
        builder.addLabel(entry.getKey(), entry.getValue());
      }
    }
```

### Possible solution:
1. Add possibility to control which MDC values should be added to log message.
2. Add Enhancer to add required values of MDC.

codyoss:
Currently all MDC values are being written as labels in a LogEntry.
This is not desirable for everyone. By adding MDCEventEnhancer a user
now has the ability to easily filter which MDC entries get converted
into labels. To maintain backwards compatability the default
implementation is to write all entries as labels. If a user wants to
filter these, they just need to extend MDCEventEnhancer, override the
filtering, and add the new classpath to their logback.xml.

Fixes #6320

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/LoggingAppenderTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/LoggingAppenderTest.java
@@ -296,9 +296,8 @@ public class LoggingAppenderTest {
     loggingAppender.addLoggingEventEnhancer(CustomLoggingEventEnhancer2.class.getName());
     loggingAppender.start();
     loggingAppender.doAppend(loggingEvent);
-
     verify(logging);
-
+    MDC.remove("mdc1");
     Map<String, String> capturedArgumentMap =
         capturedArgument.getValue().iterator().next().getLabels();
     assertThat(capturedArgumentMap.get("mdc1")).isNull();


google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/MDCEventEnhancer.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Map;
+
+/**
+ * MDCEventEnhancer takes values found in the MDC property map and adds them as labels to the {@link
+ * LogEntry}. This {@link LoggingEventEnhancer} is turned on by default. If you wish to filter which
+ * MDC values get added as labels to your {@link LogEntry}, extend this class and override {@link
+ * MDCEventEnhancer#labelFilter(Map.Entry)} accordingly. The default behavior is to add all entries
+ * as labels.
+ */
+public class MDCEventEnhancer implements LoggingEventEnhancer {
```
chingor13: Do you think it's worthwhile to maintain this extra surface area moving forward? It seems like the `LoggingEventEnhancer` interface is simple enough for someone to implement a filtered label enhancer. The user now extends a base class and overrides one method vs. implementing one interface method (which looks like it's a functional interface to begin with).

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/MDCEventEnhancerTest.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Before
+  public void setUp() {
+    logging = EasyMock.createStrictMock(Logging.class);
+    loggingAppender = new TestLoggingAppender();
+  }
+
+  @Test
+  public void testDefaultMDCEventEnhancer() {
```
chingor13: I would probably leave this test in the `AppenderTest` as we're testing the appender's default behavior. If you want to test the MDCEventEnhancer directly, I'd put that test here.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/MDCEventEnhancerTest.java
```diff
@@ -0,0 +1,121 @@
...
+    MDC.remove("mdc3");
+    Map<String, String> capturedArgumentMap =
+        capturedArgument.getValue().iterator().next().getLabels();
+    assertThat(capturedArgumentMap.get("mdc1")).isEqualTo("value1");
+    assertThat(capturedArgumentMap.get("mdc2")).isNull();
+    assertThat(capturedArgumentMap.get("mdc3")).isEqualTo("value3");
+  }
+
+  @Test
+  public void testCustomMDCEventEnhancer() {
```
chingor13: Rather than implementing a custom event enhancer, and testing its e2e behavior, I might test that the appender can configure one or more custom enhancers that we verify are called.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/MDCEventEnhancer.java
codyoss: I think so. Only because if we don't expose this as a `LoggingEventEnhancer` I can't think of a good way to give users a hook into this without breaking behavior and removing automatically adding values as labels. If we are okay with changing that behavior, that is for sure another option.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/MDCEventEnhancerTest.java
codyoss: That is fair

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/MDCEventEnhancer.java
chingor13: We could still have this class as a non-public implementation detail of that continues to implement the default behavior. By maintaining additional surface area, I mean encouraging people to extend this class vs. writing their own custom `LoggingEventEnhancer` implementation from scratch.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/MDCEventEnhancerTest.java
codyoss: okay

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/MDCEventEnhancer.java
codyoss: Ah, gottcha. I like that, good idea.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/FilteredMVCEventEnhancer.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2017 Google LLC
```
BenWhitehead: `s/2017/2019/g`

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/MDCEventEnhancerTest.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2017 Google LLC
```
BenWhitehead: `s/2017/2019/g`

codyoss:
> I thought of a possible behavior breaking change: if a user already added an additional `LoggingEventEnhancer`, they would now lose the default label setting behavior.

@chingor13 That is a good point. I did document this behavior on the class, but I suppose it is technically breaking. Short of exposing a custom hook that only relates specifically to MDCs I don't know a good way around this. I am open to suggestions.

codyoss:
What is our compatibility promise for artifacts labeled alpha? The version here is `0.111.1-alpha-SNAPSHOT`.

chingor13:
We can break the behavior, but we definitely need to document it in the release notes.

codyoss:
I vote we take the breaking behavior and document it in the release notes. I think this implementation will offer more flexibility for the users as the library. Using `LoggingEventEnhancer ` is not well documented in our examples today so I anticipate little breakage. The commit that introduced this MDC feature fbb4776 was not from too far back.

google-cloud-clients/google-cloud-contrib/google-cloud-logging-logback/src/test/java/com/google/cloud/logging/logback/MDCEventEnhancerTest.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Collections;
+import org.junit.Before;
+import org.junit.Test;
+
+public class MDCEventEnhancerTest {
+  private MDCEventEnhancer classUnderTest;
+
+  @Before
+  public void setUp() {
+    classUnderTest = new MDCEventEnhancer();
```
elharo: could just initialize the field. That lets it be final and makes this simpler.

elharo:
The alpha status is wonky. Some artifacts are alpha, some are beta, and some are GA.

codyoss:
@chingor13 Thoughts on this change?

codyoss:
I am going to merge this. Once done I will cherry-pick these changes to goolgeapis/java-logging

gendolf3d:
Hello! I'm just curious when it's being planned to merge fix to  `goolgeapis/java-logging` project?
Thank you in advance!

PR #6431, opened 2019-10-03
[Bigtable] Updated javaDoc for callable being wrapper around read & write

### LINKED ISSUE #4054: Bigtable: Improve docs to make it clear that all methods are wrappers around callables

rahulKQL:
Fixes #4054

Updated JavaDoc for callable being wrapper around read & write Operation.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -74,10 +74,19 @@ import javax.annotation.Nullable;
  *   <li>A "callable" method, like `readRowsCallable()`. This type of method takes no parameters and
  *       returns an immutable API callable object, which can be used to initiate calls to the
  *       service.
- *   <li>Each read or write operation in this client is built around the callables in the form of
- *       synchronized or asynchronized operation. These operation can also be achieved by directly
- *       using callables.
  * </ol>
+ *
+ * <p>Taking ReadRows as an example for callable:
+ *
+ * <pre>{@code
+ * // These two invocation are equivalent
+ * ServerStream<Row> stream1 = client.readRows(query);
+ * ServerStream<Row> stream2 = client.readRowsCallable().call(query);
+ *
+ * // These two invocation are also equivalent
+ * client.readRowsAsync(query, observer);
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -114,6 +114,23 @@
  *
  * BigtableDataClient client = BigtableDataClient.create(settings.build());
  * }</pre>
+ *
```
igorbernstein2: I think this should be tied more to the paragraph of "The surface of this class includes several types of Java methods"

rahulKQL: Thanks for the pointer, Have updated accordingly.

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -74,6 +74,9 @@
  *   <li>A "callable" method, like `readRowsCallable()`. This type of method takes no parameters and
  *       returns an immutable API callable object, which can be used to initiate calls to the
  *       service.
+ *   <li>Each read or write operation in this client is built around the callables in the form of
+ *       synchronized or asynchronized operation. These operation can also be achieved by directly
+ *       using callables.
```
igorbernstein2: Please provide an example

igorbernstein2: Actually looking at the docs, it seems like the callables info is already there:

```
*   <li>A "callable" method, like `readRowsCallable()`. This type of method takes no parameters and
 *       returns an immutable API callable object, which can be used to initiate calls to the	 *       returns an immutable API callable object, which can be used to initiate calls to the
 *       service.	 *       service.
```

I think all this needs is just an example:

Taking ReadRows as an example:

```java
// These two invocation are equivalent
ServerStream<Row> stream1 = client.readRows(query);
ServerStream<Row> stream2 = client.readRowsCallable().call(query);

// These two invocation are also equivalent
client.readRowsAsync(query, observer);
client.readRowsCallable().call(query, observer); 
```

PR #6432, opened 2019-10-03
[Bigtable] Overload MutationApi#setCell to accept long values

rahulKQL:
Fixes [Java-Bigtable#4](https://github.com/googleapis/java-bigtable/issues/4)

Added 4 overloaded methods in the `MutationApi` & added test case for the same.

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
+++ b/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
@@ -170,9 +170,8 @@ public final class Mutation implements MutationApi<Mutation>, Serializable {
 
   @Override
   public Mutation setCell(@Nonnull String familyName, @Nonnull ByteString qualifier, long value) {
-    long timestamp = System.currentTimeMillis() * 1_000;
 
-    return setCell(familyName, qualifier, timestamp, value);
+    return setCell(familyName, qualifier, ByteString.copyFrom(Longs.toByteArray(value)));
   }
 
   @Override


google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -155,6 +156,32 @@ public Mutation setCell(
...
+    long timestamp = System.currentTimeMillis() * 1_000;
+
+    return setCell(familyName, qualifier, timestamp, value);
+  }
+
+  @Override
+  public Mutation setCell(
+      @Nonnull String familyName, @Nonnull ByteString qualifier, long timestamp, long value) {
+
+    return setCell(familyName, qualifier, timestamp, ByteString.copyFrom(Longs.toByteArray(value)));
```
rahulKQL: @igorbernstein2 I have relied upon `Longs.toByteArray(long value)` for long value conversion instead of ByteBuffer. As ByteBuffer needs to be `rewind()` before passing it to `ByteString.copyFrom(ByteBuffer byteBuffer)`.

I hope this is acceptable.

igorbernstein2: Yep, thank you

google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -155,6 +156,32 @@ public Mutation setCell(
...
+  @Override
+  public Mutation setCell(
+      @Nonnull String familyName, @Nonnull String qualifier, long timestamp, long value) {
+
+    return setCell(familyName, wrapByteString(qualifier), timestamp, value);
+  }
+
+  @Override
+  public Mutation setCell(@Nonnull String familyName, @Nonnull ByteString qualifier, long value) {
+    long timestamp = System.currentTimeMillis() * 1_000;
```
igorbernstein2: I would prefer to duplicate the long -> byte string conversion then duplicating the timestamp logic.

rahulKQL: Thanks for the suggestion, I updated setCell with same.

PR #6451, opened 2019-10-07
don't depend on Apache commons

### LINKED ISSUE #6449: Remove dependency on org.apache.commons.io.FileUtils;
I think we use this in one file:

google-cloud-java/google-cloud-testing/google-cloud-gcloud-maven-plugin/src/main/java/com/google/cloud/DownloadComponentsMojo.java

and it should be replaceable with Guava

elharo:
Fixes #6449


chingor13:
@elharo We moved this plugin into its separate repo: https://github.com/googleapis/java-gcloud-maven-plugin. I'll transfer the corresponding issue, but unfortunately, I can't move this PR over there.

PR #6471, opened 2019-10-09
Removing outdated message about being 0.y.z

justinuang:
We are now on version 1.3.0 so the warning is untrue.


justinuang:
@googlebot I signed it!

justinuang:
https://github.com/googleapis/java-bigtable/pull/54

PR #6472, opened 2019-10-09
Refresh VPCSC tests for monitoring V3

steinwaywhw:
The codegen was updated recently, as well as the protos. This is to refresh the auto-generated tests for VPCSC for the monitoring v3 clients.


steinwaywhw:
Sure. I removed the V3 from the name. Since the file is auto-generated, the diff probably does not mean anything to you. We just need to make sure the tests pass.

PR #6484, opened 2019-10-10
Declare direct dependencies in google-cloud-bigtable pom.xml

justinuang:
Some notes:

- auto-value-annotations should be a compile time dependency: https://github.com/google/auto/blob/master/value/userguide/index.md#in-pomxml

- `BaseBigtableInstanceAdminSettings` uses iam protos

- Remove dependency on com.google.http-client:google-http-client because of a wrong import of `Lists`

- We have to specify the version of threetenbp = 1.3.3 to match gax, as enforced by the dependency convergence rule.

- jsr305 is used to provide javax.annotations.Nonnull, etc. These are added to java in java 8, but since we need to be compatible with java 7, we have an explicit import.

@igorbernstein2

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/pom.xml
+++ b/google-cloud-clients/google-cloud-bigtable/pom.xml
@@ -18,6 +18,9 @@
     <site.installationModule>google-cloud-bigtable</site.installationModule>
   </properties>
   <dependencies>
+    <!-- NOTE: Dependencies are organized into two groups, production and test.
+         Within a group, dependencies are sorted by (groupId, artifactId) -->
+
     <!-- Production dependencies -->
     <dependency>
       <groupId>${project.groupId}</groupId>


google-cloud-clients/google-cloud-bigtable/pom.xml
```diff
@@ -39,7 +39,92 @@
...
+      <artifactId>grpc-stub</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.auto.value</groupId>
+      <artifactId>auto-value-annotations</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
```
igorbernstein2: The ordering is a bit arbitrary, would you mind
- grouping like dependencies together? 
  - keep the 2 auto-values next to each other
  - introduce some vertical spacing between different groups of deps
- ordering the deps in some logical order? Consider something like the following as one possible ordering. I don't really care what you choose, as long as it's easy to keep track of the deps
```
<!-- Bigtable service specific -->
google-cloud-core-grpc
proto-google-cloud-bigtable-v2
proto-google-cloud-bigtable-admin-v2

<!-- Google client shared deps -->
proto-google-common-protos
proto-google-iam-v1

google-cloud-core
api-common
gax-grpc

grpc
protobuf

<!-- General libraries -->
auto-value
guava

<!-- Test -->
junit
```

google-cloud-clients/google-cloud-bigtable/pom.xml
```diff
@@ -39,7 +39,92 @@
...
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-protobuf</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-stub</artifactId>
```
igorbernstein2: I think grpc-stub can be scoped to test

google-cloud-clients/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -31,6 +30,7 @@
 import com.google.cloud.bigtable.data.v2.models.Query;
 import com.google.cloud.bigtable.data.v2.models.Range.ByteStringRange;
 import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.common.collect.Lists;
```
igorbernstein2: oh the joys of autocomplete, sorry about that :(

google-cloud-clients/google-cloud-bigtable/pom.xml
igorbernstein2: Also, I think you might be missing some test deps. Can you check dependency:analyze for the test scope? I know that we use grpc-testing, but I don't see it here

google-cloud-clients/google-cloud-bigtable/pom.xml
```diff
@@ -39,7 +39,92 @@
...
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-context</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-grpclb</artifactId>
```
igorbernstein2: I think this can be a runtime dep as well

google-cloud-clients/google-cloud-bigtable/pom.xml
justinuang: Reordered into two groups, production deps and test deps. Within these groups, things are alphabetized to keep similar deps together.

google-cloud-clients/google-cloud-bigtable/pom.xml
```diff
@@ -18,10 +18,23 @@
     <site.installationModule>google-cloud-bigtable</site.installationModule>
   </properties>
   <dependencies>
+    <!-- Production dependencies -->
```
igorbernstein2: Please mention that these are alphabetical, so that future devs know what pattern to follow.

google-cloud-clients/google-cloud-bigtable/pom.xml
```diff
@@ -30,16 +43,87 @@
...
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-protobuf</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.threeten</groupId>
+      <artifactId>threetenbp</artifactId>
+      <!-- For dependency convergence with gax -->
+      <version>1.3.3</version>
```
igorbernstein2: Please add a todo to pull this up to the parent pom

google-cloud-clients/google-cloud-bigtable/pom.xml
justinuang: Done

google-cloud-clients/google-cloud-bigtable/pom.xml
justinuang: We can discuss here: https://github.com/googleapis/google-cloud-java/issues/6496

kolea2:
Thanks for looking into this!

google-cloud-clients/google-cloud-bigtable/pom.xml
elharo: As long as alphabetical works, that's fine. This does affect which versions of transitive dependencies maven selects though. That is, the ordering is significant and changes in order can change the classpath and fix or break the build.

PR #6523, opened 2019-10-17
Firestore: Make Internal class constructable

schmidt-sebastian:
We need to change the constructor of `Internal` to take a Firestore class, since FirestoreImpl is a private type. This allows the Google Cloud Frameworks team to obtain an instance of the class.

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Internal.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Internal.java
@@ -30,8 +30,10 @@ public class Internal {
   private FirestoreImpl firestore;
 
   @InternalApi
-  public Internal(FirestoreImpl firestore) {
-    this.firestore = firestore;
+  public Internal(Firestore firestore) {
+    Preconditions.checkArgument(
+        firestore instanceof FirestoreImpl, "Argument is required to be of type FirestoreImpl");
+    this.firestore = (FirestoreImpl) firestore;
   }
 
   @InternalApi


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Internal.java
```diff
@@ -30,8 +30,10 @@
   private FirestoreImpl firestore;
 
   @InternalApi
-  public Internal(FirestoreImpl firestore) {
-    this.firestore = firestore;
+  public Internal(Firestore firestore) {
```
elharo: This constructor is still marked `@InternalApi` so it shouldn't b used by other code. Is this change realy needed?

schmidt-sebastian: I have a meeting with the team that requested this on Monday, after which I can follow up. The team is Internal, so they are "allowed" to use these APIs.

elharo: What's weird is that you're doing this to avoid exposing an internal class in the API, but the API is internal so why not just leave it as is?

schmidt-sebastian: FirestoreImpl is not constructible outside of our package.

schmidt-sebastian: Dropping this PR for now. Cloud Frameworks team will present their own solution.

PR #6530, opened 2019-10-18
Add list buckets sample

frankyn:
Short sample to list buckets in Java without using a prefix search.

### DIFF
--- a/google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
+++ b/google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
@@ -440,16 +440,6 @@ public class ITStorageSnippets {
   }
 
   @Test
-  public void testAuthListBuckets() {
-    ByteArrayOutputStream snippetOutputCapture = new ByteArrayOutputStream();
-    System.setOut(new PrintStream(snippetOutputCapture));
-    storageSnippets.authListBuckets();
-    String snippetOutput = snippetOutputCapture.toString();
-    System.setOut(System.out);
-    assertTrue(snippetOutput.contains("Buckets:"));
-  }
-
-  @Test
   public void testListBuckets() {
     ByteArrayOutputStream snippetOutputCapture = new ByteArrayOutputStream();
     System.setOut(new PrintStream(snippetOutputCapture));


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1003,6 +1003,19 @@ public Acl blobToPublicRead(String bucketName, String blobName, long blobGenerat
...
   }
 
+  /** Example of a simple listBuckets() */
+  public Page<Bucket> listBuckets() {
+    // [START storage_list_buckets]
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+
+    Page<Bucket> buckets = storage.list();
+    for (Bucket bucket : buckets.iterateAll()) {
+      // do something with the info
```
elharo: maybe we could print the path to the bucket or some such thing?

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -1003,6 +1003,19 @@ public Acl blobToPublicRead(String bucketName, String blobName, long blobGenerat
...
+    }
+    // [END storage_list_buckets]
+    return buckets;
```
elharo: Or why do we return and iterate the buckets? Feels like we should either use the value or return it, not both.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -988,19 +988,32 @@ public Acl blobToPublicRead(String bucketName, String blobName, long blobGenerat
   }
 
   /** Example of default auth */
-  public Page<Bucket> authListBuckets() {
+  public void authListBuckets() {
     // [START auth_cloud_implicit]
     // If you don't specify credentials when constructing the client, the
     // client library will look for credentials in the environment.
```
elharo: will look --> looks (tech writing lives in the eternal present)

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
```diff
@@ -988,19 +988,32 @@ public Acl blobToPublicRead(String bucketName, String blobName, long blobGenerat
...
+    System.out.println("Buckets:");
     for (Bucket bucket : buckets.iterateAll()) {
-      // do something with the info
+      System.out.println(bucket.getName());
     }
     // [END auth_cloud_implicit]
-    return buckets;
+  }
+
+  /** Example of a simple listBuckets() */
```
elharo: Am I missing something? These two methods look exactly the same aside from the name.

google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
```diff
@@ -441,8 +441,22 @@ public void testBlobAcl() {
 
   @Test
   public void testAuthListBuckets() {
-    Page<Bucket> bucket = storageSnippets.authListBuckets();
-    assertNotNull(bucket);
+    final ByteArrayOutputStream snippetOutputCapture = new ByteArrayOutputStream();
```
elharo: Is there a reason to make this final?

google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
```diff
@@ -441,8 +441,22 @@ public void testBlobAcl() {
...
+    System.setOut(new PrintStream(snippetOutputCapture));
+    storageSnippets.authListBuckets();
+    String snippetOutput = snippetOutputCapture.toString();
+    System.setOut(System.out);
+    assertTrue(snippetOutput.contains("Buckets:"));
+  }
+
+  @Test
+  public void testListBuckets() {
+    final ByteArrayOutputStream snippetOutputCapture = new ByteArrayOutputStream();
```
elharo: no final

google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets/ITStorageSnippets.java
frankyn: following another test. Otherwise no.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java
frankyn: authListBuckets() is used on https://cloud.google.com/docs/authentication/getting-started#auth-cloud-implicit-java with specific comments for that page. However it is duped on another sample in java-docs-samples for a related page: https://cloud.google.com/docs/authentication/production. Removing sample from this doc and updating documentation to point to the other sample.

listBuckets() will be used on https://cloud.google.com/storage/docs/listing-buckets

frankyn: Documentation was updated to not depend on the authListBuckets sample.

PR #6541, opened 2019-10-22
chore: stop release of texttospeech

codyoss:


codyoss:
Related to googleapis/java-texttospeech#1

codyoss:
Lgtm, is this spanner thing a flake? I have seen it on a previous PR as well... If so we should create an issue and get that fixed sometime

PR #6544, opened 2019-10-23
Create system test case for text to speech API.

wangbill-google:
Upload system test case for text to speech API(both v1 and v1beta1).


chingor13:
Thanks for this! Can you actually send this PR to https://github.com/googleapis/java-texttospeech?

We're in the process of splitting this repo to make releases faster and independent.

PR #6551, opened 2019-10-23
Logging : Add Label fullStacktrace

### LINKED ISSUE #4835: Stacktrace should no be written to log message
Hello.

I've noticed pretty strange things. Somehow you are adding whole stack trace information to textPayload which is weird. Also  it look very ugly in stackdriver logging web UI because you just put there extremely long string that can not be displayed in nice and verbose way.

I think that nobody wants to see it in textPayload. Why you can't add full stacktrace to property/label named "stacktrace" ? 

As for textPayload the only one reasonable modification is to add only message from exception at new line

This is part of the code that does all this things

```
  private LogEntry logEntryFor(ILoggingEvent e) {
    StringBuilder payload = new StringBuilder(e.getFormattedMessage()).append('\n');
    writeStack(e.getThrowableProxy(), "", payload);

    Level level = e.getLevel();
    LogEntry.Builder builder =
        LogEntry.newBuilder(Payload.StringPayload.of(payload.toString().trim()))
            .setTimestamp(e.getTimeStamp())
.setSeverity(severityFor(level));
```

athakor:
Fixes #4835


athakor:
@chingor13 PTAL

athakor:
Already raised same [PR](https://github.com/googleapis/java-logging/pull/40) in [java-logging](https://github.com/googleapis/java-logging) repo.

PR #6552, opened 2019-10-23
Bigquery: add range partitioning to tables, load jobs, and query jobs.

pmakani:
**PR Contains :**

`RangePartitioning` classes and properties add support for the integer range
 partitioning feature. These offer more flexibility in partitioning options than 
 time-based partitioning.


google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/RangePartitioning.java
```diff
@@ -0,0 +1,241 @@
...
+    this.range = builder.range;
+  }
+
+  private RangePartitioning(
+      com.google.api.services.bigquery.model.RangePartitioning rangePartitioning) {
+    this.field = rangePartitioning.getField();
+    this.range = Range.fromPb(rangePartitioning.getRange());
+  }
+
+  /** Returns the field of range partitioning. */
```
shollyman: /** Returns the range partitioning field. */

PR #6554, opened 2019-10-23
Import split-repo boms into google-cloud-bom

chingor13:
Rather that specifying all the artifacts from the split repos, we will import the bom artifacts that those repos provide:

* asset
* automl
* core
* datacatalog
* datalabeling
* dlp
* gameservices
* iamcredentials
* iot
* phishingprotection
* speech
* talent
* tasks
* texttospeech
* trace
* translate
* vision
* websecurityscanner

### DIFF
--- a/google-cloud-bom/pom.xml
+++ b/google-cloud-bom/pom.xml
@@ -203,7 +203,7 @@
       </dependency>
       <dependency>
         <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-automl</artifactId>
+        <artifactId>google-cloud-automl-bom</artifactId>
         <version>0.114.1-beta</version>
         <type>pom</type>
         <scope>import</scope>


anuraaga:
Hello - as we found in https://github.com/gradle/gradle/issues/11208 this PR changed the behavior of `google-cloud-bom` significantly. It used to be the one BOM to manage them all but now if a user wanted to use, e.g., trace, they have to add an unversioned dependency on the trace-bom too. Just want to confirm, is this the expected new behavior? I can't find any issue to provide more reasoning on the reason to use imported BOMs - if it's to reduce copy-paste, it's probably better for `releasetool` to inline BOMs instead of importing them to make it easier for users to use the super-bom.

elharo:
I'm looking into this. I am not certain that Maven behavior is as gradle/gradle#11208 thinks it is.

elharo:
Update: the comments on the gradle bug are incorrect.  For Maven there is no behavior change from 0.116.0 to 0.117.0. Gradle I'm not as familiar with.

jjohannes:
This is working in Gradle as well. The confusion was created because `grpc-google-cloud-trace-v2` is not listed in [google-cloud-trace-bom-0.109.0-beta.pom](https://repo1.maven.org/maven2/com/google/cloud/google-cloud-trace-bom/0.109.0-beta/google-cloud-trace-bom-0.109.0-beta.pom). And because the individual entry was removed in this PR, there is no entry for `grpc-google-cloud-trace-v2` anymore (only for `grpc-google-cloud-trace-v1`).

anuraaga:
Thanks a lot for investigating this - sorry for not confirming the contents of `google-cloud-trace-bom` first. Looks like it will be a relatively simple fix to just add to that bom :)

PR #6575, opened 2019-10-25
chore: stop release of scheduler

codyoss:


codyoss:
Being that no code was actually touched here, I think the cov is fine. Really interesting the build thinks this is a regression...

versions.txt
```diff
@@ -110,8 +110,6 @@ proto-google-cloud-pubsub-v1:1.80.0:1.80.1-SNAPSHOT
 proto-google-cloud-recaptchaenterprise-v1beta1:0.27.0:0.27.1-SNAPSHOT
 proto-google-cloud-redis-v1:0.81.0:0.81.1-SNAPSHOT
 proto-google-cloud-redis-v1beta1:0.81.0:0.81.1-SNAPSHOT
-proto-google-cloud-scheduler-v1:1.21.0:1.21.1-SNAPSHOT
```
chingor13: same as the other PR, there should be grpc- and google-cloud-scheduler versions in here too

PR #6641, opened 2019-10-29
Remove usage of shaded netty in imports in favor of none shaded one

mbril:

### DIFF
--- a/google-cloud-clients/google-cloud-bigtable/pom.xml
+++ b/google-cloud-clients/google-cloud-bigtable/pom.xml
@@ -115,7 +115,7 @@
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
-      <artifactId>grpc-netty-shaded</artifactId>
+      <artifactId>grpc-netty</artifactId>
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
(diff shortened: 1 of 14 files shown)


chingor13:
@garrettjonesgoogle I'm not sure the history on why we use grpc-netty-shaded, but I presume it's because of diamond dependencies issues with netty major versions?

garrettjonesgoogle:
Netty doesn't follow semantic versioning; they break compatibility between minor versions. Netty was the cause of some of our worse early diamond dependency conflicts. I think it is unwise to switch off of grpc-netty-shaded. cc @ejona86

mbril:
Abandoning the PR, due to issues with non shaded netty in the past

ejona86:
Yes, grpc-java recommends using grpc-netty-shaded for most users, as is visible [in our public documentation](https://github.com/grpc/grpc-java/blob/master/README.md). gRPC uses unstable APIs in Netty and even if that weren't the case Netty itself does not follow semantic versioning (4.1 was allowed to break interfaces in 4.0). But also depending on the correct netty-tcnative version can be a bit hard for users, and grpc-netty-shaded includes the correct netty-tcnative version.

PR #6660, opened 2019-10-30
chore: stop release of logging

codyoss:


google-cloud-clients/google-cloud-logging/pom.xml
```diff
@@ -77,7 +77,27 @@
       </plugin>
     </plugins>
   </build>
-
+  <build>
```
chingor13: The build is failing with 2 build tags in here

PR #6685, opened 2019-11-04
Storage :  Add support for Archive storage class

### LINKED ISSUE #6578: Storage: Add support for Archive storage class
GCS is adding a storage class called "archive" to the existing options for object storage (coldline, standard, etc). Support for this should be added to this client library. This has already been implemented in the C++ client library and there is a PR in python; reference PR [here](https://github.com/googleapis/google-cloud-python/issues/9532)

This should be straightforward as ARCHIVE should just mirror existing valid storage classes. I see the following places where changes are required in order to accomplish this:

* https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java (add ARCHIVE constant).
* Consider adding a case with ARCHIVE to one of the unit tests (BucketInfoTest.java or BlobInfoTest.java).

There are also snippets that reference storage classes in google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java but I don't __think__ we need to add references to ARCHIVE here given that there isn't a specific snippet for each possible value.

@kolea2 or @JesseLovelace could you pick this ticket up?

athakor:
Fixes #6578

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
@@ -42,44 +42,40 @@ public final class StorageClass extends StringEnumValue {
       new StringEnumType(StorageClass.class, CONSTRUCTOR);
 
   /**
-   * Regional storage class (deprecated).
-   * See:https://cloud.google.com/storage/docs/per-object-storage-class for details
+   * Standard storage class. See: https://cloud.google.com/storage/docs/storage-classes for details
+   */
+  public static final StorageClass STANDARD = type.createAndRegister("STANDARD");
+
+  /**
+   * Nearline storage class. See: https://cloud.google.com/storage/docs/storage-classes for details
+   */
+  public static final StorageClass NEARLINE = type.createAndRegister("NEARLINE");
+
+  /**
+   * Coldline storage class. See: https://cloud.google.com/storage/docs/storage-classes for details
+   */
+  public static final StorageClass COLDLINE = type.createAndRegister("COLDLINE");
+
+  /**
+   * Archive storage class. See: https://cloud.google.com/storage/docs/storage-classes for details
+   */
+  public static final StorageClass ARCHIVE = type.createAndRegister("ARCHIVE");
+
+  /**
+   * Regional storage class (deprecated). See: https://cloud.google.com/storage/docs/storage-classes
+   * for details
    */
   public static final StorageClass REGIONAL = type.createAndRegister("REGIONAL");
 
   /**
    * Multi-regional storage class (deprecated).
-   * See:https://cloud.google.com/storage/docs/per-object-storage-class for details
+   * See:https://cloud.google.com/storage/docs/storage-classes for details
    */
   public static final StorageClass MULTI_REGIONAL = type.createAndRegister("MULTI_REGIONAL");
 
   /**
-   * Nearline storage class. See: https://cloud.google.com/storage/docs/per-object-storage-class for
-   * details
-   */
-  public static final StorageClass NEARLINE = type.createAndRegister("NEARLINE");
-
-  /**
-   * Coldline storage class. See: https://cloud.google.com/storage/docs/per-object-storage-class for
-   * details
-   */
-  public static final StorageClass COLDLINE = type.createAndRegister("COLDLINE");
-
-  /**
-   * Standard storage class. See: https://cloud.google.com/storage/docs/per-object-storage-class for
-   * details
-   */
-  public static final StorageClass STANDARD = type.createAndRegister("STANDARD");
-
-  /**
-   * Archive storage class. See: https://cloud.google.com/storage/docs/per-object-storage-class for
-   * details
-   */
-  public static final StorageClass ARCHIVE = type.createAndRegister("ARCHIVE");
-
-  /**
    * Durable Reduced Availability (deprecated).
-   * See:https://cloud.google.com/storage/docs/per-object-storage-class for details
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -56,6 +56,9 @@ public StorageClass apply(String constant) {
   /** Standard storage class. */
   public static final StorageClass STANDARD = type.createAndRegister("STANDARD");
 
+  /** Archive storage class. */
```
tritone: Regional and Multi-regional are also considered deprecated now. Let's add that to the comment for each of those and also add a link to https://cloud.google.com/storage/docs/storage-classes which has the most up-to-date information.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -121,6 +122,36 @@
...
+          .setCreateTime(CREATE_TIME)
+          .setDefaultAcl(DEFAULT_ACL)
+          .setDeleteRules(DELETE_RULES)
+          .setLifecycleRules(LIFECYCLE_RULES)
+          .setIndexPage(INDEX_PAGE)
+          .setIamConfiguration(IAM_CONFIGURATION)
+          .setNotFoundPage(NOT_FOUND_PAGE)
+          .setLocation(LOCATION)
+          .setLocationType(LOCATION_TYPE)
+          .setStorageClass(ARCHIVE_STORAGE_CLASS)
```
tritone: Is it typical to set all of these options while creating a bucket in Java? Just curious.

athakor: No need to set all of these option but yes bucket name is required parameter see the [api doc](https://cloud.google.com/storage/docs/json_api/v1/buckets/insert) and example [create bucket](https://cloud.google.com/storage/docs/creating-buckets#storage-create-bucket-java)

athakor:
@tritone PLTAL

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
athakor: done

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -41,22 +41,46 @@ public StorageClass apply(String constant) {
   private static final StringEnumType<StorageClass> type =
       new StringEnumType(StorageClass.class, CONSTRUCTOR);
 
-  /** Regional storage class. */
+  /**
+   * Regional storage class (deprecated).
+   * See:https://cloud.google.com/storage/docs/per-object-storage-class for details
+   */
   public static final StorageClass REGIONAL = type.createAndRegister("REGIONAL");
```
tritone: Let's re-sort these as follows so that the non-deprecated classes come first: Standard, Nearline, Coldline, Archive, Regional, Multi-Regional, Durable Reduced Availability

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -41,22 +41,46 @@ public StorageClass apply(String constant) {
...
 
-  /** Multi-regional storage class. */
+  /**
+   * Multi-regional storage class (deprecated).
+   * See:https://cloud.google.com/storage/docs/per-object-storage-class for details
```
tritone: This link redirects to https://cloud.google.com/storage/docs/storage-classes , so let's use that.

google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
tritone: Okay, sounds good.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
athakor: done

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
athakor: done

athakor:
@tritone all the comment have been addressed **PTAL**

tritone:
> @tritone all the comment have been addressed **PTAL**

Looks good to me! I will let you know when it's okay to merge (waiting until feature has been enabled without a flag).

tritone:
This PR should no longer be merged because of the repo split. @athakor , could you re-create this PR in https://github.com/googleapis/java-storage ? Please let me know if you run into any issues.

Also, apologies for the delay on this-- we are now looking to merge the first week of Jan.

athakor:
> This PR should no longer be merged because of the repo split. @athakor , could you re-create this PR in https://github.com/googleapis/java-storage ? Please let me know if you run into any issues.
> 
> Also, apologies for the delay on this-- we are now looking to merge the first week of Jan.

done

PR #6740, opened 2019-11-12
CloudStorage: Write the storage contents to the output stream instead of a byte array

### LINKED ISSUE #6739: CloudStorage: The FakeRpcStorage read() method doesn't write the contents to the given OutputStream

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/google-cloud-java/issues
  - Check for answers on StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

1. Specify the API at the beginning of the title (for example, "BigQuery: ...")
   General, Core, and Other are also allowed as types
2. OS type and version: macOS 10.14
3. Java version: 11.0.5
4. google-cloud-java version(s): google-cloud-nio-0.118.0-alpha

#### Steps to reproduce

  1. call Blob::downloadTo(Path path) when using the FakeStorageRpc
  2. it doesn't write anything to the given file

#### Code example

```groovy
// example
def storage = LocalStorageHelper.options.service
def blob = storage.get(bucket, key)
blob.downloadTo(outputPath)
```

#### Stack trace

n/a

n/a

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

clarkperkins:
Fixes #6739 (it's a good idea to open an issue first for context and/or discussion)

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java
@@ -327,8 +327,11 @@ class FakeStorageRpc implements StorageRpc {
       // special case: you're trying to read past the end
       return 0;
     }
-    byte[] ret = new byte[bytes];
-    System.arraycopy(full, (int) position, ret, 0, bytes);
+    try {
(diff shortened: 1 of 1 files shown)


JesseLovelace:
Merging--I'll follow up to add a test.

PR #6762, opened 2019-11-13
Spanner: Remove dependency on GCS

### LINKED ISSUE #6708: Spanner: remove Cloud Storage depedency
#### Environment details

1. Specify the API at the beginning of the title (for example, "BigQuery: ...")
   General, Core, and Other are also allowed as types
2. OS type and version: OSX
3. Java version: java 8
4. google-cloud-java version(s): google-cloud-spanner-jdbc:1.7.0

#### Steps to reproduce

  1. Add dependnecy to google-cloud-spanner-jdbc:1.7.0
  2. See the dependency tree, e.g. `mvn dependency:tree -Dverbose`

#### Code example

N/A

#### Stack trace
N/A

N/A	

Spanner JDBC driver pulls in cloud storage dependency. This seems excess for a JDBC driver.

According to @olavloite this is used for potentially fetching credentials from GCS.

Master branch currently has the dependency marked as `scope` `provided`. However, the semantic of `provided` scope doesn't seem to match the intended meaning.

1. Dependency should be marked as `optional` (`<optional>true</optional>`)
1. Consider removing this dependency (and functionality) altogether. It's not documented, nor is it a JDBC driver's responsibility to fetch credentials from a remote storage.

olavloite:
Remove the dependency on Google Cloud Storage and throw a descriptive exception for users who are currently using this feature.

Fixes #6708.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
@@ -30,7 +30,7 @@ import java.io.InputStream;
 /** Service class for getting credentials from key files. */
 class CredentialsService {
   static final String GCS_NOT_SUPPORTED_MSG =
-      "Credentials that are stored on Google Cloud Storage is no longer supported. Please use the Google Cloud Storage client library to download the credentials to a local file and reference the local file in the connection URL.";
+      "Credentials that is stored on Google Cloud Storage is no longer supported. Download the credentials to a local file and reference the local file in the connection URL.";
   static final CredentialsService INSTANCE = new CredentialsService();
 
   CredentialsService() {}


google-cloud-clients/google-cloud-contrib/google-cloud-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/CredentialsService.java
```diff
@@ -20,23 +20,17 @@
...
 
-/** Service class for getting credentials from key files stored locally or on cloud storage. */
+/** Service class for getting credentials from key files. */
 class CredentialsService {
-  private static final String GOOGLE_CLOUD_STORAGE_PREFIX = "gs://";
-  private static final String INVALID_GCS_PREFIX_MSG =
-      String.format("Storage URL must start with %s", GOOGLE_CLOUD_STORAGE_PREFIX);
-
+  static final String GCS_NOT_SUPPORTED_MSG =
+      "Credentials that are stored on Google Cloud Storage is no longer supported. Please use the Google Cloud Storage client library to download the credentials to a local file and reference the local file in the connection URL.";
```
elharo: is --> are
No "Please"

http://www.textdoctor.com/webinars/please/

Most important, I don't think it's the google cloud storage client library that will be used to download these credentials. Perhaps just "download the credentials to a local file and reference the local file in the connection URL."

olavloite: Fixed.

PR #6766, opened 2019-11-13
Update with latest protos with synthtool

mbril:


skuruppu:
@mbril, I'm going to favor #6757 since that's done by Autosynth. There's something that needs fixing with the auto-generation process but I'll take care of it.

PR #6789, opened 2019-11-14
Spanner: Session pool should not retry BeginTransaction if database has been deleted

### LINKED ISSUE #6779: spanner: Client library sending a high rate of BeginTransaction requests when database is deleted
Once a database is deleted and if the Java Spanner client is still left open, it starts sending a large number of BeginTransaction requests. The additional requests fail with status NOT_FOUND.

The following error just keeps repeating in the logs:
```
Nov 07, 2019 10:20:16 PM com.google.cloud.spanner.spi.v1.LoggingInterceptor$CallLogger log FINER: google.spanner.v1.Spanner/BeginTransaction[74f0a03d]: Start
Nov 07, 2019 10:20:16 PM com.google.cloud.spanner.spi.v1.LoggingInterceptor$CallLogger log FINER: google.spanner.v1.Spanner/BeginTransaction[74f0a03d]: Send: session: "projects/<project>/instances/<instance>/databases/<database>/sessions/<session>" options { read_write { } }
Nov 07, 2019 10:20:16 PM com.google.cloud.spanner.spi.v1.LoggingInterceptor$CallLogger log FINER: google.spanner.v1.Spanner/BeginTransaction[74f0a03d]: Closed with status Status{code=NOT_FOUND, description=Database not found: projects/<project>/instances/<instance>/databases/<database> resource_type: "type.googleapis.com/google.spanner.admin.database.v1.Database" resource_name: "projects/<project>/instances/<instance>/databases/<database>" description: "Database does not exist." , cause=null} and trailers Metadata(...)
```

I'm guessing this is happening in an attempt to replenish the pool.

olavloite:
The session pool could start an infinite loop of prepare session calls (BeginTransaction) if the underlying database has been deleted. Instead, the session pool should detect this situation and not retry the call.

Fixes #6779

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -1196,6 +1196,10 @@ final class SessionPool {
     return e.getErrorCode() == ErrorCode.NOT_FOUND && e.getMessage().contains("Database not found");
   }
 
+  private boolean isPermissionDenied(SpannerException e) {
+    return e.getErrorCode() == ErrorCode.PERMISSION_DENIED;
+  }
+
   private void invalidateSession(PooledSession session) {
     synchronized (lock) {
       if (isClosed()) {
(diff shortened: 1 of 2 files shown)


google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
```diff
@@ -1185,6 +1192,11 @@ private boolean isSessionNotFound(SpannerException e) {
     return e.getErrorCode() == ErrorCode.NOT_FOUND && e.getMessage().contains("Session not found");
   }
 
+  private boolean isDatabaseDoesNotExist(SpannerException e) {
```
skuruppu: Can the logic be flipped and we call this isDatabaseExists() or databaseExists()? The suggestion is mostly for making sense of this function name.

skuruppu: Or maybe isDatabaseNotFound() to match isSessionNotFound().

olavloite: Changed to `isDatabaseNotFound()` to keep in sync with `isSessionNotFound()`. Also changed the string we are checking for to "Database not found", as that is also the string we are checking for in `isSessionNotFound()` (and the text is included in the error message).

olavloite:
@skuruppu I also added a check for PermissionDenied to the session pool, as that error would cause the exact same problem as NotFound (and the same as in the Go client).

PR #6796, opened 2019-11-15
Update README.md to 0.116.0

elharo:
@athakor


athakor:
LGTM

elharo:
I think you need to select the Approve radio button when you send a review. marking it LGTM doesn't work here like it does in google3. :-(

PR #6798, opened 2019-11-15
add libraries-bom

elharo:
@chingor13 @athakor 

Toward https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1031


versions.txt
```diff
@@ -38,6 +38,7 @@ grpc-google-cloud-pubsub-v1:1.83.0:1.83.0
 grpc-google-cloud-spanner-admin-database-v1:1.46.0:1.46.0
 grpc-google-cloud-spanner-admin-instance-v1:1.46.0:1.46.0
 grpc-google-cloud-spanner-v1:1.46.0:1.46.0
+com.google.cloud:libraries-bom:2.9.0
```
chingor13: If you add this here, our release tooling will bump this everytime this repo is released, regardless of if there's a new libraries-bom version.

elharo: OK, we don't want that. How should we configure it? Or will this work at all?

PR #6800, opened 2019-11-15
Fix naming of testIamPermissions in Bigtable

alexoneill:
The method equivalent for `TestIamPermissions` in the Instance and Table Admin Client is called
`testIamPermission`, when it should be `testIamPermission`**s**.

This will bring this client in-line with the naming schemes of other clients and be more consistent with the method name in the underlying API. The old name is deprecated (still works), the new name is preferred.

This additionally fixes an issue in the documentation for these methods where some code was not included in the neighboring code snippet.


igorbernstein2:
Can you re-open this PR in github.com/googleapis/java-bigtable?

chingor13:
The bigtable client code has moved to https://github.com/googleapis/java-bigtable. Please reopen this in that repository.

PR #6838, opened 2019-11-20
Add toString methods for object lifecycle classes

JesseLovelace:
Adds toString methods for object lifecycle classes. Strings will take the form "Action when condition". This was discovered while updating Java samples

For example "Delete when object is 10 days old and storage class is coldline", "Set storage class to COLDLINE when object is not live"


JesseLovelace:
Dropping this for now since it's inconsistent with other toString methods in this library

JesseLovelace:
Replaced by https://github.com/googleapis/google-cloud-java/pull/6856

PR #6842, opened 2019-11-20
Datastore: expose serializedSize for Entity.

### LINKED ISSUE #6389: datastore: determine entity size using api
The low level Datastore API of App Engine provides a mechanism for determining the protobuf size of an entity by using the EntityTranslator class:

```kotlin
val protoEntity = EntityTranslator.convertToPb(entity)
if (protoEntity.serializedSize >= 1048572) {
  // ....
}
```

It would be good and cloud java client library to provide such API as it will let the client code determine the size of an entity before trying to persist it.

suraj-qlogic:
Fixes #6389

### DIFF
--- a/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Entity.java
+++ b/google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Entity.java
@@ -91,9 +91,9 @@ public final class Entity extends FullEntity<Key> {
   }
 
   /**
-   * Returns the serialized protobuf size of provided entity object.
+   * Returns the size in bytes of the protobuf form of the provided entity.
    *
-   * @param entity entity object to calculate serialize size.
+   * @param entity object whose size is measured.
    */
   public static int calculateSerializedSize(BaseEntity<? extends IncompleteKey> entity) {
     checkNotNull(entity);


google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java
```diff
@@ -636,6 +640,11 @@ public Blob getBlob(String name) {
     return properties;
   }
 
+  /** Returns the entity size. */
```
elharo: "Returns the number of bytes in the serialized protobuf."

Is this definitely a fixed and reproducible value for a given protobuf?

google-cloud-clients/google-cloud-datastore/src/test/java/com/google/cloud/datastore/EntityTest.java
```diff
@@ -55,4 +55,9 @@ public void testCopyFromIncompleteEntity() throws Exception {
     assertNotEquals(INCOMPLETE_ENTITY, entity);
     assertEquals(INCOMPLETE_ENTITY.getProperties(), entity.getProperties());
   }
+
+  @Test
+  public void testEntitySize() throws Exception {
```
elharo: Does this really throw exception?

google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java
```diff
@@ -446,18 +447,21 @@ public B setNull(String name) {
   BaseEntity(Builder<K, ?> builder) {
     this.key = builder.key;
     this.properties = ImmutableSortedMap.copyOf(builder.properties);
+    this.serializedSize = this.toPb().getSerializedSize();
```
elharo: what's the cost of calculating this for a large entity? Most of the time we don't need this value. Perhaps this could be done in demand and not stored or only lazily stored in the field?

suraj-qlogic: The cost could be varied it depends on user entered properties. Does this look fine? 
```java
/** Returns the number of bytes in the serialized protobuf. */
  public int getSerializedSize() {
    if (serializedSize == 0) {
      serializedSize = this.toPb().getSerializedSize();
    }
    return serializedSize;
  }
```
Here we are calculating entity size when demand.

google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/BaseEntity.java
suraj-qlogic: Yes,Once build it would be fixed number.

google-cloud-clients/google-cloud-datastore/src/test/java/com/google/cloud/datastore/EntityTest.java
suraj-qlogic: No,This method does not throw exception.

google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Entity.java
```diff
@@ -89,4 +89,9 @@ public static Builder newBuilder(Key key, FullEntity<?> copyFrom) {
   public static Entity fromPb(com.google.datastore.v1.Entity entityPb) {
     return new Builder().fill(entityPb).build();
   }
+
+  public static int calculateSerializedSize(BaseEntity<? extends IncompleteKey> e) {
```
BenWhitehead: Please add a javadoc comment to explain the usage of this method (and to help people find it when they google for this type of thing).

google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Entity.java
```diff
@@ -90,8 +90,13 @@ public static Entity fromPb(com.google.datastore.v1.Entity entityPb) {
     return new Builder().fill(entityPb).build();
   }
 
-  public static int calculateSerializedSize(BaseEntity<? extends IncompleteKey> e) {
-    checkNotNull(e);
-    return e.toPb().getSerializedSize();
+  /**
+   * Returns the serialized protobuf size of provided entity object.
+   *
+   * @param entity entity object to calculate serialize size.
```
elharo: "object whose size is measured"

google-cloud-clients/google-cloud-datastore/src/main/java/com/google/cloud/datastore/Entity.java
```diff
@@ -90,8 +90,13 @@ public static Entity fromPb(com.google.datastore.v1.Entity entityPb) {
     return new Builder().fill(entityPb).build();
   }
 
-  public static int calculateSerializedSize(BaseEntity<? extends IncompleteKey> e) {
-    checkNotNull(e);
-    return e.toPb().getSerializedSize();
+  /**
+   * Returns the serialized protobuf size of provided entity object.
```
elharo: "Returns the size in bytes of the protobuf form of the provided entity."

PR #6843, opened 2019-11-20
Firestore: Allow passing POJOs as field values throughout API reference from Firebase

### LINKED ISSUE #6711: Firestore: support updating an individual field with pojo as a value
Current Firestore (1.30.0+) does not support updating an individual field with pojo as a value, thus the following code

```
data class Test(
    val primitiveField: String,
    val complexField: Test2
)

data class Test2(
    val primitiveField1: String,
    val primitiveField2: Int
)

val firestore = FirestoreOptions.getDefaultInstance().service

fun main() {
    val test = Test("test", Test2("test", 123))
    val docRef = firestore.collection("test").document("test")
    docRef.set(test).get()
    docRef.update("complexField", Test2("updated test", 456)).get()
}
```
yields to the following exception:

```
Exception in thread "main" com.google.cloud.firestore.FirestoreException: Cannot convert Test2(primitiveField1=updated test, primitiveField2=456) to Firestore Value
	at com.google.cloud.firestore.FirestoreException.invalidState(FirestoreException.java:51)
	at com.google.cloud.firestore.UserDataConverter.encodeValue(UserDataConverter.java:180)
	at com.google.cloud.firestore.DocumentSnapshot.fromObject(DocumentSnapshot.java:89)
	at com.google.cloud.firestore.UpdateBuilder.performUpdate(UpdateBuilder.java:513)
	at com.google.cloud.firestore.UpdateBuilder.performUpdate(UpdateBuilder.java:502)
	at com.google.cloud.firestore.UpdateBuilder.update(UpdateBuilder.java:389)
	at com.google.cloud.firestore.DocumentReference.update(DocumentReference.java:261)
```

This is because _UserDataConverter.encodeValue_ only accepts either primitive types or list or maps of primitive types (which get encoded recursively). However, I do not see any reason why pojos could not be accepted as valid field values. Further, _CustomClassMapper.convertToPlainJavaTypes_ function used to convert pojos to maps could potentially be (re)used. In other words, the following workaround works (as long as you access the CustomClassMapper class through reflection as it is package-private) as expected:

```
docRef.update(
    "complexField",
    CustomClassMapper.convertToPlainJavaTypes(Test2("updated test", 456))
).get()
```
**Related issues:**
- https://github.com/googleapis/google-cloud-java/issues/5622
- https://github.com/googleapis/google-cloud-java/issues/6039

pmakani:
Fixes #6711 

**Reference**
- firebase/firebase-android-sdk#76

### DIFF
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
@@ -170,7 +170,7 @@ public class CollectionReference extends Query {
    * Adds a new document to this collection with the specified data, assigning it a document ID
    * automatically.
    *
-   * @param fields A Map or a POJO containing the data for the new document.
+   * @param fields The Map or POJO containing the data for the new document.
    * @return An ApiFuture that will be resolved with the DocumentReference of the newly created
    *     document.
    * @see #document()
--- a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
+++ b/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
@@ -26,6 +26,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.firestore.v1.ListCollectionIdsRequest;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.Executor;
 import javax.annotation.Nonnull;
@@ -138,7 +139,7 @@ public class DocumentReference {
    * Creates a new Document at the DocumentReference location. It fails the write if the document
    * exists.
    *
-   * @param pojo A map of the fields and values for the document.
+   * @param pojo The Map or POJO that will be used to populate the document contents.
    * @return An ApiFuture that will be resolved when the write finishes.
    */
   @Nonnull
@@ -166,7 +167,7 @@ public class DocumentReference {
    * exist, it will be created. If you pass {@link SetOptions}, the provided data can be merged into
    * an existing document.
    *
-   * @param fields The fields to write to the document (e.g. a Map or a POJO containing the desired
+   * @param fields The fields to write on the document (e.g. a Map or a POJO containing the desired
    *     document contents).
    * @param options An object to configure the set behavior.
    * @return An ApiFuture that will be resolved when the write finishes.
@@ -185,7 +186,7 @@ public class DocumentReference {
    * @return An ApiFuture that will be resolved when the write finishes.
    */
   @Nonnull
-  public ApiFuture<WriteResult> update(@Nonnull Object fields) {
+  public ApiFuture<WriteResult> update(@Nonnull Map<String, Object> fields) {
     WriteBatch writeBatch = firestore.batch();
     return extractFirst(writeBatch.update(this, fields).commit());
   }
@@ -199,7 +200,7 @@ public class DocumentReference {
    * @return An ApiFuture that will be resolved when the write finishes.
    */
   @Nonnull
-  public ApiFuture<WriteResult> update(@Nonnull Object fields, Precondition options) {
+  public ApiFuture<WriteResult> update(@Nonnull Map<String, Object> fields, Precondition options) {
     WriteBatch writeBatch = firestore.batch();
     return extractFirst(writeBatch.update(this, fields, options).commit());
   }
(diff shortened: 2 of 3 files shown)


google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
```diff
@@ -171,13 +170,13 @@ public void remove() {
    * Adds a new document to this collection with the specified data, assigning it a document ID
    * automatically.
    *
-   * @param fields A Map containing the data for the new document.
+   * @param fields A Map or a POJO containing the data for the new document.
```
elharo: nit: "a map or a POJO" and no period at end of comment since it's not a complete sentence

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
```diff
@@ -171,13 +170,13 @@ public void remove() {
...
    * @return An ApiFuture that will be resolved with the DocumentReference of the newly created
    *     document.
    * @see #document()
    */
   @Nonnull
-  public ApiFuture<DocumentReference> add(@Nonnull final Map<String, Object> fields) {
+  public ApiFuture<DocumentReference> add(@Nonnull final Object fields) {
```
elharo: This looks like a breaking API change. Is that true? If so, this would require a major version bump.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
```diff
@@ -226,7 +185,7 @@ public T apply(List<T> results) {
    * @return An ApiFuture that will be resolved when the write finishes.
    */
   @Nonnull
-  public ApiFuture<WriteResult> update(@Nonnull Map<String, Object> fields) {
+  public ApiFuture<WriteResult> update(@Nonnull Object fields) {
```
schmidt-sebastian: Updates still needs to take in a Map of <String,Object> and should not accept top-level POJOs (nested POJOs are ok). The reason for this is that `update()` treats the keys of the `fields` map as field paths and not as string literals. If you passed a POJO with a property name of "foo.bar", `update()` would actually updated the nested field "bar" in map "foo".

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -146,33 +149,16 @@ private Mutation addMutation() {
...
-    return performCreate(documentReference, (Map<String, Object>) data);
-  }
-
   /**
    * Overwrites the document referred to by this DocumentReference. If the document doesn't exist
    * yet, it will be created. If a document already exists, it will be overwritten.
    *
    * @param documentReference The DocumentReference to overwrite.
-   * @param fields A map of the field paths and values for the document.
+   * @param fields The (Map or POJO) that will be used to populate the document contents.
```
schmidt-sebastian: Please drop the parenthesis.

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -342,11 +297,14 @@ private T performSet(
    * @return The instance for chaining.
    */
   @Nonnull
-  public T update(
-      @Nonnull DocumentReference documentReference, @Nonnull Map<String, Object> fields) {
+  public T update(@Nonnull DocumentReference documentReference, @Nonnull Object fields) {
+    Object data = CustomClassMapper.convertToPlainJavaTypes(fields);
```
schmidt-sebastian: As stated above, the top-level `fields` value should be a Map, but individual values can be custom POJOs.

schmidt-sebastian:
> I think we need to keep all of the existing surface methods, and treat the new methods as overloaded. When method dispatch happens the most specific method should be invoked.
> 
> [1] `mvn -Dmaven.test.skip.exec=true clean install && mvn --projects :google-cloud-firestore --also-make clirr:check`

This seems like a problem with the tooling. We rolled a similar change out on Android without a major version bump: https://github.com/firebase/firebase-android-sdk/pull/76/files

google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
schmidt-sebastian: I don't see why this would be breaking - do you have an example?

PR #6856, opened 2019-11-20
Add toString() method to LifecycleRule

JesseLovelace:
Adds toString() method to LifecycleRule


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -362,6 +362,18 @@ public LifecycleCondition getCondition() {
       return lifecycleCondition;
     }
 
+    @Override
+    public String toString() {
+      return MoreObjects.toStringHelper(this)
+          .add("action", lifecycleAction.getActionType())
```
frankyn: Could you do a toString() for ActionType? It's missing setStorageClass information in this case.

PR #6859, opened 2019-11-21
Storage: Allow null values in labels map

JesseLovelace:
The way to delete a label is to set its value in the labels map to null. Currently, the labels map is converted to an ImmutableMap, which can't have null values. It's still possible to delete values by setting their values to the empty string, but that's not intuitive, users would generally expect setting null to work based on the documentation. This converts null values to the empty string in the builder to accomplish this. This was discovered while updating Java samples


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1209,7 +1211,17 @@ public Builder setDefaultAcl(Iterable<Acl> acl) {
...
     public Builder setLabels(Map<String, String> labels) {
-      this.labels = labels != null ? ImmutableMap.copyOf(labels) : null;
+      if (labels != null) {
+        this.labels =
+            Maps.transformValues(
+                labels,
+                new Function<String, String>() {
+                  @Override
+                  public String apply(String input) {
+                    return input == null ? "" : input;
```
frankyn: For consistency could you use `Data.<String>nullOf(String.class)`.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1209,7 +1211,17 @@ public Builder setDefaultAcl(Iterable<Acl> acl) {
 
     @Override
     public Builder setLabels(Map<String, String> labels) {
-      this.labels = labels != null ? ImmutableMap.copyOf(labels) : null;
+      if (labels != null) {
```
frankyn: Add an integration test to exercise this case.

google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
JesseLovelace: Seems obtuse to me, but I'll do it for consistency. Is there a reason it wouldn't be better/more readable to do it the other way around and replace other instances of  Data.<String>nullOf(String.class) with "" for readability?

frankyn: Could you file an issue to better address this in the library for readability? So it's done all at once instead of a one-off and then we can maintain consistency for now.

frankyn:
I see you added a unit test. could you add an integration test as well? Overall LGTM.

PR #6860, opened 2019-11-21
Change SigV4 canonical request formatting.

houglum:
This change ensures that query parameters are sorted (this will
facilitate more easily adding support for users to supply signed query
params in a future PR), and changes the way whitespace normalization is
done for headers (converts all non-space whitespace to space chars, then
condenses all whitespace sequences into a single space, as is outlined
in the spec).

This addresses the issues described in https://issuetracker.google.com/issues/144304815 (for the Java client library only).

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/SignatureInfo.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/SignatureInfo.java
@@ -20,13 +20,13 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSortedMultiset;
 import com.google.common.collect.Lists;
 import com.google.common.hash.Hashing;
 import com.google.common.net.UrlEscapers;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


houglum:
@elharo - Upcoming changes (i.e. allowing users to supply query params to be signed for V4 signatures) will require additional tests, but I think the changes here just tweak the existing implementation (whitespace change is a simplification of the regex, and the ordering of the query params isn't actually being changed).

- The existing conformance tests and integration tests already test the v4 query string construction (it's needed for every test).
- The conformance tests already test the whitespace changes:
https://github.com/googleapis/java-conformance-tests/blob/650ad977e0e0ddf6fe1eee0857f8d1bcc0bf04e9/src/main/resources/com/google/cloud/conformance/storage/v1/v4_signatures.json#L73

If I'm missing anything that needs to be tested, can you describe what additional tests you'd like me to add?  Thanks!

houglum:
The changes (across all client libraries) we need to make for V4 signatures are described in the doc linked in the bug I mentioned above (see https://issuetracker.google.com/issues/144304815#comment2).

While this change does introduce new internal behavior (explicitly calling `sort()` on the collection of query params), it doesn't actually change the returned value of the method, which is just a query string with all the params in sorted order.  The old behavior was just to manually construct the string, ensuring the query params were added in code in sorted order :upside_down_face: And if this implementation detail did change even one character, it would alter the signature of the URL and cause existing tests to fail.

Introducing this sorting behavior is mostly a precursor to a future PR where we'll add support for users to supply additional query params; it's in progress in this branch -- it's not done yet, as that's the one that will need additional testing:
https://github.com/houglum/google-cloud-java/tree/signurl-2019-11-add-with-query-param

houglum:
@frankyn This should be good to pull in now.

There wasn't much to do here; this library is already testing all the things we were concerned about and handling them correctly, so I just tweaked a couple of implementation details to be consistent with the recommended behavior in the design doc you reviewed.

PR #6872, opened 2019-11-22
chore: remove split out code

codyoss:
Remove code that has been split out to its own repo.
This includes the following projects:
- compute
- logging
- dialogflow
- dlp
- dns
- errorreporting
- kms
- language
- monitoring
- resourcemanager
- talent
- video_intelligence
- vision


codyoss:
Speech has already been removed, which is why the test is failing. We still need to clean up the kokoro job

PR #6903, opened 2019-11-26
deps: google-cloud-clients should not import google-cloud-bom

chingor13:
The clients published from this repo should not depend upon the google-cloud-bom which includes these clients.

In this PR:
* google-cloud-clients parent pom no longer imports the google-cloud-bom
* google-cloud-clients parent pom now dependency manages google-cloud-core, gax, iam and common protos
* google-cloud-clients parent pom no longer dependency manages versions for artifacts used by a single client (apiary client versions)
* updates gax to 1.50.1 for dependency convergence

### DIFF
--- a/google-cloud-examples/pom.xml
+++ b/google-cloud-examples/pom.xml
@@ -68,7 +68,6 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-securitycenter</artifactId>
-      <version>0.117.0</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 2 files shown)


google-cloud-examples/pom.xml
```diff
@@ -68,6 +68,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-securitycenter</artifactId>
+      <version>0.117.0</version>
```
elharo: The examples should depend on the bom (or a BOM) if we want users to use the BOM for their own code.

chingor13: The examples project does import the google-cloud-bom. I'll remove this version and make sure it works

PR #6926, opened 2019-11-26
Extract BOM to https://github.com/googleapis/java-cloud-bom

chingor13:


chingor13:
The BOM upper bounds test is moving to the split repository

PR #6951, opened 2019-11-29
spanner: add benchwrapper.

hengfengli:
This is a benchwrapper for spanner, which is similar to the existing `storage-benchwrapper`. 

* This is mainly used for benchmarking tests. 
* It exposes a gRPC service for Cloud Spanner client libraries, which contains three basic operations, i.e., read, insert, and update. 
* The uniform gRPC service makes the caller easy to benchmark Cloud Spanner client libraries in different languages. 

This also includes a change to support `SPANNER_EMULATOR_HOST` in the spanner client library.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
+++ b/google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
@@ -405,11 +405,13 @@ public class SpannerOptionsTest {
 
   @Test
   public void testSetEmulatorHostWithoutProtocol() {
-    // If the host doesn't have a protocol as a prefix, it will automatically be prefixed with "http://".
-    SpannerOptions options = SpannerOptions.newBuilder()
-        .setProjectId("[PROJECT]")
-        .setEmulatorHost("localhost:1234")
-        .build();
+    // If the host doesn't have a protocol as a prefix, it will automatically be prefixed with
+    // "http://".
+    SpannerOptions options =
+        SpannerOptions.newBuilder()
+            .setProjectId("[PROJECT]")
+            .setEmulatorHost("localhost:1234")
+            .build();
     assertThat(options.getHost()).isEqualTo("http://localhost:1234");
     assertThat(options.getEndpoint()).isEqualTo("localhost:1234");
   }
@@ -417,10 +419,11 @@ public class SpannerOptionsTest {
   @Test
   public void testSetEmulatorHostWithProtocol() {
     // If the host has a protocol, it should not be modified.
-    SpannerOptions options = SpannerOptions.newBuilder()
-        .setProjectId("[PROJECT]")
-        .setEmulatorHost("http://localhost:1234")
-        .build();
+    SpannerOptions options =
+        SpannerOptions.newBuilder()
+            .setProjectId("[PROJECT]")
+            .setEmulatorHost("http://localhost:1234")
+            .build();
     assertThat(options.getHost()).isEqualTo("http://localhost:1234");
     assertThat(options.getEndpoint()).isEqualTo("localhost:1234");
   }


google-cloud-testing/spanner-benchwrapper/pom.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>spanner-benchwrapper</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <groupId>com.google.cloud</groupId>
+
+  <properties>
+    <grpcVersion>1.24.1</grpcVersion>
```
elharo: This is a little out of date. Could we use libraries-bom:3.0.0 to set versions?

hengfengli: Hi @elharo, the latest `grpc-bom` is 1.25.0. Here the code depends on `com.google.api.grpc:grpc-google-cloud-spanner-v1:jar:1.46.1-SNAPSHOT`, which has a dependency on `grpc-protobuf:1.24.1`. 

```bash
$mvn dependency:tree
...
[INFO] |  +- com.google.api.grpc:grpc-google-cloud-spanner-v1:jar:1.46.1-SNAPSHOT:compile
[INFO] |  |  +- io.grpc:grpc-stub:jar:1.24.1:compile
[INFO] |  |  \- io.grpc:grpc-protobuf:jar:1.24.1:compile
[INFO] |  |     \- io.grpc:grpc-protobuf-lite:jar:1.24.1:compile
...
```

Could you please explain a bit more? Thanks.

elharo: see https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/The-Google-Cloud-Platform-Libraries-BOM

hengfengli: Thanks for the doc. I can add `libraries-bom:3.0.0` into `dependencyManagement`. But I wonder here `grpcVersion` is only used to configure `pluginArtifact` but not used for specifying the version, is it necessary to add `libraries-bom:3.0.0`? Because the pom file does not explicitly import `grpc-protobuf` as a dependency.

hengfengli: I think I got what you mean here. I will update it. But I still need to specify the version for `pluginArtifact` and `protocArtifact`.

hengfengli:
Hi @elharo, I fixed the issue but I wonder why `codecov` reports a decrease of coverage. I did not modify those files in my commit. How can I rerun the `codecov` check again?

google-cloud-testing/spanner-benchwrapper/README.md
```diff
@@ -0,0 +1,54 @@
...
+    --java_out=src/main/java \
+    --grpc-java_out=src/main/java \
+    --proto_path=src/main/proto \
+    src/main/proto/*.proto
+```
+
+Note that you'll need to delete these, or uncomment the plugin, since multiple
+definitions of the same class are not allowed.
+
+Note that these should not be committed into git.
```
elharo: delete "Note that"

google-cloud-testing/spanner-benchwrapper/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Note that you can also `mvn compile -DskipTests=true` and see sources in
```
elharo: delete "Note that"

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
```diff
@@ -0,0 +1,63 @@
...
+      // consistency is nice to maintain.
+      System.err.println("Please set SPANNER_EMULATOR_HOST=localhost:8080");
+      System.exit(1);
+    }
+
+    System.out.println("Server starting up...");
+
+    int portInt = Integer.parseInt(port);
+    final Server server = NettyServerBuilder
+      .forPort(portInt)
```
elharo: four space indent for continuation lines

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,133 @@
...
+      .setChannelConfigurator(
+        new ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder>() {
+          @Override
+          public ManagedChannelBuilder apply(ManagedChannelBuilder builder) {
+            return builder.usePlaintext();
+          }
+        }
+      );
+
+    SpannerOptions options = builder
```
elharo: one line

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,133 @@
...
+import com.google.cloud.spanner.SpannerBatchUpdateException;
+import com.google.cloud.spanner.TransactionContext;
+
+class SpannerBenchWrapperImpl extends SpannerBenchWrapperImplBase {
+  private Spanner spanner;
+  private DatabaseClient dbClient;
+
+  public SpannerBenchWrapperImpl(String spannerEmulatorHost) {
+    SpannerOptions.Builder builder = SpannerOptions.newBuilder()
+      .setHost("http://" + spannerEmulatorHost)
```
elharo: four space indents

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,133 @@
...
+          }
+        }
+      );
+
+    SpannerOptions options = builder
+      .build();
+    spanner = options.getService();
+
+    dbClient = spanner.getDatabaseClient(
+      DatabaseId.of(options.getProjectId(), "test-instance", "test-db"));
```
elharo: four spaces

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public void insert(InsertQuery request, StreamObserver<EmptyResponse> responseObserver) {
+    System.out.println("insert has been called");
+
+    List<Mutation> mutations = new ArrayList<>();
+    for (User user: request.getUsersList()) {
+      mutations.add(
+          Mutation.newInsertBuilder("sometable")
+            .set("name")
```
elharo: four spaces

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,133 @@
...
+  public void update(UpdateQuery request, StreamObserver<EmptyResponse> responseObserver) {
+    System.out.println("update has been called");
+
+    dbClient
+      .readWriteTransaction()
+      .run(
+          new TransactionCallable<Void>() {
+            @Override
+            public Void run(TransactionContext transaction) throws Exception {
+              List<Statement> stmts = new ArrayList<Statement>();
```
elharo: per google style use unabbreviated variable names; here statements instead of stmts

elharo: also <Statement> --> <>

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,133 @@
...
+              for (String query: request.getQueriesList()) {
+                stmts.add(Statement.of(query));
+              }
+              long [] rowCounts;
```
elharo: no space after long

hengfengli:
> some more nits.
> You should also open a bug and get commitment from the owners that this is something they want to add to the repo. They might not.

I am from the Cloud Spanner client library team and we are running benchmarking tests regularly across all spanner client libraries. The go version has been done in https://github.com/googleapis/google-cloud-go/tree/master/spanner/internal/benchwrapper. So we are working on expanding it to other client libraries.

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
```diff
@@ -0,0 +1,64 @@
...
+    if (port == null || port.equals("")) {
+      System.err.println("Usage: mvn clean install exec:java -DskipTests=true -Dport=8081");
+      System.exit(1);
+    }
+
+    String spannerEmulatorHost = System.getenv("SPANNER_EMULATOR_HOST");
+    if (spannerEmulatorHost == null || spannerEmulatorHost.equals("")) {
+      // We could use system properties here too, but every other language uses
+      // an environment variable called SPANNER_EMULATOR_HOST, so the
+      // consistency is nice to maintain.
```
skuruppu: I don't think you need to have this comment.

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.List;
+
+class SpannerBenchWrapperImpl extends SpannerBenchWrapperImplBase {
+  private Spanner spanner;
+  private DatabaseClient dbClient;
+
+  public SpannerBenchWrapperImpl(String spannerEmulatorHost) {
+    SpannerOptions.Builder builder =
+        SpannerOptions.newBuilder()
+            .setHost("http://" + spannerEmulatorHost)
```
skuruppu: Given that we generically want the Java client lib to support an emulator so that users can use an emulator without having to change the code on how they init the client lib, I think it would be better if we do this in the client lib code.

This is how it's done in [Go](https://github.com/googleapis/google-cloud-go/blob/1118b0760db3b1989320683fd9ce284a48749a6a/spanner/client.go#L132) currently and the [Python PR](https://github.com/googleapis/google-cloud-python/pull/9902) is the same.

google-cloud-clients/google-cloud-spanner/pom.xml
```diff
@@ -150,6 +150,12 @@
       <artifactId>guava-testlib</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
hengfengli: Hi @elharo, is it okay to import this lib in `google-cloud-java` spanner? I would like to test the setting on an environment variable. Can I ask what the best practice would be for this case? Thanks.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -402,4 +404,19 @@ public void testSetClientLibToken() {
...
   }
+
+  @Test
+  public void testSetEmulatorHostEnv() {
+    environmentVariables.set("SPANNER_EMULATOR_HOST", "localhost:1234");
+    SpannerOptions options = SpannerOptions.newBuilder().setProjectId("[PROJECT]").build();
+    assertThat(options.getHost()).isEqualTo("http://localhost:1234");
+    assertThat(options.getEndpoint()).isEqualTo("localhost:1234");
+
+    environmentVariables.set("SPANNER_EMULATOR_HOST", "http://localhost:1234");
```
elharo: This looks like you're testing the same thing twice.

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
```diff
@@ -0,0 +1,61 @@
...
+
+package com.google.cloud.benchwrapper;
+
+import io.grpc.Server;
+import io.grpc.netty.NettyServerBuilder;
+import java.util.Properties;
+
+class Main {
+  public static void main(String[] args) throws Exception {
+    Properties properties = new Properties(System.getProperties());
```
elharo: Why do you make a copy of this?

google-cloud-clients/google-cloud-spanner/pom.xml
elharo: It's test scope, so this is not a big deal. I would still consider whether you truly need this. If it's only a minor convenience I wouldn't add it.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
hengfengli: Originally, the first case would throw an error when converting it to a Url object because of lack of protocol (in getEndpoint() method).

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
hengfengli: I got this from the existing code:  
https://github.com/googleapis/google-cloud-java/blob/26415a199764839f4a70640b6a8a7c4b2c953ad9/google-cloud-testing/storage-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java#L25. 

But I agree that this is not necessary. I will update it.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
elharo: Could this be two methods/ I still don't follow what's going on here. A comment m,ight be helpful.

hengfengli: The current convention is to assign `localhost:1234` to `SPANNER_EMULATOR_HOST` without the protocol prefix.

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -0,0 +1,129 @@
...
+class SpannerBenchWrapperImpl extends SpannerBenchWrapperImplBase {
+  private Spanner spanner;
+  private DatabaseClient dbClient;
+
+  public SpannerBenchWrapperImpl() {
+    SpannerOptions.Builder builder =
+        SpannerOptions.newBuilder()
+            .setProjectId("test-project-id")
+            // Channels are secure by default (via SSL/TLS). For the example we disable TLS to avoid
+            // needing certificates.
```
skuruppu: We might need to move this logic to the client library as well. Otherwise by setting SPANNER_EMULATOR_HOST, it wouldn't work directly with the client library if TLS is not disabled.

hengfengli: Thanks for pointing this out. Fixed.

google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SpannerOptionsTest.java
```diff
@@ -407,21 +405,23 @@ public void testSetInvalidClientLibToken() {
 
   @Test
   public void testSetEmulatorHostWithoutProtocol() {
-    // If the host has not protocol as prefix, it will automatically prefix it with "http://"
-    environmentVariables.set("SPANNER_EMULATOR_HOST", "localhost:1234");
-    SpannerOptions options = SpannerOptions.newBuilder().setProjectId("[PROJECT]").build();
+    // If the host has not protocol as prefix, it will automatically prefix it with "http://".
```
skuruppu:
```suggestion
    // If the host doesn't have a protocol as a prefix, it will automatically be prefixed with "http://".
```

PR #6954, opened 2019-11-29
Storage : create a signed upload policy for uploading objects

### LINKED ISSUE #6657: Allow Signing Policy Document
The current SDK doesn't support signing policy documents, only signing URLs.

See: https://cloud.google.com/storage/docs/xml-api/post-object#policydocument

Since signed URLs doesn't support limiting the size of the uploaded file, signing policy documents is a much needed feature.

athakor:
Fixes #6657 

Did the Implementation as per python [Bucket.generate_upload_policy](https://github.com/googleapis/google-cloud-python/blob/ee6214251fbe83cb83a9b5b5019c605a27a01354/storage/google/cloud/storage/bucket.py#L1997-L2057). Removing [snippets](https://github.com/googleapis/google-cloud-java/pull/6954/commits/4e2f5c6a589730936c4d41e626454e5e24dccb69), As google-cloud-bom is pulling older version of `google-cloud-storage`. I will raise a separate PR for snippets once google-cloud-bom is updated

### DIFF
--- a/google-cloud-examples/pom.xml
+++ b/google-cloud-examples/pom.xml
@@ -85,6 +85,7 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-vision</artifactId>
     </dependency>
+
     <!-- Test dependencies -->
     <dependency>
       <groupId>junit</groupId>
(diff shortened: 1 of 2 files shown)


frankyn:
@athakor, thanks for creating this PR. I'd say keep PR as an example as we go into Q12020 but we will need to modify it to work with V4 signed URLs as the documentation pointed to right now is specifically for V2. 

I'm closing this PR as we should not continue adding support for V2 features.

PR #7001, opened 2019-12-06
Add support for endpoint URLs

mbril:
Add support for endpoint URLs.


hengfengli:
@mbril Do we still need to review this PR? Could you please fix the CI check errors before we can review it? Thanks.

PR #7007, opened 2019-12-09
Updating libraries-bom and linkage-checker version

suztomo:
The latest libraries-bom is 3.1.0.
The latest linkage-checker-enforcer-rules is 1.1.0.


suztomo:
@chingor13 @codyoss Would you check this documentation update?

PR #7012, opened 2019-12-10
Add fn to correctly encode parts of signed URLs

### LINKED ISSUE #7008: Cloud Storage: Signed URL tests don't actually test that reserved characters are URI encoded
- Checked out the repo at v0.119.0, up through commit 5a2c0ae6a18.

It looks like the StorageImpl tests for encoding special characters in signed URLs doesn't actually do what it's supposed to.  The logic is here:

https://github.com/googleapis/google-cloud-java/blob/78916cf6d43aab09f92234156327fd4076e8db54/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java#L1984

But, note that `urlFragmentEscaper` does **not** escape those characters:
https://guava.dev/releases/19.0/api/docs/com/google/common/net/UrlEscapers.html#urlFragmentEscaper()

By adding in a `System.out.println` to print out the `escapedBlobName` string in that test, I verified that the resulting URL is in fact using the raw reserved characters, rather than their percent-encoded values.  That is, the URL contains `a!b` instead of `a%21b`.  I also verified that an object with such a name *should* have those characters URL encoded by using an S3 library (botocore) to generate a signed URL for it.

The test above just calls the same encoding logic used in the StorageImpl code, which (semantically) seems to be pretty useless -- the result will always be the same if you use the same encoding statements, yes, but that doesn't necessarily test that the encoding approach is right. We should be testing that each character is encoded correctly via hard-coding the expected result (e.g. making a map of characters to strings like `'!' -> "%21"`).

houglum:
The `Rfc3986UriEncode` method should be used to URI-encode various parts of
a signed URL. All path parts should be encoded the same, except for the
part identifying the GCS resource, which should not URI-encode forward
slashes. Query parameters (support coming in a later commit), should
perform this encoding on parameter keys and values, but should also
encode forward slashes.

Other Cloud providers also perform this same encoding; see examples / pseudo
code below:
https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
https://docs.aws.amazon.com/AWSECommerceService/latest/DG/Query_QueryAuth.html

Fixes #7008

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
+++ b/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
@@ -338,7 +338,7 @@ public class StorageImplTest {
 
   // List of chars under test were taken from
   // https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters
-  private static final Map<Character, String> URI_ENCODING_MAP =
+  private static final Map<Character, String> RFC3986_URI_ENCODING_MAP =
       ImmutableMap.<Character, String>builder()
           .put('!', "%21")
           .put('#', "%23")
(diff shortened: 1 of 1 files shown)


houglum:
@frankyn

houglum:
Added one more commit to also test that the chars `/~._-` are correctly handled.  This should prevent any future commits that might accidentally change this behavior by swapping out the underlying encoding functionality.

PR #7028, opened 2019-12-11
Spanner: Add block-with-timeout option to session pool

olavloite:
Adds the option to block with a timeout when the session pool is exhausted. The session pool will throw a SessionPoolExhaustedException if no session becomes available within the duration of the specified timeout. This exception contains a list of all stack traces of the threads that have checked out a session. This information can be used to track down potential session leaks in the client application.

### DIFF
--- a/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
+++ b/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java
@@ -921,8 +921,9 @@ final class SessionPool {
      * Creates a waiter to wait for a session to become available.
      *
      * @param exhausted indicates whether this waiter is created because the session pool is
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java
```diff
@@ -237,7 +245,7 @@ private Exception getFailedError() {
   @Test
   public void stressTest() throws Exception {
     int concurrentThreads = 10;
-    final int numOperationsPerThread = 1000;
+    final int numOperationsPerThread = 500;
```
olavloite: As we are adding an additional parameter value to the test set, we are also increasing the number of test cases. This means that the build time also increases from approx 30 seconds to approx 60 seconds. Reducing this value from 1000 to 500 keeps the build time at approx the same value as prior to this change.

elharo:
now in a separate repo

PR #7045, opened 2019-12-16
Allow adding query params to GCS signed URLs

### LINKED ISSUE #7044: Cloud Storage library should allow supplying query parameters when generating signed URLs
google-cloud-storage allows users to generate signed URLs via the `signUrl` method. However, it currently doesn't allow users to supply additional query parameters to be included with the URL.

This isn't as big of an issue for V2 as it is for V4 signing.  For V2 URLs, you can just add/remove/alter params after the URL has been generated - the V2 signing spec doesn't use query params when calculating the signature.  But for V4 signed URLs, the params are canonical; they're used when creating the signature and the set of params cannot be altered after the signature is calculated.  This means you can't do certain operations with V4 signed URLs today (e.g. listing objects in a bucket that start with a given prefix, since the prefix must be specified via query param).

Corresponding internal issue tracker:
https://issuetracker.google.com/issues/145017709

@frankyn

houglum:
Fixes #7044.

@frankyn

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/SignatureInfo.java
+++ b/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/SignatureInfo.java
@@ -46,6 +46,10 @@ public class SignatureInfo {
   public static final String SCOPE = "/auto/storage/goog4_request";
   private static final List<String> RESERVED_PARAMS_LOWER =
       ImmutableList.<String>of(
+          // V2:
+          "expires",
+          "googleaccessid",
+          // V4:
           "x-goog-algorithm",
           "x-goog-credential",
           "x-goog-date",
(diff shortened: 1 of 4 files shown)


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/SignatureInfo.java
```diff
@@ -167,49 +176,80 @@ private String constructV4CanonicalRequestHash() {
         .toString();
   }
 
-  public String constructV4QueryString() {
-    ArrayListMultimap<String, String> paramMap = ArrayListMultimap.create();
+  /**
+   * Returns a TreeMap containing the user-supplied query parameters that do not have reserved keys.
+   */
+  private TreeMap<String, String> getNonReservedUserQueryParams() {
```
frankyn: Instead of returning a TreeMap would it be sufficient to return a Map<> instead?

houglum: It didn't make sense to me to return any type of Map -- we want a Map that will iterate over the keys in sorted order, so TreeMap made the most sense here IMO.  If it were a public method, I'd be in favor of accepting whatever type of map, but this is a private method that just factors out some common functionality between the two query string construction methods.

frankyn: Sg, thanks @houglum. Let's merge it in.

PR #7057, opened 2019-12-18
Add disableGzipContent option for create from InputStream

### LINKED ISSUE #7056: Cloud Storage: provide option to disableGzipContent on InputStream variants
**Is your feature request related to a problem? Please describe.**

I have a use case that necessitates using the Storage/Bucket write method variants with InputStream arguments. The inputs for my use case to store in Cloud Storage are:

* predominantly already compressed,
* have a wide range of file sizes, from a few KB to GB+,
* we don't always know the size in advance.

 **Describe the solution you'd like**

Given that our content is already compressed, I would prefer to avoid spending the CPU time on compressing the content again en route to the Bucket.

 **Describe alternatives you've considered**

We have used the `byte[]` variants, with `BlobTargetOption.disableGzipContent` BlobTarget and the Compose request. This is suitable but leaves us with a tuning challenge:

* Buffering these streams into memory (`byte[]` chunks) requires additional heap space be available
* Compose is limited to only 32 chunks. If the file content we intend to store is far larger than `32 * bufferSize`, we will upload 31 small sized chunks and 1 large chunk that we have to use the `InputStream` variant for anyways, and pay the additional overhead of gzip compression.

 **Additional context**

I have done some exploration and it appears that the values on the BlobWriteOption (used on `InputStream` variants) and BlobTargetOption (used on `byte[]` variants) enums both are translated into `StorageRpc.Option`. I have a small contribution to offer in the form of a pull request to follow.

nblair:
Previously, only the methods to create blobs that take a `byte[]` argument offer the option to disable gzip compression; the methods that accept an `InputStream` argument do not. This is due to the `BlobWriteOption` enum missing a matching constant for `BlobTargetOption.IF_DISABLE_GZIP_CONTENT`.

This change set adds a matching `IF_DISABLE_GZIP_CONTENT` constant to `BlobWriteOption` including the correct translation to StorageRpc.Option. The net result is that the Storage create functions that accept an InputStream now offer the option to disable gzip compression.

Fixes #7056.

### DIFF
--- a/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
+++ b/google-cloud-clients/google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
@@ -714,10 +714,10 @@ public class StorageImplTest {
     BlobInfo infoWithHashes = infoBuilder.setMd5(CONTENT_MD5).setCrc32c(CONTENT_CRC32C).build();
     BlobInfo infoWithoutHashes = infoBuilder.setMd5(null).setCrc32c(null).build();
     EasyMock.expect(
-        storageRpcMock.create(
-            EasyMock.eq(infoWithoutHashes.toPb()),
-            EasyMock.capture(capturedStream),
-            EasyMock.eq(BLOB_TARGET_OPTIONS_CREATE_DISABLE_GZIP_CONTENT)))
+            storageRpcMock.create(
+                EasyMock.eq(infoWithoutHashes.toPb()),
+                EasyMock.capture(capturedStream),
+                EasyMock.eq(BLOB_TARGET_OPTIONS_CREATE_DISABLE_GZIP_CONTENT)))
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -699,6 +700,10 @@ public static BlobWriteOption kmsKeyName(String kmsKeyName) {
     public static BlobWriteOption userProject(String userProject) {
       return new BlobWriteOption(Option.USER_PROJECT, userProject);
     }
+
+    public static BlobWriteOption disableGzipContent() {
```
nblair: Is a javadoc desired here? If so, what do you think of the text:

> Returns an option that signals automatic gzip compression should not be performed en route to the bucket.

athakor:
closing this PR as it's already merged in respective repo.

PR #7082, opened 2020-01-11
Simple NIO tests

jean-philippe-martin:
Make sure we find the provider and it recognizes paths as expected.
These tests don't actually connect to any Google Cloud Storage
bucket.

### DIFF
--- a/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/NIOTest.java
+++ b/google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/NIOTest.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage.contrib.nio;
+
+import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth.assertWithMessage;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.is;
+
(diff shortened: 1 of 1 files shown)


google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/NIOTest.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright 2017 Google LLC
```
elharo: 2020

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/NIOTest.java
```diff
@@ -0,0 +1,77 @@
...
+@RunWith(JUnit4.class)
+public class NIOTest {
+
+  /** We can create a Path object from a gs:// URI. **/
+  @Test
+  public void testCreatePath() {
+    URI uriToCloudStorage = URI.create("gs://bucket/file.txt");
+    // Return value ignored on purpose, we just want to check
+    // no exception is thrown.
+    Paths.get(uriToCloudStorage);
```
elharo: at the least, assert this is not null

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/NIOTest.java
```diff
@@ -0,0 +1,77 @@
...
+    URI uriToCloudStorage = URI.create("gs://bucket/file.txt");
+    Path path = Paths.get(uriToCloudStorage);
+    assertThat(path.getFileSystem().provider().getScheme()).isEqualTo("gs");
+  }
+
+  /** The created Path object has the correct provider class. **/
+  @Test
+  public void testCreatedPathIsCorrectInstance() {
+    URI uriToCloudStorage = URI.create("gs://bucket/file.txt");
+    Path path = Paths.get(uriToCloudStorage);
```
elharo: You use this setup three times so it should probably be a fixture.

google-cloud-clients/google-cloud-contrib/google-cloud-nio/src/test/java/com/google/cloud/storage/contrib/nio/NIOTest.java
```diff
@@ -0,0 +1,77 @@
...
+    assertThat(path.getFileSystem().provider())
+        .isInstanceOf(CloudStorageFileSystemProvider.class);
```
elharo: We might not want to assert this. Is this part of the published API in some way? Would it matter if this were a different class?

jean-philippe-martin:
Does it still make sense to add this test here, given that the NIO library moved to https://github.com/googleapis/java-storage-nio ?

elharo:
now in a separate repo

PR #7125, opened 2020-02-06
google-cloud-java 0.35.0 is too old to worry about

elharo:
@BenWhitehead


BenWhitehead:
I'm going to defer to jeff on this one, I'm less familiar with the versions timeline than he is.

PR #7188, opened 2020-04-02
transactionManager ex: tolerate aborts during read

clifffrey:
It is possible that txn.readRow can return an aborted error, and thus the try/catch should wrap that as well, instead of just the commit operation.


skuruppu:
@clifffrey I think a run of the code formatting tool is necessary for the presubmits to pass. It's good for merging after that.

clifffrey:
ok, I think that I updated the pull request to satisfy the formatter

PR #7222, opened 2020-04-30
Add sample for V4 POST policy

JesseLovelace:
Adds sample for V4 POST policy.

Currently depends on snapshot version, will change once POST policy is released


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/GenerateSignedPostPolicyV4.java
```diff
@@ -0,0 +1,57 @@
...
+
+    // The name to give the object uploaded to GCS
+    // String blobName = "your-object-name"
+
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+
+    PostPolicyV4.PostFieldsV4 fields =
+        PostPolicyV4.PostFieldsV4.newBuilder().AddCustomMetadataField("test", "data").build();
+
+    /**
```
frankyn, line 27: Can you put this above the method signature `generateSignedPostPolicyV4`?

google-cloud-examples/pom.xml
```diff
@@ -76,6 +76,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-storage</artifactId>
+      <version>1.108.0</version>
```
frankyn, line 79: File an issue to remove once libraries-bom is updated.

PR #7229, opened 2020-05-06
chore: adding Storage team to CODEOWNERS for samples

czahedi:
Adding storage team for auth / repo triage

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,11 +2,11 @@
 # This file controls who is tagged for review for any given pull request.
 
 # Owners per module.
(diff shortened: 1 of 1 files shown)


czahedi:
But the samples do, right? In case one of those breaks it would be useful to have this.

chingor13:
Sure, but the samples live somewhere under `google-cloud-examples/`.

czahedi:
There are storage samples in both: 

/google-cloud-examples/src/test/java/com/google/cloud/examples/storage/snippets

and

/google-cloud-examples/src/test/java/com/google/cloud/examples

PR #7254, opened 2020-06-04
chore: README cleanup

chingor13:
* Remove the quickstart as that does not apply to every client. Users should refer to the individual READMEs for quickstart
* Moves IDE section to bottom (not as important of content)
* Cleans up the authentication section


chingor13:
@elharo I'm removing the quickstart with the install snippet here because no clients live here anymore and each individual library (linked in the table) has the install instructions at the top of the respective READMEs

PR #7288, opened 2020-07-08
docs(samples): remove bucket CORS configuration

athakor:


athakor:
@frankyn PTAL

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/buckets/RemoveBucketCors.java
```diff
@@ -0,0 +1,40 @@
...
+  public static void removeBucketCors(String projectId, String bucketName) {
+    // The ID of your GCP project
+    // String projectId = "your-project-id";
+
+    // The ID of your GCS bucket
+    // String bucketName = "your-unique-bucket-name";
+
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+    Bucket bucket = storage.get(bucketName);
+    bucket.toBuilder().setCors(ImmutableList.<Cors>of()).build().update();
```
frankyn, line 35: does null not work here?

athakor: setting `null `is also works here

athakor: @frankyn `null `setting doesn't works here see the respected example with response
```javascript

        Storage storage = StorageOptions.getDefaultInstance().getService();
        String bucketName = RemoteStorageHelper.generateBucketName();
        try{
            Cors cors = Cors.newBuilder()
                    .setOrigins(ImmutableList.of(Cors.Origin.of("*")))
                    .setMethods(ImmutableList.of(HttpMethod.GET))
                    .setResponseHeaders(ImmutableList.of("Content-Type"))
                    .setMaxAgeSeconds(100)
                    .build();
            Bucket bucket =
                    storage.create(BucketInfo.newBuilder(bucketName).setCors(ImmutableList.of(cors)).build());
            System.out.println("Prior size : " + bucket.getCors().size());
            Bucket updatedBucket  = bucket.toBuilder().setCors(null).build().update();
            System.out.println("After Updating operation size must be zero instead of : " + updatedBucket.getCors().size());

        }finally {
            RemoteStorageHelper.forceDelete(storage, bucketName, 5, TimeUnit.SECONDS);
        }
```
**Response :**
```response
Prior size : 1
After Updating operation size must be zero instead of : 1
```

athakor: @frankyn I have raise the [issue ](https://github.com/googleapis/java-storage/issues/437)in `storage `repo regarding this and will fixed it soon

lesv: I can see @frankyn 's argument, but following it's point, why need to call `setCors()` at all if it's empty or NULL?
Lots of folks (Effective Java 3rd ed.) prefer empty lists to using NULL, and I was going to comment w/ that, but really, if it's empty, why have to set it at all?

frankyn: In this case the example is to remove CORS from a bucket metadata. Good call, I think we can move forward with this PR with `ImmutableList.<Cors>Of()`

frankyn: Maybe one additional change that could be made here @athakor is to confirm that CORS are set before making a subsequent call. That way you reduce the need to make a patch request.

athakor: done

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/buckets/RemoveBucketCors.java
```diff
@@ -0,0 +1,52 @@
...
+
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+    Bucket bucket =
+        storage.get(bucketName, Storage.BucketGetOption.fields(Storage.BucketField.CORS));
+
+    // getCors() returns the List and copying over to an ArrayList so it's mutable.
+    List<Cors> cors = new ArrayList<>(bucket.getCors());
+
+    // Remove bucket CORS configuration.
+    for (int index = 0; index < cors.size(); index++) {
```
frankyn, line 42: Use `cors.clear()` instead if all cors will be deleted.

athakor: done

PR #7292, opened 2020-07-22
Fix POST upload sample code submit field

meltsufin:
Without ignoring the `submit` field, you get the following error:

```xml
<Error>
  <Code>InvalidPolicyDocument</Code>
  <Message>The content of the form does not meet the conditions specified in the policy document.</Message>
  <Details>Policy did not reference these fields: submit</Details>
</Error>
```

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/GenerateSignedPostPolicyV4.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/GenerateSignedPostPolicyV4.java
@@ -68,7 +68,7 @@ public class GenerateSignedPostPolicyV4 {
               + "' type='hidden' />\n");
     }
     htmlForm.append("  <input type='file' name='file'/><br />\n");
-    htmlForm.append("  <input type='submit' value='Upload File' name='x-ignore-submit'/><br />\n");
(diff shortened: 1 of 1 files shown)


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/GenerateSignedPostPolicyV4.java
```diff
@@ -68,7 +68,7 @@ public static void generateSignedPostPolicyV4(
               + "' type='hidden' />\n");
     }
     htmlForm.append("  <input type='file' name='file'/><br />\n");
-    htmlForm.append("  <input type='submit' value='Upload File' name='submit'/><br />\n");
+    htmlForm.append("  <input type='submit' value='Upload File' name='x-ignore-submit'/><br />\n");
```
frankyn, line 71: In this case can you remove `name` attribute from submit?

frankyn (dismissed):

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/GenerateSignedPostPolicyV4.java
meltsufin: Yes, I believe so.

elharo (approved):

PR #7315, opened 2020-08-18
docs(sample): storage object get kms key

athakor:


frankyn:
Hi @athakor,

This isn't a sample we need / scoped. I'm closing it as it's already supported in the viewing metadata sample here: https://cloud.google.com/storage/docs/viewing-editing-metadata

Apologies for having you go down this path. Could you tell me how you pick this work up to be careful in the future?

PR #7330, opened 2020-09-10
(feat) storage: use String instead of Path in download sample

JesseLovelace:
Based on customer feedback, updates Download sample to use String as an argument instead of Path


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/DownloadObject.java
```diff
@@ -35,12 +36,14 @@ public static void downloadObject(
     // String objectName = "your-object-name";
 
     // The path to which the file should be downloaded
-    // Path destFilePath = Paths.get("/local/path/to/file.txt");
+    // String destFilePath = "/local/path/to/file.txt";
+
+    Path path = Paths.get(destFilePath);
```
frankyn, line 41: What about inlining it?

```java
blob.downloadTo(Paths.get(destFilePath));
```

Or moving `Path path = Paths.get(destFilePath);` near `blob.downloadTo(path)`?

frankyn (changes requested):
Have a small nit, but overall LGTM

frankyn (approved):

PR #7400, opened 2020-11-20
docs: announce Java 7 end of support date

elharo:
@lesv Getting this ready for the announcement. We can commit this once we have final signoff on the date.


lesv:
I'll send some different language, but we should do the blog post first.

PR #7405, opened 2020-11-26
samples: change Random to SecureRandom

drakoleat:
java.util.Random is not cryptographically secure. switch to SecureRandom.


frankyn (approved):

lesv (approved):

lesv:
Doesn't seem to compile w/ Java 11

lesv:
Looks like a build problem - rerunning

PR #7589, opened 2021-06-17
docs: Add Java 7 support column

lesv:
Jeff - is this the right way to add a column?


lesv:
Need to do this from code. :(

PR #7725, opened 2021-11-19
Add sample for making a batch request

JesseLovelace:
Adds a sample for making a batch request, internal bug 121151093

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
@@ -20,7 +20,7 @@ public class BatchSetObjectMetadata {
 
     // The directory prefix. All objects in the bucket with this prefix will have their metadata
     // updated
-    // String objectName = "your-object-name";
+    // String directoryPrefix = "yourDirectory/";
 
     Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
     Map<String, String> newMetadata = new HashMap<>();


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
```diff
@@ -0,0 +1,51 @@
...
+      String projectId, String bucketName, String directoryPrefix) {
+    // The ID of your GCP project
+    // String projectId = "your-project-id";
+
+    // The ID of your GCS bucket
+    // String bucketName = "your-unique-bucket-name";
+
+    // The directory prefix. All objects in the bucket with this prefix will have their metadata
+    // updated
+    // String objectName = "your-object-name";
```
cojenco, line 23: Is this supposed to be `directoryPrefix`?

cojenco (commented):
Other than the comment, LGTM!

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
JesseLovelace: Good catch, thanks!

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
```diff
@@ -0,0 +1,51 @@
...
+    Page<Blob> blobs =
+        storage.list(
+            bucketName,
+            Storage.BlobListOption.prefix(directoryPrefix),
+            Storage.BlobListOption.currentDirectory());
+    StorageBatch batchRequest = storage.batch();
+
+    // Add all blobs with the given prefix to the batch request
+    for (Blob blob : blobs.iterateAll()) {
+      batchRequest.update(blob.toBuilder().setMetadata(newMetadata).build());
```
BenWhitehead, line 37: Is there a max number of updates you can put into a batch? If someones has lots of objects I could imagine this possibly resulting in an error.

JesseLovelace: According to the [docs](https://cloud.google.com/storage/docs/batch), "You should not include more than 100 calls in a single batch request. If you need to make more calls than that, use multiple batch requests. The total batch request payload must be less than 10MB."

Should i put a note about that in the comments?

cbonnie (approved):

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
frankyn: I would recommend limiting the batch operation to 100 calls per batch operation / submit combination. 

Might be overkill though?

cojenco (approved):

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
JesseLovelace: @frankyn Like, in this sample? Or do you mean update the actual batch code?

danielduhh:
@frankyn this good to merge?

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/BatchSetObjectMetadata.java
frankyn: This is more of a question; do you know if the Batch client will handle batch requests correctly if there are more than 100 requests?

Otherwise LGTM.

frankyn (approved):

cbonnie:
Thanks for your work, Jesse!

PR #7737, opened 2021-12-07
bug: fix benchwrapper

zoercai:
The current Benchwrapper does not work/run. This PR syncs the protos used by the benchmarker tests and fixes issues that prevent the benchwrapper from running successfully.

These changes have been tested and run locally, and results show on the Perfgate dashboard successfully.

### DIFF
--- a/google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
+++ b/google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
@@ -19,7 +19,6 @@ package com.google.cloud.benchwrapper;
 import static com.google.cloud.spanner.TransactionRunner.TransactionCallable;
 
 import com.google.api.core.ApiFunction;
-import com.google.api.gax.core.NoCredentialsProvider;
 import com.google.cloud.benchwrapper.SpannerBenchWrapperGrpc.SpannerBenchWrapperImplBase;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
@@ -29,7 +28,6 @@ import com.google.cloud.spanner.ResultSet;
 import com.google.cloud.spanner.Spanner;
 import com.google.cloud.spanner.SpannerBatchUpdateException;
 import com.google.cloud.spanner.SpannerOptions;
-import com.google.cloud.spanner.SpannerOptions.CallCredentialsProvider;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.TransactionContext;
 import io.grpc.ManagedChannelBuilder;


google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -39,13 +41,16 @@
...
+    final SpannerOptions.Builder optionsBuilder = SpannerOptions.newBuilder();
+    final SpannerOptions options = optionsBuilder
+        .setEmulatorHost(spannerEmulatorHost)
+        .build();
     spanner = options.getService();
 
     dbClient =
         spanner.getDatabaseClient(
-            DatabaseId.of(options.getProjectId(), "test-instance", "test-db"));
+            DatabaseId.of(options.getProjectId(), "someinstance", "somedatabase"));
```
thiagotnunes, line 53: nit: I think we mostly use `test-instance` / `test-db` in other samples

google-cloud-testing/spanner-benchwrapper/src/main/proto/spanner.proto
```diff
@@ -18,48 +18,57 @@ package spanner_bench;
 option java_multiple_files = true;
 option java_package = "com.google.cloud.benchwrapper";
 
-message ReadQuery{
-  // The query to use in the read call.
-  string Query = 1;
+message Singer {
```
thiagotnunes, line 21: Not sure why you had to update the protos here, so I won't provide feedback on this part of the code

thiagotnunes (approved):

google-cloud-testing/spanner-benchwrapper/src/main/proto/spanner.proto
zoercai: It's because I changed the protos on google3 that are used by the benchmarker tests to use with the new/real emulator, so these ones need to match. Album/Singer are the same as what we use in examples too.

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
zoercai: I might leave this as is for now if that's okay, the other existing client library benchwrappers are all using `someinstance`/`somedatabase` already, so would need to change them all and the google3 scripts if I changed these here.

zoercai:
@googleapis/yoshi-java could anyone review this PR?

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
```diff
@@ -20,7 +20,7 @@
 import io.grpc.netty.NettyServerBuilder;
 import java.util.Properties;
 
-class Main {
+public class Main {
```
olavloite, line 23: nit: I don't think you need to make this a public class.

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -19,6 +19,7 @@
 import static com.google.cloud.spanner.TransactionRunner.TransactionCallable;
 
 import com.google.api.core.ApiFunction;
+import com.google.api.gax.core.NoCredentialsProvider;
```
olavloite, line 22: nit: this import does not seem to be used

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/SpannerBenchWrapperImpl.java
```diff
@@ -28,6 +29,7 @@
 import com.google.cloud.spanner.Spanner;
 import com.google.cloud.spanner.SpannerBatchUpdateException;
 import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.SpannerOptions.CallCredentialsProvider;
```
olavloite, line 32: nit: this import does not seem to be used

google-cloud-testing/spanner-benchwrapper/src/main/proto/spanner.proto
```diff
@@ -18,48 +18,57 @@ package spanner_bench;
...
 }
 
-message EmptyResponse{
+message UpdateQuery {
+  // The queries to use in the update call.
+  repeated string queries = 1;
 }
 
-service SpannerBenchWrapper{
+// TODO(deklerk): Replace with Google's canonical Empty.
```
olavloite, line 50: nit: why is this TODO added here? Is there any reason that we cannot use the canonical Empty here that is mentioned?

olavloite (approved):

google-cloud-testing/spanner-benchwrapper/src/main/proto/spanner.proto
zoercai: Oops sorry I copied this from an internal proto file as is. I've deleted this comment for now. Since this proto needs to sync with the internal proto, which is also used for the other client libraries, I think I'll leave the `EmptyResponse` here as it is for now so that I don't break everything. I can make another fix in the future for changing to Empty.

google-cloud-testing/spanner-benchwrapper/src/main/java/com/google/cloud/benchwrapper/Main.java
zoercai: The lack of `public` is actually causing the benchmarking tests to fail according to the test logs:

```
java.lang.Exception: The specified mainClass doesn't contain a main method with appropriate signature.
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:258)
    at java.lang.Thread.run (Thread.java:748)
Caused by: java.lang.IllegalAccessException: symbolic reference class is not public: class com.google.cloud.benchwrapper.Main, from org.codehaus.mojo.exec.ExecJavaMojo$1
    at java.lang.invoke.MemberName.makeAccessException (MemberName.java:850)
    at java.lang.invoke.MethodHandles$Lookup.checkSymbolicClass (MethodHandles.java:1402)
    at java.lang.invoke.MethodHandles$Lookup.resolveOrFail (MethodHandles.java:1382)
    at java.lang.invoke.MethodHandles$Lookup.findStatic (MethodHandles.java:780)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:251)
    at java.lang.Thread.run (Thread.java:748)
```

Let me know if there's a better way to fix this :-)

olavloite: Ah, ok. No, the easiest way is probably to just make it public.

(Note: This is not a Java requirement, but apparently a requirement for the Maven exec plugin. Java allows you to define the `main` method in a package-private class.)

google-cloud-testing/spanner-benchwrapper/src/main/proto/spanner.proto
olavloite: That sounds reasonable.

PR #7753, opened 2021-12-30
samples(storage): Add samples for stream upload and download

JesseLovelace:
Samples for read channel and write channel.

Internal bug 168766907

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectDownload.java
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectDownload.java
@@ -29,11 +29,13 @@ public class StreamObjectDownload {
 
     // The path to the file to download the object to
     // String targetFile = "path/to/your/file";
+    Path targetFilePath = Paths.get(targetFile);
 
     Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
-    try (ReadChannel reader = storage.reader(BlobId.of(bucketName, objectName))) {
-      Path targetFilePath = Paths.get(targetFile);
-      FileChannel targetFileChannel = FileChannel.open(targetFilePath, StandardOpenOption.WRITE);
+    try (ReadChannel reader = storage.reader(BlobId.of(bucketName, objectName));
+        FileChannel targetFileChannel =
+            FileChannel.open(targetFilePath, StandardOpenOption.WRITE)) {
+
       ByteStreams.copy(reader, targetFileChannel);
 
       System.out.println(


google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectDownload.java
```diff
@@ -0,0 +1,49 @@
+package com.google.cloud.examples.storage.objects;
+
+import com.google.cloud.ReadChannel;
```
BenWhitehead, line 3: Missing region tags

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectDownload.java
```diff
@@ -0,0 +1,49 @@
...
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+    try (ReadChannel reader = storage.reader(BlobId.of(bucketName, objectName))) {
+      ByteBuffer buffer = ByteBuffer.allocate(64 * 1024);
+      while (reader.read(buffer) > 0) {
+        buffer.flip();
+        output.write(buffer.array(), 0, buffer.limit());
+        buffer.clear();
+      }
+      output.writeTo(new FileOutputStream(targetFile));
```
BenWhitehead, line 37: I think we can simplify this by directly creating a `FileChannel` to output to rather than doing the intermediate streams and manual buffering (We also use the guava `ByteStreams.copy` utility method which takes care of all the edgecases of using different channel types and ByteBuffers.

```suggestion
    try (ReadChannel reader = storage.reader(BlobId.of(bucketName, objectName))) {
      Path targetFilePath = Paths.get(targetFile);
      FileChannel targetFileChannel = FileChannel.open(targetFilePath, StandardOpenOption.WRITE);
      ByteStreams.copy(reader, targetFileChannel);
```

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectDownload.java
```diff
@@ -0,0 +1,49 @@
...
+    // The path to the file to download the object to
+    // String targetFile = "path/to/your/file";
+
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+    try (ReadChannel reader = storage.reader(BlobId.of(bucketName, objectName))) {
+      ByteBuffer buffer = ByteBuffer.allocate(64 * 1024);
+      while (reader.read(buffer) > 0) {
+        buffer.flip();
+        output.write(buffer.array(), 0, buffer.limit());
```
BenWhitehead, line 34: Not all `ByteBuffers` have a backing array that can be returned here, so this code sample wouldn't serve as a great example for many cases (especially those where a ByteBuffer might be passed from another library or framework).

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectUpload.java
```diff
@@ -0,0 +1,37 @@
+package com.google.cloud.examples.storage.objects;
+
+import com.google.cloud.WriteChannel;
```
BenWhitehead, line 3: Missing region tags

BenWhitehead (changes requested):
In this case am I correct to assume that "Stream" is supposed to mean "Resumable Media" and not "Java IO Streams"?

JesseLovelace:
>In this case am I correct to assume that "Stream" is supposed to mean "Resumable Media" and not "Java IO Streams"?
It's for the "[Streaming transfers](https://cloud.google.com/storage/docs/streaming#code-samples)" page. I think read channel and write channel aren't the only ways to do this but it seemed appropriate to me, plus they aren't shown in other samples, but open to suggestions.

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectDownload.java
sydney-munro: Good suggestion thanks!

google-cloud-examples/src/main/java/com/google/cloud/examples/storage/objects/StreamObjectDownload.java
```diff
@@ -0,0 +1,50 @@
...
+    // The ID of your GCS object
+    // String objectName = "your-object-name";
+
+    // The path to the file to download the object to
+    // String targetFile = "path/to/your/file";
+
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+    try (ReadChannel reader = storage.reader(BlobId.of(bucketName, objectName))) {
+      Path targetFilePath = Paths.get(targetFile);
+      FileChannel targetFileChannel = FileChannel.open(targetFilePath, StandardOpenOption.WRITE);
```
BenWhitehead, line 36: One more improvement I missed last time.

We can include the FileChannel in the try-with-resource so that it's automatically closed along with the ReadChannel when leaving the block.

```suggestion
    Path targetFilePath = Paths.get(targetFile);

    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
    try (ReadChannel reader = storage.reader(BlobId.of(bucketName, objectName));
        FileChannel targetFileChannel =
            FileChannel.open(targetFilePath, StandardOpenOption.WRITE)) {

```

BenWhitehead (approved):
Overall looks good, one small improvement I've proposed.

Feel free to merge after reviewing and accepting it.

PR #7799, opened 2022-02-15
fix(deprecate): removed older version of Compute client library samples

Sita04:
Deprecating older version of Compute client library samples.
New Compute samples can be found [here](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main/compute/cloud-client).

Ref: https://github.com/googleapis/google-cloud-java/pull/7742#issuecomment-1039631178

### DIFF
--- a/google-cloud-examples/src/main/java/com/google/cloud/examples/compute/README.md
+++ b/google-cloud-examples/src/main/java/com/google/cloud/examples/compute/README.md
@@ -1,3 +1,3 @@
 [**DEPRECATED**]
 
-The GAPIC compute samples have been moved to [Java-docs-samples repository](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main/compute/cloud-client).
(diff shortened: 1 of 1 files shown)


suztomo (approved):

google-cloud-examples/src/main/java/com/google/cloud/examples/compute/README.md
```diff
@@ -0,0 +1,3 @@
+[**DEPRECATED**]
+
+The GAPIC compute samples have been moved to [Java-docs-samples repository](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main/compute/cloud-client).
```
suztomo, line 3: This is a good note. But "GAPIC" is not good term for customers.

Sita04: Thanks. Modified it.

PR #7922, opened 2022-06-14
chore: add IT test configuration

Neenu1995:
cc: @meltsufin @suztomo

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -19,7 +19,6 @@ git clone https://github.com/newren/git-filter-repo.git
 export PATH=$PATH:`pwd`/git-filter-repo
 
 mkdir google-cloud-java
-cp -r ../templates/.kokoro google-cloud-java
 
 cd google-cloud-java
 git init -b main
@@ -48,6 +47,10 @@ cd google-cloud-java
 
 git add pom.xml
 git commit -am 'feat: create aggregator pom'
+
+cp -r ../../templates/.kokoro ./
+git add .kokoro
+git commit -am 'chore: add kokoro config files'
 
 # generate coverage report
 mkdir CoverageAggregator


meltsufin (approved):

bootstrap.sh
```diff
@@ -19,6 +19,8 @@ git clone https://github.com/newren/git-filter-repo.git
 export PATH=$PATH:`pwd`/git-filter-repo
 
 mkdir google-cloud-java
+cp -r ../templates/.kokoro google-cloud-java
```
meltsufin, line 22: Can we do it as a separate commit after line 45?

templates/.kokoro/build.sh
```diff
@@ -0,0 +1,115 @@
...
+
+set -eo pipefail
+
+## Get the directory of the build script
+scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+## cd to the parent directory, i.e. the root of the git repo
+cd ${scriptDir}/..
+
+# include common functions
+source ${scriptDir}/common.sh
```
meltsufin, line 24: I don't think this file exists.

templates/.kokoro/build.sh
```diff
@@ -0,0 +1,115 @@
...
+
+# Print out Maven & Java version
+mvn -version
+echo ${JOB_TYPE}
+
+mvn -pl -java-container clean install
```
meltsufin, line 30: Is this some kind of tests remnant?

bootstrap.sh
Neenu1995: fixed

templates/.kokoro/build.sh
Neenu1995: Fixed

templates/.kokoro/build.sh
Neenu1995: java container is a snowflake which uses JDST project for its IT tests. It will not work with gcloud-devel.
Until we figure out a way to switch project based on module or use a different project, java container is excluded from the IT tests.

meltsufin (approved):

PR #8025, opened 2022-07-19
feat: Add Cloud RAD kokoro script

lqiu96:


templates/.kokoro/release/publish_javadoc11.cfg
```diff
@@ -0,0 +1,30 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# cloud-rad production
+env_vars: {
+  key: "STAGING_BUCKET_V2"
+  value: "docs-staging-v2-dev"
```
lqiu96, line 6: Using dev bucket

templates/.kokoro/release/publish_javadoc11.sh
```diff
@@ -0,0 +1,83 @@
...
+  if [[ ! "${excluded_modules[*]}" =~ $module ]]; then
+    cd $module
+    # Extract Cloud RAD module name
+    NAME=$(jq -r '.distribution_name' .repo-metadata.json | cut -d ':' -f 2)
+    # Extract (current) version from versions.txt and remove `-SNAPSHOT`
+    VERSION=$(grep ${NAME}: versions.txt | cut -d: -f3 | sed -e 's/-SNAPSHOT//g')
+    echo "Running for ${NAME}-${VERSION}"
+
+    # cloud RAD generation
+    mvn clean javadoc:aggregate -B -P docFX-pipelineTest
```
lqiu96, line 56: Currently using a custom profile in the aggregator pom.xml (+ all submodules inherit this pom), but need to see if the plugin argument will work

templates/.kokoro/release/publish_javadoc11.cfg
meltsufin: Can you add a comment in the file itself to remind us what to change it to for prod?

templates/.kokoro/release/publish_javadoc11.sh
meltsufin: Do you need to update the aggregator pom in this branch too to include this `docFX-pipelineTest` profile?

templates/.kokoro/release/publish_javadoc11.sh
```diff
@@ -0,0 +1,83 @@
...
+     --xrefs devsite://java/protobuf \
+     --language java
+
+    # upload yml to production bucket
+    python3 -m docuploader upload . \
+     --credentials ${CREDENTIALS} \
+     --staging-bucket ${STAGING_BUCKET_V2} \
+     --destination-prefix docfx
+
+    cd ../../..
```
suztomo, line 81:
```suggestion
    popd # out of target/docfx-yml
    popd # out of $module
```

templates/.kokoro/release/publish_javadoc11.sh
```diff
@@ -0,0 +1,83 @@
...
+    # Extract (current) version from versions.txt and remove `-SNAPSHOT`
+    VERSION=$(grep ${NAME}: versions.txt | cut -d: -f3 | sed -e 's/-SNAPSHOT//g')
+    echo "Running for ${NAME}-${VERSION}"
+
+    # cloud RAD generation
+    mvn clean javadoc:aggregate -B -P docFX-pipelineTest
+    # include CHANGELOG
+    cp CHANGELOG.md target/docfx-yml/history.md
+
+    cd target/docfx-yml
```
suztomo, line 60:
```suggestion
    pushd target/docfx-yml
```

templates/.kokoro/release/publish_javadoc11.sh
```diff
@@ -0,0 +1,83 @@
...
+
+# Retrieve list of modules from aggregator pom
+modules=$(mvn help:evaluate -Dexpression=project.modules | grep '<.*>.*</.*>' | sed -e 's/<.*>\(.*\)<\/.*>/\1/g')
+excluded_modules=('CoverageAggregator' 'google-cloud-gapic-bom')
+
+for module in $modules
+do
+  # Proceed if module is not excluded
+  if [[ ! "${excluded_modules[*]}" =~ $module ]]; then
+    cd $module
```
suztomo, line 48:
```suggestion
    pushd $module
```

suztomo (approved):
pushd / popd is convenient.

templates/.kokoro/release/common.cfg
```diff
@@ -0,0 +1,49 @@
...
+# Download trampoline resources.
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
+
+# Use the trampoline script to run in docker.
+build_file: "google-cloud-java/.kokoro/trampoline.sh"
+
+# Configure the docker image for kokoro-trampoline.
+env_vars: {
+  key: "TRAMPOLINE_IMAGE"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
```
suztomo, line 12: Can you use 11?

lqiu96:
@suztomo @meltsufin I don't believe we can override a maven plugin property if it's already been set in the pom.xml (https://issues.apache.org/jira/browse/MNG-4979). I haven't been able to override the docletPath except when commenting out the <docletPath> in the configuration section. 

Shared-Config: https://github.com/googleapis/java-shared-config/blob/c805926b297617452c5ca27062d26eca917149aa/pom.xml#L715

I can create a plugin property value in <properties> and override that value, but that would also require changes in the shared-config (having <docletPath> reference the new value in properties instead of ${env.KOKORO_GFILE_DIR}).

Would it make more sense to update the existing docFX profile or create a new profile with the updated values?

suztomo (approved):

PR #8041, opened 2022-07-25
ci: gitignore for template

suztomo:
- ci: adding .gitignore to template
- boostrap to force-include gitignore

### DIFF
--- a/templates/.gitignore
+++ b/templates/.gitignore
@@ -1,5 +1,3 @@
-.gitignore
-
 # Packages
 dist
 bin
@@ -15,9 +13,6 @@ nosetests.xml
 # Translations
 *.mo
 
-# Mr Developer
-.mr.developer.cfg
-
 .metadata
 .project
 .pydevproject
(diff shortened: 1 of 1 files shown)


ddixit14 (approved):

templates/.gitignore
```diff
@@ -0,0 +1,44 @@
...
+.classpath
+
+# Built documentation
+docs/
+
+
+# Wheel directory used in Travis builds.
+gcloud-java-wheels/
+src/test/resources/gcd-head.zip
+src/test/resources/gcd-v1beta2-rev1-2.1.1.zip
```
meltsufin, line 37: Can we make a new `.gitignore` with just the ignores that we know we need?

meltsufin (changes requested):

templates/.gitignore
suztomo: Sure. Let me update them.

suztomo: Removed the irrelevant entries.

meltsufin (approved):

PR #8106, opened 2022-08-10
feat: Determine IT modules to test by files changed

lqiu96:

PR #8110, opened 2022-08-11
feat: Run integration tests on modified modules

lqiu96:
Testing for this was done in https://github.com/googleapis/google-cloud-java/pull/8106

### DIFF
--- a/templates/.kokoro/build.sh
+++ b/templates/.kokoro/build.sh
@@ -23,19 +23,14 @@ cd ${scriptDir}/..
 # include common functions
 source ${scriptDir}/common.sh
 
-# TODO: Switch monorepo_script_output to main
-# Find the root commit between the two branches
-base_commit=$(git merge-base monorepo_script_output HEAD)
-echo "Base Commit: ${base_commit}"
-
 # Find the files changed from when the PR branched to the last commit
 # Filter for java modules and get all the unique elements
 # grep returns 1 (error code) and exits the pipeline if there is no match
 # If there is no match, it will return true so the rest of the commands can run
-modified_files=$(git diff --name-only "${base_commit}" HEAD)
-echo "Modified files: ${modified_files}"
+modified_files=$(git diff --name-only $KOKORO_GITHUB_PULL_REQUEST_COMMIT $KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH)
+printf "Modified files:\n%s\n" "${modified_files}"
 directories=$(echo "${modified_files}" | grep -e 'java-.*' || true)
-echo "Files in java modules: ${directories}"
+printf "Files in java modules:\n%s\n" "${directories}"
 if [[ -n $directories ]]; then
   directories=$(echo "${directories}" | cut -d '/' -f1 | sort -u)
   dir_list=()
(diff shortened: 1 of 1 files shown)


templates/.kokoro/build.sh
```diff
@@ -23,11 +23,44 @@ cd ${scriptDir}/..
...
+# TODO: Switch monorepo_script_output to main
+# Find the root commit between the two branches
+base_commit=$(git merge-base monorepo_script_output HEAD)
+echo "Base Commit: ${base_commit}"
+
+# Find the files changed from when the PR branched to the last commit
+# Filter for java modules and get all the unique elements
+# grep returns 1 (error code) and exits the pipeline if there is no match
+# If there is no match, it will return true so the rest of the commands can run
+directories=$(git diff --name-only "${base_commit}" HEAD | grep -e 'java-.*' || true)
```
lqiu96, line 35: `git diff` was used to compare only the changes between latest commit and $base_commit instead of all the files touched between these commits.

templates/.kokoro/build.sh
```diff
@@ -23,11 +23,44 @@ cd ${scriptDir}/..
 # include common functions
 source ${scriptDir}/common.sh
 
+# TODO: Switch monorepo_script_output to main
+# Find the root commit between the two branches
+base_commit=$(git merge-base monorepo_script_output HEAD)
```
lqiu96, line 28: Kokoro pipeline is in the PR branch, so HEAD so point to the latest commit in this branch. Should this compared with the branch it is trying to be merged into instead of always `main`/`monorepo_script_output`?

templates/.kokoro/build.sh
```diff
@@ -56,21 +89,30 @@ integration)
...
-    -DskipUnitTests=true \
-    -fae \
-    -T 1C \
-    verify
-  RETURN_CODE=$?
+  if [[ -n $module_list ]]
+  then
+    echo "Running Integration Tests for: ${module_list}"
+    mvn -B -pl "${module_list}" \
+        -amd \
```
lqiu96, line 96: --also-make-dependencies to build the submodules. This also brings in CoverageAggregator (has no ITs)

templates/.kokoro/build.sh
suztomo: Hardcoding monorepo_script_output is fine for now.

suztomo (approved):
Added some comments. Overall it looks great.

templates/.kokoro/build.sh
```diff
@@ -23,11 +23,41 @@ cd ${scriptDir}/..
 # include common functions
 source ${scriptDir}/common.sh
 
+# Find the files changed from when the PR branched to the last commit
+# Filter for java modules and get all the unique elements
+# grep returns 1 (error code) and exits the pipeline if there is no match
+# If there is no match, it will return true so the rest of the commands can run
+modified_files=$(git diff --name-only $KOKORO_GITHUB_PULL_REQUEST_COMMIT $KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH)
```
lqiu96, line 30: Use `git diff` to find what changed between the two branches. No longer tries to find the `Base Commit` since the `Target Branch` could have added commits that have been committed in.

suztomo (approved):

templates/.kokoro/build.sh
suztomo: Memo: Later, we may want to run integration tests as nightly builds.

meltsufin: I think this works as long as there are no interdependencies among the modules, which I think is the case with pure GAPICs. However, it might not work in the general case of multi-module Maven projects.

PR #8155, opened 2022-08-18
fix: config to match schema

chingor13:


ddixit14 (approved):

meltsufin:
@ddixit14 Does this need to be updated in the `monorepo_script` branch?

PR #8185, opened 2022-08-24
ci: adding a check that only uses 3 repositories

suztomo:
I got an inspiration from talk with Mike.

If something is wrong with shell scripts, running it in few repositories probably detects the problem faster than the "build" job, which now takes 1 hour 47 minutes.

This "build-3repos" is not meant to replace the existing "build" job.

### DIFF
--- a/.github/workflows/monorepo.yaml
+++ b/.github/workflows/monorepo.yaml
@@ -31,18 +31,16 @@ jobs:
           git remote add monorepo https://${{ github.actor }}:${{ github.token }}@github.com/${{ github.repository }}.git
           git push -f monorepo main:monorepo_script_output
 
-  build-5repos:
+  build-3repos:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


meltsufin (approved):
Only 3 minutes to run -- cool!

.github/workflows/monorepo.yaml
```diff
@@ -29,4 +29,27 @@ jobs:
...
+  build-3repos:
+    runs-on: ubuntu-latest
+    name: Running bootstrap.sh with only 3 selected repositories
+    steps:
+    - uses: actions/checkout@v3
+    - name: Truncate repos.txt for 3 repositories
+      run: |
+        cat << EOL > repos.txt
+        java-accessapproval
+        java-dataform
```
suztomo, line 43: In your pull request, replace java-dataform with the module that somehow continues to get SNAPSHOT version.

PR #8198, opened 2022-08-24
separate release-please script

ddixit14:


templates/generation/release_please_config_generation.sh
```diff
@@ -0,0 +1,102 @@
+
```
meltsufin, line 1: First line should be `#!/bin/bash`.

suztomo: Sorry, I was demonstrating to Deepankar how to use "gh" command using this PR. It's not ready for review.

PR #8206, opened 2022-08-25
chore: Separate release please script

ddixit14:
This PR makes release-please config generation separate from bootstrap.sh
Also fixes the error where modules which did not have a bom get picked up for config generation. 
 
NOTE: This is just a modified version of release-please script which we need to use for manually staging the modules. Once release-please starts working totally fine, this will be changed again.


templates/generation/release_please_config_generation.sh
```diff
@@ -0,0 +1,171 @@
...
+
+set -e
+
+echo "{" > .release-please-manifest.json
+
+echo ""\""google-cloud-gapic-bom"\"": "\""0.0.0"\""," >> .release-please-manifest.json
+
+GENERATION_DIR=$(dirname -- "$0");
+
+num_modules="$(wc -l < ../../repos.txt)"
```
suztomo, line 11: Don't rely on repos.txt

templates/generation/release_please_config_generation.sh
```diff
@@ -0,0 +1,171 @@
...
+
+rp_config_line=""
+rp_manifest_line=""
+
+for bom_directory in $(find . -name 'google-*-bom' | sort); do
```
suztomo, line 16: Does this (Release Please config) really need to work based on "-bom" directory? To me, it just need to find versions.txt, regardless of "-bom" directories.

You have the second loop which does not rely on "-bom" directory.

templates/generation/release_please_config_generation.sh
ddixit14: What should I rely on, then? As a source of truth to calculate the total number of modules in the config files?

suztomo: The number of the modules.

bootstrap.sh
```diff
@@ -101,103 +101,7 @@ echo ""\""google-cloud-gapic-bom"\"": "\""0.0.0"\""," >> .release-please-manifes
...
-
-  #adding the line to manifest config file
-  echo "${rp_manifest_line}" >> .release-please-manifest.json
-done
-
-echo "}" >> .release-please-manifest.json
-
-awk -v "packagesList=$rp_config_line" '{gsub(/ALL_PACKAGES/,packagesList)}1' \
-    ../../release_please_config_raw.json > release-please-config.json
+bash generation/release_please_config_generation.sh
```
suztomo, line 104: Can you rename the script as "generate_release_please_config.sh"?

I put the verb first in other scripts.

templates/generation/release_please_config_generation.sh
ddixit14: This is not just generating release-please config. It is also managing versions as we want them. Sometimes it is according to release-please functioning, and right now it is according to upload to staging directory process. 
Also, there are different kind of modules (some having more modules than others), and hence their versions.txt are built differently. 
I'm writing the logic differently now. Earlier assumption that this loop covered all the modules proved to be incorrect, the main challenge here is making the versions sync.

suztomo:
Can you figure out how to stop generating the backup file?

Screen Shot 2022-08-25 at 12 09 31 AM

templates/generation/release_please_config_generation.sh
suztomo: > I'm writing the logic differently now. 

Good. Looking forward. Consider creating a loop by finding all "versions.txt" files.

ddixit14:
> Can you figure out how to stop generating (or adding) the backup files?
> 
> Screen Shot 2022-08-25 at 12 09 31 AM

Fixed.

bootstrap.sh
ddixit14: Done.

templates/generation/release_please_config_generation.sh
ddixit14: As discussed, this script will loop over the modules, generate release-please config files and sync the versions. 
However, there is one category of modules which might not get updated -> single component modules (like java-dns, java-notification, java-grafeas) with both current and released versions as non-snapshot in versions.txt. 
I can fix it later if it poses issues for us.

templates/generation/release_please_config_generation.sh
ddixit14: Done.

suztomo (approved):

templates/generation/generate_release_please_config.sh
```diff
@@ -0,0 +1,93 @@
...
+      do
+         if [[ ${line} =~ [0-9] ]] && ! [[ ${line} == *"SNAPSHOT"* ]] ; then
+           artifact_name=$(echo "${line}" | awk -F':' '{print $1}')
+           old_version=$(echo "${line}" | awk -F':' '{print $3}')
+           new_version=$(echo ${old_version} |  awk -F'.' '{print $1"."$2"."$3+1}' |  sed s/[.]$//)
+           new_version="${new_version}-SNAPSHOT"
+           sed -i "s|${artifact_name}:${old_version}:${old_version}|${artifact_name}:${old_version}:${new_version}|" ${version_file}
+           artifact_directory="${module}/${artifact_name}"
+           mvn -B -ntp -f ${artifact_directory} -U versions:set -DnewVersion=${new_version}
+           mvn versions:commit
```
suztomo, line 65: This runs against all projects, making the runtime unnecessary longer.

ddixit14: replaced it by adding " -DgenerateBackupPoms=false " in the above mvn versions:set command.

PR #8260, opened 2022-08-31
chore: Fix flaky IT test cases

lqiu96:
This PR should resolve a bunch of the flaky test cases in the monorepo.

Status:
image

Quota Metrics after VM cleanup (7 Day Graph):
image


.kokoro/build.sh
```diff
@@ -156,7 +156,7 @@ case ${JOB_TYPE} in
           -Penable-integration-tests \
           -Pnative \
           -fae \
-          verify
+          test
```
meltsufin, line 159: Would this skip ITs now?

java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/ITSmokeInstancesTest.java
```diff
@@ -72,6 +72,8 @@ public static void setUp() throws IOException {
     instances = new ArrayList<>();
     InstancesSettings instanceSettings = InstancesSettings.newBuilder().build();
     instancesClient = InstancesClient.create(instanceSettings);
+
+    Util.cleanUpComputeInstances(instancesClient, DEFAULT_PROJECT, DEFAULT_ZONE);
```
meltsufin, line 76: Isn't this a bit too aggressive to delete all compute instances? Why isn't the cleanup in `tearDown()` sufficient?

java-container/google-cloud-container/src/test/java/com/google/cloud/container/v1/it/ITSystemTest.java
```diff
@@ -90,7 +91,13 @@ public static void beforeClass() throws Exception {
 
   @AfterClass
   public static void afterClass() throws Exception {
-    Thread.sleep(TimeUnit.MINUTES.toMillis(5));
+    Operation response = client.getOperation(PROJECT_ID, ZONE, operation.getName());
+    // Sleep for one minute until Cluster CREATE operation is complete
```
meltsufin, line 95: Why are we waiting for cluster creation completion in `afterClass()`? Shouldn't we wait for it right after creation in `beforeClass()`?

java-notebooks/google-cloud-notebooks/src/test/java/com/google/cloud/notebooks/v1beta1/it/ITNotebookServiceClientTest.java
```diff
@@ -66,6 +66,8 @@ public class ITNotebookServiceClientTest {
   public static void setUp() throws IOException, ExecutionException, InterruptedException {
     // Create Test Notebook Instance
     client = NotebookServiceClient.create();
+    Util.cleanUpNotebookInstances(client, PARENT);
```
meltsufin, line 69: Wouldn't this delete the one created above before it's even used?
Also, why isn't cleanup in `tearDown()` sufficient?

meltsufin (changes requested):

.kokoro/build.sh
lqiu96: GraalVM checks do not run the failsafe (IT) plugin since the native plugin is compatible to run with it. We set the surefire plugin to use the IT prefix (https://github.com/googleapis/java-shared-config/blob/8d7f75de442cb3a274004faae430e9edf60730f3/pom.xml#L806). 

Example in the split-repos (Java-Vision): https://github.com/googleapis/java-vision/blob/359492111a9b5d20de04e4e98ecb6374e63cfb9d/.kokoro/build.sh#L74

java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/ITSmokeInstancesTest.java
lqiu96: tearDown() cleanup is sufficient for 99% of the case. There was a significant amount of compute instances that were in there (my guess from previous runs that didn't have the cleanup or when the IT jobs were cancelled halfway). 

This project is only used for ITs so it would have 24 hours to report the results before the previous instances are deleted.

java-container/google-cloud-container/src/test/java/com/google/cloud/container/v1/it/ITSystemTest.java
lqiu96: I was able to run operations of the client (get/ list) without the operation having to be complete. It was only when deleting the cluster, I got an error with `Operation in progress`. I think it would make sense to move this to wait when the operation was run.

I'll change this to to do the busy wait in beforeClass().

java-notebooks/google-cloud-notebooks/src/test/java/com/google/cloud/notebooks/v1beta1/it/ITNotebookServiceClientTest.java
lqiu96: It would only delete notebooks that have a creation time more than 24 hours earlier.

java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/ITSmokeInstancesTest.java
meltsufin: I'm worried that someone runs the test on their own project and causes inadvertent deletion of instances. If we want to do this kind of cleanup, maybe we can have instance name prefix that is consistent and can accurately filter instances created by this test.

java-notebooks/google-cloud-notebooks/src/test/java/com/google/cloud/notebooks/v1beta1/it/ITNotebookServiceClientTest.java
meltsufin: Can we have a prefix for the notebook names or something to avoid accidentally deleting unrelated notebooks.

java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/ITSmokeInstancesTest.java
lqiu96: > I'm worried that someone runs the test on their own project and causes inadvertent deletion of instances.

Actually, I was thinking that this would be fine for the ITs. I was thinking the ideal state for gcloud-devel would have 0 compute instances when there are no active ITs running. If `Util.cleanUpComputeInstances` (or more filtered version of that function) matches any existing instances, then those should be deleted. 24 hours is 12x as long as the longest GraalVM job.

> maybe we can have instance name prefix that is consistent and can accurately filter instances created by this test. 

Yep! I was planning on creating another PR that would update the instances for each module to have a more unique name instead `test-instance` or `gapic-instance`. i.e. Compute would be test-compute-instance-UUID, Container would be test-container-instance-UUID or something like that.

lqiu96: Ok nvm, I just re-read the comment. 

>  test on their own project 
That makes more sense.

java-notebooks/google-cloud-notebooks/src/test/java/com/google/cloud/notebooks/v1beta1/it/ITNotebookServiceClientTest.java
lqiu96: Yep, it's similar to the comment above about compute and container. I'll add in more specific names for the IT instances.

lqiu96:
@meltsufin @suztomo Could you re-review this PR?

java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/Util.java
```diff
@@ -0,0 +1,37 @@
...
+  private static final int DELETION_THRESHOLD_TIME_HOURS = 24;
+
+  /** Bring down any instances that are older than 24 hours */
+  public static void cleanUpComputeInstances(
+      InstancesClient instancesClient, String project, String zone) {
+    ListPagedResponse listPagedResponse = instancesClient.list(project, zone);
+    for (Instance instance : listPagedResponse.iterateAll()) {
+      if (isCreatedBeforeThresholdTime(
+              ZonedDateTime.parse(instance.getCreationTimestamp()).toInstant())
+          && instance.getName().startsWith(BaseTest.COMPUTE_PREFIX)) {
```
meltsufin, line 23: For both `Util` classes, I think I would rather have the prefix passed in as an argument into this method, but I guess this is OK too.

meltsufin (approved):

java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/Util.java
lqiu96: On second thought, this seems better. I'll create a new PR for this.

PR #8343, opened 2022-09-09
chore: GraalVM Job tests a few select modules

lqiu96:
GraalVM jobs seem to be exhausting all the resources. Each job requires 20 sub jobs, which needs 42 jobs (20 + 20 sub jobs, 2 parent jobs) to run both GraalVM tests every PR. 

This should release a bunch of jobs back to the pool. Only runs GraalVM jobs on modules that have been modified and quickly exits for the rest of the sub jobs.

```
[09/09 11:04:49] 503 Service Unavailable
{
  "code": 503,
  "errors": [
    {
      "domain": "global",
      "message": "The service is currently unavailable.",
      "reason": "backendError",
      "debugInfo": "detail: \"[ORIGINAL ERROR] throttling::THROTTLED_TASK_LIMIT: Request \\'/ZoneOperationsService.Get\\' from role \\'gce-control-esf-proxy\\' throttled: Task is overloaded (in-flight-requests) go/tr-t.\"\n"
    }
  ],
  "message": "The service is currently unavailable.",
  "status": "UNAVAILABLE",
  "details": [
    {
      "@type": "type.googleapis.com/google.rpc.DebugInfo",
      "detail": "[ORIGINAL ERROR] throttling::THROTTLED_TASK_LIMIT: Request '/ZoneOperationsService.Get' from role 'gce-control-esf-proxy' throttled: Task is overloaded (in-flight-requests) go/tr-t.&quot;
    }
  ]
}
```

Create a nightly build that runs at midnight PST to run all modules. This will test AI-Platform and Compute modules to achieve high coverage for GraalVM.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -50,7 +50,7 @@ case ${JOB_TYPE} in
       install_modules
       printf "Running Integration Tests for:\n%s\n" "${module_list}"
       mvn -B ${INTEGRATION_TEST_ARGS} \
-        -pl "${module_list}" \
+        -pl "${module_list},!CoverageAggregator" \
         -amd \
         -ntp \
         -Penable-integration-tests \


suztomo:
Can you create an independent shell script for GraalVM integration tests? (It may be called from .kokoro/build.sh)

.kokoro/build.sh is a script that are intended to be templated across many repositories and it does not have to be in that form any more in this monorepo.

suztomo:
Also, ensure we have a way to run every modules' GraalVM tests. I want to use it at the time of a release.

lqiu96:
> Can you create an independent shell script for GraalVM integration tests? (It may be called from .kokoro/build.sh)
> 
> .kokoro/build.sh is a script that are intended to be templated across many repositories and it does not have to be in that form any more in this monorepo.

I can move some of the logic to common.sh and clean up the build.sh file

> Also, ensure we have a way to run every modules' GraalVM tests. I want to use it at the time of a release.

@suztomo Are the GraalVM jobs currently running before the release process in split repos? Looking at ai-platform split repos (https://github.com/googleapis/java-aiplatform/tree/main/.kokoro/release), it doesn't seem like it runs the tests (except on presubmit).

Would it make more sense to run this on a continuous build (i.e. once a night and runs all modules) instead so it doesn't add 2+ hours to the release process?

suztomo:
What is a good way for the Kokoro build script to tell whether it's a pull request check or nightly builds? I want flexibility to run all tests in one action.

lqiu96:
For the jobs, I can pass a new environment variable (i.e something like TEST_ALL_MODULES) that will just run the ITs/ graalvm checks for every module. I can set the default values for the jobs: nightly/ continuous builds be true vs presubmits be false.

We can also override the environment variable in fusion by setting it ourselves. Should be easy to setup/ include the logic in build.sh.

lqiu96:
> What is a good way for the Kokoro build script to tell whether it's a pull request check or nightly builds? I want flexibility to run all tests in one action.

Oh I just realized I didn't answer you question. I can just change the the title of the kokoro job (i.e. prefix the nightly job with Continuous: GraalVM job or Nightly: GraalVM job).

lqiu96:
@suztomo @meltsufin The Compute module is going to hold up approvals:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Google Compute Engine Parent 1.12.1-SNAPSHOT:
[INFO] 
[INFO] Google Compute Engine Parent ....................... SUCCESS [  1.305 s]
[INFO] proto-google-cloud-compute-v1 ...................... SUCCESS [  1.161 s]
[INFO] Google Compute Engine .............................. SUCCESS [46:55 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  47:03 min
[INFO] Finished at: 2022-09-14T20:54:54Z
[INFO] ------------------------------------------------------------------------
Finished Unit and Integration Tests for GraalVM:
java-compute
```

If we mark the GraalVM jobs as required, then it'll be ~45 min before approval. I'm going to see if I can try and speed up the testing for the test cases (compilation is stuck at ~10-12 min)

lqiu96:
Runtime has been reduced to ~25 min for the testing with removing the blocking `get()` calls at the tearDown() stage.

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Google Compute Engine Parent 1.12.1-SNAPSHOT:
[INFO] 
[INFO] Google Compute Engine Parent ....................... SUCCESS [  1.293 s]
[INFO] proto-google-cloud-compute-v1 ...................... SUCCESS [  1.019 s]
[INFO] Google Compute Engine .............................. SUCCESS [26:11 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  26:19 min
[INFO] Finished at: 2022-09-15T18:25:19Z
[INFO] ------------------------------------------------------------------------
Finished Unit and Integration Tests for GraalVM:
java-compute
```

We can rely on the `instancesClient.deleteAsync()` and `addressesClient.deleteAsync()` function calls to be executed properly by the `tearDown()`. There is added `cleanUpComputeAddresses()` and `cleanUpComputeInstances()` functions in the Util class to clean up any addresses and instances that were not deleted (and meet the criteria: Matching prefix name and creation time is over 24 hours).

GraalVM Execution Times:
AI-Platform: 25 min
Compute: 35 min
Full ITs run (only on aggregator pom.xml change): 40 min

lqiu96:
https://github.com/graalvm/native-build-tools/issues/76 -- Thanks @mpeddada1 for linking me to this issue.

Seems like there was a proposal to avoid having to run the tests in JVM mode prior to being executed to native tests. Hasn't been any activity in that thread for over a year, so I think we're still going to be stuck running the the tests twice (once in JVM and once natively).

.kokoro/common.sh
```diff
@@ -52,7 +54,105 @@ function retry_with_backoff {
...
+    else
+      echo "Found no changes in the java modules"
+    fi
+  fi
+}
+
+function assign_modules_to_job() {
+  modules_assigned_list=()
+  num=0
+  for module in "${modified_module_list[@]}"; do
```
lqiu96, line 98: $modified_module_list will be an array of all the modules in pom if $TEST_ALL_MODULES is set to true (only on the nightly build)

.kokoro/common.sh
```diff
@@ -52,7 +54,105 @@ function retry_with_backoff {
...
+    -ntp \
+    -DtrimStackTrace=false \
+    -Dclirr.skip=true \
+    -Denforcer.skip=true \
+    -Dcheckstyle.skip=true \
+    -Dflatten.skip=true \
+    -Danimal.sniffer.skip=true \
+    -Pnative \
+    -fae \
+    test
```
lqiu96, line 122: Remove the -Penable-integration-tests since it only sets the skipITs user property to false (https://github.com/googleapis/java-shared-config/blob/abf28e47a74f94bc814e6ff21c7f832dce98c2eb/pom.xml#L682) which has no relation to the surefire plugin. May need to revisit this if graalvm supports integration tests with the failsafe plugin

.kokoro/common.sh
```diff
@@ -52,7 +54,105 @@ function retry_with_backoff {
...
+    # Combine each entry with a comma
+    module_list=$(
+      IFS=,
+      echo "${modules_assigned_list[*]}"
+    )
+    printf "Module list is:\n%s\n" "${module_list}"
+  else
+    module_list="${MAVEN_MODULES}"
+    printf "Running the GraalVM test on pre-selected maven modules: %s\n" "${MAVEN_MODULES}"
+  fi
```
lqiu96, line 141: Only need to call generate_modified_modules_list and assign_modules_to_job only nightly build (TEST_ALL_MODULES is set to true). Otherwise the env_var MAVEN_MODULE will specify if it's java-compute or java-aiplatform

.kokoro/nightly/graalvm-native-17.cfg
```diff
@@ -30,4 +30,9 @@ env_vars: {
 env_vars: {
   key: "SECRET_MANAGER_KEYS"
   value: "java-it-service-account, java-bigqueryconnection-samples-secrets"
+}
+
+env_vars: {
+  key: "TEST_ALL_MODULES"
+  value: "true"
```
lqiu96, line 37: I'm thinking to remove this env_var in a separate PR since we should even allow this flag to be toggled in the kokoro build. We can manually trigger the nightly build in fusion if we want to test all modules.

java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/Util.java
```diff
@@ -19,7 +24,7 @@ public static void cleanUpComputeInstances(
     ListPagedResponse listPagedResponse = instancesClient.list(project, zone);
     for (Instance instance : listPagedResponse.iterateAll()) {
       if (isCreatedBeforeThresholdTime(
-              ZonedDateTime.parse(instance.getCreationTimestamp()).toInstant())
+          ZonedDateTime.parse(instance.getCreationTimestamp()).toInstant())
           && instance.getName().startsWith(BaseTest.COMPUTE_PREFIX)) {
```
lqiu96, line 28: Going to pass in the prefix into function in a separate PR. https://github.com/googleapis/google-cloud-java/pull/8260#discussion_r968578289

.kokoro/nightly/graalvm-native-17.cfg
lqiu96: this is the wrong file (should be presubmit)

.kokoro/presubmit/graalvm-native-17-aiplatform.cfg
```diff
@@ -0,0 +1,43 @@
...
+
+env_vars: {
+  key: "MAVEN_MODULES"
+  value: "java-aiplatform"
+}
+
+env_vars: {
+  key: "TEST_ALL_MODULES"
+  value: "false"
+}
```
lqiu96, line 43: I'm thinking to remove this env_var in a separate PR since we shouldn't even allow this flag to be toggled in the presubmit kokoro build. We can manually trigger the nightly build in fusion if we want to test all modules.

lqiu96:
@suztomo @meltsufin PTAL whenever you get a chance.

meltsufin (approved):

PR #8257, opened 2022-08-31
chore: script to create a new module

suztomo:
Adding a Python script to create a new module in this monorepo.

PR #8072, opened 2022-07-28
feat: Replace bom's parent with aggregator

lqiu96:

PR #8383, opened 2022-09-13
build: script to set parent pom in each module

suztomo:
generation/set_parent_pom.sh is the script to set the parent pom in each module. I'll use this script from the upcoming script when creating a new module (https://github.com/googleapis/google-cloud-java/pull/8257).

### DIFF
--- a/generation/set_parent_pom.sh
+++ b/generation/set_parent_pom.sh
@@ -7,12 +7,14 @@ set -ef
 
 # Run this script at the root of google-cloud-java repository
 
-
+# First, read the values from the root pom.xml
 parent_version=$(perl -nle 'print $1 if m|<version>(.+)</version>|' pom.xml |head -1)
 parent_group_id=$(perl -nle 'print $1 if m|<groupId>(.+)</groupId>|' pom.xml |head -1)
 parent_artifact_id=$(perl -nle 'print $1 if m|<artifactId>(.+)</artifactId>|' pom.xml |head -1)
 
+# Then, apply the values as the parent pom of each module
 for module in $(find . -mindepth 2 -maxdepth 2 -name pom.xml |sort | xargs dirname); do
+  # example value of module is "./java-accessapproval"
   if [[ "${module}" = *google-cloud-gapic-bom ]] || [[ "${module}" = *CoverageAggregator ]]; then
     continue
   fi


generation/set_parent_pom.sh
```diff
@@ -0,0 +1,25 @@
...
+
+for module in $(find . -mindepth 2 -maxdepth 2 -name pom.xml |sort | xargs dirname); do
+  if [[ "${module}" = *google-cloud-gapic-bom ]] || [[ "${module}" = *CoverageAggregator ]]; then
+    continue
+  fi
+  echo "Processing module $module"
+  pushd $module
+  # Search for <parent> tag in module pom and replace the next three lines -- groupId, artifcatId, and version
+  sed -i.bak -e "/<parent>/{N;s|<groupId>.*</groupId>|<groupId>${parent_group_id}</groupId>|;N;s|<artifactId>.*</artifactId>|<artifactId>${parent_artifact_id}</artifactId>|;N;s|<version>.*</version>|<version>${parent_version}</version>|}" pom.xml
+  rm pom.xml.bak
```
suztomo, line 23: @lqiu96 Do you remember why we had to change the parent of BOMs?

https://github.com/googleapis/google-cloud-java/pull/8072/files#diff-99d5a2a8ca4290b61836b3ad37941f2b925e9ff3151c90dcd5ea82364e4122a1R39

suztomo: Do you still need the BOMs to use the root pom.xml as parent?

lqiu96: Let me run the DocFX plugin locally real quick. iirc, there was an problem with bom not having the correct parent, but I may be mistaken.

lqiu96: Would the preference to point the BOM back to shared-configs?

suztomo: Yes, it gives simpler inheritance.

For example if the root pom.xml happens to have dependencyManagement section, we cannot inherit it in BOMs.

suztomo: > Let me run the DocFX plugin locally real quick.

Reading this, I think we want a check (or a job) that can confirm a pull request is not breaking the DocFX work.

lqiu96: Ah, ok I see. Maybe we should change the BOM's parent to point back to shared-configs then

lqiu96: > > Let me run the DocFX plugin locally real quick.
> 
> Reading this, I think we want a check (or a job) that can confirm a pull request is not breaking the DocFX work.

iirc, this was because I couldn't set the custom doclet path (https://github.com/googleapis/java-shared-config/pull/496) to use the latest jar. Testing it locally, it doesn't seem to have an issue with bom pointing to shared-configs.

I think it could make sense to add a job that simply runs `mvn clean javadoc:aggregate` as a presubmit check and check if there are invalid values in the javadoc. That being said, do the backend teams check that the docs in the .proto files are valid? If they have checks on their end, we may not need to add this. Otherwise, it shouldn't be too hard to add a GA to run this check.

PR #8421, opened 2022-09-16
chore: speedup release please config generation

meltsufin:
* The main improvement is skipping version updates for modules that are already on `SNAPSHOT` version.
* Another improvement is skipping version backups with `-DgenerateBackupPoms=false` and the consequent `mvn versions:commit`.
* Finally, there's a fix to make the script work on macOS.

### DIFF
--- a/templates/generation/generate_release_please_config.sh
+++ b/templates/generation/generate_release_please_config.sh
@@ -60,19 +60,27 @@ for module in $(find . -mindepth 2 -maxdepth 2 -name pom.xml |sort | xargs dirna
            new_version=$(echo ${old_version} |  awk -F'.' '{print $1"."$2"."$3+1}' |  sed s/[.]$//)
            new_version="${new_version}-SNAPSHOT"
            sed -i.bak -e "s|${artifact_name}:${old_version}:${old_version}|${artifact_name}:${old_version}:${new_version}|" "${version_file}" && rm "${version_file}".bak
+           artifact_directory="${module}/${artifact_name}"
+           mvn -B -ntp -f ${artifact_directory} -U versions:set -DnewVersion=${new_version} -DgenerateBackupPoms=false
          fi
       done
 
   snapshot_version=$(echo ${module_released_version} |  awk -F'.' '{print $1"."$2"."$3+1}' |  sed s/[.]$//)
 
-  if [[ ${module_name} == "java-notification" ]]; then
-    snapshot_version="${snapshot_version}-beta-SNAPSHOT"
-  else
-    snapshot_version="${snapshot_version}-SNAPSHOT"
+  #bumping to snapshot in module bom directory, if it exists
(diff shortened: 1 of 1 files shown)


templates/generation/generate_release_please_config.sh
```diff
@@ -15,7 +15,7 @@ rp_manifest_line=""
 
 for module in $(find . -mindepth 2 -maxdepth 2 -name pom.xml |sort | xargs dirname); do
 
-  if [[ "${module}" = *google-cloud-gapic-bom ]] || [[ "${module}" = *CoverageAggregator ]]; then
+  if [[ "${module}" = *google-cloud-gapic-bom ]] || [[ "${module}" = *CoverageAggregator ]] || grep -q 'SNAPSHOT<\/version>' "${module}/pom.xml" ; then
```
meltsufin, line 18: skip the whole versions update if it's already on `SNAPSHOT`.

templates/generation/generate_release_please_config.sh
```diff
@@ -51,39 +51,8 @@ for module in $(find . -mindepth 2 -maxdepth 2 -name pom.xml |sort | xargs dirna
...
   #bumping to snapshot in module root pom
-  mvn -B -ntp -f ${module} -U versions:set -DnewVersion="${snapshot_version}-SNAPSHOT"
-  mvn versions:commit
-
-  #specific case
-  if [[ ${module_name} == "java-notification" ]]; then
-    mvn -B -ntp -f ${module} -U versions:set -DnewVersion="${snapshot_version}-beta-SNAPSHOT"
-    mvn versions:commit
-  fi
+  mvn -B -ntp -f ${module} -U versions:set -DnewVersion="${module_snapshot_version}" -DprocessAllModules -DgenerateBackupPoms=false -DprocessFromLocalAggregationRoot=false
```
meltsufin, line 55: All sub-modules should have the same version. So, I don't think you need to go through each library sub-module one by one.

ddixit14: that's the thing, they might be not. For example, in java-vision, we've got different versions for submodules -> https://github.com/googleapis/java-vision/blob/main/versions.txt

templates/generation/generate_release_please_config.sh
ddixit14: I've modified this script since this commit. The latest version of it can be found here -> https://github.com/googleapis/google-cloud-java/blob/main/generation/generate_release_please_config.sh
We can just run this one on the main branch and generate the configs.

meltsufin: The version in `main` looks the same to me.

templates/generation/generate_release_please_config.sh
meltsufin: I didn't know that was possible. I'll bring your per-module code back.

ddixit14 (approved):

PR #8425, opened 2022-09-19
chore: delete handwritten samples

meltsufin:
The only samples remaining are the ones created using SnippetsGen.
This PR was created by running the `generation/delete_non_generated_samples.sh` script added in a separate commit in this PR.


lqiu96 (approved):

tmatsuo:
@meltsufin Snippet-bot is not happy about removing region tags in use. Merging this PR will cause the devsite build to break.
I added "do not merge" label.

meltsufin:
@tmatsuo These modules were copied from split repos. Those split repos are retaining all of these samples and the docs should not break.

tmatsuo:
Yeah I manually checked the source code of the docs for the first violation reported by snippet-bot (`dataproc_create_autoscaling_cluster`).

The source code contains:

```
 'language_snippets': [{'display': 'Java',
                        'filename': 'java-dataproc/samples/snippets/src/main/java/CreateClusterWithAutoscaling.java',
                        'keyword': 'Java',
                        'org': 'googleapis',
                        'repo': 'google-cloud-java',
                        'slug': 'java'}],
```

I think the devsite build will break because of this.

meltsufin:
@tmatsuo It looks like it was autogenerated by Sample Browser. @jmdobry Do you know what happens if there are duplicates for region tags?

tmatsuo:
Ok, DRIFT is debugging the data. Also I manually checked those URLs and they all have `docs/samples` so quite likely they're all sample browser URLs (not guaranteed).

I removed the do-not-merge. Thanks for your patience.

meltsufin:
@tmatsuo Are we not worried about the Sample Browser being broken?

tmatsuo:
@meltsufin AFAIK, the sample browser will detect the removal and remove the page as well (it will be self heal).

tmatsuo:
@meltsufin JFYI, I told that it would be deleted, but in this case, the page:
https://cloud.google.com/dataproc/docs/samples/dataproc-create-autoscaling-cluster

has been rewritten with the sample code from the split repository and now the link goes to:
https://github.com/googleapis/java-dataproc/blob/HEAD/samples/snippets/src/main/java/CreateClusterWithAutoscaling.java

meltsufin:
Thanks for the update. This is perfect!

PR #8432, opened 2022-09-20
chore: Script to update versions.txt to latest

lqiu96:
Create a script to bump the versions to match the latest version in maven central (functionality is similar to release-please).

Add commit `55ea17c0427462d4ab23d6f1f18999ce6c3541fc` (which is the output/ update from running the script). If the patch bump looks good, I can revert that commit or keep it.

Retrieve the the `maven-metadata.xml` from maven central and use the latest version stored in that file. Use patch version bump to increase for the next `-SNAPSHOT` version.

Certain modules have different paths in maven central
- analytics -> com/google/analytics/
- area120 -> com/google/area120/
- grafeas -> io/grafeas/ (Grafeas has a completely different structure?)
- beyondcorp -> Does not exist in maven central

Ignore gapic-bom, CoverageAggregator, and samples directory

### DIFF
--- a/generation/update_versions.sh
+++ b/generation/update_versions.sh
@@ -8,7 +8,6 @@ function retry_with_backoff {
   shift 2
   command=$@
 
-
   # store current flag state
   flags=$-
 
@@ -25,7 +24,7 @@ function retry_with_backoff {
 
   if [[ $exit_code == 0 ]]
   then
-      return 0
+    return 0
   fi
 
   # failure


suztomo:
@ddixit14 Does this script need to modify pom.xml files or not?

meltsufin:
Shouldn't Release Please do this automatically? Why do we have this problem in the monorepo, but not in the split repos?

lqiu96:
> Shouldn't Release Please do this automatically? Why do we have this problem in the monorepo, but not in the split repos?

Release Please does this automatically, but it has the versions incorrect (lagging behind) since we are still releasing from the split repos. iirc, the purpose of this script is to be run one time after we stop releasing from the split repos so the monorepo's version.txt would have the latest versions.

java-errorreporting/versions.txt
```diff
@@ -2,5 +2,5 @@
 # module:released-version:current-version
 
 google-cloud-errorreporting:0.124.4-beta:0.124.5-beta-SNAPSHOT
```
meltsufin, line 4: Are you handling the `-beta-SNAPSHOT` versions correctly?

lqiu96: Ah, thanks for catching that. I'll update the regex.

ddixit14:
> @ddixit14 Does this script need to modify pom.xml files or not?
> 
> My current thought is that this script does not need to modify pom.xml files, because Release Please updates the files next time we create a release pull request.

I think so too. Once we merge this in, release-please will scan the current version from versions.txt, scan through the commits and suggest the next version accordingly.

generation/update_versions.sh
```diff
@@ -0,0 +1,90 @@
...
+
+    if [[ $exit_code == 0 ]]
+    then
+        return 0
+    fi
+
+    # failure
+    if [[ ${attempts_left} -gt 0 ]]
+    then
+        echo "failure (${exit_code}), sleeping ${sleep_seconds}..."
```
suztomo, line 34: Can you do 2-space indentation?

lqiu96:
There are 133 modules listed under the <modules> tag in the aggregator pom.xml.
There are 123 modules that have been modified from this script (as of the run on 9/21/22)
- Ignore CoverageAggregator (1)
- Ignore Gapic-Bom (1)
- Ignore Beyondcorp Modules (5)
- No new maven releases (3)
  - java-appengine-admin
  - java-bigquery-data-exchange
  - java-notebooks

Added retry_with_backoff to retry if the script fails to download from maven central. I see that a module will fail every now and then due to timeout/ connection refused (some artifacts have to be retried). I've added some rate limiting to prevent our request from being refused.

Since this is supposed to be invoked only once again in the future, I think we should be fine to do a quick spot check on the output/ logs to see if it missed an artifact. What do you think @suztomo ?

generation/update_versions.sh
lqiu96: Sure. Will fix.

generation/update_versions.sh
```diff
@@ -0,0 +1,90 @@
...
+    fi
+
+    return $exit_code
+}
+
+count=0
+missing_artifacts=()
+
+for path in $(find . -mindepth 2 -maxdepth 2 -name pom.xml | sort | xargs dirname); do
+  if [[ "${path}" =~ google-cloud-gapic-bom ]] || [[ "${path}" =~ CoverageAggregator ]] || [[ "${path}" =~ .*samples.* ]] || [[ "${path}" =~ .*beyondcorp.* ]]; then
```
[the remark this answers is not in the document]
lqiu96: Yep will add

lqiu96:
Also, these artifacts don't exist in maven central: `proto-google-cloud-speech-v2` `grpc-google-cloud-speech-v2` so they are kept the same.

java-errorreporting/versions.txt
```diff
@@ -1,6 +1,6 @@
 # Format:
 # module:released-version:current-version
 
-google-cloud-errorreporting:0.124.4-beta:0.124.5-beta-SNAPSHOT
```
lqiu96, line 4: Need to fix this. Forgot the `-beta` tag

suztomo (approved):

meltsufin (approved):

PR #8436, opened 2022-09-21
ci: Initial Terraform configurations (Work in Progress)

burkedavison:
See `./.terraform/test.sh` for instructions.

These terraform configurations are (currently) designed to create and destroy a new GCP project for each module being tested. To run the script, you must identify a GCP Folder ID in which the project(s) will be created, and a GCP Billing Account ID to assign to the created project(s).

This is a work in progress. We have not decided we want to do this, but it does offer full isolation between integration tests.

`terraform` and `gcloud` must be installed. Developed on MacOS. Not yet tested on Linux.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -48,11 +48,11 @@ case ${JOB_TYPE} in
         echo "${modified_module_list[*]}"
       )
 
-      terraform -version
-      source ./.terraform/helpers/init.sh "$module_list"
-      source ./.terraform/helpers/plan.sh
-      source ./.terraform/helpers/apply.sh
-      source ./.terraform/helpers/populate-env.sh
+#      terraform -version
+#      source ./.terraform/helpers/init.sh "$module_list"
+#      source ./.terraform/helpers/plan.sh
+#      source ./.terraform/helpers/apply.sh
+#      source ./.terraform/helpers/populate-env.sh
 
       install_modules
       printf "Running Integration Tests for:\n%s\n" "${module_list}"
@@ -74,7 +74,7 @@ case ${JOB_TYPE} in
         verify
       RETURN_CODE=$?
 
-      source ./.terraform/helpers/destroy.sh
+#      source ./.terraform/helpers/destroy.sh
 
       printf "Finished Integration Tests for:\n%s\n" "${module_list}"
     else


java-container/.terraform/main.tf
```diff
@@ -0,0 +1,43 @@
...
+}
+resource "google_compute_network" "java_container_network" {
+  name       = "java-container-network"
+  project    = module.project.project_id
+  depends_on = [
+    module.project,
+    google_project_service.compute_api,
+    google_project_service.container_api
+  ]
+}
```
suztomo, line 43: @ckim328 We're evaluating Terraform to prepare integration test infrastructure. You have setup networks in gcloud-devel project in past (You helped me unblock integration test failure in java-container). Do you see anything special required to create a network for java-container integration tests?

meltsufin:
Have you looked into how we will run it as part of CI/Kokoro? I guess we'll need to configure Kokoro service account to have permissions on the products directory.

burkedavison:
Kokoro is part of the discussion, but not the current focus until we can
iron out some of the issues currently being faced.

There are some issues with this create+destroy project workflow that I need
to resolve; and we haven't made a clear decision yet on whether we want to
continue down this path. If we choose to continue having Terraform
create+destroy projects, then yes - we'd need to give Kokoro's service
account project creation+deletion permissions and define a specific folder
for it to use.

However, we could also choose to have a single project in which dedicated
instances of resources (like database instances, network instances, etc)
get created for each test; but this also has issues that would need to be
resolved. (For example, we can't delete container networks due to GCE
Enforcer firewall rule dependencies, so would that mean our
`java-container` ITs litter the project with old networks, or would
`java-container` be set up to use only a single network -- or would we have
a set of networks that could be allocated+released to allow only a single
owner at a time...)

java-container/.terraform/main.tf
ckim328: this lgtm.
You might need this `auto_create_subnetworks = true` ([source](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network)) or you might get this [error])https://github.com/googleapis/java-container/pull/781#issuecomment-1248395098) again

burkedavison: Good insight. I'll add it!

.terraform/test.sh
```diff
@@ -0,0 +1,129 @@
...
+    -Penable-integration-tests \
+    -DtrimStackTrace=false \
+    -Dclirr.skip=true \
+    -Denforcer.skip=true \
+    -Dcheckstyle.skip=true \
+    -fae \
+    verify
+
+  exit_code=$?
+  popd || exit
```
suztomo, line 67: What's the situation where `popd` fails?

burkedavison: I have this primarily due to `shellcheck` providing a warning otherwise. See https://github.com/koalaman/shellcheck/wiki/SC2164

> missing directories, missing permissions, broken symlinks

.terraform/modules-under-test.txt
```diff
@@ -0,0 +1,20 @@
+# Modules to be tested when no specific module is given to test.sh as an argument
```
suztomo, line 1: I think this file can be replaced by a command to find directories that contain Terraform configuration. With that we can eliminate a possibility of having discrepancy (Terraform files are added to modules but forget updating this file). What do you think?

burkedavison: Good suggestion. TODO

java-container/.terraform/main.tf
```diff
@@ -0,0 +1,43 @@
+terraform {
+  required_providers {
+    google = {
```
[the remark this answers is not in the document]
burkedavison: Investigating

meltsufin (approved):

.terraform/test.sh
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+scriptDir="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+pushd "$scriptDir" >/dev/null || exit
+
+source ./init.sh "$1"
```
suztomo, line 19: How might we run all (applicable) modules that has terraform configuration?

burkedavison: As described in `./.terraform/README.md`, running `./.terraform/test.sh` without any arguments will perform integration tests on all modules. If a module has a terraform configuration associated with it in `./<module>/.terraform`, then its configuration will be incorporated in the requirements while setting up the project.

There isn't a command to perform integration testing only on modules with terraform configurations, though I could make that happen if that's something you'd interested in.

Alternatively, `./.terraform/test.sh <module>:<module>` will test the given modules specifically.

.terraform/helpers/generated-main.template.tf
```diff
@@ -0,0 +1,48 @@
+# Auto-generated by generate-config.sh
```
suztomo, line 1: What if we don't add generated files. Would this PR becomes much smaller?

burkedavison: These three 'template' files aren't actually generated -- they're used to copy and then append to the "actual" generated files which are not in git.

burkedavison:
See https://github.com/googleapis/google-cloud-java/pull/8555 for continuation of this effort.

PR #8446, opened 2022-09-23
chore: Set versions.txt to match latest released version

lqiu96:


generation/update_versions.sh
```diff
@@ -78,19 +78,21 @@ for path in $(find . -mindepth 2 -maxdepth 2 -name pom.xml | sort | xargs dirnam
...
-      patch_version_bump=$((patch_version + 1))
-      if [[ -z "${maven_latest_trailing}" ]]; then
-        maven_version_bump="${major_version}.${minor_version}.${patch_version_bump}"
-      else
-        maven_version_bump="${major_version}.${minor_version}.${patch_version_bump}-${maven_latest_trailing}"
-      fi
-      new_version="${artifactId}:${maven_metadata_version}:${maven_version_bump}-SNAPSHOT"
+
+      # TODO: Confirm if release-please is able to bump snapshot
+#      maven_latest_version=$(echo "${maven_metadata_version}"  | cut -d "-" -f1)
```
meltsufin, line 83: I think you can just delete this code. We have version history for the event of having to restore.

meltsufin (approved):

java-accessapproval/versions.txt
```diff
@@ -1,6 +1,6 @@
 # Format:
 # module:released-version:current-version
 
-google-cloud-accessapproval:2.4.8:2.4.9-SNAPSHOT
-grpc-google-cloud-accessapproval-v1:2.4.8:2.4.9-SNAPSHOT
-proto-google-cloud-accessapproval-v1:2.4.8:2.4.9-SNAPSHOT
+google-cloud-accessapproval:2.4.8:2.4.8
```
suztomo, line 4: We have not yet released so having "-SNAPSHOT" is correct, isn't it?

suztomo (changes requested):
I think we don't need to change the current versions.txt? Would you clarify the problem?

java-accessapproval/versions.txt
lqiu96: Hasn't it been released? https://search.maven.org/artifact/com.google.cloud/google-cloud-accessapproval/2.4.8/jar

lqiu96:
> I think we don't need to change the current versions.txt? Would you clarify the problem?

This is to run the release-please with the `released version` and `current version` matching in versions.txt. This is what release-please is expecting before it bumps to the `-SNAPSHOT` version. Once this is merged in, I'll run `release-please:force-run` to see that the versions are bumped.

generation/update_versions.sh
```diff
@@ -78,19 +78,7 @@ for path in $(find . -mindepth 2 -maxdepth 2 -name pom.xml | sort | xargs dirnam
...
-      minor_version=$(echo "${maven_latest_version}" | cut -d "." -f2)
-      patch_version=$(echo "${maven_latest_version}" | cut -d "." -f3)
-      patch_version_bump=$((patch_version + 1))
-      if [[ -z "${maven_latest_trailing}" ]]; then
-        maven_version_bump="${major_version}.${minor_version}.${patch_version_bump}"
-      else
-        maven_version_bump="${major_version}.${minor_version}.${patch_version_bump}-${maven_latest_trailing}"
-      fi
-      new_version="${artifactId}:${maven_metadata_version}:${maven_version_bump}-SNAPSHOT"
+      new_version="${artifactId}:${maven_metadata_version}:${maven_metadata_version}"
```
suztomo, line 81: Can you introduce a flag in this script so that we can use this script when we need to update the version with "-SNAPSHOT" suffix?

suztomo (approved):
Got it. It's to experiment release please behavior.

generation/update_versions.sh
lqiu96: Yeah I can. Let me do that in a new PR.

PR #8453, opened 2022-09-23
chore: Add SNAPSHOT flag to bump versions.txt

lqiu96:


meltsufin:
What's the purpose of this? I thought we verified that release-please can do the snapshot bumping.

lqiu96:
> What's the purpose of this? I thought we verified that release-please can do the snapshot bumping.

Added a flag for updating current_version to -SNAPSHOT: https://github.com/googleapis/google-cloud-java/pull/8446#discussion_r979014838

meltsufin:
> > What's the purpose of this? I thought we verified that release-please can do the snapshot bumping.
> 
> Added a flag for updating current_version to -SNAPSHOT: [#8446 (comment)](https://github.com/googleapis/google-cloud-java/pull/8446#discussion_r979014838)

@suztomo In what situation do anticipate us needed this?

suztomo:
We need that capability When the versions of the libraries in this monorepo are the same or older than the ones in Maven Central.

meltsufin:
> We need that capability when the versions of the libraries in this monorepo are the same or older than the ones in Maven Central.

@suztomo I thought we determined that release-please will automatically bump to SNAPSHOT versions. Is that actually not the case?

Per [documentation](https://github.com/googleapis/release-please/blob/main/docs/java.md):

>This strategy generates SNAPSHOT versions after each release. Snapshot is created as separate "release" Pull Request, which updates all affected files, but does not create actual release or tag.

suztomo:
Release Please bumps the version to SNAPSHOT in a "release-please:snapshot"-labelled pull request, after a normal release pull request (labelled "release-please:pending" initially). We have the latter in this repo currently.

Once we actually start releasing this repository, we don't need this snapshot flag.

meltsufin:
What would happen if we closed the pending release PR? What's the trigger for release-please to create the SNAPSHOT PR? I thought it was the `versions.txt` being in a specific state.

lqiu96:
Merging in snapshot flag to run so that we can generate `release: main`

PR #8469, opened 2022-09-27
chore: Adding release-please annotations

ddixit14:


suztomo (changes requested):
Can you remove "-parent" suffix from the key?

By convention, we don't use "-parent" as key. So "google-cloud-java" is a good key.
https://github.com/googleapis/java-accessapproval/blob/main/versions.txt

versions.txt
```diff
@@ -0,0 +1,4 @@
+# Format:
+# module:released-version:current-version
+
+google-cloud-java:0.0.0:0.0.1-SNAPSHOT
```
meltsufin, line 4: Should we start with `1.0.0-SNAPSHOT`?

meltsufin (approved):

suztomo (approved):

versions.txt
ddixit14: Done.

PR #8477, opened 2022-09-28
fix: Trying to fix release-please annotation based version updates

ddixit14:
Checking if removal of `<type>` and `<scope>` tags will enable release-please to update versions.

### DIFF
--- a/google-cloud-gapic-bom/pom.xml
+++ b/google-cloud-gapic-bom/pom.xml
@@ -51,8 +51,6 @@
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-accessapproval-bom</artifactId>
         <version>2.4.5-SNAPSHOT</version><!-- {x-version-update:google-cloud-accessapproval:current} -->
-        <type>pom</type>
-        <scope>import</scope>
       </dependency>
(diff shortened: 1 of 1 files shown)


google-cloud-gapic-bom/pom.xml
```diff
@@ -51,8 +51,6 @@
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-accessapproval-bom</artifactId>
         <version>2.4.5-SNAPSHOT</version><!-- {x-version-update:google-cloud-accessapproval:current} -->
-        <type>pom</type>
-        <scope>import</scope>
```
suztomo, line 55: The removal of these lines are incorrect.

ddixit14: Found the [root cause](https://docs.google.com/document/d/1mJZ8_ZcXnAkTLIIs99EKdAs9rXanVOObnau-UYOpaCI/edit?resourcekey=0-PTBsm1mWIUx6czBF9G0Mog#heading=h.uratr9em48ye). Here's the [issue](https://github.com/googleapis/release-please/issues/1664) raised for it.

PR #8492, opened 2022-09-30
chore: exclude and remove unnecessary owlbot postprocessor templates

meltsufin:
This PR adds a [script](https://github.com/googleapis/google-cloud-java/pull/8492/commits/51a1564abeb47ac7e943018d420cc3b21a79c115) to remove and exclude the OwlBot Post-processor templates that are no longer needed, and also [applies](https://github.com/googleapis/google-cloud-java/pull/8492/commits/1ae477730adb84f3c62ca01aa381cae2c8564a66) the script to the repo in a separate commit.


lqiu96 (approved):

generation/update_owlbot_postprocessor_config.sh
```diff
@@ -0,0 +1,32 @@
...
+    # delete files and directories
+    if [[ ${exclusion: -2} == '/*' ]]; then
+      rm -rf "${exclusion%??}" || true
+    else
+      rm "${exclusion}" || true
+    fi
+  done
+  perl_command+='\n])/s'
+  # execute the replacement in owlbot.py
+  perl -0pe "$perl_command" 'owlbot.py' > 'owlbot.py.new' && rm 'owlbot.py' && mv 'owlbot.py.new' 'owlbot.py'
```
suztomo, line 28: I think a better approach is to fix OwlBot Java PostProcessor to use a new template specific to monorepo.

Currently it uses the same template as split repos (both hand-written and auto-generated) https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/java_library

suztomo (approved):

generation/update_owlbot_postprocessor_config.sh
meltsufin: Where's the new template? Note that we can't simply replace all of the `owlbot.py` configs in every module because many of them have various customizations. Example: https://github.com/googleapis/google-cloud-java/blob/main/java-vision/owlbot.py

suztomo: New template location https://github.com/googleapis/google-cloud-java/pull/8517/files#diff-cf729d64240ae503aa465cb17da8fd060351085df2c1f39393357b54537dfcfe

Issue #8470, opened 2022-09-27
Backfill each module's CHANGELOG.md into the monorepo


lqiu96:
We should copy each split repo's CHANGLOG.md into the monorepo. Release-Please skips the previous versions so we'll lose track if the split repos have done releases.

suztomo:
We will prepare the script and will run it just before the 1st release.

PR #8498, opened 2022-09-30
chore: Copy the latest CHANGELOG.md and README.md from split repo

lqiu96:
Script to copy the latest CHANGELOG.md and README.md from split repo into the monorepo.

Resolves: https://github.com/googleapis/google-cloud-java/issues/8470

### DIFF
--- a/generation/copy_readme_changelog.sh
+++ b/generation/copy_readme_changelog.sh
@@ -3,17 +3,18 @@
 set -e
 
 for path in $(find . -mindepth 2 -maxdepth 2 -name pom.xml | sort | xargs dirname); do
-  if [[ "${path}" =~ google-cloud-gapic-bom ]] || [[ "${path}" =~ CoverageAggregator ]]; then
-    continue
-  fi
   pushd "${path}"
 
   module_name="${path:2}"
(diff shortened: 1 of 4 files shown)


generation/copy_readme_changelog.sh
```diff
@@ -0,0 +1,19 @@
...
+  if [[ "${path}" =~ google-cloud-gapic-bom ]] || [[ "${path}" =~ CoverageAggregator ]]; then
+    continue
+  fi
+  pushd "${path}"
+
+  module_name="${path:2}"
+  raw_readme_url="https://raw.githubusercontent.com/googleapis/${module_name}/main/README.md"
+  raw_changelog_url="https://raw.githubusercontent.com/googleapis/${module_name}/main/CHANGELOG.md"
+
+  curl -s "${raw_readme_url}" > README.md
```
meltsufin, line 15: You should probably handle the case where the file is not found. It could happen when we only add the new library in the monorepo before the first release. That way you also wouldn't have to have the `continue` statement above.

lqiu96: Yeah, that makes sense. I'll add the curl command to check for the URL.

meltsufin (approved):

PR #8500, opened 2022-10-01
chore: migrate bootstrap.sh to main branch

meltsufin:
This PR migrates `bootstrap.sh` and `repos.txt` from `monorepo_script` to `main`. This allows us to regenerate the monorepo and create diffs in `generation/monorepo` without needing the `monorepo_script` branch. This also allows us to continuously test most of the scripts in `generation/` via CI executions of `bootstrap.sh`.

### DIFF
--- a/generation/bootstrap.sh
+++ b/generation/bootstrap.sh
@@ -72,13 +72,13 @@ git commit -m 'chore: add template files'
 ../../print_root_pom.sh > pom.xml
 
 git add pom.xml
-git commit -am 'feat: create aggregator pom'
+git commit -am 'chore: create aggregator pom'
 
 # Point modules poms and BOMs to the aggregator pom as parent
 bash ../../set_parent_pom.sh
 
 git add --all
-git commit -am 'feat: point modules to the aggregator pom as parent'
+git commit -am 'chore: point modules to the aggregator pom as parent'
 
 ../../generate_gapic_bom.sh
 
@@ -86,7 +86,7 @@ git commit -am 'feat: point modules to the aggregator pom as parent'
 ../../print_root_pom.sh > pom.xml
 
 git add google-cloud-gapic-bom/pom.xml
-git commit -am 'feat: create google-cloud-gapic-bom'
+git commit -am 'chore: create google-cloud-gapic-bom'
 
 cp ../../gapic_bom_versions.txt google-cloud-gapic-bom/versions.txt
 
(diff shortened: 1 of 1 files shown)


generation/bootstrap.sh
```diff
@@ -0,0 +1,158 @@
...
+git add google-cloud-gapic-bom/pom.xml
+git commit -am 'feat: create google-cloud-gapic-bom'
+
+cp ../../gapic_bom_versions.txt google-cloud-gapic-bom/versions.txt
+
+# Generate Release Please configuration files
+bash ../../generate_release_please_config.sh
+
+git add --all
+git commit -am 'feat: create release please configuration'
```
ddixit14, line 97: Can we change the commit prefix to `chore` ? Otherwise it gets picked up in release-notes.

meltsufin: Done.

ddixit14 (approved):

.github/workflows/bootstrap.yaml
```diff
@@ -0,0 +1,41 @@
...
+        run: mvn compile -T C1 -B
+      - name: Unit Tests
+        working-directory: generation/monorepo/google-cloud-java
+        run: mvn test -T C1 -B
+      - name: Push monorepo to branch
+        if: ${{ github.event_name != 'pull_request' }}
+        run: |
+          cd generation/monorepo/google-cloud-java
+          git remote add monorepo https://${{ github.actor }}:${{ github.token }}@github.com/${{ github.repository }}.git
+          git push -f monorepo main:bootstrap_output
```
suztomo, line 36: Isn't the target branch name monorepo_script_output?

.github/workflows/bootstrap.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Push diff to branch
+        run: |
+          cd generation/monorepo/diff
+          git remote add monorepo https://${{ github.actor }}:${{ github.token }}@github.com/${{ github.repository }}.git
+          git push -f monorepo main:boootstrap_output_diff_${{ github.event_name }}
```
suztomo, line 41: Isn't the target branch name monorepo_script_output?

.github/workflows/bootstrap.yaml
meltsufin: I'm changing it, to reduce confusion.

.github/workflows/bootstrap.yaml
meltsufin: This is the output for the diff part of the script.
Notice `cd generation/monorepo/diff` above.

suztomo (approved):

PR #8517, opened 2022-10-04
chore: script to create a new module

suztomo:
Cont of https://github.com/googleapis/google-cloud-java/pull/8257. generation/new_client/README.md is the usage document.

https://github.com/googleapis/google-cloud-java/pull/8519 is an example output of this script. The pull request to create a new module.

`suztomo@suztomo:~/google-cloud-java/generation/new_client$ python3.9 new-client.py generate   --api_shortname=apikeys   --proto-path=google/api/apikeys   --name-pretty="API Keys API"   --product-docs="https://cloud.google.com/api-keys/"   --api-description="API Keys lets you create and manage your API keys for your projects."   --release-level=preview   --transport=grpc   --language=java   --monorepo-url="https://github.com/googleapis/google-cloud-java"`

### DIFF
--- a/generation/new_client/new-client.py
+++ b/generation/new_client/new-client.py
@@ -318,11 +318,19 @@ def generate(
         ],
         cwd=monorepo_root,
     )
+    print("Regenerating CoverageAggregator")
+    subprocess.check_call(
+        [
+            "bash", "generation/generate_coverage_aggregator.sh"
+        ],
+        cwd=monorepo_root,
+    )
 
     # Add the files to commit
     subprocess.check_call([
         "git", "add", "pom.xml", "google-cloud-gapic-bom/pom.xml",
-        "release-please-config.json", ".release-please-manifest.json"],
+        "release-please-config.json", ".release-please-manifest.json",
+        "CoverageAggregator/pom.xml"],
         cwd=monorepo_root)
 
     subprocess.check_call(


.gitignore
```diff
@@ -29,6 +37,22 @@ docs/
...
-*.pyc
+__pycache__/
+*.py[cod]
+*$py.class
 
-monorepo/
\ No newline at end of file
+
+# Environments
+.env
```
meltsufin, line 46: Why all these changes in `.gitignore`.

generation/new_client/templates/new-client/owlbot.py.j2
```diff
@@ -0,0 +1,24 @@
+# Copyright 2021 Google LLC
```
meltsufin, line 1: Can we put these template files under `generation/new_client`?

meltsufin (approved):

.gitignore
```diff
@@ -29,6 +37,22 @@ docs/
 api_key
 
 # Python utilities
-*.pyc
```
suztomo, line 32: pyc is covered by `*.py[cod]`

.gitignore
```diff
@@ -29,6 +37,22 @@ docs/
...
+__pycache__/
+*.py[cod]
+*$py.class
 
-monorepo/
```
suztomo, line 34: This moved to the end.

generation/new_client/templates/new-client/owlbot.py.j2
suztomo: I prefer to place template files in a separate directory than scripts files.

.gitignore
suztomo: Excerpt from https://github.com/github/gitignore/blob/main/Python.gitignore

generation/new_client/templates/new-client/owlbot.py.j2
meltsufin: Can you make a subdirectory within generation/?

.gitignore
meltsufin: Maybe group all of the Python-related ignores in a separate section? Or, restrict it to a specific subdirectory like `generation/`?

suztomo: Grouped the Python-related entries.

generation/new_client/templates/new-client/owlbot.py.j2
suztomo: (The better question is) What do you feel is the problem with the current location?

meltsufin: It's used only by scripts in `/generation`. So, I think it makes sense to be organized that way. For someone navigating the code, it will cause less head scratching. They can still be in a separate directory: `generation/templates`.

suztomo: For those who navigate code, the current structure is unnecessarily deep. That makes sense. Let me change it less nested.

suztomo: Moved template files from `generation/new_client/templates/new-client` to `generation/new_client/templates`.

meltsufin: I think I actually misread the original directory structure. Did you have it in `generation/templates` already? I thought it was in `templates` parallel to `generation`. In that case, sorry for the confusion.

PR #8521, opened 2022-10-05
chore: consolidate configuration in google-cloud-pom-parent and google-cloud-jar-parent

meltsufin:
* Attempts to centralize all dependency version management in the root `pom.xml` and the new `google-cloud-jar-parent`.
* Introduces the `consolidate_config.sh` script to help apply the change across all modules
* Reduces the scope of renovate bot to just the parent poms
* There are still many locations where versions are declared outside of the root pom and are not managed by release-please. That will be addressed in the follow-up PRs. For now, this PR adds a CI check to find such violations.

Fixes: #8543
Fixes: #8530
Fixes: #8522
Fixes: #8548

### DIFF
--- a/generation/generate_release_please_config.sh
+++ b/generation/generate_release_please_config.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -ex
 
 GENERATION_DIR=$(dirname -- "$0")
 
@@ -22,12 +22,8 @@ for path in $(find . -mindepth 2 -maxdepth 2 -name pom.xml | sort --dictionary-o
   module_name="${path:2}"
   version_file="${path}/versions.txt"
 
-  # Java-Grafeas is a special case
-  if [[ "${module_name}" = "java-grafeas" ]]; then
-    module_line=$(grep -E "^grafeas:[0-9]+\.[0-9]+\.[0-9]+.*:[0-9]+\.[0-9]+\.[0-9]+.*$" "${version_file}")
-  else
-    module_line=$(grep -E "^google-.*:[0-9]+\.[0-9]+\.[0-9]+.*:[0-9]+\.[0-9]+\.[0-9]+.*$" "${version_file}")
-  fi
+  module_line=$(grep -E "^((google-.*|grafeas|gapic\-libraries)).*:[0-9]+\.[0-9]+\.[0-9]+.*:[0-9]+\.[0-9]+\.[0-9]+.*$" "${version_file}")
+
   artifact_name=$(echo "${module_line}" | cut -d ":" -f1)
   module_released_version=$(echo "${module_line}" | cut -d ":" -f2)
   module_snapshot_version=$(echo "${module_line}" | cut -d ":" -f3)


suztomo:
If the root pom.xml has dependencyManagement section, ensure the small BOMs do not inherit the root pom.xml
E.g.,
https://github.com/googleapis/google-cloud-java/blob/main/java-accessapproval/google-cloud-accessapproval-bom/pom.xml#L10

meltsufin:
@suztomo Do we need a parent at all for the boms?

pom.xml
```diff
@@ -18,9 +18,42 @@
...
 
+  <developers>
+    <developer>
+      <id>chingor</id>
+      <name>Jeff Ching</name>
+      <email>chingor@google.com</email>
+      <organization>Google</organization>
+      <roles>
+        <role>Developer</role>
+      </roles>
```
suztomo, line 29: Let's update this value. Maybe "Cloud Java team"?

suztomo (commented):
In this pull request, can you add few paragraphs in README.md (or somewhere) that explain why we introduce `google-cloud-pom-parent` and `google-cloud-jar-parent`, and their roles (What kinds of elements should they have)? The paragraphs should cover the things you verbally explained yesterday and the maintainers 1 year later will immediately understand the design.

suztomo (commented):
I think google-cloud-gapic-bom, google-cloud-pom-parent, and google-cloud-jar-parent all should have the same value but currently it's maintained by different key such as `{x-version-update:google-cloud-jar-parent:current}`. Can you set them with 1 key?

java-asset/pom.xml
```diff
@@ -113,57 +82,25 @@
...
-        <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-pubsub</artifactId>
-        <version>1.120.13</version>
-        <scope>test</scope>
-      </dependency>
-      <dependency>
-        <groupId>com.google.api.grpc</groupId>
-        <artifactId>proto-google-cloud-pubsub-v1</artifactId>
-        <version>1.102.13</version>
-        <scope>test</scope>
```
suztomo, line 161: Were they unused?

google-cloud-jar-parent/pom.xml
```diff
@@ -0,0 +1,230 @@
...
+  <packaging>pom</packaging>
+
+  <parent>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>google-cloud-pom-parent</artifactId>
+    <version>1.0.0-SNAPSHOT</version><!-- {x-version-update:google-cloud-pom-parent:current} -->
+  </parent>
+  <properties>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
```
suztomo, line 18: We target Java 8.

pom.xml
```diff
@@ -161,26 +194,6 @@
...
-            <phase>test</phase>
-            <goals>
-              <goal>report</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
```
lqiu96, line 199: Do we still need the `<build>` tag? The parent pom is for the all the module's bom, which *shouldn't* have any tests to run? iirc, you already moved the build configs to the jar-parent.

pom.xml
meltsufin: Would could try that, but I'm not sure that would be compliant with https://central.sonatype.org/publish/requirements/.

java-asset/pom.xml
meltsufin: This is inside the `dependencyManagement` section, and I moved them into the parent.

google-cloud-jar-parent/pom.xml
meltsufin: good catch

pom.xml
meltsufin: Good point. I'll remove.

meltsufin:
> I think google-cloud-gapic-bom, google-cloud-pom-parent, and google-cloud-jar-parent all should have the same value but currently it's maintained by different key such as `{x-version-update:google-cloud-jar-parent:current}`. Can you set them with 1 key?

True, but for now I like to keep the convention of the annotation matching the artifactId. I now have scripts that rely on this.

pom.xml
meltsufin: I set you as the developer instead of Jeff.

meltsufin:
> In this pull request, can you add few paragraphs in README.md (or somewhere) that explain why we introduce `google-cloud-pom-parent` and `google-cloud-jar-parent`, and their roles (What kinds of elements should they have)? The paragraphs should cover the things you verbally explained yesterday and the maintainers 1 year later will immediately understand the design.

Added repository structure documentation: [512f907](https://github.com/googleapis/google-cloud-java/pull/8521/commits/512f907eca554d6a0df8bc85781319d4e779211b).

suztomo (approved):

PR #8515, opened 2022-10-04
deps: Removing annotations for release-please

ddixit14:
Removing this annotation since this should be updated by renovateBot, not release-please. 
This is also an attempt to regenerate release:main PR with the manifest config.

Issue #8520, opened 2022-10-04
bom generation script has diff depending on platforms


suztomo:
New script generation (I ran on Linux) has irrelevant changes https://github.com/googleapis/google-cloud-java/pull/8519. Probably because the BOM was generated last time on Mac.

<img width="1248" alt="Screen Shot 2022-10-04 at 7 30 50 PM" src="https://user-images.githubusercontent.com/28604/193949086-c346a72b-e351-41b3-be25-d8febfb4c195.png">

The `sort` command's behavior differs between Mac and Linux.

Linux:

```
$ cat /tmp/o |sort
java-bigqueryconnection
java-bigquery-data-exchange
```

Mac:

```
$ cat /tmp/o |sort
java-bigquery-data-exchange
java-bigqueryconnection
```

We should give `--dictionary-order` option to sort command. 

```
 $ cat /tmp/o |sort --dictionary-order
java-bigqueryconnection
java-bigquery-data-exchange
```

PR #8523, opened 2022-10-05
ci: check to ensure root pom.xml is synchronized with print_root_pom.sh

suztomo:
The check ensures that the root pom.xml file does not have manual diff from the generated file.

Screen Shot 2022-10-05 at 9 23 49 AM

This will avoid human mistakes such as https://github.com/googleapis/google-cloud-java/pull/8515/files

### DIFF
--- a/.github/workflows/generated_files_sync.yaml
+++ b/.github/workflows/generated_files_sync.yaml
@@ -1,0 +1,33 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


generation/print_root_pom.sh
```diff
@@ -5,7 +5,7 @@ set -e
 GENERATION_DIR=$(dirname -- "$0");
 
 # Find all Maven modules (a directory that contains pom.xml)
-find . -mindepth 2 -maxdepth 2 -name pom.xml |sort | xargs dirname \
+find . -mindepth 2 -maxdepth 2 -name pom.xml |sort --dictionary-order | xargs dirname \
```
suztomo, line 8: `sort` is platform dependent. This '--dictionary-order' fixes it. https://github.com/googleapis/google-cloud-java/issues/8520

PR #8554, opened 2022-10-07
chore: remove CoverageAggregator module

meltsufin:
We're not using this currently, and it's not clear if we'll ever want to track coverage for auto-generated libraries. So, removing it for now to simplify things.

### DIFF
--- a/generation/README.md
+++ b/generation/README.md
@@ -17,12 +17,6 @@ cd monorepo/google-cloud-java
 mvn test -T C1 -B
 ```
 
-To generate the coverage report, then run:
-```shell
-mvn jacoco:report-aggregate -T C1 -B
-```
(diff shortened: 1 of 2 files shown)


suztomo:
A comma has been removed?

image

.github/workflows/generated_files_sync.yaml
```diff
@@ -77,9 +77,6 @@ jobs:
       with:
         path: ~/.m2/repository
         key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.date }}
-    - name: Generate CoverageAggregator/pom.xml
-      run: |
-        bash generation/generate_coverage_aggregator.sh
     - name: Apply versions from versions.txt files
```
suztomo, line 80: Remove the entire `coverage-aggregator:` job.

meltsufin: good. catch!

suztomo (approved):

Issue #8565, opened 2022-10-10
set_versions should work cross-module dependencies


suztomo:
set_versions should work cross-module dependencies

meltsufin:
@suztomo Do you think it makes sense to have some test to validate that all annotated versions are in fact getting set?

PR #8566, opened 2022-10-10
chore: apply_current_versions.sh to take  cross-module dependency

suztomo:
Fixes https://github.com/googleapis/google-cloud-java/issues/8565

@meltsufin ~FYI, it has became really slow. During the meeting, I mentioned that it would be 10 mins but this naive implementation would take 2 hours to finish.~

~Measurement: It was 2 minutes. Now it's 23 minutes. We would need to build index of which pom.xml contain which keys in versions.txt.~

The inefficient per-module sed command is replaced with one big sed command that applies everything to every pom.xml. It now finishes in 1 minute.

### DIFF
--- a/generation/apply_current_versions.sh
+++ b/generation/apply_current_versions.sh
@@ -20,7 +20,7 @@ for versions_file in $(find . -mindepth 0 -maxdepth 2 -name versions.txt \
     SED_OPTIONS="$SED_OPTIONS -e /x-version-update:$K:current/{s|<version>.*<\/version>|<version>$V<\/version>|;}"
   done
 done
-pom_files="$(find . -maxdepth 3 -name pom.xml |sort --dictionary-order)"
-for P in $pom_files; do
-  sed -i.bak $SED_OPTIONS $P && rm ${P}.bak
-done
+
+echo "Running sed command. It may take few minutes."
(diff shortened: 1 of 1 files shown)


generation/apply_current_versions.sh
```diff
@@ -9,23 +9,18 @@
...
 
+SED_OPTIONS=""
+for versions_file in $(find . -mindepth 0 -maxdepth 2 -name versions.txt \
+    |sort --dictionary-order); do
   for KV in $(cut -f1,3 -d: $versions_file |grep -v "#"); do
     K=${KV%:*}; V=${KV#*:}
     echo Key:$K, Value:$V;
-    pom_files="$(find $MODULE -maxdepth 3 -name pom.xml) pom.xml CoverageAggregator/pom.xml google-cloud-gapic-bom/pom.xml"
-    for P in $pom_files; do
-      sed -i.bak -e "/x-version-update:$K:current/{s|<version>.*<\/version>|<version>$V<\/version>|;}" $P && rm ${P}.bak
```
suztomo, line 23: This `sed` in the nested for-loop was slow.

meltsufin (approved):
very impressive optimization!

PR #8586, opened 2022-10-12
chore: Use requirements.txt for CloudRAD

lqiu96:


.kokoro/release/publish_javadoc11.sh
```diff
@@ -27,6 +27,11 @@ fi
 # work from the git root directory
 pushd $(dirname "$0")/../../
 
+python3 --version
+
+# Force latest protobuf version
+python3 -m pip install protobuf
```
meltsufin, line 33: We should avoid using floating versions.
cc/ @mpeddada1

lqiu96: This was just to test our theory about not pulling the latest version of protobuf. Going to remove.

lqiu96:
Previous job:
image

This is using `gcp-docuploader==0.6.3`

ddixit14 (approved):

PR #8602, opened 2022-10-14
chore: Set doclet_version and module_list as env_vars

lqiu96:
MODULE_LIST allows us to enter a subset of modules to stage to Cloud RAD.

The related config change is in cl/481019109

PR #8603, opened 2022-10-14
chore: deploy to production environment

suztomo:
Wait for https://github.com/googleapis/google-cloud-java/pull/8602 is done and Lawrence confirms the effect.

### DIFF
--- a/generation/generate_release_please_config.sh
+++ b/generation/generate_release_please_config.sh
@@ -32,7 +32,7 @@ for path in $module_list; do
   # concatenating module name and module version
   rp_manifest_line="${tab}\"${module_name}\": \"${module_released_version}\""
   # Generate the JSON block with formatting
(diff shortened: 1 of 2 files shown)


.kokoro/release/stage.sh
```diff
@@ -35,11 +35,15 @@ mvn clean deploy -B \
...
-# During shadow-mode, do not release.
-#if [[ -n "${AUTORELEASE_PR}" ]]
-#then
-#  mvn nexus-staging:release -B \
-#    -DperformRelease=true \
-#    --settings=settings.xml
-#fi
+# The job triggered by Release Please (release-trigger) has this AUTORELEASE_PR
+# environment variable. Fusion also lets us to specify this variable.
+if [[ -n "${AUTORELEASE_PR}" ]]
```
meltsufin, line 40: If needed, we can also release manually after it's staged. It would allow us to do a final check.

lqiu96:
@suztomo Stage to devsite worked and no issue with all the java cls.

.kokoro/release/stage.sh
suztomo: As talked, I'll get your manual help in case it doesn't work.

meltsufin (approved):

PR #8606, opened 2022-10-14
chore: Fixing root pom

ddixit14:


suztomo (approved):
If some check fails, it means that you may need to modify the template of this pom.xml.

suztomo:
"Generated Files Sync / root pom.xml does not have diff from generated one" failed. Fix the template as well.

suztomo:
@ddixit14 Run the commands below:

```
bash generation/print_root_pom.sh > pom.xml
bash generation/apply_current_versions.sh
```

ddixit14:
Thanks Tomo. It passed now. added the automerge label.

generation/parent.pom.xml
```diff
@@ -22,15 +22,49 @@
...
+          <url>https://google.oss.sonatype.org/content/repositories/snapshots</url>
+        </snapshotRepository>
+        <repository>
+          <id>sonatype-nexus-staging</id>
+          <url>https://google.oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+        </repository>
+        <site>
+          <id>github-pages-site</id>
+          <name>Deployment through GitHub's site deployment plugin</name>
+          <url>site/google-cloud-bom</url>
```
meltsufin, line 49: @suztomo I don't think this is correct for the monorepo.

ddixit14: Sorry, correcting this.

meltsufin (approved):

ddixit14:
`Kokoro - Test: Integration` failed. A flake probably. Re-ran it.

PR #8635, opened 2022-10-18
chore: Add retry logic for the ci workflows

lqiu96:
Add retry logic for maven install to fix the random issues pulling maven dependencies from maven central.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -103,53 +103,6 @@ case ${JOB_TYPE} in
       echo "Not running GraalVM 17 checks -- No changes in relevant modules"
     fi
     ;;
-  samples)
-    generate_modified_modules_list
-    if [[ ${#modified_module_list[@]} -gt 0 ]]; then
-      module_list=$(
-        IFS=,
-        echo "${modified_module_list[*]}"
-      )
-      install_modules
-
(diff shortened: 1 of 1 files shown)


.kokoro/build.sh
```diff
@@ -92,58 +104,51 @@ case ${JOB_TYPE} in
     fi
     ;;
   samples)
```
meltsufin, line 106: You can delete this whole samples case. It's dead code.

lqiu96: Yeah, the samples aren't testing anything right now. I thought we were fine with keeping this in preparation for the handwritten libraries moving into the monorepo. Are we going to come up with different way to test samples for handwritten libraries and I should just remove this block?

meltsufin: Yeah, I think I said it's fine to keep, but seeing how complex the code is, I'd rather just get rid of it. We won't have to deal with handwritten samples in the future.

lqiu96: Ok. I'll remove it.

lqiu96: I've kept the samples.cfg for the kokoro job in for now. It just currently does nothing -- I'm considering deleting this config  and in g3 so it doesn't waste resources in our pool.

lqiu96:
@meltsufin Could you take another look when you get a chance?

meltsufin (approved):

PR #8637, opened 2022-10-19
chore: consolidate build configuration to parent

meltsufin:
Moving this block from jar modules to jar-parent.
```xml
  <build>
    <plugins>
      <plugin>
          <groupId>org.codehaus.mojo</groupId>
          <artifactId>flatten-maven-plugin</artifactId>
        </plugin>
    </plugins>
  </build>
```

Removing this block from module parents. Since `dependencies.sh` is not run at all, it's not clear if these exclusions are still needed.
```xml
  <build>
    <pluginManagement>
      <plugins>
        <plugin>
          <artifactId>maven-dependency-plugin</artifactId>
          <configuration>
            <ignoredUnusedDeclaredDependencies>
              <ignoredUnusedDeclaredDependency>org.objenesis:objenesis</ignoredUnusedDeclaredDependency>
              <ignoredUnusedDeclaredDependency>javax.annotation:javax.annotation-api</ignoredUnusedDeclaredDependency>
            </ignoredUnusedDeclaredDependencies>
          </configuration>
        </plugin>
      </plugins>
    </pluginManagement>
  </build>
```

Removing `javax.annotation` declaration for Java 9+, since the dependency is already present transitively through `api-common`.
```
  <profiles>
    <profile>
      <id>java9</id>
      <activation>
        <jdk>[9,)</jdk>
      </activation>
      <dependencies>
        <dependency>
          <groupId>javax.annotation</groupId>
          <artifactId>javax.annotation-api</artifactId>
        </dependency>
      </dependencies>
    </profile>
  </profiles>
```

### DIFF
--- a/google-cloud-jar-parent/pom.xml
+++ b/google-cloud-jar-parent/pom.xml
@@ -121,6 +121,23 @@
     </dependencies>
   </dependencyManagement>
 
+  <profiles>
+    <profile>
+      <id>java9</id>
+      <activation>
+        <jdk>[9,)</jdk>
+      </activation>
+      <dependencies>
+        <!-- needed for @Generated annotation at compile-time -->
+        <dependency>
+          <groupId>javax.annotation</groupId>
+          <artifactId>javax.annotation-api</artifactId>
+          <scope>provided</scope>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
+
   <build>
     <plugins>
       <!-- jacoco code coverage -->


google-cloud-pom-parent/pom.xml
```diff
@@ -56,4 +57,33 @@
     </license>
   </licenses>
 
+  <profiles>
+    <profile>
+      <!-- Only run checkstyle plugin on Java 11+ (checkstyle artifact only supports Java 11+) -->
+      <id>checkstyle-tests</id>
+      <activation>
+        <jdk>[1.11,)</jdk>
```
suztomo, line 65:
```suggestion
        <jdk>[11,)</jdk>
```

suztomo:
> Removing javax.annotation declaration for Java 9+, since the dependency is already present transitively through api-common.

It's a good practice to declare the dependency if a project touches a class from it, without relying on transitive dependencies (["Undeclared but used"](https://maven.apache.org/plugins/maven-dependency-plugin/examples/exclude-dependencies-from-dependency-analysis.html)). Can you move the javax.annotation part to google-cloud-jar-parent?

meltsufin:
> > Removing javax.annotation declaration for Java 9+, since the dependency is already present transitively through api-common.
> 
> It's a good practice to declare the dependency if a project touches a class from it, without relying on transitive dependencies (["Undeclared but used"](https://maven.apache.org/plugins/maven-dependency-plugin/examples/exclude-dependencies-from-dependency-analysis.html)). Can you move the javax.annotation part to google-cloud-jar-parent?

Since we flatten the poms, I guess it will be declared. Declaring and not using a dependency is also a bad practice. So, I'm torn on this one.

suztomo:
> Since we flatten the poms, I guess it will be declared. 

That's correct, it will be declared anyway. However, skipping dependency declaration (for the flatten plugin's effect) means we can/should remove lots of other dependencies as well. The reasoning would become complex. For example, you had to reason "_javax-annotation dependency is already present transitively through api-common_". I don't want people spending time on such dependency analysis to decide whether a dependency should be declared. Therefore, I think we should reason about dependency declaration without considering the effect of the flatten plugin.

suztomo (approved):

Issue #8640, opened 2022-10-19
Most of integration test times are unnecessary


suztomo:
Most of integration test times are unnecessary. For example, even when we don't touch java-compute or java-aiplatform module, these test run.

When reducing the integration test run, we need to be careful about the changes in the pom-parent and jar-parent. The changes there must be validated by the integration tests.

# How to see the diff only in the pull request

`git diff  -r origin/main...` 

What's the three dots? It's the diff from the common ancestor https://matthew-brett.github.io/pydagogue/git_diff_dots.html.

PR #8641, opened 2022-10-19
chore: Run ITs on diffs between latest and the branching commit

lqiu96:
Tries to fix: https://github.com/googleapis/google-cloud-java/issues/8640

Only run AI-Platform/ Compute Native Tests when there are changes to those modules.


meltsufin:
I think you also need to fix the root pom logic. The root pom is now in `google-cloud-jar-parent`.

lqiu96:
> I think you also need to fix the root pom logic. The root pom is now in `google-cloud-jar-parent`.

Ah nice catch! I'll update that.

meltsufin:
`google-cloud-pop-parent` too probably.

lqiu96:
> `google-cloud-pop-parent` too probably.

Ah yep, I see that shared-config is in `pom-parent` and not `jar-parent`. Adding in now.

suztomo:
Does this take care of native image test?

lqiu96:
> Does this take care of native image test?

Presubmit native tests only run AI Platform or Compute. The nightly job runs all modules and doesn't use the existing diff logic.

meltsufin:
> > Does this take care of native image test?
> 
> Presubmit native tests only run AI Platform or Compute. The nightly job runs all modules and doesn't use the existing diff logic.

I guess @suztomo is asking whether these would be skipped unless there are changes to the parent poms or compute/aiplatform.

lqiu96:
> > > Does this take care of native image test?
> > 
> > 
> > Presubmit native tests only run AI Platform or Compute. The nightly job runs all modules and doesn't use the existing diff logic.
> 
> I guess @suztomo is asking whether these would be skipped unless there are changes to the parent poms or compute/aiplatform.

Ah I see, I was resolving a different issue with the diffs. I'll add in the diffs for native tests

suztomo:
Thanks. Yes, I feel most of the presubmit native image test run (java-compute and java-aiplatform) in most pull requests are unnecessary.

lqiu96:
> Thanks. Yes, I feel most of the presubmit native image test runs (java-compute and java-aiplatform) in most pull requests are unnecessary (#8640).

Yeah, I agree that the module specific native checks don't need to be run on every PR, especially ones that don't affect Compute/ AI-Platform. With this change in, we should pay more attention to the nightly builds.

suztomo (approved):

PR #8652, opened 2022-10-21
chore: Add META-INF/* to owlbot deep-preserve-regex

lqiu96:
[WIP] 

Add META-INF/* to owlbot `deep-preserve-regex` for each module. Some GraalVM checks are failing due to owlbot post processor removing these files.

i.e. This PR #8493 for java-speech removed this file: java-speech/google-cloud-speech/src/test/resources/META-INF/native-image/com.google.cloud/google-cloud-speech/resource-config.json and the GraalVM checks for java-speech has been failing since.

These files were generated with a one time script using `yq` to add in these two lines under `deep-preserve-regex`. 

Considered writing a script to re-regenerate the owlbot.yaml file using the python code in new_client/*. This mostly works except for certain modules that already have custom files (it'll remove those and only apply the template).

Doing a `find . -name 'META-INF'` pulls up `java-speech` as the only module requiring this file. Adding this config for every module (plus in the template) just to ensure that any existing module (and any future modules) won't run into these GraalVM resource errors.

### DIFF
--- a/generation/new_client/templates/owlbot.yaml.monorepo.j2
+++ b/generation/new_client/templates/owlbot.yaml.monorepo.j2
@@ -19,8 +19,8 @@ deep-remove-regex:
 - "/{{ module_name }}/google-.*/src"
 deep-preserve-regex:
 - "/{{ module_name }}/google-.*/src/test/java/com/google/cloud/.*/v.*/it/IT.*Test.java"
-- "/{{ module_name }}/google-.*/src/main/resources/META-INF/native-image/*"
-- "/{{ module_name }}/google-.*/src/test/resources/META-INF/native-image/*"
+- "/{{ module_name }}/google-.*/src/main/resources/META-INF/native-image/.*"
+- "/{{ module_name }}/google-.*/src/test/resources/META-INF/native-image/.*"
 
 deep-copy-regex:
 - source: "/{{ proto_path }}/(v.*)/.*-java/proto-google-.*/src"
--- a/java-accessapproval/.OwlBot.yaml
+++ b/java-accessapproval/.OwlBot.yaml
@@ -33,4 +33,4 @@ deep-copy-regex:
 - source: "/google/cloud/accessapproval/(v.*)/.*-java/samples/snippets/generated"
   dest: "/owl-bot-staging/java-accessapproval/$1/samples/snippets/generated"
 
-api-name: "accessapproval"
+api-name: accessapproval
--- a/java-accesscontextmanager/.OwlBot.yaml
+++ b/java-accesscontextmanager/.OwlBot.yaml
@@ -30,4 +30,4 @@ deep-copy-regex:
 - source: "/google/identity/accesscontextmanager/type/.*-java/proto-google-.*/src"
   dest: "/owl-bot-staging/java-accesscontextmanager/type/proto-google-identity-accesscontextmanager-type/src"
 
-api-name: "accesscontextmanager"
+api-name: accesscontextmanager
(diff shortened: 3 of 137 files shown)


suztomo:
Do you plan to keep OwlBot.yaml files generated periodically?

- If not, change only java-speech.
- If yes, then add the script (that uses the yq command) and add GitHub Actions workflow to ensure there's no diff from auto-generated files and in a pull request.

generation/new_client/templates/owlbot.yaml.monorepo.j2
```diff
@@ -17,10 +17,10 @@ deep-remove-regex:
 - "/{{ module_name }}/grpc-google-.*/src"
 - "/{{ module_name }}/proto-google-.*/src"
 - "/{{ module_name }}/google-.*/src"
-
```
suztomo, line 20: Keep the empty line.

generation/new_client/templates/owlbot.yaml.monorepo.j2
```diff
@@ -17,10 +17,10 @@ deep-remove-regex:
...
 deep-preserve-regex:
 - "/{{ module_name }}/google-.*/src/test/java/com/google/cloud/.*/v.*/it/IT.*Test.java"
-
+- "/{{ module_name }}/google-.*/src/main/resources/META-INF/native-image/*"
```
suztomo, line 22:
```suggestion
- "/{{ module_name }}/google-.*/src/main/resources/META-INF/native-image/.*"
```

This is not glob pattern. It's a regular expression.

java-accessapproval/.OwlBot.yaml
```diff
@@ -12,24 +12,22 @@
...
+  - dest: /owl-bot-staging/java-accessapproval/$1/samples/snippets/generated
+    source: /google/cloud/accessapproval/(v.*)/.*-java/samples/snippets/generated
+deep-preserve-regex:
+  - /java-accessapproval/google-cloud-accessapproval/src/test/java/com/google/cloud/accessapproval/v1/it
+  - /java-accessapproval/google-.*/src/main/resources/META-INF/native-image/*
+  - /java-accessapproval/google-.*/src/test/resources/META-INF/native-image/*
+deep-remove-regex:
+  - /java-accessapproval/grpc-google-.*/src
+  - /java-accessapproval/proto-google-.*/src
+  - /java-accessapproval/google-.*/src
```
suztomo, line 32: Can you keep the order of the sections? The old order (deep-remove-regex, deep-preserve-regex, and deep-copy-regex) matches OwlBot's steps go/owlbot-howto#githubowlbotyaml.

lqiu96: I see. I can keep the order for clarity of the owl-bot steps.

generation/new_client/templates/owlbot.yaml.monorepo.j2
lqiu96: Ah gotcha. Will fix.

generation/new_client/templates/owlbot.yaml.monorepo.j2
lqiu96: Will fix.

lqiu96:
> Do you plan to keep OwlBot.yaml files generated periodically?
> 
> * If not, change only java-speech.
> * If yes, then add the script (that uses the yq command) and add GitHub Actions workflow to ensure there's no diff from auto-generated files and in a pull request.

This is generated on the bootstrap/build CI job, right? If we add new custom logic in the `.Owlbot.yaml` file in the monorepo, wouldn't it still fail since it's diffing with the `.Owlbot.yaml` file from the split repos?

suztomo:
> diffs from the split repos

No, I'm not referencing the diff from split repos. In future, when OwlBot.yaml is manually modified in future and somebody didn't update the source, then I want the pull request fail.s

lqiu96:
Closing in favor of https://github.com/googleapis/google-cloud-java/pull/8658

Issue #8659, opened 2022-10-21
Add a Github Action check for new-client.py


meltsufin:
It should be similar to the bootstrap check where it generates a client library and tests whether it compiles.
We should use a client library that is not in the repository, such as the Showcase.

PR #8670, opened 2022-10-23
chore: add test for new client generation

meltsufin:
Fixes: #8659

### DIFF
--- a/.github/workflows/newclient.yaml
+++ b/.github/workflows/newclient.yaml
@@ -47,9 +47,9 @@ jobs:
           sed -i "/${API_SHORT_NAME}/d" gapic-libraries-bom/pom.xml
       - name: Verify Client Removal
         run: |
-          test -d java-${API_SHORT_NAME} && echo "ERROR: Found java-${API_SHORT_NAME} directory" && exit 1
-          grep -q ${API_SHORT_NAME} pom.xml && echo "ERROR: Found ${API_SHORT_NAME} in pom.xml" && exit 1
-          grep -q ${API_SHORT_NAME} gapic-libraries-bom/pom.xml && echo "ERROR: Found ${API_SHORT_NAME} in gapic-libraries-bom/pom.xml" && exit 1
+          ! test -d java-${API_SHORT_NAME}
+          ! grep -q ${API_SHORT_NAME} pom.xml
+          ! grep -q ${API_SHORT_NAME} gapic-libraries-bom/pom.xml
       - name: Generate
         run: |
           python generation/new_client/new-client.py generate \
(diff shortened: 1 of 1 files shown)


suztomo:
> because the repository is private.

How about yoshi-code-bot account? https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/readme.yaml#L35 (I don't know its entitlement)

meltsufin:
> > because the repository is private.
> 
> How about yoshi-code-bot account? https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/readme.yaml#L35 (I don't know its entitlement)

Doesn't seem to be working either.

.github/workflows/newclient.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Push to branch
+        run: |
+          [ -z "`git config user.email`" ] && git config --global user.email "${USERNAME:-script}@google.com"
+          [ -z "`git config user.name`" ] && git config --global user.name "${USERNAME:-script}"
+          git checkout -b newclient_output_${{ github.event_name }}
+          git add --all
+          git commit -m 'feat: re-generated apikeys'
+          git remote add monorepo https://${{ github.actor }}:${{ github.token }}@github.com/${{ github.repository }}.git
+          git fetch -q --unshallow monorepo
+          git push -f monorepo newclient_output_${{ github.event_name }}
```
suztomo, line 58: How does "Push to branch" help verify the behavior of new-client.py?

.github/workflows/newclient.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Test
+        run: mvn test -T C1 -B
```
suztomo, line 60: Can you add source code comment that "java-apikeys is included at the aggregator pom.xml"?

.github/workflows/newclient.yaml
```diff
@@ -0,0 +1,60 @@
...
+          java-version: 11
+      - run: java -version
+      - uses: actions/cache@v3
+        id: mvn-cache
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.date }}
+      - name: Install dependencies
+        run: pip install -r generation/new_client/requirements.txt
+      - name: Generate
```
suztomo, line 40: Can you add source code comment (or name) "regenerating existing apikeys module"?

.github/workflows/newclient.yaml
suztomo: How about "-f java-apikeys/pom.xml" to skip irrelevant modules?

.github/workflows/newclient.yaml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          python generation/new_client/new-client.py generate \
```
suztomo, line 42: Can you explicitly remove java-apikeys module before new-client.py?

suztomo:
How might we verify new-client.py’s logic of adding a corresponding new entry to gapic BOM, release please configuration files, and the root pom.xml?

.github/workflows/newclient.yaml
meltsufin: It's just there for debugging, in case the job fails.

suztomo: That's helpful. Nice.

.github/workflows/newclient.yaml
meltsufin: Yep

.github/workflows/newclient.yaml
meltsufin: I do that in new-client.py. Is that not enough?

suztomo (approved):

.github/workflows/newclient.yaml
suztomo: For me it's itchy that if, by any chahce, new-client.py does nothing, the verification in this check still passes.

suztomo: It's done below:

```
      - name: Remove Client
        run: |
          rm -rf java-${API_SHORT_NAME} || true
          sed -i "/${API_SHORT_NAME}/d" pom.xml
          sed -i "/${API_SHORT_NAME}/d" gapic-libraries-bom/pom.xml
``

PR #8677, opened 2022-10-24
chore: add all modules to the gapic-libraries-bom

meltsufin:


suztomo (approved):

suztomo:
Comparing options:

- 1 bom to include everything
- 1 bom for non-GA and 1 bom for stable

The 1st option is simpler and the users don't have to worry about whether their library is in this bom or that bom.

PR #8740, opened 2022-11-08
fix: Add centralized retries to all java-vision integration tests. Simplify.

burkedavison:


meltsufin (approved):
Nice simplification, but I do wonder if we can do something better than 30 second delays between tries. For another time though.

PR #8748, opened 2022-11-08
chore: Script to generate root versions.txt

ddixit14:
This script generates versions.txt at the root. This is required for single-component setup. 
File generated by the script: [versions.txt](https://github.com/googleapis/google-cloud-java/files/9965592/versions.txt)

### DIFF
--- a/generation/generate_root_versions_txt.sh
+++ b/generation/generate_root_versions_txt.sh
@@ -4,6 +4,8 @@ set -e
 
 GENERATION_DIR=$(dirname -- "$0")
 
+rm -f versions.txt
+
 echo "# Format:" >> versions.txt
 echo "# module:released-version:current-version" >> versions.txt
 


ddixit14:
This is the file generated by the script: [versions.txt](https://github.com/googleapis/google-cloud-java/files/9965592/versions.txt)

meltsufin (approved):

generation/generate_root_versions_txt.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+GENERATION_DIR=$(dirname -- "$0")
+
+echo "# Format:" >> versions.txt
```
suztomo, line 7: Why is it appending? Is there a logic to empty the file first?

suztomo:
Do you think we will run this script periodically?

If this is one-time change, I don't think you need to commit it to the repository.

suztomo (approved):

suztomo:
jar-parent is there. Good.

meltsufin:
> Do you think we will run this script periodically?

> If this is one-time change, I don't think you need to commit it to the repository. (no harm adding it; just remember to remove it few months later)

We might be adding more split repos to this repository. So, it might be a good idea to keep scripts like this one for the future.

ddixit14:
> Do you think we will run this script periodically?
> 
> If this is one-time change, I don't think you need to commit it to the repository. (no harm adding it; just remember to remove it few months later)

We will need to run this everytime a new library/module is added to the monorepo.

generation/generate_root_versions_txt.sh
ddixit14: Oh yes, I forgot to remove the file if it exists before creating it. fixed now.

PR #8753, opened 2022-11-09
chore: Use shared-config doclet version as default

lqiu96:
Removing this kokoro env_var version since we can have `shared-configs` determine the doclet version (https://github.com/googleapis/java-shared-config/blob/29e20eb8809fc3dc985111bbaecd09c281222484/pom.xml#L698)
```
env_vars: {
  key: "DOCLET_VERSION"
  value: "1.7.0"
}
```

I'll keep the config in g3 so that we can override this env_var with any custom doclet version in any future fusion job.

### DIFF
--- a/.kokoro/release/publish_javadoc11.cfg
+++ b/.kokoro/release/publish_javadoc11.cfg
@@ -17,11 +17,6 @@ env_vars: {
   value: "github/google-cloud-java/.kokoro/release/publish_javadoc11.sh"
 }
 
-env_vars: {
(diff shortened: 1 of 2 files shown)


suztomo (approved):
Approving. But remember that the environment variables and newly added if-statements make troubleshooting a slightly difficult and might introduce new problem later.

When it was unclear what to do, these ad-hoc variables made sense. When we know the infrastructure setup is stable, it may make sense to remove the variables and if-statements altogether, in favor of a proper, standard process to update the doclet version.

PR #8755, opened 2022-11-09
chore: Single component setup

ddixit14:
This PR changes release-please manifest setup to single-component setup.


suztomo (approved):

suztomo:
I thought "generation diff / release-please-config" would fail.

suztomo:
"generation diff / root-pom" filed:

```
Running sed command. It may take few minutes.
xargs: cannot fit single argument within argument list size limit
```

ddixit14:
> I thought "generation diff / release-please-config" would fail.

Now it is failing. But it is not required, we don't need it anyway now. We can delete it later on.

suztomo:
Right. These files are no longer needed

Screen Shot 2022-11-09 at 3 44 19 PM

PR #8757, opened 2022-11-09
build: enforcer to use local Maven repository cache

suztomo:
With this pull request

- "enforcer" is a required check.
- the cache is refreshed weekly.
- newclient.yaml also uses Maven cache (speed up).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,8 +98,8 @@ jobs:
       with:
         path: ~/.m2/repository
         key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.week_of_year }}
-      - run: java -version
-      - run: mvn -B -ntp enforcer:enforce@enforce -T 1C
+    - run: java -version
+    - run: mvn -B -ntp enforcer:enforce@enforce -T 1C
 
 # TODO: Uncomment the needed Github Actions
 #  dependencies:


meltsufin (approved):
Do we really need to rebuild the cache at all?
Also, maybe we can use this built-in option: https://github.com/actions/setup-java#caching-maven-dependencies.

suztomo:
I want to control cache expiration explicitly so that the cache lifecycle is clear.

The build-in cache is too sensitive for our use case; Any change on pom.xml files make cache misses.

> The format of the used cache key is setup-java-${{ platform }}-${{ packageManager }}-${{ fileHash }}, where the hash is based on the following files:
> gradle: **/*.gradle*, **/gradle-wrapper.properties, buildSrc/**/Versions.kt, buildSrc/**/Dependencies.kt, and gradle/*.versions.toml
> maven: **/pom.xml

meltsufin:
> I want to control cache expiration explicitly so that the cache lifecycle is clear.
> 
> The build-in cache is too sensitive for our use case; Any change on pom.xml files make cache misses.
> 
> > The format of the used cache key is setup-java-${{ platform }}-${{ packageManager }}-${{ fileHash }}, where the hash is based on the following files:
> > gradle: **/_.gradle_, **/gradle-wrapper.properties, buildSrc/**/Versions.kt, buildSrc/**/Dependencies.kt, and gradle/*.versions.toml
> > maven: **/pom.xml

Indeed that is to sensitive, if you can't make use of `restore-keys` on cache misses.
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-multiple-restore-keys

PR #8769, opened 2022-11-10
chore: use gcp-docuploader 0.6.3

mpeddada1:
Using the latest version of gcp-docuploader (0.6.4) was resulting in:
```
Traceback (most recent call last):
  File "/root/.pyenv/versions/3.6.12/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/root/.pyenv/versions/3.6.12/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/root/.pyenv/versions/3.6.12/lib/python3.6/site-packages/docuploader/__main__.py", line 26, in <module>
    from docuploader.protos import metadata_pb2
  File "/root/.pyenv/versions/3.6.12/lib/python3.6/site-packages/docuploader/protos/metadata_pb2.py", line 5, in <module>
    from google.protobuf.internal import builder as _builder
ImportError: cannot import name 'builder'
cleanup
```

It looks like the latest version requires a newer version of the protobuf dependency (See: https://github.com/googleapis/docuploader/pull/137/files) . However versions greater than 3.19.6 are not compatible with Python 3.6.


lqiu96 (approved):
Tested locally and it works

suztomo:
@mpeddada1 Would you setup a job to detect such failure in “pip install”? Lawrence sees we use Python 3.6 here.

PR #8776, opened 2022-11-14
chore: excluding java-maps-* modules

suztomo:
GroupId needs to be discussed further.

### DIFF
--- a/generation/print_root_pom.sh
+++ b/generation/print_root_pom.sh
@@ -8,6 +8,4 @@ GENERATION_DIR=$(dirname -- "$0");
 find . -mindepth 2 -maxdepth 2 -name pom.xml |sort --dictionary-order | xargs dirname \
     |sed -e 's|./||' | xargs -I '{}' echo "    <module>{}</module>" > /tmp/repo-modules.txt
 
-sed -i.bak '/java-maps/d' /tmp/repo-modules.txt
-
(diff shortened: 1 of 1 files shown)


meltsufin (changes requested):
@suztomo Can you explain why?
This is going to skip the entire module when doing operations from the project root.

meltsufin:
@suztomo Can you just add skip deploy config?

suztomo:
@meltsufin I prefer removing the module from the root to figuring out the proper configuration to exclude. It's simpler than modifying pom.xml files. As you know Maven nexus deploy configuration is not simple. (Question about "-bom", "grpc-", and "proto-" artifacts).

generation/print_root_pom.sh
```diff
@@ -8,4 +8,6 @@ GENERATION_DIR=$(dirname -- "$0");
 find . -mindepth 2 -maxdepth 2 -name pom.xml |sort --dictionary-order | xargs dirname \
     |sed -e 's|./||' | xargs -I '{}' echo "    <module>{}</module>" > /tmp/repo-modules.txt
 
+sed -i.bak '/java-maps/d' /tmp/repo-modules.txt
```
meltsufin, line 11: No need for this anymore, right?

suztomo: removed.

PR #8778, opened 2022-11-14
chore: Manifest setup with single root component

ddixit14:


suztomo (approved):
Adding stamp but wait for Jeff

release-please-config.json
```diff
@@ -0,0 +1,12 @@
...
+  "release-type": "java-yoshi",
+  "versioning": "always-bump-minor",
+  "separate-pull-requests": false,
+  "include-component-in-tag": false,
+  "packages": {
+    ".": {
+      "component": "main"
+    }
+  },
+  "plugins": ["maven-workspace"]
```
chingor13, line 11: You should skip this plugin if there's only a single component unless you have pom.xml files without annotations.

It shouldn't hurt but will make extra API calls and do extra work to build the dependency tree.

chingor13 (approved):

release-please-config.json
ddixit14: @meltsufin, I remember you fixed all the pom.xml's to have the annotations. Should I remove this plugin then? or better to have it around for now?

meltsufin: All poms are fixed to have annotations and there is a check to verify it.

Issue #8767, opened 2022-11-10
Ensure generated samples directory is being copied over from googleapis-gen


alicejli:
Some modules do not have the generated samples directory copied over from googleapis-gen. For instance see `java-container` in googleapis-gen: https://github.com/googleapis/googleapis-gen/commit/73ec3348990fb308d3b600a0940ae852d47247ee, but the same PR in the monorepo doesn't contain the `samples/snippets/generated` directory: https://github.com/googleapis/java-container/pull/848

This is because some modules are missing the correct copy/remove regexes in their `.OwlBot.yaml` files. 

deep-remove-regex:
- "/samples/snippets/generated"

deep-copy-regex:
- source: "/google/cloud/${module_name}/(v.*)/.*-java/samples/snippets/generated"
  dest: "/owl-bot-staging/${module_name}/$1/samples/snippets/generated"

See https://github.com/google-cloud-java/blob/main/java-apigee-registry/.OwlBot.yaml  as an example of a correct file.

The new client generation file should get updated as well: https://github.com/googleapis/google-cloud-java/blob/main/generation/new_client/templates/owlbot.yaml.monorepo.j2

PR #8790, opened 2022-11-15
fix: update owlbot configs to copy generated samples

alicejli:
Fixes https://github.com/googleapis/google-cloud-java/issues/8767

Tested with local Owlbot invocation using the following command (using `java-analyticshub` as the example):

```
docker run --rm --user $(id -u):$(id -g) \
  -v $(pwd):/repo -v /tmp/googleapis-gen:/googleapis-gen -w /repo \
  --env HOME=/tmp \
  gcr.io/cloud-devrel-public-resources/owlbot-cli:latest copy-code \
  --source-repo=/googleapis-gen \
  --config-file=java-analyticshub/.OwlBot.yaml
```

### DIFF
--- a/generation/merge_repository.sh
+++ b/generation/merge_repository.sh
@@ -118,7 +118,8 @@ git add --all
 git commit -am 'chore: remove and disable owlbot postprocessor templates' \
     --allow-empty
 
-for F in `find . -maxdepth 2 -name '.OwlBot.yaml'`; do sh ../../set_owlbot_config.sh $F; done
+../../set_owlbot_config.sh
+
 git commit -am 'chore: set owlbot copy config' --allow-empty
 
 # create a monorepo/diff repo


generation/update_owlbot_config.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Usage:
+# update_owlbot_config.sh <.OwlBot.yaml path from the root>
+# Example:
+# $ update_owlbot_config.sh java-dataform/.OwlBot.yaml
+#
+# To apply the change to all OwlBot configuration files in all modules:
+# $ for F in `find . -maxdepth 2 -name '.OwlBot.yaml'`; do sh generation/update_owlbot_config.sh $F; done
+
+OWLBOT_FILE=$1
```
meltsufin, line 13: We already have `set_owlbot_config.sh`. Can we just update that script instead of adding another one?

alicejli: That makes sense. I kept it separately initially for testing purposes, but I didn't run into any issues combining.  I've merged the two into `set_owlbot_config.sh`.

meltsufin (changes requested):
Please merge in main and ensure that all ci checks are green.

generation/set_owlbot_config.sh
```diff
@@ -42,3 +42,33 @@ sed -i.bak "s|\"/samples|\"/${module_name}/samples|" "${OWLBOT_FILE}" && rm "${O
...
+  # Remove duplicate lines
+  perl -i -ne 'if ( /^\s*#/ ) { print } else { print if ! $SEEN{$_}++}' ${OWLBOT_FILE}
+
+  # Add back new lines between sections
+  sed -i  's/deep-copy-regex/\n&/g'  ${OWLBOT_FILE}
+  sed -i  's/deep-remove-regex/\n&/g'  ${OWLBOT_FILE}
+  sed -i  's/deep-preserve-regex/\n&/g'  ${OWLBOT_FILE}
+  sed -i  's/api-name/\n&/g'  ${OWLBOT_FILE}
+
+  fi
```
meltsufin, line 73: nit: shift left

generation/set_owlbot_config.sh
```diff
@@ -42,3 +42,33 @@ sed -i.bak "s|\"/samples|\"/${module_name}/samples|" "${OWLBOT_FILE}" && rm "${O
...
   sed -i.bak "s|owl-bot-staging|owl-bot-staging/${module_name}|" "${OWLBOT_FILE}" && rm "${OWLBOT_FILE}".bak
 fi
+
+# This section is specifically around the generated snippet directories
+# If snippets are already being copied, skip
+if ! grep -q samples/snippets/generated ${OWLBOT_FILE}; then
+  # Insert into `deep-remove-regex:` section
+  deep_remove_regex="- \"\/${module_name}\/samples\/snippets\/generated\""
+  entry_before_deep_remove_regex="${module_name}\/google-.*\/src"
+  sed -i "/${entry_before_deep_remove_regex}/a ${deep_remove_regex}" ${OWLBOT_FILE}
```
meltsufin, line 52: Please use `-i.bak` like above to make it compatible with MacOS.

meltsufin (changes requested):
How did you apply the script? Did you run this?
```
or F in `find . -maxdepth 2 -name '.OwlBot.yaml'`; do sh generation/set_owlbot_config.sh $F; done
```
I see some small differences when I run this command on the branch.

Also, would you mind adding a test to `generated_files_sync.yaml` for this script?

alicejli:
> How did you apply the script? Did you run this?
> 
> ```
> for F in `find . -maxdepth 2 -name '.OwlBot.yaml'`; do sh generation/set_owlbot_config.sh $F; done
> ```
> 
> I see some small differences when I run this command on the branch.

Yes, that's what I ran. I had to manually edit  `java-grafeas/.OwlBot.yaml` and `java-accesscontextmanager/.OwlBot.yaml` because they had an exceptional pattern compared to the other ones. Were those the same ones you noticed differences in?

> 
> Also, would you mind adding a test to `generated_files_sync.yaml` for this script?

Sorry  I'm having trouble finding where that file is - do you mind linking? Thank you!

alicejli:
Chatted with Mike offline that the test should be added here: https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/generated_files_sync.yaml

alicejli:
@meltsufin I chatted with @suztomo about this and it seemed the best way to test this was to run owlbot locally to confirm the correct directories get copied over. I did that and confirmed the updated .OwlbotYaml files are working as intended, but we're not sure of a good way to actually add a test for these files via a script. 

Do you have an idea of how you'd like to add a test? Or is testing Owlbot locally on this PR sufficient? Thanks!

meltsufin:
> @meltsufin I chatted with @suztomo about this and it seemed the best way to test this was to run owlbot locally to confirm the correct directories get copied over. I did that and confirmed the updated .OwlbotYaml files are working as intended, but we're not sure of a good way to actually add a test for these files via a script.
> 
> Do you have an idea of how you'd like to add a test? Or is testing Owlbot locally on this PR sufficient? Thanks!

The manual test is good, but I was talking more about the kind of tests that `generated_files_sync.yaml` has, where you just run the script (`generation/set_owlbot_config.sh`) for all modules and verify that no files have changed as a side-effect.

alicejli:
> The manual test is good, but I was talking more about the kind of tests that `generated_files_sync.yaml` has, where you just run the script (`generation/set_owlbot_config.sh`) for all modules and verify that no files have changed as a side-effect.

Ahh I think I understand - I added a check to run `set_owlbot_config.sh` in `generated_files_sync.yaml`. Do you mind taking another look to see if that is sufficient?

generation/check_owlbot_config.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Checks that each module's .OwlbotYaml file is correctly configured according to the template set in `set_owlbot_config.sh`
+
+OWLBOT_FILE=$1
+for F in $(find . -maxdepth 2 -name '.OwlBot.yaml'); do sh generation/set_owlbot_config.sh "$F"; done
```
meltsufin, line 6: Since this actually modifies the files, and doesn't really check, I think the file name doesn't really make sense. Would it make sense to just add this loop to `set_owlbot_config.sh` and update places where it's used? I think we always want to apply this to all `.OwlBot.yaml` files.

.github/workflows/generated_files_sync.yaml
```diff
@@ -59,6 +59,17 @@ jobs:
       - name: Fail if there's any difference
         run: git --no-pager diff --exit-code
 
+  owlbot-yaml:
```
meltsufin, line 62: Yes, this is what I had in mind. Thanks!

meltsufin (changes requested):

generation/check_owlbot_config.sh
alicejli: Hmm I see what you're saying; I did a little more digging and it looks like `set_owlbot_config.sh` is actually run as part of `merge_repository.sh` (https://github.com/googleapis/google-cloud-java/blob/main/generation/merge_repository.sh#L121) which is run regularly as a workflow: https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/merge_repository.yaml#L22. 

My sense is that actually would suffice as the check for the Owlbot.yaml files - what do you think?

meltsufin: That check doesn't apply to the entire repo and might go away in the future. I think the check your added is valuable beyond that.

alicejli: Thanks for pointers offline - let me know if this update works!

meltsufin: I see that you updated `merge_repository.sh`, but `bootstrap.sh` still needs an update.

meltsufin: Right, never mind! I think it's still in my local for some reason.

meltsufin (approved):

Issue #8574, opened 2022-10-11
Remove generation/bom.pom.xml and generation/parent.pom.xml


meltsufin:
These templates can be replaced with reliance on the actual files in `/pom.xml` and `/google-cloud-gapic-bom/pom.xml`. The scripts will have to change to deal with the existing modules list.

PR #8800, opened 2022-11-16
chore: remove parent and bom templates

meltsufin:
Use the actual files as templates instead.

Fixes: #8574.

### DIFF
--- a/generation/bootstrap.sh
+++ b/generation/bootstrap.sh
@@ -100,6 +100,7 @@ cp ../../gapic_bom_versions.txt gapic-libraries-bom/versions.txt
 git add --all
 git commit -am 'chore: delete non-auto-generated samples'
 
+../../generate_root_versions_txt.sh
 ../../update_versions.sh -s
 ../../apply_current_versions.sh
 


generation/generate_gapic_bom.sh
```diff
@@ -54,5 +54,7 @@ done
 mkdir -p gapic-libraries-bom
 
 GENERATION_DIR=$(dirname -- "$0");
+perl -0pe 's/<dependencies>.*<\/dependencies>/<dependencies>\nBOM_ARTIFACT_LIST\n    <\/dependencies>/s' ${GENERATION_DIR}/../gapic-libraries-bom/pom.xml > ${GENERATION_DIR}/bom.pom.xml
```
suztomo, line 57: It's strange to use gapic libraries BOM to generate gapic libraries BOM. Why do you have remove the template?

meltsufin: We've had several instances where the bom and parent pom checks started failing because changes were made without updating the templates. It's more accurate to call these "update" rather than "generate" scripts, which is really what we're doing when adding a new library. If you take that perspective, I think it makes sense to not use the templates.

PR #8834, opened 2022-11-22
chore: Use jq to extract field

lqiu96:
Use JQ to extract field from JSON. This will help use remove the the python3 dependency. I'll need to install JQ on all the Java and GraalVM images.

Update the docs at cl/490322376

This is to fix the issue in GraalVM:
```
./helpers/common.sh: line 43: python3: command not found
cleanup
```


lqiu96:
The latest GraalVM and Java images have JQ installed now.

ddixit14 (approved):

PR #8848, opened 2022-11-29
chore: Set parent pom to be google-cloud-java

lqiu96:


ddixit14:
@suztomo This PR will fix the annotations. `set_parent_pom.sh` will now make sure that `google-cloud-jar-parent` and `google-cloud-pom-parent` will have `google-cloud-java` as key in the annotation wherever it is declared as a parent.
For `gapic-libraries-bom`, we are using the file itself for regeneration (in `generate_gapic_bom.sh`), and it already has `google-cloud-java` in the annotation. We are not editing this portion of the file while regeneration so it should be good. 
Thank you @lqiu96.

meltsufin (approved):

suztomo:
Thank you.

PR #8854, opened 2022-11-30
chore: dashboard scripts

ddixit14:
This PR includes 4 files:

1. `dashboard/cloud_java_client_library_release_dates.csv` - Output csv that needs to be uploaded to storage (this is just a sample)
2. `dashboard/create_libraries_release_data_csv.sh` - main script that will iterate over libraries.txt, call fetch_library_release_date.sh for every artifact and add the contents to a global file, which later gets transformed to the output csv.
3. `dashboard/fetch_library_release_date.sh` - this script runs for each artifact, fetches data from maven central and adds it to the csv in desired format.
4. `dashboard/libraries.txt` - source of truth file which will have all the artifacts we need to monitor.


dashboard/fetch_library_release_date.sh
```diff
@@ -0,0 +1,32 @@
...
+  ${1}
+
+grep -E '<a href=".*">' mavenFile > mavenContents.txt
+
+awk  '/a/ {print  $2 "\t" $4}'  mavenContents.txt > finalContents.txt
+sed -i 1d  finalContents.txt
+sed -i '/maven-metadata/d' finalContents.txt
+sed -i 's/href="//g' finalContents.txt
+sed -i 's/"//g' finalContents.txt
+sed -i 's|/||g' finalContents.txt
```
suztomo, line 22: cat mavenFile |perl -nle 'print "$1,$2" if m/(\d+\.\d+\.\d+).+(\d\d\d\d\-\d\d\-\d\d)/'

PR #8926, opened 2022-12-21
Remove java notebooks

cjac:
This code has been moved to a different repository.


cjac:
I'll probably also need to update the tests to check those repos out before running the tests, eh?

PR #8997, opened 2023-01-18
test: try running graalvm jobs in parallel

mpeddada1:


mpeddada1:
Compute tests are taking ~45 mins despite `-T 1C`.

PR #9017, opened 2023-01-20
build: configure changelog.json generation

bcoe:
Configures `changelog.json` generation for `dataflow` and `secretmanager` (two of the libraries that we included in the release notes pilot on cloud.google.com).

Depends on landing: https://github.com/googleapis/release-please/pull/1810


bcoe:
we came up with a new approach using `.repo-metadata.json`.

PR #9092, opened 2023-02-07
ci: delete library-specific GraalVM presubmit configs

burkedavison:


meltsufin (approved):

.github/sync-repo-settings.yaml
```diff
@@ -36,10 +36,6 @@ branchProtectionRules:
       - lint
       - OwlBot Post Processor
       - 'Kokoro - Test: Integration'
-      - 'Kokoro - Test: Java 17 GraalVM Native Image - AI Platform'
-      - 'Kokoro - Test: Java 17 GraalVM Native Image - Compute'
-      - 'Kokoro - Test: Java GraalVM Native Image - AI Platform'
-      - 'Kokoro - Test: Java GraalVM Native Image - Compute'
```
burkedavison, line 42: I may have made a mistake by removing the old builds prior to removing the 'required' entries here.
Unless there's an 'override', I may need to rollback and perform these updates in a different order.

PR #9109, opened 2023-02-09
chore: Making dependabot ignore everything pip related

ddixit14:


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+# rules for the main branch
+- package-ecosystem: maven
```
suztomo, line 4: .kokoro/requirements.in is not maven

ddixit14: changed to pip.

suztomo (approved):
Let's do experiment.

PR #9146, opened 2023-02-21
chore: Clean up unused file ConnectionProto.java

blakeli0:
ConnectionProto.java is not used and should be removed, otherwise changes from this [cl](cl/510189478) will not compile.

This is technically a breaking change but this file is currently not used.


meltsufin (approved):

blakeli0:
@Neenu1995 Do you see any issues of removing this file? This file is not used in the monorepo, but not sure if there is any bigquery repos are dependent on it.

Neenu1995:
Neither of the bigquery handwritten clients use the ConnectionProto class. 
LGTM.

Neenu1995 (approved):

PR #9173, opened 2023-03-01
feat: [datastore] new module for datastore [DO NOT MERGE]

meltsufin:
Generated using the following command:
```
python3.9 generation/new_client/new-client.py generate  \
--api_shortname=datastore  \
--proto-path=google/datastore  \
--name-pretty="Datastore API" \
--product-docs="https://cloud.google.com/datastore" \
--api-description="Datastore is a highly scalable NoSQL database for your web and mobile applications."
```

### DIFF
--- a/gapic-libraries-bom/pom.xml
+++ b/gapic-libraries-bom/pom.xml
@@ -1031,13 +1031,6 @@
       </dependency>
       <dependency>
         <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-gapic-bom</artifactId>
-        <version>0.0.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-gapic-bom:current} -->
-        <type>pom</type>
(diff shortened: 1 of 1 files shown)


gapic-libraries-bom/pom.xml
```diff
@@ -1022,6 +1029,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-gapic-bom</artifactId>
```
meltsufin, line 1034: @suztomo Do you know why this gets inserted here when running the new client generation script?

suztomo: No, I don’t. I don’t recognize this BOM. Finding corresponding commit (and commit message) will clarify which step adds this.

PR #9186, opened 2023-03-03
chore: fix README.md versions in alloydb

diegomarquezp:


suztomo:
Add annotation for the version https://github.com/googleapis/google-cloud-java/blob/main/java-accessapproval/README.md?plain=1#L16 so that the value is automatically updated.

ddixit14:
@suztomo checking why weren't these annotations added when the module was generated. readme_update.sh should have added the annotations in here as well.

suztomo (approved):
@ddixit14 Maybe the new-client.py didn't have the commit you made for the tag?

PR #9272, opened 2023-03-23
fix: update OwlBotYaml to copy over autogenerated snippets for Compute

alicejli:
Snippets are getting autogenerated for Compute: https://github.com/googleapis/googleapis-gen/tree/master/google/cloud/compute/v1/google-cloud-compute-v1-java/samples/snippets/generated/com/google/cloud/compute/v1

This update to the .OwlBot.yaml file should copy them over from googleapis-gen to google-cloud-java.

### DIFF
--- a/java-compute/.OwlBot.yaml
+++ b/java-compute/.OwlBot.yaml
@@ -27,7 +27,7 @@ deep-copy-regex:
   dest: "/owl-bot-staging/java-compute/$1/proto-google-cloud-compute-$1/src"
 - source: "/google/cloud/compute/(v\\d)/.*-java/gapic-google-.*/src"
   dest: "/owl-bot-staging/java-compute/$1/google-cloud-compute/src"
-- source: "/google/cloud/compute/(v.*)/.*-java/samples/snippets/generated"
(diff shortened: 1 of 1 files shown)


java-compute/.OwlBot.yaml
```diff
@@ -27,7 +27,7 @@ deep-copy-regex:
   dest: "/owl-bot-staging/java-compute/$1/proto-google-cloud-compute-$1/src"
 - source: "/google/cloud/compute/(v\\d)/.*-java/gapic-google-.*/src"
   dest: "/owl-bot-staging/java-compute/$1/google-cloud-compute/src"
-- source: "/google/cloud/compute/(v.*)/.*-java/samples/snippets/generated"
```
suztomo, line 30: Would you explain why the previous regular expression didn't match the expected files?

alicejli: Sorry, just double checked and you're right - the previous expression did work. It looks like the update PR just hasn't been merged in yet (https://github.com/googleapis/google-cloud-java/pull/9262/files). I will close this.

PR #9315, opened 2023-04-05
docs: troubleshooting guide for logging

suztomo:
Document to explain how to change logging level for our libraries and underlying libraries.

Credit: I used @BenWhitehead 's snippet in bits benwhitehead/6166544285433856.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -25,12 +25,16 @@ java.util.logging.ConsoleHandler.level=ALL
 java.util.logging.ConsoleHandler.formatter=java.util.logging.SimpleFormatter
 .level=INFO
 
-# Specify logging level for certain packages
+# --- Specify logging level for certain packages ---
+# com.google.api is for HTTP 1.1 layer
 com.google.api.level=ALL
-com.google.auth.level=FINE
+# io.grpc is for gRPC + Netty layer
 io.grpc.level=FINE
+# com.google.auth is for authentication
+com.google.auth.level=FINE
 
-# Example when we want to specify storge library's level
+# Example when we want to specify storge library's level. This works when
+# the target Cloud library uses the logging API.
 com.google.cloud.storage.level=INFO
 ```
 


rajatbhatta:
very minor nit: guilde -> guide in PR title

TROUBLESHOOTING.md
```diff
@@ -1,5 +1,68 @@
...
+handlers=java.util.logging.ConsoleHandler
+java.util.logging.SimpleFormatter.format=%1$tF %1$tT,%1$tL %4$-8s %3$-50s - %5$s %6$s%n
+
+# --- ConsoleHandler ---
+java.util.logging.ConsoleHandler.level=ALL
+java.util.logging.ConsoleHandler.formatter=java.util.logging.SimpleFormatter
+.level=INFO
+
+# Specify logging level for certain packages
+com.google.api.level=ALL
```
blakeli0, line 29: Can we add notes to indicate that this is only needed for REST APIs and `io.groc.level` below is only needed for gRPC APIs?

TROUBLESHOOTING.md
```diff
@@ -1,5 +1,68 @@
...
+com.google.auth.level=FINE
+io.grpc.level=FINE
+
+# Example when we want to specify storge library's level
+com.google.cloud.storage.level=INFO
```
blakeli0, line 34: I'm afraid this is only useful for handwritten libraries, for pure generated ones, we have very little logs, which is something we are planning to improve as part of the observability project. Can we add something to indicate it as well or maybe remove this section?

blakeli0:
Thanks Tomo, this is very helpful! FYI, we are planning to improve the whole logging experience as part of the Observability project, so a lot of things could change, but this guide is pretty accurate as of now.

lqiu96:
Looks good! IIRC, I have used this page previously for logging some of the HTTP requests for REGAPIC: http://googleapis.github.io/google-http-java-client/http-transport.html

I think your guide should cover the cases. Just FYI if you wanted to add anything else.

TROUBLESHOOTING.md
suztomo: Added comments.

Note that even when you use gRPC library, some authentication happens in HTTP 1.1.

TROUBLESHOOTING.md
suztomo: Added that.

suztomo (commented):
@blakeli0 PTAL.

> Observability project, 

Looking forward.

TROUBLESHOOTING.md
blakeli0: > Note that even when you use gRPC library, some authentication happens in HTTP 1.1.

Thanks, that's good to know. I guess it would be mostly covered by the `com.google.auth.level=FINE` below?

blakeli0:
> Looks good! IIRC, I have used this page previously for logging some of the HTTP requests for REGAPIC: http://googleapis.github.io/google-http-java-client/http-transport.html
> 
> I think your guide should cover the cases. Just FYI if you wanted to add anything else.

Thanks for then note Lawrence! I think it should be already covered by `com.google.api.level=ALL` as `com.google.api.client.http` is a sub-package of `com.google.api`.

blakeli0 (approved):

TROUBLESHOOTING.md
suztomo: That does not show HTTP headers.

PR #9560, opened 2023-06-20
chore: Update changelog_generation.yaml

ddixit14:


suztomo (commented):
@ddixit14  You had some command to print the `github.event` object. Can you add that somewhere probably in a different job. (We'll remove that job once we know it starts working)

suztomo (approved):

PR #9776, opened 2023-08-22
test: understanding kokoro behavior (DO NOT MERGE)

mpeddada1:


mpeddada1:
Notes: When the mvn command from the build script is deleted, the Kokoro job will show up as "Passing" even though it doesn't run any tests.

suztomo:
It's the return code of shell script and the switch statement in bash.

PR #9800, opened 2023-09-05
feat: added iam troubleshooter v3 and v3beta api clients

alicejli:
New policytroubleshooter client protos are in a different folder than v1. Manually adjusting owlbot file to copy over correctly into the existing module

Tested the Owlbot.yaml file locally and it generates the directories for v3/v3beta correctly:

image

Command to test locally:

```
docker run --rm --user $(id -u):$(id -g)   -v $(pwd):/repo -v /tmp/googleapis-gen:/googleapis-gen -w /repo   --env HOME=/tmp   gcr.io/cloud-devrel-public-resources/owlbot-cli:latest copy-code   --source-repo=/googleapis-gen   --config-file=java-policy-troubleshooter/.OwlBot.yaml
```

### DIFF
--- a/java-policy-troubleshooter/.OwlBot.yaml
+++ b/java-policy-troubleshooter/.OwlBot.yaml
@@ -35,13 +35,13 @@ deep-copy-regex:
   dest: "/owl-bot-staging/java-policy-troubleshooter/$1/google-cloud-policy-troubleshooter/src"
 - source: "/google/cloud/policytroubleshooter/(v.*)/.*-java/samples/snippets/generated"
   dest: "/owl-bot-staging/java-policy-troubleshooter/$1/samples/snippets/generated"
-- source: "/google/cloud/policytroubleshooter/iam/(v.*)/.*-java/proto-google-.*/src"
+- source: "/google/cloud/policytroubleshooter/iam/v3/.*-java/proto-google-.*/src"
   dest: "/owl-bot-staging/java-policy-troubleshooter/$1/proto-google-cloud-policy-troubleshooter-$1/src"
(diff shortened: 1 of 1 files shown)


blakeli0:
> Tested the Owlbot.yaml file locally and it generates the directories for v3/v3beta correctly:

Did we ever get a library generation requests for v3beta version? If not, maybe just copy the v3 version? WDTY @suztomo?

alicejli:
> > Tested the Owlbot.yaml file locally and it generates the directories for v3/v3beta correctly:
> 
> Did we ever get a library generation requests for v3beta version? If not, maybe just copy the v3 version? WDYT @suztomo?

We did not get an explicit library generation request for v3beta, but v3beta does exist in the public protos, and in general (I think) our policy has been to generate libraries for whatever exists in the public protos. Can update to whichever is preferred.

suztomo:
In the ticket that requests library generation, the path to the directory containing proto is explicitly specified up to a version, isn't it?

alicejli:
> In the ticket that requests library generation, the path to the directory containing proto is explicitly specified up to a version, isn't it?

Yes, but for languages like Java we usually automatically update existing modules with new versions as they are published. This library is a bit unique where the new version is published in a different directory than the existing protos, so owlbot didn't catch the new version.

blakeli0:
> We did not get an explicit library generation request for v3beta, but v3beta does exist in the public protos, and in general (I think) our policy has been to generate libraries for whatever exists in the public protos.

I don't think this is true. There is a language section in each library generation request, which sometimes only include a subset of languages, I did see Java being excluded sometimes.

suztomo (approved):

PR #9808, opened 2023-09-08
chore: include proto-* and grpc-* in Owlbot settings for java-notebooks

mpeddada1:


alicejli (approved):
The additional lines look correct. I think you can probably remove lines 26-39 as the newly added lines would encompass them.

Issue #9837, opened 2023-09-15
When generating a new library, do not add non-Cloud artifacts to the gapic-libraries-bom/pom.xml


suztomo:
Follow-up of https://github.com/googleapis/google-cloud-java/pull/9811#discussion_r1327382602.

When a new library for a service is created, it calls `generate_gapic_bom.sh`:

https://github.com/googleapis/google-cloud-java/blob/3d9341a8ba86927e50de1de50e85b324483f90a8/generation/new_client/new-client.py#L336

```
    subprocess.check_call(
        [
            "bash", "generation/generate_gapic_bom.sh",
        ],
        cwd=monorepo_root,
    )
```

This script adds the newly generated library as part of `gapic-libraries-bom/pom.xml`.

PR #9849, opened 2023-09-19
ci: generate_gapic_bom.sh to include Cloud and existing entries

suztomo:
Fixes https://github.com/googleapis/google-cloud-java/issues/9837

The gapic bom mainly includes cloud libraries and ones that have been included already.
Let's avoid adding com.google.maps and com.google.shopping for now. We may decide to
add them later. It's more difficult to remove them later without impacting users.

PR #9853, opened 2023-09-19
ci: generate_gapic_bom.sh to include only Cloud and existing entries

suztomo:
Fixes https://github.com/googleapis/google-cloud-java/issues/9837

The gapic bom mainly includes cloud libraries and ones that have been included already.
Let's avoid adding com.google.maps and com.google.shopping for now. We may decide to
add them later. It's more difficult to remove them later without impacting users.

"New Client / generate" check was not succeeding when I created #9849 from a forked repository.

### DIFF
--- a/generation/generate_gapic_bom.sh
+++ b/generation/generate_gapic_bom.sh
@@ -25,12 +25,12 @@ for bom_directory in $(find . -maxdepth 3 -name 'google-*-bom' | sort --dictiona
     # Let's avoid adding com.google.maps and com.google.shopping for now. We may decide to
     # add them later. It's more difficult to remove them later without impacting users.
     bom_lines+="      <dependency>\n\
-        ${groupId_line}\n\
-        ${artifactId_line}\n\
-        ${version_line}\n\
-          <type>pom</type>\n\
-          <scope>import</scope>\n\
-        </dependency>\n"
+      ${groupId_line}\n\
+      ${artifactId_line}\n\
+      ${version_line}\n\
+        <type>pom</type>\n\
+        <scope>import</scope>\n\
+      </dependency>\n"
   fi
 done
 


suztomo:
"New Client / generate" succeeded.

generation/generate_gapic_bom.sh
```diff
@@ -17,8 +17,13 @@ for bom_directory in $(find . -maxdepth 3 -name 'google-*-bom' | sort --dictiona
...
-    # The gapic bom includes cloud libraries
+  if [[ "$groupId_line" != *"com.google.cloud"*
+      && "$groupId_line" != *"com.google.analytic"*
+      && "$groupId_line" != *"com.google.area120"*
+      && "$groupId_line" != *"io.grafeas"*  ]]; then
+    # The gapic bom mainly includes cloud libraries and ones that have been included already.
+    # Let's avoid adding com.google.maps and com.google.shopping for now. We may decide to
+    # add them later. It's more difficult to remove them later without impacting users.
     continue
   fi
```
mpeddada1, line 28: nit: Negative booleans can sometimes increase cognitive load. How about something like this?
```suggestion
  if [[ "$groupId_line" == *"com.google.cloud"*
      || "$groupId_line" == *"com.google.analytic"*
      ||"$groupId_line" == *"com.google.area120"*
      || "$groupId_line" == *"io.grafeas"*  ]]; then
    # The gapic bom mainly includes cloud libraries and ones that have been included already.
    # Let's avoid adding com.google.maps and com.google.shopping for now. We may decide to
    # add them later. It's more difficult to remove them later without impacting users.
   bom_lines+="      <dependency>\n\
      ${groupId_line}\n\
      ${artifactId_line}\n\
      ${version_line}\n\
        <type>pom</type>\n\
        <scope>import</scope>\n\
      </dependency>\n"
  else
     continue
  fi
```

suztomo: That was also itchy part for me too. Updated.

suztomo:
@mpeddada1 PTAL

mpeddada1 (approved):
Looks great.

PR #9931, opened 2023-10-02
docs: update new client generation instructions

### LINKED ISSUE #9902: New client library generation needs to be updated with regard to api-shortname
Currently, `new-client.py` uses `api-shortname` for directory creation. However, `api-shortname` is not always a unique value. We should determine a new methodology for creating the directory using a unique name.

For example: https://github.com/googleapis/google-cloud-java/pull/9896

alicejli:
Fixes #9902

### DIFF
--- a/generation/new_client/README.md
+++ b/generation/new_client/README.md
@@ -138,9 +138,8 @@ The example in this README uses AlloyDB's [Cloud Drop](https://github.com/google
 
 For convenience of the subsequent commands, define a variable for API short name.
 This value will be used by default to generate the following:
-* `distribution_name`
-* `destination-name`
-* `proto_path`
+* `--distribution-name`
+* `--destination-name`
 
 The corresponding value in the Cloud Drop page is `api_short_name`.
 
@@ -328,15 +327,12 @@ Let's say you get a new library request where the Cloud Drop value for `api_shor
 
 You discover that `maps-routing` is already in use by an existing client library!
 
-You need to determine a unique `destination_name` for the new library's subdirectory as well as a unique `distribution_name` for the artifact to be published (confirm these values with the service team on the buganizer ticket before proceeding).
+You need to determine a unique `destination_name` for the new library's subdirectory as well as a unique `distribution_name` for the artifact to be published. There is no hard and fast rule for determining a unique name, so some discussion will be necessary. Confirm the `distribution_name` with the service team on the buganizer ticket before proceeding, as this name is visible to customers as the Maven artifact name.
 
-You will **still use** the non-unique `api_short_name` for the `api_short_name` flag.
+You will **still use** the non-unique `api_short_name` for the `api_short_name` flag. This is important because the `api_short_name` is used to derive links within cloud.google.com to enable the API.
 
-There is no hard and fast rule for determining a unique name, so some discussion will be necessary.
+Let's say that after some discussion, `maps-routing-gps` is selected as a suitable unique subdirectory name and `com.google.maps:google-maps-routing-gps` is selected as a suitable unique artifact name. You would then use the following command:
 
-Let's say that after some discussion, `maps-routing-gps` is selected as a suitable unique subdirectory name and artifact name.
-
-You would then use the following command:
 ```
 ~/google-cloud-java$ python3.9 generation/new_client/new-client.py generate \
   --api_shortname=maps-routing \
--- a/generation/new_client/new-client.py
+++ b/generation/new_client/new-client.py
@@ -187,9 +187,6 @@ def generate(
         f"https://cloud.google.com/{language}/docs/reference/{distribution_name_short}/latest/overview"
     )
 
-    if proto_path is None:
-        proto_path = f"/google/cloud/{api_shortname}"
-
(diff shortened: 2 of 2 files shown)


generation/new_client/README.md
```diff
@@ -299,6 +304,24 @@ generated Google Maps Routes API Java client library.
   --distribution-name="com.google.maps:google-maps-routing"
 ```
 
+If `maps-routing` was already in use by an existing client library as an `api_shortname`, you
+would determine a unique `destination_name` for the new library's subdirectory as well as a unique `distribution_name` for the artifact to be published (confirm with the service team).
+Let's say `maps-routing-gps` is the unique subdirectory and artifact name that has been decided
+You would then use the following command:
```
suztomo, line 310: Which part is special?

generation/new_client/README.md
```diff
@@ -136,9 +136,14 @@ The example in this README uses AlloyDB's [Cloud Drop](https://github.com/google
 
 ### API short name
 
-For convenience of the subsequent commands, define a variable for API short name.
+For convenience of the subsequent commands, define a variable for API short name. This will be used to generate the `distribution_name` as well as the sub-directory within this repo.
```
suztomo, line 139: Not just these, but many params use it as their default values.

generation/new_client/README.md
```diff
@@ -137,8 +137,15 @@ The example in this README uses AlloyDB's [Cloud Drop](https://github.com/google
 ### API short name
 
 For convenience of the subsequent commands, define a variable for API short name.
+This value will be used to generate the `distribution_name` as well as the sub-directory for the client library within this repo. It is also the default value for many additional parameters, so take care to select a unique and reasonable value.
```
suztomo, line 140: Can you use bullet points?

generation/new_client/README.md
```diff
@@ -137,8 +137,15 @@ The example in this README uses AlloyDB's [Cloud Drop](https://github.com/google
...
+
 The corresponding value in the Cloud Drop page is `api_short_name`.
 
+> [!IMPORTANT]
+> `api_short_name` is not always unique across client libraries.
+> In the instance that the `api_short_name` is already in use by an existing client library, determine a unique name and append it to the `--destination-name`.
+> See example under [Advanced Options](#advanced-options).
+
 Example: `alloydb`
```
suztomo, line 149: This is not an example of "the `api_short_name` is already in use by an existing client library"

generation/new_client/README.md
alicejli: Good idea - updated

generation/new_client/README.md
alicejli: Sorry that example should have been above the `Note`. I added a comprehensive example about what to do with duplicate api_short_names below.

generation/new_client/README.md
```diff
@@ -137,10 +137,20 @@ The example in this README uses AlloyDB's [Cloud Drop](https://github.com/google
 ### API short name
 
 For convenience of the subsequent commands, define a variable for API short name.
+This value will be used by default to generate the following:
+* `distribution_name`
```
suztomo, line 141: Line 268 says it's "-"

```suggestion
* `distribution-name`
```

Would you check which is correct?

generation/new_client/README.md
```diff
@@ -137,10 +137,20 @@ The example in this README uses AlloyDB's [Cloud Drop](https://github.com/google
...
+* `destination-name`
+* `proto_path`
```
suztomo, line 143:
```suggestion
* `proto-path`
```

According to line 279. Can you check which is correct?

suztomo, line 143: Reading the section starting from line 154, I couldn't get how the default value would be constructed for proto-path using api_shortname. Would you clarify that?

generation/new_client/README.md
```diff
@@ -312,6 +322,34 @@ generated Google Maps Routes API Java client library.
...
 ```
+### Example with duplicate api_short_name
+
+Let's say you get a new library request where the Cloud Drop value for `api_short_name` is `maps-routing`.
+
+You discover that `maps-routing` is already in use by an existing client library!
+
+You need to determine a unique `destination_name` for the new library's subdirectory as well as a unique `distribution_name` for the artifact to be published (confirm these values with the service team on the buganizer ticket before proceeding).
+
+You will **still use** the non-unique `api_short_name` for the `api_short_name` flag.
```
suztomo, line 333: In this example, why is it important to use maps-routing-gps in api_shortname?
(An alternative is maps-routing-gps as api_shortname.)

generation/new_client/README.md
```diff
@@ -312,6 +322,34 @@ generated Google Maps Routes API Java client library.
...
+
+There is no hard and fast rule for determining a unique name, so some discussion will be necessary.
+
+Let's say that after some discussion, `maps-routing-gps` is selected as a suitable unique subdirectory name and artifact name.
```
suztomo, line 337: I see the artifact name is different from "maps-routing-gps".

suztomo (approved):

PR #9977, opened 2023-10-18
chore: mark native image checks as required

mpeddada1:


lqiu96 (approved):

burkedavison (approved):
If handwritten integration test flakes become an overly-burdensome blocker to PRs, we should increase the priority on replacing them.

PR #10050, opened 2023-11-09
fix: use path wildcards in java-iam owlbot yaml

diegomarquezp:
Alternative to https://github.com/googleapis/sdk-platform-java/pull/2233


blakeli0:
Looks like we should get identical result with the config change, can you please verify it locally?

blakeli0 (approved):
LGTM. Pending confirmation of the comment.

diegomarquezp:
> Looks like we should get identical result with the config change, can you please verify it locally?

I confirmed locally by checking out this branch and running the IT

PR #10159, opened 2023-12-15
fix: [vertexai] rely on gapic to do ADC while no scope is provided.

ZhenyiQ:


java-vertexai/google-cloud-vertexai/src/main/java/com/google/cloud/vertexai/VertexAI.java
```diff
@@ -85,21 +88,22 @@ public VertexAI(
...
-    // Disable the warning message logged in getApplicationDefault
-    Logger logger = Logger.getLogger("com.google.auth.oauth2.DefaultCredentialsProvider");
-    Level previousLevel = logger.getLevel();
-    logger.setLevel(Level.SEVERE);
     List<String> defaultScopes =
         PredictionServiceStubSettings.defaultCredentialsProviderBuilder().getScopesToApply();
-    GoogleCredentials credentials =
+    List<String> allScopes = new ArrayList<String>(Arrays.asList(scopes));
+    // defaultScopes contains the necessary scope to make service account credentials work.
+    allScopes.addAll(defaultScopes);
```
meltsufin, line 95: @blakeli0 Should default scopes be merged with user-defined scopes, if those are provided?

ZhenyiQ: I was considering sijunliu's suggestions: "If the user is required to pass scopes in order to authenticate (i.e. no user provided scopes = no authentication used), then I think we should either combine it with cloud-platform or ask users to provide cloud-platform in the public documentation. Our experiments confirmed the api call fails without the cloud-platform scope."

And then thought: if we always ask the users to provide `cloud-platform` in documentation, that's a worse user experience compared to just do that ourselves since the documentation may not be that discoverable.

blakeli0: We don't do that right now for any other GAPIC clients, we have a setter in [GoogleCredentialsProvider](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/core/GoogleCredentialsProvider.java#L122) for customers to completely override the scopes, and we don't merge the default scopes with user-defined ones.
However, per @arithmetic1728, it might be a good idea to do so. @TimurSadykov Is there any additional implications if we start doing so?

TimurSadykov: An implication I can think of is that customer might want to explicitly set a narrower scope. It might not be applicable to VertexAi atm, but technically could be the case later. 

In terms of user experience, it is not uncommon to require the cloud-platform scope. For example, all cloud run rpcs require that scope: https://cloud.google.com/run/docs/reference/rpc/google.cloud.run.v2

arithmetic1728: Based on the discussion with Shin Fan, my understanding is that: if their backend hasn't configured the allowed narrower scopes, then cloud-platform is the one for user to use; if the backend is configured with a list of narrower scopes, then users can use these narrower scopes instead of cloud-platform. So this basically leaves us 3 options: (1) always add the cloud-platform scope (2) document cloud-platform (3) configure backend to allow a narrower scope like cloud-aiplatform. 

(1) would be a good short term solution for the best user experience. Once (3) is done, we can change the code to not add cloud-platform to the user provided scopes.

ZhenyiQ: Thanks a bunch folks! I've raised this issue to the my team (we also develop the Python and JS SDKs and we all have the same concern about this too-broad scope) and we plan to escalate to a higher level - it's unclear how soon this could be resolved.

By the way, is this scope applied to only service key authentication flow? (I'm wondering what's the effect of having this scope specified when the user uses ADC but not the service account key file).

ZhenyiQ: Talked to the Vertex uber TL. It looks like we (vertex) do prefer not to use any finer-granular scope and just use cloud-platform.

I also found this: go/scope-cfg#recipe-google-cloud-platform-api-scope - to me, it signals that "we should *not* even have finer-granular scopes" the corollary of which, I believe, is "the user should stop thinking that they should manage access via scopes. They should, instead, be fully aware and strict about the IAM policy and not worry about the scopes at all".

ZhenyiQ: synced with @arithmetic1728 . I'll drop this PR and get back to "still support user specified scopes and do not merge with cloud-platform" and improve documentation on required scope.

PR #10160, opened 2023-12-15
fix: [vertexai] Use GoogleCredentialsProvider to get the default credentials.

blakeli0:
Note that this is technically a breaking change as the type of `credential` object is change from `GoogleCredentials` to `Credentials`. But since it was only released 2 days ago, and it should be `Credentials` to start with, it might be OK.


meltsufin:
@ZhenyiQ

ZhenyiQ:
I think the problem was resolved by other PRs and we can close this PR.

PR #10169, opened 2023-12-19
fix: [vertexai] Use default scopes from Prediction Service to create application default credentials if scopes are empty.

blakeli0:


ZhenyiQ (approved):
Thanks Blake!

meltsufin (approved):

Issue #10297, opened 2024-01-29
CHANGELOG.md is not generated for java-vertexai module


meltsufin:
cc: @ZhenyiQ

suztomo:
It's now fixed. The upcoming release by Joe will have the changelog for java-vertexai.

![image](https://github.com/googleapis/google-cloud-java/assets/28604/0304e9bf-96e5-427f-9235-086dbcba6e61)

PR #10303, opened 2024-01-30
ci: split_release_note.py to use repo-metadata.json too

suztomo:
Fixes https://github.com/googleapis/google-cloud-java/issues/10297

When split_release_note.py was searching for changelog files, it was looking for only OwlBot.yaml files. The java-vertexai module has a special setup where it does not need OwlBot.yaml files. With this PR, the script also looks for repo-metadata file (which java-vertexai module has).

TODO: After this is merged, confirm the Release Please pull request https://github.com/googleapis/google-cloud-java/pull/10284 gets the fixed commit for java-vertexai's CHANGELOG file.

### DIFF
--- a/.github/release-note-generation/golden/java-vertexai/CHANGELOG.md
+++ b/.github/release-note-generation/golden/java-vertexai/CHANGELOG.md
@@ -1,0 +1,28 @@
+# Changelog
+
+## 0.50.0 (2023-06-08)
+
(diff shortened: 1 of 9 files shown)


meltsufin (approved):
Should we expect the actual CHANGELOG.md for java-vertexai to appear as part of some other process?

PR #10348, opened 2024-02-07
chore: remove obsolete files

alicejli:
The kokoro release configs have moved to g3. The `requirements.txt` and `requirements.in` files should no longer be needed in this repo.

TODO: Understand what the `requirements-kokoro` test in CI is for; if that is still needed, then these files are needed.


alicejli:
Confirmed with @mpeddada1 that the ci check is obsolete, so removing that as well.

burkedavison (approved):

PR #10357, opened 2024-02-09
chore: add a generation configuration

JoeWang1127:
In this PR:
- Add a generation configuration file

To generate libraries in the config file, 
1. copy `library_generation` folder from sdk-platform-java main branch to google-cloud-java root directory.
2. in the root of google-cloud-java, run
```
python library_generation/generate_repo.py generate --generation-config-yaml=/usr/local/google/home/joewa/workspace/google-cloud-java/generation_config.yaml --repository-path=.
```

Total generation time is 1 hour 4 mins on a gLinux Cloudtop.

Note that the path of the config file has to be absolute.

Four libraries (`java-alloydb-connects`, `java-bigquery-data-exchange`, `java-securitycenter`, `java-shopping-merchant-inventories`) are commented out because they have the same `api-shortname` with other libraries and it will cause conflicts when generating libraries with the same `api-shortname` (https://github.com/googleapis/sdk-platform-java/issues/2494). I will remove the comment once the problem is fixed.

This is [step 3](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?pli=1&tab=t.0#bookmark=id.ntlh4z94hvrp) of milestone 2 in hermetic build project.

### DIFF
--- a/generation_config.yaml
+++ b/generation_config.yaml
@@ -15,6 +15,10 @@ template_excludes:
 - license-checks.xml
 - renovate.json
 - .gitignore
+
+# the libraries are ordered with respect to library name, which is
+# java-{library.library_name} or java-{library.api-shortname} when
+# library.library_name is not defined.
 libraries:
 - api_shortname: accessapproval
   name_pretty: Access Approval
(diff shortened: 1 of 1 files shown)


suztomo:
You should have run some command to generate this file. Would you write what command you ran in the pull request description?

generation_config.yaml
```diff
@@ -0,0 +1,1974 @@
...
+  product_documentation: https://cloud.google.com/access-approval/docs/
+  api_description: enables controlling access to your organization's data by Google
+    personnel.
+  release_level: stable
+  GAPICs:
+  - proto_path: google/cloud/accessapproval/v1
+
+- api_shortname: accesscontextmanager
+  name_pretty: Identity Access Context Manager
+  product_documentation: n/a
```
suztomo, line 30: YAML has null.

generation_config.yaml
```diff
@@ -0,0 +1,1974 @@
...
+- samples/*
+- CODE_OF_CONDUCT.md
+- CONTRIBUTING.md
+- LICENSE
+- SECURITY.md
+- java.header
+- license-checks.xml
+- renovate.json
+- .gitignore
+libraries:
```
suztomo, line 18: Would you explicitly define the order of the entries in `libraries` field? It will help where to add a new entry.

generation_config.yaml
JoeWang1127: This is because `product_documentation` is `n/a`.

We should verify the correctness of a field for new libraries.

JoeWang1127:
> You should have run some command to generate this file. Would you write what command you ran in the pull request description?

I created a script to generate the config file and I removed the script before sending the PR to review (https://github.com/googleapis/google-cloud-java/pull/10357/commits/3b763c33826d5b98e8fdb84ff4a9b9230c6de11d)

suztomo:
Thank you

PR #10389, opened 2024-02-15
chore: add dependency byte-buddy to enable mockStatic in Java 21

jaycee-li:
`byte-buddy` is required for mocking static methods in java-vertexai. `mockito-core 4.11.0` depends on `byte-buddy 1.12.19`, which is not compatible with Java 21. So we manually add `byte-buddy 1.14.3` (lowest version that supports Java 21) to google-cloud-jar-parent.

### DIFF
--- a/google-cloud-jar-parent/pom.xml
+++ b/google-cloud-jar-parent/pom.xml
@@ -106,12 +106,14 @@
         <version>4.11.0</version>
         <scope>test</scope>
       </dependency>
+      <!-- for java-vertexai unit tests, can be removed once we update to mockito-core >= 5.0.0 -->
       <dependency>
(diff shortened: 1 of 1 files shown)


suztomo (approved):
Scope is test. Library users won't see this. Good.

google-cloud-jar-parent/pom.xml
```diff
@@ -112,6 +112,12 @@
         <version>4.11.0</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>net.bytebuddy</groupId>
```
JoeWang1127, line 116: Could you add a code comment why this test dependency is needed so that we can remove it  later if we upgrade mockito.

PR #10503, opened 2024-03-08
feat: create pr after library generation

JoeWang1127:
In this PR:
- Update workflow to create a PR of hermetic code generation.
  - search baseline commit (B) in `generation_config.yaml`.
  - search latest commit (L) in googleapis repository.
  - update googleapis commit to L in `generation_config.yaml`.
  - search for open pull request associated with branch `generate-libraries`.
    - create a new branch if no pull request is not found; otherwise checkout the pull request.
  - generate GAPIC libraries from `generation_config.yaml`.
  - generate pull request description from B to L.
  - create a pull request or push a commit.
- Update `generation_config.yaml` to include new libraries and new versions.

Context: this is the [launch plan](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.0#bookmark=id.6l7rqzymdpwz) of hermetic build project.

### DIFF
--- a/.github/workflows/generate-from-configuration.yaml
+++ b/.github/workflows/generate-from-configuration.yaml
@@ -2,9 +2,6 @@ name: Generate GAPIC libraries from configuration
 on:
   schedule:
   - cron: '0 2 * * *' # nightly at 2 am UTC
-  push:
-    branches:
-    - feat/hermetic-pr-creation
   workflow_dispatch:
 
 


.github/workflows/verify-generation-config.yaml
```diff
@@ -2,13 +2,42 @@ name: Verify Generation Configuration
 on:
   schedule:
   - cron: '0 2 * * *' # nightly at 2 am UTC
+  push:
+    branches:
+    - feat/hermetic-pr-creation
```
JoeWang1127, line 7: I'll remove this trigger after pr is approved.

JoeWang1127:
The pull request created by this workflow is https://github.com/googleapis/google-cloud-java/pull/10516.

.github/workflows/verify-generation-config.yaml
```diff
@@ -18,6 +48,18 @@ jobs:
           docker volume rm repo-google-cloud-java
         fi
         docker volume create --name "repo-google-cloud-java" --opt "type=none" --opt "device=$(pwd)" --opt "o=bind"
+    - name: update googleapis commit to latest
+      id: update-commit
+      shell: bash
+      run: |
+        mkdir tmp-googleapis
+        git clone https://github.com/googleapis/googleapis.git tmp-googleapis
```
diegomarquezp, line 56: Maybe we can do a shallow copy of the repo since we only want the committish?

.github/workflows/verify-generation-config.yaml
```diff
@@ -30,3 +72,44 @@ jobs:
           -e "REPO_BINDING_VOLUMES=${repo_volumes}" \
           gcr.io/cloud-devrel-public-resources/java-library-generation:latest \
           python /src/generate_repo.py generate --generation-config-yaml=/workspace/google-cloud-java/generation_config.yaml --repository-path=/workspace/google-cloud-java
+      env:
+        repo_volumes: ${{ env.repo_volumes }}
```
diegomarquezp, line 76: With this variable, maybe we can remove line 76

.github/workflows/verify-generation-config.yaml
```diff
@@ -30,3 +72,44 @@ jobs:
...
+    - name: generate pull request description
+      id: generate-description
+      shell: bash
+      run: |
+        repo_volumes="-v repo-google-cloud-java:/workspace/google-cloud-java"
```
diegomarquezp, line 81: Same, this line becomes obsolete by using the env var

.github/workflows/verify-generation-config.yaml
JoeWang1127: Is `repo_volumes` available in this step if I don't specify it in `env` section?

JoeWang1127: Done.

.github/workflows/verify-generation-config.yaml
JoeWang1127: Done, using partial clone.

diegomarquezp (approved):

.github/workflows/verify-generation-config.yaml
```diff
@@ -30,3 +72,40 @@ jobs:
...
+          gcr.io/cloud-devrel-public-resources/java-library-generation:latest \
+          python /src/generate_pr_description.py generate --generation-config-yaml=/workspace/google-cloud-java/generation_config.yaml --baseline-commit="${baseline_commit}"
+      env:
+        baseline_commit: ${{ env.baseline_commit }}
+    - name: create or update the pull request
+      shell: bash
+      run: |
+        title="chore: generate libraries at $(date)"
+        git add java-* pom.xml gapic-libraries-bom/pom.xml versions.txt generation_config.yaml
+        git commit -m "chore: ${title}"
```
suztomo, line 97: "chore" appears twice.

suztomo (approved):

.github/workflows/verify-generation-config.yaml
```diff
@@ -30,3 +72,40 @@ jobs:
...
+          -e "RUNNING_IN_DOCKER=true" \
+          -e "REPO_BINDING_VOLUMES=${repo_volumes}" \
+          gcr.io/cloud-devrel-public-resources/java-library-generation:latest \
+          python /src/generate_pr_description.py generate --generation-config-yaml=/workspace/google-cloud-java/generation_config.yaml --baseline-commit="${baseline_commit}"
+      env:
+        baseline_commit: ${{ env.baseline_commit }}
+    - name: create or update the pull request
+      shell: bash
+      run: |
+        title="chore: generate libraries at $(date)"
```
suztomo, line 95: My instinct says "chore:" does not goes to Release Please pull request. We can fix it by "BEGIN_COMMIT_OVERRIDE" if it really does not work. Let's see.

blakeli0:
Is this [PR](https://github.com/googleapis/google-cloud-java/pull/10516) an example of the generated PR?

.github/workflows/verify-generation-config.yaml
JoeWang1127: Updated the commit message.

JoeWang1127:
> Is this [PR](https://github.com/googleapis/google-cloud-java/pull/10516) an example of the generated PR?

Yes.

.github/workflows/verify-generation-config.yaml
blakeli0, line 1: This file should be renamed since it does not only verify generation config, but also creates a PR.

.github/workflows/verify-generation-config.yaml
diegomarquezp: I think it was misleading to say "76". Better said with "declaration of `repo_volumes`" but I see that you changed it. Thanks!

JoeWang1127:
`generation diff/root-pom` and `generation diff/gapic-bom` failed because the sorted entries in pom.xml are in different order when using shell `sort` and python `sorted`.

.github/workflows/verify-generation-config.yaml
blakeli0: I feel the same way. Can we make sure we understand the behavior before merging? e.g. If it is a `chore`, is release please going to include the NESTE_COMMIT in the description? If it is a `feat`, is the title going to be included? The behavior of release-please will affect how we generate the PR descriptions.

.github/workflows/verify-generation-config.yaml
```diff
@@ -30,3 +72,40 @@ jobs:
...
+        git add java-* pom.xml gapic-libraries-bom/pom.xml versions.txt generation_config.yaml
+        git commit -m "${title}"
+        if [ -z "${pr_num}" ]; then
+          git remote add monorepo https://cloud-java-bot:${GH_TOKEN}@github.com/${{ github.repository }}.git
+          git fetch -q --unshallow monorepo
+          git push -f monorepo "${branch_name}"
+          set -x
+          gh pr create --title "${title}" --label "owlbot:run" --head "${branch_name}" --body "$(cat pr_description.txt)"
```
blakeli0, line 103: Why do we still need `owlbot:run` here?

.github/workflows/verify-generation-config.yaml
```diff
@@ -30,3 +72,40 @@ jobs:
           -e "REPO_BINDING_VOLUMES=${repo_volumes}" \
           gcr.io/cloud-devrel-public-resources/java-library-generation:latest \
```
blakeli0, line 73: Ideally we should not use `latest`, but rather a versioned docker image, otherwise it is not hermetic. But since this is the first time we use the scripts and there might be bug fixes in this cycle, let's keep it for now. Once the scripts are stable, we should read the gapic-generator-java version from the build config file and use that version for the docker image.  
For now, can we extract `latest` as a parameter and share it between this step and the step below?

.github/workflows/verify-generation-config.yaml
JoeWang1127: The check is not removed yet.

.github/workflows/verify-generation-config.yaml
JoeWang1127: Done, changed to `generate-from-configuration`.

.github/workflows/verify-generation-config.yaml
blakeli0: I think we should remove the label here and disable the check for this branch, otherwise owlbot postprocessing would still be run against this PR right?

.github/workflows/verify-generation-config.yaml
JoeWang1127: According to @chingor13, release-please won't skip a commit if it's a chore, it just ignores chore commit messages.

.github/workflows/verify-generation-config.yaml
diegomarquezp: +1 to removing the required check for owlbot postprocessor and the label

.github/workflows/verify-generation-config.yaml
diegomarquezp: Just FYI, we have a release job that releases `gcr.io/cloud-devrel-public-resources/java-library-generation:${GGJ_VERSION}`, so if we are to remove it, we can use that tag

JoeWang1127: > Once the scripts are stable, we should read the gapic-generator-java version from the build config file and use that version for the docker image.

I think we should group the docker image tag with generator update so that renovate bot can update them together.

> For now, can we extract latest as a parameter and share it between this step and the step below?

Sounds good.

.github/workflows/verify-generation-config.yaml
JoeWang1127: Done.

.github/workflows/generate-from-configuration.yaml
```diff
@@ -0,0 +1,113 @@
...
+    branches:
+    - feat/hermetic-pr-creation
+  workflow_dispatch:
+
+
+jobs:
+  generate-from-configuration:
+    runs-on: ubuntu-22.04
+    env:
+      branch_name: generate-libraries
```
blakeli0, line 15: This could be a follow up PR. Can we make this a parameter so that we can test the Github action in a separate branch? Otherwise it would always override the current PR if it is not merged.

JoeWang1127: > Otherwise it would always override the current PR if it is not merged

This is an intended behavior because we don't want a lot of open PRs in the repo.

This is the [description](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.0#bookmark=id.h17h2gfiy57f) of the workflow.

blakeli0: In addition, can we change the branch name to include the branch name, something like `generate-libraries-main`? As we would need to support LTS and multiple Java versions in the future.

.github/workflows/generate-from-configuration.yaml
```diff
@@ -0,0 +1,113 @@
...
+        title="chore: generate libraries at $(date)"
+        git add java-* pom.xml gapic-libraries-bom/pom.xml versions.txt generation_config.yaml
+        # use --allow-empty because (rarely) there's no change.
+        git commit --allow-empty -m "${title}"
+        if [ -z "${pr_num}" ]; then
+          git remote add monorepo https://cloud-java-bot:${GH_TOKEN}@github.com/${{ github.repository }}.git
+          git fetch -q --unshallow monorepo
+          git push -f monorepo "${branch_name}"
+          set -x
+          gh pr create --title "${title}" --head "${branch_name}" --body "$(cat pr_description.txt)"
```
blakeli0, line 105: I guess this creates a PR against main branch by default? Can we make the target branch a parameter as well?

blakeli0 (approved):
LGTM. I think this PR should be a `chore` as it does not really change any code, otherwise it would appear in the release notes. 
A few follow up enhancements that I can think of:
1. Make this Github action more generic so that it can be reused for other use cases. e.g. Extract the target branch, googleapis commitish to parameters, do not always update googleapis commitish etc.
2. Encapsulate the PR creation logic in the docker image. So that it can be tested independently and more hermetic.  
3. Create a separate Github action or reuse this Github action to automatically update renovate bot PRs for gapic-generator-java version update.

.github/workflows/generate-from-configuration.yaml
```diff
@@ -0,0 +1,113 @@
...
+  workflow_dispatch:
+
+
+jobs:
+  generate-from-configuration:
+    runs-on: ubuntu-22.04
+    env:
+      branch_name: generate-libraries
+      library_generation_image_tag: latest
+      repo_volumes: "-v repo-google-cloud-java:/workspace/google-cloud-java"
```
blakeli0, line 17: Anything that references `google-cloud-java` should be extracted to a parameter, so that we can reuse this Github action in handwritten library repos.

.github/workflows/generate-from-configuration.yaml
blakeli0: > > Otherwise it would always override the current PR if it is not merged
> 
> This is an intended behavior because we don't want a lot of open PRs in the repo.
> 
> This is the [description](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.0#bookmark=id.h17h2gfiy57f) of the workflow.

I agree with this behavior. What I'm suggesting is to make it more flexible so that we can test the Github action without overriding the PR.

.github/workflows/generate-from-configuration.yaml
JoeWang1127: Done.

.github/workflows/generate-from-configuration.yaml
JoeWang1127: Done.

.github/workflows/generate-from-configuration.yaml
JoeWang1127: I'll leave this one as-is and we'll refactor it once we prove this is working on google-cloud-java.

generation_config.yaml
```diff
@@ -167,6 +167,14 @@ libraries:
   GAPICs:
   - proto_path: google/appengine/v1
 
+- api_shortname: apphub
```
diegomarquezp, line 170: Which script do you use to populate this yaml?

JoeWang1127: I manually added these libraries, the parameters are copied from new library generation PR.

PR #10518, opened 2024-03-11
chore: format pom

JoeWang1127:
In this PR:
- Sort entries in `pom.xml` and `gapic-libaries-bom/pom.xml`

Context: this is a preparation [step](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.0#bookmark=id.wd6oq2x4fnys) for hermetic build launch plan.

### DIFF
--- a/gapic-libraries-bom/pom.xml
+++ b/gapic-libraries-bom/pom.xml
@@ -22,16 +22,30 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-accessapproval-bom</artifactId>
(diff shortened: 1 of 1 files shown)


JoeWang1127:
"generation diff/root-pom" and "generation diff/gapic-bom" are expected to fail.

suztomo:
> "generation diff/root-pom" and "generation diff/gapic-bom" are expected to fail.

What's the plan?

JoeWang1127:
> What's the plan? I think they are already sorted. It's just different order than this pull request.

I think the order is different when sorting a string using shell and python. This PR can reduce the amount of irrelvant difference in the initial generation made by hermetic build project.

PR #10528, opened 2024-03-12
chore: restore baseline commit to last owlbot update

JoeWang1127:
The last OwlBot update pr is #10465, which has googleapis commit 97e3b44f2f08bd0ffff128872519e619750c7892.

Set baseline commit to this commitish to correctly populate release note.


diegomarquezp (approved):
It would be nice, as an improvement, to have a comment above the line with the commit message of such committish

PR #10568, opened 2024-03-19
chore: [vertexai] Change the versions to 1.0.0

ZhenyiQ:


gapic-libraries-bom/pom.xml
```diff
@@ -1207,4 +1207,4 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
-</project>
```
meltsufin, line 1210: Please remove the trailing newline.

ZhenyiQ: Done!

meltsufin (approved):

ZhenyiQ:
Thank you for the review!

PR #10577, opened 2024-03-19
feat: add `v1beta2` of `secretmanager`

diegomarquezp:


diegomarquezp:
@JoeWang1127 I see that the config yaml does not have `v1beta1`, 
https://github.com/googleapis/google-cloud-java/blob/80495ada838c38f42f8aa397b1e57681f46103bb/generation_config.yaml#L1531-L1539

but there is a proto folder in the library:
image


Was this library recently removed?

diegomarquezp:
> `v1beta1` is removed from googleapis, so we should remove it from google-cloud-java.
> 
> `v1beta2` is added in this [commit](https://github.com/googleapis/googleapis/commit/061faccb53647d5a36eaebf9f52c84c78a010dca)

Gotcha, thanks for checking. I'll follow up with removing v1beta1 from the monorepo.

blakeli0:
> > `v1beta1` is removed from googleapis, so we should remove it from google-cloud-java.
> > `v1beta2` is added in this [commit](https://github.com/googleapis/googleapis/commit/061faccb53647d5a36eaebf9f52c84c78a010dca)
> 
> Gotcha, thanks for checking. I'll follow up with removing v1beta1 from the monorepo.

Is this(removal of a GAPIC version) something our scripts can handle automatically in the future?

blakeli0:
It should probably be handled by @JoeWang1127 's upcoming scripts, for now, can you manually run the generation script and add the changes as part of this PR?

diegomarquezp:
> > > `v1beta1` is removed from googleapis, so we should remove it from google-cloud-java.
> > > `v1beta2` is added in this [commit](https://github.com/googleapis/googleapis/commit/061faccb53647d5a36eaebf9f52c84c78a010dca)
> > 
> > 
> > Gotcha, thanks for checking. I'll follow up with removing v1beta1 from the monorepo.
> 
> Is this(removal of a GAPIC version) something our scripts can handle automatically in the future?

the `deep-remove` regex is meant to reset the source folders before transferring the generated files, so they should be removed with a default-generated owlbot yaml. However, these `v1beta1` folders are being explicitly preserved:

https://github.com/googleapis/google-cloud-java/blob/0416bfbcc52ed873b0cdd3639daac4799fc71b63/java-secretmanager/.OwlBot.yaml#L22-L27 

Do we want to remove this preserve regex?

java-secretmanager/.OwlBot.yaml
diegomarquezp, line 1: > It should probably be handled by @JoeWang1127 's upcoming scripts, for now, can you manually run the generation script and add the changes as part of this PR?

@blakeli0 the generation went well after a small tweak to the owlbot yaml. This is something we need to add to the release notes

blakeli0:
> > > > `v1beta1` is removed from googleapis, so we should remove it from google-cloud-java.
> > > > `v1beta2` is added in this [commit](https://github.com/googleapis/googleapis/commit/061faccb53647d5a36eaebf9f52c84c78a010dca)
> > > 
> > > 
> > > Gotcha, thanks for checking. I'll follow up with removing v1beta1 from the monorepo.
> > 
> > 
> > Is this(removal of a GAPIC version) something our scripts can handle automatically in the future?
> 
> @blakeli0 the `deep-remove` regex is meant to reset the source folders before transferring the generated files, so they should be removed with a default-generated owlbot yaml. However, these `v1beta1` folders are being explicitly preserved:
> 
> https://github.com/googleapis/google-cloud-java/blob/0416bfbcc52ed873b0cdd3639daac4799fc71b63/java-secretmanager/.OwlBot.yaml#L22-L27
> 
> Do we want to remove this preserve regex?

SGTM.

diegomarquezp:
Looks like the deep-remove regexes only delete part of the proto and grpc libraries
https://github.com/googleapis/google-cloud-java/blob/d61155fa55dac6fd93f3d93d2a48f5ba1c969e3e/java-secretmanager/.OwlBot.yaml#L18-L19

I will add a commit to manually remove them. This confirms that it's not enough with simply removing an api version from `generation_config.yaml`

JoeWang1127 (approved):

blakeli0 (approved):

Issue #10586, opened 2024-03-20
Create a guide for RetrySettings


alicejli:
RetrySettings have been a source of customer confusion. We should create an external-facing guide to using and explaining RetrySettings within the GAPIC java client libraries

blakeli0:
Currently we don't have any public guide other than Javadocs in [RetrySettings](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java), which could take customers a while to figure out. We should add a section for it to our README, similar to [Long Running Operations](https://github.com/googleapis/google-cloud-java?tab=readme-ov-file#long-running-operations) section.

lqiu96:
@alicejli @blakeli0 Added a first draft for the README. Let me know if anything is unclear or if anything else should be added.

PR #10598, opened 2024-03-21
docs: Add Retry Guide

lqiu96:
Fixes: https://github.com/googleapis/google-cloud-java/issues/10586

Add a section to the README.md to include basic information about how Google Cloud Java Client Libraries deal with retries.

### DIFF
--- a/docs/client_retries.md
+++ b/docs/client_retries.md
@@ -101,7 +101,7 @@ the total RPC's bounds.
 Exponential backoff will retry requests with an increasing delay between each retry attempt. This retry delay value
 can be capped with a maximum retry delay value.
 
-For example, with the following retry configurations:
+For example, the following retry configurations may result in the following delay times:
 ```
 Initial Retry Delay: 100ms
 Retry Delay Multiplier: 2.0
@@ -114,27 +114,33 @@ Max Retry Delay: 500ms
 - ...
 - Attempt X: Delay 500ms
 
+Note: The example above is configured to retry indefinitely. Do set an RPC operation cap by setting the TotalTimeout
+and/or MaxAttempts value.
+
 ### Jitter
-Jitter is added variance via randomness to spread out when the RPCs are invoked. By default, Google Cloud
-Client Libraries enable jitter for retries. When jitter is enabled with exponential backoff, the client libraries
-are able to spread out the retry attempts without overwhelming the server.
+Jitter is added variance via randomness to spread out when the RPCs are invoked. Google Cloud Client Libraries 
+always enable jitter for retries. This is to help spread out the retry attempts without overwhelming the server.
 
-The jitter randomness is computed on the retry delay. Before each attempt, the retry algorithm will compute
-a random value with the between `[1, RETRY_DELAY]`. This computed value is the *approximate* delay before the request
-is sent to the server.
+The jitter random value is computed as the retry delay. Before each attempt, the retry algorithm will compute a random
+value between `[1, RETRY_DELAY]`. This computed value is the *approximate* delay before the request is sent to the
+server.
 
-For example, with the following retry configurations:
+For example, the following retry configurations utilizes Jitter and Exponential Backoff, which may result in the
+following delays times:
 ```
 Initial Retry Delay: 100ms
 Retry Delay Multiplier: 2.0
 Max Retry Delay: 500ms
 ```
-- Attempt 1: Random value between [1, 100]
-- Attempt 2: Random value between [1, 200]
-- Attempt 3: Random value between [1, 400]
-- Attempt 4: Random value between [1, 500]
+- Attempt 1: Delay a random value between `[1, 100]`ms
+- Attempt 2: Delay a random value between `[1, 200]`ms
+- Attempt 3: Delay a random value between `[1, 400]`ms
+- Attempt 4: Delay a random value between `[1, 500]`ms
 - ...
-- Attempt X: Random value between [1, 500]
+- Attempt X: Delay a random value between `[1, 500]`ms
+
+Note: The example above is configured to retry indefinitely. Do set an RPC operation cap by setting the TotalTimeout
+and/or MaxAttempts value.
 
 ## Retry Examples
 The following examples below show the behavior of some retry configurations.


README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
...
+- Max Attempts: The maximum number of attempts to perform. If this value is greater than 0, and the number of attempts reaches this limit, the logic will give up retrying even if the total retry time is still lower than TotalTimeout.
+- Total Timeout: The overall total limit before the remote call gives up completely. The higher the total timeout, the more retries can be attempted.
+- Initial Retry Delay: The delay before the first retry.
+- Retry Delay Multiplier: The change in retry delay. This value is multiplied by the previous call’s retry delay to calculate the retry delay for the next call.
+- Max Retry Delay: The limit on the value of the retry delay, so that the RetryDelayMultiplier can't increase the retry delay higher than this amount.
+- Initial RPC Timeout: The timeout for the initial RPC.
+- RPC Timeout Multiplier: The change in RPC timeout. This value is multiplied by the previous call’s timeout to calculate the timeout for the next call.
+- Max RPC Timeout: The limit of the RPC timeout, so that the RpcTimeoutMultiplier can't increase the RPC timeout higher than this amount
+
+Note: It is recommended to only set _either_ the Max Attempt or the Total Timeout.
```
blakeli0, line 439: I was not aware of this recommendation, what's the reason behind it?

README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
...
+
+See the official Google Cloud Docs: https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings
+
+#### Jitter
```
blakeli0, line 443: I don't think we should mention `Jitter` here because it is already deprecated and we should always set it to true.

README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
 }
 ```
 
+## Retries
+Client Libraries use retries to handle unexpected, transient failures (i.e. Server temporarily unavailable).
+Multiple attempts, hopefully, will result in a successful response from the server.
+
+By default, retries are configured by the Cloud Service. The configured retry parameters are defined per RPC.
```
suztomo, line 403: Would you add an example (the audience is client library users)?

README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
...
+A service *may* choose to only enable retries for a subset of RPCs. It is possible that for a single service, 
+each RPC is configured differently.
+
+### Google Cloud Client Library Retry Concepts
+Enabling retries allow an RPC multiple attempts to try and achieve a successful call. A successful call
```
suztomo, line 408:
```suggestion
Enabling retries allows an RPC multiple attempts to try and achieve a successful call. A successful call
```

The subject and verb correspondence is unclear.

README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
...
+is a response from a server that returns an `OK` Status Code (from gRPC) or a `2xx` Status Code (from HttpJson).
+
+An RPC will be retried when _both_ of the following scenarios occur:
+- Non-successful status code is received by the library and the status code is marked as retryable*
+- An RPC invocation exceeds the individual RPC bounds, but still falls within total RPC bounds**
```
suztomo, line 413: How about listing the terms such as "bounds" (or "RPC bounds")?

README.md
suztomo: From the audience's perspective, it's just retries are controlled by the responses of the service.

> The configured retry parameters are defined per RPC.

If the audience doesn't have a way to read the definition, let's remove this.

README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
...
+The following table shows the attempts:
+
+| Attempt Number 	| RPC Timeout 	| Retry Delay 	| Call Invoked 	| Call Ended 	|
+|----------------	|-------------	|-------------	|--------------	|------------	|
+| 1              	| 5000ms      	| 0ms         	| 0ms          	| 50000ms    	|
+
+#### Retry
+##### Example 1
+```java
+RetrySettings.newBuilder()
```
blakeli0, line 529: Can we use the same `RetrySettings` for all examples? I think it would be easier to process.

README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
...
+1. Create the RetrySettings class with your custom configurations
+```java
+RetrySettings customRetrySettings =
+  RetrySettings.newBuilder()
+      // ... Retry Configurations
+      .build();
+```
+2. Create the StubSettings.Builder for your client and configure it for the RPC
+```java
+EchoStubSettings.Builder echoStubSettingsBuilder = EchoStubSettings.newBuilder();
```
blakeli0, line 610: I wonder if we should use a real service(maybe language/aiplatform) for demonstration purposes as it could be easier to understand.

README.md
```diff
@@ -396,6 +396,259 @@ public CloudTasksClient getService() throws IOException {
...
+**When configuring the RPC bounds, you may configure the bounds for each attempt as well as the
+total RPC's bounds. The retry algorithm will ensure that an individual attempt's bounds falls within
+the total RPC's bounds.
+
+#### Configurable Retry Params
+- Retry Status Code: Set of failure status codes to retry on
+- Retry Time/ Attempt Bounds: Configurable retry settings (via RetrySettings class) to define the retry bounds
+
+#### RetrySettings Configurable Params
+- Max Attempts: The maximum number of attempts to perform. If this value is greater than 0, and the number of attempts reaches this limit, the logic will give up retrying even if the total retry time is still lower than TotalTimeout.
```
blakeli0, line 430: These info already exist in Javadocs, I don't think we should repeat them here, a link should be good enough IMO.

README.md
lqiu96: > Would you add an example (the audience is client library users)?

I'll move the Java-Asset example up to show the one service's default retry param values

> From the audience's perspective, it's just "retries are controlled by the responses of the service." isn't it?

Controlled by response of the service and the time/ attempt bounds.

README.md
lqiu96: My recollection was that there was an internal doc that recommended setting only either param, but I can't find it (and perhaps my memory is wrong). Looking at old tickets, it seems that maxAttempts was never implemented for any language (Java seems to be the exception). 

I see this ticket I created a while back regarding maxAttempts: https://github.com/googleapis/sdk-platform-java/issues/2306, so perhaps this recommendation shouldn't exist.

README.md
lqiu96: I think disable Jitter internally quite a bit for tests to manage flakiness. I don't know if we can fully remove it in the future (perhaps we can just mark it with `@VisibleForTests`).

I mentioned this section since I think it has a bit of behavior impact on the number of retries. If a customer is either counting (or just otel), they may be expecting a certain number of retry attempts.

README.md
lqiu96: There are slight differences in each RetrySetting. I can try and highlight the differences to make it clearer.

README.md
lqiu96: Makes sense. I'll use Asset since it's already used above.

README.md
lqiu96: Sounds good. I'll remove.

README.md
lqiu96: Sounds good, added a blurb above to describe it.

docs/retries.md
```diff
@@ -0,0 +1,267 @@
+# Retries
```
blakeli0, line 1: Let's use the term `Client Side Retries`, as customers may think these info are passed to the backend services, which happened in the past.

docs/retries.md
```diff
@@ -0,0 +1,267 @@
...
+AssetServiceStubSettings.Builder assetStubSettingsBuilder = AssetServiceStubSettings.newBuilder();
+assetStubSettingsBuilder
+  .exportAssetsSettings()
+  // Set your custom Retry Settings
+  .setRetrySettings(customRetrySettings)
+  // Set your custom Retryable Codes
+  .setRetryableCodes(ImmutableSet.of(StatusCode.Code.DEADLINE_EXCEEDED));
+```
+
+The snippet above is using a mock service (i.e. EchoClient). The code snippet above is setting
```
blakeli0, line 236: We can change this part to mention `AssetServiceClient` instead now.

docs/retries.md
lqiu96: Sounds good, will update.

docs/retries.md
lqiu96: Ah, yep! Will update.

docs/client_retries.md
```diff
@@ -0,0 +1,267 @@
...
+```
+Individual RPC Bounds (an attempt) are controlled by the following settings:
+- setInitialRetryDelay
+- setRetryDelayMultiplier
+- setMaxRetryDelay
+- setInitialRpcTimeout
+- setRpcTimeoutMultiplier
+- setMaxRpcTimeout
+
+Total RPC Bounds are controlled by the following settings:
```
alicejli, line 76: maybe add `Total RPC Bounds (an operation) are controlled` - to differentiate from individual RPC Bounds?

docs/client_retries.md
```diff
@@ -0,0 +1,267 @@
...
+- setAttemptCount
+
+An RPC will be retried when _both_ of the following scenarios occur:
+- Non-successful status code is received by the library and the status code is marked as retryable*
+- An RPC invocation exceeds the individual RPC bounds, but still falls within total RPC bounds**
+
+Note: If only one (or neither) of the scenarios above are true, then the RPC will not be retried.
+i.e. If the total timeout has not been exceeded, but the latest attempt receives a non-retryable status code.
+
+*The client library will mark a status code as retryable internally. It is marked as retryable if the response's
```
alicejli, line 87: what does `retryable internally` mean?

docs/client_retries.md
```diff
@@ -0,0 +1,267 @@
...
+status code matches with any of an RPC's configured retryable status codes.
+
+**When configuring the RPC bounds, you may configure the bounds for each attempt as well as the
+total RPC's bounds. The retry algorithm will ensure that an individual attempt's bounds falls within
+the total RPC's bounds.
+
+### Jitter
+Jitter is added variance via randomness to spread out when the RPCs are invoked. By default, Google Cloud
```
alicejli, line 95: Is there an official link or something to what `Jitter` is?

README.md
```diff
@@ -444,11 +444,11 @@ Note: The client library handles the Operation's polling mechanism for you. By d
 to manually poll the status yourself.
 
 ### Default LRO Values
-Each LRO RPC has a pre-configured default values. You can find these values by 
+Each LRO RPC has a pre-configured default values. You can find these values by
```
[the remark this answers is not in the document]
lqiu96: Thanks, updated!

docs/client_retries.md
```diff
@@ -0,0 +1,267 @@
+# Client Side Retries
+Client Libraries use retries to handle unexpected, transient failures (i.e. server temporarily unavailable).
```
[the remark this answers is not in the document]
lqiu96: Updated. I think it should be clearer now that it's not a status code or message back.

docs/client_retries.md
```diff
@@ -0,0 +1,267 @@
...
+Example:
+```java
+      definitions.put(
+          "retry_policy_1_codes",
+          ImmutableSet.copyOf(
+              Lists.<StatusCode.Code>newArrayList(
+                  StatusCode.Code.DEADLINE_EXCEEDED, StatusCode.Code.UNAVAILABLE)));
+```
+
+Retry Params are configured [here](https://github.com/googleapis/google-cloud-java/blob/d9da511b4b56302e509abe8b2d919a15ea7dcae7/java-asset/google-cloud-asset/src/main/java/com/google/cloud/asset/v1/stub/AssetServiceStubSettings.java#L1086-L1155)
```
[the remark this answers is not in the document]
lqiu96: Done

docs/client_retries.md
```diff
@@ -0,0 +1,267 @@
...
+      settings =
+          RetrySettings.newBuilder()
+              .setInitialRpcTimeout(Duration.ofMillis(60000L))
+              .setRpcTimeoutMultiplier(1.0)
+              .setMaxRpcTimeout(Duration.ofMillis(60000L))
+              .setTotalTimeout(Duration.ofMillis(60000L))
+              .build();
+```
+
+The default configurations are mapping [here](https://github.com/googleapis/google-cloud-java/blob/d9da511b4b56302e509abe8b2d919a15ea7dcae7/java-asset/google-cloud-asset/src/main/java/com/google/cloud/asset/v1/stub/AssetServiceStubSettings.java#L1306-L1474)
```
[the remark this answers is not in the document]
lqiu96: Thanks, done.

docs/client_retries.md
lqiu96: Added an explanation for operation/ attempt.

docs/client_retries.md
lqiu96: Removed it. Meant that gax-java will handle marking it as retryable or not.

docs/client_retries.md
lqiu96: We don't have an official guide of jitter. I can look for a more public guide regarding this, but I don't think google has one.

docs/client_retries.md
```diff
@@ -0,0 +1,267 @@
...
+Multiple attempts, hopefully, will result in a successful response from the server.
+
+By default, retries are configured by the Cloud Service. The configured retry parameters are defined per RPC.
+A service *may* choose to only enable retries for a subset of RPCs. It is possible that for a single service,
+each RPC is configured differently.
+
+## RPC Retry Configurations
+Using Java-Asset v3.41.0 as an example, the default configurations are configured in the following files:
+
+Retry Status Codes are configured [here](https://github.com/googleapis/google-cloud-java/blob/d9da511b4b56302e509abe8b2d919a15ea7dcae7/java-asset/google-cloud-asset/src/main/java/com/google/cloud/asset/v1/stub/AssetServiceStubSettings.java#L1058-L1082)
```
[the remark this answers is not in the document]
lqiu96: I see. Let me try and move `Configurable Retry Params` section up.

docs/client_retries.md
alicejli: Okay sorry I think the capitalization threw me off; I thought it was like an official library or something. But seems like its just a general software term.

docs/client_retries.md
```diff
@@ -0,0 +1,320 @@
...
+per RPC. A service *may* choose to only enable retries for a subset of RPCs. It is possible that each RPC for
+a service is configured differently.
+
+## Retry Parameters
+Client libraries have two types of retry parameters to configure:
+1. Retry Status Code: Set of failure status codes to retry on
+2. Retry Time/ Attempt Bounds: Configurable [RetrySettings](https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings) to define the bounds
+
+### RPC Retry Configurations
+Using Java-Asset v3.41.0 as an example, the default retry configurations are defined in the following places:
```
burkedavison, line 15: The formatting here is a bit confusing as it is meant to be read as a list, but isn't formatted as one.

Consider adding bullets, and indenting each example with its bullet.

## For example...

Using Java-Asset v3.41.0 as an example, the default retry configurations are defined in the following places:
* Retry status codes...

    Example: 
    ```java
    java.logic;
    ```

* Retry parameters are...

    Example:
    ```java
    java.logic;
    ````

docs/client_retries.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+The configurations above are set [here](https://github.com/googleapis/google-cloud-java/blob/d9da511b4b56302e509abe8b2d919a15ea7dcae7/java-asset/google-cloud-asset/src/main/java/com/google/cloud/asset/v1/stub/AssetServiceStubSettings.java#L1306-L1474)
+
+Example:
+```java
+builder
+  .exportAssetsSettings()
+  .setRetryableCodes(RETRYABLE_CODE_DEFINITIONS.get("no_retry_0_codes"))
+  .setRetrySettings(RETRY_PARAM_DEFINITIONS.get("no_retry_0_params"));
```
burkedavison, line 48: I'm not understanding how this information is to be used by a developer. Is this section just FYI on how the internals are implemented, or is this information important because developers will need to modify this when X.

(The examples below don't seem to require knowledge of the `RETRYABLE_CODE_DEFINITIONS` and `RETRY_PARAM_DEFINITIONS` collections.)

docs/client_retries.md
lqiu96: Updating to be be indented in a list.

docs/client_retries.md
lqiu96: I intended this section to be to show where users can find the default retry configurations. I'll update this section to be clearer.

> (The examples below don't seem to require knowledge of the RETRYABLE_CODE_DEFINITIONS and RETRY_PARAM_DEFINITIONS collections.)

Those definitions aren't important and are internal details to gax.

docs/client_retries.md
```diff
@@ -0,0 +1,321 @@
...
+The default retry configurations are defined in the generated `{Client}StubSettings` file. Using the ExportAssets RPC in
+Java-Asset v3.41.0 as an example, the default retry configurations are defined in the following places:
+<br>
+- Retry Status Codes are configured [here](https://github.com/googleapis/google-cloud-java/blob/d9da511b4b56302e509abe8b2d919a15ea7dcae7/java-asset/google-cloud-asset/src/main/java/com/google/cloud/asset/v1/stub/AssetServiceStubSettings.java#L1058-L1082). Example:
+```java
+ImmutableMap.Builder<String, ImmutableSet<StatusCode.Code>> definitions = ImmutableMap.builder();
+definitions.put("no_retry_0_codes", ImmutableSet.copyOf(Lists.<StatusCode.Code>newArrayList()));
+// ... More StatusCode configurations
+RETRYABLE_CODE_DEFINITIONS = definitions.build();
+```
```
burkedavison, line 24: nit: consider indenting these snippets by ~4 spaces to make them render as indented under their parent bullet.

docs/client_retries.md
```diff
@@ -0,0 +1,321 @@
...
+- [setInitialRpcTimeout](https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings.Builder#com_google_api_gax_retrying_RetrySettings_Builder_setInitialRpcTimeout_org_threeten_bp_Duration_)
+- [setRpcTimeoutMultiplier](https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings.Builder#com_google_api_gax_retrying_RetrySettings_Builder_setRpcTimeoutMultiplier_double_)
+- [setMaxRpcTimeout](https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings.Builder#com_google_api_gax_retrying_RetrySettings_Builder_setMaxRpcTimeout_org_threeten_bp_Duration_)
+
+Total RPC Bounds (an operation) are controlled by the following settings:
+- [setTotalTimeout](https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings.Builder#com_google_api_gax_retrying_RetrySettings_Builder_setTotalTimeout_org_threeten_bp_Duration_)
+- [setMaxAttempts](https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings.Builder#com_google_api_gax_retrying_RetrySettings_Builder_setMaxAttempts_int_)
```
burkedavison, line 82: Consider adding a minimal explanatory phrase to each of these, rather than requiring the user to click on each of the links to see the full documentation on what the setting means.

Example:
* setInitialRetryDelay : delay time before the first retry
* setRetryDelayMultiplier : exponential backoff multiplier
* setMaxRetryDelay : upper bound of delay time after applying the multiplier
* ...

lqiu96: Added a short blurb for each one

docs/client_retries.md
lqiu96: Done

lqiu96:
@burkedavison @suztomo @blakeli0 @alicejli Could I get a re-review on this whenever you all get time? Plan is to add details for retries, LRO, endpoint, client initialization, etc. so that we can reference this on future customer issues.

burkedavison (approved):

alicejli (approved):
thanks for putting together! this will be very helpful to reference for customers. at some point it'd be nice to get a version of this guide onto Cloud RAD.

docs/client_retries.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Exponential Backoff
+Exponential backoff will retry requests with an increasing delay between each retry attempt. This retry delay value
+can be capped with a maximum retry delay value.
+
+For example, with the following retry configurations:
+```
+Initial Retry Delay: 100ms
+Retry Delay Multiplier: 2.0
+Max Retry Delay: 500ms
```
JoeWang1127, line 108: Is this setting will cause retry indefinite times?

docs/client_retries.md
```diff
@@ -0,0 +1,321 @@
...
+- Attempt 1: Delay 100ms
+- Attempt 2: Delay 200ms
+- Attempt 3: Delay 400ms
+- Attempt 4: Delay 500ms
+- ...
+- Attempt X: Delay 500ms
+
+### Jitter
+Jitter is added variance via randomness to spread out when the RPCs are invoked. By default, Google Cloud
+Client Libraries enable jitter for retries. When jitter is enabled with exponential backoff, the client libraries
```
blakeli0, line 119: "When jitter is enabled" implies that it can be disabled, but `setJittered` is now deprecated. Can we change the wording to indicate that `Jitter is always enabled`?

blakeli0 (approved):
LGTM other than one minor comment.

docs/client_retries.md
lqiu96: Will do, thanks!

docs/client_retries.md
lqiu96: Yep, this current configuration implies infinite retry times. I'll make that clearer in the example. Thanks!

PR #10679, opened 2024-04-10
chore: Remove owlbot yaml files

blakeli0:
Remove owlbot yaml files under `.github`. Hopefully this change would stop Owlbot PRs from creating, otherwise we may have to rename all the `.Owlbot.yaml` files in sub modules.


diegomarquezp (approved):

diegomarquezp:
Also hoping for this being the only necessary change to stop the owlbot service from creating PRs

PR #10693, opened 2024-04-12
chore: update hermetic library generation workflow

JoeWang1127:
In this PR:
- Add a workflow, when `generation_config.yaml` is changed in a pull request, to generate changed libraries and push commits to the corresponding pull request.
- Update workflow only to update googleapis commit.
- Update workflow only to add library configuration to `generation_config.yaml`.
- Change image tag to `2.39.0`

Follow up of https://github.com/googleapis/sdk-platform-java/pull/2616

For the goal of series of PRs, please refer to [improvement proposal](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.g3vua2kd06gx#bookmark=id.72s3ukwwzevo).

### DIFF
--- a/.github/workflows/update_googleapis_commit.yaml
+++ b/.github/workflows/update_googleapis_commit.yaml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       # the branch into which the pull request is merged
-      base_branch: lts-11
+      base_branch: main
     steps:
     - uses: actions/checkout@v4
     - name: Update googleapis commit to latest


JoeWang1127:
The workflows are tests in a forked repo:
https://github.com/JoeWang1127/google-cloud-java/pull/14 (update a library)
https://github.com/JoeWang1127/google-cloud-java/pull/13 (update googleapis commit)

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -14,14 +12,59 @@ jobs:
...
-      with:
-        ref: ${{ github.base_ref }}
-        path: baseline
-        sparse-checkout: generation_config.yaml
-    - name: diff
       shell: bash
       run: |
-        diff generation_config.yaml baseline/generation_config.yaml
+        git show ${{ github.base_ref }}:generation_config.yaml > baseline_generation_config.yaml
+    - name: show configuration diff
```
blakeli0, line 28: Can we extract most of the logics here to a shell script so that we can call the script directly here? That way we can easily call it in local and possibly in a non-Github environment in the future.
Ideally, we should try to put this logic into Docker container as well, so that this logic could be reused by handwritten libraries, otherwise we would need to duplicate this in every handwritten repos. Bake it into Docker can be a future enhancement, but I think we should at least extract it to a separate script.

.github/workflows/update-googleapis-commit.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+jobs:
+  update-googleapis-commit:
+    runs-on: ubuntu-22.04
+    env:
+      # the branch into which the pull request is merged
+      base_branch: main
+    steps:
+    - uses: actions/checkout@v4
+    - name: Setup branch for pull request
```
blakeli0, line 16: Same comment as above. Let's extract it to a script for now, and think about ways to make it more reusable.

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: Move command lines to a script.

.github/workflows/update-googleapis-commit.yaml
JoeWang1127: Moved command lines to a script.

JoeWang1127:
Should we remove "Verify GAPIC client library generation" or rewrite it using hermetic build script?

.github/workflows/generate_new_client_hermetic_build.yaml
```diff
@@ -1,7 +1,7 @@
 name: Generate new GAPIC client library (Hermetic Build)
 on:
   workflow_dispatch:
-      # some inputs are ommited due to limit of 10 input arguments
+      # some inputs are emitted due to limit of 10 input arguments
```
diegomarquezp, line 4: nit: `ommited` should be `omitted`

.github/workflows/generated_files_sync.yaml
diegomarquezp, line 1: nit (maybe not in scope): What's the purpose of this script? Maybe a one-liner comment can help.

generation/hermetic_library_generation.sh
```diff
@@ -0,0 +1,113 @@
...
+fi
+
+volume_name="repo"
+workspace_name="/workspace/repo"
+repo_volumes="${volume_name}:${workspace_name}"
+baseline_generation_config="baseline_generation_config.yaml"
+message="chore: generate libraries at $(date)"
+
+git checkout "${target_branch}"
+git checkout "${current_branch}"
```
diegomarquezp, line 77: qq: Has this double `checkout` statement a special effect?

diegomarquezp (approved):
LGTM, just a couple nits and a question

.github/workflows/generated_files_sync.yaml
JoeWang1127: There's a one-liner comment in each job in the workflow.

.github/workflows/generate_new_client_hermetic_build.yaml
JoeWang1127: done.

generation/hermetic_library_generation.sh
JoeWang1127: In the action, the fetch-depth is 0 so it needs `git checkout "${target_branch}"` to load target_branch.

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -1,27 +1,40 @@
...
-      - generation_config.yaml
+    - "generation_config.yaml"
+
 jobs:
   library_generation:
     runs-on: ubuntu-latest
     env:
-      library_generation_image_tag: latest
-      repo_volumes: "-v repo-google-cloud-java:/workspace/google-cloud-java"
+      library_generation_image_tag: 2.39.0
```
blakeli0, line 26: Is this configured to be updated renovate bot?

generation/new_client/new-client.py
```diff
@@ -250,7 +250,7 @@ def generate(
     )
 
     # In monorepo, .OwlBot.yaml needs to be in the directory of the module.
-    owlbot_yaml_location_from_module = ".OwlBot.yaml"
+    owlbot_yaml_location_from_module = ".OwlBot-hermetic.yaml"
```
blakeli0, line 253: I purposely didn't change this in my PR because `new-client.py` is the old process and should be removed soon.

.github/workflows/generated_files_sync.yaml
```diff
@@ -20,7 +20,7 @@ on:
   pull_request:
 name: generation diff
 env:
-  library_generation_image_tag: latest
+  library_generation_image_tag: 2.39.0
```
blakeli0, line 23: Is this configured to be updated renovate bot?

generation/hermetic_library_generation.sh
```diff
@@ -0,0 +1,113 @@
...
+# copy generation configuration from target branch to current branch.
+git show "${target_branch}":"${generation_config}" > "${baseline_generation_config}"
+diff "${generation_config}" "${baseline_generation_config}" || echo "config diff"
+# bind docker volume to include google-cloud-java in docker running environment.
```
blakeli0, line 81: This script is not specific to `google-cloud-java` anymore, can we mention something more generic?

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -1,27 +1,40 @@
...
+# GitHub action job to test core java library features on
+# downstream client libraries before they are released.
 name: Hermetic library generation upon generation config change through pull requests
 on:
   pull_request:
-    types:
-      - synchronize
     paths:
-      - generation_config.yaml
+    - "generation_config.yaml"
```
blakeli0, line 20: Is the trigger only for main branch? If yes, can this be configured to be triggered for other branches, like a future LTS branch?

generation/hermetic_library_generation.sh
blakeli0, line 1: Really like this script! This script is generic enough to be reused by any repo. We need to come up with a way to reused this script, but worst case we can just copy this script to every handwritten repos.

generation/hermetic_library_generation.sh
```diff
@@ -0,0 +1,113 @@
...
+  -v /var/run/docker.sock:/var/run/docker.sock \
+  -e "RUNNING_IN_DOCKER=true" \
+  -e "REPO_BINDING_VOLUMES=-v ${repo_volumes}" \
+  gcr.io/cloud-devrel-public-resources/java-library-generation:"${image_tag}" \
+  python /src/cli/entry_point.py generate \
+  --baseline-generation-config-path="${workspace_name}/${baseline_generation_config}" \
+  --current-generation-config-path="${workspace_name}/${generation_config}" \
+  --repository-path="${workspace_name}"
+# commit the change to the pull request.
+[ -z "$(git config user.email)" ] && git config --global user.email "cloud-java-bot@google.com"
```
blakeli0, line 104: I think we can keep the `git config` part in the Github action, so that everything else in this script can be run in local environment.

.github/workflows/generated_files_sync.yaml
blakeli0, line 1: Can you remind me what this file is used for?

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: I plan to update renovate config in separate PR.

.github/workflows/generated_files_sync.yaml
JoeWang1127: I plan to update renovate config in separate PR.

generation/new_client/new-client.py
JoeWang1127: Sure, I'll restore this change.

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: This can be triggered when a pull request changes `generation_config.yaml`.

Therefore, it can be used in any branch.

generation/hermetic_library_generation.sh
JoeWang1127: Done.

generation/new_client/new-client.py
JoeWang1127: Done.

.github/workflows/generated_files_sync.yaml
JoeWang1127: This workflow is used to verify generated files, e.g., `owlbot.py`, are not modified by pull request. Each job verifies one file.

blakeli0: What if we do want to update `owlbot.py` for a certain library? Also does it mean we run the generation for the whole repo on every PR?

.github/workflows/hermetic_library_generation.yaml
blakeli0: I don't see any configuration for a target branch, does it mean if I create a PR with `generation_config.yaml` changes against a `release/2.x` branch, this workflow would be triggered too?

.github/workflows/update_googleapis_commit.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - cron: '* 2 * * *'
+  workflow_dispatch:
+
+
+jobs:
+  update-googleapis-commit:
+    runs-on: ubuntu-22.04
+    env:
+      # the branch into which the pull request is merged
+      base_branch: lts-11
```
blakeli0, line 28: Is this `lts-11` intentional?

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: Yes.

blakeli0:
> Should we remove "Verify GAPIC client library generation" or rewrite it using hermetic build script?

I think we can remove it as the logic of "new client generation" is part of the hermetic build scripts now.

.github/workflows/generated_files_sync.yaml
JoeWang1127: > What if we do want to update owlbot.py for a certain library?

The check will fail. This is a not required check only to remain the author the change is indeed intentional.

> Also does it mean we run the generation for the whole repo on every PR?

root pom and gapic-libraries-bom are generated through hermetic build cli script (just like repo-level post processing). Other checks are done through shell script in this repo, which are not generating any java file, so they are relatively fast.

.github/workflows/update_googleapis_commit.yaml
JoeWang1127: No, this is a test value. I changed it to `main`.

JoeWang1127:
> I think we can remove it as the logic of "new client generation" is part of the hermetic build scripts now.

I'll remove this check in another PR.

blakeli0 (approved):
LGTM. Please update the PR title to make it more meaningful before merging, and update relevant docs after merging.

PR #10696, opened 2024-04-13
chore: Disable OwlBot PRs.

blakeli0:
## Why
Now that we set up [daily hermetic build generation](https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/generate-from-configuration.yaml) which would creates one PR for googleapis proto changes for all services, and it has been working consistently for the last ~4 weeks, we can safely disable OwlBot PRs.
## How
OwlBot is installed org wide and we can not disable it for a single repo. The bot uses the existence of `.OwlBot.yaml` to determine if a PR needs to be created. Hence, in order to disable OwlBot PRs, all `.OwlBot.yaml` files are renamed to `.Owlbot-hermetic.yaml`, all references to `.OwlBot.yaml` are also updated.

## Next
Once this PR is merged, the hermetic build generation will start to fail due to the assumption in hermetic build scripts that each library must contains a `.OwlBot.yaml` file. A follow up PR will be created in `sdk-platform-java` to address it.

### DIFF
--- a/generation/new_client/new-client.py
+++ b/generation/new_client/new-client.py
@@ -250,7 +250,7 @@ def generate(
     )
 
     # In monorepo, .OwlBot.yaml needs to be in the directory of the module.
-    owlbot_yaml_location_from_module = ".OwlBot-hermetic.yaml"
+    owlbot_yaml_location_from_module = ".OwlBot.yaml"
     # create owlbot config
     templates.render(
         template_name="owlbot.yaml.monorepo.j2",


JoeWang1127:
Do you use an command to rename the files?

JoeWang1127 (approved):

blakeli0:
> Do you use an command to rename the files? Could you add the command in the description?

Yes, this is the command I used 
```find . -name ".OwlBot.yaml" -type f -exec bash -c 'mv "$0" "${0%.OwlBot.yaml}.OwlBot-hermetic.yaml"' {} \```

diegomarquezp (approved):

PR #10729, opened 2024-04-22
chore: reduce circular generation

JoeWang1127:
In this PR:
- Only generate the libraries if the last commit in the current branch has changed `generation_config.yaml`.
- Restore trigger in `.github/workflows/update_googleapis_commit.yaml` and `.github/workflows/hermetic_library_generation.yaml`
- Remove `.github/workflows/generate-from-configuration.yaml`

Example pull request: https://github.com/JoeWang1127/google-cloud-java/pull/29, the `hermetic_library_generation` workflow triggered twice in this PR, but the second time exited early.

### DIFF
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -24,7 +24,7 @@ jobs:
   library_generation:
     runs-on: ubuntu-latest
     env:
-      library_generation_image_tag: 2.40.0
+      library_generation_image_tag: 2.39.0
     steps:
     - uses: actions/checkout@v4
       with:


generation/hermetic_library_generation.sh
```diff
@@ -75,6 +75,13 @@ message="chore: generate libraries at $(date)"
 
 git checkout "${target_branch}"
 git checkout "${current_branch}"
+# if the last commit doesn't contain changes to generation configuration,
```
blakeli0, line 78: In addition to or in place of the comments here, can we add a description to the hermetic_library_generation workflow to describe the behavior?

generation/hermetic_library_generation.sh
```diff
@@ -75,6 +75,13 @@ message="chore: generate libraries at $(date)"
...
+# do not generate again as the result will be the same.
+contains_config_change=$(git diff-tree --no-commit-id --name-only HEAD~1..HEAD -r | grep "${generation_config}")
+if [[ "${contains_config_change}" == "" ]]; then
+    echo "The last commit doesn't contain configuration change, exit."
```
blakeli0, line 82:
```suggestion
    echo "The last commit doesn't contain any changes to the generation_config.yaml, skipping the whole generation process."
```

generation/hermetic_library_generation.sh
```diff
@@ -102,7 +109,13 @@ docker run \
   --repository-path="${workspace_name}"
 # commit the change to the pull request.
 git add java-* pom.xml gapic-libraries-bom/pom.xml versions.txt "${generation_config}"
-git commit --allow-empty -m "${message}"
+changed_files=$(git diff --cached --name-only)
+if [[ "${changed_files}" == "" ]]; then
+    echo "Nothing to commit, exit."
```
blakeli0, line 114: Can we provide a more verbose message? Something like "There is no generated code change with the generation config change {insert config diff}."
In addition, this should be a very rare case, right? The only case I can think of is there are googleapis committish change but the change is not for Java.

generation/hermetic_library_generation.sh
```diff
@@ -102,7 +109,13 @@ docker run \
   --repository-path="${workspace_name}"
 # commit the change to the pull request.
 git add java-* pom.xml gapic-libraries-bom/pom.xml versions.txt "${generation_config}"
```
blakeli0, line 111: I might missed this from last PR, why we need to add `versions.txt` and `generation_config`? I don't think our scripts are touching those files.

generation/hermetic_library_generation.sh
JoeWang1127: I added a comment at the beginning of this script.

blakeli0: Actually I mean the workflow `hermetic_library_generation.yaml` not the script, as that is usually the entry point people start troubleshooting things. But this change looks good too.

generation/hermetic_library_generation.sh
JoeWang1127: Done.

generation/hermetic_library_generation.sh
JoeWang1127: > Something like "There is no generated code change with the generation config change {insert config diff}."

Done.

> In addition, this should be a very rare case, right? The only case I can think of is there are googleapis committish change but the change is not for Java.

Yes, it is a rare case.

generation/hermetic_library_generation.sh
JoeWang1127: I removed `generation_config`.

A new client generation will change `versions.txt` (adding new entry).

generation/hermetic_library_generation.sh
```diff
@@ -75,9 +78,19 @@ message="chore: generate libraries at $(date)"
...
+contains_config_change=$(git diff-tree --no-commit-id --name-only HEAD~1..HEAD -r | grep "${generation_config}")
+if [[ "${contains_config_change}" == "" ]]; then
+    echo "The last commit doesn't contain any changes to the generation_config.yaml, skipping the whole generation process."
+    exit 0
+fi
 # copy generation configuration from target branch to current branch.
 git show "${target_branch}":"${generation_config}" > "${baseline_generation_config}"
-diff "${generation_config}" "${baseline_generation_config}" || echo "config diff"
+config_diff=$(diff "${generation_config}" "${baseline_generation_config}")
+echo "Configuration diff:"
```
blakeli0, line 91: Can we move this diff logging to right before `git commit` to prevent possible double logging?

blakeli0 (approved):

blakeli0:
Not sure if it matters, `Verify GAPIC client library generation` is failing, please take a look.

JoeWang1127:
> Not sure if it matters, `Verify GAPIC client library generation` is failing, please take a look.

I don't think it's related to this pull request. It used new-client.py.

generation/hermetic_library_generation.sh
JoeWang1127: done.

diegomarquezp (approved):
IIUC the PR title should be "reduce [circular | cyclical] generation"

JoeWang1127:
We need to update the version in the template.

```
diff --git a/pom.xml b/pom.xml
index 3a[8](https://github.com/googleapis/google-cloud-java/actions/runs/8898222544/job/24434897316?pr=10729#step:5:9)11bd99..aba7f9fe2 100644
--- a/pom.xml
+++ b/pom.xml
@@ -207,7 +207,7 @@
       <plugin>
         <artifactId>maven-deploy-plugin</artifactId>
-        <version>3.1.2</version>
+        <version>3.1.1</version>
         <configuration>
           <skip>true</skip>
         </configuration>
```

blakeli0:
> We need to update the version in the template.
> 
> ```
> diff --git a/pom.xml b/pom.xml
> index 3a[8](https://github.com/googleapis/google-cloud-java/actions/runs/8898222544/job/24434897316?pr=10729#step:5:9)11bd99..aba7f9fe2 100644
> --- a/pom.xml
> +++ b/pom.xml
> @@ -207,7 +207,7 @@
>        <plugin>
>          <groupId>org.apache.maven.plugins</groupId>
>          <artifactId>maven-deploy-plugin</artifactId>
> -        <version>3.1.2</version>
> +        <version>3.1.1</version>
>          <configuration>
>            <skip>true</skip>
>          </configuration>
> ```
> 
> Created [googleapis/sdk-platform-java#2704](https://github.com/googleapis/sdk-platform-java/pull/2704)

I think we should either 
1. Disable the renovate bot config for any Maven dependencies managed in our template.
2. Or extend google-cloud-java's root pom from java-share-config so that we don't have to manage the versions of `maven-deploy-plugin` and `nexus-staging-maven-plugin`.
WDYT @suztomo ?

JoeWang1127:
> I think we should either
> 
> 1. Disable the renovate bot config for any Maven dependencies managed in our template.
> 2. Or extend google-cloud-java's root pom from java-share-config so that we don't have to manage the versions of `maven-deploy-plugin` and `nexus-staging-maven-plugin`.
>    WDYT @suztomo ?

For me, I prefer the 2nd option. 

However, I don't think this issue will block merging this PR as the sdk-platform-java update will fix this issue temporarily and we can implement a fix later.

PR #10773, opened 2024-05-01
chore: remove path filter

JoeWang1127:
In this PR:
- Remove path filter in hermetic library generation workflow because the script that called by the workflow can handle a pull request which doesn't contain changes to `generation_config.yaml`.
- Do not allow manually invocation because the workflow requires a pull request.
- Do not run hermetic library generation workflow after merging a pull request.

After merging this PR, the workflow will be triggered whenever there's a pull request is opened or edited so that we can make it required.

### DIFF
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -16,8 +16,7 @@
 name: Hermetic library generation upon generation config change through pull requests
 on:
   pull_request:
-    # do not run this workflow when merging the pull request.
-    types: [ opened, reopened, edited, synchronize ]
+
 jobs:
   library_generation:
     runs-on: ubuntu-latest


.github/workflows/hermetic_library_generation.yaml
```diff
@@ -15,10 +15,7 @@
 # downstream client libraries before they are released.
 name: Hermetic library generation upon generation config change through pull requests
 on:
-  workflow_dispatch:
   pull_request:
-    paths:
```
blakeli0, line 20: Can you remind me what is the trigger now? I guess we don't need it because the logic [here](https://github.com/googleapis/google-cloud-java/blob/main/generation/hermetic_library_generation.sh#L84-L87) should cover all the cases?
I guess it is the same reason we don't allow manual invocation anymore, because manual invocation would not do anything due to no generation_config.yaml change?

JoeWang1127: > Can you remind me what is the trigger now?

Whenever there's a pull request open/edited in the repo, the workflow will be triggered.

We don't need the `paths` since the script can handle it (the logic you mentioned). All we need is a pull request and manual invocation doesn't have one.

blakeli0: Thanks, do you mind adding these info to the PR description?

blakeli0 (approved):
LGTM. Please update the PR description before merging.

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: Done.

PR #10782, opened 2024-05-03
deps: update gapic generator java to 2.40.0

JoeWang1127:


blakeli0 (approved):

java-admanager/ad-manager-bom/pom.xml
```diff
@@ -24,7 +24,7 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>com.google.api-ads</groupId>
+        <groupId>com.google.api-ads.api.grpc</groupId>
```
blakeli0, line 27: Is this intentional?

java-admanager/pom.xml
```diff
@@ -46,4 +46,4 @@
     <module>ad-manager-bom</module>
   </modules>
 
-</project>
+  </project>
```
blakeli0, line 49: Can we revert this change as well? Also do we know where this change is from?

JoeWang1127: This also comes from `fix-poms.py`.

pom.xml is not rendered from templates when updates so it's not clear to me how this indentation come.

blakeli0 (approved):

PR #10849, opened 2024-05-16
chore: remove consolidate config check

JoeWang1127:
In this PR:
- Remove consolidate config check.

Pom.xml is generated/updated by hermetic build, this check is redundant.

### DIFF
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -19,6 +19,8 @@ on:
 
 jobs:
   library_generation:
+    # skip pull requests come from a forked repository
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     env:
       library_generation_image_tag: 2.40.0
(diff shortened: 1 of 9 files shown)


generation/consolidate_config.sh
```diff
@@ -80,7 +86,9 @@ removeElement 'organization'
 removeElement 'scm'
 removeElement 'issueManagement'
 removeElement 'licenses'
-removeElement 'profiles'
+# Do not remove profiles in java-compute because the profile is used to
+# speed up native image test, see https://github.com/googleapis/google-cloud-java/pull/10827
+removeElement 'profiles' 'java-compute'
```
suztomo, line 91: > removeElement 'profiles' 'java-compute'

This reads to remove the profiles element from (only) java-compute. I know the implementation is the opposite. Can you make the input design of "removeElement" better?

JoeWang1127: I changed the command to
```
removeElement --element 'profiles' --excludes 'java-compute'
```

suztomo (approved):

PR #10955, opened 2024-06-11
chore: update generation config

JoeWang1127:
In this PR:
- Update generator version to latest in `generation_config.yaml`.
- Update libraries-bom version to latest in `generation_config.yaml`.
- Move workflow related shell scripts to `.github/scripts`.
- Always update pr title/body.

Example run: https://github.com/JoeWang1127/google-cloud-java/actions/runs/9472827989/job/26098842644

### DIFF
--- a/.github/scripts/update_generation_config.sh
+++ b/.github/scripts/update_generation_config.sh
@@ -11,20 +11,11 @@ set -e
 # 3. jq
 
 # Utility functions
-
-# Get the major version from a key in the generation config.
-function get_major_version() {
-    local key_word=$1
-    major=$(grep "${key_word}" "${generation_config}" | cut -d ':' -f 2 | xargs | cut -d '.' -f 1)
-    echo "${major}"
-}
-
-# Get the latest released version within a major version of a Maven artifact.
+# Get the latest released version of a Maven artifact.
 function get_latest_released_version() {
     local group_id=$1
     local artifact_id=$2
-    local major_version=$3
-    latest=$(curl -s "https://search.maven.org/solrsearch/select?q=g:${group_id}+AND+a:${artifact_id}&core=gav&rows=500&wt=json" | jq -r '.response.docs[] | select(.v | test("^[0-9]+(\\.[0-9]+)*$")) | .v' | grep ^"${major_version}". | sort -V | tail -n 1)
+    latest=$(curl -s "https://search.maven.org/solrsearch/select?q=g:${group_id}+AND+a:${artifact_id}&core=gav&rows=500&wt=json" | jq -r '.response.docs[] | select(.v | test("^[0-9]+(\\.[0-9]+)*$")) | .v' | sort -V | tail -n 1)
     echo "${latest}"
 }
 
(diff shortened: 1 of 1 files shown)


.github/scripts/update_generation_config.sh
```diff
@@ -52,7 +76,7 @@ if [ -z "${generation_config}" ]; then
 fi
 
 current_branch="generate-libraries-${base_branch}"
-title="chore: update googleapis commit at $(date)"
+title="chore: update generation configuration at $(date)"
```
blakeli0, line 79:
```suggestion
title="chore: Update generation configuration at $(date)"
```

.github/scripts/update_generation_config.sh
```diff
@@ -1,12 +1,36 @@
...
+    local key_word=$1
+    major=$(grep "${key_word}" "${generation_config}" | cut -d ':' -f 2 | xargs | cut -d '.' -f 1)
+    echo "${major}"
+}
+
+function get_latest_release_version() {
+    local group_id=$1
+    local artifact_id=$2
+    local major_version=$3
+    latest=$(curl -s "https://search.maven.org/solrsearch/select?q=g:${group_id}+AND+a:${artifact_id}&core=gav&rows=500&wt=json" | jq -r '.response.docs[] | select(.v | test("^[0-9]+(\\.[0-9]+)*$")) | .v' | grep ^"${major_version}". | sort -V | tail -n 1)
```
blakeli0, line 23: Why do we need a `major_version` to get the latest released version?

.github/scripts/update_generation_config.sh
```diff
@@ -1,12 +1,36 @@
...
+# 3. jq
+
+# Utility functions
+function get_major_version() {
+    local key_word=$1
+    major=$(grep "${key_word}" "${generation_config}" | cut -d ':' -f 2 | xargs | cut -d '.' -f 1)
+    echo "${major}"
+}
+
+function get_latest_release_version() {
```
diegomarquezp, line 19: Should this be "get_latest_released_version"?
Also, could you add a quick comment on these helper functions to explain what they do?

.github/scripts/update_generation_config.sh
JoeWang1127: We don't want to introduce major version bump to the generator in a branch.

.github/scripts/update_generation_config.sh
```diff
@@ -1,12 +1,36 @@
 #!/bin/bash
 set -e
 # This script should be run at the root of the repository.
-# This script is used to update googleapis commit to latest in generation
+# This script is used to update parameters in generation
```
diegomarquezp, line 4: nit: maybe we can expand this explanation to mention which parameters are affected

.github/scripts/update_generation_config.sh
JoeWang1127: Done.

.github/scripts/update_generation_config.sh
blakeli0: I'm not sure that's something should be guarded in google-cloud-java. Major version bump of the generator should be rare and it should be guarded in the sdk-platform-java. Assuming there is indeed a major version bump, the automation would not work and we have to manually update the generator version.

blakeli0: Same thing for libraries bom version, it is only used for generating README, and I don't think there is a need of guarding libraries bom.

JoeWang1127: This is used in supporting multi-version of gapic generator.

For example, we have two version of gapic generator, v2 and v3. In google-cloud-java, one branch (`branch-a`) is using v2 and another one is using v3. I think we don't want to update the generator to v3 in `branch-a` in this case.

.github/scripts/update_generation_config.sh
JoeWang1127: Done.

.github/scripts/update_generation_config.sh
JoeWang1127: Done.

.github/scripts/update_generation_config.sh
blakeli0: I see your point now. I think it is a good consideration and could work well for multi-JDK version scenario, but we probably don't want to guard it this way at this moment because there are other scenarios of multi-version support as well 
1. We could have multiple branches using the same major version but different minor version of gapic generator.
2. We could have multiple branches using the exact same version of gapic generator but different protoc version.
3. We may not want any automated version updates in a branch(LTS).

I think for now we can always update to the latest, and adjust our scripts as needed in the future.

JoeWang1127: OK, I'll remove the major version limitation here. We can adjust the script when new use case arises.

JoeWang1127: Done.

blakeli0 (approved):

PR #10964, opened 2024-06-13
chore: Move the logic of getting image_tag to after generation_config is populated.

blakeli0:
Move the logic of getting `image_tag` to after `generation_config` is populated. Otherwise `generation_config` could be empty when we try to get `image_tag` from it.


JoeWang1127:
Can you paste the issue in the pr description?

JoeWang1127 (approved):

PR #11114, opened 2024-09-03
chore: update hermetic build workflow

JoeWang1127:
In this PR:
- Secure generation workflow (use environment variable to avoid script injection), inspired by https://github.com/googleapis/java-bigtable/pull/2317
- Generalize script so that it can run both in google-cloud-java and handwritten libraries.

This workflow will serve as a source of true for generation workflows in handwritten libraries.

### DIFF
--- a/.github/scripts/hermetic_library_generation.sh
+++ b/.github/scripts/hermetic_library_generation.sh
@@ -17,14 +17,11 @@ set -e
 # 1. git
 # 2. gh
 # 3. docker
-# 4. mvn
 
 # The parameters of this script is:
 # 1. target_branch, the branch into which the pull request is merged.
 # 2. current_branch, the branch with which the pull request is associated.
-# 3. [optional] image_tag, the tag of gcr.io/cloud-devrel-public-resources/java-library-generation.
-# The value will be parsed from the generation configuration if not specified.
-# 4. [optional] generation_config, the path to the generation configuration,
+# 3. [optional] generation_config, the path to the generation configuration,
 # the default value is generation_config.yaml in the repository root.
 while [[ $# -gt 0 ]]; do
 key="$1"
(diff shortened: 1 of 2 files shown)


diegomarquezp (approved):
LGTM, thanks!

.github/scripts/hermetic_library_generation.sh
```diff
@@ -54,22 +47,18 @@ shift
...
 
 if [ -z "${current_branch}" ]; then
-  echo "missing required argument --current_branch"
+  echo "Missing required argument --current_branch"
   exit 1
 fi
 
 if [ -z "${generation_config}" ]; then
   generation_config=generation_config.yaml
-  echo "Use default generation config: ${generation_config}"
```
blakeli0, line 68: Can we keep the image tag as a parameter? In long term, we do want to pass it in and remove the generator version from the generation_config.yaml.

.github/scripts/hermetic_library_generation.sh
```diff
@@ -101,8 +94,15 @@ docker run \
   gcr.io/cloud-devrel-public-resources/java-library-generation:"${image_tag}" \
   --baseline-generation-config-path="${workspace_name}/${baseline_generation_config}" \
   --current-generation-config-path="${workspace_name}/${generation_config}"
+
 # commit the change to the pull request.
-git add java-* pom.xml gapic-libraries-bom/pom.xml versions.txt
+if [[ $(basename "$(pwd)") == "google-cloud-java" ]]; then
```
blakeli0, line 99: Do we have to differentiate between google-cloud-java and java-* repos? Can we just do `git add .`?

.github/scripts/hermetic_library_generation.sh
JoeWang1127: Sorry I missed this comment. I'll create a follow up.

PR #11115, opened 2024-09-03
chore: do not distinguish monorepo or split repo when using `git add` in generation workflow

JoeWang1127:
In this PR:
- Restore `image_tag` parameter.
- Do not distinguish monorepo and split repo when using `git add`.

A follow up of #11114.


blakeli0 (approved):

blakeli0 (commented):
Can you please update the PR title to make it more specific?

PR #11148, opened 2024-09-16
feat!: Remove orphaned protoc generated files from bigquerydatatransfer, monitoring and speech.

blakeli0:
There are some protoc generated files are not updated in over 3 years, which made them not compatible with latest protobuf runtime anymore. Since the source proto of these files are already deleted, we should delete them from our client library as well.
See go/incompatible-protoc-files-java-sdk for more details.


java-bigquerydatatransfer/.OwlBot-hermetic.yaml
```diff
@@ -20,52 +20,7 @@ deep-remove-regex:
 - "/java-bigquerydatatransfer/samples/snippets/generated"
 
 deep-preserve-regex:
-- "/java-bigquerydatatransfer/google-.*/src/test/java/com/google/cloud/.*/v.*/it/IT.*Test.java"
```
lqiu96, line 23: Do we want to delete this IT?

versions.txt
```diff
@@ -481,10 +481,8 @@ grpc-google-cloud-shell-v1:2.48.0:2.49.0-SNAPSHOT
 proto-google-cloud-shell-v1:2.48.0:2.49.0-SNAPSHOT
 google-cloud-speech:4.44.0:4.45.0-SNAPSHOT
 grpc-google-cloud-speech-v1:4.44.0:4.45.0-SNAPSHOT
-grpc-google-cloud-speech-v1beta1:2.44.0:2.45.0-SNAPSHOT
 grpc-google-cloud-speech-v1p1beta1:2.44.0:2.45.0-SNAPSHOT
 proto-google-cloud-speech-v1:4.44.0:4.45.0-SNAPSHOT
-proto-google-cloud-speech-v1beta1:2.44.0:2.45.0-SNAPSHOT
```
lqiu96, line 487: qq, why are these two removed?

blakeli0: For speech, the whole v1beta1 version does not exist anymore, see [googleapis](https://github.com/googleapis/googleapis/tree/master/google/cloud/speech)

java-bigquerydatatransfer/.OwlBot-hermetic.yaml
blakeli0: There is actually no integration tests for bigquerydatatransfer, so this is like a not-directly-related clean up. I can revert this as well if it is causing confusion.

versions.txt
lqiu96: I can't really tell from the diffs on Github, but I'm guessing the entire subfolder is removed in this PR?

java-bigquerydatatransfer/.OwlBot-hermetic.yaml
lqiu96: Ah ok. If there are no ITs to being with then it doesn't matter. I'm fine with removing

versions.txt
lqiu96: Actually I think still see the folders in the branch. Can we just remove the folders in google-cloud-java?

lqiu96 (approved):
LGTM

versions.txt
blakeli0: I think they should be removed already. Let's merge this in first, I'll create a follow up cleanup PR if the folders still exist.

JoeWang1127 (approved):

PR #11162, opened 2024-09-19
test: force protobuf 25.5 on v1.37.0

zhumin8:
for test only.


zhumin8:
kokoro integration failures on dataproc, compute and vertexai. Not in scope for lts.

PR #11166, opened 2024-09-20
[test]force protobuf 3.25.5 to test with 1.25.x.

zhumin8:
test for lts6 patch


zhumin8:
kokoro integration test failures in compute and vision (test fixed in #10567). these modules not in scope.

PR #11196, opened 2024-09-26
fix: revert "feat: [marketingplatformadminapi] new module for marketingplatformadminapi"

suztomo:
Reverts googleapis/google-cloud-java#11194

Preparing revert pull request just in case if we need it.

b/348679150.


lqiu96 (approved):

suztomo:
Checks are green. The affected files look good.

image

PR #11235, opened 2024-10-08
chore: change deep remove regex in ad-manager

JoeWang1127:
The deep remove regex is wrong in ad-manager:
Error: https://github.com/googleapis/google-cloud-java/pull/11229#issuecomment-2398459246


blakeli0 (approved):
Thanks Joe! Looks like we have `google` hardcoded in our [templates](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/templates/owlbot.yaml.monorepo.j2#L16-L20). As a follow up, can you please update the template as well? And audit if the same issue would happen for other non-cloud libraries?

PR #11294, opened 2024-11-08
fix: Propagate repo level dependency change sdk-platform-java-config to each module's release notes

blakeli0:
`google-cloud-shared-config` and `google-cloud-shared-dependencies` are consolidated into `sdk-platform-java-config`. The only repo level dependency change in google-cloud-java is `sdk-platform-java-config` now. 

This PR removes the outdated logic in release notes splitting that still looks for `google-cloud-shared-config` and `google-cloud-shared-dependencies`, add the logic to propagate `sdk-platform-java-config` change to each module's release notes.

This PR is dependent on the behavior that `sdk-platform-java-config` update is in the main release notes at least, which is not happening at this moment because it is a [chore](https://github.com/googleapis/google-cloud-java/pull/11188). There will be a follow up PR to update the `sdk-platform-java-config` dependency update to a `fix`.


JoeWang1127 (approved):
Thank you.

PR #11296, opened 2024-11-08
chore: Change the conventional commit type to fix for sdk-platform-java-config.

blakeli0:
This PR removes the grouping of `sdk-platform-java-config` and `gapic-generator-java`. Change the conventional commit type for `sdk-platform-java-config` to `fix`. 

In the future, maybe we should change all the deps update PR to `fix`, which is consistent with what we are doing in other repos. See relevant section in [sdk-platform-java ](https://github.com/googleapis/sdk-platform-java/blob/816b1608e52543b4b5ade40056c89740fe35a459/renovate.json#L87-L93)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -11,13 +11,13 @@
         "*"
       ],
       "enabled": false
-    },
+    }, 
     {
       "matchPackagePatterns": [
-        "^com.google.api:gapic-generator-java",
         "^com.google.cloud:sdk-platform-java-config"
       ],
-      "groupName": "SDK platform Java dependencies"
+      "semanticCommitType": "fix",
+      "semanticCommitScope": null
     }
   ],
   "includePaths": [


renovate.json
```diff
@@ -11,13 +11,13 @@
         "*"
       ],
       "enabled": false
-    },
+    }, 
     {
       "matchPackagePatterns": [
-        "^com.google.api:gapic-generator-java",
```
JoeWang1127, line 17: Why do we need to remove `com.google.api:gapic-generator-java` from this group?

blakeli0: Because the title of the grouped dependency update PR would not include versions. e.g. https://github.com/googleapis/google-cloud-java/pull/11271. In additional, even though they are from the same repo, `gapic-generator-java` and `sdk-platform-java-config` have different versions, so I think we should keep them separate.

suztomo (approved):
I want to see how gapic-generator-java removal affects the behavior.

blakeli0:
> I want to see how gapic-generator-java removal affects the behavior.

It would probably like this https://github.com/googleapis/google-cloud-java/pull/11121.

PR #11308, opened 2024-11-15
Fix: Update java-managedkafka .repo-metadata.json

acocuzzo:
Fixes product name.


noahdietz:
Hey @acocuzzo I don't work on this repo but was notified of the mismatched names via YAQS and was coming to do the same thing. There are a few spots that need updating, mostly in the various `pom.xml` files: https://github.com/search?q=repo%3Agoogleapis%2Fgoogle-cloud-java%20%22Google%20Apache%20Kafka%20for%20BigQuery%20API%22&type=code

Should those be done at the same time here?

acocuzzo:
Fixed old mentions

generation_config.yaml
```diff
@@ -1231,7 +1231,7 @@ libraries:
   - proto_path: google/cloud/managedidentities/v1beta1
 
 - api_shortname: managedkafka
-  name_pretty: Apache Kafka for BigQuery API
+  name_pretty: Managed Service for Apache Kafka API
```
suztomo, line 1234: I believe this is the source of other lines.

suztomo (approved):

generation_config.yaml
JoeWang1127: We shouldn't allow manual change in other files. I'll run the generation job for this library.

PR #11315, opened 2024-11-18
fix!: Remove grpc module for gke-connect-gateway

blakeli0:
`gke-connect-gateway` was recently changed to a [rest only](https://github.com/googleapis/google-cloud-java/blob/main/java-gke-connect-gateway/.repo-metadata.json#L8) library. Grpc transport was never supported in `gke-connect-gateway`, hence removing the generated grpc module in this PR.


blakeli0:
> I think we need to change `.OwlBot-hermetic.yaml` as well.

Thank! Updated.

JoeWang1127 (approved):

burkedavison (approved):

burkedavison:
Thank you for the fast attention on this.

PR #11481, opened 2025-03-11
feat!: [dataform] remove deprecated v1alpha2

zhumin8:
`v1alpha2` version of dataform was removed in  https://github.com/googleapis/googleapis/commit/c3640e234b86a4129234015461d22ef80b3e9ad0

This PR relects these changes in config yaml and delete the removed modules.

### DIFF
--- a/java-dataform/pom.xml
+++ b/java-dataform/pom.xml
@@ -41,24 +41,12 @@
         <artifactId>grpc-google-cloud-dataform-v1beta1</artifactId>
         <version>0.59.0-SNAPSHOT</version><!-- {x-version-update:grpc-google-cloud-dataform-v1beta1:current} -->
       </dependency>
-      <dependency>
-        <groupId>com.google.api.grpc</groupId>
-        <artifactId>grpc-google-cloud-dataform-v1alpha2</artifactId>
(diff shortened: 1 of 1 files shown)


generation_config.yaml
```diff
@@ -681,7 +681,6 @@ libraries:
   product_documentation: https://cloud.google.com/dataform/docs
   api_description: Help analytics teams manage data inside BigQuery using SQL.
   GAPICs:
-  - proto_path: google/cloud/dataform/v1alpha2
```
blakeli0, line 684: Thanks @zhumin8 for creating the fix! I thought our default behavior is that if the path does not exist, we would throw errors, @JoeWang1127 can you confirm it? If this is not expected, I think we should create an issue in the backlog to track it.

zhumin8: It does throw no such file error, if that's what you mean: https://github.com/googleapis/google-cloud-java/pull/11462#issuecomment-2715669430

blakeli0: OK, that makes sense then. It was by design that we don't want to automatically remove a deprecated module, because it is technically a breaking change.
Can you please change the title to `feat!` to indicate that it's a breaking change?

suztomo (approved):

suztomo:
This should unblock the generation check failure in https://github.com/googleapis/google-cloud-java/pull/11462. Thank you.

PR #11483, opened 2025-03-13
docs: change logging title

JoeWang1127:
In this PR:
- Rename Logging section to Advanced Logging Setup. A preparation step to introduce Client Library Debug Logging.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -12,7 +12,7 @@ including:
 - Verbose messages in underlying dependency libraries
 
 While there are various ways to configure the logging,
-to quickly enable verbose logging for the gRPC or HTTP transportation, create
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -1,6 +1,6 @@
 # Troubleshooting
 
-## Logging
+## Advanced Logging Setup
 
 Our libraries use the Java logging API via `java.util.logging` package.
```
blakeli0, line 5: I think a lot of the content needs to be updated along with the title. For example, `Our libraries use the Java logging API` is not valid anymore, we should say something like "our core dependencies and some handwritten libraries use the Java logging API".

JoeWang1127: Updated the contents, please take a look.

PR #11557, opened 2025-05-05
feat: Generate IAM v3 Library

lqiu96:
See b/406872289 for more information

### DIFF
--- a/java-iam/google-iam-policy/pom.xml
+++ b/java-iam/google-iam-policy/pom.xml
@@ -50,6 +50,10 @@
       <artifactId>proto-google-iam-v2</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-iam-v3</artifactId>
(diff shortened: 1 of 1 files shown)


lqiu96:
Turns out that that the proto/ grpc IAM versions are set by the gapic-generator-bom... 

For this release, I will just hardcode the IAM v3 proto and grpc versions so that I don't need to re-release sdk-platform-java. Next release we can remove this workaround have the bom manage the versions.

lqiu96:
> Turns out that that the proto/ grpc IAM versions are set by the gapic-generator-bom...
> 
> For this release, I will just hardcode the IAM v3 proto and grpc versions so that I don't need to re-release sdk-platform-java. Next release we can remove this workaround have the bom manage the versions.

Actually, do not need to hard-code the versions for this release as I had to re-release sdk-platform-java. IAM v3 has the versions set by the upstream gapic-generator-bom

jinseopkim0 (approved):

PR #11567, opened 2025-05-06
feat: [secretmanager] add method to extract secretId from name

### LINKED ISSUE #11566: [secretmanager] Get secretId from secret name
I'm submitting this feature request in the hope that the `Secret` class can be extended to include a method that I find is currently missing.

When I'm working with `Secret` objects, I often need to access the SecretId. Although I know I can extract it from the resource name or access it via the `SecretName`, I find myself repeatedly implementing the same small piece of code. This made me think that it would be more convenient if this method were simply built into the `Secret` class.

The `SecretName` class possesses a `getSecret()` method, whereas this method is absent from the `Secret` class. The existing `getName()` method returns the complete resource name (`/projects/*/secrets/*`). It would be beneficial to have a method that provides access to the SecretId itself. In line with the `SecretName.getSecret()` method, a possible name for this method could be `Secret.getSecret()`, although this might cause confusion. A more suitable option might be `Secret.getSecretId()`.

mieseprem:
Fixes #11566

This PR adds the method `getSecretId()` to `Secret` class.

First I must confess, that I didn't got the whole proto-thing. While the resource name is part of the SecretResponse, the secretId is not. So, I'm totally unsure if
* ... it is allowed to add a `getYXZ()` method without having a field with this name in the corresponding proto file.
* ... there is a established convention governing the order of members and methods

So, regarding the code: The idea is to parse the resource name only once on first access (while myself needs this information, others may not). I had doubts if thread safety is required. Using the double-checked locking pattern here may be a little to much. Anyway, better safe than sorry :wink:

### DIFF
--- a/java-secretmanager/google-cloud-secretmanager/src/test/java/com/google/cloud/secretmanager/v1/SecretManagerServiceClientTest.java
+++ b/java-secretmanager/google-cloud-secretmanager/src/test/java/com/google/cloud/secretmanager/v1/SecretManagerServiceClientTest.java
@@ -587,6 +587,31 @@ public class SecretManagerServiceClientTest {
   }
 
   @Test
+  public void getSecretTest3() throws Exception {
+    SecretName secretName = SecretName.ofProjectSecretName("[PROJECT]", "[SECRET]");
+    Secret secret =
+        Secret.newBuilder()
+            .setName(secretName.toString())
+            .setReplication(Replication.newBuilder().build())
+            .setCreateTime(Timestamp.newBuilder().build())
+            .putAllLabels(new HashMap<String, String>())
+            .addAllTopics(new ArrayList<Topic>())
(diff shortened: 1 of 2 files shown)


lqiu96:
Thanks for raising this @mieseprem. The `Secret.java` file is actually auto generated from the protoc compiler and I'm afraid your changes will probably be overwritten on the next generation.  The changes will have to come from upstream. 

Could you raise a feature request [here](https://cloud.google.com/support/docs/issue-trackers) so the secretmanager team would be able to evaluate and make the changes?

mieseprem:
> Could you raise a feature request [here](https://cloud.google.com/support/docs/issue-trackers) so the secretmanager team would be able to evaluate and make the changes

I guess this would mean to add the same information multiple times in the payload (which is really not my intention).

So be it :shrug:

PR #11597, opened 2025-05-29
docs(maps-places): add link to client library examples

meltsufin:
b/421118724

### DIFF
--- a/java-maps-places/.readme-partials.yaml
+++ b/java-maps-places/.readme-partials.yaml
@@ -1,4 +1,4 @@
 custom_content: |
   ### Client Library Usage Examples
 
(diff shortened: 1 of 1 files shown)


meltsufin:
> README is templated and fully-generated. We should either add this to the [generic template](https://github.com/googleapis/sdk-platform-java/blob/main/hermetic_build/library_generation/owlbot/templates/java_library/README.md) or custom `.readme-partials.yaml` template in each module. For example, `.readme-partials.yaml` in [java-compute](https://github.com/googleapis/google-cloud-java/blob/main/java-compute/.readme-partials.yaml).

Thanks! I replaced it with `.readme-partials.yaml`. PTAL.

PR #11625, opened 2025-06-17
chore: gapic-libraries-bom not to import other group IDs

suztomo:
The new Central Portal release API doesn't accept a release
bundle with multiple namespaces (group IDs) and the
gapic-libraries-bom imports the 3 BOMs from other group IDs than
com.google.cloud.

Let's cut the references by flattening the BOM imports.

b/424838882

### DIFF
--- a/gapic-libraries-bom/pom.xml
+++ b/gapic-libraries-bom/pom.xml
@@ -21,82 +21,27 @@
 
   <dependencyManagement>
     <dependencies>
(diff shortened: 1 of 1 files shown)


suztomo:
The change was overwritten by the automation. The template is in sdk-platform-java (./hermetic_build/library_generation/templates/gapic-libraries-bom.xml.j2). This idea doesn't work.

PR #11713, opened 2025-08-06
fix: add v1 to shopping-merchant-notifications

zhumin8:
Bring back the configuration for v1beta from [original commit](https://github.com/googleapis/google-cloud-java/pull/10748/files#diff-70450badb57c93e27292190b5470ead474f3b2071e682e3d8274ec45e9510410). It was accidentally deleted.

Add proto path for v1 and let automation generate the library.


zhumin8:
Will do 2 separate PRs, so review can be more focused.

PR #11723, opened 2025-08-12
feat: bump shopping-merchant sub-APIs to 1.0.0

zhumin8:
Bump shopping-merchant sub-APIs to 1.0.0, this includes 12 sub-APIs with recent v1 publications.

Does not include google-shopping-merchant-productstudio and google-shopping-merchant-reviews.


lqiu96:
LGTM. qq ,do we need to update the `current-version` value in `versions.txt`?

If we don't, would release-please revert this to the older version number on the next run?

zhumin8:
> LGTM. qq ,do we need to update the `current-version` value in `versions.txt`?
> 
> If we don't, would release-please revert this to the older version number on the next run?

Good point. Will update.

PR #11781, opened 2025-10-28
test: Update KMS ITs to use GAPICs instead of gRPC stub

lqiu96:
Discovered in https://github.com/googleapis/sdk-platform-java/pull/3942, it looks like KMS ITs are using gRPC generated stubs instead of the GAPIC stubs.

While this IT may not actually have that much use to simply test the GET/ CREATE/ LIST functionality of a service, we do use KMS in sdk-platform-java as a downstream test case.


blakeli0:
Can we use `test:` instead of `chore:`?

lqiu96:
> Can we use `test:` instead of `chore:`?

done

PR #11791, opened 2025-11-13
chore: Disable scorecard check

blakeli0:
Scorecard check [has been failing](https://github.com/googleapis/google-cloud-java/actions/runs/19339016837/job/55322042174) for a while. Disable it for now until we figure out what to do with it.


blakeli0:
> Do you have context on this test? If we still want this test, can we create an issue to track investigation and how to resolve and link it here?

I don't. Created https://github.com/googleapis/google-cloud-java/issues/11792 to track it.

PR #11805, opened 2025-12-03
chore: add configuration for protobuf-4.x-rc release

diegomarquezp:
Configuration in https://github.com/googleapis/google-cloud-java/blob/protobuf-4.x-rc/release-please-config.json


gapic-libraries-bom/pom.xml
blakeli0: Changes in this pom is unrelated?

diegomarquezp: Merging the generation PR fixed the issue.

PR #11837, opened 2025-12-17
feat: Upgrade protobuf gen code to 4.33

diegomarquezp:
BEGIN_COMMIT_OVERRIDE
BEGIN_NESTED_COMMIT
fix(deps): update the Java code generator (gapic-generator-java) to 2.65.0-rc1
END_NESTED_COMMIT
END_COMMIT_OVERRIDE


blakeli0:
It might be better to regenerate if there are too many conflicts.

diegomarquezp:
> It might be better to regenerate if there are too many conflicts.

Yes, I just rebased and force-pushed.

diegomarquezp:
Retrying integration test after terraform failure:

```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ module.java_iam_admin.google_service_account.service_account, provider
│ "provider[\"registry.terraform.io/hashicorp/google\"]" produced an
│ unexpected new value: Root object was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```

PR #11873, opened 2026-01-22
chore: update workflow github token

diegomarquezp:


diegomarquezp:
> The change does not look right

@blakeli0 I spoiled my find-replace. It should be OK now. Thanks for the catch.

PR #11874, opened 2026-01-22
Migrate java datastore

chingor13:


.github/CODEOWNERS
```diff
@@ -7,3 +7,4 @@
 
 # java-vertexai has maintainers
 /java-vertexai/   @googleapis/vertex-java-sdk @googleapis/cloud-java-team-teamsync
+/java-datastore/ @googleapis/cloud-native-db-dpes @googleapis/api-datastore-sdk @googleapis/api-firestore-partners @googleapis/cloud-java-team-teamsync
```
blakeli0: `cloud-java-team-teamsync` can be changed to Tomo's new alias `cloud-sdk-java-team` now.

java-datastore/google-cloud-datastore-utils/src/main/java/com/google/datastore/utils/testing/MockCredential.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2026 Google LLC
```
blakeli0: Why we decide to use current year `2026`? I think it's OK if we squash all the commits, but we probably want to keep the original year if we keep all the commit history.

PR #11902, opened 2026-02-03
chore: migrate java-logging into monorepo

chingor13:


jinseopkim0:
CIs appear to be failing, saying "Pull request exceeds maximum commit limit". Is there a good way to resolve this? Or is the plan to ignore this?

.github/CODEOWNERS
```diff
@@ -7,3 +7,4 @@
 
 # java-vertexai has maintainers
 /java-vertexai/   @googleapis/vertex-java-sdk @googleapis/cloud-sdk-java-team
+/java-logging/ @googleapis/cloud-sdk-java-team @googleapis/cloud-java-team-teamsync
```
blakeli0: `@googleapis/cloud-java-team-teamsync` can be removed now.

chingor13:
> CIs appear to be failing, saying "Pull request exceeds maximum commit limit". Is there a good way to resolve this? Or is the plan to ignore this?

For the CLA, we plan to manually override as these commits all came from a different repository where the CLA has been enforced. The maximum commit limit is unavoidable and a limitation of the check (managed by OSPO).

For the header-check, we plan to ignore, because we do want to preserve the original creation date of those files.

blakeli0:
I see that integration tests and GraalVM tests are run as part of `Kokoro - Test: Integration` and `Kokoro - Test: GraalVM Native Image`. Can they be a separate CI that is only triggered by changes to the java-logging folder?

chingor13:
> I see that integration tests and GraalVM tests are run as part of `Kokoro - Test: Integration` and `Kokoro - Test: GraalVM Native Image`. Can they be a separate CI that is only triggered by changes to the java-logging folder?

Both integration and graalvm tests already are conditional based on which folders are touched and are intentionally running IT tests.

We could separate them if we really want to, but it is additional work across internal configs and in-repo. We could also do this separately after monitoring if these tests are adversely affecting other packages (perhaps the next codegen that touches logging + other modules in the same PR).

jinseopkim0 (approved):

blakeli0:
> Both integration and graalvm tests already are conditional based on which folders are touched and are intentionally running IT tests.

I might be missing something, do you mind sharing where this is configured?

chingor13:
> > Both integration and graalvm tests already are conditional based on which folders are touched and are intentionally running IT tests.
> 
> I might be missing something, do you mind sharing where this is configured?

`generate_graalvm_modules_list` (defined in common.sh) builds a list of modules based on which ones have been modified in the PR.

https://github.com/googleapis/google-cloud-java/blob/21bfc98ba7bc56b6ca6dba3216e517818e24d8ee/.kokoro/build.sh#L100

https://github.com/googleapis/google-cloud-java/blob/21bfc98ba7bc56b6ca6dba3216e517818e24d8ee/.kokoro/common.sh#L265-L298

generation_config.yaml
```diff
@@ -637,21 +580,19 @@ libraries:
...
   product_documentation: https://cloud.google.com/compute/
-  api_description: "delivers virtual machines running in Google's innovative data
-    centers and worldwide fiber network. Compute Engine's tooling and workflow support
+  api_description: 'delivers virtual machines running in Google''s innovative data
+    centers and worldwide fiber network. Compute Engine''s tooling and workflow support
     enable scaling from single instances to global, load-balanced cloud computing.
-    Compute Engine's VMs boot quickly, come with persistent disk storage, deliver
-    consistent performance and are available in many configurations. "
+    Compute Engine''s VMs boot quickly, come with persistent disk storage, deliver
+    consistent performance and are available in many configurations. '
```
lqiu96: seems like there may be some added changes in this file. Are some of these intended? Looks like some added `\` and other characters.

chingor13: This was read/output by a python yaml parser. It should just be a formatting thing with regards to quoting/unquoting.

blakeli0:
@chingor13 I see, thanks! I think this is a good start and we can merge it for now. But I would prefer to have them as separate CIs in the future. The main benefit is that we can run them in parallel. In case we need to re-run a flaky tests, we don't have to re-run the whole suite. Even if there are no flaky tests, the overall running time can be significantly reduced as well.

chingor13:
> @chingor13 I see, thanks! I think this is a good start and we can merge it for now. But I would prefer to have them as separate CIs in the future. The main benefit is that we can run them in parallel. In case we need to re-run a flaky tests, we don't have to re-run the whole suite. Even if there are no flaky tests, the overall running time can be significantly reduced as well.

From our sync, I think we decided to move forward with this PR, then split the integration tests after merge. Then we can adjust our migration script to get to the final desired state.

lqiu96 (approved):

blakeli0 (approved):

PR #11905, opened 2026-02-05
chore: replace old java team with cloud-sdk-java-team

chingor13:


.github/CODEOWNERS
```diff
@@ -7,4 +7,4 @@
 
 # java-vertexai has maintainers
 /java-vertexai/   @googleapis/vertex-java-sdk @googleapis/cloud-sdk-java-team
-/java-logging/ @googleapis/cloud-sdk-java-team @googleapis/cloud-java-team-teamsync
+/java-logging/ @googleapis/cloud-sdk-java-team @googleapis/cloud-sdk-java-team
```
jinseopkim0: `@googleapis/cloud-sdk-java-team` is repeated. Was this intentional?

chingor13: not intentional -- we can also remove the entry all together since the root is also already owned by this same team

PR #11951, opened 2026-02-18
chore: truncate changelog.json to work with release please

ldetmer:
Currently the changelog.json file is too large to work with github trees api, so release please is unable to run due to a 422 error (file size too large).  Wiping changelog history.  This will have no affect on release-please and full history will be kept via git.


suztomo:
My approval is not enough.

PR #11960, opened 2026-02-20
test: skip unit tests for modules that have their own split tests

chingor13:
As the new split repos are migrating to the monorepo, we actually want to skip those unit tests when running the unit tests on the entire repo. This is because the unit tests for the migrated repos may be non-trivial and take a lot longer than the rest of the tests.

This is accomplished by adding a new profile called `bulkTests` which we set when running the "full" unit tests suite (no run on the split unit tests). When activated, the modules that migrated into the monorepo have that profile defined that will set the `skipTests` maven property (which skips those tests).


blakeli0:
I thought we agreed that unit tests are OK to be included because they should be relatively fast. Do you have examples of unit tests that are non-trivial and take a lot longer?

PR #12042, opened 2026-03-02
test(bqstorage): Add test for default keep alive settings in the Read Client

lqiu96:


jinseopkim0:
Some required CIs are failing. Could you please take a look?

PR #12044, opened 2026-03-02
chore: partial release of bigquerystorage as 1.80.1

lqiu96:
See b/424828238 for more info


lqiu96:
Ran with `mvn -B dependency:analyze -DfailOnWarning=true -Dcheckstyle.skip` locally to skip the checkstyle warning for BQStorage. It passes. 

I think it should be fine to merge and I'll resolve the issue in a separate PR.

PR #12104, opened 2026-03-17
feat: add response attribute tracking

ldetmer:


java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -1769,6 +1769,12 @@ private <T> T executeWithSpan(Span span, SpanOperation<T> operation) throws IOEx
     }
     try (Scope scope = span.makeCurrent()) {
       return operation.execute(span);
+    } catch (Exception e) {
```
ldetmer: sorry Jin Seop I know you requested this earlier, I just forgot about client side exceptions, so in the end I did need it!

PR #12107, opened 2026-03-18
chore(bq jdbc): fix docker environment commands

logachev:
- Update `make install` to work in monorepo
- Update dockerfile to handle monorepo. Due to its size, using repo root for context is taking too long (20+ minutes to just copy files to Docker image). Replaced it with `git clone`. It makes it not suitable to validate cross-library changes, but vast majority of JDBC work is within JDBC folder.


logachev:
gemini /review

PR #12113, opened 2026-03-18
feat: Add v1beta BigLake Hive client library

lqiu96:
See b/493631142 for more info


lqiu96:
Library should regenerate now that the generation config has been updated.

PR #12115, opened 2026-03-19
feat(bigquery): added error attributes to span tracing

ldetmer:
This adds capture of error attributes to BigQuery span tracing. 

[Example with GoogleJsonResponseException](https://screenshot.googleplex.com/54scDAtJYwhijST.png)
[Example with exception](https://screenshot.googleplex.com/7X787su6MSP8v3T.png)
see note [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/85b304b4916f2775a6f06b582fc866b24b8cea44/exporters/trace/README.md?plain=1#L125) about exception.type mapping to error/name

### DIFF
--- a/java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/telemetry/ErrorTypeUtil.java
+++ b/java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/telemetry/ErrorTypeUtil.java
@@ -60,7 +60,7 @@ class ErrorTypeUtil {
   static ErrorType getClientErrorType(Exception e) {
     if (isClientTimeout(e)) {
       return ErrorType.CLIENT_TIMEOUT;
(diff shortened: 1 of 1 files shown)


java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java
```diff
@@ -1769,6 +1771,16 @@ private <T> T executeWithSpan(Span span, SpanOperation<T> operation) throws IOEx
     }
     try (Scope scope = span.makeCurrent()) {
       return operation.execute(span);
+    } catch (Exception e) {
```
ldetmer: sorry Jin Seop I know you asked me to add this originally, but I didn't realize at the time that client side errors would only be available here, so you were correct this is the appropriate place to include error tracing!

PR #12124, opened 2026-03-20
feat(jdbc): basic OpenTelemetry tracing integration for BigQuery JDBC Statement

keshavdandeva:
b/491239772
b/491239773

### Changes

- New connection properties: `enableGcpTraceExporter` (Boolean, default: false) and `enableGcpLogExporter` (Boolean, default: false)

- `customOpenTelemetry` (Instance): Programmatic injection of a custom SDK (User Application-Managed setup) via `BigQueryDataSource.setCustomOpenTelemetry()`

- Added the core initialization logic for `OpenTelemetry`. During connection setup, it evaluates whether tracing is enabled and constructs an OpenTelemetry Tracer instance. Then, it passes this newly minted tracer strictly downward into the core `BigQueryOptions.Builder` via `.setOpenTelemetryTracer()`

- Intercepted the execution functions (`execute`, `executeQuery`, `executeLargeUpdate`, `executeBatch`) to spawn child spans wrapping each database call.


java-bigquery/google-cloud-bigquery-jdbc/pom.xml
```diff
@@ -133,6 +133,10 @@
                     <relocation>
                       <pattern>io</pattern>
                       <shadedPattern>com.google.bqjdbc.shaded.io</shadedPattern>
+                      <excludes>
+                        <exclude>io.opentelemetry.api.*</exclude>
```
logachev, line 144: Can you add comments on why it needs to remain unshaded? We saw Looker running into version mismatch, so apps using opentelemetry can hit it as well

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -939,6 +949,16 @@ private BigQuery getBigQueryConnection() {
       bigQueryOptions.setTransportOptions(this.httpTransportOptions);
     }
 
+    OpenTelemetry openTelemetry =
+        BigQueryJdbcOpenTelemetry.getOpenTelemetry(
+            this.enableGcpTraceExporter, this.enableGcpLogExporter, this.customOpenTelemetry);
+    if (this.enableGcpTraceExporter
+        || this.enableGcpLogExporter
+        || this.customOpenTelemetry != null) {
+      this.tracer = BigQueryJdbcOpenTelemetry.getTracer(openTelemetry);
```
logachev, line 956: Does tracer depend on `this.enableGcpLogExporter`?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -233,40 +239,49 @@ private BigQuerySettings generateBigQuerySettings() {
...
     LOG.finest("++enter++");
-    logQueryExecutionStart(sql);
-    try {
-      QueryJobConfiguration jobConfiguration =
-          setDestinationDatasetAndTableInJobConfig(getJobConfig(sql).build());
-      runQuery(sql, jobConfiguration);
-    } catch (InterruptedException ex) {
-      throw new BigQueryJdbcException(ex);
-    }
+    return withTracing(
```
logachev, line 244: I'd suggest using `executeQueryImpl` method for core logic instead of moving it to lambda. So that way public interface method will be tracer-wrapper on top of actual `executeQueryImpl`. 

Another idea is to keep methods in `BIgQueryStatement` class without traces & add new `BigQueryStatementWithTraces` class that will override methods we're interested in & use non-overrided version.

java-bigquery/google-cloud-bigquery-jdbc/pom.xml
keshavdandeva: Added the comment:
```
OpenTelemetry API and Context must remain unshaded to ensure interoperability.
Shading these would prevent the driver from participating in the application's 
existing tracing context. We are aware that unshaded dependencies can lead to 
version mismatches, but this is a necessary trade-off for the OpenTelemetry 
integration to function correctly across different applications.
```

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: Nope, good catch! Updated it

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: Hm yeah both approach sounds good but I guess will go with `executeQueryImpl` one so we dont add more factory complexity or maintenance overhead

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -1406,13 +1490,35 @@ public void clearBatch() {
...
+    Span span = tracer.spanBuilder("BigQueryStatement.executeBatch").startSpan();
+    try (Scope scope = span.makeCurrent()) {
+      span.setAttribute("db.statement.count", this.batchQueries.size());
+
+      StringBuilder sb = new StringBuilder();
+      for (String query : this.batchQueries) {
+        sb.append(query);
+      }
+      String combinedQueries = sb.toString();
+      span.setAttribute("db.statement", combinedQueries);
```
logachev: Should we define a separate attribute for a list of strings instead?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -1035,51 +1091,73 @@ Thread runNextPageTaskAsync(
...
-        () -> {
-          String currentPageToken = firstPageToken;
-          TableResult currentResults = result;
-          TableId destinationTable = null;
-          if (firstPageToken != null) {
-            destinationTable = getDestinationTable(jobId);
-          }
+        Context.current()
+            .wrap(
+                () -> {
```
logachev: At this point, should we move lambdas to functions instead?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -288,6 +322,21 @@ int checkUpdateCount(long updateCount) {
   @Override
   public boolean execute(String sql) throws SQLException {
     LOG.finest("++enter++");
+    Tracer tracer = getSafeTracer();
```
logachev: Can probably add  a generic method `private T traceMethod(String spanName, Runnable<T> )` and pass `Impl` method as a lambda

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: Yes, and I also realised I can store the batch queries as a structured string array attribute (`db.batch.statements`) instead of a concatenated string.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: Hm I am a bit confused because the previous version of PR had that exact pattern (using a `withTracing` helper that took a lambda), but I refactored it because I interpreted your previous comment:

> I'd suggest using executeQueryImpl method for core logic instead of moving it to lambda. So that way public interface method will be tracer-wrapper on top of actual executeQueryImpl

as a preference for avoiding lambdas in these execution methods.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -138,6 +140,11 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
   Long connectionPoolSize;
   Long listenerPoolSize;
   String partnerToken;
+  boolean enableGcpTraceExporter;
```
Neenu1995: nit: Don't we usually use Boolean class here?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -939,6 +949,14 @@ private BigQuery getBigQueryConnection() {
       bigQueryOptions.setTransportOptions(this.httpTransportOptions);
     }
 
+    OpenTelemetry openTelemetry =
+        BigQueryJdbcOpenTelemetry.getOpenTelemetry(
+            this.enableGcpTraceExporter, this.enableGcpLogExporter, this.customOpenTelemetry);
+    if (this.enableGcpTraceExporter || this.customOpenTelemetry != null) {
```
Neenu1995, line 955: Can these be null tested since they are primitive boolean and not Object Wrapper?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -939,6 +949,14 @@ private BigQuery getBigQueryConnection() {
...
+      this.tracer = BigQueryJdbcOpenTelemetry.getTracer(openTelemetry);
+      bigQueryOptions.setOpenTelemetryTracer(this.tracer);
```
Neenu1995, line 957: What about the other clients? Do they support OTel?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: Yeah, it is getting quite messy.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: Yes, we do. Good catch

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: `customOpenTelemetry` is an object instance (not a boolean)

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: Yes, the `BigQueryReadSettings` also supports it. I have added it.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: I moved these two to new functions

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
Neenu1995: There is a write client as well.

Neenu1995 (approved):
LGTM. But add the write client as well if it supports OTel.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: Yeah, I checked the source for `BigQueryWriteSettings` in the parent project, and unlike the read client settings, it does not expose a `setOpenTelemetryTracerProvider(...)` method. 
Also, I think its not needed, its network calls should be traced as children of our active spans via context propagation, once it is fully implemented

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: > Can probably add a generic method private T traceMethod(String spanName, Runnable<T> ) and pass Impl method as a lambda

Done

PR #12131, opened 2026-03-23
chore: migrate java-spanner-jdbc to monorepo

suztomo:
Migrate java-spanner-jdbc to monorepo.

b/477663564


suztomo:
I wasn't using correct branch for the migration script.

PR #12188, opened 2026-03-24
ci: Add Maven `quick-build` profile to skip maven lifecycle plugins

lqiu96:
Adding a maven profile that makes it easier to build modules faster locally. Added benefit is to make it easier to run faster builds in the CI.


lqiu96:
> How do you feel about naming the profiles based on their intent?
> 
> For example `installOnly` for when you want to do a module compile/install into your local maven repository. For this task, you would skip tests, clirr, enforcer, checkstyle etc.

Open to a different name. Intention was just to create one simple profile that could be run for whatever maven lifecycle and skip plugin executions. It wasn't aimed to be specific to a maven lifecycle (compile, test, or install) in particular.

> Another separate profile for `lint`, `native`, `integrationTests`, etc. Most of these would be setting similar options (e.g. `lint` would also skip Tests and clirr)

Probably could be added if it's needed, though I think it may be easier to just do something like `-Pfast -{plugin}.skip=false` to explicitly enable a specific plugin. Perhaps this can be added in a future PR if needed?

lqiu96:
How about something like `quickBuild` for the profile name to describe the intention?

lqiu96:
> I see warning in some [logs](https://github.com/googleapis/google-cloud-java/actions/runs/24052277905/job/70150592894?pr=12188)
> 
> ```
> Warning: 328 [WARNING] The requested profile "quickBuild" could not be activated because it does not exist.
> ```
> 
> Does this matter?

I renamed it from quickBuild to `quick-build`. Seems like I forgot a few references

blakeli0 (approved):

PR #12209, opened 2026-03-26
impl(o11y): introduce `url.full`

diegomarquezp:
- Adds HTTP_URL_FULL_ATTRIBUTE to ObservabilityAttributes
- Introduces requestUrlResolved into ApiTracer
- Implements redaction logic in ObservabilityUtils
- Passes tracer into HTTP/JSON transport to record the URL

Note: Following Footnote 3, the spec indicates logging implementations MAY expose a configuration to override redacted fields. This implementation does not expose this configurability to provide a consistent and optimally safe baseline out of the box.


diegomarquezp:
auth failures in SonarCloud Build

PR #12291, opened 2026-03-27
feat(bqjdbc): run getStatementType in parallel

logachev:
Core change in this PR:
Running `getStatementType` in parallel with a regular query. It saves ~150ms on small queries (where it takes e2e with reading results ~0.6s). Additionally, even if Job was created, we'll use `getStatementType` results (achieved with dry-run) for better performance. SDK doesn't have Job type after query is complete, it still has to query it from BQ (similar ~150ms).

Additional change:
Moving Executors to the Connection layer to be shared with all statements within a connection. Removed existing static Executor (from Statement object) due to:
- It was not used
- It was not cleaned up (and no meaningful way to call shutdown())


logachev:
This approach has issues when DROP TABLE is involved; Working on a different approach

PR #12310, opened 2026-03-30
feat(gax): Implement trace context extraction and injection interceptors

westarle:
See https://github.com/googleapis/google-cloud-java/compare/main...westarle:google-cloud-java:trace-propagation-integration-test for the end-to-end integration test.


westarle:
#12625 is adding tests as well

PR #12313, opened 2026-03-30
feat(bigquery): add resend attribute to span tracing + integration tests

ldetmer:
This adds the http.request.resend_count to span tracing in the case of a retry.  We achieve this by adding a context to the parent span that tracks all retries.  This remains at null for the initial retry and then gets incremented for all subsequent retries.

This PR also introduces integration tests to fully validate the resent attribute.

[example trace with resent attribute set ](https://screenshot.googleplex.com/BTjLU6v2stpYPWf.png)


java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/telemetry/HttpTracingRequestInitializer.java
```diff
@@ -84,6 +86,14 @@ public void initialize(HttpRequest request) throws IOException {
 
     addInitialHttpAttributesToSpan(span, request);
 
+    AtomicInteger attemptTracker = Context.current().get(BigQueryRetryHelper.RETRY_ATTEMPT_KEY);
+    if (attemptTracker != null) {
+      int attempt = attemptTracker.getAndIncrement();
+      if (attempt > 0) {
+        span.setAttribute(HTTP_REQUEST_RESEND_COUNT, (long) attempt);
```
blakeli0: This logic looks good to me. However, is there a way to implement it without using OpenTelemetry Context so it can be reused for other things? 

For example, metrics may also need this attribute in the future. We can call `Context.current()` as well if traces are enabled, but it is also possible that customers only enable metrics but not traces.

PR #12329, opened 2026-03-31
test(gax): Error Logging Integration Test

westarle:
Add an automated integration test (`ITActionableErrorsLogging`) to ensure end-to-end functionality of the Actionable Errors logging framework against a live`gapic-showcase` server.

### DIFF
--- a/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelEndpoint.java
+++ b/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/ChannelEndpoint.java
@@ -41,18 +41,26 @@ public interface ChannelEndpoint {
   String getAddress();
 
   /**
-   * Returns whether this server is ready to accept RPCs.
+   * Returns whether this server's channel is in {@code READY} state and can accept location-aware
+   * RPCs.
    *
-   * <p>A server is considered unhealthy if:
+   * <p>Only endpoints in {@code READY} state are eligible for location-aware routing. Endpoints in
+   * {@code IDLE}, {@code CONNECTING}, {@code TRANSIENT_FAILURE}, or {@code SHUTDOWN} are not
+   * considered healthy for location-aware routing purposes.
    *
-   * <ul>
-   *   <li>The underlying channel is shutdown or terminated
-   *   <li>The channel is in a transient failure state
-   * </ul>
-   *
-   * @return true if the server is healthy and ready to accept RPCs
+   * @return true if the channel is in READY state
    */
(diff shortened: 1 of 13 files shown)


sdk-platform-java/java-showcase/gapic-showcase/pom.xml
```diff
@@ -406,6 +419,51 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>envVarTest</id>
```
blakeli0: Can we reuse the existing profile [slf4j2_logback](https://github.com/googleapis/google-cloud-java/blob/c08d55316125faad1a61571203ea8383d5a6668e/sdk-platform-java/java-showcase/gapic-showcase/pom.xml#L258)?

There is an existing `ITLogging` that uses a [TestAppender](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/java-showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/TestAppender.java), which dose not require us to set up `slf4j-test-provider`.

That being said, I think the proposed approach in this PR also makes sense and I don't feel too strongly to reuse the existing pattern.

sdk-platform-java/java-showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITActionableErrorsLogging.java
```diff
@@ -0,0 +1,243 @@
...
+    EchoRequest request = EchoRequest.newBuilder().build();
+    assertThrows(ApiException.class, () -> mockHttpJsonClient.echo(request));
+
+    assertThat(testLogger.getMessageList().size()).isAtLeast(1);
+    String loggedMessage = testLogger.getMessageList().get(testLogger.getMessageList().size() - 1);
+
+    assertThat(loggedMessage).contains("This is a mock JSON error generated by the server");
+
+    Map<String, Object> kvps = testLogger.getKeyValuePairsMap();
+    assertThat(kvps).containsEntry("rpc.system.name", "http");
```
blakeli0: Can we reuse the static String in [ObservabilityAttributes](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/tracing/ObservabilityAttributes.java)?

sdk-platform-java/java-showcase/gapic-showcase/pom.xml
westarle: much better, thank you!

westarle:
fixed a real test failure

Issue #12328, opened 2026-03-31
SonarCloud Build is failing


blakeli0:
SonarCloud Build has been failing since the monorepo migration. See recent [failure logs](https://github.com/googleapis/google-cloud-java/actions/runs/23802827789/job/69367956326?pr=12321).

chingor13:
We are also missing the `SONAR_TOKEN` GHA secret

ldetmer:
google-cloud-java is using project key googleapis_gapic-generator-java which is associated with [this sonar project](https://sonarcloud.io/project/overview?id=googleapis_gapic-generator-java).  This is currently linked with cloud-sdk-platform not google-cloud-java.

I went to the admin settings to try to update the [repository binding](https://sonarcloud.io/project/settings?category=project_binding&id=googleapis_gapic-generator-java), however google-cloud-java is not an available repository.

When I went to [add it](https://sonarcloud.io/projects/create), I don't have access in the organization, only @meltsufin [is listed as org admin](https://sonarcloud.io/organizations/googleapis/members?view=ADMIN).  Mike can you please:

1) [add google-cloud-java as a project](https://sonarcloud.io/projects/create) we are tracking in sonar cube
2) either [update the repository binding for existing gapic-generator project](https://sonarcloud.io/project/settings?category=project_binding&id=googleapis_gapic-generator-java) to point to google-cloud-java OR update the github repo to point to a new sonar project/key for a new google-cloud-java sonar project

meltsufin:
I created the [google-cloud-java](https://sonarcloud.io/project/configuration?id=googleapis_google-cloud-java) Sonar project and gave @ldetmer and @blakeli0 admin access. Do you have what you need to complete the setup now?

meltsufin:
I updated the `SONAR_TOKEN` in google-cloud-java.
cc @jinseopkim0

blakeli0:
This is still failing in recent [PRs](https://github.com/googleapis/google-cloud-java/actions/runs/24403465976/job/71280351299?pr=12791).

meltsufin:
I see an unrelated issue in the log that causing the failure.
```
2026-04-14T14:28:47.6900903Z /home/runner/work/_temp/95b0da69-7862-4039-9c79-5ebee6177dc1.sh: line 2: cd: java-showcase: No such file or directory
```

blakeli0:
[...]

PR #12334, opened 2026-03-31
ci: fix paths for sonar checks

chingor13:
Fixes #12328


chingor13:
We need to adjust some configuration on the sonar cloud side

PR #12564, opened 2026-04-01
feat(gax): Add error attributes to golden signal metrics.

blakeli0:
This PR
- Added error attributes to metrics
- Added version as instrumentation version
- Added ITOtelGoldenMetrics that verifies all attributes


run_test.sh
diegomarquezp: same here

PR #12619, opened 2026-04-02
build: add option to force regenerate all

### LINKED ISSUE #12319: nightly codegen is not updating all libraries
We used to get new generator updates from the version bump of the gapic generator. Now, we only get updates for libraries whose config changed for generation_config.yaml

chingor13:
Adds an option to the hermetic codegen action for `force_regenerate_all`. We set this option if the code action is running against the `generate-libraries-main` branch.

So nightly:

1. the `.github/workflows/update_generation_config.yaml` updates the `generate-libraries-main` branch (usually with a new googleapis/googleapis SHA) 
2. the `.github/workflows/hermetic_library_generation.yaml` runs on this PR with the `force_regenerate_all` option enabled. This runs codegen for all libraries, not just ones that changed between the 2 googleapis/googleapis commits.

Sample run (used this branch's name instead of `generate-libraries-main`: https://github.com/googleapis/google-cloud-java/actions/runs/23887924105/job/69654601450

Fixes #12319

### DIFF
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -43,5 +43,5 @@ jobs:
         base_ref: ${{ github.base_ref }}
         head_ref: ${{ github.head_ref }}
         token: ${{ secrets.CLOUD_JAVA_BOT_GITHUB_TOKEN }}
-        force_regenerate_all: ${{ github.event.pull_request.head.ref == 'snapshot-generator' }}
+        force_regenerate_all: ${{ github.event.pull_request.head.ref == 'generate-libraries-main' }}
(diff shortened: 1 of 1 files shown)


chingor13:
> Thanks! I think this should work for now. In the future though, can we have a solution that works for branches as well? For example, we may need a generator fix for an LTS branch.

How we set the `force_regenerate_all` can definitely be adjusted. We don't want to set it all the time, because it does take > 1 hour to run over the entire codebase.

blakeli0:
> > Thanks! I think this should work for now. In the future though, can we have a solution that works for branches as well? For example, we may need a generator fix for an LTS branch.
> 
> How we set the `force_regenerate_all` can definitely be adjusted. We don't want to set it all the time, because it does take > 1 hour to run over the entire codebase.

Do you mean adding an allowlist of branches? That should work.

PR #12676, opened 2026-04-03
chore: Add style guide to gemini.md

blakeli0:
This PR adds common coding style guide to gemini.md.

### DIFF
--- a/sdk-platform-java/GEMINI.md
+++ b/sdk-platform-java/GEMINI.md
@@ -134,7 +134,7 @@ Showcase integration tests are run against a local server that implements the Sh
 
 ## 5. Style Guide
 
-1. Minimize scope; avoid `public` unless necessary.
+1. Minimize visibility scopes by defaulting to the most restrictive access level, avoiding the `public` modifier unless required.
 2. Use short names over fully qualified names.
 3. Avoid calling `@ObsoleteApi` or `@Deprecated` methods unless there are no alternatives.
 4. Avoid unnecessary formatting changes to keep diffs clean.


sdk-platform-java/GEMINI.md
```diff
@@ -131,14 +131,22 @@ Showcase integration tests are run against a local server that implements the Sh
...
-    
-## 5. Dependency Management
+
+## 5. Style Guide
+
+1. Minimize scope; avoid `public` unless necessary.
+2. Use short names over fully qualified names.
+3. Avoid calling `@ObsoleteApi` or `@Deprecated` methods unless there are no alternatives.
+4. Avoid unnecessary formatting changes to keep diffs clean.
+5. Use `mvn` for everything other than the `test/integration` folder.
```
lqiu96: not something we need to do in this PR, but I wonder if we can move some of these sections to a root GEMINI.md file. 

I think style guide and dependency management apply to all modules now that we have a monorepo

sdk-platform-java/GEMINI.md
```diff
@@ -131,14 +131,22 @@ Showcase integration tests are run against a local server that implements the Sh
     cd java-showcase
     mvn verify -P enable-integration-tests
     ```
-    
-## 5. Dependency Management
+
+## 5. Style Guide
+
+1. Minimize scope; avoid `public` unless necessary.
```
lqiu96: I think `minimize scope` wording might be a bit confusing/ unclear. Is this to minimize feature scope or scope visibility?

Also, maybe we can change this from `avoid` to `prefer`:
`Prefer private or package-private scope for code implementation`. or `Expose interfaces and hide implementation`

sdk-platform-java/GEMINI.md
```diff
@@ -131,14 +131,22 @@ Showcase integration tests are run against a local server that implements the Sh
...
+2. Use short names over fully qualified names.
+3. Avoid calling `@ObsoleteApi` or `@Deprecated` methods unless there are no alternatives.
+4. Avoid unnecessary formatting changes to keep diffs clean.
```
lqiu96: What does this line covering? Doesn't everything get run through the formatter?

PR #12681, opened 2026-04-05
fix(auth): Address ClientSideCredentialAccessBoundary RefreshTask race condition

vverman:
This change addresses a race condition in ClientSideCredentialAccessBoundaryFactory that occurred when multiple concurrent calls were made to generateToken.

The fix involves:
- Waiting on the RefreshTask itself rather than its internal task.
- Using a single listener in RefreshTask to ensure finishRefreshTask completes before the outer future unblocks waiting threads.
- Adding a regression test generateToken_freshInstance_concurrent_noNpe.

### DIFF
--- a/google-auth-library-java/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/google-auth-library-java/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -248,7 +248,7 @@ public class ClientSideCredentialAccessBoundaryFactory {
         }
         try {
           // Wait for the refresh task to complete.
-          currentRefreshTask.task.get();
+          currentRefreshTask.get();
         } catch (InterruptedException e) {
           // Restore the interrupted status and throw an exception.
           Thread.currentThread().interrupt();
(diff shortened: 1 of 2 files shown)


google-auth-library-java/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -495,31 +495,17 @@ class RefreshTask extends AbstractFuture<IntermediateCredentials> implements Run
...
-            }
-
-            @Override
-            public void onFailure(@Nullable Throwable t) {
-              RefreshTask.this.setException(
-                  t != null ? t : new IOException("Refresh failed with null Throwable."));
+              RefreshTask.this.setException(e.getCause());
+            } catch (Exception e) {
+              RefreshTask.this.setException(e);
             }
```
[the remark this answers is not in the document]
vverman: Addressed

macastelaz (approved):
LGTM!

google-auth-library-java/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -495,31 +493,18 @@ class RefreshTask extends AbstractFuture<IntermediateCredentials> implements Run
...
-            @Override
-            public void onSuccess(IntermediateCredentials result) {
-              RefreshTask.this.set(result);
-            }
-
-            @Override
-            public void onFailure(@Nullable Throwable t) {
-              RefreshTask.this.setException(
-                  t != null ? t : new IOException("Refresh failed with null Throwable."));
+              RefreshTask.this.setException(cause != null ? cause : e);
```
lqiu96, line 505: nit: can `ExecutionException` be nullable? From the original logic, it seems like this can't be the case. It seems like only the Throwable may be nullable, but the actual reason why seems to be lost on me.

vverman: ExecutionException isn't null here, it is the cause that could be null.

An ExecutionException is supposed to always have a cause. However, it is technically possible for an ExecutionException to be created without one (so e.getCause() returns null. In that case A NullPointerException will be thrown.

lqiu96 (approved):

lqiu96:
Thanks for the fix @vverman ! Changes LGTM

PR #12696, opened 2026-04-07
chore(bqjdbc): move extended test suite to monorepo

Neenu1995:


java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITBase.java
```diff
@@ -16,14 +16,141 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.junit.Assert;
 
 public class ITBase extends BigQueryJdbcBaseTest {
 
+  public static final String DEFAULT_CATALOG = ServiceOptions.getDefaultProjectId();
```
logachev: I'd suggest to keep only functions that can be re-used by different classes of tests in `ITBase` class. 

E.g. `connectionUrl`, `DEFAULT_CATALOG` 100% makes sense to keep there, but various `verify` methods that are very specific to a specific test (e.g. validate metadata) I'd suggest to keep with tests that leverage them.

e.g. `verifyAllBooleanMethods` is very specific (and it is just a single test of its own from what I understand) and I'd prefer to keep it separate

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITBase.java
Neenu1995: Done

PR #12718, opened 2026-04-08
feat(Datastore): Introduce Client Side Metrics

lqiu96:


java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/CompositeDatastoreMetricsRecorder.java
```diff
@@ -0,0 +1,93 @@
...
+   * own their {@link io.opentelemetry.api.OpenTelemetry} instance (built-in path) will flush and
+   * shut down; recorders backed by a user-provided instance will no-op. All recorders are closed
+   * even if one throws an exception.
+   */
+  @Override
+  public void close() {
+    for (DatastoreMetricsRecorder recorder : recorders) {
+      recorder.close();
+    }
+  }
```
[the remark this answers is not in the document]
lqiu96: Updated.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,218 @@
...
+      SdkMeterProviderBuilder sdkMeterProviderBuilder = SdkMeterProvider.builder();
+      // Register Datastore-specific views and the PeriodicMetricReader.
+      DatastoreBuiltInMetricsView.registerBuiltinMetrics(
+          DatastoreCloudMonitoringExporter.create(projectId, credentials), sdkMeterProviderBuilder);
+      // Configure the monitored resource attributes for this specific client.
+      sdkMeterProviderBuilder.setResource(
+          Resource.create(createResourceAttributes(projectId, databaseId)));
+      SdkMeterProvider sdkMeterProvider = sdkMeterProviderBuilder.build();
+      // Ensure cleanup on shutdown.
+      Runtime.getRuntime().addShutdownHook(new Thread(sdkMeterProvider::close));
```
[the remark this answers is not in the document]
lqiu96: Removed

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,218 @@
...
+   * @return a new {@link OpenTelemetry} instance, or {@code null} if it could not be created.
+   */
+  @Nullable
+  public OpenTelemetry createOpenTelemetry(
+      @Nonnull String projectId, @Nonnull String databaseId, @Nullable Credentials credentials) {
+    try {
+      SdkMeterProviderBuilder sdkMeterProviderBuilder = SdkMeterProvider.builder();
+      // Register Datastore-specific views and the PeriodicMetricReader.
+      DatastoreBuiltInMetricsView.registerBuiltinMetrics(
+          DatastoreCloudMonitoringExporter.create(projectId, credentials), sdkMeterProviderBuilder);
```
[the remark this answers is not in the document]
jinseopkim0: Is this relevant?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,213 @@
...
+   * pid@hostname}). The UUID prefix ensures uniqueness across process restarts that reuse the same
+   * PID, preventing Cloud Monitoring from conflating time series from different process lifecycles.
+   *
+   * <p>For Java 9 and later, the PID is obtained using the ProcessHandle API. For Java 8, the PID
+   * is extracted from ManagementFactory.getRuntimeMXBean().getName().
+   *
+   * @return a unique identifier string.
+   */
+  private static String getDefaultTaskValue() {
+    if (taskId == null) {
```
jinseopkim0: Would this be thread-safe? IIUC `static volatile` are not thread-safe, so there could be a race condition here. 

Also, given that taskId is a static variable, we'll have just 1 per JVM, so if multiple exporters are created, they will have the same ids. IIUC, this would be an issue with TimeSeries. I'm thinking we would need to generate a unique UUID for every client instance. WDYT?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreMetricsRecorder.java
```diff
@@ -49,25 +68,75 @@ public interface DatastoreMetricsRecorder extends MetricsRecorder {
...
+   * @return a {@link DatastoreMetricsRecorder} that fans out to all configured backends
    */
   static DatastoreMetricsRecorder getInstance(@Nonnull DatastoreOptions datastoreOptions) {
     DatastoreOpenTelemetryOptions otelOptions = datastoreOptions.getOpenTelemetryOptions();
+    List<DatastoreMetricsRecorder> recorders = new ArrayList<>();
 
+    // Default provider: export built-in metrics to Cloud Monitoring
+    String emulatorHost = System.getenv(DatastoreOptions.LOCAL_HOST_ENV_VAR);
+    boolean emulatorEnabled = emulatorHost != null && !emulatorHost.isEmpty();
+    String metricsEnvVar = System.getenv(TelemetryConstants.ENABLE_METRICS_ENV_VAR);
```
jinseopkim0: `ENABLE_METRICS_ENV_VAR` would be ignored given `isExportBuiltinMetricsToGoogleCloudMonitoring` is set to `false`, and I'm thinking this could be confusing if users set the environment variable to `true`. Perhaps naming the environment variable to something like `DATASTORE_DISABLE_BUILTIN_METRICS` would be better, so that it's more clear that it's just for turning it off but not necessarily turning it on. WDYT?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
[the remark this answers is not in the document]
lqiu96: I originally intended to have each client create it's own `MetricServiceClient` (tradeoff of simplicity vs memory). However, I think there is a good point that I can try and create a shared cache as I don't think it would be too much added complexity.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreMetricsRecorder.java
lqiu96: Ok gotcha. I think the confusion is a good point.

I actually don't think we really need an env var (I copied this from Spanner). I think we also can initially just use the setting `isExportBuiltinMetricsToGoogleCloudMonitoring` only to make this simpler.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: Ah ok, good catch. Let me take another look at this. We probably shouldn't make this static. I think the pid should be unique so that there can be multiple clients per JVM and shouldn't interfere.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Returns common client attributes added to every exported metric data point.
+   *
+   * <p>The returned map is pre-computed at construction time and shared across all export calls,
+   * since {@code client_name}, {@code client_uid}, and {@code service} are stable for the lifetime
+   * of the process.
+   *
+   * @return an unmodifiable map of client attributes.
+   */
+  Map<String, String> getClientAttributes() {
```
jinseopkim0: Does this get used anymore?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporter.java
```diff
@@ -0,0 +1,340 @@
...
+   * @param credentials the credentials used to authenticate with Cloud Monitoring.
+   * @return a new {@link DatastoreCloudMonitoringExporter} instance.
+   */
+  @Nullable
+  static DatastoreCloudMonitoringExporter create(
+      String projectId,
+      String databaseId,
+      Credentials credentials,
+      Map<String, String> clientAttributes) {
+    String key = projectId + ":" + databaseId;
```
jinseopkim0, line 137: Would it be possible for two clients to use the same project/database but use different `credentials`? If so, would it make sense to append something like `credentials.hashCode()` to the key?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,288 @@
...
+
+    return builder.build();
+  }
+
+  /** Maps OpenTelemetry metric type to Cloud Monitoring {@link MetricKind}. */
+  private static MetricKind convertMetricKind(MetricData metricData) {
+    switch (metricData.getType()) {
+      case HISTOGRAM:
+      case EXPONENTIAL_HISTOGRAM:
+        return convertHistogramType(metricData.getHistogramData());
```
jinseopkim0, line 182: Should this be using `getExponentialHistogramData()` instead? And in that case, `convertHistogramData` would not work. I'm thinking we don't need to support `EXPONENTIAL_HISTOGRAM`, since the exporter will never receive it.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,288 @@
...
+  private static Point createPoint(
+      MetricDataType type, PointData pointData, TimeInterval timeInterval) {
+    Point.Builder builder = Point.newBuilder().setInterval(timeInterval);
+    switch (type) {
+      case HISTOGRAM:
+      case EXPONENTIAL_HISTOGRAM:
+        return builder
+            .setValue(
+                TypedValue.newBuilder()
+                    .setDistributionValue(convertHistogramData((HistogramPointData) pointData))
```
jinseopkim0, line 252: Does it need to use `ExponentialHistogramPointData` instead of `HistogramPointData`? And in that case, `convertHistogramData` would not work. I'm thinking we don't need to support `EXPONENTIAL_HISTOGRAM`, since the exporter will never receive it.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,222 @@
...
+   * creation) to avoid performance overhead from UUID generation and hostname lookup.
+   *
+   * @return a unique identifier string.
+   */
+  private static String getDefaultTaskValue() {
+    String identifier = UUID.randomUUID().toString();
+    String pid = getProcessId();
+
+    try {
+      String hostname = InetAddress.getLocalHost().getHostName();
```
jinseopkim0: This has the potential to ake a long time, right? Would it make sense to have something like this?
```
private static final String PID_AND_HOSTNAME = getProcessId() + "@" + getHostnameSafely();

private static String getDefaultTaskValue() {
    return UUID.randomUUID().toString() + "@" + PID_AND_HOSTNAME;
}
```

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/TelemetryConstants.java
```diff
@@ -78,6 +105,7 @@ public class TelemetryConstants {
 
   // Metric attribute keys (used on metric data points)
   public static final AttributeKey<String> CLIENT_UID_KEY = AttributeKey.stringKey("client_uid");
+  public static final AttributeKey<String> CLIENT_NAME_KEY = AttributeKey.stringKey("client_name");
```
jinseopkim0: Is this unused?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: Removed

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
lqiu96: Good catch. I removed it

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
lqiu96: Removed

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: Good idea. Made it a static constant for each client.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/TelemetryConstants.java
lqiu96: Removed

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporter.java
lqiu96: Hmm, I think I went back and forth on this a few times and I'm leaning towards not hashing the credential. I think for multiple classes loaded together that share the same classpath, they can re-use the credential if they are targeting the same projectId + databaseId. Metrics (for now) will be explicit toggle on so users will need to supply a credential that is able to send metrics to Cloud Monitoring.

The original intention for having an isolated client per Datastore client was to avoid having to do filtering for metrics. For example, multiple Datastore clients share a Monitoring client singleton. If each Datastore client writes to a different project or database, then we would need to filter accordingly before we are able to send it as we don't want metrics from project A to be sent to project B. It was a tradeoff of increasing memory and the cache idea was to try and minimize/ optimize it a bit more. WDYT?

jinseopkim0: Thanks for the explanation, SGTM!

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporter.java
```diff
@@ -0,0 +1,343 @@
...
+   */
+  @Override
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
+  }
+
+  /** Shuts down the exporter and the underlying {@link MetricServiceClient}. */
+  @Override
+  public CompletableResultCode shutdown() {
+    if (client.isShutdown()) {
```
jinseopkim0, line 308: Would this be an issue since `client` is shared among multiple Datastore clients? E.g. there are 2 Datastore clients, and one of them calls shutdown twice. `client.isShutdown()` would be false both times.

If so, would it make sense to use something like this?
```
private final AtomicBoolean isShutDown = new AtomicBoolean(false);
```

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,285 @@
...
+
+  private DatastoreCloudMonitoringExporterUtils() {}
+
+  /**
+   * Extracts the project ID from the OpenTelemetry {@link Resource}.
+   *
+   * @param resource the OTel resource.
+   * @return the project ID string, or null if not present.
+   */
+  static String getProjectId(Resource resource) {
```
jinseopkim0, line 81: nit: is this unused?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporter.java
lqiu96: Updated!

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
lqiu96: Removed!

jinseopkim0 (approved):

java-datastore/samples/snippets/pom.xml
```diff
@@ -41,6 +41,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-datastore</artifactId>
+      <version>2.40.0</version>
```
blakeli0: Can this version be updated by release-please?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,202 @@
...
+  private static String getHostnameSafely() {
+    try {
+      return InetAddress.getLocalHost().getHostName();
+    } catch (UnknownHostException e) {
+      logger.log(Level.CONFIG, "Unable to get the hostname.", e);
+      return "localhost";
+    }
+  }
+
+  private static String getProcessId() {
```
blakeli0: Is getting a process id a requirement? Or we just need a unique identifier?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,202 @@
...
+    try {
+      // Check if Java 9+ and ProcessHandle class is available
+      Class<?> processHandleClass = Class.forName("java.lang.ProcessHandle");
```
blakeli0: If process id is required, can we always use a Java 8 compatible method instead of using reflections?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
```diff
@@ -0,0 +1,274 @@
...
+ * Utility class for converting OpenTelemetry metrics to Google Cloud Monitoring format.
+ *
+ * <p>This class contains the logic to map OpenTelemetry {@link MetricData} and {@link PointData} to
+ * Cloud Monitoring {@link TimeSeries}, including resource label mapping and attribute conversion.
+ *
+ * <p>The implementation in this file is inspired from the original work done in the Spanner client
+ * library (SpannerCloudMonitoringExporterUtils) to export metrics. The logic has been adapted for
+ * Datastore's use case.
+ */
+class DatastoreCloudMonitoringExporterUtils {
```
blakeli0, line 66: The converting logic seems pretty generic, what is unique to Datastore other than the resource type?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreMetricsRecorder.java
```diff
@@ -49,25 +66,73 @@ public interface DatastoreMetricsRecorder extends MetricsRecorder {
...
+          new OpenTelemetryDatastoreMetricsRecorder(customOtel, TelemetryConstants.METRIC_PREFIX));
     }
 
-    return new NoOpDatastoreMetricsRecorder();
+    // Default metrics are disabled and user has not custom Otel instance
+    if (recorders.isEmpty()) {
+      return new NoOpDatastoreMetricsRecorder();
+    }
+    // CompositeMetricsRecorder is not needed for one recorder
+    if (recorders.size() == 1) {
```
blakeli0: I think it's OK to use CompositeMetricsRecorder even if there is only 1 recorder.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/OpenTelemetryDatastoreMetricsRecorder.java
```diff
@@ -78,6 +101,22 @@ OpenTelemetry getOpenTelemetry() {
...
   }
 
+  /**
+   * Closes this recorder. If this recorder owns the underlying {@link OpenTelemetry} instance
+   * (i.e., it was created by the built-in metrics provider), it will be shut down, flushing any
+   * pending metrics. If the instance was provided by the user, this is a no-op.
+   */
+  @Override
+  public void close() {
+    if (isBuiltIn && openTelemetry instanceof OpenTelemetrySdk) {
```
blakeli0: Can we manage the lifecycle of the builtin OpenTelemetrySdk instance outside of the recorder? The creator of the OpenTelemetrySdk instance should be responsible for it.

java-datastore/samples/snippets/pom.xml
lqiu96: Yeah, will update

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: ProcessId itself is not required. Added to make it easier to identify when metrics are coming from different environments and different processes.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: From what I see online, the `ProcessHandle.current().pid` API is the best practice for Java 9+ as there are some weird edge cases with `ManagementFactory.getRuntimeMXBean().getName()`. Is there a concern with using reflection?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
lqiu96: There isn't really anything specific to datastore at all (I will just swapping the resource type and the attributes). The logic is mostly copied from Spanner.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreMetricsRecorder.java
lqiu96: There should be no real difference. I just opted to return the specific type

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/OpenTelemetryDatastoreMetricsRecorder.java
lqiu96: This should be closed only when the user calls `Datastore.close()`. Since we don't expose the built in otel object, users won't be able to access it otherwise.

I believe the Datastore client should be able to use this Otel instance for any other reason until the user initiates the shutdown.

blakeli0: I meant outside the recorder but still managed by us. For example, we can initialize `OpenTelemetry` object in `DatastoreImpl` and close it  in the same class. This also creates clear separation of concerns that recorder is only responsible for recording and always takes `OpenTelemetry` as an argument. Instead of responsible for both recording and managing `OpenTelemetry`.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreMetricsRecorder.java
blakeli0: If there are no difference, I think it's better to always use `CompositeDatastoreMetricsRecorder` for consistency.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/OpenTelemetryDatastoreMetricsRecorder.java
```diff
@@ -43,20 +46,40 @@
 class OpenTelemetryDatastoreMetricsRecorder extends OpenTelemetryMetricsRecorder
     implements DatastoreMetricsRecorder {
 
+  private static final Logger logger =
+      Logger.getLogger(OpenTelemetryDatastoreMetricsRecorder.class.getName());
+
   private final OpenTelemetry openTelemetry;
+  // True when this recorder created the OpenTelemetry instance (built-in path) and therefore
+  // owns its lifecycle. False when the instance was provided by the user.
+  private final boolean isBuiltIn;
```
blakeli0: If we manage the lifecycle of OpenTelemetry [outside of the recorder](https://github.com/googleapis/google-cloud-java/pull/12718/changes/BASE..05d20aa0cb7cb8da69f4ed613cdcb08efba38311#r3149538221), I don't think we need this flag anymore?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
blakeli0: We should probably refactor this to a common place then. But it can be done in a separate PR.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
blakeli0: Using reflection is general discouraged due to performance overhead, error-prone etc. We would also need to add GraalVm configs for any reflection usages. I would avoid it strongly if there are no hard requirement for it.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
blakeli0: How much easier would it be? I feel it is the same as just a UUID since  `getDefaultTaskValue` is always prefixed with a UUID?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporterUtils.java
lqiu96: Almost all of the exporting logic can be extracted to a common place (only real difference is in attributes, monitored resource, metric name) with some configurations toggleable in settings

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/OpenTelemetryDatastoreMetricsRecorder.java
lqiu96: Oh I see what you mean. I don't think it's impossible to do, but I believe I would just need to open up a few internal classes that I would prefer to keep package-private (BuiltInDatastoreMetricsProvider + DatastoreCloudMonitoringExporter + DatastoreBuiltInMetricsView). 

I think this would be easier if we have a stable, shared Gax utility for the creating all the necessary exporter + views since that would need to be public.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: > How much easier would it be? 

I think that `UUID@PID@hostname` *should* be relatively informative and easy to identify. The traceability is for the event of any specific process or host related metric issues.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: The performance should be minimal and a JDK native API should be safe from changes. Though, I'm less certain about GraalVM even if it would be a one-time addition. I'll use the Java 8 variant even if it isn't recommended.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: Hmm actually on second thought, it might just be easier if I just hash the taskID value. In the future, the client_uid is going to be aggregated away.

That way, I can try and create an artificial bound (no longer unbounded) for customers that are paying for metrics and I don't need to handle PID + hostname logic. Loses out on some traceability for metrics, but I think that tradeoff is OK.

lqiu96: Hashing the client_uid value instead

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: Hashing the client_uid value instead

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/OpenTelemetryDatastoreMetricsRecorder.java
lqiu96: Yes, that's true. Right now OpenTelemetryDatastoreMetricsRecorder is shared by both the builtin and user settings. 

Current tradeoff is that I can keep the builtin classes internal (package-private) and use a flag to manage the lifecycle. The separation could be cleaner, but I think it should be OK for this case. WDYT?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreMetricsRecorder.java
lqiu96: I do prefer using the specific type for clarity, but a benefit of being able to always return `CompositeDatastoreMetricsRecorder` is that I can get rid of the `NoOpDatastoreMetricsRecorder` class. 

I refactored this so that it always returns a `CompositeDatastoreMetricsRecorder`

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/OpenTelemetryDatastoreMetricsRecorder.java
blakeli0: I might be missing something, why we need to open up a few internal classes? I think we may need to open one public method that takes `OpenTelemetry` in `DatastoreMetricsRecorder`, but not sure we have to open more. 

In addition, I see that `DatastoreMetricsRecorder` is being cached in `DatastoreOptions` which is not a good pattern IMHO. The option/setting classes are supposed to hold static configuration values, creating a metric recorder based on configurations should be the responsibility of the caller class (in this case `DatastoreImpl`).

lqiu96: > I think we may need to open one public method that takes OpenTelemetry in DatastoreMetricsRecorder

Ah, I misinterpreted your earlier message. I thought by `initialize OpenTelemetry object in DatastoreImpl`, you meant to move the the initialization logic over to DatastoreImpl. I can split the creation of OtelSdk to be managed by DatastoreImpl. 

> In addition, I see that DatastoreMetricsRecorder is being cached in DatastoreOptions which is not a good pattern IMHO. The option/setting classes are supposed to hold static configuration values, creating a metric recorder based on configurations should be the responsibility of the caller class (in this case DatastoreImpl).

I'm not sure what you mean by cached in DatastoreOptions. Settings is a good point, I'll move that out to DatastoreImpl since that is a better spot to hold the recorder.

lqiu96: Updated. The built-in OtelSdk should be managed by DatastoreImpl and the recorder should be initialized there as well

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
```diff
@@ -162,13 +169,31 @@ public T call() throws DatastoreException {
...
+   */
   @Override
   public void close() throws Exception {
     try {
       datastoreRpc.close();
     } catch (Exception e) {
       logger.log(Level.WARNING, "Failed to close channels", e);
     }
+    // Shut down the built-in OTel SDK as we manage its lifecycle
+    if (builtInOpenTelemetry instanceof OpenTelemetrySdk) {
```
blakeli0, line 190: Is this instanceof check required because builtInOpenTelemetry is nullable? Can we return a `OpenTelemetry.noop()` instead when `BuiltInDatastoreMetricsProvider.createOpenTelemetry()` returns null?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -0,0 +1,191 @@
...
+    Preconditions.checkNotNull(credentials, "Credentials cannot be null for built in metrics");
+    SdkMeterProviderBuilder sdkMeterProviderBuilder = SdkMeterProvider.builder();
+
+    Map<String, String> clientAttributes = buildClientAttributes();
+
+    if (credentials instanceof NoCredentials) {
+      logger.log(
+          Level.WARNING,
+          "Built-in metrics exporting is disabled when using NoCredentials (emulator).");
+      return null;
```
blakeli0: Consider returning a `OpenTelemetry.noop()` instead when returning null.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporter.java
```diff
@@ -0,0 +1,346 @@
...
+   * ensures a full collection cycle completes before returning.
+   */
+  @Override
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
+  }
+
+  /** Shuts down the exporter and the underlying {@link MetricServiceClient}. */
+  @Override
+  public CompletableResultCode shutdown() {
```
blakeli0, line 310: Where is this shutdown method being called?

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreImpl.java
lqiu96: The instanceof check is needed because builtInOpenTelemetry is type OpenTelemetry. Only the implementation (OpenTelemetrySdk) implements AutoCloseable

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: Done! Updated

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporter.java
lqiu96: IIUC, called when the OpenTelemetry is closed. When testing this locally, it seems to be invoked via a shutdown hook from the Otel instance. Otel -> MeterProvider -> MetricReader -> Exporter

blakeli0: I see, so it will be called from `DatastoreImple#close` which should close the monitoring client as well. 

In general, I think it is a nice idea to share monitoring clients between datastore clients, but this is the first time we introduce shared states between clients, which may have other unexpected behaviors. It is possible to share `ChannelProviders` and `CredentialProviders` as well, but we don't manage the lifecycle of channels and credentials in those cases.

Since this is behind a client setting, I think it's fine for now but let's keep a close eye on it.

blakeli0 (approved):
Please update the title to `feat` and add descriptions to the PR body before merging.

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreCloudMonitoringExporter.java
lqiu96: > share monitoring clients between datastore clients, but this is the first time we introduce shared states between clients

Sounds good. I was aiming to try and reduce the memory a bit as it would end up creating 1 Otel instance + 1 Monitoring client per Datastore client. Not ideal as Spanner and BigTable were able to reduce memory by having a single Otel instance shared across clients. I made the tradeoff of isolated Otel instances to reduce filter complexity for metrics.

If this does introduce problems, I can try to isolate and create separate clients.

PR #12723, opened 2026-04-08
chore: Migrate auth commits

lqiu96:


lqiu96:
Hmm, wait if this is squash and merged, wouldn't this lose the individual commits? @chingor13 Is there a way to change this to rebase and merge (I think that should preserve the history)? Or is there a better way to port the history over?

diegomarquezp (approved):

PR #12729, opened 2026-04-09
chore(spanner): add LatencyTracker interface and default implementation

olavloite:
Adds an internal LatencyTracker interface and a default implementation that allows the client to track the latency of requests. This can be used for automatic replica selection and load balancing.


java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EwmaLatencyTracker.java
```diff
@@ -0,0 +1,85 @@
...
+  public EwmaLatencyTracker(double alpha) {
+    Preconditions.checkArgument(alpha > 0.0 && alpha <= 1.0, "alpha must be in (0, 1]");
+    this.alpha = alpha;
+  }
+
+  @Override
+  public double getScore() {
+    synchronized (lock) {
+      return score;
+    }
```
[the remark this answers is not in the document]
olavloite: Fixed

java-spanner/google-cloud-spanner/src/test/java/com/google/cloud/spanner/spi/v1/EwmaLatencyTrackerTest.java
```diff
@@ -0,0 +1,87 @@
...
+@RunWith(JUnit4.class)
+public class EwmaLatencyTrackerTest {
+
+  @Test
+  public void testInitialization() {
+    EwmaLatencyTracker tracker = new EwmaLatencyTracker();
+    tracker.update(100);
+    assertEquals(100.0, tracker.getScore(), 0.001);
+  }
+
```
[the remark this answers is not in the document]
olavloite: Fixed

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/LatencyTracker.java
```diff
@@ -0,0 +1,52 @@
...
+   * @return the latency score, where lower is better.
+   */
+  double getScore();
+
+  /**
+   * Updates the latency score with a new observation.
+   *
+   * @param latencyMillis the observed latency in milliseconds.
+   */
+  void update(long latencyMillis);
```
rahul2393: This flatten most “60us vs 500us vs 700us” differences into the same bucket. If we want this to drive bypass selection, the score needs to be at least micros, and ideally nanos or Duration.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EwmaLatencyTracker.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  public EwmaLatencyTracker(double alpha) {
+    Preconditions.checkArgument(alpha > 0.0 && alpha <= 1.0, "alpha must be in (0, 1]");
+    this.alpha = alpha;
+  }
+
+  @Override
+  public double getScore() {
+    synchronized (lock) {
+      return initialized ? score : Double.MAX_VALUE;
```
rahul2393, line 66: getScore() returns Double.MAX_VALUE until the tracker has seen traffic. That means a new endpoint, or one recreated after eviction, will always lose against any sampled endpoint that has historical data. In other words, it never gets traffic, so it never learns. We should be probing / low-rate exploration so a replica can “come back to the game”; this implementation bakes in starvation unless some separate mechanism guarantees exploration.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/LatencyTracker.java
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner.spi.v1;
+
+import com.google.api.core.BetaApi;
+import com.google.api.core.InternalApi;
+
+/**
+ * Interface for tracking latency scores of Spanner servers.
```
rahul2393, line 24: nit: The abstraction is attached to the wrong identity unless you are very careful in the follow-up work. The doc wants a score “for a given spanner server”, but this branch introduces a generic tracker with no ownership model. 

In the current routing code, CachedTablet instances are reused across cache updates and can change serverAddress in place. If someone later stores the EWMA on CachedTablet, the latency history from the old server will bleed into the new one after a cache update. The stable identity in this codebase  is the per-address endpoint cached, not the tablet object.

rahul2393 (approved):
Thanks, looks good to me overall with some questions open. I would like to see follow-up PRs soon for answers

Open Question: Eligibility filtering for stale read/query only, score ownership, score updates from successful/errorful routed calls, and the actual Po2 selection logic.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/LatencyTracker.java
olavloite: Good point, changed to `Duration`.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EwmaLatencyTracker.java
olavloite: Yeah, that is a good point. We will fix this in a follow-up PR in combination in the `ReplicaSelector` by allowing some of the traffic to just choose a random endpoint.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/LatencyTracker.java
olavloite: Ack

PR #12737, opened 2026-04-09
chore: untangling auth library parent and imported BOM

suztomo:
Maven Central does not allow uploading a set of artifacts having different group IDs.

- The shared dependencies BOM (which uses gapic-generator-java-bom) should not import the auth BOM.
- The google-auth-library-java/pom.xml should not use POM file of `com.google.cloud`.

### DIFF
--- a/google-auth-library-java/pom.xml
+++ b/google-auth-library-java/pom.xml
@@ -14,12 +14,10 @@
   <url>https://github.com/googleapis/google-cloud-java</url>
 
   <parent>
-    <!-- Because we cannot release com.google.auth and com.google.cloud together,
-      we skip the deployment of this module <maven.deploy.skip>true</maven.deploy.skip> -->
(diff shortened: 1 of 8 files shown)


google-auth-library-java/pom.xml
```diff
@@ -82,19 +80,73 @@
     <project.appengine.version>2.0.33</project.appengine.version>
     <project.findbugs.version>3.0.2</project.findbugs.version>
     <deploy.autorelease>false</deploy.autorelease>
-    <project.error-prone.version>2.38.0</project.error-prone.version>
+    <project.error-prone.version>2.42.0</project.error-prone.version>
```
blakeli0, line 83: Anything that requires this update?

suztomo: Guava depends on this newer version and upper-bound check was failing.

blakeli0 (approved):

PR #12747, opened 2026-04-10
chore(o11y): remove beta api from observability classes

blakeli0:
This PR remove `@BetaApi` from observability classes. `@InternalApi` annotations are also removed because they are either package private or need to be used externally.


sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryMetricsFactory.java
diegomarquezp, line 1: qq: should the recorder also be marked as non-beta?

https://github.com/googleapis/google-cloud-java/blob/a805ac7a883a6adc7d0d223e5b5cef81f81e73d7/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryMetricsRecorder.java#L54

diegomarquezp (approved):

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryMetricsFactory.java
blakeli0: The recorder was introduced three years and only used by built-in metrics. [GoldenSignalsMetricsRecorder](https://github.com/googleapis/google-cloud-java/blob/050187d934fc78139ec2790c04dd4c1e256591d4/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/tracing/GoldenSignalsMetricsRecorder.java#L49)  is the one used by this feature and is not `@BetaApi`.

PR #12765, opened 2026-04-11
ci: Add env vars for split units in sdk-platform-java

lqiu96:


blakeli0:
I think we may want to remove the split-units for sdk-platform-java since we already have a separate sdk-platform-java-ci.yaml

PR #12769, opened 2026-04-13
test(bigquery): Fix flaky ITBigQueryTest#testOpenTelemetryTracingQuery

jinseopkim0:
b/467063732


java-bigquery/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
```diff
@@ -7889,7 +7889,15 @@ public void testOpenTelemetryTracingQuery() throws InterruptedException {
     bigquery.getOptions().setDefaultJobCreationMode(JobCreationMode.JOB_CREATION_OPTIONAL);
     TableResult tableResult = executeSimpleQuery(bigquery);
     assertNotNull(tableResult.getQueryId());
-    assertNull(tableResult.getJobId());
+
+    // Safely handle the fallback where BigQuery decides to create a job anyway
```
lqiu96: nit: IIUC, BQ engine determines if a job is to be created even if we set the mode to be optional. Can you add a small comment that details that so it explains the `create a job anyways` part of the comment?

e.g. most requests will be stateless, but it is still possible that the BQ engine will create a job even for tiny requests

lqiu96 (approved):

java-bigquery/google-cloud-bigquery/src/test/java/com/google/cloud/bigquery/it/ITBigQueryTest.java
jinseopkim0: Thanks for the feedback, updated comment.

PR #12776, opened 2026-04-13
fix(gax): make FakePagedApi inner classes public to fix compilation error

diegomarquezp:
This PR fixes a compilation error in `PagingTest` where `FakePagedApi.ListIntegersSizedPage` was reported as having private access.

In `FakePagedApi.java`, `ListIntegersSizedPage` and `ListIntegersPage` were declared as `private static class`. However, they were used as type parameters in the public class `ListIntegersPagedResponse`. This exposure of private types via public type parameters caused compilation failures in stricter compiler environments (as seen in the CI failure).

## Changes

- Changed visibility of `ListIntegersPage` and `ListIntegersSizedPage` from `private` to `public` in `FakePagedApi.java`.

## Verification Results

- **Compilation**: Verified that `mvn test-compile -pl sdk-platform-java/gax-java/gax` succeeds.

### DIFF
--- a/sdk-platform-java/gax-java/gax/src/test/java/com/google/api/gax/rpc/testing/FakePagedApi.java
+++ b/sdk-platform-java/gax-java/gax/src/test/java/com/google/api/gax/rpc/testing/FakePagedApi.java
@@ -133,7 +133,7 @@ public class FakePagedApi {
     }
   }
 
-  private static class ListIntegersPage
+  public static class ListIntegersPage
       extends AbstractPage<Integer, List<Integer>, Integer, ListIntegersPage> {
 
     public ListIntegersPage(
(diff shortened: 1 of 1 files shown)


sdk-platform-java/gax-java/gax/src/test/java/com/google/api/gax/rpc/testing/FakePagedApi.java
```diff
@@ -133,7 +133,7 @@ public ListIntegersPagedResponse apply(ListIntegersPage input) {
     }
   }
 
-  private static class ListIntegersPage
+  public static class ListIntegersPage
```
blakeli0, line 136: I don't think we should make this change. This class has existed for a while and we didn't run into any issues. I think it is because the build infra messed up somewhere after monorepo migration.

diegomarquezp: SG, I'll backtrack and find the origin of this problem. Thanks for the quick review.

PR #12778, opened 2026-04-13
fix(bqjdbc): Revert DatabaseMetaData field to be non-static in BigQueryConnection

Neenu1995:


logachev (approved):
Should we also revert dynamic Statement within Metadata? 

Not sure what the right call is, but might be easier to track telemetry around metadata calls if statement is being reused in a future?

Neenu1995:
> Should we also revert dynamic Statement within Metadata?
> 
> Not sure what the right call is, but might be easier to track telemetry around metadata calls if statement is being reused in a future?

I feel like the current pattern is better in terms of thread safety(each call uses an isolated instance, no collision) and resource management(can be closed easily with closeOnCompletion). Also, considering metadata fetching is not a very frequent operation should be ok to have a per method statement instance.

PR #12785, opened 2026-04-13
ci: trigger split repo tests on auth, gax, and shared-deps changes

lqiu96:
Some "upstream" modules have a bit more impact and we want to test the impact of the changes across the modules inside the monorepo.

Changes in auth, gax, or shared-deps can trigger downstream unit tests


.github/workflows/ci.yaml
```diff
@@ -117,14 +117,44 @@ jobs:
...
-          java-datastore: java-datastore/**
-          java-logging-logback: java-logging-logback/**
-          java-logging: java-logging/**
-          java-spanner: java-spanner/**
-          java-storage: java-storage/**
-          sdk-platform-java: sdk-platform-java/**
+          java-bigquery:
+            - 'java-bigquery/**'
+            - 'google-auth-library-java/**/*.java'
+            - 'sdk-platform-java/gax-java/**/*.java'
```
blakeli0: Can we do `sdk-platform-java/**`? Changes in api-common and java-core may need to be verified as well. Same thing for other handwritten repos.

lqiu96: I added a few more cases. I set it to be a bit selective with the trigger (Java code changes or dep changes) to avoid our CIs running extra since our CIs currently take a bit

blakeli0 (approved):

PR #12797, opened 2026-04-14
chore: Add a Java development skill for the monorepo

lqiu96:
Adding a skill from the existing [sdk-platform-java GEMINI.md file](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/GEMINI.md). Try to break the Java development specific portions out so that it can be used by the rest of the modules in the monorepo.

### DIFF
--- a/.agents/skills/java-development/SKILL.md
+++ b/.agents/skills/java-development/SKILL.md
@@ -51,7 +51,7 @@ Code formatting is enforced using the `fmt-maven-plugin`.
 Follow these general rules to maintain code quality and consistency:
 
 1.  **Minimize Visibility**: Default to the most restrictive access level possible. Avoid using `public` unless the class or method is intended to be part of the public API.
-2.  **Short Names**: Prefer short names over fully qualified names when importing classes.
+2.  Avoid Fully Qualified Names: Use imports to keep class names short and readable, rather than using fully qualified names in the code.
 3.  **Avoid Obsolete APIs**: Do not call methods marked with `@ObsoleteApi` or `@Deprecated` unless there are no viable alternatives.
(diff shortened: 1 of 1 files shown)


.agents/skills/java-development/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+description: General guidance on Java development practices, building, testing, and style in the monorepo. Use this skill when working on Java code across the repository.
+---
+
+# Java Development Guide
+
+This skill provides general guidelines for Java development inside the monorepo. It covers building, formatting, testing, and style conventions to ensure consistency across modules.
+
+## Workflow
+
+### 1. Building the Project
```
blakeli0, line 12: I feel 1/2/3 are proper skills since they are telling the AI exactly how to do things. But 4/5/6 might be more suitable for a repo level gemini.md?

.agents/skills/java-development/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+
+The repository uses Maven as its primary build system.
+
+*   **Build All Modules**: To build all modules from the root of the repository, run:
+    ```bash
+    mvn install -T 1C -P quick-build
```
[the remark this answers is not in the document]
lqiu96: Use `install` to build to local ~/.m2 for snapshot artifacts

.agents/skills/java-development/SKILL.md
lqiu96: I'm not sure. I think it would be relevant as it tells the agent how to generate code to be used for the features. Shall we try a skill to see how it works?

blakeli0 (approved):

PR #12811, opened 2026-04-15
build: create additional tags on release

chingor13:
Towards #12724

Renamed the workflow and removed the path filter which doesn't apply to release events.

### DIFF
--- a/.github/workflows/create_additional_release_tag.yaml
+++ b/.github/workflows/create_additional_release_tag.yaml
@@ -4,9 +4,6 @@ on:
   release:
     types: [published]
   workflow_dispatch:
-
-env:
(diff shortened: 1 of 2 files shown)


meltsufin:
@blakeli0 Do we really need them still?

meltsufin (approved):

blakeli0:
I don't have the full context of this. It seems @suztomo reviewed the [original PR](https://github.com/googleapis/sdk-platform-java/pull/1692). 
From the description of the original issue though, it would be useful for troubleshooting. So if it's not too much toil, I think we should keep it.

blakeli0 (approved):

PR #12812, opened 2026-04-15
ci: fix unmanaged_dependency_check module install

chingor13:
Fixes #12704

### DIFF
--- a/.github/workflows/unmanaged_dependency_check.yaml
+++ b/.github/workflows/unmanaged_dependency_check.yaml
@@ -10,9 +10,13 @@ jobs:
       with:
         distribution: temurin
         java-version: 11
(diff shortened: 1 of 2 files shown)


meltsufin (commented):
Don't we already have it here: .github/workflows/versions.yaml?
There should probably just be one across the entire repo.

chingor13:
> Don't we already have it here: .github/workflows/versions.yaml? There should probably just be one across the entire repo.

versions.yaml check is trying to check that the release-please release PR doesn't have any -SNAPSHOT versions or if an existing release pom version already exists.

The unmanaged-dependency-check is for dependencyManagement versions that should be in shared-dependencies

blakeli0 (approved):

Issue #12736, opened 2026-04-09
[java-managed-identities] both v1 and v1beta in config but only v1 code present


zhumin8:
Both versions configured in generation_config.yaml, but only v1 is present in code.
https://github.com/googleapis/google-cloud-java/blob/55c985776700b1219ece39a519021030eb18d927/generation_config.yaml#L1541-L1543

https://github.com/googleapis/google-cloud-java/tree/main/java-managed-identities

zhumin8:
`java-os-login` also have same issue.

zhumin8:
This v1beta1 was added as part of the intial hermetic build onboarding [here](https://github.com/googleapis/google-cloud-java/commit/14b73467ea8e5037eb713610714a532fbba1a769).
owlbot config has `(v\\d)` instead of `(v.*)` for gapic-, grpc, and proto- modules.
https://github.com/googleapis/google-cloud-java/blob/8095342d164db5b0d2dc5141cf062986d8815458/java-managed-identities/.OwlBot-hermetic.yaml#L26-L34

 It is correct for samples though, so [samples](https://github.com/googleapis/google-cloud-java/tree/main/java-managed-identities/samples/snippets/generated/com/google/cloud/managedidentities) does contain both v1 and v1beta.

zhumin8:
I tried to raise PRs ( #12829) with changes to the .OwlBot-hermetic.yaml file, expecting hermetic build workflow to introduce code changes. Looks like the workflow succeeded, but no code changes. 

Ah, so the workflow detected no change in generation_config.yaml, thus not actually running the generator
https://screenshot.googleplex.com/5Koj7fGGJipF8nH

PR #12829, opened 2026-04-17
fix(java-managed-identities): fix copy rules to allow generated v1beta1

zhumin8:
Fix deep-copy-regex in java-managed-identities/.OwlBot-hermetic.yaml to allow generated code for v1beta1 to be copied to output destination.

v1beta1 of java-managed-identities was added to the generation_config.yaml as part of initial migration to hermetic build: https://github.com/googleapis/google-cloud-java/commit/14b73467ea8e5037eb713610714a532fbba1a769, but client library code for v1beta1 is not present in `java-managed-identities/google-cloud-managed-identities` dir. This is due to `deep-copy-regex` being too strict and does not copy /v1beta1 from staging dir to final output.

Expect generation workflow to add code for v1beta1.

For #12736


zhumin8:
close in favor of #12856

PR #12833, opened 2026-04-17
feat(bqjdbc): Per connection logs - Add BigQueryJdbcMdc

Neenu1995:
PR 1 of Per connection logging implementation.
- Implements BigQueryJdbcMdc using an InheritableThreadLocal design and corresponding unit tests for verification.


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -0,0 +1,97 @@
...
+  /** Clears the connection ID context from all active connection contexts on the current thread. */
+  public static void removeInstance(BigQueryConnection connection) {
+    if (connection != null) {
+      InheritableThreadLocal<String> local = instanceLocals.remove(connection);
+      if (local != null) {
+        local.remove();
+      }
+      instanceIds.remove(connection);
+    }
+  }
```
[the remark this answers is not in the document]
Neenu1995: removed in clear methods.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public static void clear() {
```
keshavdandeva, line 89: So, this method clears the global `instanceLocals` and `instanceIds` maps. I understand its use in the tests (to reset state) but if this method is used in production code in finally blocks after every operation (as suggested in the design doc), it will wipe the context for other concurrent connections, no?

maybe im misreading it but just wanted to confirm if this is intended.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Lightweight MDC implementation for the BigQuery JDBC driver using InheritableThreadLocal.
+ * Allocates a dedicated, independent InheritableThreadLocal object per concrete BigQueryConnection
+ * instance.
+ */
+public class BigQueryJdbcMdc {
+  private static final AtomicLong nextId = new AtomicLong(1);
```
keshavdandeva, line 28: Doubt: I think design doc says we use connection ID or UUID but here we are using incremental atomicLong? Am I confusing two different things?

Neenu1995: I am using this Atomic counter to generate a connection ID like JdbcConnection-N. Figured this pattern has better readability in logs.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
Neenu1995: Good catch. You're right, it gets cleared but then gets re-registered on a new method entry to preserve the intended behaviour, so I missed it. Fixed it now.

keshavdandeva (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -0,0 +1,95 @@
...
+
+      currentConnectionId.set(cleanId);
+      InheritableThreadLocal<String> threadLocal =
+          instanceLocals.computeIfAbsent(connection, k -> new InheritableThreadLocal<>());
+      threadLocal.set(cleanId);
+    }
+  }
+
+  /** Retrieves the connection ID mapped to a specific BigQueryConnection instance. */
+  public static String getConnectionId(BigQueryConnection connection) {
```
logachev: I still don't understand the purpose of that method.. If we have `BigQueryConnection` object, it should know its own connectionId. ConnectionId should be immutable across threads... So why using ThreadLocal to retrieve Id instead of `connection.getConnectionId()`?

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
Neenu1995: The connection object doesn't store the connectionID like JdbcConnection-N. It only generates the N part. The actual connectionID generation happens in Mdc class. But looking at the prototype code again, this method is not being used anywhere. I must have changed the usage for this method in a later iteration. I will remove it for now and add it back if the use case pops up again.

Issue #12735, opened 2026-04-09
[sdk-platform-java] generated modules are not in monorepo config after merge


zhumin8:
showcase, common-protos and iam from https://github.com/googleapis/sdk-platform-java/blob/main/generation_config.yaml are not in google-cloud-java/generation-config.yaml

chingor13:
Need to figure out how to move the protos into a non-root folder

suztomo:
> Need to figure out how to move the protos into a non-root folder

Joe @JoeWang1127 also wants to relocate the modules from sdk-platform-java to the root (https://github.com/googleapis/librarian/issues/5326).

chingor13:
Waiting for the current generation PR to regenerate everything. We expect to see some new classes in java-iam

chingor13:
The generation PR is generating IAM correctly now: https://github.com/googleapis/google-cloud-java/pull/12807/changes/0a5fe8e0a744784a026eb882576fcf967990a552

PR #12834, opened 2026-04-17
refactor: move iam-policy code to java-iam-policy folder

chingor13:
This folder contains the google-cloud-iam-policy client

Towards #12735
Towards https://github.com/googleapis/librarian/issues/5326

There is also an existing sdk-platform-java/java-iam folder which will need to move to the top level to be generated


suztomo (approved):

generation_config.yaml
```diff
@@ -1299,12 +1299,13 @@ libraries:
   - proto_path: google/cloud/hypercomputecluster/v1beta
   - proto_path: google/cloud/hypercomputecluster/v1
   requires_billing: true
-- api_shortname: iam
+- api_shortname: iam-policy
```
blakeli0, line 1302: Why we need to change the api_shortname?

blakeli0:
There are some assumptions in the hermetic build scripts regarding iam and common-protos, such as [here](https://github.com/googleapis/google-cloud-java/blob/64e7c9e0c766d6e838565982891e03710c8fb1df/sdk-platform-java/hermetic_build/common/model/gapic_inputs.py#L39) and [here](https://github.com/googleapis/google-cloud-java/blob/64e7c9e0c766d6e838565982891e03710c8fb1df/sdk-platform-java/hermetic_build/library_generation/owlbot/src/fix_poms.py#L187). Can we test locally to make sure the hermetic build scripts can generate IAM correctly?

PR #12838, opened 2026-04-17
feat(gax-grpc): add configurable resize delta and warning for repeated resizing

lqiu96:


sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
jinseopkim0, line 324: Would this mean that `getMaxChannelCount()` may be ignored, if `minChannels` is greater than `maxChannels`? It says in https://github.com/googleapis/google-cloud-java/blob/4e566af673e41af33663077e324ee949f20158da/sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java#L93: "The absolute maximum size of the channel pool." Would it make sense to set the max limit on `minChannels` as well? E.g.
```
if (minChannels > settings.getMaxChannelCount()) {
      minChannels = settings.getMaxChannelCount();
    }
```

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -69,6 +69,11 @@
  * <p>Package-private for internal use.
  */
 class ChannelPool extends ManagedChannel {
+  static final String CHANNEL_POOL_CONSECUTIVE_RESIZING_WARNING =
+      "Channel pool is repeatedly resizing. "
+          + "Consider adjusting `initialChannelCount` or `maxResizeDelta` to a more reasonable value. "
```
jinseopkim0, line 74: nit: would it make sense to also add max/minRpcsPerChannel in the message? E.g.
```
"Consider adjusting `initialChannelCount`, `maxResizeDelta`, `minRpcsPerChannel`, or `maxRpcsPerChannel` to a more reasonable value. "
```

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
jinseopkim0, line 201: Should this be the following?
```
Preconditions.checkState(
    s.getMinChannelCount() <= s.getMaxChannelCount(), "absolute channel range is invalid");
```

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -313,9 +329,25 @@ void resize() {
     int currentSize = localEntries.size();
     int delta = tentativeTarget - currentSize;
     int dampenedTarget = tentativeTarget;
-    if (Math.abs(delta) > ChannelPoolSettings.MAX_RESIZE_DELTA) {
-      dampenedTarget =
-          currentSize + (int) Math.copySign(ChannelPoolSettings.MAX_RESIZE_DELTA, delta);
+    if (Math.abs(delta) > settings.getMaxResizeDelta()) {
```
blakeli0, line 332: Do we know why there was a limit in the first place? Were there any technical limitations?

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
```diff
@@ -132,6 +139,9 @@ public static ChannelPoolSettings staticallySized(int size) {
         .setMaxRpcsPerChannel(Integer.MAX_VALUE)
         .setMinChannelCount(size)
         .setMaxChannelCount(size)
+        // Static pools don't resize so this value doesn't affect operation. However,
+        // validation still checks that resize delta doesn't exceed channel pool size.
+        .setMaxResizeDelta(Math.min(DEFAULT_MAX_RESIZE_DELTA, size))
```
blakeli0, line 153: I think we should still give it a max limit instead of being unbounded. Otherwise customers might expect this to handle 100x request spike as well.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: Yeah, I think we will need to enforce these bounds. This seems to be outside the scope of this PR. I will create a follow up PR for this

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: My plan is to change this to reference a public guide that I'm going to write (after metrics). WDYT if I leave this for now (for the immediate Datastore ask) and then update it to the public guide?

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
lqiu96: Yep, that should be fixed. I think I'll do this in a follow up PR

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: IIUC, it looks to just be a choice. Dampening and rate limit the channel growth to not overwhelm the client for a sudden burst.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
lqiu96: The delta is capped to between [1, MAX_CHANNEL_COUNT]. The javadocs already mention that the number of channels can never exceed the total number of channels configured (Default 200)

blakeli0: But is 200 a realistic number? I think it makes sense to allow customers configure resizeDelta to be more than 2, but more than 10 (We need to come up with an more acurate number) may introduce other performance issues. 

For example
- resize is in a [synchronized block](https://github.com/googleapis/google-cloud-java/blob/0b6a924544f7f23219e306f7b43fd5b686a8a560/sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java#L261-L263). 
- We use AtomicInteger to keep tracking the number of outstanding RPCs, which could cause issues in high contention scenarios. In which case [LongAdder](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/util/concurrent/atomic/LongAdder.html) may be preferred. 

Can we reach out to the gRPC team and get some suggestions?

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: https://github.com/googleapis/google-cloud-java/pull/12863

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
lqiu96: https://github.com/googleapis/google-cloud-java/pull/12863

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
lqiu96: This PR does not intend to fix channel pool's issues. It only exposing a setting to allow users to configure a value that they want that fits within the current bounds of the existing logic. The default resize delta remains 2 and if they choose a different value, then they can test it and figure if it works for their workload.

There are limitations to the existing channel pool logic that can use some broader changes. There is a project proposal to investigate (b/503856499) how to make this better overall. 

> But is 200 a realistic number? I think it makes sense to allow customers configure resizeDelta to be more than 2, but more than 10 (We need to come up with an more acurate number) may introduce other performance issues.

I just aim provide a safe default and let customers tinker with that they think best. Delta of 10 or 25 may work better for different workloads. 

If we want to fix ChannelPooling, then I think other changes would be more beneficial than investigating the resize delta value.

blakeli0: I agree that this PR is not to fix ChannelPooling. However, the new setter could make it easier for customers to exploit the current limitations of ChannelPooling, hence I think it would still be better to set an upper limit.

lqiu96: > exploit the current limitations of ChannelPooling

I'm not sure what you mean by exploiting here. If they configure a high value where performance degrades and doesn't work for them, they can rollback this change. If a high delta works for their use case, then they can opt to keep it.

---
Regardless, I think we may have to agree to disagree on setting a hard bound here. I don't think we should set an arbitrary bound for user configuration regardless of current limitations, barring something that doesn't fit the logic like negative resize delta. 

IMO, if it has drastic performance concerns, it would be beneficial to see user workload configurations as well as the issue reports. It gives us signal about their requirements and helps us see what would be needed for a future ChannelPool overhaul (e.g. channel priming, power-of-two, etc). And the channelpool issues give us more direct data point to prioritize the project (instead of slightly tangential reports of increased client-side request latency). I'm worried that limiting this "exploit" hides the need to fix the underlying channelpool issues.

Rather than having to talk to gRPC and having to investigate the possible default upper bound limits, how we compromise and I set this to a higher default upper bound value (e.g. 25?). I'll add javadocs about *potential performance concerns* for setting a higher delta?

lqiu96: Updated to clamp to 25 max for the resize delta with warning about performance

sdk-platform-java/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/ChannelPoolTest.java
```diff
@@ -523,6 +523,53 @@ void channelCountShouldNotChangeWhenOutstandingRpcsAreWithinLimits() throws Exce
     assertThat(pool.entries.get()).hasSize(2);
   }
 
+  @Test
+  void customResizeDeltaIsRespected() throws Exception {
+    ScheduledExecutorService executor = Mockito.mock(ScheduledExecutorService.class);
+    FixedExecutorProvider provider = FixedExecutorProvider.create(executor);
+
+    List<ManagedChannel> channels = new ArrayList<>();
```
jinseopkim0, line 532: Are these unused? There is a SonarCloud warning below.

jinseopkim0: Could you please take a look at other similar errors in the PR?

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -84,6 +89,16 @@ class ChannelPool extends ManagedChannel {
...
   private final String authority;
 
+  // The number of consecutive resize cycles to wait before logging a warning about repeated
+  // resizing. This value was chosen to detect repeated requests for changes (multiple continuous
+  // increase or decrease attempts) without being too sensitive.
+  private static final int CONSECUTIVE_RESIZE_THRESHOLD = 5;
+
+  // Tracks the number of consecutive resize cycles where a resize actually occurred (either expand
+  // or shrink). Used to detect repeated resizing activity and log a warning.
+  // Note: This field is only accessed safely within resizeSafely() and does not need to be atomic.
```
jinseopkim0, line 99: nit: should it say `resize()` instead of `resizeSafely()`, since the field is accessed there?

lqiu96: I've updated this to AtomicInteger in the followup PR. I'll address the comment in that one.

sdk-platform-java/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/ChannelPoolTest.java
lqiu96: Updated

jinseopkim0 (approved):

Issue #12702, opened 2026-04-07
[sdk-platform-java] Make the CI of sdk-platform-java required


blakeli0:
The CI of [sdk-platform-java](https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/sdk-platform-java-ci.yaml) is currently not required. This could cause mistakes of merging in changes with failing tests. 

We should make this required and always pass if there are no changes to `sdk-platform-java` folder.

chingor13:
We already conditionally run these tests - we will mark them as required checks after we ensure they are not broken

blakeli0:
Can we have a tracking list? Also can kokoro integration tests be conditionally required too?

chingor13:
> Can we have a tracking list? 

updated the issue with the list

> Also can kokoro integration tests be conditionally required too?

These are conditionally skipped by Kokoro already. We should mark them also as required

chingor13:
For making conditionally run matrix jobs required, we can use a final job that "needs" other jobs in a workflow to run. It can check the status of those jobs and return failure.

Example: https://github.com/chingor13/release-please-playground/pull/72/changes#diff-e178e3e7f9ded80fcbd00bccca0fb7ebfafcb78777a50280feea5e2006788cc4R40-R49

blakeli0:
Adding a datapoint that https://github.com/googleapis/google-cloud-java/pull/12838 introduced a few failing tests on main and we had to fix them in https://github.com/googleapis/google-cloud-java/pull/12882 and https://github.com/googleapis/google-cloud-java/pull/12886

chingor13:
`conditional-required-check` is now a required check on the repository

#12839 adds the `conditional-required-check` for the showcase tests

blakeli0:
I'm testing the fix in [this PR](https://github.com/googleapis/google-cloud-java/pull/12846). I made changes to auth, sdk-platform-java and most migrated repos but didn't see expected checks to be required. cc: @lqiu96 

Ideally:
- Auth triggers the following required checks: 
  - sdk-platform-java
  - showcase
  - split-units
  - integration tests (nice to have) 
- sdk-platform-java 
  - showcase
  - split-units
  - integration tests (nice to have) 
- Each individual handwritten library trigger their own required CIs.

chingor13:
[...]

PR #12839, opened 2026-04-17
refactor: move java-common-protos, java-iam, java-showcase to top level

chingor13:
Currently, the hermetic code generator can only move files to top level folders. Rather than changing the generator, we will move the generated code to the top level to also make it compatible with librarian.

Other fixes:
* The hermetic_build scripts had special case logic to force regeneration of the entire repository if the common protos were found in the generation_config.yaml. This is no longer necessary in google-cloud-java as the nightly build regenerates everything
* Includes the OTel wait for metrics fix similar to #12880
* Removes configuration for GCS Maven Mirror as that mirror is no longer updated and is missing dependencies 

Towards #12735
Towards #12702


generation_config.yaml
```diff
@@ -3014,4 +3015,26 @@ libraries:
   GAPICs:
   - proto_path: google/cloud/workstations/v1
   - proto_path: google/cloud/workstations/v1beta
-
+- api_shortname: common-protos
```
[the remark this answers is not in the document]
chingor13: This is a good idea, but it is not sorted this way currently

generation/check_non_release_please_versions.sh
```diff
@@ -13,6 +13,10 @@ for pomFile in $(find . -mindepth 2 -name pom.xml | sort ); do
       [[ "${pomFile}" =~ .*java-logging-logback.* ]] || \
       [[ "${pomFile}" =~ .*java-bigquery.* ]] || \
       [[ "${pomFile}" =~ .*sdk-platform-java.* ]] || \
+      [[ "${pomFile}" =~ .*java-common-protos.* ]] || \
+      [[ "${pomFile}" =~ .*java-showcase.* ]] || \
+      [[ "${pomFile}" =~ .*java-common-protos.* ]] || \
```
blakeli0: `java-common-protos` is duplicated. Also why are we excluding them? Especially for java-common-protos and java-iam, they should be pure generated.

java-showcase/README.md
blakeli0, line 1: There are a few scripts mentioned in README regarding how to run and update showcase tests. Can we make sure those scripts still work after migration?

generation/check_non_release_please_versions.sh
chingor13: This is a check that any declared dependency versions are all managed by release-please. common-protos and iam declare dependency versions for protobuf, guava, and grpc. We can't use the first-party-dependencies import because common-protos and iam are in first-party-dependencies.

java-showcase/README.md
chingor13: All the commands in the java-showcase/README.md are working. The test ones are exercised as part of CI. The regeneration showcase worked locally (takes a long time to rebuild the generator image)

chingor13:
The showcase-clirr check fails as we are moving the showcase source and it's comparing against head (which doesn't have the new location)

suztomo (approved):

blakeli0 (approved):

diegomarquezp (approved):

PR #12845, opened 2026-04-17
feat(spanner): add shared endpoint cooldowns for location-aware rerouting

rahul2393:
## Summary

This PR improves Java Spanner's location-aware bypass routing when routed replicas are overloaded or unavailable, and  extends score-based replica selection

  The client now:

  - avoids recently overloaded routed endpoints using shared cooldowns
  - records RESOURCE_EXHAUSTED / UNAVAILABLE as EWMA error penalties
  - uses EWMA-based selection for both preferLeader=false and strong preferLeader=true read/query routing when
    operation_uid is available

  It also keeps the location-aware read path lock-free via immutable group snapshots.

  ## What changed

  - Added shared channel-level cooldown tracking for routed endpoints that return RESOURCE_EXHAUSTED / UNAVAILABLE, while still keeping request-scoped exclusions for same-logical-request retries.
  - Updated bypass retry behavior so eligible reads/queries can reroute to another replica instead of immediately
    returning to the same failed endpoint.
  - Recorded RESOURCE_EXHAUSTED / UNAVAILABLE as EWMA error penalties for routed replicas, so unhealthy endpoints are deprioritized even after the immediate retry/cooldown window.
  - Extended score-based routing to strong preferLeader=true read/query traffic when operation_uid is present, using  leader preference as a bias instead of a hard override.
  - Kept preferLeader=true behavior unchanged for paths without operation_uid such as mutation/commit routing.
  - Refactored KeyRangeCache group state to immutable snapshots and removed per-group synchronization from the routing hot path.


java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EndpointLatencyRegistry.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  private static long defaultRttMicros() {
+    return DEFAULT_RTT.toNanos() / 1_000L;
+  }
+
+  @VisibleForTesting
+  static final class TrackerKey {
+    private final long operationUid;
+    private final String address;
```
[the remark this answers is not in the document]
rahul2393: Added database scope

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EwmaLatencyTracker.java
```diff
@@ -92,4 +123,13 @@ public void recordError(Duration penalty) {
     // Treat the error as a sample with high latency (penalty)
     update(penalty);
   }
+
+  private double calculateTimeBasedAlpha(long nowNanos) {
+    long deltaNanos = nowNanos - lastUpdatedAtNanos;
+    if (deltaNanos <= 0L) {
+      return 1.0;
+    }
```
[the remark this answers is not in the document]
rahul2393: Intentional

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -431,8 +436,15 @@ public GapicSpannerRpc(final SpannerOptions options) {
...
-        this.readRetryableCodes =
+        Set<Code> streamingReadRetryableCodes =
             options.getSpannerStubSettings().streamingReadSettings().getRetryableCodes();
+        this.readRetryableCodes =
+            enableLocationApi
+                ? ImmutableSet.<Code>builder()
+                    .addAll(streamingReadRetryableCodes)
+                    .add(Code.RESOURCE_EXHAUSTED)
+                    .build()
+                : streamingReadRetryableCodes;
```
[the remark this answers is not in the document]
rahul2393: Thanks, added

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EndpointLatencyRegistry.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+    double activeRequests = getInflight(endpointLabelOrAddress);
+    LatencyTracker tracker = TRACKERS.getIfPresent(trackerKey);
+    if (tracker != null) {
+      return tracker.getScore() * (activeRequests + 1.0);
+    }
+    if (activeRequests > 0.0) {
+      return DEFAULT_PENALTY_VALUE + activeRequests;
+    }
+    return defaultRttMicros() * (activeRequests + 1.0);
```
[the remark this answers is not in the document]
rahul2393: We are adding a tracker key/operation_id for valid bypass traffic always

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
olavloite: Not introduced in this PR, but: This map could in theory be a source of a memory leak: If an application does not call Commit or Rollback for a transaction, then that transaction affinity will not be removed from this map.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EndpointLatencyRegistry.java
```diff
@@ -0,0 +1,225 @@
...
+
+  static final Duration DEFAULT_ERROR_PENALTY = Duration.ofSeconds(10);
+  static final Duration DEFAULT_RTT = Duration.ofMillis(10);
+  static final double DEFAULT_PENALTY_VALUE = 1_000_000.0;
+  @VisibleForTesting static final Duration TRACKER_EXPIRE_AFTER_ACCESS = Duration.ofMinutes(10);
+  @VisibleForTesting static final long MAX_TRACKERS = 100_000L;
+
+  private static volatile Cache<TrackerKey, LatencyTracker> TRACKERS =
+      newTrackerCache(Ticker.systemTicker());
+  private static final ConcurrentHashMap<String, AtomicInteger> INFLIGHT_REQUESTS =
```
olavloite: We seem to only be adding data to this map, and never removing anything. However, it will only get new entries when a new endpoint address is seen, meaning that it will probably not grow far. (There is a `clear` method, but it is only being called from tests)

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EwmaLatencyTracker.java
```diff
@@ -77,12 +104,16 @@ public void update(Duration latency) {
...
       latencyMicros = Long.MAX_VALUE;
     }
+    long nowNanos = nanoTimeSupplier.getAsLong();
     synchronized (lock) {
       if (!initialized) {
         score = latencyMicros;
         initialized = true;
+        lastUpdatedAtNanos = nowNanos;
       } else {
+        double alpha = fixedAlpha != null ? fixedAlpha : calculateTimeBasedAlpha(nowNanos);
```
olavloite, line 114: Is it strictly necessary to do this calculation while holding the lock? Or could we move the calculation to outside the lock, and only take the lock when we are reading/writing the `lastUpdatedAtNanos`?

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyRangeCache.java
```diff
@@ -698,19 +565,40 @@ String debugString() {
...
+  private static final class GroupSnapshot {
+    final ByteString generation;
+    final int leaderIndex;
+    final List<TabletSnapshot> tablets;
+
+    private GroupSnapshot(ByteString generation, int leaderIndex, List<TabletSnapshot> tablets) {
+      this.generation = generation;
+      this.leaderIndex = leaderIndex;
+      this.tablets = Collections.unmodifiableList(new ArrayList<>(tablets));
+    }
```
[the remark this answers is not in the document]
olavloite: Please address this

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EndpointLifecycleManager.java
```diff
@@ -104,6 +106,7 @@ static final class EndpointState {
 
   private final ChannelEndpointCache endpointCache;
   private final Map<String, EndpointState> endpoints = new ConcurrentHashMap<>();
+  private final Set<String> evictedAddresses = ConcurrentHashMap.newKeySet();
```
olavloite: We seem to only remove elements from this map when an endpoint is being re-created. That means that if an endpoint goes away and never 'comes back', then it will remain in this map. If that happens often, then that could be a potential source of a memory leak.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyRangeCache.java
```diff
@@ -720,115 +752,363 @@ synchronized void update(Group groupIn) {
...
+        GroupSnapshot snapshot, boolean hasDirectedReadOptions) {
+      if (!hasDirectedReadOptions
+          && snapshot.hasLeader()
+          && snapshot.leader().distance <= MAX_LOCAL_REPLICA_DISTANCE) {
+        return snapshot.leader();
+      }
+      return null;
+    }
+
+    private TabletSnapshot selectScoreAwareTablet(
```
olavloite, line 887: This method is hard to read, partly due to the length and partly due to the number of branches. And I think that there are multiple opportunities to optimize it. Now, it potentially loops over the eligibleTablets more than once. It also creates two ArrayLists and a HashMap each time it is invoked. I think that it would be worth taking a second look at this method, considering how often it will be invoked.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
olavloite: We can probably remove this call and move it to a background thread, or we can just execute it once every 1,000 requests, and still get the same result. That would probably be more efficient than the current implementation that calls it every time and then tries to iterate over the queue.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/EwmaLatencyTracker.java
rahul2393: I will be keeping this for now, moving the calculation out would either introduce unsynchronized access or compute alpha from stale state under contention, will do in follow-up if needed

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyRangeCache.java
rahul2393: I have refactored it for readability, will improve in followup

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -399,33 +415,14 @@ private static boolean shouldExcludeEndpointOnRetry(io.grpc.Status.Code statusCo
 
   private Predicate<String> consumeExcludedEndpointsForCurrentCall(
```
olavloite: nit: can we get rid of this method? Or otherwise rename it, as I don't think the name corresponds with what it actually does anymore.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyRangeCache.java
```diff
@@ -861,7 +835,7 @@ private TabletSnapshot selectTablet(
           && snapshot.hasLeader()
           && snapshot.leader().distance <= MAX_LOCAL_REPLICA_DISTANCE) {
         checkedLeader = true;
-        selectionStats.matchingReplicas++;
+        selectionStats.sawMatchingReplica = true;
```
olavloite, line 838: nit: we should consider renaming `SelectionStats` to something like `SelectionFlags` or `SelectionState`. It does not contain stats (anymore), it only contains flags.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyRangeCache.java
```diff
@@ -948,124 +950,105 @@ private TabletSnapshot selectScoreAwareTablet(
...
-              operationUid,
-              selectionCostLookup(operationUid, preferredLeader),
-              selectionReason,
-              selected,
-              scoredCandidates);
-      return selected;
+      return lowestCost;
+    }
+
+    private List<ChannelEndpoint> endpointView(List<EligibleReplica> eligibleReplicas) {
```
olavloite, line 993: nit: You can use `Lists.transform(..)` for this

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/RequestIdTargetTracker.java
olavloite, line 36: I think that it would be good to set an explicit value for `concurrency` for this cache. The default is 4, but the documentation explicitly states that this could change in the future. And as this cache is heavily used, it might be better to set a higher concurrency level than the default 4. Maybe one that is equal to the number of processors (with a minimum of 4)?

olavloite (approved):

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/RequestIdTargetTracker.java
rahul2393: Update to use processor count

PR #12848, opened 2026-04-18
build: fall back to previous image version on release PRs

### LINKED ISSUE #12825: Library-generation fails on release PRs
```
+ docker run --rm -u 1001:1001 -v /home/runner/work/google-cloud-java/google-cloud-java:/workspace -v /tmp/tmp.YRQLyh2Yi8:/workspace/googleapis -e GENERATOR_VERSION=2.71.0 gcr.io/cloud-devrel-public-resources/java-library-generation:2.71.0 --generation-config-path=/workspace/generation_config.yaml --library-names= --api-definitions-path=/workspace/googleapis
Unable to find image 'gcr.io/cloud-devrel-public-resources/java-library-generation:2.71.0' locally
docker: Error response from daemon: manifest for gcr.io/cloud-devrel-public-resources/java-library-generation:2.71.0 not found: manifest unknown: Failed to fetch "2.71.0"
```

meltsufin:
The library generation pipeline fails on release PRs because it attempts to pull a Docker image version that has not yet been built or pushed to the registry.

This PR adds a fallback mechanism in hermetic_library_generation.sh to use the previous version of the image from the main branch if the requested version fails to pull on a release PR.

To verify this in CI, this branch is named with the prefix 'release-please--' to simulate a release PR.

Fixes #12825


meltsufin:
Fallback validated in a separate release simulation PR #12849:
```
+ [[ release-please--simulation =~ ^release-please-- ]]
+ echo 'Detected release PR branch: release-please--simulation'
+ docker pull gcr.io/cloud-devrel-public-resources/java-library-generation:9.99.9
Detected release PR branch: release-please--simulation
Error response from daemon: manifest for gcr.io/cloud-devrel-public-resources/java-library-generation:9.99.9 not found: manifest unknown: Failed to fetch "9.99.9"
+ echo 'Image not found for version 9.99.9. Falling back to previous version from main.'
Image not found for version 9.99.9. Falling back to previous version from main.
++ git show main:.github/workflows/hermetic_library_generation.yaml
++ grep -m 1 '^[[:space:]]*image_tag:'
++ cut -d : -f 2-
++ cut -d '#' -f 1
++ xargs
+ previous_tag=2.71.0
+ '[' -n 2.71.0 ']'
+ echo 'Using previous image version: 2.71.0'
+ image_tag=2.71.0
Using previous image version: 2.71.0
++ id -u
++ id -g
++ pwd
+ docker run --rm -u 1001:1001 -v /home/runner/work/google-cloud-java/google-cloud-java:/workspace -v /tmp/tmp.dG9h7KnPTo:/workspace/googleapis -e GENERATOR_VERSION=2.71.0 gcr.io/cloud-devrel-public-resources/java-library-generation:2.71.0 --generation-config-path=/workspace/generation_config.yaml --library-names= --api-definitions-path=/workspace/googleapis
Unable to find image 'gcr.io/cloud-devrel-public-resources/java-library-generation:2.71.0' locally
2.71.0: Pulling from cloud-devrel-public-resources/java-library-generation
```

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
```diff
@@ -123,13 +124,28 @@ changed_libraries="$(cat "${changed_libraries_file}")"
 echo "Changed libraries are: ${changed_libraries:-"No changed library"}."
 
 # run hermetic code generation docker image.
+# Attempt to pull the image to see if it exists on release PRs.
+if [[ "$current_branch" =~ ^release-please-- ]]; then
+  echo "Detected release PR branch: $current_branch"
+  if ! docker pull "${IMAGE_NAME}:${image_tag}"; then
+    echo "Image not found for version ${image_tag}. Falling back to previous version from ${target_branch}."
+    previous_tag=$(git show "${target_branch}":.github/workflows/hermetic_library_generation.yaml | grep -m 1 "^[[:space:]]*image_tag:" | cut -d ':' -f 2- | cut -d '#' -f 1 | xargs || true)
```
suztomo: The main branch is updated through a snapshot version of the container. Using an old version brings the change back.

Blake wrote:

```
The flow after migration is: 
every commit to main triggers a Cloud Build job -> 
create a snapshot hermetic image -> 
nightly generation uses the latest snapshot image to regenerate the whole repo 
-> google-cloud-java release
```

meltsufin: Can you clarify what is the problem?
Since Release Please PRs only update versions and changelogs and do not modify the generator itself, using the previously pushed image from the latest commit in main should be safe. It should not produce regressions because the generation logic is identical to what produced the current state of main.

suztomo: > using the previously pushed image from the latest commit in main

If that's the case (it's using the latest pushed snapshot image), then that's good. I was under the impression that the "previous image version" to be the one published in the previous release cycle (2 weeks ago). Can you confirm which is the case?

For reference, I got `2.71.0` from this command:

```
suztomo@suztomo:~/migration-work/google-cloud-java$ git show "${target_branch}":.github/workflows/hermetic_library_generation.yaml | grep -m 1 "^[[:space:]]*image_tag:" | cut -d ':' -f 2- | cut -d '#' -f 1 | xargs || true
2.71.0
```

blakeli0: > Since Release Please PRs only update versions and changelogs and do not modify the generator itself

But the image contains the generator, I think this does introduce risks as @suztomo mentioned.

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
```diff
@@ -123,13 +124,28 @@ changed_libraries="$(cat "${changed_libraries_file}")"
 echo "Changed libraries are: ${changed_libraries:-"No changed library"}."
 
 # run hermetic code generation docker image.
+# Attempt to pull the image to see if it exists on release PRs.
+if [[ "$current_branch" =~ ^release-please-- ]]; then
```
blakeli0: We could create release-please PRs from CLI as well.

meltsufin: That's fine, as long as the branch prefix is `release-please--`.

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
suztomo: Why don't we just use the "latest" tag? The intention is more clear than these if-statements and `$(git show "${target_branch}":.github/workflows/hermetic_library_generation.yaml | grep -m 1 "^[[:space:]]*image_tag:" | cut -d ':' -f 2- | cut -d '#' -f 1 | xargs || true)`.

suztomo (approved):
Approving but please consider whether to use the "latest" tag.

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
meltsufin: @blakeli0 Can you please explain the risk a bit more? As far as I understand, the image gets rebuilt on push to main and tagged with `:latest` and `:{version}`. The version tag is floating to the latest build from that branch. So, a release pull request targeting that branch should be able to use the latest image at last version safely because it should have no functional difference with the release commit version of the generator.

meltsufin:
> Approving but please consider whether to use the "latest" tag.

@suztomo  I think using the version in the target branch is more robust because it allows the code to support multiple branches with different versions of the generator.

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
blakeli0: But that means we have to remember using `release-please--` as the branch name, is there any harm we do this for every branches?

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
blakeli0: Does the script checking for previous SNAPSHOT versions as well? If yes, then we should be good.

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
meltsufin: You shouldn't have to use the Release Please CLI going forward which should eliminate the need to remember the branch prefix. The risk of always falling back is that if someone configures and invalid generator version, they won't get a CI failure as feedback.

sdk-platform-java/.github/scripts/hermetic_library_generation.sh
meltsufin: Yes, it should always use `gapic-generator-java:current` version, which would at times have `-SNAPSHOT`.

blakeli0 (approved):

generation_config.yaml
```diff
@@ -1,4 +1,3 @@
-gapic_generator_version: 2.70.0
```
meltsufin, line 1: @blakeli0  Note that I'm removing this version declaration because it's confusing as it doesn't get updated, and it's not used anyway.

sdk-platform-java/.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -43,6 +43,10 @@ ARG GRPC_VERSION=1.80.0
 ENV HOME=/home
 ENV OS_ARCHITECTURE="linux-x86_64"
 
+# {x-version-update-start:gapic-generator-java:current}
+ENV GENERATOR_VERSION="2.71.0"
```
blakeli0, line 47: I don't think this is needed?

generation_config.yaml
blakeli0: Agreed. We used to update this version in the daily generation PR to trigger a full generation, but we always use the latest snapshot version now after monorepo migration.

suztomo: Do you intend to fix this file in the split repositories?

E.g, https://github.com/googleapis/java-firestore/blob/main/generation_config.yaml#L1

suztomo (approved):

generation_config.yaml
meltsufin: I think it's OK to wait until the other repos migrate to the monorepo.

sdk-platform-java/.cloudbuild/library_generation/library_generation.Dockerfile
meltsufin: Yes, `sdk-platform-java/hermetic_build/common/model/generation_config.py` has a method `__set_generator_version` that checks if the `gapic_generator_version` is specified in the YAML file. If it is not specified in the YAML, the code falls back to reading the `GENERATOR_VERSION` env var.

generation_config.yaml
blakeli0: Yep, this is resulted from monorepo migration. I don't think we want to do it for split repos.

sdk-platform-java/.cloudbuild/library_generation/library_generation.Dockerfile
blakeli0: I see. We are already setting it in the CI [hermetic_library_generation.yaml](https://github.com/googleapis/google-cloud-java/pull/12848/changes/BASE..7dc57e7e0e70ceb0ff9c83fefb137572a419e65c#diff-7e628bfb722b6b3b22179bce41c2518f4fe12b7e9b7e4b91aaea1b2ffa51aa28) though, so this is for scenarios that are not triggered through CI?

PR #12856, opened 2026-04-20
chore(managed-identities): rm unused config API path

zhumin8:
Removing unused API path from config for managed-identities, so that it matches the generated code status. 

v1beta1 of java-managed-identities was added to the generation_config.yaml as part of initial migration to hermetic build: https://github.com/googleapis/google-cloud-java/commit/14b73467ea8e5037eb713610714a532fbba1a769, but client library code for v1beta1 is not present in java-managed-identities/google-cloud-managed-identities dir. This is due to deep-copy-regex being too strict and does not copy /v1beta1 from staging dir to final output.

Expect generation workflow to remove sample code related to v1beta1.

For https://github.com/googleapis/google-cloud-java/issues/12736


zhumin8:
hmm, we got no changed libraries
```
+ changed_libraries=
+ echo 'Changed libraries are: No changed library.'
Changed libraries are: No changed library.
```

zhumin8:
Forced a code generation for this library by a temp change (3e4f22a)
Generated sample code under this dir are removed as expected: java-managed-identities/samples/snippets/generated/com/google/cloud/managedidentities/v1beta1

reverted above temp change.

zhumin8:
@blakeli0 PTAL.
The removed files are sample code within
`java-managed-identities/samples/snippets/generated/com/google/cloud/managedidentities/v1beta1/` I think it's reasonable to remove since the relevant code for v1beta1 does not exists and samples are not published.

lqiu96:
> This is due to deep-copy-regex being too strict and does not copy /v1beta1 from staging dir to final output.

Please correct me if I'm misunderstanding this, but does this mean that the reason we don't have v1beta1 library is because of a config error on our end? If so, do we want to fix this so that managedidentities v1beta1 is generated properly?

lqiu96:
NVM, I see in the librarian ticket:

> Agreed. If they were never generated, I think it's OK to not generate them before migration.

Ok, I think removing the samples if fine for now. Do we have a ticket to generate it after migration (if we still want to)?

lqiu96 (approved):

blakeli0:
> Do we have a ticket to generate it after migration (if we still want to)?

Since this is v1beta1 and we never generated before, I think it's OK to not generate it for now. We can always add it back if someone needs it.

blakeli0 (approved):

zhumin8:
updated branch in attempt to trigger the required "conditional-required-check" test

PR #12858, opened 2026-04-20
chore(os-login): rm unused config API path

zhumin8:
Removing unused API path from config for os-login, so that it matches the generated code status.

v1beta of java-os-login was added to the generation_config.yaml as part of initial migration to hermetic build: https://github.com/googleapis/google-cloud-java/commit/14b73467ea8e5037eb713610714a532fbba1a769, but client library code for v1beta is not present in java-os-login/google-cloud-os-login dir. This is due to deep-copy-regex being too strict and does not copy /v1beta from staging dir to final output.

Expect generation workflow to run successfully with not change. (no sample code exists for [os-login](https://github.com/googleapis/google-cloud-java/tree/main/java-os-login))

For https://github.com/googleapis/google-cloud-java/issues/12736

### DIFF
--- a/generation_config.yaml
+++ b/generation_config.yaml
@@ -1977,7 +1977,6 @@ libraries:
   issue_tracker: https://issuetracker.google.com/savedsearches/559755
   GAPICs:
   - proto_path: google/cloud/oslogin/v1
-  - proto_path: google/cloud/oslogin/v1beta
 - api_shortname: parallelstore
   name_pretty: Parallelstore API
   product_documentation: https://cloud/parallelstore?hl=en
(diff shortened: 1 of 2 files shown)


zhumin8:
hmm, this config change did not trigger hermetic build:
```
+ changed_libraries=
+ echo 'Changed libraries are: No changed library.'
Changed libraries are: No changed library.
```
Forcing it to run with temp change to script

zhumin8:
After forcing generate on this library, now hermetic build workflow correctly triggers generate and completes with no code change:
```
Changed libraries are: os-login.
```
full log: https://github.com/googleapis/google-cloud-java/actions/runs/24739089474/job/72373657169?pr=12858

Now proceed to revert temp change.

blakeli0 (approved):

PR #12862, opened 2026-04-20
fix: manual preservation of Version.java files

diegomarquezp:
### Description
This PR addresses an issue where `Version.java` files were either incorrectly preserved or not generated during a library's initial generation due to an overly relaxed `deep-preserve-regex` rule in the OwlBot configuration templates. 

To fix this edge case, we are moving the responsibility of preserving existing `Version.java` files away from the OwlBot YAML regex and directly into our hermetic build scripts. 

### Implementation Details
1. **Removed Obsolete Regex**: Removed the `deep-preserve-regex` rule targeting `Version.java` from `owlbot.yaml.monorepo.j2`. 
2. **Conditional Backup/Restore**: Extracted new backup and restore utilities into `utilities.sh` and invoked them in `postprocess_library.sh`. The script now dynamically scans for existing `Version.java` files, backs them up to a temporary directory before executing `owl-bot copy-code`, and restores them immediately after.

### Why this approach?
- **First Generation**: Since there are no existing `Version.java` files, the script does nothing, allowing OwlBot to generate and retain the new file as intended.
- **Release Please Interoperability**: Release Please regularly updates the version string inside existing `Version.java` files. Because our `backup_version_java` utility copies the *existing* file out of the repository *before* `owl-bot copy-code` executes, the file containing the updated Release Please version string is safely stored in `mktemp -d`. After `owl-bot copy-code` writes the newly generated, generic `Version.java` file to the repo, `restore_version_java` overwrites it with our backup. This guarantees the Release Please version updates are perfectly preserved.
- **Hermetic Safety**: The backups are stored in `mktemp -d` which guarantees they are unaffected by OwlBot's `deep-remove-regex` folder wipes. 

### Testing
- [x] Added unit tests
- [ ] Once approach is validated: modify the `chore/test-hermetic-build` branch with arbitrary values in `Version.java` files and confirm they remain untouched via integration test


sdk-platform-java/hermetic_build/library_generation/utils/utilities.sh
```diff
@@ -334,6 +334,45 @@ error_if_not_exists() {
   fi
 }
 
+# Backups existing Version.java files in the specified target directory
+# Arguments:
+# 1 - target_directory: The directory where to search for Version.java
+# 2 - backup_directory: The directory where the backups will be stored
+backup_version_java() {
```
blakeli0, line 341: Is it possible to treat Version.java the same way as other bootstrap files and use [owlbot.py](https://github.com/googleapis/google-cloud-java/blob/main/java-ces/owlbot.py#L26-L38) to exclude it?

blakeli0:
cc: @zhumin8 since this may affect librarian migration.

sdk-platform-java/hermetic_build/library_generation/utils/utilities.sh
diegomarquezp: That logic only handles [owlbot templates for java libraries](https://github.com/googleapis/google-cloud-java/tree/main/sdk-platform-java/hermetic_build/library_generation/owlbot/templates/java_library), unfortunately.

Furthermore, owlbot.py is called after copy-code, so the overwrite of Version.java would already have been issued by then.

blakeli0: Thanks, makes sense.

blakeli0 (approved):

PR #12863, opened 2026-04-20
fix(gax): Remove strict validation that resize delta must be less than max channel count

lqiu96:
Address some comments in https://github.com/googleapis/google-cloud-java/pull/12838

It is possible that users use the DEFAULT_RESIZE_DELTA (2) with a statically sized channel pool (1). We should not enforce that the resize delta needs to be less than or equal to the channel pool. Instead, the channel pool's tentative target is calculated to be clamped within the channelpoolsetting's configured bounds.


sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
```diff
@@ -178,6 +191,11 @@ public ChannelPoolSettings build() {
           "initial channel count must be less than maxChannelCount");
       Preconditions.checkState(
           s.getInitialChannelCount() > 0, "Initial channel count must be greater than 0");
+      Preconditions.checkState(
+          s.getMaxResizeDelta() > 0, "Max resize delta must be greater than 0");
+      Preconditions.checkState(
+          s.getMaxResizeDelta() <= s.getMaxChannelCount(),
+          "Max resize delta cannot be greater than max channel count");
       return s;
```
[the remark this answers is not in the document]
jinseopkim0: Would it make sense to remove the `checkState` validation?

sdk-platform-java/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/ChannelPoolTest.java
```diff
@@ -899,4 +849,75 @@ void testDoubleRelease() throws Exception {
       ChannelPool.LOG.removeHandler(logHandler);
     }
   }
+
+  @Test
+  void settingsValidationFailsWhenMinChannelsExceedsMaxChannels() {
+    Assertions.assertThrows(
```
jinseopkim0: To resolve the warning, would it make sense to do something like this?
```
ChannelPoolSettings.Builder builder = ChannelPoolSettings.builder()
    .setMinChannelCount(2)
    .setMaxChannelCount(1);
Assertions.assertThrows(IllegalStateException.class, () -> builder.build());
```

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -84,6 +89,16 @@ class ChannelPool extends ManagedChannel {
...
 
+  // The number of consecutive resize cycles to wait before logging a warning about repeated
+  // resizing. This value was chosen to detect repeated requests for changes (multiple continuous
+  // increase or decrease attempts) without being too sensitive.
+  private static final int CONSECUTIVE_RESIZE_THRESHOLD = 5;
+
+  // Tracks the number of consecutive resize cycles where a resize actually occurred (either expand
+  // or shrink). Used to detect repeated resizing activity and log a warning.
+  // Note: This field is only accessed safely within resizeSafely() and does not need to be atomic.
+  private int consecutiveResizes = 0;
```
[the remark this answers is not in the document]
jinseopkim0: Would it make sense to do something like this?
```
// Tracks the number of consecutive resize cycles where a resize actually occurred.
private final AtomicInteger consecutiveResizes = new AtomicInteger(0);
```
If so, then `resize()` would need to be updated accordingly.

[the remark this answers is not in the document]
lqiu96: I don't think this is relevant (atleast not for now). `consecutiveResizes` is only used inside `resize()` and locked by entryWriteLock, whereas `indexTicker` is used outside and needs to be an AtomicInteger.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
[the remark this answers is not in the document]
lqiu96: Ah yeah, I think we can remove this. Since we already have logic that handles clamping within ChannelPool itself.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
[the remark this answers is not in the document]
lqiu96: Actually, probably no harm in changing this. Performance would be minimal anyways

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java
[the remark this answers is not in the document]
lqiu96: Hmm wait actually, channelpool logic doesn't clamp the delta value.

Should resolvedd now

jinseopkim0:
There're many CI failures.
```
Error: 9:580 [ERROR] Errors: 
Error: 9:580 [ERROR]   BigQueryConnectionTest.testBigQueryReadClientKeepAliveSettings:374 » IllegalState Max resize delta cannot be greater than max channel count
Error: 9:580 [ERROR]   BigQueryConnectionTest.testGetWriteClient:231 » IllegalState Max resize delta cannot be greater than max channel count
Error: 9:580 [ERROR]   BigQueryJdbcProxyUtilityTest.testGetTransportChannelProvider_withSslAndProxy:288 » IllegalState Max resize delta cannot be greater than max channel count
Error: 9:580 [ERROR]   BigQueryJdbcProxyUtilityTest.testGetTransportChannelProvider_withSslTrustStore_noPassword:265 » IllegalState Max resize delta cannot be greater than max channel count
Error: 9:580 [ERROR]   BigQueryJdbcProxyUtilityTest.testGetTransportChannelProvider_withSslTrustStore_withCorrectPassword:275 » IllegalState Max resize delta cannot be greater than max channel count
```
Can you please take a look?

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -319,34 +322,49 @@ void resize() {
...
-    }
+    // Dampen the rate of change if the desired delta exceeds the maximum allowed step size.
+    // Ensure that the step size is capped by the max channel count to handle small pool
+    // configurations.
+    int effectiveMaxResizeDelta =
+        Math.min(settings.getMaxResizeDelta(), settings.getMaxChannelCount());
+    if (Math.abs(delta) > effectiveMaxResizeDelta) {
+      // Limit the change to effectiveMaxResizeDelta, maintaining the correct direction (positive or
+      // negative).
+      dampenedTarget = currentSize + (int) Math.copySign(effectiveMaxResizeDelta, delta);
```
jinseopkim0: nit: `Math.copySign` is for `double` or `float` (https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#copySign-double-double-), so this would be ineffcient. Would it make sense to replace this with something like this?
```
int step = delta > 0 ? effectiveMaxResizeDelta : -effectiveMaxResizeDelta;
dampenedTarget = currentSize + step;
```

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
jinseopkim0, line 556: (Please feel free to address this in a follow-up.) Is this prone to race conditions? The sequence of `outstandingRpcs.incrementAndGet()`, `maxOutstanding.get()`, then `maxOutstanding.incrementAndGet()` is not an atomic operation. Would it make sense to update this to the following?
```
maxOutstanding.updateAndGet(prev -> Math.max(prev, currentOutstanding));
```

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -319,34 +322,49 @@ void resize() {
...
+    }
+
+    // Ensure that the calculated dampedTarget value will never exceed the maxChannelCount or fall
```
jinseopkim0: nit: s/dampedTarget/dampenedTarget/

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: Updated!

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: Updated!

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: I don't know if this causes any direct issues off the top of my head. There maybe be some less than optimal channel pool calculations/ behaviors because of this and definitely may warrant another look.

From what I understand, the behavior between outstandingRpcs and maxOutstanding is just an approximation to try and figure out how to resize the number of channels in the channel pool. I don't think this is intended to be an exact calculation.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -319,34 +322,51 @@ void resize() {
...
+      // Maintaining the correct direction (positive or negative) to handle expand/shrink
+      int step = delta > 0 ? effectiveMaxResizeDelta : -effectiveMaxResizeDelta;
+      dampenedTarget = currentSize + step;
+    }
+
+    // Ensure that the calculated dampenedTarget value will never exceed the maxChannelCount or fall
+    // below minChannelCount. This ensures that `currentSize + resizeDelta` remains within bounds.
+    dampenedTarget =
+        Math.max(
+            settings.getMinChannelCount(), Math.min(settings.getMaxChannelCount(), dampenedTarget));
```
lqiu96: on a second look, I'm not sure we still need this check here. max and min are already enforced within the bounds

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -319,34 +322,51 @@ void resize() {
...
+    // Rate-limit the change to not exceed the effectiveMaxResizeDelta
+    if (Math.abs(delta) > effectiveMaxResizeDelta) {
+      // Maintaining the correct direction (positive or negative) to handle expand/shrink
+      int step = delta > 0 ? effectiveMaxResizeDelta : -effectiveMaxResizeDelta;
+      dampenedTarget = currentSize + step;
+    }
+
+    // Ensure that the calculated dampenedTarget value will never exceed the maxChannelCount or fall
+    // below minChannelCount. This ensures that `currentSize + resizeDelta` remains within bounds.
+    dampenedTarget =
```
blakeli0: IIUC, the problem we are trying to solve is that the resized channel count could exceed the configured max channel count? Other than the configured max channel count is not respected, are there any harm if it happens? Also it seems like an existing issue, and we are fixing it because the new setting makes it easier to exceed the configured count?

sdk-platform-java/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/ChannelPoolTest.java
blakeli0, line 1: Can we add some unit tests for the new logics? If it's hard to test the whole channel pool resizing logic, we could extract the part that calculates target channel size and only test that part.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: I think I misunderstood the original logic. I don't think it ever goes past the min and max channel bounds.

sdk-platform-java/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/ChannelPoolTest.java
lqiu96: Yep, will do

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
lqiu96: I updated the title to better reflect the changes. I think I misunderstood some of the original logic and certain edge cases were already handled.

jinseopkim0 (approved):

PR #12867, opened 2026-04-20
chore(auth): Address remaining Regional Access Boundary feedback

vverman:
1. The RAB refresh uses a direct executor with a fixed thread pool as opposed to instantiating a new thread each time.

2. The RAB env gate -> GOOGLE_AUTH_TRUST_BOUNDARY_ENABLE_EXPERIMENT has been removed. This means RAB refresh triggers by default.

3. Added other fixes/suggestions made in the previous Java [PR](https://github.com/googleapis/google-auth-library-java/pull/1880).

### DIFF
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
@@ -67,24 +67,6 @@ final class RegionalAccessBoundary implements Serializable {
   static final String X_ALLOWED_LOCATIONS_HEADER_KEY = "x-allowed-locations";
   private static final long serialVersionUID = -2428522338274020302L;
 
-  private static final ThreadLocal<Boolean> DISABLE_RAB_FOR_TESTS =
-      ThreadLocal.withInitial(() -> false);
-
-  @VisibleForTesting
-  static void disableForTests() {
-    DISABLE_RAB_FOR_TESTS.set(true);
-  }
-
-  @VisibleForTesting
-  static void enableForTests() {
-    DISABLE_RAB_FOR_TESTS.set(false);
-  }
-
-  @VisibleForTesting
-  static void resetForTests() {
-    DISABLE_RAB_FOR_TESTS.remove();
-  }
-
   static final long TTL_MILLIS = 6 * 60 * 60 * 1000L; // 6 hours
   static final long REFRESH_THRESHOLD_MILLIS = 1 * 60 * 60 * 1000L; // 1 hour
 
@@ -191,10 +173,7 @@ final class RegionalAccessBoundary implements Serializable {
    *
    * @return True if the regional access boundary feature is enabled, false otherwise.
    */
-  @InternalApi
-  static boolean isEnabled() {
-    return !DISABLE_RAB_FOR_TESTS.get();
-  }
+
 
   /**
    * Refreshes the regional access boundary by making a network call to the lookup endpoint.
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
@@ -40,10 +40,11 @@ import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpResponse;
 import com.google.api.client.http.HttpUnsuccessfulResponseHandler;
 import com.google.api.client.json.GenericJson;
-import com.google.api.client.json.JsonParser;
+import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.util.Clock;
 import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.client.util.Key;
+import com.google.api.core.InternalApi;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
@@ -53,7 +54,6 @@ import java.io.ObjectInputStream;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.List;
-import javax.annotation.Nullable;
 
 /**
  * Represents the regional access boundary configuration for a credential. This class holds the
@@ -67,10 +67,24 @@ final class RegionalAccessBoundary implements Serializable {
   static final String X_ALLOWED_LOCATIONS_HEADER_KEY = "x-allowed-locations";
   private static final long serialVersionUID = -2428522338274020302L;
 
-  // Note: this is for internal testing use use only.
-  // TODO: Fix unit test mocks so this can be removed
-  // Refer -> https://github.com/googleapis/google-auth-library-java/issues/1898
-  static final String ENABLE_EXPERIMENT_ENV_VAR = "GOOGLE_AUTH_TRUST_BOUNDARY_ENABLE_EXPERIMENT";
+  private static final ThreadLocal<Boolean> DISABLE_RAB_FOR_TESTS =
+      ThreadLocal.withInitial(() -> false);
+
+  @VisibleForTesting
+  static void disableForTests() {
+    DISABLE_RAB_FOR_TESTS.set(true);
+  }
+
+  @VisibleForTesting
+  static void enableForTests() {
+    DISABLE_RAB_FOR_TESTS.set(false);
+  }
+
+  @VisibleForTesting
+  static void resetForTests() {
+    DISABLE_RAB_FOR_TESTS.remove();
+  }
+
   static final long TTL_MILLIS = 6 * 60 * 60 * 1000L; // 6 hours
   static final long REFRESH_THRESHOLD_MILLIS = 1 * 60 * 60 * 1000L; // 1 hour
 
@@ -78,8 +92,6 @@ final class RegionalAccessBoundary implements Serializable {
   private final List<String> locations;
   private final long refreshTime;
   private transient Clock clock;
-
-  private static EnvironmentProvider environmentProvider = SystemEnvironmentProvider.getInstance();
 
   /**
    * Creates a new RegionalAccessBoundary instance.
@@ -172,28 +184,16 @@ final class RegionalAccessBoundary implements Serializable {
     }
   }
 
-  @VisibleForTesting
-  static void setEnvironmentProviderForTest(@Nullable EnvironmentProvider provider) {
-    environmentProvider = provider == null ? SystemEnvironmentProvider.getInstance() : provider;
-  }
-
   /**
-   * Checks if the regional access boundary feature is enabled. The feature is enabled if the
-   * environment variable or system property "GOOGLE_AUTH_TRUST_BOUNDARY_ENABLE_EXPERIMENT" is set
-   * to "true" or "1" (case-insensitive).
+   * Checks if the regional access boundary feature is enabled.
+   *
+   * <p>This method is for internal use only and may be changed or removed in future releases.
    *
    * @return True if the regional access boundary feature is enabled, false otherwise.
    */
+  @InternalApi
   static boolean isEnabled() {
-    String enabled = environmentProvider.getEnv(ENABLE_EXPERIMENT_ENV_VAR);
-    if (enabled == null) {
-      enabled = System.getProperty(ENABLE_EXPERIMENT_ENV_VAR);
-    }
-    if (enabled == null) {
-      return false;
-    }
-    String lowercased = enabled.toLowerCase();
-    return "true".equals(lowercased) || "1".equals(enabled);
+    return !DISABLE_RAB_FOR_TESTS.get();
   }
 
   /**
@@ -249,15 +249,20 @@ final class RegionalAccessBoundary implements Serializable {
     HttpIOExceptionHandler ioExceptionHandler = new HttpBackOffIOExceptionHandler(backoff);
     request.setIOExceptionHandler(ioExceptionHandler);
 
+    request.setParser(new JsonObjectParser(OAuth2Utils.JSON_FACTORY));
+
     RegionalAccessBoundaryResponse json;
+    HttpResponse response = null;
     try {
-      HttpResponse response = request.execute();
-      String responseString = response.parseAsString();
-      JsonParser parser = OAuth2Utils.JSON_FACTORY.createJsonParser(responseString);
-      json = parser.parseAndClose(RegionalAccessBoundaryResponse.class);
+      response = request.execute();
+      json = response.parseAs(RegionalAccessBoundaryResponse.class);
     } catch (IOException e) {
       throw new IOException(
           "RegionalAccessBoundary: Failure while getting regional access boundaries:", e);
+    } finally {
+      if (response != null) {
+        response.disconnect();
+      }
     }
     String encodedLocations = json.getEncodedLocations();
     // The encodedLocations is the value attached to the x-allowed-locations header, and
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
@@ -199,6 +199,8 @@ final class RegionalAccessBoundary implements Serializable {
 
     HttpRequestFactory requestFactory = transportFactory.create().createRequestFactory();
     HttpRequest request = requestFactory.buildGetRequest(new GenericUrl(url));
+    // Disable automatic logging by google-http-java-client to prevent leakage of sensitive tokens.
+    request.setLoggingEnabled(false);
     request.getHeaders().setAuthorization("Bearer " + accessToken.getTokenValue());
 
     // Add retry logic
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
@@ -165,14 +165,6 @@ final class RegionalAccessBoundary implements Serializable {
   }
 
   /**
-   * Checks if the regional access boundary feature is enabled.
-   *
-   * <p>This method is for internal use only and may be changed or removed in future releases.
-   *
-   * @return True if the regional access boundary feature is enabled, false otherwise.
-   */
-
-  /**
    * Refreshes the regional access boundary by making a network call to the lookup endpoint.
    *
    * @param transportFactory The HTTP transport factory to use for the network request.
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
@@ -44,9 +44,7 @@ import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.util.Clock;
 import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.client.util.Key;
-import com.google.api.core.InternalApi;
 import com.google.auth.http.HttpTransportFactory;
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
@@ -173,7 +171,6 @@ final class RegionalAccessBoundary implements Serializable {
    *
    * @return True if the regional access boundary feature is enabled, false otherwise.
    */
-
 
   /**
    * Refreshes the regional access boundary by making a network call to the lookup endpoint.
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
@@ -37,7 +37,6 @@ import com.google.auth.http.HttpTransportFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.SettableFuture;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
@@ -86,14 +86,24 @@ final class RegionalAccessBoundaryManager {
   // on concurrent refresh tasks, while threadCount provides unique names
   // for easier debugging.
   private static final AtomicInteger threadCount = new AtomicInteger(0);
-  private static final ExecutorService EXECUTOR =
-      Executors.newFixedThreadPool(
-          5,
-          r -> {
-            Thread t = new Thread(r, "RAB-refresh-" + threadCount.getAndIncrement());
-            t.setDaemon(true);
-            return t;
-          });
+  private static final ExecutorService EXECUTOR;
+
+  static {
+    java.util.concurrent.ThreadPoolExecutor executor =
+        new java.util.concurrent.ThreadPoolExecutor(
+            5, // corePoolSize: threads to keep alive
+            5, // maximumPoolSize: max threads allowed
+            1, // keepAliveTime: time to wait before terminating idle threads
+            java.util.concurrent.TimeUnit.HOURS, // unit for keepAliveTime
+            new java.util.concurrent.LinkedBlockingQueue<>(), // work queue
+            r -> {
+              Thread t = new Thread(r, "RAB-refresh-" + threadCount.getAndIncrement());
+              t.setDaemon(true);
+              return t;
+            });
+    executor.allowCoreThreadTimeOut(true);
+    EXECUTOR = executor;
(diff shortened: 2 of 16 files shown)


google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -78,6 +81,20 @@ final class RegionalAccessBoundaryManager {
...
+  // for easier debugging.
+  private static final AtomicInteger threadCount = new AtomicInteger(0);
+  private static final ExecutorService EXECUTOR =
+      Executors.newFixedThreadPool(
+          5,
+          r -> {
+            Thread t = new Thread(r, "RAB-refresh-" + threadCount.getAndIncrement());
+            t.setDaemon(true);
+            return t;
+          });
```
lqiu96: The tradeoff we are making is between:
1. Executor - Possible memory leak where the executor and the threads cannot be released as the auth library does not have a lifecycle.
2. Thread - Possible unbounded threads created (note: we aim to limit the amount of threads spawning per credential, but it may be possible to spawn an unbounded number of credentials).

I don't think there will be perfect solution. For a middle ground, is it possible to add `allowCoreThreadTimeOut()` and set a keep-alive value of something like ~1ish hour(s)? Single credentials would end up just invoking `new Thread()` in the pool as the RAB call should be idle for the other ~5 hours. If there are multiple credentials that need RAB calls, the executor should be able to use the existing threads as they'll no longer be marked as idle. WDYT?

lqiu96:
@vverman The sdk-platform-java-ci CI issues are not relevant for this PR (existing issues as part of the monorepo migration).

PTAL at the conventialcommit CI complaints: https://github.com/googleapis/google-cloud-java/pull/12867/checks?check_run_id=72739487063

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
vverman: I think this is a good idea to manage memory usage, especially in situations where the application is redeployed. 

In the present implementation, since we don't shutdown the threads within the pool, they would not be garbage collected and the reference would live on while the server doesn't use the old application.


With this thread shutdown suggestion, we avoid that pitfall.


I agree that the threads which are idle for an hour can be timed-out and removed from the pool.

vverman:
Conventional commits addressed.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -78,6 +80,30 @@ final class RegionalAccessBoundaryManager {
...
 
+  // Unbounded thread creation is discouraged in library code to avoid resource
+  // exhaustion. A shared, bounded executor service ensures a hard limit (5)
+  // on concurrent refresh tasks, while threadCount provides unique names
+  // for easier debugging.
+  private static final AtomicInteger threadCount = new AtomicInteger(0);
+  private static final ExecutorService EXECUTOR;
+
+  static {
+    java.util.concurrent.ThreadPoolExecutor executor =
```
lqiu96: nit: Can you use the short names and import the ThreadPoolExecutor?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -78,6 +80,30 @@ final class RegionalAccessBoundaryManager {
...
+            5, // maximumPoolSize: max threads allowed
+            1, // keepAliveTime: time to wait before terminating idle threads
+            java.util.concurrent.TimeUnit.HOURS, // unit for keepAliveTime
+            new java.util.concurrent.LinkedBlockingQueue<>(), // work queue
+            r -> {
+              Thread t = new Thread(r, "RAB-refresh-" + threadCount.getAndIncrement());
+              t.setDaemon(true);
+              return t;
+            });
+    executor.allowCoreThreadTimeOut(true);
```
lqiu96, line 116: nit: Can you add a comment to explain the addition of `allowCoreThreadTimeOut` for future maintainers?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
```diff
@@ -67,10 +67,24 @@ final class RegionalAccessBoundary implements Serializable {
...
+
+  @VisibleForTesting
+  static void enableForTests() {
+    DISABLE_RAB_FOR_TESTS.set(false);
+  }
+
+  @VisibleForTesting
+  static void resetForTests() {
+    DISABLE_RAB_FOR_TESTS.remove();
+  }
```
lqiu96: can you explain this part a bit more? I'm not sure I'm following why this is needed

lqiu96 (approved):
Changes LGTM overall. Added a few comments if you could take a look. I'm not entirely sure what the DISABLE_RAB_TESTS_ENV is for

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
vverman: With this PR, the env variable gate on RAB refresh is removed. Which means on each request call, the RAB refresh will be triggered.

This causes all the tests which test the request headers flow in the auth library to trigger RAB refresh. 

I did spend some time trying to fix this but it is complicated for the following reasons:

1. Tests that verify the header flow (such as getDefaultCredentials_compute_providesToken or tests calling the testUserProvidesToken helper) explicitly call getRequestMetadata to check the bearer token. Because getRequestMetadata now triggers the async RAB refresh on every call, these tests will spawn background requests that interfere with assertions on mock transports.

2. Without this isolation, a background thread started in one test could leak into subsequent tests and interfere with their mocks, leading to flaky tests in CI


Hence we have an injectable disableRABForTests which ensures the tests for the non-RAB flow don't trigger the RAB refresh.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
vverman: Done!

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
vverman: Done!

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
```diff
@@ -172,28 +184,16 @@ public String toString() {
     }
   }
 
-  @VisibleForTesting
-  static void setEnvironmentProviderForTest(@Nullable EnvironmentProvider provider) {
```
lsirac, line 176: This is called in LoggingTest.java which also needs to be updated.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -78,6 +83,32 @@ final class RegionalAccessBoundaryManager {
...
+  private static final ExecutorService EXECUTOR;
+
+  static {
+    ThreadPoolExecutor executor =
+        new ThreadPoolExecutor(
+            5, // corePoolSize: threads to keep alive
+            5, // maximumPoolSize: max threads allowed
+            1, // keepAliveTime: time to wait before terminating idle threads
+            TimeUnit.HOURS, // unit for keepAliveTime
+            new LinkedBlockingQueue<>(), // work queue
```
lsirac: `LinkedBlockingQueue<>()` is unbounded, so a process with many credentials could still build up unlimited pending refreshes.

Can we use a bounded queue here?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
lsirac: But nothing calls `disableForTests()`, `enableForTests()`, or `resetForTests()`?

Also, since this is only disables RAB on the thread that calls it. That may not cover async `getRequestMetadata(...)` tests where the callback runs on an executor thread.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
```diff
@@ -172,28 +184,16 @@ public String toString() {
...
-    String enabled = environmentProvider.getEnv(ENABLE_EXPERIMENT_ENV_VAR);
-    if (enabled == null) {
-      enabled = System.getProperty(ENABLE_EXPERIMENT_ENV_VAR);
-    }
-    if (enabled == null) {
-      return false;
-    }
-    String lowercased = enabled.toLowerCase();
-    return "true".equals(lowercased) || "1".equals(enabled);
+    return !DISABLE_RAB_FOR_TESTS.get();
```
lsirac: Why was this behavior changed to remove the gate? Are you planning on holding this PR until we're ready to launch?

vverman: This PR is to a feature branch so it won't be released till we're ready to launch.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
vverman: Done!

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
vverman: Thanks for the catch. Your secondary comment spurred a discussion which led to consider 2 options for disabling RAB for pre-existing tests:

1. Thread local disable_rab_for_tests with a ContextPropagatingExecutor. While this would centralize our tests, they would make the tests brittle with bad debuggability.

2. Per test disabling of RAB. code intensive but cleaner to debug.

We decided to go with option 2. Additionally incoming tests in the future would have to add the disableRAB for test if their logic checks for getRequestMetadata.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
vverman: Done, we are using a bounded queue of size 100. Added a unit test as well.

vverman:
Added per-test disablement of RAB refresh and bounded queue.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -161,14 +203,7 @@ void triggerAsyncRefresh(
...
       try {
-        // We use new Thread() here instead of
-        // CompletableFuture.runAsync() (which uses ForkJoinPool.commonPool()).
-        // This avoids consuming CPU resources since
-        // The common pool has a small, fixed number of threads designed for
-        // CPU-bound tasks.
-        Thread refreshThread = new Thread(refreshTask, "RAB-refresh-thread");
-        refreshThread.setDaemon(true);
-        refreshThread.start();
+        EXECUTOR.submit(refreshTask);
```
lsirac: If the executor is full, this puts the credential into the normal 15-minute RAB failure cooldown. That seems too harsh: the RAB lookup didn't fail, we just couldn't schedule more local background work right now. Instead, we should just release `refreshFuture` and skip/log this refresh attempt, but not start the 15-minute cooldown.

This is an extreme edge cases (e.g. process has more than 5 active RAB refreshes + 100 queued refreshes at the same time) which can happen if a service is creating many cred instances close together, but worth fixing!

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
```diff
@@ -449,6 +450,9 @@ void getRequestMetadata_hasAccessToken() throws IOException {
     transportFactory.transport.setServiceAccountEmail("SA_CLIENT_EMAIL");
     ComputeEngineCredentials credentials =
         ComputeEngineCredentials.newBuilder().setHttpTransportFactory(transportFactory).build();
+    credentials.regionalAccessBoundaryManager.setCachedRAB(
```
lsirac: This seeds a fake cached RAB, but that isn't the same as disabling RAB. 

We should replace the fake cached RAB setup with a real test-only disable path. 

For non-RAB tests, we want two things: 
1) no background RAB refresh
2) no `x-allowed-locations` header added to request metadata

Adding fake cached RAB only solves the first part, and changes the headers/logs under test.

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/RegionalAccessBoundaryTest.java
```diff
@@ -200,11 +220,72 @@ public void testManagerTriggersRefreshInGracePeriod() throws InterruptedExceptio
...
     assertTrue(
-        "Refresh should have completed and updated the cache within 5 seconds",
-        resultRab != null && newerEncoded.equals(resultRab.getEncodedLocations()));
+        resultRab != null && newerEncoded.equals(resultRab.getEncodedLocations()),
+        "Refresh should have completed and updated the cache within 5 seconds");
     assertEquals(newerEncoded, resultRab.getEncodedLocations());
   }
 
+  @Test
+  public void testExecutorQueueCapacityLimit() throws Exception {
```
lsirac, line 229: This test is brittle because it depends on the shared static executor being completely empty when the test starts. If any previous or parallel test has already submitted RAB refresh work, the queue may not have exactly 105 available slots, so the “106th task” assertion can fail for the wrong reason or pass without proving the intended behavior.

I'd suggest making the executor a test-injectable dep instead of hard-coding the static one in triggerAsyncRefresh.

lsirac (changes requested):

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
lsirac, line 196: This is missing request.setLoggingEnabled(false);

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
vverman: > For non-RAB tests we want ... no x-allowed-locations header added to request metadata

For those test that do check the presence of the headers, they check for the presence of the authorization header. Which is rightly present, thereby not affecting the existing testing logic. 

It is not that the x-allowed-locations header is incorrect, that we're 'disabling RAB' rather it is because async refreshes might non-deterministically consume mocks. That problem is avoided here.

Additionally, this method doesn't add any production logic exclusively for testing.

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/RegionalAccessBoundaryTest.java
vverman: Done!

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
vverman: Thanks for the tip, got rid of it!

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
vverman: Great point, now the logic logs the inability to schedule an RAB refresh and returns without throwing an exception.

The cooldown isn't activated for that credential.

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
lsirac: The downside with your approach is you need to make sure to update all the assertions. There are now 5 log events, not 3, so the test is failing. Please make sure the necessary places are updated.

vverman: Thank you for the suggestion. 

I checked the failure in LoggingTest and found the test which was failing "getRequestMetadata_hasAccessToken" was caused by the lazy loading of the universe domain, triggered by the new RAB check, rather than the RAB refresh itself."

For ComputeEngineCredentials, the first call to getUniverseDomain() fetches it from the metadata server, generating 2 extra logs. Since the RAB check now calls this lazily during getRequestMetadata(), these logs were captured in the test.

So I bumped the count for the test to 5. The rest of the tests weren't updated because they don't trigger a lazy load of the universe domain and aren't at risk of triggering the increased logs.

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
```diff
@@ -445,19 +445,19 @@ void sign_sameAs() throws IOException {
...
   void getRequestMetadata_hasAccessToken() throws IOException {
-    TestAppender testAppender = setupTestLogger(ComputeEngineCredentials.class);
     MockMetadataServerTransportFactory transportFactory = new MockMetadataServerTransportFactory();
     transportFactory.transport.setServiceAccountEmail("SA_CLIENT_EMAIL");
     ComputeEngineCredentials credentials =
         ComputeEngineCredentials.newBuilder().setHttpTransportFactory(transportFactory).build();
     credentials.regionalAccessBoundaryManager.setCachedRAB(
         new RegionalAccessBoundary(
             "dummy-locations", Arrays.asList("dummy-loc"), credentials.clock));
+    TestAppender testAppender = setupTestLogger(ComputeEngineCredentials.class);
```
lsirac: test setup should happen at the beginning of the test - why the change?

lsirac (approved):

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
vverman: Thanks for pointing out! That was a vestige of something I was trying. I got rid of it.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
```diff
@@ -172,29 +164,13 @@ public String toString() {
...
   /**
-   * Checks if the regional access boundary feature is enabled. The feature is enabled if the
-   * environment variable or system property "GOOGLE_AUTH_TRUST_BOUNDARY_ENABLE_EXPERIMENT" is set
-   * to "true" or "1" (case-insensitive).
+   * Checks if the regional access boundary feature is enabled.
+   *
+   * <p>This method is for internal use only and may be changed or removed in future releases.
    *
    * @return True if the regional access boundary feature is enabled, false otherwise.
    */
```
lqiu96: Since we deleted the `isEnabled()` method, can we remove the associated method?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
```diff
@@ -67,10 +65,6 @@ final class RegionalAccessBoundary implements Serializable {
   static final String X_ALLOWED_LOCATIONS_HEADER_KEY = "x-allowed-locations";
   private static final long serialVersionUID = -2428522338274020302L;
 
-  // Note: this is for internal testing use use only.
-  // TODO: Fix unit test mocks so this can be removed
-  // Refer -> https://github.com/googleapis/google-auth-library-java/issues/1898
```
lqiu96, line 72: can we close this ticket created here? Or we want to wait until this is merged into main?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -161,19 +211,15 @@ void triggerAsyncRefresh(
...
       } catch (Exception | Error e) {
         // If scheduling fails (e.g., RejectedExecutionException, OutOfMemoryError for threads),
         // the task's finally block will never execute. We must release the lock here.
-        handleRefreshFailure(
-            new Exception("Regional Access Boundary background refresh failed to schedule", e));
+        LoggingUtils.log(
+            LOGGER_PROVIDER,
+            java.util.logging.Level.WARNING,
+            null,
+            "Regional Access Boundary background refresh failed to schedule: " + e.getMessage());
```
lqiu96: I forgot to ask about this earlier. Do we plan to have any public docs regarding this to explain this behavior? I wonder if something like this may be confusing to the user and they may be asking what RAB is.

If not, can we use something that either has a bit softer wording or adding something that assures this is nothing to worry about

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
```diff
@@ -76,7 +76,7 @@
  * credentials test classes with addition of test logging appender setup and test logic for logging.
  * This duplicates tests setups, but centralizes logging test setup in this class.
  */
-class LoggingTest {
+public class LoggingTest {
```
lqiu96: nit: is this needed? Or just slipped in

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/AwsCredentialsTest.java
```diff
@@ -139,6 +134,9 @@ void refreshAccessToken_withoutServiceAccountImpersonation() throws IOException
             .setHttpTransportFactory(transportFactory)
             .setCredentialSource(buildAwsCredentialSource(transportFactory))
             .build();
+    awsCredential.regionalAccessBoundaryManager.setCachedRAB(
+        new RegionalAccessBoundary(
+            "dummy-locations", Arrays.asList("dummy-loc"), awsCredential.clock));
```
lqiu96: nit: changes make sense. thanks I like this more than the env var idea.

Is possible just to create one constant `RegionalAccessBoundary` object per file?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -161,19 +211,15 @@ void triggerAsyncRefresh(
...
-        refreshThread.start();
+        this.executor.submit(refreshTask);
       } catch (Exception | Error e) {
         // If scheduling fails (e.g., RejectedExecutionException, OutOfMemoryError for threads),
         // the task's finally block will never execute. We must release the lock here.
-        handleRefreshFailure(
-            new Exception("Regional Access Boundary background refresh failed to schedule", e));
+        LoggingUtils.log(
+            LOGGER_PROVIDER,
+            java.util.logging.Level.WARNING,
```
lqiu96: nit: can you use the short name and import instead?

lqiu96 (approved):
added a few small nits. Changes look fine on my end. PTAL at the nits and see if the make sense/ should be addressed. Feel free to merge afterwards

nbayati (approved):

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
vverman: Done!

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
vverman: Done!

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/AwsCredentialsTest.java
vverman: RAB needs the clock from its instantiating credential for testing purposes.

Instead I defined a dummy rab function which takes in the clock, this avoids duplication.

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
vverman: Accidental change, reverted it. Thanks for the catch.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
vverman: I changed the log level to be fine and changed the message to say this will be retried on next request.

We do have plans for public documentation. Not sure of the timeline for those

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundary.java
vverman: Removed the comment. Thanks for the catch.

PR #12876, opened 2026-04-21
chore(spanner): accomodate experimental host in samples

Adarsh-Shrivastava-001:
Add the support for experimental host in sample command.

### DIFF
--- a/java-spanner/samples/snippets/pom.xml
+++ b/java-spanner/samples/snippets/pom.xml
@@ -34,7 +34,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>libraries-bom</artifactId>
-        <version>26.76.0</version>
+        <version>26.79.0</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -228,7 +228,7 @@
           <archive>
             <index>false</index>
             <manifest>
-              <mainClass>com.example.spanner.admin.archived.SpannerSample</mainClass>
+              <mainClass>com.example.spanner.SpannerSample</mainClass>
               <addClasspath>true</addClasspath>
               <classpathPrefix>lib/</classpathPrefix>
             </manifest>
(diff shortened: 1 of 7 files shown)


java-spanner/samples/snippets/pom.xml
```diff
@@ -228,7 +228,7 @@
           <archive>
             <index>false</index>
             <manifest>
-              <mainClass>com.example.spanner.admin.archived.SpannerSample</mainClass>
```
rahul2393, line 231: I don't think this change is intentional?

java-spanner/samples/snippets/src/main/java/com/example/spanner/MutableCredentialsExample.java
```diff
@@ -33,6 +33,7 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
```
rahul2393, line 36: Is this intentional?

rahul2393 (changes requested):
Samples are snippets which customer can copy-paste as it is and run in there application, We should not be using testing helper methods here, also can you please share the use-case of these sample changes are they going to land on docs.cloud.google.com?

java-spanner/samples/snippets/src/main/java/com/example/spanner/MutableCredentialsExample.java
Adarsh-Shrivastava-001: Yes, else this cause compilation failure. essentially its missing is current version of code.

java-spanner/samples/snippets/pom.xml
Adarsh-Shrivastava-001: This change is intentional as we want to pick the newer version the mentioned class. Do you see any problems with this?

Adarsh-Shrivastava-001:
> Samples are snippets which customer can copy-paste as it is and run in there application, We should not be using testing helper methods here, also can you please share the use-case of these sample changes are they going to land on docs.cloud.google.com?

I have created a new helper class and migrated required methods with proper error handling to examples package.

Regarding the use case and docs, that will be a followup as discussed.

PR #12877, opened 2026-04-21
fix(bqjdbc): optimize formatter in BigQueryJdbcRootLogger

Neenu1995:
- Replaced `String.format` with `StringBuilder` and Guava's Strings for faster padding.
- Eliminated `Thread.getAllStackTraces()` (which triggered heavy JVM safepoints) by using `Thread.currentThread()` and `Thread.enumerate()` fallbacks.
- Cached PROCESS_ID instead of parsing it dynamically on every log line.
- Swapped in `ThreadLocal<SimpleDateFormat>` to safely reuse expensive date formatters.

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
@@ -104,6 +104,10 @@ class BigQueryJdbcRootLogger {
         String sourceClassName = record.getLoggerName();
         String sourceMethodName = record.getSourceMethodName();
 
+        // Expected log format: yyyy-MM-dd HH:mm:ss.SSS LEVEL PID --- [THREAD] CLASS METHOD: MESSAGE
+        // Example: 2026-04-22 10:16:00.123  INFO 12345 --- [main   ]
+        // com.google.cloud.bigquery.jdbc.BigQueryConnection connect             : Connection
+        // successful
         StringBuilder sb = new StringBuilder(256);
         sb.append(date)
             .append(" ")
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
@@ -16,8 +16,12 @@
 
 package com.google.cloud.bigquery.jdbc;
 
+import com.google.common.base.Strings;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
```diff
@@ -62,49 +88,39 @@ class BigQueryJdbcRootLogger {
...
-            record.getLevel().getName(),
-            processId,
-            threadName,
-            sourceClassName,
-            sourceMethodName,
-            record.getMessage(),
-            System.lineSeparator());
+
+        StringBuilder sb = new StringBuilder(256);
+        sb.append(date)
```
logachev, line 112: nit: can you add a comment with expected format?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
```diff
@@ -62,49 +88,39 @@ class BigQueryJdbcRootLogger {
...
-                .orElse("");
-        long processId =
-            Long.parseLong(ManagementFactory.getRuntimeMXBean().getName().split("@")[0]);
+        String date = DATE_FORMATTER.get().format(new Date(record.getMillis()));
+
+        long threadId = record.getThreadID();
+        String threadName = getThreadName(threadId);
+
+        if (threadName.length() > MAX_THREAD_NAME_LENGTH) {
+          threadName = threadName.substring(threadName.length() - MAX_THREAD_NAME_LENGTH);
```
logachev, line 101: Is it intentional that we cut start of the name rather than end?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
```diff
@@ -49,6 +49,32 @@ class BigQueryJdbcRootLogger {
...
+    ThreadGroup rootGroup = current.getThreadGroup();
+    while (rootGroup.getParent() != null) {
+      rootGroup = rootGroup.getParent();
+    }
+
+    int count = rootGroup.activeCount();
+    Thread[] threads = new Thread[count * 2];
+    int actualCount = rootGroup.enumerate(threads);
+    for (int i = 0; i < actualCount; i++) {
+      if (threads[i].getId() == threadId) {
```
logachev, line 71: Maybe worth adding local cache for id->name?

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
Neenu1995: My reasoning being, thread identifiers are _typically_ at the end of the thread name.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
Neenu1995: This fallback is an expensive, low-frequency edge case. The memory overhead of maintaining the cache exceeds the performance gains from faster lookups.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
Neenu1995: done.

PR #12899, opened 2026-04-23
feat(bqjdbc): Per connection logs - Add PerConnectionFileHandler

Neenu1995:
- PerConnectionFileHandler.java: Introduced a logging handler mapping logs via InheritableThreadLocal constructs.
- BigQueryJdbcMdc.java: Reduced interface exposure to strict package scopes.
- PerConnectionFileHandlerTest.java: Introduced coverage validations verifying default execution routing.

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
@@ -37,7 +37,7 @@ class PerConnectionFileHandler extends Handler {
   private FileHandler defaultHandler;
 
   PerConnectionFileHandler(String baseLogPath, Level level) {
-    this.baseLogPath = baseLogPath;
+    this.baseLogPath = baseLogPath != null ? baseLogPath : "";
     this.level = level;
 
     try {
@@ -53,7 +53,8 @@ class PerConnectionFileHandler extends Handler {
       this.defaultHandler.setLevel(level);
       this.defaultHandler.setFormatter(BigQueryJdbcRootLogger.getFormatter());
     } catch (IOException e) {
-      System.err.println("Failed to initialize default log file: " + e.getMessage());
+      reportError(
+          "Failed to initialize default log file", e, java.util.logging.ErrorManager.OPEN_FAILURE);
     }
   }
 
@@ -86,8 +87,10 @@ class PerConnectionFileHandler extends Handler {
                   fh.setFormatter(BigQueryJdbcRootLogger.getFormatter());
                   return fh;
                 } catch (IOException e) {
-                  System.err.println(
-                      "Failed to create log file for connection " + id + ": " + e.getMessage());
+                  reportError(
+                      "Failed to create log file for connection " + id,
+                      e,
+                      java.util.logging.ErrorManager.OPEN_FAILURE);
                   return defaultHandler;
                 }
               });
@@ -109,14 +112,18 @@ class PerConnectionFileHandler extends Handler {
   }
 
   @Override
-  @Override
   public void close() throws SecurityException {
     for (FileHandler h : handlers.values()) {
-      try { h.close(); } catch (Exception e) {}
+      try {
+        h.close();
+      } catch (Exception e) {
+      }
     }
     try {
       if (defaultHandler != null) defaultHandler.close();
-    } finally { handlers.clear(); }
+    } finally {
+      handlers.clear();
+    }
   }
 
   public void closeHandler(String connectionId) {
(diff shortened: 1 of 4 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.logging.LogRecord;
+
+/**
+ * Custom logging handler that dynamically creates and routes log records to per-connection specific
+ * log files using the connection ID retrieved from BigQueryJdbcMdc.
+ */
+class PerConnectionFileHandler extends Handler {
+  private final String baseLogPath;
+  private final Level level;
+  private final ConcurrentHashMap<String, FileHandler> handlers = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
Neenu1995: Integration to follow in future PRs.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
```diff
@@ -0,0 +1,137 @@
...
+
+    String connectionId = BigQueryJdbcMdc.getConnectionId();
+    FileHandler handler = defaultHandler;
+
+    if (connectionId != null && !connectionId.isEmpty()) {
+      handler =
+          handlers.computeIfAbsent(
+              connectionId,
+              id -> {
+                try {
```
logachev: Can you move it to a helper function instead of lambda? (And can be merged with setup for defaultHandler)

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
```diff
@@ -0,0 +1,137 @@
...
+                  reportError(
+                      "Failed to create log file for connection " + id,
+                      e,
+                      java.util.logging.ErrorManager.OPEN_FAILURE);
+                  return defaultHandler;
+                }
+              });
+    }
+
+    if (handler != null) {
```
logachev, line 91: Can `handler` be null?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+    try {
+      if (defaultHandler != null) defaultHandler.close();
+    } finally {
+      handlers.clear();
+    }
+  }
+
+  public void closeHandler(String connectionId) {
+    if (connectionId != null) {
```
logachev, line 122: nit: I think this check is not required. If it is `null`, `.remove()` is going to return `null` which is handled.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
```diff
@@ -0,0 +1,137 @@
...
+      this.defaultHandler.setLevel(level);
+      this.defaultHandler.setFormatter(BigQueryJdbcRootLogger.getFormatter());
+    } catch (IOException e) {
+      reportError(
+          "Failed to initialize default log file", e, java.util.logging.ErrorManager.OPEN_FAILURE);
+    }
+  }
+
+  private String getLogFilePath(String id) {
+    String path = baseLogPath;
```
logachev: Can we store `baseLogPath` as a `Path` (maybe using `toAbsolutePath` when initialized and use `baseLogPath.resolve("BigQuery-" + id + ".log")?

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
Neenu1995: Done

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
Neenu1995: Here, handler can be null if the createFileHandler() method fails to create the default handler.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
Neenu1995: The handlers map is a ConcurrentHashMap. If connectionId were null and we passed it directly to handlers.remove(connectionId), it would throw a NullPointerException. So I think it is safer to keep it than throw an exception in log layer.

PR #12900, opened 2026-04-23
chore: add script to generate release notes based on commit history

### LINKED ISSUE #12864: Changelog was not updated in the last release
CHANGELOGs in each module was not updated since the March release. For example, [CHANGELOG.md in java-spanner ](https://github.com/googleapis/google-cloud-java/blob/main/java-spanner/CHANGELOG.md).

This is likely due to the releases (1.83 to 1.85) was manually performed through release-please cli and didn't trigger the [changelog generation logic](https://github.com/googleapis/google-cloud-java/blob/0b6a924544f7f23219e306f7b43fd5b686a8a560/.github/workflows/changelog_generation.yaml#L22-L24).

There are three aspects we want to fix:

1. Manually re-run the changelog generation CI with all the commit history to populate the CHANGELOG for last release.
2. Make sure the changelog generation logic works for recently migrated repos.
3. Make sure the changelog generation logic works for releases performed from release-please cli as well. Or provide a way to manually trigger it.

meltsufin:
The script takes the following inputs:
- module name: as specified in versions.txt
- module directory: path in the monorepo
- version: version as found in versions.txt

It scans backwards through the git history of versions.txt to find the commit where the version was changed to the provided version. It then finds the commit where the previous non-snapshot version was set. It uses this commit range to generate the commit history affecting that directory.

This also adds missing release notes for spanner and bigquery.

Fixes #12864.


java-bigquery/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## [2.65.0](https://github.com/googleapis/google-cloud-java/compare/6bef068b4d8...v1.85.0) (2026-04-13)
+
+### Bug Fixes
+
+* fix(bqjdbc): lazily instantiate Statement in BigQueryDatabaseMetaData (#12752) ([72e5508](https://github.com/googleapis/google-cloud-java/commit/72e5508669ea48cde28f02adfeedfb05cd73fc57))
```
meltsufin, line 7: @logachev Does this look right for Bigquery for filling the missed versions?

java-spanner/CHANGELOG.md
```diff
@@ -1,5 +1,37 @@
 # Changelog
 
+## [6.116.1](https://github.com/googleapis/google-cloud-java/compare/966e8a4d99b...cd6048171fc) (2026-04-21)
```
meltsufin, line 3: @olavloite Does this look like the right release notes for these recent releases?

.github/release-note-generation/generate_module_notes.py
blakeli0, line 1: Do we expect this script to replace the existing script [split_release_note.py](https://github.com/googleapis/google-cloud-java/blob/main/.github/release-note-generation/split_release_note.py) and used in the automation [changelog_generation.yaml](https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/changelog_generation.yaml)

meltsufin: We could potentially, but for now, it's meant just for backfilling.
It has specific logic for dealing with the complex commit history resulting from split repo migrations. That won't be relevant in the future.

blakeli0 (approved):

PR #12904, opened 2026-04-23
ci: dependency check with BUILD_SUBDIR=java-spanner-jdbc

suztomo:


suztomo:
BUILD_SUBDIR is already declared at the workflow level.

PR #12913, opened 2026-04-24
fix(auth): handle null APPDATA environment variable in GoogleAuthUtils.getWellKnownCredentialsFile

### LINKED ISSUE #12565: [google-auth-library-java] NullPointerException when APPDATA environment variable is unset on Windows in GoogleAuthUtils.getWellKnownCredentialsFile
In [GoogleAuthUtils.getWellKnownCredentialsFile](https://github.com/googleapis/google-auth-library-java/blob/0f925933f3e94877ef84fd6c160ad91643f0d4b3/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java#L67), when the OS is detected as Windows and the CLOUDSDK_CONFIG environment variable is not set, the code attempts to resolve the gcloud configuration directory using the APPDATA environment variable. If the APPDATA environment variable is missing, provider.getEnv("APPDATA") returns null.

Passing null to the new File(String) constructor instantly throws a raw NullPointerException. This causes the library to crash with an unhelpful stack overflow or server runtime crash, instead of failing gracefully with a FileNotFoundException or IOException.

anishesg:
In `GoogleAuthUtils.getWellKnownCredentialsFile()`, on Windows systems where `CLOUDSDK_CONFIG` is unset, the code calls `provider.getEnv("APPDATA")` and passes the result directly to `new File(String)`. When `APPDATA` is not present in the environment, `getEnv()` returns `null`, and `new File(null)` immediately throws a raw `NullPointerException` with no useful context.

The fix adds an explicit null check after retrieving the `APPDATA` value. When it is null, the method now throws an `UncheckedIOException` wrapping a `FileNotFoundException` with a descriptive message indicating that `APPDATA` is not set, rather than propagating an opaque `NullPointerException`. This is consistent with the library's overall approach of surfacing meaningful errors when credential paths cannot be resolved.

A regression test in `GoogleAuthUtilsTest` verifies that the new exception type and message are produced when `APPDATA` is absent on a simulated Windows environment.

Fixes #12565

### DIFF
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
@@ -32,6 +32,8 @@
 package com.google.auth.oauth2;
 
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.UncheckedIOException;
 
 /**
  * This public class provides shared utilities for common OAuth2 utils or ADC. It also exposes
@@ -70,7 +72,14 @@ public class GoogleAuthUtils {
     if (envPath != null) {
       cloudConfigPath = new File(envPath);
     } else if (provider.getOsName().indexOf("windows") >= 0) {
-      File appDataPath = new File(provider.getEnv("APPDATA"));
+      String appDataEnv = provider.getEnv("APPDATA");
+      if (appDataEnv == null) {
+        throw new UncheckedIOException(
+            new FileNotFoundException(
+                "APPDATA environment variable is not set; cannot locate the well-known"
+                    + " credentials file on Windows."));
+      }
+      File appDataPath = new File(appDataEnv);
       cloudConfigPath = new File(appDataPath, provider.CLOUDSDK_CONFIG_DIRECTORY);
     } else {
       File configPath = new File(provider.getProperty("user.home", ""), ".config");
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
@@ -73,10 +73,10 @@ public class GoogleAuthUtils {
       cloudConfigPath = new File(envPath);
     } else if (provider.getOsName().indexOf("windows") >= 0) {
       String appDataEnv = provider.getEnv("APPDATA");
-      if (appDataEnv == null) {
+      if (appDataEnv == null || appDataEnv.trim().isEmpty()) {
         throw new UncheckedIOException(
             new FileNotFoundException(
-                "APPDATA environment variable is not set; cannot locate the well-known"
+                "APPDATA environment variable is not set or empty; cannot locate the well-known"
                     + " credentials file on Windows."));
       }
       File appDataPath = new File(appDataEnv);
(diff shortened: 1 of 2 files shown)


google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -70,7 +72,14 @@ static final File getWellKnownCredentialsFile(DefaultCredentialsProvider provide
...
       cloudConfigPath = new File(envPath);
     } else if (provider.getOsName().indexOf("windows") >= 0) {
-      File appDataPath = new File(provider.getEnv("APPDATA"));
+      String appDataEnv = provider.getEnv("APPDATA");
+      if (appDataEnv == null) {
+        throw new UncheckedIOException(
+            new FileNotFoundException(
+                "APPDATA environment variable is not set; cannot locate the well-known"
+                    + " credentials file on Windows."));
+      }
```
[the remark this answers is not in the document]
anishesg: Good catch — updated the check to `appDataEnv == null || appDataEnv.trim().isEmpty()` and adjusted the error message accordingly. Also added tests for the empty-string and blank-string cases.

anishesg:
Friendly ping — all tests pass, just waiting on the multi-approver requirement. Let me know if there's anything else needed on this one.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -70,7 +72,14 @@ static final File getWellKnownCredentialsFile(DefaultCredentialsProvider provide
     if (envPath != null) {
       cloudConfigPath = new File(envPath);
     } else if (provider.getOsName().indexOf("windows") >= 0) {
-      File appDataPath = new File(provider.getEnv("APPDATA"));
+      String appDataEnv = provider.getEnv("APPDATA");
+      if (appDataEnv == null || appDataEnv.trim().isEmpty()) {
```
lqiu96, line 76: nit: Guava has [Strings.isNullOrEmpty](https://www.javadoc.io/doc/com.google.guava/guava/32.0.0-jre/com/google/common/base/Strings.html#isNullOrEmpty(java.lang.String)) for a convenient API

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -70,7 +72,14 @@ static final File getWellKnownCredentialsFile(DefaultCredentialsProvider provide
...
+        throw new UncheckedIOException(
+            new FileNotFoundException(
+                "APPDATA environment variable is not set or empty; cannot locate the well-known"
```
lqiu96, line 79: Can you update the message to be something like `... is not set or is empty` for clarity?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
nbayati: I think this check would reject if the env var was set to a string like "  ", which is probably more robust than just checking for Null or Empty. If we decide to keep it this way, we might want to update it in[ here](https://github.com/googleapis/google-cloud-java/blob/85f51b299e187645add14d39fa38a738df9b3574/google-auth-library-java/oauth2_http/java/com/google/auth/mtls/MtlsUtils.java#L127) too, so that our checks for `APPDATA` are consistent.

nbayati (approved):

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
lqiu96: From a quick search, I don't think `" "` is a valid in Windows and probably not realistic (IIUC `APPDATA` is used throughout Windows and should be valid). Since the other location already uses Guava, let's just make it consistent and use Guava as well (both impls resolve the main NPE possibility and are fine with me).

anishesg:
Friendly ping — this was approved by @nbayati but has picked up merge conflicts. I'll rebase shortly to resolve them. Is there a second approver who can help clear the multi-approvers gate?

PR #12918, opened 2026-04-24
feat(bqjdbc): extend OpenTelemetry instrumentation for metadata and pagination

keshavdandeva:
b/491245568

#### Core Instrumentation Logic
*   **Database Metadata Tracing**: Added OTel spans to key methods in `BigQueryDatabaseMetaData.java` (`getCatalogs`, `getSchemas`, `getTables`, `getColumns`) to capture underlying API calls.
*   **Pagination Span Links**: Captured the parent span context at the start of `fetchNextPages` in `BigQueryStatement.java` and linked background pagination spans back to it, avoiding timeline anomalies.
*   **Cross-Thread Context Propagation**: Stored the `SpanContext` in `BigQueryBaseResultSet.java` at creation time and made it current during `next()` in `BigQueryJsonResultSet.java` and `BigQueryArrowResultSet.java` to survive thread hops.
*   **Tracer Reuse**: Extracted `getSafeTracer` to `BigQueryJdbcOpenTelemetry.java` as a static utility to ensure consistent fallback behavior across the driver.
*   **Lambda Extraction**: Extracted the large lambda function in `populateArrowBufferedQueue` in `BigQueryStatement.java` to its own private method `processArrowStream` to improve readability and maintainability.

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
@@ -216,68 +216,69 @@ class BigQueryArrowResultSet extends BigQueryBaseResultSet {
   @Override
   public boolean next() throws SQLException {
     checkClosed();
-    try (Scope scope = Context.current().with(Span.wrap(originalSpanContext)).makeCurrent()) {
-      if (this.isNested) {
-        if (this.currentNestedBatch == null || this.currentNestedBatch.getNestedRecords() == null) {
-          throw new IllegalStateException(
-              "currentNestedBatch/JsonStringArrayList can not be null working with the nested record");
-        }
-        if (this.nestedRowIndex < (this.toIndexExclusive - 1)) {
-          /* Check if there's a next record in the array which can be read */
-          this.nestedRowIndex++;
-          return true;
-        }
+    if (this.isNested) {
+      if (this.currentNestedBatch == null || this.currentNestedBatch.getNestedRecords() == null) {
+        throw new IllegalStateException(
+            "currentNestedBatch/JsonStringArrayList can not be null working with the nested record");
+      }
+      if (this.nestedRowIndex < (this.toIndexExclusive - 1)) {
+        /* Check if there's a next record in the array which can be read */
+        this.nestedRowIndex++;
+        return true;
+      }
+      this.afterLast = true;
+      return false;
+    } else {
+      /* Non nested */
+      if (this.hasReachedEnd || this.isLast()) {
         this.afterLast = true;
         return false;
-      } else {
-        /* Non nested */
-        if (this.hasReachedEnd || this.isLast()) {
-          this.afterLast = true;
-          return false;
-        }
-        try {
-          if (this.currentBatchRowIndex == -1
-              || this.currentBatchRowIndex == (this.vectorSchemaRoot.getRowCount() - 1)) {
-            /* Start of iteration or we have exhausted the current batch */
-            // Advance the cursor. Potentially blocking operation.
-            BigQueryArrowBatchWrapper batchWrapper = this.buffer.take();
-            if (batchWrapper.getException() != null) {
-              throw new BigQueryJdbcRuntimeException(batchWrapper.getException());
-            }
-            if (batchWrapper.isLast()) {
-              /* Marks the end of the records */
-              if (this.vectorSchemaRoot != null) {
-                // IMP: To avoid memory leak: clear vectorSchemaRoot as it still holds
-                // the last batch
-                this.vectorSchemaRoot.clear();
-              }
-              this.hasReachedEnd = true;
-              this.rowCount++;
-              return false;
-            }
-            // Valid batch, process it
-            ArrowRecordBatch arrowBatch = batchWrapper.getCurrentArrowBatch();
-            // Populates vectorSchemaRoot
-            this.arrowDeserializer.deserializeArrowBatch(arrowBatch);
-            // Pointing to the first row in this fresh batch
-            this.currentBatchRowIndex = 0;
-            this.rowCount++;
-            return true;
-          }
-          // There are rows left in the current batch.
-          else if (this.currentBatchRowIndex < this.vectorSchemaRoot.getRowCount()) {
-            this.currentBatchRowIndex++;
-            this.rowCount++;
-            return true;
-          }
-        } catch (InterruptedException | SQLException ex) {
-          throw new BigQueryJdbcException(
-              "Error occurred while advancing the cursor. This could happen when connection is closed while the next method is being called.",
-              ex);
-        }
       }
-      return false;
+      try {
+        if (this.currentBatchRowIndex == -1
+            || this.currentBatchRowIndex == (this.vectorSchemaRoot.getRowCount() - 1)) {
+          /* Start of iteration or we have exhausted the current batch */
+          // Advance the cursor. Potentially blocking operation.
+          BigQueryArrowBatchWrapper batchWrapper;
+          try (Scope scope = Context.current().with(Span.wrap(originalSpanContext)).makeCurrent()) {
+            batchWrapper = this.buffer.take();
+          }
+          if (batchWrapper.getException() != null) {
+            throw new BigQueryJdbcRuntimeException(batchWrapper.getException());
+          }
+          if (batchWrapper.isLast()) {
+            /* Marks the end of the records */
+            if (this.vectorSchemaRoot != null) {
+              // IMP: To avoid memory leak: clear vectorSchemaRoot as it still holds
+              // the last batch
+              this.vectorSchemaRoot.clear();
+            }
+            this.hasReachedEnd = true;
+            this.rowCount++;
+            return false;
+          }
+          // Valid batch, process it
+          ArrowRecordBatch arrowBatch = batchWrapper.getCurrentArrowBatch();
+          // Populates vectorSchemaRoot
+          this.arrowDeserializer.deserializeArrowBatch(arrowBatch);
+          // Pointing to the first row in this fresh batch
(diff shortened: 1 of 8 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
```diff
@@ -236,7 +239,10 @@ public boolean next() throws SQLException {
             || this.currentBatchRowIndex == (this.vectorSchemaRoot.getRowCount() - 1)) {
           /* Start of iteration or we have exhausted the current batch */
           // Advance the cursor. Potentially blocking operation.
-          BigQueryArrowBatchWrapper batchWrapper = this.buffer.take();
+          BigQueryArrowBatchWrapper batchWrapper;
+          try (Scope scope = Context.current().with(Span.wrap(originalSpanContext)).makeCurrent()) {
```
logachev: Helper method to get context?

logachev: Why not including entire block in the span? Line 261 performs deserialization, so it's still part of the operation

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJsonResultSet.java
```diff
@@ -158,7 +161,9 @@ public boolean next() throws SQLException {
       }
       try {
         // Advance the cursor,Potentially blocking operation
-        this.cursor = this.buffer.take();
+        try (Scope scope = Context.current().with(Span.wrap(originalSpanContext)).makeCurrent()) {
```
logachev: Same as for Arrow

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
```diff
@@ -49,4 +49,12 @@ public static OpenTelemetry getOpenTelemetry(
   public static Tracer getTracer(OpenTelemetry openTelemetry) {
     return openTelemetry.getTracer(INSTRUMENTATION_SCOPE_NAME);
   }
+
+  /** Gets a Tracer for the JDBC driver, fallback to noop if connection has no tracer. */
+  public static Tracer getSafeTracer(BigQueryConnection connection) {
+    if (connection != null && connection.getTracer() != null) {
```
logachev: Any reason why `connection` can't set the same default value on init? So everywhere we can use `connection.getTracer()`

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -2096,102 +2139,119 @@ public ResultSet getColumns(
...
-              DatasetId datasetId = dataset.getDatasetId();
-              LOG.info("Processing dataset: " + datasetId.getDataset());
+          Span backgroundSpan =
+              tracer
+                  .spanBuilder("BigQueryDatabaseMetaData.getColumns.background")
+                  .setNoParent()
+                  .addLink(parentSpanContext)
+                  .startSpan();
+
+          try (Scope scope = backgroundSpan.makeCurrent()) {
```
logachev, line 2153: Can we move `startSpan()` there too? So no need for manual cleanup at the end

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
keshavdandeva: Yeah so, earlier I did that but then Gemini review said:

> The next() method is a high-frequency operation in JDBC, often called millions of times during result set iteration. Wrapping the entire method body in an OpenTelemetry Scope introduces significant overhead due to ThreadLocal access and object allocations for every single row. This is particularly inefficient when next() is simply incrementing an index for rows already available in memory (e.g., when isNested is true or when iterating within the current Arrow batch). Consider moving the context propagation to only the specific sections where blocking operations or external API calls occur, such as the block starting at line 243 where buffer.take() is called. Ensure that the scope is managed in an exception-safe manner (e.g., using try-with-resources) to prevent resource leaks.

But I guess, for Arrow, as this is batch of rows, maybe its fine

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJsonResultSet.java
keshavdandeva: Same comment from Gemini about this as well:

> Similar to the Arrow implementation, wrapping the entire next() method in a Scope creates substantial performance overhead for every row processed. Since next() is a hot path, the cost of managing the OpenTelemetry context for every iteration can lead to a noticeable regression in throughput for large result sets. It is recommended to limit the scope of context propagation to the parts of the method that actually perform work requiring context, such as the blocking buffer.take() call. Ensure the scope is closed in an exception-safe manner to prevent leaks.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
keshavdandeva: Done

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
keshavdandeva: Yeah, initially I did not have a no-op tracer upon initialization, but now we can simply do `connection.getTracer()` and remove `getSafeTracer`

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
keshavdandeva: I looked into this, and unfortunately, the OTel does not make the Span interface AutoCloseable. Only the Scope returned by `span.makeCurrent()` is AutoCloseable to handle thread-local cleanup. That is why we are forced to manually call `span.end()` in the finally block

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJsonResultSet.java
```diff
@@ -158,7 +159,9 @@ public boolean next() throws SQLException {
       }
       try {
         // Advance the cursor,Potentially blocking operation
-        this.cursor = this.buffer.take();
+        try (Scope scope = makeOriginalContextCurrent()) {
```
logachev: Can re-use above `try/catch` block

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
logachev: I think we can still expand it to the entire if block (lines 237-268), it'd include batch de-serializing (which is kind of part of "read the next batch" operation)

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJsonResultSet.java
logachev: Can expand to the try block (162-180), there is not much else happening either way

logachev: Actually in this version, `this.cursor = this.buffer.take()` happens on every call to `next()`, so it'd already create new span each time. Should we check performance & see if it is needed? All we do here is taking an item from the queue.. So maybe no real tracing in `next` function needed. (in arrow tracing makes sense because we're reading next block in the same thread.. So maybe we can squeeze more from Arrow by reading in the background thread)

logachev (approved):

logachev:
Please use `bqjdbc` in the title, so it will be easier to find PRs in the future (e.g. spanner has their own jdbc in that repo)

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJsonResultSet.java
keshavdandeva: Yeah, I guess you are right. Since `this.buffer.take()` is called for every row in and it doesn't trigger any new spans or heavy processing (unlike Arrow where we deserialize a whole batch), maintaining the scope here only adds overhead without any _real_ benefit. 
I will remove it for now, and create a task to investigate performance with and without to finalise it later

Neenu1995 (approved):

PR #12920, opened 2026-04-24
fix(bqjdbc): Log exception messages - part 3

Neenu1995:
b/505826451

**1. Centralization of Exception Logging**

Moved the responsibility of logging exceptions from the throw sites to the constructors of the custom exception classes themselves. This ensures all instances are logged automatically and consistently.

Added loggers and logging calls to constructors of the following custom exception classes:

- BigQueryJdbcRuntimeException
- BigQueryJdbcException
- BigQueryConversionException
- BigQueryJdbcCoercionException
- BigQueryJdbcCoercionNotFoundException
- BigQueryJdbcSqlFeatureNotSupportedException
- BigQueryJdbcSqlSyntaxErrorException

**2. Updates to BigQueryJdbcCustomLogger**
To support the above changes, the custom logger was updated:

- Modified caller inference to skip classes in the com.google.cloud.bigquery.exception package. This ensures that the logs point to the method that threw the exception rather than the exception constructor itself.
- Made the class, its constructor, and severe methods public so they can be accessed by exception classes residing in a different package.

**3. Cleanup of Redundant Logging**
After centralizing the logging, removed manual LOG.severe calls that were followed by a throw of the updated custom exceptions in functional classes.

Files cleaned up:

- BigQueryConnection.java
- BigQueryStatement.java
- BigQueryPreparedStatement.java
- BigQueryJdbcOAuthUtility.java
- BigQueryArrowResultSet.java
- BigQueryJsonResultSet.java
- BigQueryBaseResultSet.java
- BigQueryTypeCoercer.java
- BigQueryJdbcTypeMappings.java
- PooledConnectionDataSource.java

**4. Addition of Logging for Standard Exceptions**
For standard Java exceptions (like IllegalArgumentException and IllegalStateException) that do not log themselves, added manual severe logging before they are thrown to ensure they are not missed in the logs.

Files updated: 

- BigQueryConnection.java
- BigQueryCallableStatement.java
- BigQueryStatement.java.

**5. Logger Signature Refactor** 
Updated BigQueryJdbcCustomLogger.java to take Throwable as the second parameter in severe methods, and updated calls across 17 files to match this signature.


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
```diff
@@ -108,7 +108,10 @@ private BigQueryArrowResultSet(
       try {
         this.arrowDeserializer = new ArrowDeserializer(arrowSchema);
       } catch (IOException ex) {
-        throw new BigQueryJdbcException(ex);
+        BigQueryJdbcException e =
+            new BigQueryJdbcException("IOException during ArrowDeserializer creation", ex);
+        LOG.severe(e, e.getMessage());
```
logachev: When we get Mdc working, can this LOG be a part of the exception? At least for exceptions that we own they could log sever errors, so we'll never forget to log it.

Neenu1995: Done.

Neenu1995:
> When we get Mdc working, can this LOG be a part of the exception? At least for exceptions that we own they could log sever errors, so we'll never forget to log it.
 

I updated the exception message boiler plate with the existing infra. After MDC impl is complete. I will update this loggers to match.

keshavdandeva (approved):

logachev (approved):

PR #12933, opened 2026-04-25
feat(bqjdbc): Integrate the PerConnectionFileHandler with BigQueryJdbcRootLogger

Neenu1995:
- **BigQueryConnection.java:**
   - Integrated the MdcCloseable pattern with try-with-resources across all methods requiring MDC context.[b/501081433]
   - Extracted rollbackImpl() and closeImpl() to separate business logic from logging setup.
   - Added connectionId generation in the constructor.[b/501081158] 
- **BigQueryStatement.java:**
   - Refactored public JDBC execution methods to use MdcCloseable and extracted Impl methods (like executeQueryImpl).[b/501081433]
- **BigQueryJdbcMdc.java:**
   - Added the MdcCloseable interface to support the try-with-resources pattern.
   - Update the Log Record format to inject the Connection-ID tag.[b/501081558] 
- **BigQueryJdbcRootLogger.java:**
   - Integrated PerConnectionFileHandler for log isolation.[b/501081671]


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
```diff
@@ -104,13 +101,13 @@ public String format(LogRecord record) {
         String sourceClassName = record.getLoggerName();
         String sourceMethodName = record.getSourceMethodName();
 
-        // Expected log format: yyyy-MM-dd HH:mm:ss.SSS LEVEL PID --- [THREAD] CLASS METHOD: MESSAGE
-        // Example: 2026-04-22 10:16:00.123  INFO 12345 --- [main   ]
-        // com.google.cloud.bigquery.jdbc.BigQueryConnection connect             : Connection
-        // successful
+        // Expected log format: yyyy-MM-dd HH:mm:ss.SSS [CONNECTION_ID] LEVEL PID --- [THREAD] CLASS
```
logachev, line 104: Why do we need Connection_ID? Isn't it 1 file per connection now?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -378,31 +411,49 @@ public Statement createStatement(int resultSetType, int resultSetConcurrency)
...
+    try {
+      BigQueryJdbcMdc.registerInstance(this, this.connectionId);
+      LOG.finest("++enter++");
+      checkClosed();
+      PreparedStatement currentStatement = new BigQueryPreparedStatement(this, sql);
+      LOG.fine("Prepared Statement %s created.", currentStatement);
+      addOpenStatements(currentStatement);
+      return currentStatement;
+    } finally {
+      BigQueryJdbcMdc.clear();
```
logachev: Can we make `registerInstance()` to return AutoCloseable() object so we don't need to call `clear()` every time?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -706,102 +782,125 @@ public void clearWarnings() {
 
   @Override
   public boolean getAutoCommit() {
-    checkClosed();
-    return this.autoCommit;
+    try {
```
logachev: I'd suggest using the same approach from Telemetry.

Moving real logic to the private `Impl()` so it is easier to distinguish business logic vs logging setup\tracer setup etc.

Now multiple places have try/catch/finally blocks inside another try\catch\finally and that's difficult to read.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
Neenu1995: Yes. But I figured connection ID in the logs will be useful when sharing log snippets.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
Neenu1995: Done

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
Neenu1995: Done.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -868,12 +946,14 @@ public boolean isClosed() {
 
   private void checkClosed() {
     if (isClosed()) {
+      LOG.severe("This " + getClass().getName() + " has been closed");
```
logachev: nit: not very important for such one-off use-cases, but it's better to use formatted string option, so computation is skipped with disabled loggers

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -77,4 +78,14 @@ static void clear() {
       local.remove();
     }
   }
+
+  /**
+   * Functional interface that extends AutoCloseable to avoid throwing checked exceptions in
+   * try-with-resources.
+   */
+  @FunctionalInterface
```
logachev, line 86: Can you please add a unit-test validating that `clear()` is properly called as expected?

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
Neenu1995: done.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
Neenu1995: done.

PR #12941, opened 2026-04-27
chore(deps): update netty to 4.1.132.Final

jinseopkim0:
b/504629399


blakeli0:
`dependencies.properties` is only used by self-service flows. I think we need to first reproduce the issue by identifying where netty is coming from (likely grpc). And see if it is something we need to fix in our repo or in our dependencies' repo.

jinseopkim0:
> `dependencies.properties` is only used by self-service flows. I think we need to first reproduce the issue by identifying where netty is coming from (likely grpc). And see if it is something we need to fix in our repo or in our dependencies' repo.

Thanks for the feedback. I've left an update on b/504629399#comment6 regarding this.

sdk-platform-java/java-shared-dependencies/third-party-dependencies/pom.xml
```diff
@@ -262,6 +263,13 @@
         <artifactId>tink</artifactId>
         <version>${com.google.crypto.tink.version}</version>
       </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
```
blakeli0: Can you run `mvn dependency:tree` against a locally built client library to confirm that netty is upgraded?

blakeli0: I don't think third-party-dependencies are included in the libraries-bom, so this may not work.

jinseopkim0: Thanks for the feedback. I ran this:
```
jinseop@jinseop:/workspace/google-cloud-java$ mvn dependency:tree -pl :google-cloud-bigquery-jdbc -Dverbose -Dincludes=io.netty
...
[INFO] --- dependency:3.8.1:tree (default-cli) @ google-cloud-bigquery-jdbc ---
[INFO] com.google.cloud:google-cloud-bigquery-jdbc:jar:0.9.0
[INFO] \- com.google.cloud:google-cloud-bigquery:jar:2.65.0:compile
[INFO]    +- io.netty:netty-buffer:jar:4.1.132.Final:compile (version managed from 4.1.110.Final)
[INFO]    \- io.netty:netty-common:jar:4.1.132.Final:compile (version managed from 4.1.110.Final)
...
```
This confirms that the override correctly patches unshaded Netty paths (like the one from Apache Arrow in BigQuery) that are not controlled by the gRPC version. In addition to this PR, we can upgrade grpc to v1.81.0 (see b/504629399#comment10, https://github.com/grpc/grpc-java/pull/12779) to address the shaded transport version. Note that this version has the netty version we need: https://github.com/grpc/grpc-java/blob/4813c6d06089d49a479f7f1786297dc3d8c0e549/gradle/libs.versions.toml#L104)

blakeli0: Thanks! In general, I don't think we want to manage netty versions in our bom because we don't use it directly. If grpc-java is going to upgrade it soon, we should wait for it and upgrade grpc-java instead. 

And we need a separate solution for bigquery since it gets netty from arrow not grpc-java. Maybe manually overriding it in bigquery for now if the latest arrow does not have it.

jinseopkim0: Thanks for the feedback. I've updated the PR to override in bigquery only as suggested.

PR #12953, opened 2026-04-29
deps: upgrade opentelemetry

jinseopkim0:
b/507025240


google-cloud-pom-parent/pom.xml
```diff
@@ -120,6 +120,9 @@
                           <exclude>io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi</exclude>
                           <exclude>io.opentelemetry:opentelemetry-sdk-common</exclude>
                           <exclude>io.opentelemetry:opentelemetry-sdk</exclude>
+                          <exclude>com.fasterxml.jackson.core:jackson-core</exclude>
```
blakeli0: I don't think we want to exclude more things here. And actually we should remove the above exclusions for opentelemetry as well, because it was a temp solution originally.

sdk-platform-java/java-shared-dependencies/third-party-dependencies/pom.xml
```diff
@@ -36,14 +36,14 @@
     <!-- ensure checker-qual version matches what Guava uses -->
     <checker-qual.version>3.49.0</checker-qual.version>
     <perfmark-api.version>0.27.0</perfmark-api.version>
-    <google.cloud.opentelemetry.version>0.33.0</google.cloud.opentelemetry.version>
-    <opentelemetry-grpc-instrumentation.version>2.1.0-alpha</opentelemetry-grpc-instrumentation.version>
-    <opentelemetry-semconv.version>1.29.0-alpha</opentelemetry-semconv.version>
+    <google.cloud.opentelemetry.version>0.36.0</google.cloud.opentelemetry.version>
```
blakeli0: Why we need to upgrade these opentelemetry versions as well? Because they don't work with 1.57.0 anymore?

google-cloud-pom-parent/pom.xml
jinseopkim0: Thanks for the feedback. I've removed them as suggested.

sdk-platform-java/java-shared-dependencies/third-party-dependencies/pom.xml
jinseopkim0: I initially thought this would be good-to-have. But I don't think they're necessary. Removed them.

blakeli0 (approved):

PR #12957, opened 2026-04-29
chore(bqjdbc): switch connection identifier to UUID

keshavdandeva:
b/507856382

#### `BigQueryConnection`
- Generates a full `UUID.randomUUID().toString()` as the `connectionId`.

#### `BigQueryJdbcMdc`
- Prepends `BQ-JDBC-` to the full `UUID` for thread context logging.

#### `PerConnectionFileHandler`
- Extracts `UUID` from `BQ-JDBC-` prefix and constructs log filenames with the format: `BQ-JDBC-yyyyMMddHHmmss-<first-4-chars-of-UUID>.log`.

#### Tests
- Updated `BigQueryJdbcMdcTest` to match the new prefix.
- Refactored `PerConnectionFileHandlerTest` to use a helper method for finding dynamic log files, making tests cleaner.


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -16,16 +16,15 @@
...
+import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * Lightweight MDC implementation for the BigQuery JDBC driver using InheritableThreadLocal.
  * Allocates a dedicated, independent InheritableThreadLocal object per concrete BigQueryConnection
  * instance.
  */
 class BigQueryJdbcMdc {
```
logachev, line 27: iirc, `Connection` itself generates ID now too.

Neenu1995: Yes. Connection generates the actual ID. This is a fallback in case it is absent.

keshavdandeva: Updated the logic there as well

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -35,15 +36,20 @@ class BigQueryJdbcMdc {
...
     if (connection != null) {
       String cleanId =
           instanceIds.computeIfAbsent(
               connection,
               k -> {
-                String suffix =
-                    (id != null && !id.isEmpty()) ? id : String.valueOf(nextId.getAndIncrement());
-                return "JdbcConnection-" + suffix;
+                String baseId = (id != null && !id.isEmpty()) ? id : generateConnectionId();
+                return "JdbcConnection-" + baseId;
```
Neenu1995: Won't this result in a file name like `BigQuery-JdbcConnection-yyyyMMdd-HHmmSS-12345678`. Isn't that a bit too long?
 Also the connection ID looks like this- `JdbcConnection-yyyyMMdd-HHmmSS-12345678`. Too long?

keshavdandeva: Yeah, so it seems like currently it is:
- Connection ID: `yyyyMMdd-HHmmSS-12345678`
- Connection ID in logs: `JdbcConnection-yyyyMMdd-HHmmSS-12345678`
- Log Filenames: `BigQuery-JdbcConnection-yyyyMMdd-HHmmSS-12345678`

How do you suggest we reduce it? @Neenu1995 @logachev

Neenu1995 (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -41,9 +40,8 @@ static MdcCloseable registerInstance(BigQueryConnection connection, String id) {
           instanceIds.computeIfAbsent(
               connection,
               k -> {
-                String suffix =
-                    (id != null && !id.isEmpty()) ? id : String.valueOf(nextId.getAndIncrement());
-                return "JdbcConnection-" + suffix;
+                String baseId = (id != null && !id.isEmpty()) ? id : UUID.randomUUID().toString();
+                return "BQ-JDBC-" + baseId;
```
logachev: nit: don't think we need `BQ-JDBC-` or any other prefix.. They're internal and only relevant to stitch together telemetry/logs. Can be just a uuid

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/PerConnectionFileHandler.java
```diff
@@ -53,7 +55,13 @@ class PerConnectionFileHandler extends Handler {
   }
 
   private String getLogFilePath(String id) {
-    return baseLogPath.resolve("BigQuery-" + id + ".log").toString();
+    String uuid = id;
+    if (id.startsWith("BQ-JDBC-")) {
+      uuid = id.substring("BQ-JDBC-".length());
+    }
+    String timestamp = LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyyMMddHHmmss"));
+    String shortUuid = uuid.length() >= 4 ? uuid.substring(0, 4) : uuid;
```
logachev, line 63: nit: no need for that check, uuid will never be shorter than 4

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
Neenu1995: The user application many have logs generated from other sources. For the file name I think it is useful if we keep the BQ-JDBC prefix. Same with the connection ID, if the user Application is logging from multiple sources , it will be helpful to identify the logs generated by the JDBC Driver.

logachev: Yes, but that's the internal ID, not the filename

PR #12967, opened 2026-04-30
fix(bqjdbc): support Service Account Impersonation in ADC

Neenu1995:
b/506258278

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOAuthUtility.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOAuthUtility.java
@@ -230,27 +230,22 @@ final class BigQueryJdbcOAuthUtility {
         break;
     }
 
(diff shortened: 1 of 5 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOAuthUtility.java
```diff
@@ -232,7 +232,8 @@ static Map<String, String> parseOAuthProperties(DataSource ds, String callerClas
 
     if (authType == AuthType.GOOGLE_SERVICE_ACCOUNT
```
logachev: Maybe remove this entirely? Don't see a reason not to support it for external auth as well

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOAuthUtility.java
Neenu1995: done

PR #12975, opened 2026-04-30
docs(logging): update quickstart guidance

### LINKED ISSUE #11924: [java-logging] Sample: update quickstart snippet
Improve the Logging quickstart [in Java](https://cloud.google.com/logging/docs/samples/logging-quickstart#logging_quickstart-java). Quickstart should work out of the box and contain all the instructions (in comments) so users can immediately see their logs. 

Additional criteria: 
- [ ] include comments on various [ways to authenticate ](https://cloud.google.com/docs/authentication)
- [ ] encourage users to log with standard log wrapper [Logback](https://github.com/googleapis/java-logging-logback) rather than directly writing [LogEntries](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry). Context: In an ideal world, customers should be able to use existing logging frameworks native to their language. With some simple appender/handler to also ship logs to Cloud Logging. Customers aren't locked into logging library syntax, thus it's easier to both onboarding and offboard off of these client libraries. 
- [ ] comment that if users need more customizable logs, redirect to [writing log entries](https://cloud.google.com/logging/docs/samples/logging-write-log-entry)
- [ ] if applicable, a test for this sample

MukundaKatta:
Fixes #11924.

Updates the Java Cloud Logging quickstart to:
- include authentication guidance in the sample comments
- recommend standard logging framework integration, specifically Logback, for production use
- point users who need custom LogEntry fields to the WriteLogEntry sample
- provide a default log name so the sample can run without command-line arguments

Verification:
- git diff --check

Not run locally:
- mvn test (mvn is not installed in this environment)

### DIFF
--- a/java-logging/samples/snippets/src/main/java/com/example/logging/QuickstartSample.java
+++ b/java-logging/samples/snippets/src/main/java/com/example/logging/QuickstartSample.java
@@ -28,13 +28,15 @@ import java.util.Collections;
 /**
  * This sample demonstrates writing logs directly with the Cloud Logging API.
  *
- * <p>Authenticate by running `gcloud auth application-default login`, by setting the
- * GOOGLE_APPLICATION_CREDENTIALS environment variable to a service account key path, or by running
- * this sample on Google Cloud where Application Default Credentials are available automatically.
+ * <p>Authenticate by running {@code gcloud auth application-default login}, by setting the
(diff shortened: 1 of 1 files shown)


MukundaKatta:
Thanks for the review. I'll address the Javadoc style points (use `{@code …}` and `<a>` tags instead of Markdown backticks/raw URLs) in a follow-up commit.

PR #12977, opened 2026-05-01
feat(Datastore): Swap the default transport from HttpJson to gRPC

lqiu96:


java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/DatastoreOptions.java
blakeli0, line 1: Not directly related to this PR, but when we bump [major versions](https://github.com/googleapis/google-cloud-java/blob/d690333e4ffe8dda982d61723387ecda971d7d21/versions.txt#L960-L966) for Datastore, we may have to manually bump it because google-cloud-java [always bump minor](https://github.com/googleapis/google-cloud-java/blob/d690333e4ffe8dda982d61723387ecda971d7d21/release-please-config.json#L4).

lqiu96: Thanks, yep. @jinseopkim0 and I are following up on that in a new PR

jinseopkim0 (approved):

PR #12989, opened 2026-05-04
feat(datastore): Bump to v3 major version

lqiu96:
Datastore V3 MajorVersion upgrade:
- Introduction of client side metrics 
- Removal of @Deprecated methods and classes and bumping of existing @ObsoleteApi to @Deprecated
- Swapping the default transport from HttpJson to gRPC
- Configuring a larger default initial channel count and increasing the channel `resize_delta` to better handle spikes in requests


java-datastore/samples/snippets/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-datastore</artifactId>
-      <version>2.40.0</version><!-- {x-version-update:google-cloud-datastore:current} -->
+      <version>3.0.0</version><!-- {x-version-update:google-cloud-datastore:current} -->
```
jinseopkim0: I see this is to 3.0.0, not 3.0.0-SNAPSHOT. Is that intended?

lqiu96: Ah, thanks. Not intentional, I copied the wrong value

jinseopkim0 (approved):

PR #12991, opened 2026-05-04
fix(bqjdbc): pre-populate Named Job ID to enable robust query cancellation waits

rhgoogle:
This PR pre-populates a unique named JobId during statement execution when jobful mode is active, and registers it to the cancellation list before the SDK wait blocks. This prevents thread hangs during manual query cancellation.


rhgoogle:
Context note: This PR is a follow-up fix to #12961 (feat(bqjdbc): Bypass dry-run job for read-only tokens.) which caused query cancellation thread hangs by blocking query wait blocks before the Job ID could be registered in time.

rhgoogle:
not needed anymore, BQ driver team fixes it in a better way.

PR #12995, opened 2026-05-04
build(deps): upgrade grpc to 1.81.0

jinseopkim0:
b/504629399


sdk-platform-java/gax-java/dependencies.properties
```diff
@@ -28,7 +28,7 @@ version.gax_httpjson=2.80.0-SNAPSHOT
 
 version.com_google_protobuf=3.25.8
 version.google_java_format=1.15.0
-version.io_grpc=1.76.2
+version.io_grpc=1.81.0
```
blakeli0: The Bazel build is failing, which I think is because of this change. There might be conflict between the latest grpc bazel build and existing bazel dependencies. If this is hard to fix, I think it is OK to not upgrade it for self-service clients.

jinseopkim0: Thanks for the feedback, reverted this change.

blakeli0 (approved):

PR #12996, opened 2026-05-04
chore(bqjdbc): enhance exception messages to include chained exception message

logachev:
Some tools display to the user only top-level exception message. In order to avoid hiding underlying issues, modify JDBC messages to always include underlying reason if available.


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/exception/BigQueryJdbcCoercionException.java
```diff
@@ -34,7 +34,7 @@ public class BigQueryJdbcCoercionException extends RuntimeException {
    * @param cause the actual cause which was thrown while performing the coercion.
    */
   public BigQueryJdbcCoercionException(Exception cause) {
-    super("Coercion error", cause);
-    LOG.severe("Coercion error", this);
+    super(BigQueryJdbcExceptionUtils.formatMessage("Coercion error", cause), cause);
+    LOG.severe(this.getMessage(), this);
```
Neenu1995, line 38: nit: My latest PR removes the severe logs from the exception classes. You can leave these unedited, it will be removed later.

Neenu1995 (approved):

PR #12999, opened 2026-05-04
fix(bq): do not call 'getQueryResults' for stateful queries via 'query' api

logachev:
`queryWithTimeout` is not supposed to call `getQueryResults` when job is forcefully created. Unit test was meant to cover this scenario, but it didn't set a JOB_CREATION_REQUIRED flag, so it was not using proper API. 

There are multiple issues with calling `getQueryResults` directly:
- timeout is not honored, long-running queries are non-cancellable if JobID is not provided
- `maxResults` per page to read results is not honored

### DIFF
--- a/java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
+++ b/java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
@@ -2104,7 +2104,7 @@ final class BigQueryImpl extends BaseService<BigQueryOptions> implements BigQuer
 
         return queryRpc(projectId, content, options);
       }
-      return create(JobInfo.of(jobId, configuration), options).getQueryResults();
+      return create(JobInfo.of(jobId, configuration), options);
     } finally {
       if (querySpan != null) {
         querySpan.end();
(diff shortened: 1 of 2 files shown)


java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -2104,7 +2104,7 @@ && getOptions().getOpenTelemetryTracer() != null) {
 
         return queryRpc(projectId, content, options);
       }
-      return create(JobInfo.of(jobId, configuration), options).getQueryResults();
+      return create(JobInfo.of(jobId, configuration), options);
```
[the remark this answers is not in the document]
logachev: queryWithTimeout is expected to return either Job or TableResult.

lqiu96, line 2107: doesn't this change the return type? From TableResult to a Job?

lqiu96: nvm, I just realized there is alredy a comment.

lqiu96 (approved):

PR #13000, opened 2026-05-05
chore: add permanent and conditional file exclusions

blakeli0:
This PR adds permanent exclusions for legacy showcases and conditional exclusions for the test/ folder to GEMINI.md to prevent context window bloat in agentic workflows.


lqiu96 (approved):
LGTM let's try it

PR #13001, opened 2026-05-05
feat(bqjdbc): implement Per connection logging with Context proxy

Neenu1995:
b/501081433

1. **Foundational MDC & Global File Handler Routing**
 Simplifies the BigQueryJdbcMdc ThreadLocal context slots, sets up "Jdbc-global" fallback file routing, and adds isFileLoggingEnabled() to the root logger.

2.  **Dynamic Context Proxy Wrapper & Connection Wrap**
 Implements the BigQueryJdbcContextProxy dynamic proxy interceptor, delegating standard wrappers, and wrapping all new JDBC entries in BigQueryDriver.

3. **Statement Modifications & Connection Pool unwrapping**
 Adapts Statements and Connection ID builders to utilize the dynamic proxy, and updates connection pool managers  to safely extract raw connection properties via standard unwrap calls.

4. **0%-Overhead Raw ResultSet & Metadata Routing**
 Completely un-proxies ResultSet to achieve 0% query execution overhead, while implementing targeted ResultSetMetaData proxy wrapping inside getMetaData() to keep metadata queries connection-routed.

5. **Decoupled Exceptions & Clean Integration Tests**
 Decouples all custom exception constructors to eliminate double-logging, converts all 17 integration test class casts to standard unwrap calls, and migrates all deprecated github_timeline public dataset queries to the active shakespeare dataset.


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcContextProxy.java
```diff
@@ -0,0 +1,116 @@
...
+        return wrap(result, java.sql.PreparedStatement.class, connectionId);
+      } else if (result instanceof java.sql.Statement) {
+        return wrap(result, java.sql.Statement.class, connectionId);
+      } else if (result instanceof java.sql.DatabaseMetaData) {
+        return wrap(result, java.sql.DatabaseMetaData.class, connectionId);
+      } else if (result instanceof java.sql.ParameterMetaData) {
+        return wrap(result, java.sql.ParameterMetaData.class, connectionId);
+      } else if (result instanceof java.sql.ResultSetMetaData) {
+        return wrap(result, java.sql.ResultSetMetaData.class, connectionId);
+      }
```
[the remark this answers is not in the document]
Neenu1995: This is intentional.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryBaseResultSet.java
```diff
@@ -46,7 +46,8 @@
 
 public abstract class BigQueryBaseResultSet extends BigQueryNoOpsResultSet
     implements BigQueryResultSet {
-  protected final BigQueryJdbcCustomLogger LOG = new BigQueryJdbcCustomLogger(this.toString());
+  protected final BigQueryJdbcCustomLogger LOG =
+      new BigQueryJdbcCustomLogger(this.getClass().getName());
```
logachev, line 50: nit: We should probably set it to `null` here, that's abstract class, child classes are supposed to instantiate it.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
```diff
@@ -173,7 +173,8 @@ public Connection connect(String url, Properties info) throws SQLException {
             logLevel,
             logPath,
             this.toString());
-        return connection;
+        return BigQueryJdbcContextProxy.wrap(
+            connection, Connection.class, connection.getConnectionId());
```
logachev: nit: `wrap` can extract `connectionId` itself

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLogger.java
```diff
@@ -67,6 +67,21 @@ private void logWithCaller(Level level, Throwable thrown, Supplier<String> msgSu
     }
   }
 
+  @Override
```
logachev, line 70: Why adding these overrides here? 

When we talked earlier, I meant for `JdbcRootLogger()` we will need to remove (throw an exception) when `finest`/`finer` is used & adding another logger `JdbcResultSetLogger` with `finer`/`finest` levels.

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITAuthTests.java
```diff
@@ -72,8 +72,7 @@ private void validateConnection(String connection_uri) throws SQLException {
     assertNotNull(connection);
     assertFalse(connection.isClosed());
     String query =
-        "SELECT DISTINCT repository_name FROM `bigquery-public-data.samples.github_timeline` LIMIT"
-            + " 850";
+        "SELECT DISTINCT word FROM `bigquery-public-data.samples.shakespeare` LIMIT" + " 850";
```
logachev: nit: remove ` + `, just a single string

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITBigQueryJDBCTest.java
```diff
@@ -150,8 +150,7 @@ public void testValidAllDataTypesSerializationFromSelectQueryArrowDataset() thro
   @Test
   public void testFastQueryPathSmall() throws SQLException {
     String query =
-        "SELECT DISTINCT repository_name FROM `bigquery-public-data.samples.github_timeline` LIMIT"
-            + " 850";
+        "SELECT DISTINCT word FROM `bigquery-public-data.samples.shakespeare` LIMIT" + " 850";
```
logachev: nit: Similar: a few ` + ` in that file are no longer required

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLoggerTest.java
```diff
@@ -79,6 +79,38 @@ public void testLogWithCallerInference() {
     assertEquals(BigQueryJdbcCustomLoggerTest.class.getName(), record.getSourceClassName());
   }
 
+  @Test
+  public void testHotPathLoggerLogToDefaultWhenContextIsNull() {
```
logachev, line 83: Can we add a test when context is null & exception is throw?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryBaseResultSet.java
```diff
@@ -107,6 +108,24 @@ public void close() {
     }
   }
 
+  protected SQLException logAndCreateException(SQLException ex) {
+    if (BigQueryJdbcRootLogger.isFileLoggingEnabled() && this.statement != null) {
```
logachev: I don't think we need `isFileLoggingEnabled` check.. That happens right before the exception is thrown, so not a hotpath.. 

And I think `statement` can't be `null` in any situation?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcContextProxy.java
```diff
@@ -0,0 +1,142 @@
...
+
+      return result;
+    } catch (InvocationTargetException e) {
+      Throwable cause = e.getCause();
+
+      // Unified Context Logger: Captures and logs every exception exactly once with the Connection
+      // context
+      if (BigQueryJdbcRootLogger.isFileLoggingEnabled()) {
+        try (BigQueryJdbcMdc.MdcCloseable mdc = BigQueryJdbcMdc.registerInstance(connectionId)) {
```
logachev: It seems like top-level try/catch that handles ResultSet is missing.

This `mdc` object is not required, when this `catch` is processed, `mdc` created on line 108 is still alive.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLogger.java
Neenu1995: This is not for resultSet log handling. Because our JDBC driver uses dynamic proxies and abstract base classes, standard Java logger automatic class-detection gets confused (it would print proxy classes or abstract frameworks as the log source) sometimes. I overloaded these methods to avoid that confusion.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryBaseResultSet.java
Neenu1995: Setting it to null may cause an NPE. Isn't it safer to leave it as-is. The subclasses when they use it, will automatically get their name in the logs.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcContextProxy.java
Neenu1995: The missingResultSet in the top-level try-catch is intentional, since we decided we don't want context proxy invoking methods in RS for context registration.

Java closes try-with-resources before entering the catch block, so the second try (registerInstance) at line 168 is essential to re-establish the connection context thread-local slot specifically for the exception logging handler.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryBaseResultSet.java
Neenu1995: Removed the isFileLoggingEnabled check. But Statement can be null, so kept that check.
Example: 
```
BigQueryJsonResultSet() {
  super(null, null, null, false); // <--- statement is passed as null
  ...
}
```

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLoggerTest.java
Neenu1995: done.

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITBigQueryJDBCTest.java
Neenu1995: fixed.

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITAuthTests.java
Neenu1995: fixed.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
Neenu1995: done

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryBaseResultSet.java
```diff
@@ -107,6 +108,24 @@ public void close() {
     }
   }
 
+  protected SQLException logAndCreateException(SQLException ex) {
+    if (this.statement != null) {
```
logachev: nit: usually it is better to avoid nesting, for multiple conditions easier to read

```
if (this.statement == null) {
  return ex;
}
...
```

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcContextProxy.java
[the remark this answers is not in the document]
logachev: Can you add a comment about it?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcContextProxy.java
```diff
@@ -0,0 +1,176 @@
...
+    this.interfaceType = interfaceType;
+  }
+
+  /** Wraps a target JDBC object with a dynamic proxy carrying the connection context. */
+  @SuppressWarnings("unchecked")
+  static <T> T wrap(Object target, Class<T> interfaceType) {
+    if (target == null) {
+      return null;
+    }
+    String connectionId = extractConnectionId(target);
```
logachev, line 50: I meant that specifically for `Connection` type, it can be extracted.. e.g. for everything else it 100% makes sense to propagate value from the proxy

```
  static <T> T wrap(Connection connection) {
    return wrap(connectio, Connection.class, connection.getConnectionId());
```

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryBaseResultSet.java
Neenu1995: done.

PR #13002, opened 2026-05-05
feat(bqjdbc): Implement Correlated OpenTelemetry Logging Bridge

keshavdandeva:
b/496678357

This PR implements the **Correlated GCP Logging Bridge** for the OpenTelemetry integration in the BigQuery JDBC driver. It enables bridging standard Java logs (`java.util.logging`) to the OpenTelemetry Logs API, allowing users to correlate logs with distributed traces and isolate them by connection session.

### Changes

- `BigQueryDriver.java`: Implemented Cloud-Only Mode matrix logic to suppress local file creation when `enableGcpLogExporter=true` and `LogPath` is omitted.
- `BigQueryJdbcRootLogger.java`: Updated `setLevel` to handle `Level.OFF` properly and skip file handler creation if path is null.
- `BigQueryConnection.java`: Attached `OpenTelemetryJulHandler` to the `"com.google.cloud.bigquery"` namespace during initialization.
- `OpenTelemetryJulHandler.java`: Created a new handler that bridges JUL logs to OTel Logs API with context harvesting and connection ID filtering.
- `pom.xml`: Added `google-cloud-logging` dependency with version `3.33.0-SNAPSHOT` and auto-update marker.
- `OpenTelemetryJulHandlerTest.java`: Created unit tests using `OpenTelemetryExtension` to verify log emission and filtering.

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
@@ -64,7 +64,6 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
-import java.util.logging.Logger;
 
 /**
  * An implementation of {@link java.sql.Connection} for establishing a connection with BigQuery and
@@ -955,6 +954,7 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
     } finally {
       BigQueryJdbcMdc.removeInstance(this);
       BigQueryJdbcRootLogger.closeConnectionHandler(this.connectionId);
+      BigQueryJdbcOpenTelemetry.unregisterConnection(this.connectionId);
     }
     this.isClosed = true;
   }
@@ -1059,10 +1059,8 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
             this.enableGcpTraceExporter, this.enableGcpLogExporter, this.customOpenTelemetry);
 
     if (this.enableGcpLogExporter || this.customOpenTelemetry != null) {
-      OpenTelemetryJulHandler otelHandler =
-          new OpenTelemetryJulHandler(
-              null, openTelemetry, this.enableGcpLogExporter, this.connectionId);
-      Logger.getLogger("com.google.cloud.bigquery").addHandler(otelHandler);
+      BigQueryJdbcOpenTelemetry.registerConnection(
+          this.connectionId, openTelemetry, null, this.enableGcpLogExporter);
     }
 
     if (this.enableGcpTraceExporter || this.customOpenTelemetry != null) {
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
@@ -64,6 +64,7 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
 
 /**
  * An implementation of {@link java.sql.Connection} for establishing a connection with BigQuery and
@@ -1056,6 +1057,14 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
     OpenTelemetry openTelemetry =
         BigQueryJdbcOpenTelemetry.getOpenTelemetry(
             this.enableGcpTraceExporter, this.enableGcpLogExporter, this.customOpenTelemetry);
+
+    if (this.enableGcpLogExporter || this.customOpenTelemetry != null) {
+      OpenTelemetryJulHandler otelHandler =
+          new OpenTelemetryJulHandler(
+              null, openTelemetry, this.enableGcpLogExporter, this.connectionId);
+      Logger.getLogger("com.google.cloud.bigquery").addHandler(otelHandler);
+    }
+
     if (this.enableGcpTraceExporter || this.customOpenTelemetry != null) {
       this.tracer = BigQueryJdbcOpenTelemetry.getTracer(openTelemetry);
       bigQueryOptions.setOpenTelemetryTracer(this.tracer);
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
@@ -157,13 +157,10 @@ public class BigQueryDriver implements Driver {
         if (logPath == null) {
           logPath = System.getenv(BigQueryJdbcUrlUtility.LOG_PATH_ENV_VAR);
         }
-        if (logPath == null) {
-          // Cloud-Only Mode: Suppress local file creation if GCP log exporter is enabled
-          if (ds.getEnableGcpLogExporter() && logLevel != Level.OFF) {
-            logPath = null;
-          } else {
-            logPath = BigQueryJdbcUrlUtility.DEFAULT_LOG_PATH;
-          }
+
+        // Fallback to default path only if not specified and not in Cloud-Only mode
+        if (logPath == null && !ds.getEnableGcpLogExporter()) {
+          logPath = BigQueryJdbcUrlUtility.DEFAULT_LOG_PATH;
         }
 
         BigQueryJdbcRootLogger.setLevel(logLevel, logPath);
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
@@ -158,7 +158,12 @@ public class BigQueryDriver implements Driver {
           logPath = System.getenv(BigQueryJdbcUrlUtility.LOG_PATH_ENV_VAR);
         }
         if (logPath == null) {
-          logPath = BigQueryJdbcUrlUtility.DEFAULT_LOG_PATH;
+          // Cloud-Only Mode: Suppress local file creation if GCP log exporter is enabled
+          if (ds.getEnableGcpLogExporter() && logLevel != Level.OFF) {
+            logPath = null;
+          } else {
+            logPath = BigQueryJdbcUrlUtility.DEFAULT_LOG_PATH;
+          }
         }
 
         BigQueryJdbcRootLogger.setLevel(logLevel, logPath);
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
@@ -28,6 +28,7 @@ public class BigQueryJdbcOpenTelemetry {
 
   static final String INSTRUMENTATION_SCOPE_NAME = "com.google.cloud.bigquery.jdbc";
   static final String BIGQUERY_NAMESPACE = "com.google.cloud.bigquery";
+  public static final String CONNECTION_ID_BAGGAGE_KEY = "jdbc.connection_id";
 
   static class TelemetryConfig {
     final OpenTelemetry openTelemetry;
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
@@ -16,14 +16,59 @@
 
 package com.google.cloud.bigquery.jdbc;
 
+import com.google.cloud.logging.Logging;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
+import java.util.Collection;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.Logger;
 
 public class BigQueryJdbcOpenTelemetry {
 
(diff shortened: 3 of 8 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
```diff
@@ -138,14 +138,17 @@ public static Logger getRootLogger() {
 
   public static void setLevel(Level level, String logPath) throws IOException {
     if (level != Level.OFF) {
-      setPath(logPath, level);
+      if (logPath != null) {
+        setPath(logPath, level);
+      }
```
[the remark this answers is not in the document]
keshavdandeva: explicitly closing and removing the `fileHandler` here would cause issues if multiple connections are active simultaneously with different configurations.

For example, if Connection 1 enables file logging and Connection 2 later starts with Cloud-Only mode (passing null as `logPath`), removing the handler here would abruptly stop file logging for Connection 1 as well.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1056,6 +1057,12 @@ private BigQuery getBigQueryConnection() {
     OpenTelemetry openTelemetry =
         BigQueryJdbcOpenTelemetry.getOpenTelemetry(
             this.enableGcpTraceExporter, this.enableGcpLogExporter, this.customOpenTelemetry);
+
+    if (this.enableGcpLogExporter || this.customOpenTelemetry != null) {
```
logachev, line 1061: Feels weird to check `this.customOpenTelemetry`, but use `openTelemetry`. Is it intended?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
```diff
@@ -158,7 +158,12 @@ public Connection connect(String url, Properties info) throws SQLException {
           logPath = System.getenv(BigQueryJdbcUrlUtility.LOG_PATH_ENV_VAR);
         }
         if (logPath == null) {
-          logPath = BigQueryJdbcUrlUtility.DEFAULT_LOG_PATH;
+          // Cloud-Only Mode: Suppress local file creation if GCP log exporter is enabled
+          if (ds.getEnableGcpLogExporter() && logLevel != Level.OFF) {
```
logachev: What's special about `Level.OFF`? If it is `off`, does it matter what `logPath` value is?

Also based on the comment, if intent is to suppress file logging, it should be done outside of `if (logPath == null)` check

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
```diff
@@ -138,14 +138,17 @@ public static Logger getRootLogger() {
...
       logger.setLevel(level);
     } else {
       for (Handler h : logger.getHandlers()) {
         h.close();
         logger.removeHandler(h);
       }
       fileHandler = null;
+      logger.setLevel(Level.OFF);
```
logachev: Better to move `logger.setLevel(level)` outside of the `if`, it happens in both code paths

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
```diff
@@ -16,15 +16,76 @@
...
 import io.opentelemetry.api.trace.Tracer;
+import java.util.Collection;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.Handler;
+import java.util.logging.Logger;
 
 public class BigQueryJdbcOpenTelemetry {
 
   static final String INSTRUMENTATION_SCOPE_NAME = "com.google.cloud.bigquery.jdbc";
+  static final String BIGQUERY_NAMESPACE = "com.google.cloud.bigquery";
```
logachev, line 30: nit: Should it be `jdbc` namespace?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
```diff
@@ -0,0 +1,190 @@
...
+  @Override
+  public void publish(LogRecord record) {
+    if (!isLoggable(record)) {
+      return;
+    }
+
+    try {
+      // Extract connection ID from baggage
+      String connectionId =
+          Baggage.fromContext(Context.current()).getEntryValue("jdbc.connection_id");
```
logachev: nit: move magic string to some constant

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
```diff
@@ -0,0 +1,190 @@
...
+
+      // Fallback to MDC if not in baggage (if MDC is available and used)
+      if (connectionId == null) {
+        connectionId = BigQueryJdbcMdc.getConnectionId();
+      }
+
+      if (connectionId == null) {
```
logachev, line 62: Should we proceed with `null` connectionId` instead?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  private void publishToGcp(LogRecord record, String connectionId, Logging loggingClient) {
+    Context context = Context.current();
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
+    String traceId = spanContext.isValid() ? spanContext.getTraceId() : null;
+    String spanId = spanContext.isValid() ? spanContext.getSpanId() : null;
+
+    // TODO(b/491238299): May require refinement for structured logging or error handling
+    if (loggingClient == null) {
```
logachev: nit: not needed, you already checked it is non-null before; If there is another code path with `null` loggingClient, it is better to catch the crash in unit/integration tests instead of silently dropping logs

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+    if (connectionId != null) {
+      builder.addLabel("jdbc.connection_id", connectionId);
+    }
+
+    loggingClient.write(Collections.singleton(builder.build()));
+  }
+
+  private com.google.cloud.logging.Severity mapGcpSeverity(Level level) {
+    if (level == Level.SEVERE) return com.google.cloud.logging.Severity.ERROR;
```
logachev, line 109: Please replace with `switch/case`

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
```diff
@@ -0,0 +1,190 @@
...
+    if (level == Level.WARNING) return com.google.cloud.logging.Severity.WARNING;
+    if (level == Level.INFO) return com.google.cloud.logging.Severity.INFO;
+    if (level == Level.CONFIG) return com.google.cloud.logging.Severity.INFO;
+    if (level == Level.FINE) return com.google.cloud.logging.Severity.DEBUG;
+    return com.google.cloud.logging.Severity.DEBUG;
+  }
+
+  private void publishToOTel(LogRecord record, String connectionId, OpenTelemetry openTelemetry) {
+    if (openTelemetry == null) {
```
logachev: similar, not needed

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
```diff
@@ -16,15 +16,76 @@
...
+
+  static class TelemetryConfig {
+    final OpenTelemetry openTelemetry;
+    final Logging loggingClient;
+    final boolean useDirectGcpLogging;
+
+    TelemetryConfig(
```
logachev, line 38: What's the purpose of this config? Is there difference between openTelemetry clients/loggingClients?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: Yes, this is intended. We want to register the connection for telemetry if either GCP logging is explicitly enabled OR if the user provided a `customOpenTelemetry` instance (which might want to receive logs)

Since the local `openTelemetry` variable above is resolved to `customOpenTelemetry` when it is present, passing `openTelemetry` to `registerConnection` is correct.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDriver.java
keshavdandeva: > Also based on the comment, if intent is to suppress file logging, it should be done outside of if (logPath == null) check

The intent here, as we discussed earlier, was to enable 'Cloud-Only Mode' by default when the user omits the `LogPath` property but enables GCP logging. If the user explicitly provides a `LogPath` AND enables GCP logging, we assume they want dual logging to both destinations. Moving this check outside would prevent users from logging to both file and GCP simultaneously.

> What's special about Level.OFF? If it is off, does it matter what logPath value is?

Yeah, that seems redundant. Will remove it and refactor this if-else block

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcRootLogger.java
keshavdandeva: Done

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
keshavdandeva: I did have that earlier but then realised the existing `BigQueryJdbcRootLogger` uses`com.google.cloud.bigquer` namespace as the root logger for the driver. So, changing it here would make it inconsistent with the file handler unless we change it there too

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
keshavdandeva: Done

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
keshavdandeva: If we proceed with a null `connectionId` (subsequently a null `config`), it will still return early because we don't have any configuration registered for global logs in our map. Gemini suggested to use `GlobalOpenTelemetry.get()`. But I didn't wanna pollute the host application's global telemetry space with driver logs by default. 
Thinking about it now, I guess we can do the default config thing for all global logs. So, it would act as a catch-all fallback in our registry for any logs not mapped to a specific connection ID. WDYT?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
keshavdandeva: Done

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
keshavdandeva: `Switch/case` doesn't work for this function as `java.util.logging.Level` is a class with static constants, not an enum, so we cannot use it directly in a `switch` statement. 

Doing so would require using hardcoded integer values in the `case` labels

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
keshavdandeva: The `TelemetryConfig` stores the telemetry settings for a specific JDBC connection. We are using a single global handler to prevent memory leaks, that handler uses the `connectionId` from the log context to look up this config and know how to route the logs for that specific connection.

> Is there difference between openTelemetry clients/loggingClients?

- `openTelemetry` is the standard OTel SDK instance
- `loggingClient` is the direct Google Cloud Logging client. We use it for the "export to GCP Mode" to send logs directly to Cloud Logging, bypassing OTel, because OTel logs are not yet natively supported in GA by Google Cloud.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcOpenTelemetry.java
logachev: I think we should change both

logachev (approved):

PR #13017, opened 2026-05-06
fix(spanner): preserve call options when applying statement timeout

akash329d:
Follow-up to #12726. The grpc-gcp affinity routing it (and googleapis/grpc-gcp-java#235) introduced is bypassed when a `statement_timeout` is set. With multiplexed sessions every data RPC then falls back to the api-config `affinity_key: "session"` rule and has the same original issue.

Build the configurator's result from the input `context` instead so the call options survive the merge. The relative timeout is still applied; `merge()` already requires a `GrpcCallContext`, so the cast is safe.


rahul2393 (approved):

rahul2393:
@akash329d can you please sign cla https://cla.developers.google.com/.

PR #13018, opened 2026-05-06
fix(bqjdbc): fallback to RestAPI if ReadAPI is not accessible

logachev:
b/506217300

ReadAPI requires separate set of permissions, if authenticated user has no access to use it, fallback to a regular paginated API instead of returning an error.

Added info-level log with job/query ID and result row #.


Neenu1995 (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -929,25 +932,47 @@ Thread populateArrowBufferedQueue(
...
-    } else {
-      // read API cannot be used.
-      LOG.fine("Standard API used.");
-      this.currentResultSet = processJsonResultSet(results);
+    JobId jobId = results.getJobId();
+    String queryId = results.getQueryId();
+    LOG.info(
+        "Processing query response. JobId: %s, QueryId: %s, Total rows: %s",
+        jobId, queryId, results.getTotalRows());
+    LOG.fine("Processing query response. Query: %s} ", query);
```
keshavdandeva: nit: Is the `}` a typo or intended

keshavdandeva (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
logachev: thanks!

Issue #13024, opened 2026-05-06
[iam-policy] v1 and v1beta is not generated to java-iam-policy/google-iam-policy


zhumin8:
`google/iam/v1` and  `google/iam/v1beta` are configured to in generation_config.yaml:

https://github.com/googleapis/google-cloud-java/blob/2312aae6be8dcac757e896eb394548531173ed78/generation_config.yaml#L1327-L1343

However, they are not copied in [.OwlBot-hermetic.yaml](https://github.com/googleapis/google-cloud-java/blob/main/java-iam-policy/.OwlBot-hermetic.yaml). So even though generated, it is not copied to final location.

Is this intentional? 
For now, I propose to remove these paths from generation_config.yaml for consistency.

Then, If we desire to add these to generated code, we can: 
- update .OwlBot-hermetic.yaml to copy v1 and v1beta now before migration
- update librarian.yaml after migration to librarian

zhumin8:
Put up https://github.com/googleapis/google-cloud-java/pull/13025 to see if I am missing anything.

zhumin8:
@blakeli0 WDYT? Should we merge https://github.com/googleapis/google-cloud-java/pull/13025?

blakeli0:
I agree. If they were never generated, we should remove them from the config. Separately, we should probably merge the [java-iam](https://github.com/googleapis/google-cloud-java/tree/main/java-iam) module (which only contains grpc and protoc gen code) and the java-iam-policy module.

zhumin8:
What was the reason java-iam they had to be separate? Did this change with the monorepo merge?

At this point, I would prefer to merge them after inital migrate to librarian. I've done most of the work to deal with this special case of separation in librarian, and it should be easier after migrate so we don't need to change 2 systems.

blakeli0:
Agreed that we don't need to merge them before Librarian migration.

> What was the reason java-iam they had to be separate? Did this change with the monorepo merge?

They were separate because java-iam was in sdk-platform-java and java-iam-policy was in google-cloud-java. Now that both are in the same repo, we can merge them.

PR #13025, opened 2026-05-06
chore(iam-policy): remove proto path that is not copied from generation config

zhumin8:
These proto path are configured in generation_config.yaml, but not copied in [.OwlBot-hermetic.yaml](https://github.com/googleapis/google-cloud-java/blob/main/java-iam-policy/.OwlBot-hermetic.yaml). Expect that this change should not lead to code changes after generate workflow runs.

For #13024
For https://github.com/googleapis/librarian/issues/5730


zhumin8:
only generated change is README version.

blakeli0 (approved):

PR #13034, opened 2026-05-07
feat(bqjdbc): Add custom logger to handle ResultSet logs

Neenu1995:
b/510827966

Optimized JDBC driver performance by implementing a custom logger for ResultSet classes, designed for low impact when logging is off and efficient handling when on.

- **Lazy Logging Overloads:** Added Java 8 Supplier overloads to custom loggers, eliminating dynamic string allocations and GC pressure when logging is disabled.

- **Explicit Trace Logging:** Introduced Supplier-based trace overloads (like finestTrace) with explicit method parameters, bypassing CPU-intensive stack-walking.

- **ResultSet-Wide Refactoring:** Refactored 11 ResultSet-related classes to use these lazy trace overloads, removing hot-path logging overhead at the bytecode level.

- **Per-Connection Log Routing:** Enhanced PerConnectionFileHandler to extract connection IDs directly from log parameters, preventing log leakage into the global log.

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowArray.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowArray.java
@@ -41,7 +41,7 @@ class BigQueryArrowArray extends BigQueryBaseArray {
 
   @Override
   public Object getArray() {
-    LOG.finest("++enter++");
+    LOG.finestTrace("getArray", () -> "++enter++");
     ensureValid();
     if (values == null) {
       return null;
@@ -51,7 +51,7 @@ class BigQueryArrowArray extends BigQueryBaseArray {
 
   @Override
   public Object getArray(long index, int count) {
-    LOG.finest("++enter++");
+    LOG.finestTrace("getArray", () -> "++enter++");
     ensureValid();
     if (values == null) {
       return null;
@@ -62,7 +62,7 @@ class BigQueryArrowArray extends BigQueryBaseArray {
 
   @Override
   public ResultSet getResultSet() throws SQLException {
-    LOG.finest("++enter++");
+    LOG.finestTrace("getResultSet", () -> "++enter++");
     ensureValid();
     if (values == null) {
       return new BigQueryArrowResultSet();
@@ -75,7 +75,7 @@ class BigQueryArrowArray extends BigQueryBaseArray {
 
   @Override
   public ResultSet getResultSet(long index, int count) throws SQLException {
-    LOG.finest("++enter++");
+    LOG.finestTrace("getResultSet", () -> "++enter++");
     ensureValid();
     if (values == null) {
       return new BigQueryArrowResultSet();
@@ -89,14 +89,14 @@ class BigQueryArrowArray extends BigQueryBaseArray {
 
   @Override
   public void free() {
-    LOG.finest("++enter++");
+    LOG.finestTrace("free", () -> "++enter++");
     this.values = null;
     markInvalid();
   }
 
   @Override
   Object getCoercedValue(int index) {
-    LOG.finest("++enter++");
+    LOG.finestTrace("getCoercedValue", () -> "++enter++");
     Object value = this.values.get(index);
     return this.arrayOfStruct
         ? new BigQueryArrowStruct(schema.getSubFields(), (JsonStringHashMap<?, ?>) value)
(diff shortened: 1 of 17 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowArray.java
```diff
@@ -41,7 +41,7 @@ public BigQueryArrowArray(Field schema, JsonStringArrayList<?> values) {
 
   @Override
   public Object getArray() {
-    LOG.finest("++enter++");
+    LOG.finestTrace("getArray", () -> "++enter++");
```
logachev: I don't think there is a reason to leverage msgSupplier for fixed strings.

That's my assumption, but I think constant string operations are extremely well optimized in the JVM. From GC perspective, there is still 1 object.. 

Suppliers are valuable when you have some computation (e.g. String.format()).

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
```diff
@@ -449,7 +449,7 @@ private String formatRangeElement(Object element, StandardSQLTypeName elementTyp
 
   @Override
   public void close() {
-    LOG.fine("Closing BigqueryArrowResultSet %s.", this);
+    LOG.finestTrace("close", () -> String.format("Closing BigqueryArrowResultSet %s.", this));
```
logachev: nit: `close()` is not a hot-path, so can still be a regular `fine` logging

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcResultSetLogger.java
```diff
@@ -0,0 +1,138 @@
...
+      }
+    }
+    super.log(record);
+  }
+
+  /**
+   * Log a message at Level.FINEST with predefined class name and method name to avoid stack trace
+   * parsing on the hot-path.
+   */
+  public void finestTrace(String methodName, String msg) {
```
logachev, line 77: All usages of this function are actually logging `+++enter+++`. Should we just replace it with `finestEnter` & no need to provide message or msg supplier at all?

Neenu1995: Since the RS methods don't use proxy setup, it won't know which method it called from without explicitly supplying the method name(or doing a stacktrace, which we want to avoid for perf reasons). But we can remove the ++enter++ message and have it printed here once. Which was one of the things I have planned for the next PR.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowResultSet.java
Neenu1995: done.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryArrowArray.java
Neenu1995: Agreed. Changes made.

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcResultSetLogger.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  @Override
+  public void log(LogRecord record) {
+    if (connectionId != null) {
+      Object[] existingParams = record.getParameters();
+      if (existingParams == null || existingParams.length == 0) {
+        record.setParameters(new Object[] {connectionId});
+      } else {
+        Object[] newParams = new Object[existingParams.length + 1];
```
keshavdandeva, line 64: Doubt: Could changing the params of log record cause race condition, if it is accessed by different threads at same time? Or is that case not valid?

keshavdandeva (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcResultSetLogger.java
Neenu1995: The parameter addition to the LogRecord is completely thread-safe because every log event generates a brand new LogRecord instance with its own freshly allocated parameter array. Since threads never share these individual record objects or array containers with one another, concurrent access conflicts are avoided entirely. Furthermore, the connection ID payload placed inside the array is an immutable string that cannot be altered. Because downstream log handlers only read this data and never modify its contents, there is no shared mutable state, making race conditions impossible.

PR #13036, opened 2026-05-08
chore(spanner): add missing commit ts logging

skuruppu:
If the commit timestamp is missing from a commit response, we can have more visibility into the state of the commit by looking at whether it's in the commit-retry phase or not.


skuruppu:
Closing as #13037 was merged instead.

PR #13090, opened 2026-05-08
fix(bqjdbc):  optimize meetsReadRatio latency to achieve faster page counting

Neenu1995:
b/511231568

- **Collection size check inside meetsReadRatio** — Speeds up page counting by 300x by replacing a slow $O(N)$ loop over 10,000 rows with a fast $O(1)$ collection size call.
- **All-data-already-fetched short-circuit safeguard (totalRows <= pageSize)** — Reduces E2E latency by 30%-40% by preventing redundant gRPC stream setups and double-fetching when results are already fully loaded in memory.
- **O(1) page size approximation fallback** — Prevents slow iterations on custom non-Collection iterables by using a fast math fallback (Math.min(totalRows, maxResultPerPage)).
- **Updated default HighThroughputMinTableSize to 10,000** — Ensures small queries under 10,000 rows bypass gRPC connection setup overhead and run on the faster REST execution path.
- **Added 4 comprehensive unit tests** — Validates all optimization, fallback, and safeguard paths under various dataset sizes and configurations.

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
@@ -72,7 +72,7 @@ final class BigQueryJdbcUrlUtility {
   static final String QUERY_PROPERTIES_NAME = "QueryProperties";
   static final int DEFAULT_HTAPI_ACTIVATION_RATIO_VALUE = 2;
   static final String HTAPI_MIN_TABLE_SIZE_PROPERTY_NAME = "HighThroughputMinTableSize";
-  static final int DEFAULT_HTAPI_MIN_TABLE_SIZE_VALUE = 100;
+  static final int DEFAULT_HTAPI_MIN_TABLE_SIZE_VALUE = 10000;
   static final int DEFAULT_OAUTH_TYPE_VALUE = -1;
   static final String LOCATION_PROPERTY_NAME = "Location";
   static final String ENDPOINT_OVERRIDES_PROPERTY_NAME = "EndpointOverrides";
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
@@ -55,7 +55,6 @@ import com.google.cloud.bigquery.storage.v1.ReadRowsResponse;
 import com.google.cloud.bigquery.storage.v1.ReadSession;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterators;
 import com.google.common.util.concurrent.Uninterruptibles;
 import java.lang.ref.ReferenceQueue;
 import java.sql.Connection;
@@ -952,7 +951,23 @@ public class BigQueryStatement extends BigQueryNoOpsStatement {
     // below log iterates and counts. This is inefficient and we may eventually want to expose
     // PageSize with TableResults
     // TODO(Obada): Scope for performance optimization.
-    int pageSize = Iterators.size(results.getValues().iterator());
+    int pageSize;
+    Iterable<FieldValueList> values = results.getValues();
+    if (values instanceof java.util.Collection) {
+      pageSize = ((java.util.Collection<?>) values).size();
+    } else {
+      // O(1) Fast Page Size Approximation:
+      // If the values iterable is not a collection, approximate the page size rather than
+      // performing a slow O(N) iteration over the entire page of query results.
+      pageSize = (int) Math.min(totalRows, querySettings.getMaxResultPerPage());
+    }
+
+    // SAFEGUARD: If all data has already been retrieved in the first page,
+    // NEVER switch to the Read API as it would discard in-memory data and cause a double-fetch.
+    if (totalRows <= pageSize) {
+      return false;
+    }
+
     return totalRows / pageSize > querySettings.getHighThroughputActivationRatio();
   }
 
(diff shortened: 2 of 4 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -952,7 +951,23 @@ private boolean meetsReadRatio(TableResult results) {
...
+      // If the values iterable is not a collection, approximate the page size rather than
+      // performing a slow O(N) iteration over the entire page of query results.
+      pageSize = (int) Math.min(totalRows, querySettings.getMaxResultPerPage());
+    }
+
+    // SAFEGUARD: If all data has already been retrieved in the first page,
+    // NEVER switch to the Read API as it would discard in-memory data and cause a double-fetch.
+    if (totalRows <= pageSize) {
+      return false;
+    }
```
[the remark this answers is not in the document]
Neenu1995: Fixed.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryTypeCoercionUtility.java
```diff
@@ -198,14 +197,10 @@ private static class LongToTime implements BigQueryCoercion<Long, Time> {
...
-      int HH = (int) TimeUnit.MICROSECONDS.toHours(value);
-      int MM = (int) (TimeUnit.MICROSECONDS.toMinutes(value) % 60);
-      int SS = (int) (TimeUnit.MICROSECONDS.toSeconds(value) % 60);
-
       // Note: BQ Time has a precision of up to six fractional digits (microsecond precision)
-      // but java.sql.Time do not. So data after seconds is not returned.
-      return new Time(HH, MM, SS);
+      // but java.sql.Time only supports up to millisecond precision. So data after milliseconds is
+      // truncated.
+      return new Time(value / 1000);
```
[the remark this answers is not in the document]
Neenu1995: Rebased.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -952,7 +951,28 @@ private boolean meetsReadRatio(TableResult results) {
...
+      pageSize = ((java.util.Collection<?>) values).size();
+    } else {
+      // O(1) Fast Page Size Approximation:
+      // If the values iterable is not a collection, approximate the page size rather than
+      // performing a slow O(N) iteration over the entire page of query results.
+      pageSize = Math.min(totalRows, querySettings.getMaxResultPerPage());
+    }
+
+    // Prevent division by zero or negative pageSize due to potential overflows/empty sets:
+    if (pageSize <= 0) {
```
logachev, line 957: That's an impossible condition. It can't be < 0 and if totalRows is 0 is already checked

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -952,7 +951,28 @@ private boolean meetsReadRatio(TableResult results) {
...
+      pageSize = 1;
+    }
+
+    // SAFEGUARD: If all data has already been retrieved in the first page,
+    // NEVER switch to the Read API as it would discard in-memory data and cause a double-fetch.
+    if (totalRows <= pageSize) {
```
logachev: Instead of doing this, we can check `TableResult.getNextPageToken()`. 
This check is actually dangerous because pageSize is an estimate

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -952,7 +951,28 @@ private boolean meetsReadRatio(TableResult results) {
...
-    int pageSize = Iterators.size(results.getValues().iterator());
+    long pageSize;
+    Iterable<FieldValueList> values = results.getValues();
+    if (values instanceof java.util.Collection) {
+      pageSize = ((java.util.Collection<?>) values).size();
+    } else {
+      // O(1) Fast Page Size Approximation:
+      // If the values iterable is not a collection, approximate the page size rather than
+      // performing a slow O(N) iteration over the entire page of query results.
+      pageSize = Math.min(totalRows, querySettings.getMaxResultPerPage());
```
logachev: Can `values` even be a Collection here? I think we can just use `getMaxResultsPerPage()` here. If you add next page token check, totalRows is always going to be <= maxResultsPerPage.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -72,7 +72,7 @@ protected boolean removeEldestEntry(Map.Entry<String, Map<String, String>> eldes
   static final String QUERY_PROPERTIES_NAME = "QueryProperties";
   static final int DEFAULT_HTAPI_ACTIVATION_RATIO_VALUE = 2;
   static final String HTAPI_MIN_TABLE_SIZE_PROPERTY_NAME = "HighThroughputMinTableSize";
-  static final int DEFAULT_HTAPI_MIN_TABLE_SIZE_VALUE = 100;
+  static final int DEFAULT_HTAPI_MIN_TABLE_SIZE_VALUE = 10000;
```
keshavdandeva, line 75: We would also need to update this in documentation/user guide. Idk if there is already a bug to track this or if that's part of this task.

Neenu1995: created the bug.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
logachev: I still think we don't need this. getMaxResultPerPage() shouldn't be <= 0. We can add this check during connection creation, but we shouldn't check it when we use it

logachev (approved):

PR #13164, opened 2026-05-11
chore: partial release of datastore as 1.86.0 #2

jinseopkim0:
b/511245639

### DIFF
--- a/gapic-libraries-bom/pom.xml
+++ b/gapic-libraries-bom/pom.xml
@@ -15,7 +15,7 @@
   <parent>
     <artifactId>google-cloud-pom-parent</artifactId>
     <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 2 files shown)


gapic-libraries-bom/pom.xml
```diff
@@ -4,7 +4,7 @@
   <groupId>com.google.cloud</groupId>
   <artifactId>gapic-libraries-bom</artifactId>
   <packaging>pom</packaging>
-  <version>1.87.0-SNAPSHOT</version><!-- {x-version-update:google-cloud-java:current} -->
+  <version>1.86.1</version><!-- {x-version-update:google-cloud-java:current} -->
```
blakeli0, line 7: I think this version needs to be used in the datastore?

PR #13169, opened 2026-05-11
feat: Enable Bound Token for Agentic Identities

vverman:
This PR introduces a feature which enables the auth library to acquire bound **access-tokens** and **bound id-tokens** in Agentic Environments.

1. We detect certs in default paths and check if they match the SPIFFE format for agents.

2. If 1. is a yes then we call the MDS endpoint in a POST request with the certificate in the body.


google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,399 @@
...
+        timeService.sleep(sleepInterval);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new IOException(
+            "Interrupted while waiting for well-known certificate files.",
+            e);
+      }
+    }
+    throw new IOException(
+        "Unable to find well-known certificate file for bound token request after multiple retries.");
```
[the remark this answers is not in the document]
vverman: IIUC This is a limitation of the present design.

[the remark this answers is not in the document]
nbayati: We actually had several issues reported in python and had to add additional checks that were not present in the original design (go/sdk-agent-identity):
https://github.com/googleapis/google-cloud-python/pull/17377
https://github.com/googleapis/google-cloud-python/pull/17568

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+   * locations.
+   */
+  static ResolvedCertAndKeyPaths resolveCertAndKeyPaths(String certConfigPath) throws IOException {
+    String certPath = null;
+    String keyPath = null;
+
+    if (!Strings.isNullOrEmpty(certConfigPath)) {
+      // Read cert path from config file. We use retry with backoff to handle transient race
+      // conditions where the config file might be being updated by a rotation process.
+      certPath = getCertificatePathWithRetry(certConfigPath);
```
macastelaz, line 199: Doesn't this ("getCertificatePathWithRetry" followed by "extractKeyPathFromConfig") open and parse the file at certConfigPath twice in succession? Ideally, we'd only open and parse the file only once

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+        throw new IOException(
+            "Certificate intent inferred via config, but cert files are missing.");
+      }
+      // Neither cert-config nor certsexist, do not enable
+      return false;
+    }
+  }
+
+  /** Retrieves the bound token payload (certificate chain) if applicable. */
+  static String getBoundTokenPayload() throws IOException {
```
macastelaz, line 452: Looks like this is called in every refreshAccessToken() and idTokenWithAudience() call on ComputeEngineCredentials? getBoundTokenPayload seems to do a lot of I/O. Are the performance implications to this acceptable/expected? Is there an opportunity to do any caching of some/all of the bound token payload?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+    CertInfo info = getAgentIdentityCertInfo();
+    if (info != null && shouldRequestBoundToken(info.certificate)) {
+      return readCertificateChain(info.path);
```
macastelaz, line 455: I think this "re-parses"/re-reads the cert file.  I wonder if there is a way to optimize this?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+    return null;
+  }
+
+  @SuppressWarnings("unchecked")
+  /** Extracts the certificate path from the JSON configuration file. */
+  private static String extractCertPathFromConfig(String certConfigPath) throws IOException {
+    try (InputStream stream = new FileInputStream(certConfigPath)) {
+      JsonObjectParser parser = new JsonObjectParser(OAuth2Utils.JSON_FACTORY);
+      GenericJson config = parser.parseAndClose(stream, StandardCharsets.UTF_8, GenericJson.class);
+      Map certConfigs = (Map) config.get("cert_configs");
```
macastelaz, line 466: Do we want this to throw exceptions for misconfigured JSON? E.g. I think if cert_configs or workload ends up being a String or JSON array we would see a ClassCastException?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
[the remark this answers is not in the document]
macastelaz: I'd be curious to learn more about this too - based on the AI feedback, this definitely doesn't sound to be desirable (even if "working as designed").

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+            "Certificate intent established via config, but cert files are missing.");
+      }
+      // Neither exist, do not enable
+      return false;
+    }
+    // Case 2: Explicitly disabled via environment variable
+    else if ("false".equalsIgnoreCase(useClientCert)) {
+      if (certsPresent) {
+        // Warn that we are ignoring present certs because it was explicitly disabled
+        Slf4jUtils.log(
```
macastelaz, line 426: Doesn't this run the risk of spamming logs as this could trigger WARN messages on every refresh or idTokenWithAudience

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+      if (certsPresent) {
+        // Infer mTLS is enabled because certs are present
+        return true;
+      }
+      if (configExists) {
+        // Config exists but files are missing - fail fast
+        throw new IOException(
+            "Certificate intent inferred via config, but cert files are missing.");
+      }
+      // Neither cert-config nor certsexist, do not enable
```
macastelaz, line 446: nit: typo "certsexist" should be "certs exist"

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+            LOGGER,
+            org.slf4j.event.Level.WARN,
+            Collections.emptyMap(),
+            String.format(
+                "Certificate config file not found at %s (from %s environment variable). Retrying for up to %d seconds.",
+                certConfigPath, GOOGLE_API_CERTIFICATE_CONFIG, TOTAL_TIMEOUT_MS / 1000));
+        warned = true;
+      }
+      try {
+        timeService.sleep(sleepInterval);
```
macastelaz, line 303: I think the AI review flagged something similar, but I'm concerned about this sleeping in what seems to be such a critical/main thread of applications. Are we sure sleeping like this on the "hot path" of token generation is acceptable?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
macastelaz: Also worth noting I believe this technically introduces a race condition by reading twice (the file content can change between reads)

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+        }
+      }
+
+      if (!matched) {
+        throw new IOException(
+            "Agent Identity certificate and private key mismatch or read failure after 3 retries.");
+      }
+    } else if (!Strings.isNullOrEmpty(certPath)) {
+      // Bundle or only cert available
+      cert = parseCertificate(certPath);
```
macastelaz, line 261: Does this need a proactive retry mechanism like the other flow?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
macastelaz: I also think there may be some concurrency issues here that lead to CPU starvation if there are a bunch of threads concurrently requesting an ID token.

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
```diff
@@ -1177,6 +1219,117 @@ void getProjectId_explicitSet_noMDsCall() {
...
+    assertNotNull(token);
+    com.google.api.client.testing.http.MockLowLevelHttpRequest request =
+        transportFactory.transport.getRequest();
+    assertEquals("POST", transportFactory.transport.getRequestMethod());
+    String body = request.getContentAsString();
+    assertTrue(body.contains("certificate_chain"));
+  }
+
+  @Test
+  void idTokenWithAudience_withValidCertAndKey_requestsBoundToken() throws IOException {
```
macastelaz, line 1301: Is it worth testing when there is a cert but it is not a spiffe x509 cert?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,572 @@
...
+    CertInfo info = getAgentIdentityCertInfo();
+    if (info != null && shouldRequestBoundToken(info.certificate)) {
+      return readCertificateChain(info.path);
+    }
+    return null;
+  }
+
+  @SuppressWarnings("unchecked")
+  /** Extracts the certificate path from the JSON configuration file. */
+  private static String extractCertPathFromConfig(String certConfigPath) throws IOException {
```
macastelaz, line 462: On a similar note to my next comment, I don't think we have test coverage for malformed JSON. Probaby worth adding one

google-auth-library-java/oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
```diff
@@ -1177,6 +1219,117 @@ void getProjectId_explicitSet_noMDsCall() {
...
+    Map<String, Object> config = new HashMap<>();
+    config.put("cert_configs", certConfigs);
+
+    String configContent = OAuth2Utils.JSON_FACTORY.toString(config);
+    Files.write(configPath, configContent.getBytes(StandardCharsets.UTF_8));
+    envProvider.setEnv("GOOGLE_API_CERTIFICATE_CONFIG", configPath.toAbsolutePath().toString());
+  }
+
+  @Test
+  void refreshAccessToken_withValidCertAndKey_requestsBoundToken() throws IOException {
```
macastelaz, line 1280: Is it worth testing what happens if the metadata server gets mocked to return an error (400 or 403, for example)?

PR #13174, opened 2026-05-12
fix(bqjdbc): validate integer connection properties

Neenu1995:
- **Property Validation:** The driver now rejects negative integers for properties like pool sizes and timeouts.
- **Consistent Enforcement:** This validation applies to programmatic DataSource setters.
- **Remove Double Parsing:** Removed the `BigQueryJdbcUrlUtility.parseUrl(connectionUri)` usage in `BigQueryDriver.connect()` to eliminate double parsing. 
- **Better Error Handling:** Invalid inputs now throw clear, driver-specific exceptions instead of low-level runtime errors.
- **Expanded Testing:** Added tests to verify behavior for negative values, non-numeric inputs, and unrecognized properties.

### DIFF
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
@@ -72,7 +72,7 @@ final class BigQueryJdbcUrlUtility {
   static final String QUERY_PROPERTIES_NAME = "QueryProperties";
   static final int DEFAULT_HTAPI_ACTIVATION_RATIO_VALUE = 2;
   static final String HTAPI_MIN_TABLE_SIZE_PROPERTY_NAME = "HighThroughputMinTableSize";
-  static final int DEFAULT_HTAPI_MIN_TABLE_SIZE_VALUE = 100;
+  static final int DEFAULT_HTAPI_MIN_TABLE_SIZE_VALUE = 10000;
   static final int DEFAULT_OAUTH_TYPE_VALUE = -1;
   static final String LOCATION_PROPERTY_NAME = "Location";
   static final String ENDPOINT_OVERRIDES_PROPERTY_NAME = "EndpointOverrides";
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
@@ -858,15 +858,15 @@ final class BigQueryJdbcUrlUtility {
   }
 
   private static boolean isNonNegativeIntegerProperty(String propertyName) {
-    return propertyName.equals(MAX_RESULTS_PROPERTY_NAME)
-        || propertyName.equals(CONNECTION_POOL_SIZE_PROPERTY_NAME)
-        || propertyName.equals(LISTENER_POOL_SIZE_PROPERTY_NAME)
-        || propertyName.equals(HTAPI_MIN_TABLE_SIZE_PROPERTY_NAME)
-        || propertyName.equals(HTAPI_ACTIVATION_RATIO_PROPERTY_NAME)
-        || propertyName.equals(METADATA_FETCH_THREAD_COUNT_PROPERTY_NAME)
-        || propertyName.equals(HTTP_CONNECT_TIMEOUT_PROPERTY_NAME)
-        || propertyName.equals(HTTP_READ_TIMEOUT_PROPERTY_NAME)
-        || propertyName.equals(SWA_ACTIVATION_ROW_COUNT_PROPERTY_NAME)
-        || propertyName.equals(SWA_APPEND_ROW_COUNT_PROPERTY_NAME);
+    return MAX_RESULTS_PROPERTY_NAME.equals(propertyName)
+        || CONNECTION_POOL_SIZE_PROPERTY_NAME.equals(propertyName)
+        || LISTENER_POOL_SIZE_PROPERTY_NAME.equals(propertyName)
+        || HTAPI_MIN_TABLE_SIZE_PROPERTY_NAME.equals(propertyName)
+        || HTAPI_ACTIVATION_RATIO_PROPERTY_NAME.equals(propertyName)
+        || METADATA_FETCH_THREAD_COUNT_PROPERTY_NAME.equals(propertyName)
+        || HTTP_CONNECT_TIMEOUT_PROPERTY_NAME.equals(propertyName)
+        || HTTP_READ_TIMEOUT_PROPERTY_NAME.equals(propertyName)
+        || SWA_ACTIVATION_ROW_COUNT_PROPERTY_NAME.equals(propertyName)
+        || SWA_APPEND_ROW_COUNT_PROPERTY_NAME.equals(propertyName);
   }
 }
--- a/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
+++ b/java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
@@ -723,7 +723,6 @@ final class BigQueryJdbcUrlUtility {
       }
       String propertyName = PROPERTY_NAME_MAP.get(key);
       String value = CharEscapers.decodeUriPath(kv[1].replace("+", "%2B"));
-      validateNonNegativeIntegerProperty(propertyName, value);
       map.put(propertyName, value);
     }
     return Collections.unmodifiableMap(map);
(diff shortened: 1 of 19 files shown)


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -721,7 +721,10 @@ private static Map<String, String> parseUrlInternal(String url) {
           continue;
         }
       }
-      map.put(PROPERTY_NAME_MAP.get(key), CharEscapers.decodeUriPath(kv[1].replace("+", "%2B")));
+      String propertyName = PROPERTY_NAME_MAP.get(key);
+      String value = CharEscapers.decodeUriPath(kv[1].replace("+", "%2B"));
+      validateNonNegativeIntegerProperty(propertyName, value);
+      map.put(propertyName, value);
```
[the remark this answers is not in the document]
Neenu1995: This is already handled upstream.

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtilityTest.java
```diff
@@ -206,4 +208,42 @@ public void testAppendPropertiesToURL_propertyWithSemicolon_isEscaped() throws E
...
+    String url = "jdbc:bigquery://;MaxResults=-1";
+    assertThrows(BigQueryJdbcRuntimeException.class, () -> DataSource.fromUrl(url));
+
+    String url2 = "jdbc:bigquery://;ConnectionPoolSize=-2";
+    assertThrows(BigQueryJdbcRuntimeException.class, () -> DataSource.fromUrl(url2));
+  }
+
+  @Test
+  public void testValidZeroConnectionProperties() {
+    String url = "jdbc:bigquery://;MaxResults=0";
```
logachev: MaxResults actually can't be 0 :)

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -833,4 +836,37 @@ static Map<String, String> parsePropertiesMapFromValue(
...
+  static void validateNonNegative(long val, String propertyName) {
+    if (val < 0) {
+      throw new BigQueryJdbcRuntimeException(
+          String.format(
+              "Invalid value for %s. It must be greater than or equal to 0.", propertyName));
+    }
+  }
+
+  private static void validateNonNegativeIntegerProperty(String propertyName, String value) {
+    if (isNonNegativeIntegerProperty(propertyName)) {
```
logachev: nit: isNonNegativeIntegerProperty

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -721,7 +721,10 @@ private static Map<String, String> parseUrlInternal(String url) {
           continue;
         }
       }
-      map.put(PROPERTY_NAME_MAP.get(key), CharEscapers.decodeUriPath(kv[1].replace("+", "%2B")));
+      String propertyName = PROPERTY_NAME_MAP.get(key);
```
logachev, line 724: Do we need this check here? iirc we'll construct DataSource from the Connection string now (@keshavdandeva is it true?), so it will perform the validation.

keshavdandeva: Yeah, it is redundant. `DataSource.fromUrl` calls `parseUrl` and then applies the properties using the setters in DataSource, which performs the same validation.

But while confirming this flow, I realised that `BigQueryDriver.connect` calls `parseUrl` before `DataSource.fromUrl` specifically to catch `BigQueryJdbcRuntimeException` and wrap it in a `BigQueryJdbcException` (which is a `SQLException`, which we should throw per JDBC spec)

keshavdandeva: So, I think correct/cleaner way would be to:

- Remove this validation from here `BigQueryJdbcUrlUtility.parseUrlInternal`
- Fix `BigQueryDriver.connect` logic by either:
a) wrap the `DataSource.fromUrl` call in the try-catch block instead of the standalone `parseUrl` call
b) or totally replacing line `BigQueryJdbcUrlUtility.parseUrl(connectionUri);` -> `ds = DataSource.fromUrl(connectionUri);` as it will parse the URL AND call setters that perform validation

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/DataSource.java
```diff
@@ -664,6 +664,10 @@ public void setProjectId(String projectId) {
   }
```
keshavdandeva, line 672: Doubt: Should we also add non-negative validations for timeout/delay properties like `Timeout,` `JobTimeout,` `RetryInitialDelay,` and `RetryMaxDelay`?

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -833,4 +836,37 @@ static Map<String, String> parsePropertiesMapFromValue(
...
+        long val = Long.parseLong(value);
+        validateNonNegative(val, propertyName);
+      } catch (NumberFormatException ex) {
+        throw new BigQueryJdbcRuntimeException(
+            String.format("Invalid value for %s. It must be a valid integer.", propertyName), ex);
+      }
+    }
+  }
+
+  private static boolean isNonNegativeIntegerProperty(String propertyName) {
```
keshavdandeva: If we go with the approach of relying on DataSource setters for validation (and wrapping `DataSource.fromUrl` in BigQueryDriver), we can remove the `isNonNegativeIntegerProperty` method entirely (and wont need to maintain this list as well)

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtilityTest.java
Neenu1995: Max results can be zero. It is an accepted use case to only retrieve query metadata in getQueryResults.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
Neenu1995: Replacing `BigQueryJdbcUrlUtility.parseUrl` makes sense since `DataSource.fromUrl` is basically calling `parseUrl` inside it, we are double parsing right now.
Made the changes.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
Neenu1995: Agreed. Needed this before for the connection string verification.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/DataSource.java
Neenu1995: Good catch. I overlooked those. adding them now.

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtilityTest.java
```diff
@@ -206,4 +208,25 @@ public void testAppendPropertiesToURL_propertyWithSemicolon_isEscaped() throws E
...
   }
+
+  @Test
+  public void testInvalidConnectionProperties() {
+    String url = "jdbc:bigquery://;MaxResults=-1";
+    assertThrows(BigQueryJdbcRuntimeException.class, () -> DataSource.fromUrl(url));
+
+    String url2 = "jdbc:bigquery://;ConnectionPoolSize=-2";
+    assertThrows(BigQueryJdbcRuntimeException.class, () -> DataSource.fromUrl(url2));
+  }
```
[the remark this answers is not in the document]
Neenu1995: Already present.

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtilityTest.java
logachev: hm, I'd argue that driver should not allow that. If it is set to 0, we'll get background thread that will (likely) infinitely try to retrieve results. I'm actually surprised to see that BQ backend does honor it & provides "nextPageToken"

logachev (approved):

keshavdandeva (approved):

PR #13184, opened 2026-05-13
chore(spanner): migrate grpc-gcp into google-cloud-java

rahul2393:
## Summary

  Migrate grpc-gcp from GoogleCloudPlatform/grpc-gcp-java/grpc-gcp into google-cloud-java/grpc-gcp as a top-level base Maven module with existing com.google.cloud:grpc-gcp coordinates.

  ## What changed

  - Imported grpc-gcp source under root grpc-gcp.
  - Added grpc-gcp to root Maven reactor.
  - Kept google-cloud-spanner depending on com.google.cloud:grpc-gcp, avoiding duplicate classes in google-cloud-spanner.
  - Added Release Please version marker for next grpc-gcp version 1.11.0-SNAPSHOT.
  - Updated sdk-platform-java/java-shared-dependencies/first-party-dependencies to manage grpc-gcp at 1.11.0-SNAPSHOT.
  - Kept grpc-gcp dependency management self-contained to avoid circular dependency:
      - grpc-gcp does not import google-cloud-shared-dependencies.
      - google-cloud-shared-dependencies can continue managing grpc-gcp through first-party-dependencies.
  - Updated CI package filters so grpc-gcp runs as its own package and still triggers Spanner checks when changed.
  - Excluded live integration tests from default grpc-gcp test compilation.
  - Stabilized GcpManagedChannelTest.testLogMetrics.
  - Added Java 8-only exclusion for Spanner native-image sources because current GraalVM nativeimage artifact requires Java 11 bytecode.

  ## Validation

  - mvn -q -pl grpc-gcp test -Pquick-build
  - cd grpc-gcp && mvn -q dependency:analyze -DfailOnWarning=true -DskipTests -Pquick-build
  - mvn -q -pl java-spanner/google-cloud-spanner -am -DskipTests compile -Pquick-build
  - cd java-spanner && mvn -q dependency:analyze -Pquick-build -DfailOnWarning=true -Dmdep.analyze.skip=false
  - ./generation/check_non_release_please_versions.sh


java-spanner/grpc-gcp/pom.xml
```diff
@@ -0,0 +1,296 @@
...
+    <clirr.skip>true</clirr.skip>
+    <jacoco.skip>true</jacoco.skip>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-shared-dependencies</artifactId>
+        <version>${google-cloud-shared-dependencies.version}</version>
```
blakeli0: This would cause circular dependency issues that `google-cloud-shared-dependencies` manages `grpc-gcp` version and `grpc-gcp` also imports `google-cloud-shared-dependencies`. 
Maven build does not error out because it does not validate declared versions in `dependencyManagement` sections. But it will cause issues in releasing and confusion in dependency management.

I think we have two options:
1. Treat `grpc-gcp` a top-level base library like auth. It manages its own dependencies (for now). Its version is managed by first-party-dependencies. It is imported through first-party-dependencies by downstream libraries such as Spanner.
2. Treat is as a sub module of Spanner and remove it from first-party-dependencies. This is a cleaner solution but it would introduce breaking changes to customers. Now my question is: Is there any `grpc-gcp` classes exposed on the surface and used by customers directly? If not, there are probably minimum risks.

blakeli0:
Can you update [CODEOWNERS](https://github.com/googleapis/google-cloud-java/blob/main/.github/CODEOWNERS) so that grpc-gcp is owned by the Spanner team?

grpc-gcp-java/src/main/proto/google/grpc/gcp/proto/grpc_gcp.proto
blakeli0, line 1: How is this file used? Are the gen code of the this proto file generated during compilation phase and shipped with the jar?

java-grpc-gcp/pom.xml
```diff
@@ -0,0 +1,311 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>google-cloud-shared-config</artifactId>
+    <!-- Do not depend on the version in this monorepo. It causes publication order problems. -->
```
blakeli0: This is an issue for auth but not an issue for `grpc-grpc`. Because the groupId of `grpc-gcp` is also `com.google.cloud`

grpc-gcp-java/pom.xml
```diff
@@ -0,0 +1,311 @@
...
+    <google-http-client.version>2.1.0</google-http-client.version>
+    <gson.version>2.13.2</gson.version>
+    <guava.version>33.5.0-jre</guava.version>
+    <protobuf.version>4.33.2</protobuf.version>
+    <grpc.version>1.81.0</grpc.version>
+    <jsr305.version>3.0.2</jsr305.version>
+    <junit.version>4.13.2</junit.version>
+    <opencensus.version>0.31.1</opencensus.version>
+    <opentelemetry.version>1.57.0</opentelemetry.version>
+    <spanner.version>6.118.0-SNAPSHOT</spanner.version><!-- {x-version-update:google-cloud-spanner:current} -->
```
blakeli0, line 77: Reminder: the Spanner test should be moved to Spanner module right after this PR is merged.

grpc-gcp-java/README.md
```diff
@@ -0,0 +1,116 @@
+# gRPC-GCP for Java
+
+grpc-gcp provides gRPC support for Google Cloud Clients.
```
blakeli0, line 3: Add a blurb that this module is only used by Spanner?

pom.xml
```diff
@@ -125,6 +125,7 @@
     <module>java-gkehub</module>
     <module>java-gkerecommender</module>
     <module>java-grafeas</module>
+    <module>java-grpc-gcp</module>
```
blakeli0: I think this module name can stay as `grpc-gcp-java`. All other `java-*` are a client library but this is more like an implementation details of the client libraries.

rahul2393: I can rename it to grpc-gcp-java if preferred, but we would also need to update the root POM generator/metadata flow so root-pom CI keeps it. Otherwise, the generated root POM will drop the module,  root-pom CI regenerates pom.xml, and removes non-java-* modules.

grpc-gcp-java/src/main/proto/google/grpc/gcp/proto/grpc_gcp.proto
rahul2393: Yes. grpc_gcp.proto defines ApiConfig, MethodConfig, AffinityConfig, and ChannelPoolConfig. Those generated classes are used directly by the main code. The protobuf-maven-plugin runs during compile and it includes in the shipped jar.

pom.xml
rahul2393: I tested grpc-gcp-java against the current root POM generator image, and it has been dropped from the generated pom.xml. So using that path requires a generator change/release, not just this PR.

blakeli0: It should be fixed in https://github.com/googleapis/google-cloud-java/pull/13211 now. I would prefer it to stay as `grpc-gcp-java`, similar to `google-auth-library-java`

rahul2393: Done

rahul2393: root-pom CI may still depend on generator Docker image to get latest main change, does that mean we can't merge this PR?

blakeli0: It takes some time for the image to be regenerated and uploaded, it should work soon.

blakeli0 (approved):

PR #13208, opened 2026-05-15
chore(firestore): remove redundant files

sofisl:
For https://github.com/googleapis/librarian/issues/6012


java-firestore/google-cloud-firestore/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-firestore/reflect-config.json
blakeli0, line 1: I don't think they are redundant, the [reflect-config.json](https://github.com/googleapis/google-cloud-java/blob/main/java-firestore/google-cloud-firestore/src/main/resources/META-INF/native-image/com.google.cloud.firestore.v1/reflect-config.json) under `com.google.cloud.firestore.v1` is for proto gen code from `proto-google-cloud-firestore-v1`. 

This file seems to be the GraalVM config for `proto-google-cloud-firestore-bundle-v1`

PR #13211, opened 2026-05-15
fix(hermetic_build): add grpc-gcp-java to module allowlist in root pom generator

blakeli0:
This PR adds grpc-gcp-java to the MODULE_ALLOWLIST in pom_generator.py. This ensures that when grpc-gcp-java is migrated to the monorepo, the root pom generator will correctly preserve its module entry in the root pom.xml without requiring a java- prefix.

This unblocks https://github.com/googleapis/google-cloud-java/pull/13184


blakeli0:
cc: @zhumin8 A new repo [grpc-gcp-java](https://github.com/googleapis/google-cloud-java/pull/13184) is being added to the monorepo, this PR is to make changes in the generator to unblock it.

jinseopkim0 (approved):

rahul2393 (approved):

PR #13212, opened 2026-05-15
feat(gax): implement dynamic channel refreshing on 401 retries

blakeli0:
This PR implements dynamic channel refreshing on 401 Unauthenticated retries under the isMwlidEnvironment environment variable. It introduces compile-time type-safe refresh contracts across TransportChannel and ApiCallContext, with debouncing protection in ChannelPool to prevent connection stampedes.

### DIFF
--- a/sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
+++ b/sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
@@ -82,6 +82,7 @@ class ChannelPool extends ManagedChannel {
   private ScheduledFuture<?> resizeFuture = null;
 
   private final Object entryWriteLock = new Object();
+  private long lastRefreshTimeNanos = 0;
   @VisibleForTesting final AtomicReference<ImmutableList<Entry>> entries = new AtomicReference<>();
   private final AtomicInteger indexTicker = new AtomicInteger();
   private final String authority;
@@ -441,6 +442,13 @@ class ChannelPool extends ManagedChannel {
     // - then thread2 will shut down channel that thread1 will put back into circulation (after it
     //   replaces the list)
     synchronized (entryWriteLock) {
+      long now = System.nanoTime();
+      if (now - lastRefreshTimeNanos < TimeUnit.SECONDS.toNanos(5)) {
+        LOG.fine("Channel pool was refreshed recently, skipping duplicate refresh");
+        return;
+      }
+      lastRefreshTimeNanos = now;
+
       LOG.fine("Refreshing all channels");
       ArrayList<Entry> newEntries = new ArrayList<>(entries.get());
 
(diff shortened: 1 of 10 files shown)


sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/ScheduledRetryingExecutor.java
```diff
@@ -122,4 +126,27 @@ public ApiFuture<ResponseT> submit(RetryingFuture<ResponseT> retryingFuture) {
...
   }
+
+  private void checkForFailedChannelRefresh(RetryingFuture<ResponseT> retryingFuture) {
+    ApiFuture<ResponseT> lastAttemptResult = retryingFuture.peekAttemptResult();
+    if (lastAttemptResult != null && lastAttemptResult.isDone()) {
+      try {
+        lastAttemptResult.get();
+      } catch (java.util.concurrent.ExecutionException e) {
+        Throwable cause = e.getCause();
+        if (cause instanceof com.google.api.gax.rpc.UnauthenticatedException) {
```
vverman: Having com.google.api.gax.rpc classes (e.x. ApiCallContext, TransportChannel) inside ScheduledRetryingExecutor creates a circular package dependency with com.google.api.gax.retrying package.

We can define the MtlsRotationHandler within gax.retrying which can be inferred from the RetryingContext. The executor only interacts with the MtlsRotationHandler.

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiResultRetryAlgorithm.java
```diff
@@ -51,6 +55,10 @@ public boolean shouldRetry(Throwable previousThrowable, ResponseT previousRespon
   @Override
   public boolean shouldRetry(
       RetryingContext context, Throwable previousThrowable, ResponseT previousResponse) {
+    if ("true".equalsIgnoreCase(System.getenv("isMwlidEnvironment"))
+        && previousThrowable instanceof UnauthenticatedException) {
```
vverman, line 59: nit: This checks only for UNAUTHENTICATED requests, instead we should check for cert-mismatch which can be done by passing some param with the context. However, it is key that we avoid redundant disk-reads since many requests could fail simultaneously.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -441,6 +442,13 @@ void refresh() {
     // - then thread2 will shut down channel that thread1 will put back into circulation (after it
     //   replaces the list)
     synchronized (entryWriteLock) {
+      long now = System.nanoTime();
+      if (now - lastRefreshTimeNanos < TimeUnit.SECONDS.toNanos(5)) {
```
vverman, line 446: nit: One potential concern here is if in the 5 nano second gap the certs rotate and the request fails due to a cert-mismatch. This could lead to valid requests failing. 

I think we can build a workaround by using the cert-mismatch as a trigger.

vverman (commented):
In general I think this design is elegant and handles the unary approach well. There might be changes needed to accomodate cert-mismatches and avoid circular dependencies.

The only open concern remains streaming requests which we cannot leverage the RetryExecutor for. Since we don't want to retry a failed stream request and instead just refresh the channel, IIUC, we would be restricted to using a per call interceptor.

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/ScheduledRetryingExecutor.java
blakeli0: Agreed. Even though they belong to the same jar and there are no compilation issues, it is a good design practice to keep the relationship one-direction.

That being said, this is more of a proof of concept to demonstrate a way to do channel refresh outside of grpc/httpjson, and use transport neutral way to refresh channels.  I updated the approach to use future callbacks instead, there could be other approaches.

PR #13214, opened 2026-05-18
feat(spanner): add option for auto-tagging transactions

sakthivelmanii:
Adds an option that automatically adds transaction tags to all read/write transactions that do not already include an explicit transaction tag. For read-only transactions, this option automatically adds a request tag to each statement that is executed in the read-only transaction. The tag name is equal to the class name + method name of the method that started the transaction. Tags are limited to max 50 characters. If the combination of class name + method name exceeds 50 characters, then the last 50 characters are used as the tag.

This option can be enabled with `SpannerOptions$Builder.enableAutoTagging()` or by setting either the environment variable SPANNER_ENABLE_AUTO_TAGGING=true or the system property `spanner.enable_auto_tagging=true`.

### DIFF
--- a/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -644,6 +644,7 @@ abstract class AbstractReadContext
   private final DirectedReadOptions defaultDirectedReadOptions;
   private final DecodeMode defaultDecodeMode;
   private final Clock clock;
+  private volatile String cachedRequestTag;
 
   @GuardedBy("lock")
   private boolean isValid = true;
@@ -841,7 +842,22 @@ abstract class AbstractReadContext
       builder.setClientContext(clientContextBuilder.build());
     }
     if (getTransactionTag() != null) {
-      builder.setTransactionTag(getTransactionTag());
+      // Read-write transactions support transaction-level tags only. We populate the
+      // transaction tag on the builder if it is non-empty.
+      if (!getTransactionTag().isEmpty()) {
+        builder.setTransactionTag(getTransactionTag());
+      }
+    } else if (session.getSpanner().getOptions().isAutoTaggingEnabled()
+        && builder.getRequestTag().isEmpty()) {
+      // Read-only contexts (both single-use and multi-use) do not support transaction-level tags.
+      // We lazily resolve and populate the request tag instead.
+      if (this.cachedRequestTag == null) {
+        String autoTag = AutoTagHelper.getAutoTag(session.getSpanner().getOptions());
+        this.cachedRequestTag = autoTag == null ? "" : autoTag;
+      }
+      if (!this.cachedRequestTag.isEmpty()) {
+        builder.setRequestTag(this.cachedRequestTag);
+      }
     }
     return builder.build();
   }
--- a/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AutoTagHelper.java
+++ b/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AutoTagHelper.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spanner;
+
+import java.util.List;
+
+/** Helper for Spanner transaction tags. */
+final class AutoTagHelper {
+
+  /** Maximum allowed character length for resolved tags. */
+  private static final int MAX_TAG_LENGTH = 50;
+
+  /** Ignored packages. */
+  private static final String[] INTERNAL_PACKAGES;
+
+  static {
+    INTERNAL_PACKAGES =
+        new String[] {
+          "java.",
+          "javax.",
+          "jdk.",
+          "sun.",
+          "io.grpc.",
+          "com.google.cloud.spanner.",
+          "com.google.api."
+        };
+  }
+
+  private AutoTagHelper() {
+    // prevent instantiation
+  }
+
+  static String getAutoTag(final SpannerOptions options) {
+    StackTraceElement[] stackTrace = new Throwable().getStackTrace();
+    int tracerLimit = options.getAutoTaggingTracerLimit();
+    int limit = Math.min(stackTrace.length, tracerLimit);
+    List<String> targetPackages = options.getAutoTaggingPackages();
+    boolean hasTarget = targetPackages != null && !targetPackages.isEmpty();
+
+    for (int i = 0; i < limit; i++) {
+      StackTraceElement element = stackTrace[i];
+      String className = element.getClassName();
+      if (hasTarget) {
+        for (String targetPackage : targetPackages) {
+          if (className.startsWith(targetPackage)) {
+            return formatTag(className, element.getMethodName());
+          }
+        }
+      } else if (isInternalPackage(className)) {
+        continue;
+      } else {
+        return formatTag(className, element.getMethodName());
+      }
+    }
+    return null;
+  }
+
+  private static boolean isInternalPackage(final String cls) {
+    for (String internalPackage : INTERNAL_PACKAGES) {
+      if (cls.startsWith(internalPackage)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private static String formatTag(final String cls, final String method) {
+    int lastDot = cls.lastIndexOf('.');
+    String simpleClassName;
+    if (lastDot == -1) {
+      simpleClassName = cls;
+    } else {
+      simpleClassName = cls.substring(lastDot + 1);
+    }
+    String tag = simpleClassName + "." + method;
+    if (tag.length() > MAX_TAG_LENGTH) {
+      tag = tag.substring(tag.length() - MAX_TAG_LENGTH);
+    }
+    return tag;
+  }
+}
--- a/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
+++ b/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
@@ -96,6 +96,7 @@ import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -308,6 +309,9 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
   private final String monitoringHost;
   private final TransactionOptions defaultTransactionOptions;
   private final RequestOptions.ClientContext clientContext;
+  private final boolean autoTaggingEnabled;
+  private final List<String> autoTaggingPackages;
+  private final int autoTaggingTracerLimit;
 
   enum TracingFramework {
     OPEN_CENSUS,
@@ -993,6 +997,9 @@ public class SpannerOptions extends ServiceOptions<Spanner, SpannerOptions> {
     monitoringHost = builder.monitoringHost;
     defaultTransactionOptions = builder.defaultTransactionOptions;
     clientContext = builder.clientContext;
+    autoTaggingEnabled = builder.autoTaggingEnabled;
+    autoTaggingPackages = builder.autoTaggingPackages;
+    autoTaggingTracerLimit = builder.autoTaggingTracerLimit;
   }
 
   private String getResolvedUniverseDomain() {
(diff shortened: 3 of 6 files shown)


java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -2120,6 +2143,31 @@ public Builder setDefaultClientContext(RequestOptions.ClientContext clientContex
...
+
+    public Builder setAutoTagTransactionsPackage(String autoTagTransactionsPackage) {
+      this.autoTagTransactionsPackages = Collections.singletonList(autoTagTransactionsPackage);
+      return this;
+    }
+
+    public Builder setAutoTagTransactionsPackages(List<String> autoTagTransactionsPackages) {
+      this.autoTagTransactionsPackages = Preconditions.checkNotNull(autoTagTransactionsPackages);
+      return this;
+    }
```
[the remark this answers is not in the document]
sakthivelmanii: Done

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AutoTagHelper.java
```diff
@@ -0,0 +1,77 @@
...
+    "io.grpc.",
+    "com.google.cloud.spanner.",
+    "com.google.api."
+  };
+
+  static String getAutoTag(SpannerOptions options) {
+    if (!options.isAutoTagTransactionsEnabled()) {
+      return null;
+    }
+    StackTraceElement[] stackTrace = new Throwable().getStackTrace();
```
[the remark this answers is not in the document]
sakthivelmanii: Yes. but since we use animal-sniffer, it's not possible to do it without using reflection which might be difficult and adds latency given that we execute for every transaction

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -831,6 +831,12 @@ QueryOptions buildQueryOptions(QueryOptions requestOptions) {
 
   RequestOptions buildRequestOptions(Options options) {
     RequestOptions.Builder builder = options.toRequestOptionsProto(false).toBuilder();
+    if (builder.getRequestTag().isEmpty()) {
+      String autoTag = AutoTagHelper.getAutoTag(session.getSpanner().getOptions());
```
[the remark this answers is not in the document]
sakthivelmanii: done

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -831,6 +832,14 @@ QueryOptions buildQueryOptions(QueryOptions requestOptions) {
 
   RequestOptions buildRequestOptions(Options options) {
     RequestOptions.Builder builder = options.toRequestOptionsProto(false).toBuilder();
+    if (builder.getRequestTag().isEmpty()) {
+      if (this.cachedRequestTag == null) {
+        this.cachedRequestTag = AutoTagHelper.getAutoTag(session.getSpanner().getOptions());
```
olavloite: This seems to add auto-request tags, although the option name seems to indicate that this is only for transactions. So we either need to change the name of the option (and maybe also add this in more places than only here), or remove it from here.

Also, unconditionally calling `getAutoTag(..)` here without checking whether auto-tagging has been enabled, is not really readable. I had to look into `getAutoTag(..)` to check whether it checks the option first or not. Instead, we should add a check before calling the method. That would also fix a small caching error here: If `this.cachedRequestTag == null`, then we call `getAutoTag(..)` for every invocation. And while it returns quickly, it slightly defies the entire idea of caching it here.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AutoTagHelper.java
```diff
@@ -0,0 +1,71 @@
...
+        for (String internalPackage : INTERNAL_PACKAGES) {
+          if (className.startsWith(internalPackage)) {
+            isInternal = true;
+            break;
+          }
+        }
+        if (isInternal) {
+          continue;
+        }
+        return formatTag(className, element.getMethodName());
```
olavloite: Can we simplify this a bit.

1. Move the check whether it is an internal package to a separate helper method:

```java
private static boolean isInternalPackage(String className) {
  for (String internalPackage : INTERNAL_PACKAGES) {
    if (className.startsWith(internalPackage)) {
      return true;
    }
  }
  return false;
}
```

Then simplify the block to something like this:

```java
if (isInternalPackage(className)) {
  continue;
}
return formatTag(className, element.getMethodName());
```

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AutoTagHelper.java
```diff
@@ -0,0 +1,71 @@
...
+    return null;
+  }
+
+  private static String formatTag(String className, String methodName) {
+    int lastDot = className.lastIndexOf('.');
+    String simpleClassName = lastDot == -1 ? className : className.substring(lastDot + 1);
+    String tag = simpleClassName + "." + methodName;
+    if (tag.length() > 50) {
+      tag = tag.substring(0, 50);
+    }
```
olavloite: This returns the 50 __first__ characters if the generated name is too long. I think that taking the __last__ 50 characters would make more sense, as that will then include the method name. I would expect that it is more useful for a customer to know the method name, than (part of) the package and class name. This is especially true if the reason that the 50-chars limit was exceeded was caused by a very long package name that is reused across the entire application.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -315,13 +315,19 @@ public CommitResponse writeAtLeastOnceWithOptions(
 
   private RequestOptions getRequestOptions(TransactionOption... transactionOptions) {
     Options requestOptions = Options.fromTransactionOptions(transactionOptions);
-    if (requestOptions.hasPriority() || requestOptions.hasTag()) {
+    String autoTag = null;
+    if (!requestOptions.hasTag()) {
+      autoTag = AutoTagHelper.getAutoTag(spanner.getOptions());
```
olavloite: I'm not sure if this really is the best place to capture the auto-tag. Do we know all the possible code paths for getting to this point? Could it be that it is also called from a worker thread in the client library? Or from some internal method that is called when a transaction is retried? Etc...

Would it maybe make more sense to move the actual capturing of the auto-tag as close as possible to the entry point in the client library. Best case would basically be the first line of code in the implementation of `DatabaseClient#readWriteTransaction(..)` (although that literally might not be feasible)

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionImpl.java
```diff
@@ -501,6 +507,12 @@ ApiFuture<Transaction> beginTransactionAsync(
     }
     RequestOptions.Builder optionsBuilder =
         transactionOptions.toRequestOptionsProto(true).toBuilder();
+    if (!transactionOptions.hasTag()) {
+      String autoTag = AutoTagHelper.getAutoTag(spanner.getOptions());
```
olavloite: Same question as above, especially as this is an async method: Is this really the best place to capture the auto-tag?

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1168,6 +1179,12 @@ public boolean isEnableLocationApi() {
       return Boolean.parseBoolean(System.getenv(EXPERIMENTAL_LOCATION_API_ENV_VAR));
     }
 
+    @Override
+    public boolean isAutoTagTransactionsDisabled() {
+      return Boolean.parseBoolean(System.getenv("SPANNER_DISABLE_AUTO_TAGGING"))
+          || Boolean.parseBoolean(System.getProperty("spanner.disable_auto_tagging"));
```
olavloite, line 1189: I would rather expect this feature to be something that you want to be able to __enable__ with an environment variable or system property. The most common case is:
1. Someone writes an application and it works just fine in test.
2. They deploy it to prod and stuff goes wrong and everyone panics
3. During debugging, the question comes up: Do you have tagging?
4. ...

Could we instead add an env var / system property that allows you to flip it both ways? (And should it then really always override what is set in code?)

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/TransactionRunnerImpl.java
```diff
@@ -780,7 +781,10 @@ String getTransactionTag() {
       if (this.options.hasTag()) {
         return this.options.tag();
       }
-      return null;
+      if (this.cachedTransactionTag == null) {
```
olavloite: Here also the same as for request tag above: I think it would be much more readable if we do the check for whether auto-tagging is enabled here, instead of having `getAutoTag(..)` implicitly return `null` in that case. That then also fixes the (small) caching issue that repeatedly assigns `null` to this variable when auto-tagging is disabled.

java-spanner/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockAutoTaggingTest.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.cloud.NoCredentials;
+import com.google.spanner.v1.CommitRequest;
+import com.google.spanner.v1.ExecuteSqlRequest;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class MockAutoTaggingTest extends AbstractMockServerTest {
```
olavloite, line 34: - These tests do not cover the async versions (so `runAsync` and if we also want to support requests; `executeQueryAsync`)
- There are no tests for combining auto-tagging with explicit tags
- Can we test for truncation (auto-tag longer than 50 chars)
- Can we test the system property / env var function for disabling (and enabling?) the feature?

olavloite:
Can we update both the pull request title and add a new paragraph at the start of the pull request description? The current title and description are very technical and 'internal', instead of focusing on what's in it for a customer.

Suggestion:

```
feat(spanner): add option for auto-tagging transactions

Adds an option that automatically adds transaction tags to all read/write transactions that do not already include an explicit transaction tag. For read-only transactions, this option automatically adds a request tag to each statement that is executed in the read-only transaction. The tag name is equal to the class name + method name of the method that started the transaction. Tags are limited to max 50 characters. If the combination of class name + method name exceeds 50 characters, then the last 50 characters are used as the tag.

This option can be enabled with `SpannerOptions$Builder.enableAutoTagging()` or by setting either the environment variable SPANNER_ENABLE_AUTO_TAGGING=true or the system property spanner.enable_auto_tagging=true.
```

java-spanner/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockAutoTaggingTest.java
```diff
@@ -0,0 +1,334 @@
...
+          databaseClient,
+          transaction -> {
+            transaction.executeUpdate(Statement.of("UPDATE Venues SET Capacity = 100"));
+            return null;
+          });
+
+      // Verify transaction tag populated on ExecuteSqlRequest
+      assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+      ExecuteSqlRequest sqlRequest = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0);
+      String transactionTag = sqlRequest.getRequestOptions().getTransactionTag();
```
olavloite, line 87: Can we add a test verification here that asserts that the statement does not contain a request tag? See also my comment above about my confusion as to why request tags are not being added to read/write transactions.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -842,6 +843,15 @@ RequestOptions buildRequestOptions(Options options) {
     }
     if (getTransactionTag() != null) {
       builder.setTransactionTag(getTransactionTag());
+    } else if (session.getSpanner().getOptions().isAutoTagTransactionsEnabled()
```
olavloite: I had to look long and hard to understand why this in in the end leads to the following behavior:
1. Multi-use and single-use read-only transactions set a request tag when auto-tagging is enabled (unless an explicit tag has been set)
2. Read/write transactions do __not__ set a request tag. The reason for that is that `getTransactionTag()` in __most__ cases returns a non-null value when the transaction type is `ReadWriteTransaction`.

The only case where `getTransactionTag()` returns null for a read/write transaction is when the following happens:

If AutoTagHelper.getAutoTag(...) cannot find a matching application stack frame:
- cachedTransactionTag gets set to "".
- getTransactionTag() returns null.
- buildRequestOptions falls into the else if block.
- cachedRequestTag evaluates AutoTagHelper.getAutoTag(...), which again returns null.
- cachedRequestTag gets set to "".
- Since cachedRequestTag is empty, no request tag is added.

Can we add a test verification to the MockAutoTaggingTest to verify this behavior, and prevent accidental changes in the future? (And/or make the behavior here a bit more explicit)

olavloite (approved):
LGTM, but please also take a look at https://github.com/googleapis/google-cloud-java/pull/13214#issuecomment-4486559717

java-spanner/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MockAutoTaggingTest.java
sakthivelmanii: Done

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
sakthivelmanii: Done

olavloite (approved):

PR #13226, opened 2026-05-19
test(bigquery-jdbc): add e2e otel test and code refinements

keshavdandeva:
b/499079838

#### 1. BigQueryJdbcOpenTelemetry.java
*   **Feature**: Added safe, generous default attribute value length limits of **`60KB (61,440 characters)`** to the autoconfigured OpenTelemetry instance properties.
*   **Why**: Prevents GCP Cloud Trace from silently rejecting and dropping span batches when we log massive `BigQueryException` stack traces or Arrow schema payloads exceeding the hard 64KB Cloud Trace backend limits.
*   **Design**: If the user explicitly configures their own limits the driver automatically skips the defaults and respects their overrides.

#### 2. OpenTelemetryJulHandler.java
*   **Fix**: Configured the handler level to `Level.ALL` in the constructor.
*   **Why**: Bypasses a standard Java Logging (JUL) constraint where handlers default to `Level.INFO` and silently drop `FINE`/`DEBUG` queries. Delegates log filtering exclusively to the Connection loggers.

#### 3. BigQueryConnection.java
*   **Visibility**: Exposed the visibility of the connection session identifier by changing `getConnectionId()` from package-private to `public`.
*   **Why**: Allows automated E2E tests to retrieve the UUID and harvest specific logs/traces accurately.

#### 4. ITOpenTelemetryTest.java
*   **Feature**: Implemented a new standalone E2E integration test suite verifying the live GCP OTel egress.
*   **Test 1 (`testExecute_withOpenTelemetryGcpExporter`)**: Natively resolves target project via `ServiceOptions.getDefaultProjectId()`. Runs an optimized in-memory array query and iterates results to trigger small-page JSON pagination. Queries Cloud Trace E2E to strictly assert that async pagination child spans are parented perfectly under the root JDBC span.
*   **Test 2 (`testExecute_withErrorCorrelation`)**: Triggers database failures, captures `SQLException`, harvests Trace IDs from standard logs, and verifies failed span ingestion in Cloud Trace.

#### 5. BigQueryConnectionTest.java
*   **Feature**: Added a new unit test (`testConnect_withCustomOpenTelemetry_usesCustomInstance`) verifying the custom OTel injection pipeline.
*   **What it does**: Leverages `OpenTelemetryExtension` to mock an OTel provider locally. Injecting the custom SDK via properties, it validates that `BigQueryConnection` resolves the instance and routes spans exclusively to the custom provider

#### 6. pom.xml
*   **Dependencies**: Added `google-cloud-trace` test-scoped dependency to query Cloud Trace v1 API programmatically during E2E validation.


gapic-libraries-bom/pom.xml
```diff
@@ -301,13 +301,6 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-      <dependency>
```
logachev: assume some merge mismatch

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -363,7 +363,7 @@ String getConnectionUrl() {
     return connectionUrl;
   }
 
-  String getConnectionId() {
+  public String getConnectionId() {
```
logachev, line 445: `@VisibleForTesting` instead?

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
```diff
@@ -42,7 +42,9 @@
 public class OpenTelemetryJulHandler extends Handler {
   private static final Pattern UNSAFE_LOG_CHARACTERS = Pattern.compile("[^a-zA-Z0-9./_-]");
 
-  public OpenTelemetryJulHandler() {}
+  public OpenTelemetryJulHandler() {
+    setLevel(Level.ALL);
```
logachev, line 46: Is this intentional? I don't think we should alter log level from non-test code

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
```diff
@@ -0,0 +1,297 @@
...
+  public void testExecute_withOpenTelemetryGcpExporter() throws Exception {
+    assumeTrue(
+        PROJECT_ID != null && !PROJECT_ID.trim().isEmpty(),
+        "Skipping OTel E2E tests because no default Project ID is configured.");
+
+    // Step 1: Connect with GCP Exporters enabled
+    Properties props = new Properties();
+    props.setProperty("enableGcpTraceExporter", "true");
+    props.setProperty("enableGcpLogExporter", "true");
+    props.setProperty("LogLevel", "3"); // Triggers FINE log generation
```
logachev: 3 is `INFO`

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
```diff
@@ -0,0 +1,297 @@
...
+
+  private static final String PROJECT_ID = ServiceOptions.getDefaultProjectId();
+  private static final String CONNECTION_URL =
+      String.format(
+          "jdbc:bigquery://https://www.googleapis.com/bigquery/v2:443;ProjectId=%s;OAuthType=3;Timeout=3600;",
+          PROJECT_ID);
+
+  @Test
+  public void testExecute_withOpenTelemetryGcpExporter() throws Exception {
+    assumeTrue(
```
logachev: I'd advise against conditional skip for tests based on runtime.. We can select tests at suite level.

Can add tags for these tests & skip based on tags, but not on config-step

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
```diff
@@ -0,0 +1,297 @@
...
+        PROJECT_ID != null && !PROJECT_ID.trim().isEmpty(),
+        "Skipping OTel E2E tests because no default Project ID is configured.");
+
+    // Step 1: Connect with GCP Exporters enabled
+    Properties props = new Properties();
```
logachev: nit: is it better to use data source to guarantee that there are no typos & if parameters are changed it is reflecte

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
```diff
@@ -0,0 +1,297 @@
...
+      // Retrieve the Connection UUID programmatically
+      BigQueryConnection bqConnection = connection.unwrap(BigQueryConnection.class);
+      connectionUuid = bqConnection.getConnectionId();
+      assertNotNull(connectionUuid, "Connection UUID should be generated");
+
+      // Execute an in-memory array query (scans 0 bytes, extremely fast) and force pagination
+      String paginationQuery = "SELECT * FROM UNNEST(GENERATE_ARRAY(1, 1000)) AS id;";
+      try (ResultSet paginatedRs = statement.executeQuery(paginationQuery)) {
+        int rowCount = 0;
+        while (paginatedRs.next() && rowCount < 1000) {
```
logachev: nit: do we need 20 pages here? That's ~15-18 seconds iirc

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
```diff
@@ -0,0 +1,297 @@
...
+      assertNotNull(connectionUuid, "Connection UUID should be generated");
+
+      // Execute a query designed to fail due to non-existent table
+      boolean caughtException = false;
+      try {
+        statement.executeQuery("SELECT * FROM invalid_dataset.invalid_table;");
+      } catch (SQLException e) {
+        caughtException = true;
+      }
+      assertTrue(caughtException, "Expected SQLException to be thrown");
```
logachev: nit: assertThrows()`

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    // Step 2: Retrieve logs from Cloud Logging and assert error logs
+    String traceId = null;
```
logachev: can trace\log validation be shared?

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/OpenTelemetryJulHandler.java
keshavdandeva: Yeah, so this does not modify or alter any global logging levels. By default, Java Handler thresholds default to `INFO.` Without `setLevel(Level.ALL)`, our custom handler would silently drop all `FINE` query logs before exporting them, even if the user explicitly enabled lower log level on their connection.

Overriding it here just ensures log filtering is done strictly at the Logger level and not here

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
keshavdandeva: Okay, removed

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
keshavdandeva: Yes

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
keshavdandeva: No, I can reduce them. Also did some other optimizations

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
keshavdandeva: Yes, refactored

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -912,8 +912,7 @@ private void processArrowStream(
       enqueueError(arrowBatchWrapperBlockingQueue, e);
       Thread.currentThread().interrupt();
     } catch (Exception e) {
-      if (e.getCause() instanceof InterruptedException
-          || Thread.currentThread().isInterrupted()) {
+      if (e.getCause() instanceof InterruptedException || Thread.currentThread().isInterrupted()) {
```
logachev: That can never happen, if it is InterruptedException, it is handled above (and if it is interrupted, no need to call interrupt again); 
I see that you just changed formatting, but lets clean up.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -1684,8 +1683,7 @@ private void parseAndPopulateRpcData(
       }
 
     } catch (Exception ex) {
-      if (ex.getCause() instanceof InterruptedException
-          || Thread.currentThread().isInterrupted()) {
+      if (ex.getCause() instanceof InterruptedException || Thread.currentThread().isInterrupted()) {
```
logachev: same here

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITOpenTelemetryTest.java
```diff
@@ -0,0 +1,266 @@
...
+      BigQueryConnection bqConnection = connection.unwrap(BigQueryConnection.class);
+      connectionUuid = bqConnection.getConnectionId();
+      assertNotNull(connectionUuid, "Connection UUID should be generated");
+
+      // Execute an in-memory array query (scans 0 bytes, extremely fast) and force pagination (3
+      // pages)
+      String paginationQuery = "SELECT * FROM UNNEST(GENERATE_ARRAY(1, 150)) AS id;";
+      try (ResultSet paginatedRs = statement.executeQuery(paginationQuery)) {
+        int rowCount = 0;
+        while (paginatedRs.next() && rowCount < 150) {
```
logachev: nit: let's remove rowCount here. If we get into infinite loop with `next()`, better to fail the test (it will eventually die with timeout)

logachev (approved):

PR #13231, opened 2026-05-20
chore(spanner): use channel affinity

rahul2393:
Internal reference: go/grpc-gcp-fixes#bookmark=id.q4x3oa8l672


java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -2239,12 +2237,8 @@ static void resetActiveTracingFramework() {
   }
 
   public static TracingFramework getActiveTracingFramework() {
-    synchronized (lock) {
```
rahul2393, line 2242: FYI: this change is unrelated, but since it was small perf change I included it here, this was on critical request path being shown up in benchmark mutex profile

rahul2393:
Screenshot 2026-05-20 at 7 11 55 PM

Prober running stale_read at 400QPS on DirectPath shows 200us save

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
```diff
@@ -1961,6 +2018,12 @@ public String authority() {
   @Override
   public <ReqT, RespT> ClientCall<ReqT, RespT> newCall(
       MethodDescriptor<ReqT, RespT> methodDescriptor, CallOptions callOptions) {
+    ChannelAffinityRef channelAffinityRef = callOptions.getOption(CHANNEL_AFFINITY_REF_KEY);
+    if (channelAffinityRef != null) {
+      return new GcpClientCall.SimpleGcpClientCall<>(
+          this, getChannelRefByAffinityRef(channelAffinityRef), methodDescriptor, callOptions);
```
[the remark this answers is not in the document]
rahul2393: Addressed: 
pickLeastBusyChannel never returns null and NoopGcpClientCall doesn't exist — same null-handling (none) as the sibling getChannelRef(key) path, so no change needed.

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
```diff
@@ -1678,6 +1698,42 @@ protected ChannelRef getChannelRef(@Nullable String key) {
...
+      }
+
+      ChannelRef selectedChannelRef =
+          useDifferentChannel
+              ? pickLeastBusyChannelDifferentFrom(channelRef)
+              : pickLeastBusyChannel(/* forFallback= */ false);
+      if (affinityRef.channelRef.compareAndSet(channelRef, selectedChannelRef)) {
+        return selectedChannelRef;
+      }
+    }
```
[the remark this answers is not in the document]
rahul2393: Both versions current/suggested consume the flag once on the first iteration and drop it after — the suggested rewrite is the same behavior with extra lines.

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
```diff
@@ -1678,6 +1698,42 @@ protected ChannelRef getChannelRef(@Nullable String key) {
...
+        || channelRef != excludedChannelRef
+        || !excludedChannelRef.isActive()
+        || channelRefs.size() <= 1) {
+      return channelRef;
+    }
+    for (ChannelRef candidate : channelRefs) {
+      if (candidate != excludedChannelRef) {
+        return candidate;
+      }
+    }
```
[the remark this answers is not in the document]
rahul2393: Stale review

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
```diff
@@ -1678,6 +1696,51 @@ protected ChannelRef getChannelRef(@Nullable String key) {
...
+      if (!useDifferentChannel && channelRef != null && channelRef.isActive()) {
+        return channelRef;
+      }
+
+      ChannelRef selectedChannelRef =
+          useDifferentChannel
+              ? pickLeastBusyChannelDifferentFrom(channelRef)
+              : pickLeastBusyChannel(/* forFallback= */ false);
+      if (affinityRef.channelRef.compareAndSet(channelRef, selectedChannelRef)) {
+        return selectedChannelRef;
```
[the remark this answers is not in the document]
rahul2393: Neither pickLeastBusyChannel nor pickLeastBusyChannelDifferentFrom can return null — empty-pool case throws on channelRefs.get(0) long before this point, so the NPE path doesn't exist.

rahul2393:
image


YCSB run with 50 threads on database with 10GB data shows improvement of around 100us for 30mins run

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
```diff
@@ -1678,6 +1715,58 @@ protected ChannelRef getChannelRef(@Nullable String key) {
     return mappedChannel;
   }
 
+  /**
+   * Pick a {@link ChannelRef} using a caller-owned reference instead of grpc-gcp's affinity map.
+   */
+  protected ChannelRef getChannelRefByAffinityRef(ChannelAffinityRef affinityRef) {
+    maybeDynamicUpscale();
+    while (true) {
```
olavloite, line 1724: nit: can we add a clarifying comment for why we are looping here?

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -537,10 +526,6 @@ public void close() {
       } finally {
         txnLock.unlock();
       }
-      ByteString id = getTransactionId();
-      if (id != null && !id.isEmpty()) {
-        rpc.clearTransactionAndChannelAffinity(id, Option.CHANNEL_HINT.getLong(channelHint));
```
olavloite, line 542: This not only cleaned up channel affinity, but also transaction affinity for location-aware routing, right? I think that we still need the latter.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
```diff
@@ -633,16 +634,6 @@ public void clearTransactionAffinity(ByteString transactionId) {
     }
   }
 
-  @Override
-  public void clearTransactionAndChannelAffinity(
-      ByteString transactionId, @Nullable Long channelHint) {
-    if (keyAwareChannel != null) {
-      keyAwareChannel.clearTransactionAndChannelAffinity(transactionId, channelHint);
```
olavloite, line 640: Are you sure that we do not need this call anymore? (See also my other comment on AbstractReadContext)

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
```diff
@@ -366,21 +366,6 @@ void clearTransactionAffinity(ByteString transactionId) {
     clearAffinity(transactionId);
   }
 
-  void clearTransactionAndChannelAffinity(ByteString transactionId, @Nullable Long channelHint) {
-    String address = transactionAffinities.asMap().remove(transactionId);
-    readOnlyTxPreferLeader.invalidate(transactionId);
```
olavloite, line 371: Is this also no longer needed? (I mean specifically the `readOnlyTxPreferLeader.invalidate` call)

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
```diff
@@ -107,6 +107,40 @@ public class GcpManagedChannel extends ManagedChannel {
   public static final CallOptions.Key<Integer> CHANNEL_ID_KEY =
       CallOptions.Key.create("GcpChannelId");
 
+  /** CallOptions key for sticky channel routing without affinity-key map state. */
```
olavloite, line 110: Overall, there are quite big changes to grpc-gcp being made in this pull request, but there are no tests that verify these changes. Can we add tests that cover the changes that we make to grpc-gcp here? Relying on tests in the Spanner client is not enough, as this is a standalone library that can be used by other clients.

olavloite: The test coverage for the changes to the Spanner client is also quite thin, but the existing tests generally do cover these changes. One interesting test (if possible) would be a test that really verifies that all requests in a single read/write or multi-use read-only transaction really all use the same gRPC channel (so basically checking the local port where the requests are coming from on the mock server).

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
rahul2393: Yes, it's intended.
For Omni, irrespective of whether it's a strong/stale single/multi-use read-only, each RPC can land to the server based on the routing hint, so pinning to the same endpoint for read-only transactions is not needed.

Only read-write transactions need pinning to the same endpoint, and since there is only one gRPC channel per endpoint, there is no need to maintain affinities, so we can safely remove it.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/KeyAwareChannel.java
rahul2393: Yes not needed.

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
rahul2393: Added some more tests

grpc-gcp-java/src/main/java/com/google/cloud/grpc/GcpManagedChannel.java
rahul2393: added

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java
rahul2393: Intended

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
olavloite: Hmm... OK. But this code is for cleaning up the various cache entries. KeyAwareChannel also caches preferLeader routing metadata inside readOnlyTxPreferLeader. Without calling rpc.clearTransactionAffinity(id), those entries are not being cleaned up, right? (There seems to have been a test for this, but that test is also being removed/refactored in this PR, so that it does not catch this. Can we re-instate that test, assuming that readOnlyTxPreferLeader entries should be cleaned up?)

olavloite (approved):
The entire

PR #13233, opened 2026-05-20
feat(spanner): add option for inline begin with multi-use read only txn

rahul2393:
- Add optional inline begin support for multi-use read-only transactions.
  - Preserve explicit BeginTransaction as default behavior.

### DIFF
--- a/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
+++ b/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
@@ -423,6 +423,8 @@ abstract class AbstractReadContext
     void beforeReadOrQuery() {
       super.beforeReadOrQuery();
       if (shouldUseInlinedBegin()) {
+        // Keep the same nested transaction guard as the explicit BeginTransaction path. This checks
+        // TransactionRunner's thread-local pending state, not the session's active transaction.
         SessionImpl.throwIfTransactionsPending();
       } else {
         initTransaction();
@@ -434,7 +436,8 @@ abstract class AbstractReadContext
     TransactionSelector getTransactionSelector() {
       if (!shouldUseInlinedBegin()) {
         // No need for synchronization: super.readInternal() is always preceded by a check of
-        // "transactionId" that provides a happens-before from initialization, and the value is never
+        // "transactionId" that provides a happens-before from initialization, and the value is
+        // never
         // changed afterwards.
         @SuppressWarnings("GuardedByChecker")
         TransactionSelector selector =
@@ -461,19 +464,18 @@ abstract class AbstractReadContext
 
       try {
         return TransactionSelector.newBuilder()
-            .setId(
-                futureToWaitFor.get(
-                    WAIT_FOR_INLINE_BEGIN_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS))
+            .setId(futureToWaitFor.get(WAIT_FOR_INLINE_BEGIN_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS))
             .build();
       } catch (ExecutionException e) {
         throw SpannerExceptionFactory.asSpannerException(e.getCause());
       } catch (TimeoutException e) {
         throw SpannerExceptionFactory.newSpannerException(
-            ErrorCode.ABORTED,
+            ErrorCode.DEADLINE_EXCEEDED,
             "Timeout while waiting for an inlined read-only transaction to be returned by another"
                 + " statement.",
             e);
       } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
         throw SpannerExceptionFactory.newSpannerExceptionForCancellation(null, e);
       }
     }
@@ -620,6 +622,13 @@ abstract class AbstractReadContext
                 "ResultSet was closed before a read-only transaction id was returned"));
       }
       super.onDone(withBeginTransaction);
+    }
+
+    @Override
+    void onStartFailed(boolean withBeginTransaction, Throwable t) {
+      if (withBeginTransaction) {
+        failTransactionIdFuture(t);
+      }
     }
 
     private void failTransactionIdFuture(Throwable t) {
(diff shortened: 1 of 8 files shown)


rahul2393:
Screenshot 2026-05-20 at 5 22 56 PM


[Probers](https://github.com/rahul2393/spanner-client-libraries-prober/blob/main/java/probers/src/main/java/com/google/bypass/MultiUseReadOnlyReadProbe.java) running multi read with 2 read at fixed QPS of 10 show improvement of around 600-700us using DirectPath

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -398,21 +415,67 @@ protected boolean isRouteToLeader() {
...
+
     @Override
     void beforeReadOrQuery() {
       super.beforeReadOrQuery();
-      initTransaction();
+      if (shouldUseInlinedBegin()) {
+        SessionImpl.throwIfTransactionsPending();
+      } else {
+        initTransaction();
+      }
```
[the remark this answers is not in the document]
rahul2393: stale review

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -398,21 +415,67 @@ protected boolean isRouteToLeader() {
...
+      try {
+        if (transactionId != null) {
+          return TransactionSelector.newBuilder().setId(transactionId).build();
+        }
+        if (transactionIdFuture == null) {
+          transactionIdFuture = SettableApiFuture.create();
+          return TransactionSelector.newBuilder()
+              .setBegin(createReadOnlyTransactionOptions())
+              .build();
+        }
```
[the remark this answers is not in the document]
rahul2393: After creating an inline-begin selector, both read and query start paths wrap
  the RPC start in a try/catch and call `onStartFailed(withBeginTransaction, t)`, which completes `transactionIdFuture`  exceptionally. `onError()` and `onDone()` also fail the future if the stream starts but no transaction id is returned.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java
```diff
@@ -384,6 +395,32 @@ ServerStream<BatchWriteResponse> batchWriteAtLeastOnce(
...
   ReadOnlyTransaction readOnlyTransaction(TimestampBound bound);
 
+  /**
+   * Returns a read-only transaction context in which multiple reads and/or queries can be performed
+   * at the given timestamp bound and with the given read-only transaction options.
+   *
+   * <p>Options can include:
+   *
+   * <ul>
+   *   <li>{@link Options#beginTransactionOption(Options.BeginTransactionOption)}: Controls whether
```
olavloite, line 422: This part of the documentation should also be added to the method above that does not accept a TimestampBound. Also, I think it would be good to include some arguments for when it makes sense to use this option, and when it does not make sense.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -398,21 +415,69 @@ protected boolean isRouteToLeader() {
...
+        futureToWaitFor = transactionIdFuture;
+      } finally {
+        txnLock.unlock();
+      }
+
+      try {
+        return TransactionSelector.newBuilder()
+            .setId(futureToWaitFor.get(WAIT_FOR_INLINE_BEGIN_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS))
```
olavloite, line 466: This means that when using inline-begin-tx and you run multiple queries concurrently at the start of the transaction, the queries that do not include the BeginTransaction option effectively get a different timeout behavior than today; if the query that starts the transaction is slow, then the other queries could fail. It might be good to include this in the documentation for this option (or maybe more in general for the documentation: Basically advice against using this option when the read-only transaction executes queries in parallel)

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
```diff
@@ -503,6 +568,80 @@ public ListenableAsyncResultSet executeQueryAsync(Statement statement, QueryOpti
           bufferRows);
     }
 
+    @Override
+    public void onTransactionMetadata(Transaction transaction, boolean shouldIncludeId) {
+      Timestamp readTimestamp = null;
+      if (transaction.hasReadTimestamp()) {
+        try {
+          readTimestamp = Timestamp.fromProto(transaction.getReadTimestamp());
```
olavloite, line 575: Note that this change does introduce a behavioral difference between inline-begin and explicit-begin, albeit one that is perfectly valid. With explicit-begin, you can call `getReadTimestamp()` right away starting the transaction. With inline-begin, the `getReadTimestamp()` method will throw an error if you try to call it before at least one `PartialResultSet` has been returned by the first query. This is an additional reason why we cannot make this behavior the default without a major version bump (or we would need to change this behavior into something that would be safe).

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClientImpl.java
```diff
@@ -218,21 +219,26 @@ public ReadOnlyTransaction singleUseReadOnlyTransaction(TimestampBound bound) {
...
+
+  @Override
+  public ReadOnlyTransaction readOnlyTransaction(ReadOnlyTransactionOption... options) {
+    return readOnlyTransaction(TimestampBound.strong(), options);
   }
 
   @Override
   public ReadOnlyTransaction readOnlyTransaction(TimestampBound bound) {
+    return readOnlyTransaction(
+        bound, Options.beginTransactionOption(Options.BeginTransactionOption.EXPLICIT));
```
olavloite: Would it not make more sense to pass in an empty array of read-only transaction options here? By passing in an explicit value here, you are essentially overriding any future changes to the default that we want to use, and/or if we for example want to add a field to `SpannerOptions` for setting a client-wide default.

java-spanner/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionImplTest.java
```diff
@@ -628,6 +629,46 @@ private void mockRead(final PartialResultSet myResultSet) {
             });
   }
 
+  @Test
+  public void multiUseReadOnlyTransactionCanUseInlineBegin() throws ParseException {
```
olavloite, line 667: This is the only test that is added for this entire feature. That is not enough. We should add tests that verify:
1. The happy path (does inline-begin actually do anything?)
2. Unhappy-paths and other non-standard cases: What if the first query fails? What if there are multiple concurrent queries at start? Does this also work for Read operations? Does this work for our async API? What happens if an application calls executeQuery(..) and then closes the ResultSet without ever reading anything?

rahul2393: Added more tests

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbstractReadContext.java
rahul2393: Thanks, updated the comment

olavloite (approved):

PR #13236, opened 2026-05-20
chore(spanner): deprecate experimental host option/parameter to replace with Spanner Omni

sagnghos:
This PR deprecates the `experimentalHost` / `isExperimentalHost` option and property across the Cloud Spanner client libraries (`SpannerOptions` and `ConnectionOptions`) and introduces a unified `type` option and connection parameter (supporting the values `cloud`, `omni`,).

To connect to a Spanner Omni instance, users should now use the `setType(InstanceType.OMNI)` builder option or pass the connection string query parameter `?type=omni`.

Refer to discussion: [Spanner Client Library Configuration for Omni](https://docs.google.com/document/d/1n-rsHkNAEwQbaICHKlFhPl3NN8a3QuOknn_tEhaowZQ/edit?resourcekey=0-LnCt0jSFXCoSuVttphq3FQ&tab=t.0)

Currently, setting the type to `cloud` or `emulator` acts as a **no-op** (since the client library automatically resolves default behaviors for standard cloud and emulator setups). However, introducing this option makes the client library highly extendable, enabling us to enforce or optimize custom behaviors tailored to specific Spanner instance types in the future.


java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1795,26 +1809,49 @@ public Builder setDecodeMode(DecodeMode decodeMode) {
...
+              "Please remove the 'https:' protocol prefix from the host string when using plain text"
+                  + " communication");
+          if (!this.host.startsWith("http")) {
+            this.host = "http://" + this.host;
+            setHost(this.host);
+          }
         }
+        setBuiltInMetricsEnabled(false);
+        super.setProjectId(SPANNER_OMNI_PROJECT_ID);
+        setSessionPoolOption(SessionPoolOptions.newBuilder().setExperimentalHost().build());
```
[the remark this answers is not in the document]
sagnghos: This is intentional

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1795,26 +1809,49 @@ public Builder setDecodeMode(DecodeMode decodeMode) {
...
+     * Specifies the type of Spanner instance to connect to (CLOUD, OMNI, or EMULATOR). Currently,
+     * this is a no-op for most types, but setting it to OMNI is mandatory when connecting to a
+     * Spanner Omni instance.
+     */
+    public Builder setType(InstanceType instanceType) {
+      this.instanceType = instanceType;
+      if (instanceType == InstanceType.OMNI) {
+        if (Strings.isNullOrEmpty(this.host)) {
+          throw new IllegalStateException("Host must be set before setting Type to OMNI");
+        }
```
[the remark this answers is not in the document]
sagnghos: Done

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -314,6 +315,18 @@ enum TracingFramework {
...
   }
 
+  /**
+   * Specifies the type of Spanner instance to connect to (CLOUD, OMNI, or EMULATOR). Currently,
+   * this is a no-op for most types, but setting it to OMNI is mandatory when connecting to a
+   * Spanner Omni instance.
+   */
+  public enum InstanceType {
+    UNSPECIFIED,
+    CLOUD,
```
rahul2393, line 327: We should just put a comment on  UNSPECIFIED that default value will be CLOUD

Put another way, do you see any difference in code path when `UNSPECIFIED` or `CLOUD` is set?

sagnghos: Currently both are no-op. Any value except OMNI is no-op. Removed  the UNSPECIFIED enum as it seems redundant right now. Thanks!

rahul2393 (approved):

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -1202,8 +1223,17 @@ boolean isUsePlainText() {
         || getInitialConnectionPropertyValue(USE_PLAIN_TEXT);
   }
 
-  boolean isExperimentalHost() {
-    return getInitialConnectionPropertyValue(IS_EXPERIMENTAL_HOST);
+  boolean isSpannerOmni() {
+    return getInitialConnectionPropertyValue(IS_EXPERIMENTAL_HOST)
+        || getInitialConnectionPropertyValue(TYPE) == SpannerOptions.InstanceType.OMNI;
```
sakthivelmanii: When someone calls `getInstanceType`, we are calling `getInitialConnectionPropertyValue(TYPE)` twice. One in `getInstanceType` and second in  `isSpannerOmni()`

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -318,6 +319,17 @@ enum TracingFramework {
...
 
+  /**
+   * Specifies the type of Spanner instance to connect to (CLOUD, OMNI, or EMULATOR). Currently,
+   * this is a no-op for most types, but setting it to OMNI is mandatory when connecting to a
+   * Spanner Omni instance.
+   */
+  public enum InstanceType {
+    CLOUD,
+    OMNI,
+    EMULATOR
```
sakthivelmanii: `CLOUD` is default behaviour so no changes are required. I would suggest to support `EMULATOR` in this PR itself or remove it. Why do we need to keep an enum which is a no-op?

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1827,27 +1842,50 @@ public Builder setDecodeMode(DecodeMode decodeMode) {
 
     @Override
     public Builder setHost(String host) {
-      super.setHost(host);
+      this.host = host;
+      if (this.instanceType == InstanceType.OMNI
```
sakthivelmanii, line 1865: probably move this to a method.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
```diff
@@ -1827,27 +1842,50 @@ public Builder setDecodeMode(DecodeMode decodeMode) {
...
+
+    /**
+     * Specifies the type of Spanner instance to connect to (CLOUD, OMNI, or EMULATOR). Currently,
+     * this is a no-op for most types, but setting it to OMNI is mandatory when connecting to a
+     * Spanner Omni instance.
+     */
+    public Builder setType(InstanceType instanceType) {
+      this.instanceType = instanceType;
+      if (instanceType == InstanceType.OMNI) {
+        setBuiltInMetricsEnabled(false);
```
sakthivelmanii: I think someone can still override these since this is happening in the `Builder`. Can we add this in `build()`?

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
```diff
@@ -895,4 +896,28 @@ public Dialect convert(String value) {
...
+
+    @Override
+    public Class<SpannerOptions.InstanceType> getParameterClass() {
+      return SpannerOptions.InstanceType.class;
+    }
+
+    @Override
+    public SpannerOptions.InstanceType convert(String value) {
+      SpannerOptions.InstanceType converted = values.get(value);
+      return converted;
```
sakthivelmanii:
```suggestion
      return values.get(value);
```

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -151,6 +153,7 @@ public class ConnectionOptions {
       new LocalConnectionChecker();
   static final boolean DEFAULT_USE_PLAIN_TEXT = false;
   static final boolean DEFAULT_IS_EXPERIMENTAL_HOST = false;
+  static final SpannerOptions.InstanceType DEFAULT_TYPE = null;
```
sakthivelmanii: Shouldn't this be `CLOUD`? why does it have to be `NULL`

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
```diff
@@ -449,13 +459,16 @@ private Builder() {}
 
     public static final String EXTERNAL_HOST_FORMAT =
```
sakthivelmanii, line 460: Can't we rename this? @olavloite

sakthivelmanii: This is a static public variable.

sakthivelmanii (changes requested):

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
sagnghos: No we first need to uptake the change in java-jdbc - 
 https://github.com/googleapis/google-cloud-java/blob/df67a3c8d8cb9ce9f5c3edfcee226e83fc84dc06/java-spanner-jdbc/src/main/java/com/google/cloud/spanner/jdbc/JdbcDriver.java#L150 then only we can make the change here

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
sagnghos: Done

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ClientSideStatementValueConverters.java
sagnghos: Done

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
sagnghos: Removed EMULATOR

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
sagnghos: Corrected

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
sagnghos: Done

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
sagnghos: Instead of moving to `Builder` I moved it to `SpannerOptions` constructor using a prepareBuilder method. As for cases like `SpannerPool.createSpanner` directly passes the builder into the constructor instead of going via build() as it uses ConnectionSpannerOptions.Builder - https://github.com/googleapis/google-cloud-java/blob/df67a3c8d8cb9ce9f5c3edfcee226e83fc84dc06/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/SpannerPool.java#L417

I have added a unit test SpannerOptionsTest#testOmniEnforcedInConstructor to catch this specific case

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionOptions.java
sakthivelmanii: Sure. Please create a ticket to deprecate setExperimentalHost method and this variable and other references to ExperimentalHost

sagnghos: I already have a P0 tracking bug to deprecate from all libraries - b/509773235 as we intend to deprecate all references of experimental host in public libraries by GA

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java
sakthivelmanii: Makes sense.

sakthivelmanii (approved):

PR #13238, opened 2026-05-20
fix(bqjdbc): pass rowsInPage with TableResult

Neenu1995:
b/511229401

This PR introduces the `rowsInPage` metadata to `TableResult` and ensures it is populated for all pages (both the first page and subsequent pages fetched via pagination).

Previously, metadata about the number of rows in the page was not easily accessible or propagated across multiple paginated requests (e.g., when calling `getNextPage()`). This change:
1. Adds a `rowsInPage` field and builder method to the `@AutoValue` `TableResult` class.
2. Populates `rowsInPage` during initial page creation in `BigQueryImpl` and empty results in `Job.java`.
3. Updates `TableResult.getNextPage()` to calculate the size of values for subsequent pages via `Iterables.size` (which is $O(1)$ for materialized lists) and sets it on the returned `TableResult` pages.
4. Updates unit test assertions in `TableResultTest.java` and `SerializationTest.java` to verify that `rowsInPage` is populated and propagated correctly.

## Key Changes
- **`TableResult.java`**: Added `getRowsInPage()`, builder method `setRowsInPage()`, and updated `getNextPage()` to pass down the calculated row count for the next page.
- **`BigQueryImpl.java`**: Set `rowsInPage` on initial query results page creation.
- **`TableResultTest.java`**: Added assertions to verify that `rowsInPage` is correctly populated for both the first page and subsequent pages.


java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1685,6 +1685,7 @@ && getOptions().getOpenTelemetryTracer() != null) {
           .setSchema(schema)
           .setTotalRows(data.y())
           .setPageNoSchema(data.x())
+          .setRowsInPage(data.x() != null ? (long) Iterables.size(data.x().getValues()) : 0L)
```
logachev: Can we guarantee somehow that it would be using `.size()` to avoid O(N) here?

Neenu1995: We can guarantee $O(1)$ performance because the row values in TableResult pages are always materialized into an ImmutableList under the hood. Since ImmutableList implements Collection, any size check—whether through our explicit instanceof check or Guava's Iterables.size—calls the list's size() method directly. This allows Java to look up the count instantly instead of looping through each row one by one.

But if we want we can add an explicit instanceof check here. Should I add it?

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
```diff
@@ -49,6 +49,8 @@ public abstract static class Builder {
 
     public abstract TableResult.Builder setJobCreationReason(JobCreationReason jobCreationReason);
 
+    public abstract TableResult.Builder setRowsInPage(Long rowsInPage);
```
lqiu96: nit: Is is possible to keep this setter package-private for now? I think it's only used internally and we can make it public if there is a need.

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
```diff
@@ -94,12 +100,18 @@ public String getNextPageToken() {
   @Override
   public TableResult getNextPage() {
     if (getPageNoSchema().hasNextPage()) {
+      Page<FieldValueList> nextPageNoSchema = getPageNoSchema().getNextPage();
+      long nextRows =
+          nextPageNoSchema.getValues() != null
+              ? (long) Iterables.size(nextPageNoSchema.getValues())
+              : 0L;
```
lqiu96: qq, similar to the question Kirill mentioned above. If `Page.getValues()` is not a Collection, this would be `O(n)` time instead of `O(1)` right? 

Would that end up increasing the overall time? For example O(n) here in the `getNextPage()` call and then O(n) again as the user iterates through the page?

.gitignore
```diff
@@ -86,3 +86,7 @@ monorepo
 *.tfstate.lock.info
 
 .jqwik-database
+
+# Agent configuration and local skills
+.agent/
+.agents/
```
lqiu96: qq, is this intended to be added?

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
Neenu1995: In BigQueryImpl.java, the transformTableData() method eagerly copies the rows for each page  into an ImmutableList (which implements Collection). So,

1. getValues() always returns a collection.
2. Iterables.size() resolves to Collection.size(), making it a pure $O(1)$ lookup.
3. It does not duplicate iteration or risk consuming any lazy stream.

.gitignore
Neenu1995: removed.

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
Neenu1995: done.

Neenu1995:
@lqiu96 gentle ping

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
lqiu96: Gotcha. If getValues() always returns a collection, do we still need a null check?

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
lqiu96: Similar to below will `data.x()` ever be null? Or would it just return something like an empty collection. I see below that `getRowsInPage` is marked as nullable and getRowsInPage == null is still possible?

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
Neenu1995: You are right. This null check is not really necessary. Removed.

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
Neenu1995: Removed this null check as well.

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
```diff
@@ -81,6 +83,10 @@ public static Builder newBuilder() {
   @Nullable
   public abstract JobCreationReason getJobCreationReason();
 
+  /** Returns the number of rows in the current page of results. */
+  @Nullable
```
lqiu96, line 87: nit: Can this be null now? I think we already check in the setter to ensure that will always default to either 0 or the page size

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -981,7 +981,9 @@ private boolean meetsReadRatio(TableResult results) {
       return false;
     }
 
-    long pageSize = querySettings.getMaxResultPerPage();
+    Long rowsInPage = results.getRowsInPage();
+    long pageSize =
+        (rowsInPage != null && rowsInPage > 0) ? rowsInPage : querySettings.getMaxResultPerPage();
```
lqiu96, line 986: nit: I don't think `rowsInPage` can be null now since we default this to be 0.

Also, do we still need the `rowsInPage > 0` check? I assumed it was to avoid the divide by 0 error possibility, but below we set the pageSize to 1 if that is the case

lqiu96 (approved):
added some small comments, but the PR LGTM

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/TableResult.java
Neenu1995: Removing `@Nullable` makes `rowsInPage` a required property, which throws an `IllegalStateException` for any existing unit tests or external client calls that build `TableResult` without setting it. Since `setRowsInPage` is package-private, external callers also have no way to set this value to prevent the exception.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
Neenu1995: Even though production paths populate the field, `rowsInPage` can still be null in unit tests/mocks because `getRowsInPage()` remains annotated with `@Nullable`. The `rowsInPage > 0` check is required so that empty pages fall back to the configured max page size rather than defaulting to `1`, which would trigger the Read API prematurely. Finally, the defensive check below handles cases where the user explicitly configures the page size to an invalid non-positive number.

PR #13241, opened 2026-05-20
feat(spanner): add setMetricsProjectId to fix metrics export

### LINKED ISSUE #13240: Spanner: client-side metrics export uses wrong project
## Bug

When `SpannerOptions.newBuilder().build()` is used without an explicit `.setProjectId()`, the project ID defaults to the GCP metadata server value — which on GKE with shared VPC is the **host project** (e.g., `a-gke-project-unrelated`), not the application project.

`DatabaseId.of("my-app-project", instance, db)` correctly routes the database connection to the application project. However, the **client-side built-in metrics export** uses `SpannerOptions.getProjectId()` to determine the Cloud Monitoring target project. This causes metrics to be exported to the host project, which the service account typically lacks `monitoring.metricWriter` permission on.

## Environment

- GKE with shared VPC (host project != application project)
- Spanner Java client with built-in metrics enabled (default in recent versions)
- Service account has Spanner + Monitoring permissions on the application project only

## Error

```
createServiceTimeSeries request failed for spanner metrics.
Need monitoring metric writer permission on project=a-gke-project-unrelated
```

## Impact

- Silent metric data loss
- Repeated permission errors in logs
- Potential memory pressure from accumulated retry buffers
- Affects any Java Spanner client on GKE shared VPC without explicit `.setProjectId()`

## Workaround

Explicitly set `.setProjectId()` on `SpannerOptions.newBuilder()`:

```java
SpannerOptions options = SpannerOptions.newBuilder()
    .setProjectId("my-app-project")
    .build();
```

## Suggested fix

Add a `setMetricsProjectId(String)` to `SpannerOptions.Builder` that allows configuring the metrics export project independently. When not set, fall back to `getProjectId()`. Additionally, log a warning in `getDatabaseClient()` when the `DatabaseId` project differs from the metrics project.

Reference implementation: https://github.com/RRap0so/google-cloud-java/tree/fix/metrics-project-default

RRap0so:
Fixes #13240

## Problem

On GKE with shared VPC, `SpannerOptions.getProjectId()` defaults to the host project via the metadata server. While `DatabaseId` correctly routes database operations to the application project, the client-side metrics exporter uses `getProjectId()` — causing `createServiceTimeSeries` to target the wrong project and fail with permission errors.

## Changes

- Add `setMetricsProjectId(String)` to `SpannerOptions.Builder`
- Add `resolveMetricsProjectId()` that prefers explicit value, falls back to `getProjectId()`
- Update 3 metrics export call sites to use `resolveMetricsProjectId()`
- Log `WARNING` in `getDatabaseClient()` when `DatabaseId` project differs from metrics project

## Usage

```java
SpannerOptions options = SpannerOptions.newBuilder()
    .setMetricsProjectId("my-app-project")
    .build();
```

When not set, behavior is unchanged (uses `getProjectId()`).


kunheeha (approved):

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -306,6 +306,17 @@ public DatabaseClient getDatabaseClient(DatabaseId db) {
       if (dbClients.containsKey(db)) {
         return dbClients.get(db);
       } else {
+        SpannerOptions opts = getOptions();
```
sakthivelmanii: can we do this in SpannerOptions#build?

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java
```diff
@@ -306,6 +306,17 @@ public DatabaseClient getDatabaseClient(DatabaseId db) {
...
+        String metricsProject = opts.resolveMetricsProjectId();
+        if (opts.isEnableBuiltInMetrics()
+            && metricsProject != null
+            && !metricsProject.equals(db.getInstanceId().getProject())) {
+          logger.log(
+              Level.INFO,
+              "Built-in metrics project ''{0}'' differs from DatabaseId project ''{1}''."
+                  + " Use setMetricsProjectId() to override if needed.",
```
sakthivelmanii: only if you override the metrics project ID using `setMetricsProjectId` then only it differs from database project. I feel this log is misleading.

sakthivelmanii (changes requested):
Can you address these comments and rebase it with the latest main?

RRap0so:
Hey @sakthivelmanii, thank you for the review. Simplified it a bit and removed the log line probably not necessary.

rahul2393:
@RRap0so thanks for raising the PR, can you check if this fix https://github.com/googleapis/google-cloud-java/pull/13262 is sufficient to solve your issue? by exposing a public method it is possible to export built-in metric for a database to separate project(though we can log warning) which is not expected, and the fix I shared handled those cases without exposing an option.

RRap0so:
@rahul2393 you folks will know best. If it's not intended for users to define a different Metrics Project ID then your PR should do it.

rahul2393:
Ok, From spanner side we ideally would not like to expose that as public setter, let me try getting reviews on the PR, if it's not reviewed or I see concern from team, we will go ahead with the public setter

sakthivelmanii:
Fixed it as part of https://github.com/googleapis/google-cloud-java/pull/13262

Issue #3277, opened 2018-05-16
Exception in thread "grpc-default-executor-1" java.lang.IllegalArgumentException : Jetty ALPN/NPN has not been properly configured.         at io.grpc.netty.GrpcSslContexts.selectApplicationProtocolConfig(GrpcSsl Contexts.java:153)


mandeep1443:
May 16, 2018 11:43:39 AM com.example.flexible.speak.TranscribeSocket onWebSocket
Text
INFO: Got sampleRate: 48000
May 16, 2018 11:43:39 AM io.grpc.internal.ManagedChannelImpl <init>
INFO: [ManagedChannelImpl@117866a] Created with target speech.googleapis.com:443

Exception in thread "grpc-default-executor-1" java.lang.IllegalArgumentException
: Jetty ALPN/NPN has not been properly configured.
        at io.grpc.netty.GrpcSslContexts.selectApplicationProtocolConfig(GrpcSsl
Contexts.java:153)
        at io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:130)
        at io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:119)
        at io.grpc.netty.GrpcSslContexts.forClient(GrpcSslContexts.java:90)
        at io.grpc.netty.NettyChannelBuilder.createProtocolNegotiator(NettyChann
elBuilder.java:265)
        at io.grpc.netty.NettyChannelBuilder$NettyTransportFactory.newClientTran
sport(NettyChannelBuilder.java:324)
        at io.grpc.internal.CallCredentialsApplyingTransportFactory.newClientTra
nsport(CallCredentialsApplyingTransportFactory.java:62)
        at io.grpc.internal.TransportSet.startNewTransport(TransportSet.java:215
)
        at io.grpc.internal.TransportSet.obtainActiveTransport(TransportSet.java
:192)
        at io.grpc.internal.ManagedChannelImpl$3.getTransport(ManagedChannelImpl
.java:651)
        at io.grpc.internal.ManagedChannelImpl$3.getTransport(ManagedChannelImpl
.java:592)
        at io.grpc.DummyLoadBalancerFactory$DummyLoadBalancer$1.get(DummyLoadBal
ancerFactory.java:135)
        at io.grpc.internal.DelayedClientTransport$2.run(DelayedClientTransport.
java:262)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.
[...]

PR #13243, opened 2026-05-20
fix(gax): adjust RetryAlgorithm logic for timeout v. exception

### LINKED ISSUE #12373: [sdk-platform-java] flaky test: showcase test ITLongRunningOperation
```
Error:  Failures: 
Error:    ITLongRunningOperation.testHttpJson_LROUnsuccessfulResponse_exceedsTotalTimeout_throwsDeadlineExceededException:187 Unexpected exception type thrown, expected: <java.util.concurrent.CancellationException> but was: <java.util.concurrent.ExecutionException>
```
full log: https://github.com/googleapis/sdk-platform-java/actions/runs/18358328703/job/52295976819

whowes:
The current RetryAlgorithm logic is susceptible to an edge case if a short RPC deadline is [set near the end of polling](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java#L161).  If an exception results from a too-short deadline, the exception thrown by the associated Future will be an ExecutionException rather than e.g. CancellationException (for LRO polling). This likely causes flakiness observed in #12373 and [#3277.](https://github.com/googleapis/sdk-platform-java/issues/3277)

This change ensures that shouldRetryBasedOnTiming executes if the operation hasn't completed successfully, even when shouldRetryBasedOnResult is false (e.g. due to non-retriable RPC exception due to the short deadline). This way a timeout-related exception will be thrown instead if the overall timeout has been reached.

Resolves #12373

### DIFF
--- a/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetryAlgorithm.java
+++ b/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetryAlgorithm.java
@@ -232,14 +232,17 @@ public class RetryAlgorithm<ResponseT> {
       ResponseT previousResponse,
       TimedAttemptSettings nextAttemptSettings)
       throws CancellationException {
-    boolean retryBasedOnResult = shouldRetryBasedOnResult(context, previousThrowable, previousResponse);
+    boolean retryBasedOnResult =
+        shouldRetryBasedOnResult(context, previousThrowable, previousResponse);
 
-    // Short-circuit when operation has succeeded to avoid erroneously throwing CancellationException below
+    // Short-circuit when operation has succeeded to avoid erroneously throwing
+    // CancellationException below
     if (!retryBasedOnResult && previousThrowable == null) {
       return false;
     }
 
-    boolean retryBasedOnTiming = shouldRetryBasedOnTiming(context, nextAttemptSettings); // throws CancellationException
+    // throws CancellationException
+    boolean retryBasedOnTiming = shouldRetryBasedOnTiming(context, nextAttemptSettings);
     return retryBasedOnResult && retryBasedOnTiming;
   }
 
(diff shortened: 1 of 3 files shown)


sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetryAlgorithm.java
```diff
@@ -153,19 +153,25 @@ public TimedAttemptSettings createNextAttempt(
...
-    TimedAttemptSettings newSettings =
-        createNextAttemptBasedOnResult(
-            context, previousThrowable, previousResponse, previousSettings);
-    if (newSettings == null) {
-      newSettings = createNextAttemptBasedOnTiming(context, previousSettings);
+    // If we shouldn't retry based on result and the last attempt failed with an
+    // exception, defer to any exception thrown by shouldRetryBasedOnTiming.
+    // This lets a timeout-related exception get propagated when justified
+    // rather than e.g. exceptions caused by very short RPC deadlines on a
+    // final polling attempt.
```
lqiu96, line 169: qq, I'm not sure I'm following this comment. I might not be understanding this fully.

 What do you mean by `defer to any exception thrown by `shouldRetryBasedOnTiming`? Is this referring to an exception thrown by the ExponentialRetryAlgorithm?

IIRC, short RPC deadlines that can't be processed in time will result in a `DEADLINE_EXCEEDED` status code and that may or may not be retried as a retryable status code

whowes: **"Is this referring to an exception thrown by the ExponentialRetryAlgorithm?"**  Not by ExponentialRetryAlgorithm itself, but by its subclass [OperationTimedPollAlgorithm](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java#L86-L102) used for LRO polling, which throws a CancellationException when the cumulative timeout for all polls is exhausted. 

With the current structure of RetryAlgorithm.createNextAttempt() the CancellationException-throwing codepath doesn't get reached if the final LRO polling RPC fails. That results in an ExecutionException getting propagated instead, which is misleading when that failure is triggered by an [overly-truncated deadline](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java#L150-L162). The mismatch of expectations is IMO likely at the root of the test flakes from #12373 and [#3277](https://github.com/googleapis/sdk-platform-java/issues/3277).

PR #13244, opened 2026-05-20
chore(bqjdbc): log connection properties set by the user

Neenu1995:


java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -152,6 +152,24 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
     this.connectionId = UUID.randomUUID().toString();
     try (BigQueryJdbcMdc.MdcCloseable mdc = BigQueryJdbcMdc.registerInstance(this.connectionId)) {
       this.connectionUrl = url;
+      if (LOG.isLoggable(java.util.logging.Level.CONFIG)) {
+        Properties connectionProps = ds.createProperties();
+        Properties maskedProps = new Properties();
+        for (String name : connectionProps.stringPropertyNames()) {
```
logachev, line 223: I'd suggest using reverse approach. Instead of "opt-out" have an "opt-in". 
We can create a list of properties that are "safe" to log.

Neenu1995: Wouldn't that be a long list and one more place to add property name when a new connection is added. We already have multiple lists of properties and I didn't want to add one more just for logging.

Neenu1995: I made the changes.

java-bigquery/google-cloud-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -76,6 +77,68 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
   private final String connectionId;
   private static final String DEFAULT_JDBC_TOKEN_VALUE = "Google-BigQuery-JDBC-Driver";
   private static final String DEFAULT_VERSION = "0.0.0";
+  private static final Set<String> SAFE_TO_LOG_PROPERTIES =
```
logachev, line 80: Can we use ImmutableSortedSet instead, it allows case insensitive `contains`, so no need to `lower()` every time

```
ImmutableSortedSet<String> SAFE_TO_LOG_PROPERTIES = ImmutableSortedSet.orderedBy(String.CASE_INSENSITIVE_ORDER)
.add(BigQueryJdbcUrlUtility.PROJECT_ID_PROPERTY_NAME)
.add(...)
.build()
```

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryConnectionTest.java
```diff
@@ -456,4 +456,54 @@ public void testIsReadOnlyTokenProvided(String readonlyProp, boolean expectedIsR
       assertEquals(expectedIsReadOnly, connection.isReadOnlyTokenUsed());
     }
   }
+
+  @Test
+  public void testConnectionPropertiesLoggingAndMasking() throws IOException, SQLException {
+    java.util.logging.Logger rootLogger = BigQueryJdbcRootLogger.getRootLogger();
```
logachev: We have `BigQueryJdbcLoggingBaseTest` to handle some of that setup

logachev (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryConnectionTest.java
```diff
@@ -456,4 +456,54 @@ public void testIsReadOnlyTokenProvided(String readonlyProp, boolean expectedIsR
...
+          public void close() throws SecurityException {}
+        };
+    rootLogger.addHandler(handler);
+
+    try {
+      String url =
+          "jdbc:bigquery://https://www.googleapis.com/bigquery/v2:443;"
+              + "OAuthType=2;ProjectId=MyTestProjectId;"
+              + "OAuthAccessToken=secretAccessToken;Location=US;"
+              + "PartnerToken=GPN:secretPartnerToken;";
```
keshavdandeva: I dont see any assertion for partner token.

keshavdandeva (approved):

java-bigquery/google-cloud-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryConnectionTest.java
Neenu1995: I intentionally removed it because it exposed the partner token format easily. I should clean this up as well.

PR #13246, opened 2026-05-21
feat(gax): Implement cert-rotation retries for grpc and http-json.

vverman:


sdk-platform-java/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/RefreshingHttpJsonChannel.java
```diff
@@ -0,0 +1,233 @@
...
+
+  public RefreshingHttpJsonChannel(Supplier<ManagedHttpJsonChannel> channelFactory) {
+    this.channelFactory = channelFactory;
+    this.activeEntry = new AtomicReference<>(new ChannelEntry(channelFactory.get()));
+  }
+
+  @Override
+  public void refresh() {
+    // 1. Lock-free CAS coalescing check to prevent duplicate queueing/blocking of concurrent threads
+    if (!refreshInProgress.compareAndSet(false, true)) {
```
blakeli0: I'm not sure lock free is the best way. Because this mean other failed request would skip this refreshing block and be scheduled with a retry, which would fail again if refreshing is still in progress.

sdk-platform-java/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/RefreshingHttpJsonChannel.java
blakeli0, line 1: Looks good overall. When we actually get to implementation phase, we need test it very carefully to prevent resource leaking.

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiResultRetryAlgorithm.java
```diff
@@ -51,6 +50,16 @@ public boolean shouldRetry(Throwable previousThrowable, ResponseT previousRespon
   @Override
   public boolean shouldRetry(
       RetryingContext context, Throwable previousThrowable, ResponseT previousResponse) {
+    if (previousThrowable instanceof UnauthenticatedException) {
+      if (context instanceof ApiCallContext) {
+        TransportChannel transportChannel = ((ApiCallContext) context).getTransportChannel();
+        if (transportChannel != null && transportChannel.shouldRefresh()) {
+          transportChannel.refresh();
```
lqiu96, line 57: IIUC, I think this should be done in the AttemptCallable itself, right? Other callable types have the refresh logic Bidi, Client, and Server. Otherwise this may potentionally refresh twice for those.

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -425,6 +446,74 @@ private void refreshSafely() {
...
+  boolean shouldRefresh() {
+    String certPath = getWorkloadCertPath();
+    if (certPath == null) {
+      return false;
+    }
+    String currentDiskFingerprint = getOrUpdateDiskFingerprint(certPath);
+    if (currentDiskFingerprint.isEmpty()) {
+      return false;
+    }
+    return !currentDiskFingerprint.equalsIgnoreCase(activeCertFingerprint);
```
lqiu96, line 514: IIUC, this is the new gated logic right? Basically to ensure that we only check the fingerprint if the cert exists on the system. No need to have the rotation if the file doesn't exist

sdk-platform-java/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/RefreshingHttpJsonChannel.java
```diff
@@ -0,0 +1,319 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * An implementation of {@link ManagedHttpJsonChannel} that supports dynamic mTLS certificate
+ * rotation by thread-safely hot-swapping the underlying active HTTP/JSON channel while gracefully
+ * retiring older connections after all active in-flight requests complete.
+ */
+@InternalApi
+public class RefreshingHttpJsonChannel extends ManagedHttpJsonChannel {
```
lqiu96, line 53: IIUC, this is just a 1:1 copy of the gRPC ChannelPool logic?

sdk-platform-java/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -425,6 +446,74 @@ private void refreshSafely() {
...
+        return cached.fingerprint;
+      }
+      String fingerprint = getCertificateFingerprint(certPath);
+      lastDiskCheck.set(new DiskCheckResult(fingerprint, System.nanoTime()));
+      return fingerprint;
+    }
+  }
+
+  boolean shouldRefresh() {
+    String certPath = getWorkloadCertPath();
```
blakeli0, line 506: I think we need to split this shouldRefresh logic to two parts
1. getWorkloadCertPath() should be a static field loaded on app start up. So that we can comfortably skip refreshing for most cases.
2. Get current finger print and compare with the active ones. This part should also be moved to a transport neutral place since I see the logic is duplicated in RefreshingHttpJsonChannel.

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiResultRetryAlgorithm.java
blakeli0: Agreed. Also it is mixing concerns by putting refreshing channels logic in a `shouldRetry` method.

sdk-platform-java/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/RefreshingHttpJsonChannel.java
vverman: Done!

PR #13262, opened 2026-05-25
fix(spanner): derive built-in metrics project from database client

### LINKED ISSUE #13240: Spanner: client-side metrics export uses wrong project

rahul2393:
Fixes https://github.com/googleapis/google-cloud-java/issues/13240

### Problem

  The built-in-metrics Cloud Monitoring exporter can initialize before any DatabaseClient exists, so the OpenTelemetry Resource may carry the host/GKE project instead of the Spanner project — metrics get attributed to the wrong project. A  single resource-level project also misattributes metrics for multi-project clients (RPCs to databases in other  projects).

 ### Solution

  Route each metric by its point-level project_id instead of the resource project:

  1. HeaderInterceptor adds project_id to its existing cached built-in attribute map (cache-miss only, only when parsed from the resource prefix and not the undefined-project sentinel). No new parsing or per-RPC work.
  2. The exporter ignores the resource-level project_id, derives each TimeSeries' project from the point, groups by
  project, and calls createServiceTimeSeries once per project (still batched at the 200 quota limit).
  3. Fallback for points with no usable project is meter-scoped: grpc-java/grpc-gcp core metrics use a fallback project (first DatabaseClient/BatchClient project, set in initializeBuiltInMetrics); Spanner/GAX DB-scoped metrics are skipped rather than exported under an unrelated project. No fallback yet → skip.

#### Notes / trade-offs

  - gRPC/core metrics for multi-project clients are attributed to the first client's project (they carry no point project).
  - gRPC/core metrics before the first data/batch client, or in admin-only processes, are dropped.
  - Per-project export-failure logging (once per project, with PermissionDeniedException setup guidance + project id; cleared on a successful export).

### DIFF
--- a/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsProvider.java
+++ b/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsProvider.java
@@ -78,6 +78,7 @@ final class BuiltInMetricsProvider {
   private OpenTelemetry openTelemetry;
   private String projectId;
   private boolean mismatchedProjectIdLogged;
+  private Thread shutdownHook;
 
   private BuiltInMetricsProvider() {}
 
@@ -96,7 +97,8 @@ final class BuiltInMetricsProvider {
         sdkMeterProviderBuilder.setResource(Resource.create(createResourceAttributes(projectId)));
         SdkMeterProvider sdkMeterProvider = sdkMeterProviderBuilder.build();
         this.openTelemetry = OpenTelemetrySdk.builder().setMeterProvider(sdkMeterProvider).build();
-        Runtime.getRuntime().addShutdownHook(new Thread(sdkMeterProvider::close));
+        this.shutdownHook = new Thread(sdkMeterProvider::close);
+        Runtime.getRuntime().addShutdownHook(this.shutdownHook);
       }
       return this.openTelemetry;
     } catch (IOException ex) {
@@ -116,8 +118,9 @@ final class BuiltInMetricsProvider {
       mismatchedProjectIdLogged = true;
       logger.log(
           Level.WARNING,
-          "Built-in metrics are already initialized for project {0}. Metrics for project {1} will"
-              + " be exported using the existing project.",
+          "Built-in metrics fallback project is already initialized to project {0}. Non-Spanner"
+              + " metrics without project information will be exported using that project instead"
+              + " of project {1}.",
           new Object[] {this.projectId, projectId});
     }
   }
@@ -127,7 +130,6 @@ final class BuiltInMetricsProvider {
     return this.openTelemetry;
   }
 
-  @VisibleForTesting
   synchronized String getProjectId() {
     return this.projectId;
   }
@@ -137,9 +139,17 @@ final class BuiltInMetricsProvider {
     if (this.openTelemetry instanceof OpenTelemetrySdk) {
       ((OpenTelemetrySdk) this.openTelemetry).getSdkMeterProvider().close();
     }
+    if (this.shutdownHook != null) {
+      try {
+        Runtime.getRuntime().removeShutdownHook(this.shutdownHook);
+      } catch (IllegalStateException ignored) {
+        // The JVM is already shutting down.
+      }
+    }
     this.openTelemetry = null;
     this.projectId = null;
     this.mismatchedProjectIdLogged = false;
+    this.shutdownHook = null;
   }
 
   // TODO: Remove when
--- a/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsProvider.java
+++ b/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsProvider.java
@@ -32,6 +32,7 @@ import com.google.auth.Credentials;
 import com.google.cloud.opentelemetry.detection.AttributeKeys;
 import com.google.cloud.opentelemetry.detection.DetectedPlatform;
 import com.google.cloud.opentelemetry.detection.GCPPlatformDetector;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.hash.HashFunction;
 import com.google.common.hash.Hashing;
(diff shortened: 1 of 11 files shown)


java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/BuiltInMetricsProvider.java
```diff
@@ -88,7 +91,7 @@ OpenTelemetry getOrCreateOpenTelemetry(
         SdkMeterProviderBuilder sdkMeterProviderBuilder = SdkMeterProvider.builder();
         BuiltInMetricsView.registerBuiltinMetrics(
             SpannerCloudMonitoringExporter.create(
-                projectId, credentials, monitoringHost, universeDomain),
+                this::getProjectId, credentials, monitoringHost, universeDomain),
```
surbhigarg92, line 95: @rahul2393 I did not understand this solution. getOrCreateOpenTelemetry is called from GapicSpannerRPC while creating SpannerClient. At the time `projectId` shared here could be the projectId of GKE instance for example. 

So in this case we will be initialising SpannerCloudMonitoringExporter with null projectId ? As by this time `setProjectIdIfAbsent` won't be called, it is called later during database init.

surbhigarg92, line 95: We are also passing the projectId in next line to create OpenTelemetry Resource

rahul2393: So flow is:

  1. `SpannerClient` init → OpenTelemetry/exporter may be created, project supplier returns `null`
  2. no export happens yet because no database project is known
  3. `getDatabaseClient(DatabaseId)` → database project is set once
  4. future metric exports use that database project

rahul2393: We also do
  ```java
  monitoredResourceBuilder.putLabels(PROJECT_ID_KEY.getKey(), projectId);
```
 So the resource created during SDK initialization may contain the early/default project, but before sending
  createServiceTimeSeries, we overwrite the monitored resource label with the database project from the exporter supplier.

File: java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java [hunk 2]
```diff
@@ -140,37 +152,20 @@ MetricServiceClient getMetricServiceClient() {
 
   /** Export client built in metrics */
   private CompletableResultCode exportSpannerClientMetrics(Collection<MetricData> collection) {
-    // Filter spanner metrics. Only include metrics that contain a valid project.
```
olavloite, line 143: I don't think you can safely remove this code when using this strategy. Or technically you can, but this then changes the current behavior of clients that use multiple different projects, which I don't think is a side-effect that we want from this fix. Previously, metrics with mismatched project IDs would be filtered out and not exported. Now, they are all set to whatever project ID is used by the first `DatabaseClient` that is created and then exported. I think that the latter is a degradation from the previous behavior, where they were dropped, as now you risk getting metrics from a 'wrong' database into a different project.

An alternative to setting a fixed project ID that is used for all metrics, is to dynamically collect and then batch export the metrics per project ID (which in the vast majority of cases would be just one project). That would remove the requirement to try to set a project ID the first time a `DatabaseClient` is created, which would simplify the code a bit. See here for an example: https://github.com/googleapis/google-cloud-java/compare/spanner-export-metrics-per-project (Note: That sample has not been further refined, so it might need a bit of polishing before it is ready to use)

The up and downsides of the strategy in https://github.com/googleapis/google-cloud-java/compare/spanner-export-metrics-per-project are:
- Upside: It correctly exports Spanner metrics to the correct project, even when a client creates multiple `DatabaseClient`s with different project IDs.
- Downside: It uses the default Project ID of the environment (so for example the GKE project ID) for core gRPC metrics that are not handled by the Spanner interceptor.

The downside mentioned above could partly be mitigated by combining it with the strategy in this pull request, and dynamically setting the project ID that is used for non-Spanner metrics to the project ID of the first DatabaseClient.

java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerCloudMonitoringExporter.java
```diff
@@ -218,16 +227,23 @@ public void onSuccess(List<Empty> empty) {
...
-    return !spannerProjectId.equals(SpannerCloudMonitoringExporterUtils.getProjectId(resource));
-  }
-
-  boolean lastExportSkippedData() {
-    return this.lastExportSkippedData.get();
+  private void logExportFailure(Throwable throwable, ProjectName projectName) {
+    if (spannerExportFailureLoggedProjects.add(projectName.getProject())) {
+      String msg = "createServiceTimeSeries request failed for spanner metrics.";
+      if (throwable instanceof PermissionDeniedException) {
+        // TODO: Add the link of public documentation when available in the log message.
```
olavloite: I think that we can remove this TODO. The link has been added to the log message below.

olavloite (approved):

PR #13267, opened 2026-05-26
fix(bigquery-jdbc): handle EXPORT DATA, EXPORT MODEL, and LOAD DATA statements

keshavdandeva:
b/510498449

This PR resolves a blocking issue where executing `EXPORT DATA`, `EXPORT MODEL`, or `LOAD DATA` statements triggered a `BigQueryJdbcException: Unexpected value: OTHER`. 

### 1. `EXPORT_DATA`
*   Mapped to a new `SqlType.EXPORT` in the converter and added a specific handler in `BigQueryStatement.java`.
*   BigQuery exposes the actual number of rows exported in `QueryStatistics.ExportDataStats.getRowCount()`. We fetch this job statistic and return it as the update count, providing full functionality for this statement.

### 2. `EXPORT_MODEL` & `LOAD_DATA` (Resilient Fallback via `SqlType.OTHER`)
*   We kept these case labels in `BigQuerySqlTypeConverter` but let them fall through to `default` (returning `OTHER`). 
*   We then updated the `OTHER` case in `BigQueryStatement.java` to log a warning and treat it as a standard update statement (returning a `0` update count) instead of crashing.
*   Any new or unmapped SQL statements added by BigQuery in the future will now fall back to this safe `OTHER` flow and succeed with a `0` update count rather than throwing an exception. 
*   The warning log is highly actionable for developers: it logs the **BigQuery Job ID** (for easy lookup in the GCP Console) and a **truncated version of the SQL query** (truncated to 60 characters to prevent log bloat and protect sensitive inline data).

### 3. Architectural Refactoring
*   Consolidated job-retrieval and waiting logic into a shared helper method `getQueryStatisticsFromJob` in `BigQueryStatement.java`.
*   This applies uniform, robust defensive programming (null-checks, `instanceof` checks, and proper thread interruption signal restoration) to **both DML and EXPORT paths**.

### DIFF
--- a/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
+++ b/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
@@ -686,7 +686,10 @@ public class BigQueryStatement extends BigQueryNoOpsStatement {
       case DML:
       case DML_EXTRA:
         QueryStatistics dmlStats = getQueryStatisticsFromJob(results);
-        Long dmlRowCount = (dmlStats != null) ? dmlStats.getNumDmlAffectedRows() : null;
+        Long dmlRowCount =
+            (dmlStats != null && dmlStats.getNumDmlAffectedRows() != null)
+                ? dmlStats.getNumDmlAffectedRows()
+                : 0L;
         updateAffectedRowCount(dmlRowCount);
         break;
       case TCL:
@@ -731,12 +734,12 @@ public class BigQueryStatement extends BigQueryNoOpsStatement {
         updateAffectedRowCount(exportRowCount);
         break;
       case OTHER:
-        String truncatedQuery =
-            (query != null && query.length() > 60) ? query.substring(0, 60) + "..." : query;
-        String jobId = (results.getJobId() != null) ? results.getJobId().getJob() : "unknown";
+        String truncatedQuery = truncateQuery(query);
+        String id =
+            (results.getJobId() != null) ? results.getJobId().getJob() : results.getQueryId();
         LOG.warning(
-            "Encountered unmapped SQL statement type [Job ID: %s]. Treating as update statement: %s",
-            jobId, truncatedQuery);
+            "Encountered unmapped SQL statement type [Job/Query ID: %s]. Treating as update statement: %s",
+            id, truncatedQuery);
         updateAffectedRowCount(results.getTotalRows());
         break;
     }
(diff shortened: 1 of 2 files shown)


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -685,15 +685,9 @@ void handleQueryResult(String query, TableResult results, SqlType queryType)
...
-          Job completedJob = this.bigQuery.getJob(results.getJobId()).waitFor();
-          JobStatistics.QueryStatistics statistics = completedJob.getStatistics();
-          updateAffectedRowCount(statistics.getNumDmlAffectedRows());
-        } catch (InterruptedException ex) {
-          throw new BigQueryJdbcRuntimeException(ex);
-        } catch (NullPointerException ex) {
-          throw new BigQueryJdbcException(ex);
-        }
+        QueryStatistics dmlStats = getQueryStatisticsFromJob(results);
+        Long dmlRowCount = (dmlStats != null) ? dmlStats.getNumDmlAffectedRows() : null;
```
logachev: Should it be `0L` instead of `null`? Will match `EXPORT` behavior

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -725,8 +719,43 @@ void handleQueryResult(String query, TableResult results, SqlType queryType)
...
+          QueryStatistics.ExportDataStats dataStats = exportStats.getExportDataStats();
+          if (dataStats != null && dataStats.getRowCount() != null) {
+            exportRowCount = dataStats.getRowCount();
+          }
+        }
+        updateAffectedRowCount(exportRowCount);
+        break;
       case OTHER:
-        throw new BigQueryJdbcException(String.format("Unexpected value: " + queryType));
+        String truncatedQuery =
```
logachev: nit: move to separate function & re-use within `logQueryExecutionStart`?

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -725,8 +719,43 @@ void handleQueryResult(String query, TableResult results, SqlType queryType)
...
+            (query != null && query.length() > 60) ? query.substring(0, 60) + "..." : query;
+        String jobId = (results.getJobId() != null) ? results.getJobId().getJob() : "unknown";
```
logachev: nit: in another place we just use `results.getJobId()` for logging, will it properly format it? 
Also `results.getQueryId()` instead of `unknown`?

logachev (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: Yeah

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: Done

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
keshavdandeva: > Also results.getQueryId() instead of unknown?

Yes, nice

> nit: in another place we just use results.getJobId() for logging, will it properly format it?

So, `getJobId()` would output a verbose string containing the project and location (e.g., JobId{project=..., location=..., job=...} but for a user-facing warning log, I guess it makes sense to keep the message clean by printing just the short job name string. WDYT?

logachev: Agreed.

PR #13269, opened 2026-05-26
feat(storage): add checksum validation on json read paths

nidhiii-27:
Enabled default full object checksum validation for the following:
`Storage#readAllBytes(bucket,blob)`
`Storage.downloadTo(blobId, destination)`
`Storage.downloadTo(blob, outputStream)`


java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -919,7 +923,18 @@ public long read(
       }
       MediaHttpDownloader mediaHttpDownloader = req.getMediaHttpDownloader();
       mediaHttpDownloader.setDirectDownloadEnabled(true);
-      req.executeMedia().download(outputStream);
+
+      // Check if this is a full object download (no Range header set)
+      boolean isFullObjectDownload = (req.getRequestHeaders().getRange() == null);
```
Dhriti07: Are we also planning to do checksum for cases where the user specifies a range that spans a full object ?

java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -860,9 +861,12 @@ public byte[] load(StorageObject from, Map<Option, ?> options) {
       if (Option.RETURN_RAW_INPUT_STREAM.getBoolean(options) != null) {
         getRequest.setReturnRawInputStream(Option.RETURN_RAW_INPUT_STREAM.getBoolean(options));
       }
+      HttpResponse response = getRequest.executeMedia();
       ByteArrayOutputStream out = new ByteArrayOutputStream();
-      getRequest.executeMedia().download(out);
-      return out.toByteArray();
+      response.download(out);
```
Dhriti07: Can we also have used wrap here similar to read ?

Dhriti07 (approved):

ShreyasSinha (approved):

PR #13271, opened 2026-05-26
chore: add markers in pom.xml

zhumin8:
Add comment markers in pom.xml for generated proto and grpc artifacts.
These markers are added in parent, BOM and client pom.xml files for each library.

Also included in this PR is a one line fix for sdk-platform-java/hermetic_build/library_generation/owlbot/src/fix_poms.py, so that it ignores comments during extra module detection, and does not produce broken pom.xml files when the original has comments. [45f17c4](https://github.com/googleapis/google-cloud-java/pull/13271/commits/45f17c41dfa4377d6b31f549d95bfcba764b359a).
The consequence of this minimal fix is that comments added will be wiped out when regenerated (see [557b1ed](https://github.com/googleapis/google-cloud-java/pull/13271/commits/557b1ed00c61d4d5f57cd46cf8fb420299a4974d)). This is acceptable as is only during migration and will re-run and add markers as needed.

For https://github.com/googleapis/librarian/issues/6134


zhumin8:
Closing this PR to reopen when generation for release is done to avoid repeating the work.

PR #13272, opened 2026-05-26
chore: adjust download-maven-plugin usages to reduce release flakiness

### LINKED ISSUE #13026: [sdk-platform-java] maven-download-plugin is flaky when building in parallel.
The release job failed due to the following errors
```
00:01:33:876 [ERROR] Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.6.8:wget (download-compliance-suite) on project gapic-showcase: IO Error: OverlappingFileLockException -> [Help 1]
00:01:33:876 [ERROR] Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.6.8:wget (download-common-resources-proto) on project gapic-generator-java: IO Error: java.io.StreamCorruptedException: invalid type code: 00 -> [Help 1]
```

See [error log](https://fusion2.corp.google.com/invocations/088b16e2-e068-46da-be73-8504fd93da75/targets/cloud-sdk%2Fclient-libraries%2Fjava%2Frelease%2Fgoogle-cloud-java%2Frelease/log) for details.

`download-maven-plugin:1.6.8:wget` is not thread safe and could run into errors when building in parallel. We should remove the need of it in build. For this specific case, the protos we are trying to download are only used in tests, we should have a copy of them in test folders instead of always downloading from the internet.

whowes:
As observed in issue #13026, the download-maven-plugin's wget goal is currently flaky in release jobs. 

This change 
- checks in protos for gapic-generator-java instead of downloading them 
- adjusts the plugin configuration for showcase to skip caching to prevent parallel jobs from trying to acquire the same lock

Fixes #13026


sdk-platform-java/java-showcase-3.25.8/gapic-showcase/pom.xml
```diff
@@ -68,6 +68,9 @@
                 https://raw.githubusercontent.com/googleapis/gapic-showcase/v${gapic-showcase.version}/server/services/compliance_suite.json
               </url>
               <outputDirectory>src/test/resources</outputDirectory>
+              <cacheDirectory>
+                ${project.build.directory}/download-cache
+              </cacheDirectory>
```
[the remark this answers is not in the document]
whowes: /gemini PTAL after changes

sdk-platform-java/gapic-generator-java/src/main/proto/gapic/metadata/gapic_metadata.proto
blakeli0, line 1: I'm kind of surprised that it just worked. I though there might be additional configurations needed in the poms. I guess the existing protoc configurations automatically generates everything in `main/proto`. 

Also I think this `gapic_metadata.proto` probably needs a separate solution because it is a googleapis proto and used in production code not tests. For example, we may want to use automations to sync it from [googleapis](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) to [java-common-protos](https://github.com/googleapis/google-cloud-java/tree/main/java-common-protos). 

That being said, for the scope of this PR, I think it's better to still use wget to get the latest gapic_metadata from googleapis and set `skipCache` to true.

whowes: SGTM - I restored wget + compilation for gapic_metadata.proto and reverted the checked-in copy. The restored download execution now has skipCache and overwrite set, mirroring what's done for showcase.

blakeli0 (approved):

PR #13274, opened 2026-05-26
chore(bigquery-jdbc): bump next BQ JDBC version to 1.0.0

logachev:


lqiu96 (approved):
LGTM.

2 small notes/ questions:
- Do you want to match the BQ version (e.g. v2.66.0) or prefer it to be 1.0 version for GA?
- Can you update the conventional commit to something like `feat(bigquery-jdbc): ... `, so the GA change would show up in the release notes?

logachev:
- I don't think we should match BQ version.. We also rely on bq-storage and it's version is different.. I think if we ever decide to align all BQ versions, we'll join, but without it I think it's ok to keep as-is
- Updated, I had my doubts about it, but have no opinion; It seemed to be more like a service change rather than component-specific (e.g. no need to highlight version change in release notes, it is already included)

PR #13282, opened 2026-05-27
feat(bigquery-jdbc): add globalOtel support

keshavdandeva:
b/517124136

This PR introduces support for opting into the global OpenTelemetry instance via connection properties in the BigQuery JDBC driver. It also establishes a clear hierarchy of precedence for various OpenTelemetry configuration flags to ensure predictable behavior

### 1. Global OpenTelemetry Support
*   Added a new connection property `useGlobalOpenTelemetry` (default: `false`).
*   When set to `true`, the driver will explicitly use the globally registered OpenTelemetry instance (`GlobalOpenTelemetry.get()`) for both tracing and logging.

### 2. Precedence Rules for Telemetry Flags
Implemented the following hierarchy of precedence in `BigQueryConnection`:
1.  **`customOpenTelemetry`** (Provided via `DataSource`): Highest precedence. Overrides everything else.
2.  **`useGlobalOpenTelemetry`**: Second precedence. Overrides driver-managed GCP fallback flags.
3.  **`enableGcpTraceExporter` / `enableGcpLogExporter`**: Lowest precedence. Used only if above are false/null.

### 3. Code Modifications
*   **`BigQueryJdbcUrlUtility.java`**: Added property definition and added it to valid properties.
*   **`DataSource.java`**: Added field, getter/setter, and URL parsing support for the new property.
*   **`BigQueryConnection.java`**: Updated constructor to read the new property and implemented the precedence logic for both tracer initialization and logging client registration.
*   **`BigQueryJdbcOpenTelemetry.java`**: Updated `getOpenTelemetry()` to handle the `useGlobalOpenTelemetry` flag.

### 4. Testing
*   **`BigQueryJdbcOpenTelemetryTest.java`**: Added unit test to verify that `useGlobalOpenTelemetry` returns the global instance.
*   **`BigQueryConnectionTest.java`**: Added a comprehensive parameterized test (`testOpenTelemetryPrecedenceHierarchy`) covering all 8 combinations of the 4 flags to verify the correct behavior of `OpenTelemetry` resolution and logging client creation.

### DIFF
--- a/java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryConnectionTest.java
+++ b/java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryConnectionTest.java
@@ -17,15 +17,25 @@
 package com.google.cloud.bigquery.jdbc;
 
 import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 
 import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.api.gax.rpc.HeaderProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
+import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.QueryJobConfiguration.JobCreationMode;
 import com.google.cloud.bigquery.exception.BigQueryJdbcException;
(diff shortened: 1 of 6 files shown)


logachev (approved):

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryConnectionTest.java
```diff
@@ -522,4 +534,134 @@ public void testConnectionPropertiesLoggingAndMasking() throws IOException, SQLE
...
+    DataSource ds = DataSource.fromUrl(BASE_URL);
+    ds.setUseGlobalOpenTelemetry(useGlobal);
+    ds.setEnableGcpTraceExporter(enableTrace);
+    ds.setEnableGcpLogExporter(enableLog);
+
+    OpenTelemetry mockCustomOtel = mock(OpenTelemetry.class);
+    OpenTelemetry mockGlobalOtel = mock(OpenTelemetry.class);
+    OpenTelemetry mockDriverManagedOtel = mock(OpenTelemetry.class);
+    Logging mockLogging = mock(Logging.class);
+    org.mockito.Mockito.when(mockCustomOtel.getTracer(anyString())).thenReturn(mock(Tracer.class));
```
logachev: nit: import Mockito.when?

PR #13285, opened 2026-05-28
fix(bigquery-jdbc): fix manual commit mode failure for non-US regions

yagudin10:
## Problem

Manual commit mode fails when using BigQuery JDBC with non-US datasets.

**STR (Steps to Reproduce)**

1. Create BigQuery JDBC connection
2. Set a non-US location (e.g. `EU`) via `Location` property
3. Enable session mode (`EnableSession = 1`)
4. Disable auto-commit:
   ```java
   conn.setAutoCommit(false);
   ```
5. Execute a simple query:
 ```sql
SELECT 1;
```

**ER (Expected Result)**
- Query executes successfully in manual commit mode
- BigQuery uses configured Location (e.g. EU)
- No region mismatch errors occur

**AR (Actual Result)**

Query execution fails with:

```
GoogleJsonResponseException: 400 Bad Request
INVALID_ARGUMENT
Error: 3848323
```

**Sample java code:**
```java
String jdbcUrl = "jdbc:bigquery://https://www.googleapis.com/bigquery/v2:443";

Properties props = new Properties();
props.put("OAuthType", "0");
props.put("OAuthPvtKeyPath", "<key-file-path>");
props.put("ProjectId", "<project-id>");
props.put("Location", "EU");
props.put("EnableSession", "1");

try (Connection conn = DriverManager.getConnection(jdbcUrl, props)) {
    conn.setAutoCommit(false);

    try (PreparedStatement ps = conn.prepareStatement("SELECT 1;")) {
        ps.execute();
    }
}
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,20 +3,20 @@
 # For syntax help see:
 # https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-*   @googleapis/cloud-sdk-java-team
+*   @googleapis/cloud-sdk-java-team @googleapis/cloud-sdk-librarian-team
 
 # java-vertexai has maintainers
-/java-vertexai/   @googleapis/vertexai-team @googleapis/cloud-sdk-java-team
-/java-bigquerystorage/ @googleapis/bigquery-team @googleapis/cloud-sdk-java-team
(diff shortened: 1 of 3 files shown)


logachev (approved):
Thank you for finding & fixing this!

lqiu96 (approved):

jinseopkim0 (approved):

PR #13294, opened 2026-05-28
fix(bigquery-jdbc): add proper version to BigQueryDriver

logachev:
Update `BigQueryDriver`  getMinorVersion/getMajorVersion to return proper value


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/utils/BigQueryJdbcVersionUtility.java
```diff
@@ -0,0 +1,116 @@
...
+    ensureLoaded();
+    return parsedDriverVersion.get();
+  }
+
+  /**
+   * Gets the driver major version.
+   *
+   * @return the major version number
+   */
+  public static int getDriverMajorVersion() {
```
keshavdandeva: So, this and the minorVersion function with this new impl could throw `IllegalStateException` for `BigQueryDriver.getMajorVersion()` and `getMinorVersion()`, right? 
I think we need to confirm JDBC spec if that's okay or should it throw SQL exception or should it default to something and not throw error at all

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/utils/BigQueryJdbcVersionUtility.java
```diff
@@ -0,0 +1,116 @@
...
+    ensureLoaded();
+    return parsedDriverMajorVersion.get() != null ? parsedDriverMajorVersion.get() : 0;
+  }
+
+  /**
+   * Gets the driver minor version.
+   *
+   * @return the minor version number
+   */
+  public static int getDriverMinorVersion() {
```
keshavdandeva: same as above

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/utils/BigQueryJdbcVersionUtility.java
```diff
@@ -0,0 +1,116 @@
...
+      String errorMessage =
+          "Error reading dependencies.properties. Driver version information is"
+              + " unavailable. Error: "
+              + e.getMessage();
+      IllegalStateException ex = new IllegalStateException(errorMessage, e);
+      LOG.severe(errorMessage, ex);
+      throw ex;
+    }
+  }
+}
```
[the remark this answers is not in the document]
keshavdandeva: Agree with gemini, we should address concurrency issue here

keshavdandeva:
Can we also update the README in this PR for the maven link

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/utils/BigQueryJdbcVersionUtility.java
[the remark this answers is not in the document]
logachev: Fixed by making 'ensureLoaded' synchronized.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/utils/BigQueryJdbcVersionUtility.java
logachev: Good point, removed an exception;

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/utils/BigQueryJdbcVersionUtility.java
logachev: Good point, removed an exception;

keshavdandeva (approved):

PR #13301, opened 2026-05-29
fix(bigquery): prevent NullPointerException in OTel tracing when JobId is null

keshavdandeva:
b/517929704

This PR addresses a `NullPointerException` that occurs in the BigQuery client library when OpenTelemetry tracing is enabled and operations are performed on jobs that do not have a client-assigned `JobId` at the start of the call.

A notable case where this happens is during dry-run queries (often used to determine statement type or fetch statistics) where `JobInfo.getJobId()` returns null. The OTel instrumentation was attempting to extract attributes from the `JobId` without checking for nullability.

**Changes:**
*   Added null checks before calling `JobId.getOtelAttributes()` in `BigQueryImpl.create(JobInfo)` (line 455).
*   Added null checks before calling `JobId.getOtelAttributes()` in `BigQueryImpl.getQueryResults(JobId)` (line 2128).
*   These changes align with the safe-check pattern already implemented in `BigQueryImpl.queryWithTimeout` (line 2075).


lqiu96 (approved):

lqiu96:
Thanks for the info and the context. I don't have any concerns with this

PR #13304, opened 2026-05-29
build (deps): upgrade opentelemetry to 1.62.0 to patch CVE-2026-45292

blakeli0:
Upgrades opentelemetry.version to 1.62.0 to address CVE-2026-45292 (GHSA-rcgg-9c38-7xpx). This repo is actually NOT affected by this vulnerability but it is a good practice to upgrade it so it does not show up in customers' reports. 

Separately, removed unnecessary version declaration of opentelemetry in Spanner.

### DIFF
--- a/grpc-gcp-java/pom.xml
+++ b/grpc-gcp-java/pom.xml
@@ -73,7 +73,7 @@
     <jsr305.version>3.0.2</jsr305.version>
     <junit.version>4.13.2</junit.version>
     <opencensus.version>0.31.1</opencensus.version>
-    <opentelemetry.version>1.57.0</opentelemetry.version>
+    <opentelemetry.version>1.62.0</opentelemetry.version>
     <spanner.version>6.118.0-SNAPSHOT</spanner.version><!-- {x-version-update:google-cloud-spanner:current} -->
     <truth.version>1.4.5</truth.version>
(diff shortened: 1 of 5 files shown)


rahul2393 (approved):

lqiu96:
Do we care about this sample's version? https://github.com/googleapis/google-cloud-java/blob/8183d088d63bfa77ef5d70e4d20677249ed7848b/java-bigquery/samples/install-without-bom/pom.xml#L40

lqiu96 (approved):

blakeli0:
> Do we care about this sample's version?
> 
> https://github.com/googleapis/google-cloud-java/blob/8183d088d63bfa77ef5d70e4d20677249ed7848b/java-bigquery/samples/install-without-bom/pom.xml#L40

I don't think it matters much but it is more of a best practice to make all versions update together.

Separately, I also briefly looks into if we can remove the version from bigquery samples as well. I think we probably can because it is not used in between the [region tags](https://github.com/blakeli0/google-cloud-java/blob/342a513abc9c7d75a843420db714723b938e8d18/java-bigquery/samples/install-without-bom/pom.xml#L45-L51). Do you have more info about how it is used (not in scope of this PR)?

PR #13305, opened 2026-05-29
chore: create workflow to check showcase commit

### LINKED ISSUE #6200: google-cloud-java: create workflow check for showcase commit used
Create a workflow in google-cloud-java to check showcase commit used in librarian.yaml is in sync with java-showcase/gapic-showcase/pom.xml gapic-showcase.version property

See [go/sdk-librarian-migration-java](http://goto.google.com/sdk-librarian-migration-java)

sofisl:
A workflow to make sure the showcase commit is the same as what is in librarian. If the workflow fails, or there is a mismatch, an issue will be created.

Fixes https://github.com/googleapis/librarian/issues/6200


sofisl:
The logic is pretty complicated so I think maybe putting it all in a go file would be better, but I'm taking from [this](https://github.com/googleapis/librarian/issues/6200#issuecomment-4577853982) that maybe you'd prefer some more bash scripting.

.github/workflows/showcase-version-check.yaml
```diff
@@ -0,0 +1,88 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Extract showcase version from pom.xml
+        id: extract_version
+        shell: bash
+        run: |
+          version=$(awk -F'[<>]' '/gapic-showcase.version/{print $3; exit}' java-showcase/gapic-showcase/pom.xml)
```
zhumin8, line 38: use maven command for this: https://github.com/googleapis/librarian/issues/6200#issuecomment-4577853982

.github/workflows/showcase-version-check.yaml
```diff
@@ -0,0 +1,88 @@
...
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+      - name: Extract showcase commit from librarian.yaml
+        id: extract_commit
+        shell: bash
+        run: |
+          commit=$(grep -A 1 "showcase:" librarian.yaml | grep "commit:" | awk '{print $2}')
```
zhumin8: use `librarian config get sources.showcase.commit`

.github/workflows/showcase-version-check.yaml
sofisl: it doesn't work: https://github.com/googleapis/google-cloud-java/actions/runs/26656080361/job/78566438275

.github/workflows/showcase-version-check.yaml
sofisl: right! missed that

.github/scripts/verify_showcase_version.go
```diff
@@ -0,0 +1,76 @@
...
+			continue
+		}
+		if fields[1] == "refs/tags/"+tagName+"^{}" {
+			expectedCommit = fields[0]
+			break
+		}
+		if fields[1] == "refs/tags/"+tagName {
+			expectedCommit = fields[0]
+		}
+	}
```
zhumin8: Looks like we can simplify and avoid the loop here. For a specific version tag (like v0.39.0), git ls-remote will return either one or two lines depending on whether it's a lightweight tag or an annotated tag. And because of how Git natively sorts and streams this output, the real commit we want will always be on the very last line of the output, regardless of whether the tag is annotated or lightweight.
```suggestion
lines := strings.Split(strings.TrimSpace(string(outputBytes)), "\n")
if len(lines) == 0 || lines[0] == "" {
    fmt.Fprintf(os.Stderr, "Error: Tag %s not found on remote %s\n", tagName, remoteUrl)
    os.Exit(1)
}
finalLineFields := strings.Fields(lines[len(lines)-1])
expectedCommit = finalLineFields[0]
```

zhumin8: After this change, the logic seems to be simple enough to inline it?
I would vote for go script if we got enough test-worthy logic. In this case, it does not seem worth it? WDYT?

Something like (taken from jetski suggestion, not verified): 
```
- name: Verify showcase version and commit match
        shell: bash
        run: |
          # 1. Query the remote repository for the specific version tag
          REMOTE_URL="https://github.com/googleapis/gapic-showcase.git"
          TAG_NAME="v${{ steps.extract_version.outputs.version }}"
          
          OUTPUT=$(git ls-remote "$REMOTE_URL" "refs/tags/$TAG_NAME")
          
          if [ -z "$OUTPUT" ]; then
            echo "Error: Tag $TAG_NAME not found on remote $REMOTE_URL"
            exit 1
          fi
          
          # 2. Grab the commit hash from the very last line of the output
          EXPECTED_COMMIT=$(echo "$OUTPUT" | tail -n 1 | awk '{print $1}')
          
          echo "Expected commit for tag $TAG_NAME: $EXPECTED_COMMIT"
          echo "Current librarian commit: ${{ steps.extract_commit.outputs.commit }}"
          
          # 3. Cross-reference with librarian.yaml
          if [ "${{ steps.extract_commit.outputs.commit }}" != "$EXPECTED_COMMIT" ]; then
            echo "Mismatch: librarian.yaml has commit '${{ steps.extract_commit.outputs.commit }}', but tag '$TAG_NAME' is at commit '$EXPECTED_COMMIT'"
            echo "MISMATCH=true" >> "$GITHUB_ENV"
            exit 2
          fi
          
          echo "Showcase version and commit are in sync!"
```

.github/workflows/showcase-version-check.yaml
zhumin8: Looks like it is failing because the command printed a warning. The error is throw not because the maven command failed, but `echo "version=$version" >> "$GITHUB_OUTPUT"` failed because `version` is multiline.

Try this: 
```
            run: |
              version=$(mvn help:evaluate \
                -f java-showcase/gapic-showcase/pom.xml \
                -Dexpression=gapic-showcase.version \
                -q \
                -DforceStdout \
                -Dorg.slf4j.simpleLogger.defaultLogLevel=ERROR 2>/dev/null \
                | grep -E '^[0-9]' || true)

              echo "version=version" > >"GITHUB_OUTPUT"
```

.github/scripts/verify_showcase_version.go
sofisl: yay, this worked!

.github/workflows/showcase-version-check.yaml
sofisl: this is still not working. asking jetski

```
The reason mvn help:evaluate is failing on the clean GitHub Actions runner (and returning an empty version, which causes the Tag v not found error) is that Maven attempts to parse the entire dependency tree and resolve imports, including google-cloud-shared-dependencies:3.63.0-SNAPSHOT (which is a local snapshot in the monorepo that hasn't been compiled or installed yet on a clean runner).

To solve this without modifying any POM files and without running Maven, we can use awk to extract the version string directly from the XML
```

zhumin8 (approved):
Merge with main you should be able to avoid the irrelevant ci/unit tests (https://github.com/googleapis/google-cloud-java/pull/13299).

PR #13309, opened 2026-05-29
chore: change to post submit and create an issue on failure

sofisl:
Converts the presubmit to a postsubmit check to not delay PRs. Creates issues if necessary. Per https://github.com/googleapis/librarian/issues/5240#issuecomment-4578899009.

Fixes [5240](https://github.com/googleapis/librarian/issues/5240)


JoeWang1127 (approved):

zhumin8 (changes requested):
Change LGTM. However, I don't think we are ready to merge this now:
- we should to sync librarian.yaml with generation_config.yaml before this, or it will be always failing and creates issues on each PR (see https://github.com/googleapis/google-cloud-java/actions/workflows/librarian_generation_check.yaml).
- Currently, we got generation diff (see https://github.com/googleapis/google-cloud-java/pull/13319) before the release is cut or fix for https://github.com/googleapis/librarian/issues/6244.

sofisl:
From Min:

```
I would pause and merge this in toward the end of next week (when we turn done hermetic build workflow and librarian.yaml is the source of truth).
```

.github/workflows/librarian_generation_check.yaml
```diff
@@ -66,3 +72,11 @@ jobs:
     - name: Check for generated code changes
       run: |
         git diff --exit-code
+    - name: Create issue if previous step fails
```
zhumin8, line 75: Considering current generate all runs for 1 hour +, If one PR is merged causing generation diff, it's likely subsequent PRs are merged. This will create repeating issues. I think we want to combine them to not over-pollute the issue board.
In fact, I believe we can reuse this action from librarian that already does exactly this. https://github.com/googleapis/librarian/tree/main/.github/actions/create-issue-on-failure

zhumin8:
closing in favor of https://github.com/googleapis/google-cloud-java/pull/13422

PR #13321, opened 2026-06-01
fix(bigquery-jdbc): fallback to standard precision and scale for numeric, integer, and temporal types

Neenu1995:
b/516471484

## Problem
Currently, `ResultSetMetaData.getPrecision()` and `getScale()` fall back to `0` for all standard BigQuery types unless explicitly returned by the server's field metadata (which is rarely populated for basic columns). Returning `0` violates the JDBC specification for exact numeric, integer, and temporal types, which can cause ORMs and schema inspectors to misbehave.

Additionally, the type mappings inside `BigQueryDatabaseMetaData` used for schema queries (like `getColumns()`) were inconsistent with the JDBC specification and standard string length representations for temporal types, reporting a precision of `29` with `null` scale for `TIMESTAMP` and `DATETIME`, and missing scale definitions for `TIME` and `DATE`.

### Solution
1. **`BigQueryResultSetMetadata`**:
Modified BigQueryResultSetMetadata.java to fall back to the standard SQL type definition properties when explicit schema-defined precision/scale are absent.

The driver now returns standard fallback values based on the column's StandardSQLTypeName:

| BigQuery Type | JDBC SQL Type | Fallback Precision | Fallback Scale | Justification |
| :--- | :--- | :--- | :--- | :--- |
| **`INT64`** | `Types.BIGINT` | `19` | `0` | Signed 64-bit integer range limit |
| **`NUMERIC`** | `Types.NUMERIC` | `38` | `9` | Standard BigQuery NUMERIC specifications |
| **`BIGNUMERIC`** | `Types.NUMERIC` | `77` | `38` | Standard BigQuery BIGNUMERIC specifications |
| **`DATE`** | `Types.DATE` | `10` | `0` | Display length of `"YYYY-MM-DD"` |
| **`TIME`** | `Types.TIME` | `15` | `6` | Display length of `"HH:MM:SS.ffffff"` (microseconds) |
| **`TIMESTAMP`** / **`DATETIME`** | `Types.TIMESTAMP` | `26` | `6` | Display length of `"YYYY-MM-DD HH:MM:SS.ffffff"` |
| **`BOOL`** | `Types.BOOLEAN` | `1` | `0` | Single-digit bit/flag |
| **`FLOAT64`** | `Types.DOUBLE` | `15` | `0` | Double-precision floating point precision limit |

2. **`BigQueryDatabaseMetaData`**:
   - Aligned `STANDARD_TYPE_INFO` mappings for `TIMESTAMP`, `DATETIME`, `DATE`, and `TIME` to match the fallback precision and scale values reported by `BigQueryResultSetMetadata`.
   - Added explanatory comments in the code justifying the fallback/precision values.

### Testing
- Updated unit tests in BigQueryResultSetMetadataTest.java to verify correct fallback values across all 13 dataset columns.


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -4835,6 +4835,8 @@ private Tuple<String, String> determineEffectiveCatalogAndSchema(
 
   // BigQuery STRING represents Unicode character strings (UTF-8).
   // Standard JDBC maps UTF-8/Unicode data to Types.NVARCHAR rather than Types.VARCHAR.
+  // Precision and scale values for temporal types (TIMESTAMP, DATETIME, DATE, TIME) are
```
logachev: Can we move this mapping to some share file & re-use from ResultSetMetadata to ensure they're not out of sync?

Neenu1995: done.

logachev (approved):

keshavdandeva (approved):

PR #13326, opened 2026-06-01
chore: always run units jobs with step-level skips

blakeli0:
### Description

This PR solves the issue where required checks `units (11)`, `units (17)`, `units (21)`, `units (25)`, and `units (8)` get stuck in a pending "stuck forever" state on GitHub.

For example, https://screenshot.googleplex.com/9bpNwWP7jParbuM

### Root Cause
The `units` and `units-8-runtime` jobs in `.github/workflows/ci.yaml` are conditionally skipped using job-level `if: ${{ needs.bulk-filter.outputs.runnable == 'true' }}`.
Because the entire job is skipped at the job-level, the matrix versions are not expanded, and status checks like `units (11)` are never reported to GitHub. When these checks are configured directly as required status checks in branch protection rules, PRs that skip the jobs (e.g., workflow-only PRs) are stuck forever.

### Solution
We removed the job-level `if: runnable == true` conditional from both `units` and `units-8-runtime` jobs so they always start. Instead, we applied the `if: runnable == true` check to every individual step inside these jobs.
When a PR doesn't need to run these checks:
1. The jobs will still start and run.
2. All steps will skip instantly.
3. The jobs will finish successfully in 1–2 seconds, reporting success and satisfying GitHub's required status checks.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,6 @@ jobs:
   units:
     runs-on: ubuntu-latest
     needs: bulk-filter
-    if: ${{ needs.bulk-filter.outputs.runnable == 'true' }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


lqiu96:
IIUC, this seems to be a limitation of the Github CI. The required status for the matrix jobs aren't reported because they're never run (though Github CI docs does say that skipped jobs are reported as a success). I guess just not for matrix jobs.

Might be another data point to try and move away from bulk jobs (needing to run all GAPIC unit tests).

lqiu96 (approved):

blakeli0:
> this seems to be a limitation of the Github CI. The required status for the matrix jobs aren't reported because they're never run

Yes exactly.

PR #13331, opened 2026-06-02
fix(auth): Skip RAB lookup in case of non-email response from MDS.

vverman:
In ComputeEngineCredentials when running on GKE platform, the getAccount() call may return a value which isn't an email.

In this case the right behaviour is to skip RAB lookup which is what this PR does.

Added tests.

### DIFF
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -805,6 +805,11 @@ public class ComputeEngineCredentials extends GoogleCredentials
     // Since RAB lookup requires a valid email-based service account, we skip RAB lookup
     // in non-email scenarios by returning null.
     if (account == null || !account.contains("@")) {
+      LoggingUtils.log(
+          LOGGER_PROVIDER,
+          Level.INFO,
+          Collections.emptyMap(),
+          "Regional Access Boundary lookup is skipped for this instance because it is a non-email instance.");
       return null;
     }
     return String.format(
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -800,8 +800,15 @@ public class ComputeEngineCredentials extends GoogleCredentials
   @InternalApi
   @Override
   public String getRegionalAccessBoundaryUrl() throws IOException {
+    String account = getAccount();
+    // In GKE environments, the default service account might return a non-email placeholder.
+    // Since RAB lookup requires a valid email-based service account, we skip RAB lookup
+    // in non-email scenarios by returning null.
+    if (account == null || !account.contains("@")) {
+      return null;
+    }
     return String.format(
-        OAuth2Utils.IAM_CREDENTIALS_ALLOWED_LOCATIONS_URL_FORMAT_SERVICE_ACCOUNT, getAccount());
+        OAuth2Utils.IAM_CREDENTIALS_ALLOWED_LOCATIONS_URL_FORMAT_SERVICE_ACCOUNT, account);
   }
 
   /**
--- a/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -72,6 +72,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 
 /**
  * OAuth2 credentials representing the built-in service account for a Google Compute Engine VM.
@@ -117,6 +118,7 @@ public class ComputeEngineCredentials extends GoogleCredentials
 
   private static final String PARSE_ERROR_PREFIX = "Error parsing token refresh response. ";
   private static final String PARSE_ERROR_ACCOUNT = "Error parsing service account response. ";
+  private static final Pattern EMAIL_PATTERN = Pattern.compile("^[^@]+@[^@]+\\.[^@]+$");
   private static final long serialVersionUID = -4113476462526554235L;
 
   private final String transportFactoryClassName;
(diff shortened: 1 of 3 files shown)


google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -800,8 +800,15 @@ public String getAccount() {
   @InternalApi
   @Override
   public String getRegionalAccessBoundaryUrl() throws IOException {
+    String account = getAccount();
+    // In GKE environments, the default service account might return a non-email placeholder.
+    // Since RAB lookup requires a valid email-based service account, we skip RAB lookup
+    // in non-email scenarios by returning null.
+    if (account == null || !account.contains("@")) {
```
lqiu96: I know the document may have changed since this PR was raised, but I think the the standard email pattern is a regex and we will need to confirm to that.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -800,8 +800,15 @@ public String getAccount() {
...
+      return null;
+    }
```
lqiu96: I think we will also need to follow the recommendation for a one-time log message that the RAB instance is skipped (maybe here, maybe in the RABManager)

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/RegionalAccessBoundaryManager.java
```diff
@@ -196,6 +196,10 @@ void triggerAsyncRefresh(
           () -> {
             try {
               String url = provider.getRegionalAccessBoundaryUrl();
+              if (url == null) {
+                future.set(null);
+                return;
+              }
```
lqiu96, line 201: qq, can you remind me again why we need to do `future.set(null);`?

vverman: I was using a SettableFuture as a gate to deduplicate RAB refreshes which let's us set the future to indicate to dependent threads as to the result of the refresh.

However, I realized the same can be done with AtomicBoolean which is more easier to maintain. Hence updated the logic.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
vverman: Done!

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
vverman: Done, we now use email based regex.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -800,8 +802,20 @@ public String getAccount() {
   @InternalApi
   @Override
   public String getRegionalAccessBoundaryUrl() throws IOException {
+    String account = getAccount();
+    // In GKE environments, the default service account might return a non-email placeholder.
+    // Since RAB lookup requires a valid email-based service account, we skip RAB lookup
+    // in non-email scenarios by returning null.
```
lqiu96: nit: Since this is a temp change, can you link to the internal ticket tracking this (b/XXXX)

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -800,8 +802,20 @@ public String getAccount() {
...
+    if (account == null || !EMAIL_PATTERN.matcher(account).matches()) {
+      LoggingUtils.log(
+          LOGGER_PROVIDER,
+          Level.INFO,
+          Collections.emptyMap(),
+          "Regional Access Boundary lookup is skipped for this instance because it is a non-email instance.");
```
lqiu96: two small nits:
1. I know the spec says 1-time INFO log (I think this should be fine given the skipRAB flag). For Java SDK, I think that might be too noisy for an internal/ hidden impl. Is it possible to change the warning level? 
2. Can we try a bit softer wording without mentioning RAB? It may be me, but `lookup skipped ... non-email instance` sounds a bit like an error and not super actionable for users. 

Maybe something like (ff to change this to something clearer): `Unable to retrieve this instance's email and will skip the regional request routing. Proceeding with request`

vverman: Added the log you suggested. I think the design mentions a one time INFO log hence I left that as is.

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
vverman: This actualy isn't a temp change. MDS already returns a non-email value for certain GKE instances so this PR is actually to handle that scenario.

lsirac (approved):

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: Apologies, I meant that the skip RAB is a temp change for now. Can you link the internal tracking ticket in the PR description or as a comment in the code `TODO(b/xxxx): ..`?

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: Thanks, on a second look, I think INFO with a one-time non error message is OK.

lqiu96:
Some errors in the CI:
```
external/com_google_api_gax_java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java:357: error: incompatible types: String cannot be converted to URI
    return ((GdchCredentials) credentials).createWithGdchAudience(audienceString);
```

Hmm, this may be related to some changes we made into GDCH to resolve a customer issue. Can you try and pull in the latest changes into the RAB feature branch?

lqiu96 (approved):

google-auth-library-java/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
vverman: Ah I see, the correct and expected behaviour is for RAB lookup to be skipped when a non-email account is returned from the MDS.

This won't be a temp change rather this is the fix.

vverman:
Plan to raise a PR with the rebased main later as it'd introduce a huge delta in this PR.

PR #13335, opened 2026-06-02
chore: POC for PQC in gax-grpc

diegomarquezp:


sdk-platform-java/gax-java/gax-grpc/pom.xml
```diff
@@ -70,7 +70,6 @@
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-netty-shaded</artifactId>
-      <scope>runtime</scope>
```
diegomarquezp, line 73: this must be evaluated

diegomarquezp: This is not viable nor recommended. We should expose something in `ManagedChannelBuilder`

PR #13336, opened 2026-06-02
fix(build): support generating unversioned libraries in OwlBot config

blakeli0:
This PR adds support for generating unversioned libraries (such as backstory) in the hermetic build OwlBot configuration. Historically, only versioned libraries (with v1, v1beta, etc. directories) were supported in deep-copy regex. For example, new library generation for [backstory](https://github.com/googleapis/google-cloud-java/pull/13334) didn't trigger any code changes.

This PR checks if a version component is present and conditionally outputs versioned or unversioned copy patterns.

### DIFF
--- a/sdk-platform-java/hermetic_build/library_generation/generate_composed_library.py
+++ b/sdk-platform-java/hermetic_build/library_generation/generate_composed_library.py
@@ -32,6 +32,7 @@ import os
 from pathlib import Path
 from typing import List
 import library_generation.utils.utilities as util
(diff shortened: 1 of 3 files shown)


blakeli0:
FYI @zhumin8, we have to change the owlbot template but I know we are moving away from it.

zhumin8 (approved):
LGTM. 
Thanks for the note, librarian already supports non-versioned paths from existing libraries.

diegomarquezp (approved):

PR #13338, opened 2026-06-02
feat: enable self-signed JWTs by default in ServiceOptions

whowes:
This should improve efficiency/reliability by avoiding OAuth token exchange - see https://google.aip.dev/auth/4111.

BigQueryOptions (b/523372553) and StorageOptions (b/523372960) are disabled by default for now.

### DIFF
--- a/java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
+++ b/java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
@@ -77,7 +77,9 @@ public class BigQueryOptions extends ServiceOptions<BigQuery, BigQueryOptions> {
     private Tracer openTelemetryTracer;
     private ResultRetryAlgorithm<?> resultRetryAlgorithm;
 
-    private Builder() {}
+    private Builder() {
+      setUseJwtAccessWithScope(false);
+    }
 
     private Builder(BigQueryOptions options) {
       super(options);
@@ -211,11 +213,6 @@ public class BigQueryOptions extends ServiceOptions<BigQuery, BigQueryOptions> {
 
   public static HttpTransportOptions getDefaultHttpTransportOptions() {
     return HttpTransportOptions.newBuilder().setReadTimeout(DEFAULT_READ_API_TIME_OUT).build();
-  }
-
-  @Override
-  protected boolean useSelfSignedJwt() {
-    return false;
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


sdk-platform-java/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -650,9 +650,17 @@ public Credentials getScopedCredentials() {
...
       credentialsToReturn = ((GoogleCredentials) credentials).createScoped(getScopes());
     }
+    if (useSelfSignedJwt() && credentialsToReturn instanceof ServiceAccountCredentials) {
+      credentialsToReturn =
+          ((ServiceAccountCredentials) credentialsToReturn).createWithUseJwtAccessWithScope(true);
+    }
     return credentialsToReturn;
   }
 
+  protected boolean useSelfSignedJwt() {
```
lqiu96: nit: I believe the AIP marked the name as `UseJWTAccessWithScope`. Can we use that name as well? Similar to what we have in GoogleCredentialProvider, I think we may want to add a setter: https://github.com/googleapis/google-cloud-java/blob/e1e40592073c33709937e9ed47847bd6e266a405/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/core/GoogleCredentialsProvider.java#L156

nit: Also a small javadoc blurb for this.

whowes: Thanks for the review!

- adjusted naming to consistently reference "useJwtAccessWithScope"
- added a setter (via the ServiceOptions.Builder)
- added javadoc

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
```diff
@@ -77,7 +77,9 @@ public static class Builder extends ServiceOptions.Builder<BigQuery, BigQueryOpt
     private Tracer openTelemetryTracer;
     private ResultRetryAlgorithm<?> resultRetryAlgorithm;
 
-    private Builder() {}
+    private Builder() {
+      setUseJwtAccessWithScope(false);
```
lqiu96, line 81: For BQ and Storage, are there any internal tracking tickets that we can link for these exceptions (e.g. default false)? IIRC, these are more so temporary and not permanent as we want this for default. I think would be helpful for future context

lqiu96 (approved):

java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryOptions.java
whowes: Good suggestion, I added tracking tickets to the PR description.

PR #13342, opened 2026-06-02
refactor(gax): implement clean stateless channel refresh & retry mapping

lqiu96:
POC code to have a separation of refreshing vs retrying

### DIFF
--- a/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/CallbackChainRetryingFuture.java
+++ b/sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/CallbackChainRetryingFuture.java
@@ -32,9 +32,6 @@ package com.google.api.gax.retrying;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.core.ApiFuture;
-import com.google.api.gax.rpc.ApiCallContext;
-import com.google.api.gax.rpc.TransportChannel;
-import com.google.api.gax.rpc.UnauthenticatedException;
 import com.google.common.util.concurrent.MoreExecutors;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -56,7 +53,6 @@ import java.util.concurrent.Future;
  */
 class CallbackChainRetryingFuture<ResponseT> extends BasicRetryingFuture<ResponseT> {
   private final ScheduledRetryingExecutor<ResponseT> retryingExecutor;
-  private final RetryingContext context;
   private volatile AttemptCompletionListener attemptFutureCompletionListener;
 
   CallbackChainRetryingFuture(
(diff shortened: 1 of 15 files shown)


sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingAttemptCallable.java
```diff
@@ -234,6 +235,28 @@ public void onResponseImpl(ResponseT response) {
...
+                cause = new UnauthenticatedException(
+                    causeEx.getMessage(),
+                    causeEx.getCause(),
+                    causeEx.getStatusCode(),
+                    true, // isRetryable = true
+                    causeEx.getErrorDetails());
+                
+                t = new com.google.api.gax.retrying.ServerStreamingAttemptException(
+                    cause,
+                    true, // canResume = true
```
lqiu96, line 256: probably should be resumptionStrategy.canResume() instead of hard coding true

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/retrying/CallbackChainRetryingFuture.java
```diff
@@ -114,7 +119,23 @@ public void run() {
         ResponseT response = attemptFuture.get();
         handle(null, response);
       } catch (ExecutionException e) {
-        handle(e.getCause(), null);
+        Throwable cause = e.getCause();
+        if (cause instanceof UnauthenticatedException) {
+          if (context instanceof ApiCallContext) {
+            TransportChannel transportChannel = ((ApiCallContext) context).getTransportChannel();
+            if (transportChannel != null && transportChannel.shouldRefresh()) {
+              transportChannel.refresh();
```
vverman: IIUC, we're still doing the refresh within the retry here.

What's changing is that as opposed to ApiResultRetryAlgo, we're attempting the refresh inside the CallbackChainRetryingFuture which I assume is one layer below .

Additionally: We might introduce a circular dependency of the retry layer on the rpc layer while the latter already depends on the former.

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannel.java
```diff
@@ -47,4 +47,20 @@ public interface TransportChannel extends BackgroundResource {
...
+   * <p>By default, this is a no-op for transports that do not require stateful connection lifecycle
+   * management.
+   */
+  default void refresh() {}
+
+  /**
+   * Returns true if a certificate rotation has been detected on disk and the transport channel
+   * should be refreshed, or false otherwise.
+   */
+  default boolean shouldRefresh() {
```
lqiu96, line 63: does this need to be on the interface? Is there any transport specific logic here?

PR #13344, opened 2026-06-03
feat(bigquery-jdbc): add `EnableProjectDiscovery` connection property for metadata methods

keshavdandeva:
b/499078725

This PR implements the `EnableProjectDiscovery` connection property into the BigQuery JDBC driver. Enabling this property (default `false`) allows JDBC database metadata methods (like `getCatalogs()` and `getSchemas()`) to discover and query datasets across all Google Cloud projects accessible to the client credentials, rather than being confined to the single default `ProjectId` specified in the connection URL.

#### Changes
- **Connection Parameter Parsing**: Added parsing for `EnableProjectDiscovery` in `BigQueryJdbcUrlUtility` and configured it in `DataSource`.
- **SDK Integration**: Implemented `BigQueryConnection.getDiscoveredProjects()` to utilize the `BigQuery.listProjects()` core SDK method
- **Resilience and Caching**:
  - Connection-scoped caching is implemented via `discoveredProjectsCache`.
- **Metadata Integration**: Integrated discovered projects inside `BigQueryDatabaseMetaData.getAccessibleCatalogNames()` so that catalog and schema metadata fetches automatically query across all accessible projects when `EnableProjectDiscovery` is enabled.


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1221,6 +1231,42 @@ private boolean checkIsReadOnlyTokenUsed(Map<String, String> authProps) {
...
+        pageToken = projectList.getNextPageToken();
+      } while (pageToken != null);
+
+      this.discoveredProjectsCache = ImmutableList.copyOf(projects);
+    } catch (Exception e) {
+      LOG.warning(e, "Failed to list all accessible projects, falling back to connection default.");
+      this.discoveredProjectsCache = ImmutableList.of();
+    }
+    return this.discoveredProjectsCache;
+  }
```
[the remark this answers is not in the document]
keshavdandeva: Done

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -3652,60 +3652,61 @@ public ResultSet getSchemas(String catalog, String schemaPattern) {
             return;
           }
 
+          ExecutorService apiExecutor = null;
+          final List<Future<List<Dataset>>> apiFutures = new ArrayList<>();
           try {
+            apiExecutor = Executors.newFixedThreadPool(API_EXECUTOR_POOL_SIZE);
```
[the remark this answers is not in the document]
keshavdandeva: Will resolve this in a separate PR for all the metadata methods

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1221,6 +1237,61 @@ private boolean checkIsReadOnlyTokenUsed(Map<String, String> authProps) {
...
+    return this.enableProjectDiscovery;
+  }
+
+  public synchronized List<String> getDiscoveredProjects() {
+    if (this.discoveredProjectsCache != null) {
+      return this.discoveredProjectsCache;
+    }
+
+    try {
+      BigQueryOptions options = (BigQueryOptions) getBigQuery().getOptions();
```
logachev: We have already code that provisions BigQuery client, we should not be creating new one (e.g. feels like HttpTransport is missing proxy/private endpoint and other properties that are configured).

Are there issues with SDK client?

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -77,6 +85,10 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
   private final String connectionId;
   private static final String DEFAULT_JDBC_TOKEN_VALUE = "Google-BigQuery-JDBC-Driver";
   private static final String DEFAULT_VERSION = "0.0.0";
+  private static final String BIGQUERY_SERVICE_NAME = "bigquery";
+  private static final long MAX_PROJECTS_PER_PAGE = 10000L;
```
logachev: nit: lets use `maxResults` property we already have

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -77,6 +85,10 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
...
+  private static final String PROJECT_LIST_FIELDS =
+      "projects/projectReference/projectId,nextPageToken";
```
logachev: Better to query projectName rather than projectId imo. This is used in some UI tools

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
logachev: instead of create threadPool, we should have one available. We have `MetaDataFetchThreadCount` property available, but we actually lack connection-layer threadPool.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
logachev: I'd suggest some refactoring for this method, few ideas to simplify:

- Move code to fetch list of schemas in a specific catalog to a separate function;
- Keep this method simple - single catalog only, essentially calls into helper & transforms to jsonResultSet. No need for background threads since it is single catalog
- Refactor `getSchemas()` to be the one that fans out multiple requests & assembles data.

Probably breaking up CLs in 2 will be easier:
- Add proper threadPool to connection-layer (and remove static threadPool in statement, it is not used)
- update getSchemas

Also I'd suggest to reuse more code. There are 4 metadata methods that can generate large # of rest calls:
- getCatalogs()
- getSchemas()
- getTables()
- getColumns()
- getProcedures()

In a way, they build results on top of each other. Right now we duplicate a lot, e.g. `listDatasets()` is called in 7 different places. 

Also please don't tackle all of them in a single CL, lets do it incrementally :)
(Otherwise it is pain to review)

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: Yeah, so the SDK does not expose a public `listProjects()` method. To invoke the BigQuery REST /projects endpoint, we must drop down to the low-level client's `projects().list()` call.

I used reflection to extract the underlying client from `BigQueryImpl` but gemini said it is bad practice and is fragile and can throw dynamic access exceptions (e.g. `InaccessibleObjectException` under strict modular Java 17+ runtimes) or fail when BigQuery is mocked in tests.

Hence, the current implementation. And it should **not** bypass proxy, endpoint, or auth configurations.

- Proxy: Calling `transportOptions.getHttpTransportFactory().create()` creates the exact HttpTransport configured in the connection provider (which contains any custom proxy factory settings).
- Private Endpoints: `options.getResolvedApiaryHost(BIGQUERY_SERVICE_NAME)` resolves the correct API host, incorporating user overrides or custom private service endpoints.
- Auth/Timeouts/Headers: `transportOptions.getHttpRequestInitializer(options)` automatically provisions the auth token initialization, user agent headers, and connection timeouts."

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: In the BigQuery REST API, there is no `projectName` field (only `projectId` and `friendlyName).` Using `projectId` is the standard way to reference GCP projects. Also, the catalog name returned by `getCatalogs()` must be the alphanumeric `projectId.`

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
keshavdandeva: Yes, completely agree. I noticed this as well and this is happening with all major metadata methods (`getTables,` `getColumns,` `getProcedures,` etc.). I created b/520400589 and will work on it in separate PR

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
keshavdandeva: Yeah, makes sense. I have created 3 bugs:

1. b/520400589 - Refactor Metadata Thread Pool Management to Reuse Connection-Scoped Executor 
2. b/520407325 - Refactor getSchemas for Catalog-Based Routing (Synchronous & Async Fan-out)
3. b/520406763 - Deduplicate metadata API calls

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1221,6 +1236,64 @@ private boolean checkIsReadOnlyTokenUsed(Map<String, String> authProps) {
...
+    if (this.discoveredProjectsCache != null) {
+      return this.discoveredProjectsCache;
+    }
+
+    try {
+      BigQueryOptions options = (BigQueryOptions) getBigQuery().getOptions();
+      HttpTransportOptions transportOptions = (HttpTransportOptions) options.getTransportOptions();
+      HttpTransport transport = transportOptions.getHttpTransportFactory().create();
+      HttpRequestInitializer initializer = transportOptions.getHttpRequestInitializer(options);
+      Bigquery lowLevelBq =
```
Neenu1995: I don't think we should use both of the Bigquery clients the JDBC Driver. If a feature in the Apiary client is required in JDBC, lets first bring it into the cloud client and use it from there.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1312,6 +1320,39 @@ private boolean checkIsReadOnlyTokenUsed(Map<String, String> authProps) {
...
+      BigQuery bigQuery = getBigQuery();
+      List<String> projects = new ArrayList<>();
+      Page<Project> projectPage =
+          bigQuery.listProjects(ProjectListOption.pageSize(getMaxResults()));
+      for (Project project : projectPage.iterateAll()) {
+        projects.add(project.getProjectId());
+      }
+      this.discoveredProjectsCache = ImmutableList.copyOf(projects);
+    } catch (BigQueryException e) {
+      LOG.warning(e, "Failed to list all accessible projects due to BigQuery error.");
```
logachev: I think we shouldn't be obscuring errors. If we can't query, we should properly raise exception all the way to the caller. That's no different from being unable to run a query

keshavdandeva: Done

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1312,6 +1320,33 @@ private boolean checkIsReadOnlyTokenUsed(Map<String, String> authProps) {
...
+      for (Project project : projectPage.iterateAll()) {
+        projects.add(project.getProjectId());
+      }
+      this.discoveredProjectsCache = ImmutableList.copyOf(projects);
+    } catch (BigQueryException e) {
+      throw new BigQueryJdbcException(
+          "Failed to list all accessible projects due to BigQuery error.", e);
+    } catch (Exception e) {
+      throw new BigQueryJdbcException("Failed to list all accessible projects.", e);
+    }
```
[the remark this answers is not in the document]
keshavdandeva: Removed this logic

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/BigQueryConnectionTest.java
```diff
@@ -519,4 +529,81 @@ public void testWrapperMethods() throws Exception {
...
+      // Subsequent call should retry since no cache is set
+      assertThrows(
+          BigQueryJdbcException.class,
+          () -> {
+            connection.getDiscoveredProjects();
+          });
+      verify(mockBigQuery, times(2)).listProjects(any(BigQuery.ProjectListOption.class));
+    }
+  }
+
```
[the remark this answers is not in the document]
keshavdandeva: Not needed

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1312,6 +1320,33 @@ private boolean checkIsReadOnlyTokenUsed(Map<String, String> authProps) {
...
+  public synchronized List<String> getDiscoveredProjects() throws SQLException {
+    if (this.discoveredProjectsCache != null) {
+      return this.discoveredProjectsCache;
+    }
+
+    try {
+      BigQuery bigQuery = getBigQuery();
+      List<String> projects = new ArrayList<>();
+      Page<Project> projectPage =
+          bigQuery.listProjects(ProjectListOption.pageSize(getMaxResults()));
```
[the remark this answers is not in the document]
keshavdandeva: Interesting, I did look up about this and I agree with Gemini. Removing `getMaxResults()`

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -1312,6 +1319,32 @@ private boolean checkIsReadOnlyTokenUsed(Map<String, String> authProps) {
...
+      List<String> projects = new ArrayList<>();
+      Page<Project> projectPage = bigQuery.listProjects();
+      for (Project project : projectPage.iterateAll()) {
+        projects.add(project.getProjectId());
+      }
+      this.discoveredProjectsCache = ImmutableList.copyOf(projects);
+    } catch (BigQueryException e) {
+      throw new BigQueryJdbcException(
+          "Failed to list all accessible projects due to BigQuery error.", e);
+    } catch (Exception e) {
```
logachev, line 1347: nit: no need to handle BQ in a different way, child exception will be available either way.

logachev (approved):

PR #13347, opened 2026-06-03
chore: sync librarian.yaml with generation_config.yaml

zhumin8:


zhumin8:
Code changes in backstory:
- some are due to proto change. Latest generation PR did not included these because it was created before backstory is added. So the main right now has backstory code generated from an earlier googeapis commitish and updated generated_config.yaml after.
https://github.com/googleapis/google-cloud-java/commit/a90bc3dec325583d0e751ad2c8171a376de2fd16
- java-backstory/proto-google-cloud-backstory/clirr-ignored-differences.xml file is created
- repometadata and readme diff: defaults to both instead of grpc transport. But since this repo is proto only, there is no practical difference.

cc. @blakeli0

zhumin8:
generation diff found for bigtable, due to proto change: 

```
diff --git a/google/bigtable/v2/bigtable.proto b/google/bigtable/v2/bigtable.proto
index 1a2bb37ccc..32c133bd9a 100644
--- a/google/bigtable/v2/bigtable.proto
+++ b/google/bigtable/v2/bigtable.proto
@@ -1187,6 +1187,15 @@ message ExecuteQueryRequest {
   // `Value.type` must match the corresponding `param_types` entry, or be
   // rejected with `INVALID_ARGUMENT`.
   map<string, Value> params = 7 [(google.api.field_behavior) = REQUIRED];
+
+  // Optional. This map provides the runtime values returned by the
+  // VIEW_PARAMETERS() function calls, typically used for user-level scoping of
+  // data based on identity.
+  //
+  // The key is the name of the view parameter e.g. `user_id`, and
+  // the value is the parameter value e.g. `alice@example.com`.
+  map<string, Value> view_parameters = 12
+      [(google.api.field_behavior) = OPTIONAL];
 }
 
 // Response message for Bigtable.ExecuteQuery
(END)
```

This is the only change `git diff 0ff3c706a5e0964c6430529c406208f2ff57cf65 5481332007e57ad3d9cb81e83ff6bf9c449476b6`
Root cause is last night's generation workflow only updated commitish but no code generation.
https://github.com/googleapis/google-cloud-java/pull/13315/changes/3ccc31063a469b10d2d117c622c83747505dd51f#diff-70450badb57c93e27292190b5470ead474f3b2071e682e3d8274ec45e9510410L1
@blakeli0 is going to revert this change

zhumin8:
closing this one due to these changes:
https://github.com/googleapis/google-cloud-java/pull/13350
https://github.com/googleapis/google-cloud-java/pull/13351

PR #13365, opened 2026-06-04
fix(bigquery-jdbc): lazy caller inference and zero-overhead coercion logging

Neenu1995:
### Problem
When the BigQuery JDBC driver is run with JUL-to-SLF4J bridges (like `SLF4JBridgeHandler` in Spring Boot environments), JUL's logging level is configured to `ALL` to allow filtering to occur at the SLF4J/Logback layer. As a result, the driver's custom logger always proceeded to log, eagerly capturing and walking stack traces (`new Throwable().getStackTrace()`) to find the caller class and method. On hot paths (such as `BigQueryTypeCoercer.coerceTo` converting values for millions of rows), this caused severe CPU and memory overhead (representing up to ~99% CPU time in profile runs), even if SLF4J immediately discarded the logs.
### Summary of Changes
* **Lazy Caller Inference (`BigQueryJdbcLogRecord`)**:
  * Introduced `BigQueryJdbcLogRecord` (a custom subclass of `LogRecord`) inside `BigQueryJdbcCustomLogger` to defer caller stack-trace walking until requested by formatters/bridges.
  * Ensures zero stack-walk overhead when logging is disabled under SLF4J bridges.
  * Added framework packages (like `org.slf4j.bridge.*`) to the caller skipper list to maintain inference correctness.
* **Coercion Hot Path Optimization (`BigQueryTypeCoercer`)**:
  * Switched type coercion logger to the specialized `BigQueryJdbcResultSetLogger`.
  * Replaced generic `effectiveLog.finest` with explicit context passing using `effectiveLog.finestTrace("coerceTo", ...)`.
  * Avoids stack walks entirely on the result set coercion hot path, regardless of whether logging is enabled.
* **Unit Testing**:
  * Added `testLazyCallerInference` in `BigQueryJdbcCustomLoggerTest` to verify that stack-trace walking is deferred until queried and accurately resolves the correct caller class/method.
### Performance Impact
* **Benchmark results**: Query execution on 100k rows × 5 cols with high-throughput enabled yielded a **~5.9% end-to-end execution time reduction** (and up to 3.3× speedup on larger 2M × 34 datasets where JDBC coercion overhead is the main bottleneck).
* Eliminated millions of garbage collection allocations of `StackTraceElement[]` on the JDBC data retrieval hot paths.

### DIFF
--- a/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLogger.java
+++ b/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLogger.java
@@ -97,28 +97,26 @@ public class BigQueryJdbcCustomLogger extends Logger {
 
       for (StackTraceElement element : stackTrace) {
         String className = element.getClassName();
-        if (shouldSkipClass(className)) {
-          continue;
+        if (isDriverClass(className) && !isLoggerClass(className)) {
+          sourceClass = className;
+          sourceMethod = element.getMethodName();
+          break;
         }
-        sourceClass = className;
-        sourceMethod = element.getMethodName();
-        break;
       }
       super.setSourceClassName(sourceClass);
       super.setSourceMethodName(sourceMethod);
     }
 
-    private static boolean shouldSkipClass(String className) {
+    private static boolean isDriverClass(String className) {
+      return className.startsWith("com.google.cloud.bigquery.jdbc.")
+          || className.startsWith("com.google.cloud.bigquery.exception.");
+    }
+
+    private static boolean isLoggerClass(String className) {
       return className.equals("com.google.cloud.bigquery.jdbc.BigQueryJdbcCustomLogger")
           || className.equals("com.google.cloud.bigquery.jdbc.BigQueryJdbcResultSetLogger")
           || className.startsWith("com.google.cloud.bigquery.jdbc.BigQueryJdbcRootLogger")
-          || className.equals(BigQueryJdbcLogRecord.class.getName())
-          || className.startsWith("java.util.logging.")
-          || className.startsWith("sun.util.logging.")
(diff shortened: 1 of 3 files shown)


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLogger.java
```diff
@@ -42,28 +42,83 @@ private void logWithCaller(Level level, Throwable thrown, Supplier<String> msgSu
...
-      LogRecord record = new LogRecord(level, msgSupplier.get());
-      record.setSourceClassName(sourceClass);
-      record.setSourceMethodName(sourceMethod);
-      record.setThrown(thrown);
-      log(record);
+    private static boolean shouldSkipClass(String className) {
+      return className.equals("com.google.cloud.bigquery.jdbc.BigQueryJdbcCustomLogger")
+          || className.equals("com.google.cloud.bigquery.jdbc.BigQueryJdbcResultSetLogger")
+          || className.startsWith("com.google.cloud.bigquery.jdbc.BigQueryJdbcRootLogger")
+          || className.equals(BigQueryJdbcLogRecord.class.getName())
```
logachev: Is it more reliable to ensure that className is from our driver rather than listing various logging libraries? Do we have any logs from non-jdbc here?

keshavdandeva (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcCustomLogger.java
Neenu1995: Changed this to use a whitelist of jdbc classes instead of blacklisting external classes.

logachev (approved):

PR #13372, opened 2026-06-05
chore(.github/workflows): let librarian update workflow also sync new commit to generation_config.yaml

zhumin8:
Add this step temporarily so that this workflow also triggers hermetic build generation workflow, and we can verify changes in the same PR.
This change is intended to be set till 06/10/2026 to verify librarian generate against hermetic build. 

Here is the sequence:
- nightly, this scheduled workflow updates librarian.yaml with new googleapis commit.
- [additional step] update generation_config.yaml with same commit.
- run `librarian generate -all` and add code change to commit
- raise PR.
- hermetic build [generation workflow](https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/hermetic_library_generation.yaml) triggered because generation_config.yaml `googleapis_commitish` is updated.

On the next day, we verify the results on this PR, expect hermetic build [generation workflow] runs succesfully with no code change added. If any, we should investigate the diff.

Note: hermetic build workflow will not trigger on this PR, because I pushed from fork. This is to avoid it adding code changes to this PR.

### DIFF
--- a/.github/workflows/update_librarian_googleapis.yaml
+++ b/.github/workflows/update_librarian_googleapis.yaml
@@ -76,14 +76,14 @@ jobs:
         sed -i -e "s/^googleapis_commitish.*$/googleapis_commitish: ${{ steps.commit.outputs.new_commit }}/" generation_config.yaml
         echo "After update:"
         grep "^googleapis_commitish" generation_config.yaml
-    - name: Detect Changes To Configs
-      id: detect_config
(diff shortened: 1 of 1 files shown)


zhumin8:
dry run updates commit as expected: https://screenshot.googleplex.com/4JrQaWBWzaMKCLh

JoeWang1127 (approved):

PR #13377, opened 2026-06-05
feat(bigquery-jdbc): parse endpoint overrides and proxy settings from URL authority

keshavdandeva:
b/515051004

Added support for extracting endpoint overrides and proxy routing configurations directly from the authority segment (the host part preceding the first semicolon `;`) of the JDBC connection URL. 

*   **Endpoint Overrides**: If the authority segment starts with `http://` or `https://`, it is mapped to the `EndpointOverrides` property (`BIGQUERY=...`).
*   **Proxy Settings**: If the authority does not contain a scheme, it is parsed as a proxy (e.g., `host:port`) and mapped to the `ProxyHost` and `ProxyPort` properties.
*   **Parameter Priority**: Properties specified after the semicolon take precedence. In the case of `EndpointOverrides`, settings from the authority and parameter segments are merged, with parameter values overwriting duplicate keys.


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcUrlUtility.java
```diff
@@ -866,4 +881,111 @@ static Map<String, String> parsePropertiesMapFromValue(
...
+          BIGQUERY_ENDPOINT_OVERRIDE_PROPERTY_NAME + "=" + normalizeEndpoint(authority));
+      return;
+    }
+
+    if (authority.startsWith("[")) {
+      int closingBracketIndex = authority.indexOf(']');
+      if (closingBracketIndex != -1) {
+        String host = authority.substring(0, closingBracketIndex + 1).trim();
+        String rest = authority.substring(closingBracketIndex + 1).trim();
+        map.put(PROXY_HOST_PROPERTY_NAME, host);
```
logachev, line 912: What is the spec saying about this? Unless spec prescribes it to be proxy, it's not obvious behavior..

e.g. `jdbc:bigquery:https://bigquery.googleapis.com:443;` and `jdbc:bigquery:bigquery.googleapis.com:443` seem identical, but removing `https://` part turns Endpoint override into a proxy..

keshavdandeva: The official JDBC spec is agnostic and leaves the URL structure after `jdbc:<subprotocol>:` entirely up to individual driver implementations.

PR #13378, opened 2026-06-05
chore: avoid GitHub Search API rate-limit during README generation

jinseopkim0:
b/520464446


generate-readme.py
```diff
@@ -122,34 +124,29 @@ def client_for_module(module) -> Optional[CloudClient]:
...
-
-def _fetch_repo_list(page):
-  url = "https://api.github.com/search/repositories"
-  response = requests.get(url, params = {
-    'q': 'org:googleapis is:public archived:false language:java',
-    'per_page': 100,
-    'page': page,
-  })
-  return response.json()['items']
+EXTERNAL_REPOS = [
```
blakeli0: Is this generate-readme.py used by these external repos as well?

generate-readme.py
blakeli0, line 1: I'm not sure what the problem was. Do you mind sharing more background info? And why we chose this fix?

generate-readme.py
jinseopkim0: No, the script is not used by external repos. These external repos are included in the readme though.

generate-readme.py
jinseopkim0: `_fetch_repo_list` was causing issues, because it was making GitHub Search API calls to get the list of `java-*` modules. This was rate-limited, causing errors (e.g. `KeyError: 'items'`). `_fetch_repo_list` has now been replaced with a glob of `java-*` within the monorepo + `EXTERNAL_REPOS` (which is a static list).

generate-readme.py
blakeli0: I see, this actually could be a good chance to remove them from our README. We don't maintain them and they are not really Cloud APIs either (except for pubsublite which is also deprecating soon).

README.md
```diff
@@ -226,12 +260,18 @@ Libraries are available on GitHub and Maven Central for developing Java applicat
...
+| [Security Compliance API](https://github.com/googleapis/google-cloud-java/tree/main/java-cloudsecuritycompliance) | [![preview][preview-stability]][preview-description] | [![Maven](https://img.shields.io/maven-central/v/com.google.cloud/google-cloud-cloudsecuritycompliance.svg)](https://search.maven.org/search?q=g:com.google.cloud%20AND%20a:google-cloud-cloudsecuritycompliance&core=gav) |
 | [Service Health API](https://github.com/googleapis/google-cloud-java/tree/main/java-servicehealth) | [![preview][preview-stability]][preview-description] | [![Maven](https://img.shields.io/maven-central/v/com.google.cloud/google-cloud-servicehealth.svg)](https://search.maven.org/search?q=g:com.google.cloud%20AND%20a:google-cloud-servicehealth&core=gav) |
+| [Showcase](https://github.com/googleapis/google-cloud-java/tree/main/java-showcase) | [![preview][preview-stability]][preview-description] | [![Maven](https://img.shields.io/maven-central/v/com.google.cloud/gapic-showcase.svg)](https://search.maven.org/search?q=g:com.google.cloud%20AND%20a:gapic-showcase&core=gav) |
```
blakeli0: Can we exclude showcase from the list since it is a test API?

generate-readme.py
jinseopkim0: Thanks for the feedback, done.

README.md
jinseopkim0: Thanks for the feedback, done.

blakeli0 (approved):

PR #13383, opened 2026-06-06
fix(bigquery): retry getTable server errors

### LINKED ISSUE #12885: Default retry logic does not retry HTTP 503 errors from BigQuery API on `BigQuery.getTable()`
# [bigquery]: Default retry logic does not retry HTTP 503 errors from BigQuery API on `BigQuery.getTable()`

When a BigQuery RPC such as `BigQuery.getTable(...)` receives a 503 `backendError` response, the call fails immediately on the first attempt. The configured `RetrySettings` (`maxAttempts`, backoff, etc.) are not honored, even though 503 is declared as retryable by the SDK itself in `BigQueryException.RETRYABLE_ERRORS`.

#### Environment details

1. API: BigQuery
3. Java version: 21
4. Version(s):
   - `com.google.cloud:google-cloud-bigquery:2.62.0`
   - `com.google.http-client:google-http-client-test:1.44.1` (used only by the reproduction below to mock the transport)

#### Steps to reproduce

1. Build and run the code under "Code example" with `com.google.cloud:google-cloud-bigquery:2.62.0`.
2. The mock transport returns HTTP 503 on the first request and HTTP 200 on the second. With `RetrySettings.setMaxAttempts(5)` configured, two requests are expected (first is retried, second succeeds).
3. Observe that only **one** request is sent and a `BigQueryException` is thrown immediately - the retry never happens.

#### Code example

```java
package com.example;

import com.google.api.client.testing.http.MockHttpTransport;
import com.google.api.client.testing.http.MockLowLevelHttpRequest;
import com.google.api.client.testing.http.MockLowLevelHttpResponse;
import com.google.api.gax.retrying.RetrySettings;
import com.google.auth.oauth2.AccessToken;
import com.google.auth.oauth2.GoogleCredentials;
import com.google.cloud.bigquery.BigQuery;
import com.google.cloud.bigquery.BigQueryOptions;
import com.google.cloud.bigquery.Table;
import com.google.cloud.bigquery.TableId;
import com.google.cloud.http.HttpTransportOptions;
import org.threeten.bp.Duration;

import java.util.Date;

public class Repro {

        MockHttpTransport mockTransport = new MockHttpTransport() {
            private int requestCount = 0;

            @Override
            public MockLowLevelHttpRequest buildRequest(String method, String url) {
                return new MockLowLevelHttpRequest() {
                    @Override
                    public MockLowLevelHttpResponse execute() {
                        requestCount++;
                        System.out.println("[Mock] Intercepted request #" + requestCount + " to: " + url);

                        if (requestCount < 2) {
                            String jsonError = """
                                    {
                                      "error": {
                                        "code": 503,
                                        "message": "Visibility check was unavailable. Please retry the request",
                                        "errors": [{"reason": "backendError", "domain": "global"}]
                                      }
                                    }
                                    """;
                            return new MockLowLevelHttpResponse()
                                    .setStatusCode(503)
                                    .setContentType("application/json")
                                    .setContent(jsonError);
                        }

                        String tableJson = """
                                {
                                  "kind": "bigquery#table",
                                  "id": "test-project:test_dataset.test_table",
                                  "type": "TABLE",
                                  "tableReference": {
                                    "projectId": "test-project",
                                    "datasetId": "test_dataset",
                                    "tableId": "test_table"
                                  }
                                }
                                """;
                        return new MockLowLevelHttpResponse()
                                .setStatusCode(200)
                                .setContentType("application/json")
                                .setContent(tableJson);
                    }
                };
            }
        };

        BigQueryOptions options = BigQueryOptions.newBuilder()
                .setProjectId("test-project")
                .setCredentials(GoogleCredentials.create(
                        new AccessToken("dummy-token", new Date(System.currentTimeMillis() + 3600000))))
                .setRetrySettings(RetrySettings.newBuilder()
                        .setMaxAttempts(5)
                        .setInitialRetryDelay(Duration.ofSeconds(1))
                        .setRetryDelayMultiplier(2.0)
                        .setMaxRetryDelay(Duration.ofSeconds(10))
                        .setTotalTimeout(Duration.ofSeconds(30))
                        .build())
                .setTransportOptions(HttpTransportOptions.newBuilder()
                        .setHttpTransportFactory(() -> mockTransport)
                        .build())
                .build();

        BigQuery bigquery = options.getService();

        Table table = bigquery.getTable(TableId.of("test_dataset", "test_table"));
        System.out.println("Got table: " + table.getTableId().getTable());
    }
}
```

#### Stack trace

```
[Mock] Intercepted request #1 to: https://bigquery.googleapis.com/...
Exception in thread "main" com.google.cloud.bigquery.BigQueryException: Visibility check was unavailable. Please retry the request
    at com.google.cloud.bigquery.BigQueryRetryHelper.runWithRetries(...)
    ...
```

Only one request is made. Expected: two requests, then a successful `Table`.

#### External references such as API reference guides

- `com.google.cloud.bigquery.BigQueryException.RETRYABLE_ERRORS` in `googleapis/java-bigquery` - lists `{500, 502, 503, 504}` as retryable.
- `BigQueryOptions.Builder.setResultRetryAlgorithm(...)` - the public hook that currently has to be used to make 5xx retries actually happen.

**Expected behavior.** A 503 `backendError` response is retried according to the configured `RetrySettings`, because:

- `BigQueryException.RETRYABLE_ERRORS` explicitly lists `{500, 502, 503, 504}` as retryable.
- BigQuery's public documentation describes 5xx as transient and retryable.
- A user configuring `RetrySettings.setMaxAttempts(5)` reasonably expects 5xx to be included.

**Actual behavior.** The request fails on the first attempt. Total elapsed time is a single RPC round-trip - no backoff, no retry.

**Why this matters / how it was discovered.** The issue surfaced during a production incidents: the BigQuery API (`GET https://bigquery.googleapis.com/bigquery/v2/projects/<EXMAPLE_PROJECT>/datasets/<EXAMPLE_DATASET>>/tables/<EXAMPLE_TABLE>?prettyPrint=false&view=STORAGE_STATS`) sporadically returned

```json
{
  "error": {
    "code": 503,
    "message": "Visibility check was unavailable. Please retry the request",
    "errors": [{"reason": "backendError", "domain": "global"}]
  }
}
```

in response to `bigquery.getTable(...)` calls. The default retry policy in `google-cloud-bigquery` did not retry these errors, and there is no straightforward configuration-level way for users to enable retrying of 5xx responses without replacing the retry algorithm wholesale via `setResultRetryAlgorithm(...)`.

Happy to provide any additional context if useful.

goutamadwant:
Translate HTTP response exceptions thrown by BigQuery.getTable before retry evaluation so the existing BigQueryException retryability rules can retry transient server errors such as HTTP 503.

This keeps the change scoped to getTable. It does not change job retry behavior or raw network IOException handling.

Fixes #12885

Tests:

mvn -pl java-bigquery/google-cloud-bigquery -Dtest=BigQueryImplTest#testGetTableFailureShouldRetryServerErrors -DskipITs test
mvn -pl java-bigquery/google-cloud-bigquery -Dtest=BigQueryImplTest -DskipITs test
mvn -pl java-bigquery/google-cloud-bigquery -DskipITs test
mvn -pl java-bigquery/google-cloud-bigquery -DskipITs package


java-bigquery/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java
```diff
@@ -1123,11 +1124,15 @@ && getOptions().getOpenTelemetryTracer() != null) {
...
-                      optionsMap);
+                  try {
+                    return bigQueryRpc.getTableSkipExceptionTranslation(
+                        completeTableId.getProject(),
+                        completeTableId.getDataset(),
+                        completeTableId.getTable(),
+                        optionsMap);
+                  } catch (HttpResponseException e) {
+                    throw new BigQueryException(e);
+                  }
```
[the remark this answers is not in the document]
goutamadwant: @lqiu96 As per the gemini's suggestion does below scope make sense?

the current change only covers `getTable`, but `getDataset`, `getModel` and `getRoutine` use the same pattern - the `*SkipExceptionTranslation` RPC call runs inside `BigQueryRetryHelper.runWithRetries(...)`, so a raw `HttpResponseException` can avoid the normal `BigQueryException` translation before retry handling.

So my plan is to keep the follow-up scoped to those resource getter paths only:

- `getDataset`
- `getModel`
- `getRoutine`

I am **not** planning to wrap every `*SkipExceptionTranslation` call in this PR, since that would widen the behavior change beyond this retry issue.

I will add matching tests for the three paths with a first 503 response and a successful second response, similar to the new `getTable` coverage.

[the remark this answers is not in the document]
lqiu96: Thanks, I think that makes sense. We can create the necessary follow up for the other relevant RPCs. For now, let's try this with getTable.

lqiu96: Thanks for this. IIUC the root of the issue is that the default configuration isn't retrying on a `HttpResponseException` due to how it's configured in BaseService. 

I think this will work, but I have a small concern about behavioral compatibility now that we are throwing a new exception (it may be a small possibility, but I'll need to do a bit more digging into this). BigQuery is an old library and I think it is possible that users may have written a custom ResultRetryAlgorithm to handle these cases since we weren't natively supporting it (e.g. ResultRetryAlgorithm catch specifically for `HttpResponseException` and parse out the status code). I don't want to wrap the exception and have the behavior changing for existing users.

One idea I have in mind is something for max compatibility is like:
```
private ResultRetryAlgorithm<Object> getDefaultRetryAlgorithm() {
  final ResultRetryAlgorithm<?> configuredAlgorithm = getOptions().getResultRetryAlgorithm();
  // 1. If the user configured a custom algorithm, respect it completely (do not apply fallback)
  if (configuredAlgorithm != BigQueryBaseService.DEFAULT_BIGQUERY_EXCEPTION_HANDLER) {
    ResultRetryAlgorithm<Object> customAlgorithm = (ResultRetryAlgorithm<Object>) configuredAlgorithm;
    return customAlgorithm;
  }
  // 2. If they are using the default handler, wrap it to add transient HTTP retries (like 503)
  return new ResultRetryAlgorithm<Object>() {
    @Override
    public TimedAttemptSettings createNextAttempt(
        Throwable previousThrowable, Object previousResponse, TimedAttemptSettings previousSettings) {
      return null; // Delegate timing to TimedRetryAlgorithm
    }
    @Override
    public boolean shouldRetry(Throwable previousThrowable, Object previousResponse) {
      // Check default transient HTTP rules first
      if (previousThrowable instanceof HttpResponseException) {
        int statusCode = ((HttpResponseException) previousThrowable).getStatusCode();
        if (statusCode == 500 || statusCode == 502 || statusCode == 503 || statusCode == 504) {
          return true;
        }
      }
      
      // Fall back to default ExceptionHandler rules (SocketException, ConnectException, etc.)
      ResultRetryAlgorithm<Object> delegate = (ResultRetryAlgorithm<Object>) configuredAlgorithm;
      return delegate.shouldRetry(previousThrowable, previousResponse);
    }
  };
}
```

And if this makes sense, we use this default for the relevant RPCs (getTable, ... , etc).

goutamadwant: Thanks @lqiu96 i think that makes sense. I agree the current wrapping can change what a custom ResultRetryAlgorithm sees.

I am planning to change the approach like this:

- leave getTableSkipExceptionTranslation throwing the original HttpResponseException
- if the user configured a custom ResultRetryAlgorithm, pass it through unchanged
- only when the configured algorithm is the default BigQuery exception handler, wrap it with a small default-only retry check for transient HTTP response codes
- keep this PR scoped to getTable and leave other RPCs for a follow up

**Rough shape of the change ( Same to what you proposed ) :**

```
ResultRetryAlgorithm<?> getTableRetryAlgorithm() {
  ResultRetryAlgorithm<?> configured = getOptions().getResultRetryAlgorithm();

  if (configured != BigQueryBaseService.DEFAULT_BIGQUERY_EXCEPTION_HANDLER) {
    return configured;
  }

  return new ResultRetryAlgorithm<Object>() {
    @Override
    public boolean shouldRetry(Throwable previousThrowable, Object previousResponse) {
      if (previousThrowable instanceof HttpResponseException) {
        int code = ((HttpResponseException) previousThrowable).getStatusCode();
        if (code == 500 || code == 502 || code == 503 || code == 504) {
          return true;
        }
      }

      return defaultHandler.shouldRetry(previousThrowable, previousResponse);
    }

    @Override
    public TimedAttemptSettings createNextAttempt(
        Throwable previousThrowable, Object previousResponse, TimedAttemptSettings previousSettings) {
      return defaultHandler.createNextAttempt(previousThrowable, previousResponse, previousSettings);
    }
  };
}
```

I will also add a regression test for the custom retry case, so we verify custom algorithms still see the raw HttpResponseException.

Let me know if this is good and i can make these changes

lqiu96: I think this looks fine to me. 

I took a another look and it's hard for us to know all the possible use cases. I think this provides the most safety (granted it's not the clean as just wrapping it) for all users. Thanks for accomodating the larger change.

goutamadwant: Thanks @lqiu96 Updated this now with code and tests as well. Let me know if this is good or any suggestions.

lqiu96: Hey @goutamadwant, your changes look fine to me and thanks for our discussions. We have some internal asks regarding this issue and we are prioritizing the same changes in https://github.com/googleapis/google-cloud-java/pull/13416 to help meet them sooner. Basically the changes over there based on our discussions and use this PR as a reference/ idea.

We need these changes sooner, so I'm going to close this PR in favor of that one. I know this is less than ideal, but I want to thank you again for all your time spent on this.

lqiu96:
Closing in favor of https://github.com/googleapis/google-cloud-java/pull/13416 (ideas are the same but quickly expanded due to timelines)

PR #13397, opened 2026-06-08
fix(firestore): rename internal Version class to FirestoreAdminVersion to avoid duplicate class conflict

blakeli0:
bom-content-test [failed](https://github.com/googleapis/java-cloud-bom/actions/runs/27166666222/job/80195524697?pr=7496) in the releasing PR of java-cloud-bom due to duplicated class error.
This would result in build-time failures in customer's applications if they have static analysis tools enabled.

As a short term fix, replaced the internal Version class in google-cloud-firestore-admin with FirestoreAdminVersion to avoid the error. 

In the long term, we should have static checks in this repo to prevent this from happening.


java-firestore/owlbot.py
```diff
@@ -38,10 +38,29 @@
...
+        )
+    else:
+        s.move(library)
+
+    # Replace Version.VERSION with FirestoreAdminVersion.VERSION in stub files
+    s.replace(
+        "google-cloud-firestore-admin/src/main/java/com/google/cloud/firestore/v1/stub/**/*.java",
+        "Version.VERSION",
+        "FirestoreAdminVersion.VERSION",
+    )
```
jinseopkim0: Would it make sense to move this to be within the `if "firestore-admin" in str(library):` block?

blakeli0: Thanks, updated!

jinseopkim0 (approved):

java-firestore/google-cloud-firestore-admin/src/main/java/com/google/cloud/firestore/v1/stub/FirestoreAdminStubSettings.java
```diff
@@ -757,7 +757,7 @@ protected LibraryMetadata getLibraryMetadata() {
     return LibraryMetadata.newBuilder()
         .setArtifactName("com.google.cloud:google-cloud-firestore")
         .setRepository("googleapis/google-cloud-java")
-        .setVersion(Version.VERSION)
+        .setVersion(FirestoreAdminVersion.VERSION)
```
zhumin8, line 760: @blakeli0 I think this is a generated class. If you really need to keep this change, you can add this 
class to `keep` in librarian.yaml. But that preserves the class, risking not getting any new updates in the future.
https://github.com/googleapis/google-cloud-java/actions/runs/27170876639

blakeli0: Yes I think that's fine. I think Version.java in other modules are getting the same treatment as well.

zhumin8: And the reason the daily generation PR did not had check failed is that it re-generated, basically reverted changes to java-firestore/google-cloud-firestore-admin/src/main/java/com/google/cloud/firestore/v1/stub/FirestoreAdminStubSettings.java, and created a new Version.java file since no one existed (renamed to FirestoreAdminVersion.java)
See https://github.com/googleapis/google-cloud-java/pull/13399/changes#diff-092f3430cf5fa92d46108113931ad91f6cb58246f3c560a78a09a457bd0ed9a5

PR #13412, opened 2026-06-10
feat(deps): add jspecify to shared dependencies

nnicolee:
b/522246866

### DIFF
--- a/sdk-platform-java/java-shared-dependencies/third-party-dependencies/pom.xml
+++ b/sdk-platform-java/java-shared-dependencies/third-party-dependencies/pom.xml
@@ -262,6 +262,11 @@
         <artifactId>tink</artifactId>
         <version>${com.google.crypto.tink.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.jspecify</groupId>
+        <artifactId>jspecify</artifactId>
+        <version>${jspecify.version}</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 </project>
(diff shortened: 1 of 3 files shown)


lqiu96 (approved):
Approving following approval from internal ticket

diegomarquezp (approved):

sdk-platform-java/java-shared-dependencies/third-party-dependencies/pom.xml
```diff
@@ -45,6 +45,7 @@
     <com.google.crypto.tink.version>1.16.0</com.google.crypto.tink.version>
     <io.opentelemetry.contrib.opentelemetry-gcp-resources.version>1.45.0-alpha</io.opentelemetry.contrib.opentelemetry-gcp-resources.version>
     <json.version>20250517</json.version>
+    <jspecify.version>1.0.0</jspecify.version>
```
blakeli0: Do we plan to use it in gax? If yes, we may want to follow how opentelemetry is managed. That the version is managed in [gapic-generator-java-pom-parent](https://github.com/googleapis/google-cloud-java/blob/9aab7eb513adecc44bcc85530971f9dc83d633dd/sdk-platform-java/gapic-generator-java-pom-parent/pom.xml#L34), and both `gax-java` and `third-party-dependencies` extend from `gapic-generator-java-pom-parent`.

nnicolee: Hi yes, we plan to use it in gax! Thanks for the suggestion, I will update the PR!

PR #13413, opened 2026-06-10
feat(bigquery-jdbc): support connection-scoped executor isolation and dynamic queue warnings

Neenu1995:
b/519201498

This PR implements connection-scoped thread pool isolation and dynamic saturation monitoring for the BigQuery JDBC driver.

* **Connection-Scoped Pools**: Spawns independent `queryExecutor` and `metadataExecutor` instances per connection, properly cleaning them up on connection close.
* **Daemon Thread Factory**: Configures the connection thread factory to mark all executor threads as daemon threads, preventing JVM shutdown hangs.
* **Hysteresis Saturation Warnings**: Implements lightweight queue saturation logging that triggers dynamically when the queue reaches `max(10, corePoolSize * 5)` and resets only when it recovers below `max(4, corePoolSize * 2)`.
* **Testing Improvements**: Refactors `BigQueryJdbcMdcTest` to inherit from `BigQueryJdbcLoggingBaseTest`, and adds unit tests to verify lifecycle, saturation logging, and hysteresis behavior.


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -954,6 +965,44 @@ private void closeImpl() throws SQLException {
         statement.close();
       }
       this.openStatements.clear();
+
+      if (this.metadataExecutor != null) {
```
logachev, line 953: I think we should update this `closeImpl()` to call all shutdown ops first & do await later, right now it awaits each shutdown individually.

We could force `shutdownNow` in Interrupted exception handler that already exists & we need to do it for all (including read/write clients), no need to handle them independently (I'd assume shutdownNow is no-op if was already successful)

Also, if Interrupted is thrown during `bigQueryWriteClient.awaitTermination`, openStatements is not properly cleaned up.

logachev (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
Neenu1995: Done.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
logachev: We don't need coreThreads to match maxThreads imo; I think coreThreads being 3-4 would be sufficient.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -102,11 +133,37 @@ public MdcThreadPoolExecutor(
       super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, threadFactory);
     }
 
+    private final AtomicBoolean warningLogged = new AtomicBoolean(false);
+
+    private void monitorQueueSaturation(int queueSize) {
+      int corePoolSize = getCorePoolSize();
```
logachev: I guess it needs to be maxPoolSize? Exceeding core is fine, new threads are being added.

logachev (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
Neenu1995: done.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
Neenu1995: With our unbounded queue design, if corePoolSize is set to 3-4 while maximumPoolSize is 32, the executor will never spawn more than 4 threads, ignoring the user's config. To allow the pool to scale up to the full 32 threads under load, the core size must match the max size. I added core thread timeouts to clean them up when idle.

PR #13417, opened 2026-06-10
test(bigquery): warm up client in beforeClass to stabilize testFastSQLQueryMultiPage

jinseopkim0:
b/467064659


lqiu96 (approved):
LGTM. Let's see if this helps with the query tests. If not, maybe we can try and extend the timeouts/ add retries

PR #13427, opened 2026-06-11
feat(storage): log additional bytes received from GCS in read path

nidhiii-27:
Fixes bug 475824752. Tracks the number of bytes received from the transport layer and emits a warning log if GCS sends more bytes than requested by the user during explicit range requests.

[Generated-by: AI]


nidhiii-27:
Addressed all PR review comments.

Dhriti07:
Can you add a tests for Apiary path as well ?

nidhiii-27:
Added unit tests verifying the extra bytes logging for the Apiary path in ApiaryUnbufferedReadableByteChannelTest.

_Co-authored by AI Agent_

java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryUnbufferedReadableByteChannel.java
```diff
@@ -182,9 +184,25 @@ public boolean isOpen() {
...
   @Override
   public void close() throws IOException {
-    open = false;
-    if (sbc != null) {
-      sbc.close();
+    try {
+      long requestedLength = apiaryReadRequest.getByteRangeSpec().length();
+      if (requestedLength >= 0
+          && requestedLength < ByteRangeSpec.EFFECTIVE_INFINITY
+          && totalBytesReadFromNetwork > requestedLength) {
```
Dhriti07, line 191: Could you double-check if we are actually measuring the total bytes sent by GCS over the network? I was wondering if it might just be measuring the bytes our client chooses to consume from the InputStream instead.

nidhiii-27: You are right—in Apiary, `totalBytesReadFromNetwork` strictly measures the bytes our client actively consumes from the `InputStream`, not the raw bytes arriving over the network socket.

Because Apiary uses a pull-based HTTP stream (unlike Gapic's background gRPC thread which tracks `fetchOffset` directly off the wire), if GCS sends extra bytes but the user's application stops reading exactly at the requested limit and calls `close()`, we won't see those extra bytes and the warning won't trigger. We only catch the over-fetch if the application continues reading until it hits `EOF`.

Given the limitations of the HTTP client's pull model, this is the closest proxy we have for detecting over-fetching.

_Co-authored by AI Agent_

Dhriti07: Given that this is measuring the client consumed bytes, Does that mean that this check will never be triggered ? Can we check how other languages like go and handling these cases or if they have a different architecture that doesn't have this problem ?

nidhiii-27: 1. **Triggering of the warning:** Yes, this check will trigger if the client application reads in a loop until EOF (i.e. until `read()` returns `-1`), which is standard practice when consuming streams or when using copy utilities. If the caller manually stops reading at exactly the requested length and immediately closes the stream, the warning won't fire. But this is the closest proxy we have to track if GCS sent extra bytes.
2. **How Go handles this:** Go's implementation is identical. The Go client tracks the read count at the user-facing `Reader` layer, decrementing a `remain` counter on each `Read()` call. On `Close()`, if `remain` is less than 0, it logs a warning: `storage: received %d more bytes than requested from GCS...` (see [reader.go:398](https://github.com/googleapis/google-cloud-go/blob/main/storage/reader.go#L398)).

_Co-authored by AI Agent_

Dhriti07 (approved):

PR #13435, opened 2026-06-11
feat(bigquery-jdbc): respect standard JVM trustStore properties by default

keshavdandeva:
b/515129164

### Problem
In enterprise corporate networks (e.g. Zscaler), outbound HTTPS traffic is intercepted by transparent proxies doing SSL MITM (man-in-the-middle) decryption. The proxy signs its re-encrypted connections with dynamically generated root CA certificates. 

The BigQuery JDBC driver failed to establish TLS connections under these environments because the underlying client library ignored JVM trust stores (the standard Java `cacerts` file or custom system properties set via `-Djavax.net.ssl.trustStore`). 
This happened because when direct connections had empty SSL/proxy settings, the driver returned `null` for `HttpTransportOptions`. This fallback triggered classpath SPI overrides or legacy defaults which invoked `GoogleNetHttpTransport.newTrustedTransport()`. That convenience constructor hardcodes trust exclusively to a bundled `google.p12` keystore, completely overriding JVM system properties.

### Solution
Simplified the driver's transport instantiation to align with the core Google Cloud Java SDK's network defaults:
1. **Direct Connections:** Modified `getHttpTransportOptions(...)` to unconditionally return a transport factory configured with a single `NetHttpTransport` instance (`new NetHttpTransport.Builder().build()`). This allows JSSE to handle TLS certificate validation using standard JVM system properties and `cacerts` natively. Bypassing the SPI loader prevents classpath hijacking.
2. **Explicit Proxy Connections:** Configured the Apache HTTP client builder inside `getHttpTransportFactory(...)` to unconditionally call `httpClientBuilder.useSystemProperties()`. This ensures that even when a proxy is set in the JDBC URL, Apache HttpClient still honors system-level properties like `-Djavax.net.ssl.trustStore`.

### Integration Testing: SSL/TLS Validation (`ITLocalSslValidationTest`)
Added `ITLocalSslValidationTest.java` to validate the loading and enforcement of custom SSL truststore configurations end-to-end.

* **Local Mock HTTPS Server:** Starts a lightweight local HTTPS server on a random port presenting a self-signed certificate. It mocks necessary BigQuery backend endpoints (`/queries` and `/jobs`) to satisfy basic driver query execution.
* **Process Isolation:** Runs each connection check in a separate, isolated JVM subprocess via `ProcessBuilder`. This is required to bypass JSSE's JVM-wide caching of the `-Djavax.net.ssl.trustStore` property.
* **Test Coverage:**
  * **Negative Case:** Verifies that connection attempts without a truststore fail with the expected `PKIX path building failed` handshake error (exit code `1`).
  * **Positive Case:** Verifies that connection attempts using our custom truststore (`localhost-truststore.jks`) succeed and complete query executions successfully (exit code `0`).
* **CI Integration:** Added to `ITPresubmitTests` to run automatically on every pull request. Since it uses local mocks, it requires **no GCP credentials** and executes in **under 2 seconds**.


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcProxyUtility.java
```diff
@@ -136,17 +137,15 @@ static HttpTransportOptions getHttpTransportOptions(
...
-      return null;
-    }
 
     HttpTransportOptions.Builder httpTransportOptionsBuilder = HttpTransportOptions.newBuilder();
     if (hasProxyOrSsl) {
       httpTransportOptionsBuilder.setHttpTransportFactory(
           getHttpTransportFactory(
               proxyProperties, sslTrustStorePath, sslTrustStorePassword, callerClassName));
+    } else {
+      final HttpTransport defaultTransport = new NetHttpTransport.Builder().build();
```
logachev: Why using final here?

logachev (commented):
lgtm, but before approving I'd want us to have an environment where we could validate it

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcProxyUtility.java
keshavdandeva: Removed. It was originally added because the variable is captured by a lambda, but since it is effectively final, the final keyword is redundant.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcProxyUtility.java
```diff
@@ -136,17 +137,15 @@ static HttpTransportOptions getHttpTransportOptions(
...
-      return null;
-    }
 
     HttpTransportOptions.Builder httpTransportOptionsBuilder = HttpTransportOptions.newBuilder();
     if (hasProxyOrSsl) {
       httpTransportOptionsBuilder.setHttpTransportFactory(
           getHttpTransportFactory(
               proxyProperties, sslTrustStorePath, sslTrustStorePassword, callerClassName));
+    } else {
+      HttpTransport defaultTransport = new NetHttpTransport.Builder().build();
```
Neenu1995: nit: This is an expensive call. We might want to make it private static.

Neenu1995 (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcProxyUtility.java
keshavdandeva: Done

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcProxyUtility.java
```diff
@@ -58,6 +59,7 @@ final class BigQueryJdbcProxyUtility {
       new BigQueryJdbcCustomLogger(BigQueryJdbcProxyUtility.class.getName());
   static final String validPortRegex =
       "^([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$";
+  private static final HttpTransport DEFAULT_TRANSPORT = new NetHttpTransport.Builder().build();
```
logachev, line 62: Is this the default used by the SDK?

logachev (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcProxyUtility.java
keshavdandeva: Yes

PR #13445, opened 2026-06-11
chore(release): bump biglake to 0.81.1 and gapic-libraries-bom to 1.87.1

blakeli0:
Bumps java-biglake to 0.81.1 and gapic-libraries-bom to 1.87.1.


versions.txt
```diff
@@ -1,7 +1,7 @@
 # Format:
 # module:released-version:current-version
 
-google-cloud-java:1.87.0:1.87.0
+google-cloud-java:1.87.1:1.87.1
```
zhumin8, line 4: you also need to update https://github.com/googleapis/google-cloud-java/blob/291add8ed59c4323c003714db5728ce39a15270f/librarian.yaml#L1820-L1822

blakeli0: Updated

JoeWang1127 (approved):

PR #13452, opened 2026-06-12
chore: add backstory to librarian.yaml and re-generate

### LINKED ISSUE #6409: google-cloud-java: add back backstory to librarian.yaml
Right before librarian migration, a new library backstory is added. However, it is a special case that hermetic build cannot handle without changes. @blakeli0 made necessary changes to the old system and was able to generate and release this library. However, hermetic build is not fully compatible with this library and cannot continue to re-generate with it.

To allow daily hermetic build generation to continue to work, so during migration, we could compare to librarian re-generation results, this library is temporarily removed from generation_config.yaml:
https://github.com/googleapis/google-cloud-java/pull/13351

After turning down hermetic build daily workflow, add this library back to librarian.yaml. Note, this library needs extra tweaks in librarian.yaml because it is path under google/cloud and is a proto only library. Reference priors attempt [here](https://github.com/googleapis/google-cloud-java/pull/13347/changes/0865302c5436fc19dc5509053f05267908d8cb88).

zhumin8:
Followup to https://github.com/googleapis/google-cloud-java/pull/13351, add backstory back to librarian.yaml and re-generate code with librarian.

Manual add config to librarian.yaml and run:
```
V=$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
go run github.com/googleapis/librarian/cmd/librarian@${V} install
export PATH=$PATH:$HOME/.cache/librarian/bin/java_tools/bin
go run github.com/googleapis/librarian/cmd/librarian@${V} generate backstory
```
Then manually reverted version change in [gapic-libraries-bom/pom.xml](https://github.com/googleapis/google-cloud-java/pull/13452/commits/48525721316044f5b530b78145a65f50604d6a0b), this is known limitation during partial release. See https://github.com/googleapis/librarian/issues/6411.

Fixes https://github.com/googleapis/librarian/issues/6409


zhumin8:
note: librarian defaults to grpc+rest as transport for generation. For this library, since it is proto only, these are purely documentation diff.

librarian.yaml
```diff
@@ -433,6 +433,25 @@ libraries:
       api_description_override: n/a
       name_pretty_override: Cloud Batch
       product_documentation_override: https://cloud.google.com/
+  - name: backstory
+    version: 0.0.1-SNAPSHOT
```
blakeli0: The versions seem to be wrong

zhumin8: Good catch! My bad, did not update this config now that we already released this library.

Updated according to versions.txt and re-generated, also run `librarian tidy` that removed some unnecessary config that matches defaults.

blakeli0 (approved):

PR #13454, opened 2026-06-12
feat(bigquery-jdbc): migrate statement execution thread tracking to connection-scoped executor

Neenu1995:
b/519201498

This PR migrates the asynchronous processing tasks of the JDBC driver away from the legacy static global thread pool (`queryTaskExecutor`) to the modern, connection-scoped `ExecutorService` (`connection.getExecutorService()`).

### Key Changes
1. **Statement & Stream Refactoring**:
   - Removed the static `queryTaskExecutor` from `BigQueryStatement`.
   - Updated `populateArrowBufferedQueue` to submit streaming and buffering tasks to the connection-scoped executor pool.
   - Converted background thread tracking in `BigQueryStatement` and `BigQueryArrowResultSet` from raw `Thread` references to modern, cancelable `Future<?>` task handles.

2. **Isolated Backward-Compatibility Wrapper**:
   - Isolated the legacy raw thread wrapping to `BigQueryDatabaseMetaData` (which is still pending future refactoring to the executor pool).
   - Implemented a temporary `wrapThread(Thread)` helper method inside `BigQueryDatabaseMetaData` to wrap raw background threads into cancelable `Future<?>` adapters on the fly before passing them to the result set.
   - Hardened the anonymous `Future` wrapper state tracking to strictly satisfy the `Future` specification contract (handling `cancel`, `isCancelled`, and `isDone` correctly using a volatile boolean flag to prevent duplicate cancellation triggers).


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -348,7 +348,11 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
       this.headerProvider = createHeaderProvider();
       this.bigQuery = getBigQueryConnection();
       this.metadataExecutor = BigQueryJdbcMdc.newFixedThreadPool(metadataFetchThreadCount);
-      this.queryExecutor = BigQueryJdbcMdc.newCachedThreadPool();
+      // Use a bounded cached thread pool to prevent unbounded thread creation (and OOMs)
+      // under heavy load, while ensuring a limit (e.g., 100) high enough to prevent deadlocks
+      // between interdependent producer/consumer tasks (like nextPageWorker and
+      // populateBufferWorker).
+      this.queryExecutor = BigQueryJdbcMdc.newBoundedCachedThreadPool(100);
```
keshavdandeva: Should we also have a connection property similar to `metadataFetchThreadCount` for this with default being 100?

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
keshavdandeva, line 157: Is this log warning for `metadataFetchThreadCount` or the new `cachedTreadPool` for `statements` or both? In any case, let's make it more clearer

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
Neenu1995: done.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
Neenu1995: done.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -340,15 +341,20 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
...
       this.listenerPoolSize = ds.getListenerPoolSize();
       this.partnerToken = ds.getPartnerToken();
 
       this.headerProvider = createHeaderProvider();
       this.bigQuery = getBigQueryConnection();
+      // Fixed thread pool limits concurrent calls to prevent BigQuery metadata API throttling
+      // (independent tasks).
       this.metadataExecutor = BigQueryJdbcMdc.newFixedThreadPool(metadataFetchThreadCount);
-      this.queryExecutor = BigQueryJdbcMdc.newCachedThreadPool();
+      // Cached thread pool prevents deadlocks between interdependent producer/consumer query
```
logachev: nit: not sure this comment is necessary. thread pool doesn't prevent any deadlocks, it just provides a thread orchestration.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -5264,4 +5266,88 @@ private void loadDriverVersionProperties() {
...
+          long deadline = System.nanoTime() + remainingNanos;
+          while (thread.getState() != Thread.State.TERMINATED) {
+            if (isCancelled()) {
+              throw new CancellationException();
+            }
+            if (remainingNanos <= 0) {
+              throw new TimeoutException();
+            }
+            long remainingMillis = TimeUnit.NANOSECONDS.toMillis(remainingNanos);
+            if (remainingMillis <= 0) {
```
logachev: This will never happen, exception is thrown in that case.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJsonResultSet.java
```diff
@@ -291,10 +292,10 @@ private FieldValue getObjectInternal(int columnIndex) throws SQLException {
   public void close() {
     LOG.fineTrace("close", () -> String.format("Closing BigqueryJsonResultSet %s.", this));
     this.isClosed = true;
-    if (ownedThreads != null) {
-      for (Thread ownedThread : ownedThreads) {
-        if (!ownedThread.isInterrupted()) {
-          ownedThread.interrupt();
+    if (ownedTasks != null) {
+      for (Future<?> ownedTask : ownedTasks) {
+        if (ownedTask != null) {
```
logachev: We should not have `null` items in `ownedTasks` list.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryResultSetFinalizers.java
```diff
@@ -27,44 +28,44 @@ class BigQueryResultSetFinalizers {
...
     // Free resources. Remove all the hard refs
     public void finalizeResources() {
       LOG.finestTrace("finalizeResources");
-      if (ownedThreads != null) {
-        for (Thread ownedThread : ownedThreads) {
-          if (!ownedThread.isInterrupted()) {
-            ownedThread.interrupt();
+      if (ownedTasks != null) {
+        for (Future<?> ownedTask : ownedTasks) {
+          if (ownedTask != null) {
```
logachev: same here

logachev (commented):
Please split this PR in two separate PRs. Bridge for the Future<?> objects & changes to the thread pool are unrelated to each other. It will simplify the review.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
Neenu1995: Do you mean it should be ==0 instead of <= 0?

Neenu1995: done.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryJdbcMdc.java
```diff
@@ -86,6 +87,7 @@ static ExecutorService newFixedThreadPool(int nThreads) {
    */
   static ExecutorService newCachedThreadPool(ThreadFactory threadFactory) {
     return new MdcThreadPoolExecutor(
+        "Query Executor Pool",
```
logachev: nit: That should be an argument.. It is true now that cached is query pool, metadata is fixed, but this is not guaranteed

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -1048,34 +1064,60 @@ BigQueryJsonResultSet processJsonResultSet(TableResult results, Job job) {
...
-        parseAndPopulateRpcDataAsync(
-            schema, this.bigQueryFieldValueListWrapperBlockingQueue, rpcResponseQueue);
-    threadList.add(populateBufferWorker);
+      // Task to parse data received from the server to client library objects
+      Future<?> populateBufferWorker =
+          parseAndPopulateRpcDataAsync(
+              schema, this.bigQueryFieldValueListWrapperBlockingQueue, rpcResponseQueue);
+      taskList.add(populateBufferWorker);
+    } catch (Exception | OutOfMemoryError e) {
+      for (Future<?> task : taskList) {
```
logachev, line 1099: - What if there are tasks that were not created by the above code? Why would we want to cancel them?
- Do we need to cleanup taskList?

logachev (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
Neenu1995: taskList is a local list that only contains tasks created in the current execution of processJsonResultSet.
On exception we should cancel them to immediately terminate background execution and prevent resource leaks when query setup fails.
We don't have to explicitly cancel the taskList object. As a local variable, taskList is garbage-collected automatically when the method exits.

PR #13481, opened 2026-06-15
chore: migrate java-shared-config to monorepo

jinseopkim0:
b/519273974


.github/workflows/ci.yaml
```diff
@@ -219,6 +219,13 @@ jobs:
             - 'sdk-platform-java/**/*.java'
             - 'sdk-platform-java/java-shared-dependencies/**/pom.xml'
             - 'sdk-platform-java/gapic-generator-java-pom-parent/pom.xml'
+          java-shared-config:
+            - 'java-shared-config/**'
+            - 'google-auth-library-java/**/*.java'
```
blakeli0: java-shared-config is the parent of everything else, I don't think we want to trigger its CI when there are auth or sdk-platform-java changes

.github/workflows/java-shared-config-downstream-dependencies.yaml
```diff
@@ -0,0 +1,63 @@
...
+  dependencies:
+    needs: filter
+    if: ${{ needs.filter.outputs.library == 'true' }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [11]
+        repo:
+        - java-bigquery
```
blakeli0, line 31: Does this test against downstream repos? If yes, we need to migrate them to test against the mono repo.

.github/workflows/java-shared-config-downstream-maven-plugins.yaml
```diff
@@ -0,0 +1,107 @@
...
+    - run: .kokoro/client-library-check.sh ${{matrix.repo}} ${{matrix.job-type}}
+  javadoc-with-doclet:
+    needs: filter
+    if: ${{ needs.filter.outputs.library == 'true' }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        repo:
+        - java-bigtable
```
blakeli0, line 121: Same questions as https://github.com/googleapis/google-cloud-java/pull/13481/changes#r3417255625

java-shared-config/.cloudbuild/cloudbuild-test-a.yaml
blakeli0, line 1: Cloud build triggers in the GCP project need to be migrated as well.

.github/workflows/ci.yaml
jinseopkim0: Thanks for the feedback, updated.

.github/workflows/java-shared-config-downstream-dependencies.yaml
jinseopkim0: Thanks for the feedback. Yes, and I've updated the tests accordingly.

.github/workflows/java-shared-config-downstream-maven-plugins.yaml
jinseopkim0: Thanks for the feedback. Yes, and I've updated the tests accordingly.

java-shared-config/.cloudbuild/cloudbuild-test-a.yaml
jinseopkim0: Thanks for the feedback, updated.

.kokoro/presubmit/logging-logback-graalvm-native-presubmit.cfg
blakeli0, line 1: Is the deletion of this file expected?

.kokoro/presubmit/shared-config-integration.cfg
blakeli0, line 1: It does not seem right that we rename logging-logback-integration.cfg‎  to shared-config-integration.cfg to pass CI?

java-shared-config/native-image-shared-config/pom.xml
```diff
@@ -0,0 +1,291 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.cloud</groupId>
+  <artifactId>native-image-shared-config</artifactId>
+  <packaging>pom</packaging>
+  <version>1.17.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-shared-config:current} -->
+  <name>Native Image Shared Config</name>
+  <url>https://github.com/googleapis/java-shared-config</url>
```
blakeli0: The URL and SCM info below needs to be updated. There could be more links need to be updated to point to google-cloud-java now.

blakeli0:
Existing places (e.g. [sdk-platform-java-config/pom.xml](https://github.com/googleapis/google-cloud-java/blob/7f4c73a447832836fd76ca7bff349b2fb43d678f/sdk-platform-java/sdk-platform-java-config/pom.xml#L13-L18)) that point to released version of java-shared-config in this repo need to be updated to use relative path now.

.github/workflows/java-shared-config-downstream-maven-plugins.yaml
```diff
@@ -0,0 +1,164 @@
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+
+# Keeping this file separate as the dependencies check would use more
+# repositories than needed this downstream check for GraalVM native image and
+# other Maven plugins.
+name: java-shared-config downstream
```
blakeli0: both this yaml and the previous yaml have the same name

.github/workflows/java-shared-config-downstream-dependencies.yaml
blakeli0, line 1: Do you mind sharing how the downstream tests work now? I guess we test the downstream libraries in this repo with the changes in this PR?

.kokoro/presubmit/logging-logback-graalvm-native-presubmit.cfg
jinseopkim0: Thanks for the question. This is expected, as graalvm nightly tests of the monorepo will include these tests (e.g. https://github.com/googleapis/google-cloud-java/blob/64cde7f23a2a0186664bc459e5e91d773ebb0b29/.kokoro/nightly/graalvm-sub-jobs/native-a/common.cfg#L62). This is essentially a clean-up.

.kokoro/presubmit/shared-config-integration.cfg
jinseopkim0: Thanks for the question. This is actually not a rename, but a deletion and a new file creation. `logging-logback-integration.cfg` was deleted since `java-logging-logback` unit tests are now run as part of the monorepo CI (https://github.com/googleapis/google-cloud-java/blob/64cde7f23a2a0186664bc459e5e91d773ebb0b29/.github/workflows/ci.yaml#L172). This was essentially a clean-up.

`shared-config-integration.cfg` was created for build and compilation checks for the newly migrated java-shared-config module.

java-shared-config/native-image-shared-config/pom.xml
jinseopkim0: Thanks for the feedback, updated.

.github/workflows/java-shared-config-downstream-maven-plugins.yaml
jinseopkim0: Thanks for the feedback, updated.

.github/workflows/java-shared-config-downstream-dependencies.yaml
jinseopkim0: Thanks for the question. Yes, the downstream checks verify that the proposed changes to `java-shared-config` in the PR do not break dependency convergence or packaging for the main client libraries in the monorepo.

1. **Local Installation**: The workflow first compiles and installs the modified local version of `google-cloud-shared-config` from the PR branch into the runner's local Maven cache (`~/.m2/repository`) using `.kokoro/build.sh`.
2. **Mocking Downstream Usage**: The [.kokoro/client-library-check.sh](file:///.kokoro/client-library-check.sh) script is invoked. Since this script was migrated from the standalone repo, it currently:
   * Clones `google-cloud-java` at the latest stable tagged release.
   * Rewrites the parent pom reference in `sdk-platform-java-config` in the cloned repo to point to the local `google-cloud-shared-config` version we just built in Step 1.
3. **Validation**: It then runs `dependencies.sh` (to run the upper-bound dependency convergence check) and the `flatten-plugin-check` (to verify flattened POM dependency resolution) on the target client libraries (like BigQuery, Spanner, Storage) using the newly modified configuration.

There may be some optimizations possible to improve this test in the future, but for now, most of the existing testing logic has been re-used.

.kokoro/presubmit/logging-logback-graalvm-native-presubmit.cfg
jinseopkim0: On second thought, it would be better to follow-up for a clean-up. I've restored the file and created b/525280086 to track the clean-up.

.kokoro/presubmit/shared-config-integration.cfg
jinseopkim0: On second thought, it would be better to follow-up for a clean-up. I've restored the file and created b/525280086 to track the clean-up.

blakeli0 (approved):

PR #13482, opened 2026-06-15
chore(bigquery-jdbc): add the bridge Thread wrap ahead of Executor migration

Neenu1995:


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -5264,4 +5265,89 @@ private void loadDriverVersionProperties() {
...
+        }
+
+        @Override
+        public boolean isDone() {
+          return cancelled || thread.getState() == Thread.State.TERMINATED;
+        }
+
+        @Override
+        public Object get() throws InterruptedException, CancellationException {
+          if (isCancelled()) {
```
logachev: Instead of the code duplication, we can call `get(timeout, unit)` with some absurdly large limit, like 100 days.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -5264,4 +5265,89 @@ private void loadDriverVersionProperties() {
...
+              throw new TimeoutException();
+            }
+            long remainingMillis = TimeUnit.NANOSECONDS.toMillis(remainingNanos);
+            if (remainingMillis == 0) {
+              remainingMillis = 1;
+            }
+
+            long delay = Math.min(remainingMillis, 50);
+            if (thread.getState() == Thread.State.NEW) {
+              Thread.sleep(delay);
```
logachev, line 5333: Why the difference sleep vs join? 

Also we can just call `thread.join(1000);`, there is no need to check every 50ms if object was cancelled or not. We have 10-60 second cleanup waiters either way (or worry about < 50 ms remaining time)

Neenu1995: We use `sleep` for `NEW` threads because `join` returns immediately on unstarted threads, which would cause the `get()` loop to busy-wait and consume 100% CPU until the thread is started. Additionally, checking every 50ms allows the thread calling `get()` to detect a cancellation request quickly and throw a `CancellationException` without remaining blocked for a long `join` timeout.

logachev (approved):

PR #13484, opened 2026-06-15
refactor(test): migrate remaining wait loops to Awaitility in GAX tests

blakeli0:
This PR migrates all remaining wait loops and polling blocks in GAX tests to use Awaitility. This is the second part of b/505479343, following the initial introduction of Awaitility in #13458.

Specifically, this migrates:
- Semaphore64Test: replaced mock thread wait loops with Awaitility state waiting.
- ScheduledRetryingExecutorTest: migrated future attempt checking, attempt cancellation wait, and first attempt latch count down to Awaitility.
- BatcherImplTest: converted GC loops and thread state polling blocks to Awaitility.

### DIFF
--- a/sdk-platform-java/gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
+++ b/sdk-platform-java/gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
@@ -30,6 +30,7 @@
 package com.google.api.gax.retrying;
 
 import static com.google.api.gax.retrying.FailingCallable.FAST_RETRY_SETTINGS;
+import static org.awaitility.Awaitility.await;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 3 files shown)


sdk-platform-java/gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
```diff
@@ -99,26 +103,32 @@ void testSuccessWithFailuresPeekAttempt() throws Exception {
...
-              failedAttempts++;
-            }
-          }
-        }
-        Thread.sleep(0L, 100);
-      }
+      final AtomicInteger failedAttempts = new AtomicInteger(0);
+      final AtomicReference<ApiFuture<String>> lastSeenAttempt = new AtomicReference<>();
+      await()
+          .pollInterval(Duration.ofMillis(2))
```
lqiu96, line 109: do we want to poll more frequently for this? Seems like it would try 2/3 times at most?

lqiu96 (approved):
LGTM. simple awaitility refactor for tests. we can address any issues that come up in follow up PRs

sdk-platform-java/gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
blakeli0: Thanks for the suggestion! I think 2 millis are a good start, let's see if it causes issues.

lqiu96: Oh that's millis... I read that as seconds. Yeah, I think 2ms is more than fine.

PR #13497, opened 2026-06-16
chore(librarian): update librarian version to v0.21.0

sofisl:


librarian.yaml
```diff
@@ -1848,6 +1848,7 @@ libraries:
     keep:
       - src/main/java/io/grafeas/v1/stub/Version.java
     java:
+      alternate_headers: java-grafeas/license-header.txt
```
JoeWang1127: Why do we need to specify `java-grafeas`?

sofisl: oops that was a mistake, should be in a separate PR/

JoeWang1127 (approved):

blakeli0 (approved):

PR #13506, opened 2026-06-17
feat(storage): adding disabling option for bidi reads

Dhriti07:
Add option to disable Bidi Checksum using checksum disable system property


nidhiii-27 (approved):

java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/ReadAsChannel.java
```diff
@@ -111,7 +111,7 @@ boolean getCrc32cValidationEnabled() {
    */
   @BetaApi
   ReadAsChannel withCrc32cValidationEnabled(boolean enabled) {
-    if (enabled && Hasher.enabled().equals(hasher)) {
```
ShreyasSinha, line 114: why is did required?

ShreyasSinha (approved):

PR #13508, opened 2026-06-17
fix(bigquery-jdbc): explicitly set location when available for temporary dataset

Neenu1995:
b/524850173

### DIFF
--- a/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
+++ b/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
@@ -1375,11 +1375,15 @@ public class BigQueryStatement extends BigQueryNoOpsStatement {
     Dataset dataset = bigQuery.getDataset(DatasetId.of(datasetName));
     if (dataset == null) {
       LOG.info("Creating a hidden dataset: %s ", datasetName);
-      DatasetInfo datasetInfo =
+      DatasetInfo.Builder datasetInfoBuilder =
           DatasetInfo.newBuilder(datasetName)
-              .setDefaultTableLifetime(this.querySettings.getDestinationDatasetExpirationTime())
-              .build();
-      bigQuery.create(datasetInfo);
+              .setDefaultTableLifetime(this.querySettings.getDestinationDatasetExpirationTime());
+      if (this.connection != null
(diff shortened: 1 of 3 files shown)


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryStatement.java
```diff
@@ -1375,11 +1375,15 @@ private void checkIfDatasetExistElseCreate(String datasetName) {
...
+      DatasetInfo.Builder datasetInfoBuilder =
           DatasetInfo.newBuilder(datasetName)
-              .setDefaultTableLifetime(this.querySettings.getDestinationDatasetExpirationTime())
-              .build();
-      bigQuery.create(datasetInfo);
+              .setDefaultTableLifetime(this.querySettings.getDestinationDatasetExpirationTime());
+      if (this.connection != null
+          && this.connection.getLocation() != null
+          && !this.connection.getLocation().isEmpty()) {
+        datasetInfoBuilder.setLocation(this.connection.getLocation());
```
logachev: nit: we don't need all these checks. connection can't be null & if Location is null, so be it.. SDK will handle null Location the same way as no location.

E.g. we do it in executeJob()

java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITStatementTest.java
```diff
@@ -412,4 +416,42 @@ int getSizeOfResultSet(ResultSet resultSet) throws SQLException {
...
+    BigQuery bigQuery = BigQueryOptions.getDefaultInstance().getService();
+    try (Connection connection = DriverManager.getConnection(customConnectionUrl)) {
+      Statement statement = connection.createStatement();
+      String query = "SELECT 1 as val;";
+      try (ResultSet rs = statement.executeQuery(query)) {
+        assertTrue(rs.next());
+        assertEquals(1, rs.getInt("val"));
+      }
+
+      Dataset dataset = bigQuery.getDataset(DatasetId.of(tempDatasetName));
```
logachev, line 449: nit: should we validate that this dataset actually has data? (e.g. ensure that it is indeed the dataset that was used to write results to). Might require larger query

logachev (approved):

Issue #13296, opened 2026-05-28
Java SDK CIs Improvement Tracking Ticket


lqiu96:
Problems:
1. GAPIC Unit Tests Long Duration

On changes that impact all of google-cloud-java, all the GAPIC unit tests are run. These test jobs take well over 30+ min (https://github.com/googleapis/google-cloud-java/actions/runs/26462248492/job/77913201176?pr=13222). One latest log results timed this to be ~38 min.

These GAPIC unit tests are autogenerated within each module and do not vary too much. These look to test happy and sad paths for each RPC method, but generally are the same tests with different inputs and outputs. We have dedicated showcase tests that can test these RPCs functionality once and ensure that the common variations are tested properly.

Question(s):
- Can we stop autogenerating these unit tests and instead rely on showcase unit tests for complete coverage? Additionally, we can rely one showcase ITs to ensure that the functionality is able to work E2E

2. GraalVM Tests Duration

The GraalVM jobs (A,B,C) run through test cases (units and ITs) in each module twice (once to investigate all the possible code paths and a second time to run through the tests). This doubles the amount of time needed to ensure GraalVM compatibility. One of the latest runs for GraalVM ran for ~90 min (https://btx.cloud.google.com/invocations/a827908d-a2fb-49f2-8c39-01b51e4baad8/targets).

Is there a way that we can streamline the tests to run faster?
- Remove the GAPIC unit tests so that they do not need to run? Instead, we rely on showcase unit tests and integration tests
- Only run the required tests once?
- Use GraalVM fast build mode?

2.1 GraalVM Tests Flakiness

Additionally, the test suite is already quite flaky and this flakiness compounds into an even more expensive operation when there are long test times.

3. Enforcer and Fmt Jobs Duration

Similar to issues above, running these jobs is expensive and take a long time if they require building and compiling all the GAPIC modules.

PR #13518, opened 2026-06-18
chore: Update owlbot to avoid copying generated unit tests for pure GAPIC libraries

lqiu96:
Update all the owlbot hermetic configs to not copy over the GAPIC generated unit tests. We will instead rely on the showcase tests for coverage for functionality.

This also updates the template `owlbot.yaml.monorepo.j2` to prevent new client libraries from copying over the generated unit tests. All future library generations should only contain the source code.

Part of https://github.com/googleapis/google-cloud-java/issues/13296 to speed up the CIs (reducing the bulk test times).


blakeli0:
Is next generation going to remove the unit tests? Or we have to manually remove them first?

lqiu96:
> Is next generation going to remove the unit tests? Or we have to manually remove them first?

I believe the next generation should remove it. The remove regex should remove the specific autogenerated tests and the copy regex only copies over the files in `/src/main`. The existing handwritten test files in `/src/test` should remain.

blakeli0 (approved):

blakeli0:
cc: @zhumin8 The next generation should remove generated tests from pure GAPICs.

PR #13522, opened 2026-06-18
feat: port secure_context testing support to executor proxy

aseering:
Add support for our executor proxy, so that tools can drive tests/etc using secure parameters

### DIFF
--- a/java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
+++ b/java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
@@ -744,7 +744,16 @@ public class CloudClientExecutor extends CloudExecutor {
             Level.INFO, String.format("executeBatchDml [%d]: %s", i + 1, stmts.get(i).toString()));
       }
       List<Options.UpdateOption> allOptions = new ArrayList<>(java.util.Arrays.asList(options));
-      allOptions.add(Options.tag("batch-update-transaction-tag"));
+      boolean hasTag = false;
+      for (Options.UpdateOption opt : options) {
+        if (opt.getClass().getSimpleName().equals("TagOption")) {
+          hasTag = true;
+          break;
+        }
+      }
+      if (!hasTag) {
+        allOptions.add(Options.tag("batch-update-tag"));
+      }
       return getTransactionForWrite()
           .batchUpdate(stmts, allOptions.toArray(new Options.UpdateOption[0]));
     }
@@ -863,7 +872,10 @@ public class CloudClientExecutor extends CloudExecutor {
       throws IOException {
     // Create a cloud spanner client
     Credentials credentials;
-    if (WorkerProxy.serviceKeyFile.isEmpty()) {
+    if (WorkerProxy.serviceKeyFile == null
+        || WorkerProxy.serviceKeyFile.isEmpty()
+        || !new File(WorkerProxy.serviceKeyFile).exists()
+        || new File(WorkerProxy.serviceKeyFile).isDirectory()) {
       credentials = NoCredentials.getInstance();
     } else {
       credentials =
@@ -945,7 +957,10 @@ public class CloudClientExecutor extends CloudExecutor {
     }
     // Create a trace service client
     Credentials credentials;
-    if (WorkerProxy.serviceKeyFile.isEmpty()) {
+    if (WorkerProxy.serviceKeyFile == null
+        || WorkerProxy.serviceKeyFile.isEmpty()
+        || !new File(WorkerProxy.serviceKeyFile).exists()
+        || new File(WorkerProxy.serviceKeyFile).isDirectory()) {
       credentials = NoCredentials.getInstance();
     } else {
       credentials =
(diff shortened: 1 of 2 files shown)


java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
```diff
@@ -2198,13 +2198,51 @@ private Status executeGenerateDbPartitionsRead(
...
+      for (Map.Entry<String, com.google.spanner.executor.v1.Value> entry :
+          secureContextMap.entrySet()) {
+        com.google.protobuf.Value.Builder valueBuilder = com.google.protobuf.Value.newBuilder();
+        if (entry.getValue().getValueTypeCase() == com.google.spanner.executor.v1.Value.ValueTypeCase.IS_NULL
+            && entry.getValue().getIsNull()) {
+          valueBuilder.setNullValue(com.google.protobuf.NullValue.NULL_VALUE);
+        } else if (entry.getValue().getValueTypeCase() == com.google.spanner.executor.v1.Value.ValueTypeCase.STRING_VALUE) {
+          valueBuilder.setStringValue(entry.getValue().getStringValue());
+        } else {
+          throw new IllegalArgumentException("Unsupported secure parameter value type in GitHub proxy");
```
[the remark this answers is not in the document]
aseering: This text is already used elsewhere.  I think we want to keep the code in sync.  If a reviewer with broader context thinks this is worthwhile, please ping here and I'll make the change in both locations.

java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
```diff
@@ -737,14 +737,16 @@ public synchronized void bufferMutations(List<Mutation> mutations) throws Spanne
...
-    public synchronized long[] executeBatchDml(@Nonnull List<Statement> stmts)
-        throws SpannerException {
+    public synchronized long[] executeBatchDml(
+        @Nonnull List<Statement> stmts, Options.UpdateOption... options) throws SpannerException {
       for (int i = 0; i < stmts.size(); i++) {
         LOGGER.log(
             Level.INFO, String.format("executeBatchDml [%d]: %s", i + 1, stmts.get(i).toString()));
       }
+      List<Options.UpdateOption> allOptions = new ArrayList<>(java.util.Arrays.asList(options));
+      allOptions.add(Options.tag("batch-update-transaction-tag"));
```
olavloite: This will override any request tag that has been set in the new input argument `options` for this method. Is that intentional?

Also, the text of this tag implies that it is a transaction tag, but it is being set as a request tag. Tags that are passed in to a `batchUpdate(..)` call as an `UpdateOption` is translated to the `request_tag` field on the `RequestOptions` proto that is created for this `ExecuteBatchDml` request. So I think it would be better to remove the `transaction` part of this tag.

java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
```diff
@@ -2199,13 +2207,54 @@ private Status executeGenerateDbPartitionsRead(
...
+        }
+        clientContextBuilder.putSecureContext(entry.getKey(), valueBuilder.build());
+      }
+      return Options.clientContext(clientContextBuilder.build());
+    }
+    return null;
+  }
+
+  @SuppressWarnings("unchecked")
+  private <T> void addSecureContextOption(
```
olavloite: This method can be simplified (and made type safer) like this:

```java
  private void addSecureContextOption(
      Map<String, com.google.spanner.executor.v1.Value> secureContextMap,
      List<? super Options.ReadQueryUpdateTransactionOption> optionsList) {
    Options.ReadQueryUpdateTransactionOption secureContextOption =
        buildSecureContextOption(secureContextMap);
    if (secureContextOption != null) {
      optionsList.add(secureContextOption);
    }
  }
```

java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
```diff
@@ -2770,10 +2829,14 @@ private Status executeCloudDmlUpdate(
                     update.getParams(i).getType(), update.getParams(i).getValue()));
       }
       sender.initForQuery();
+      List<Options.QueryOption> queryOptions = new ArrayList<>();
+      queryOptions.add(Options.tag("dml-transaction-tag"));
```
olavloite, line 2838: nit: this is also translated to a request_tag, not a transaction_tag. (But in this case, I can see that it is an existing flaw in the code, so feel free to leave as-is)

java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
```diff
@@ -2804,7 +2867,16 @@ private Status executeCloudBatchDmlUpdates(
         }
         queries.add(stmt.build());
       }
-      long[] rowCounts = executionContext.executeBatchDml(queries);
+      Map<String, com.google.spanner.executor.v1.Value> secureContextMap =
+          new java.util.HashMap<>();
+      for (int i = 0; i < action.getUpdatesCount(); ++i) {
+        secureContextMap.putAll(action.getUpdates(i).getSecureContextMap());
+      }
```
olavloite: nit: the `putAll(..)` call will silently overwrite any duplicate entries in the different updates. Is that OK for this test framework?

Also, use an enhanced for loop:

```java
      for (QueryAction update : action.getUpdatesList()) {
        secureContextMap.putAll(update.getSecureContextMap());
      }
```

java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
aseering: Updated.

java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
aseering: Updated.

java-spanner/google-cloud-spanner-executor/src/main/java/com/google/cloud/executor/spanner/CloudClientExecutor.java
aseering: Done.

olavloite (approved):

PR #13542, opened 2026-06-23
feat(bigtable): route point read rows to shim

mutianf:


java-bigtable/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -302,16 +327,20 @@ public <RowT> UnaryCallable<Query, RowT> createReadRowCallable(RowAdapter<RowT>
     BigtableUnaryOperationCallable<Query, RowT> classic =
         new BigtableUnaryOperationCallable<>(
             readRowCallable,
```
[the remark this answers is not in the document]
mutianf: I don't think this review is right. readRows callable is converted to readRow callable with Transforming callable.

java-bigtable/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/MaybePointReadCallable.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Routes ReadRows calls whose query identifies a single row through a unary point-read callable,
+ * letting them benefit from the same session-shim diversion as {@code ReadRow}. Queries that cannot
+ * be reduced to a point read fall through to the classic {@code ReadRows} callable.
+ */
+@InternalApi
+public class MaybePointReadCallable<RowT> extends ServerStreamingCallable<Query, RowT> {
```
igorbernstein2, line 38: This looks mostly good. The only thought I have is that this will change the timeouts being used for the RPC. ie previously this would use the ReadRowsSettings and now it will use ReadRowSettings. 
I suspect that this might be ok. But worth calling out in the change log

igorbernstein2 (approved):
lgtm, but please think about the effects of the settings change

java-bigtable/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/MaybePointReadCallable.java
mutianf: We pass the RetryCode and RetrySettings in the callable.

PR #13543, opened 2026-06-23
fix(datastore): disable built-in OpenTelemetry SDK when metrics export is disabled

### LINKED ISSUE #13469: java-datastore: Error messages from OpenTelemetry
## Issue Details

In our logs, we see lots of messages

```
i.o.s.m.e.PeriodicMetricReader$Scheduled:275 - Exporter failed
```

## Environment

  - OS Type and Version: cos-121-18867-381-125
  - Java Version and JDK Vendor: eclipse-temurin:25.0.3_9-jdk

## Dependencies

    - Libraries-Bom ([com.google.cloud:libraries-bom](https://central.sonatype.com/artifact/com.google.cloud/libraries-bom)): 26.83.0

## Reproducer

Caused by creating a Datastore:

```
        var builder = DatastoreOptions.newBuilder()
            .setRetrySettings(Gcloud.getDefaultRetrySettings())
            .setProjectId(gcloudProjectId)
            .setDatabaseId(databaseId)
            .setCredentials(Gcloud.DEFAULT_CREDENTIALS.createCredentials());
        var result = builder.build().getService();
```

Even with            

```
   setOpenTelemetryOptions(
                DatastoreOpenTelemetryOptions.newBuilder()
                    .setTracingEnabled(false)
                    .setMetricsEnabled(false)
                    .setOpenTelemetry(ourOwnInstance)
                    .build());
```

the issue persists.

It seems the reason is the built-in `OpenTelemetry` instance which gets created even when the `DatastoreOpenTelemetryOptions` are turned off.

A rather clumsy workaround is setting an environment variable
```
DATASTORE_EMULATOR_HOST=https://datastore.googleapis.com 
```

There should be an option to turn off OpenTelemetry completely, or it should be opt-in and turned off by default.

## Logs and Stack Trace

The export failures are initially caused by

```
"com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: One or more TimeSeries could not be written: timeSeries[0] (metric.type="otel.sdk.metric_reader.collection.duration", metric.labels={"client_uid": "0002bd", "otel_component_type": "periodic_metric_reader", "otel_component_name": "periodic_metric_reader/0", "service": "datastore.googleapis.com"}, resource.type="global", resource.labels={"project_id": "nd-staging"}): The metric type must be a URL-formatted string with a domain and non-empty path."
```

lqiu96:
Fixes: https://github.com/googleapis/google-cloud-java/issues/13469

### DIFF
--- a/java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
+++ b/java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
@@ -120,6 +120,12 @@ public class BuiltInDatastoreMetricsProvider {
    */
   @Nullable
   public OpenTelemetry createOpenTelemetry(@Nonnull DatastoreOptions options) {
+    // If built-in metrics export is disabled, return no-op OpenTelemetry to avoid instantiating
+    // a real SdkMeterProvider. Otherwise, the SDK-internal PeriodicMetricReader will start
+    // and attempt to export diagnostic metrics, leading to log spam if the exporter fails.
+    if (!options.getOpenTelemetryOptions().isExportBuiltinMetricsToGoogleCloudMonitoring()) {
+      return OpenTelemetry.noop();
+    }
     Credentials credentials =
         Preconditions.checkNotNull(
             options.getCredentials(), "Credentials cannot be null for built in metrics");
--- a/java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
+++ b/java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
@@ -36,7 +36,6 @@ import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 
 /**
  * Provides a built-in {@link OpenTelemetry} instance for Datastore client-side metrics.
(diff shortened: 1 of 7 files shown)


java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -120,6 +120,12 @@ private static String hashClientUId(String uuid) {
    */
   @Nullable
   public OpenTelemetry createOpenTelemetry(@Nonnull DatastoreOptions options) {
+    // If built-in metrics export is disabled, return no-op OpenTelemetry to avoid instantiating
+    // a real SdkMeterProvider. Otherwise, the SDK-internal PeriodicMetricReader will start
+    // and attempt to export diagnostic metrics, leading to log spam if the exporter fails.
+    if (!options.getOpenTelemetryOptions().isExportBuiltinMetricsToGoogleCloudMonitoring()) {
+      return OpenTelemetry.noop();
```
jinseopkim0, line 125: Given this change, would it make sense to update https://github.com/googleapis/google-cloud-java/blob/9492aa2b01382d731abf465aa69444092068a814/java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/DatastoreMetricsRecorder.java#L106 to:

```
        if (builtInOtel != null && builtInOtel != OpenTelemetry.noop()) {
```

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
```diff
@@ -120,6 +120,12 @@ private static String hashClientUId(String uuid) {
    */
   @Nullable
```
jinseopkim0: Should we remove this `@Nullable` annotation?

lqiu96: Updated!

java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/telemetry/BuiltInDatastoreMetricsProvider.java
lqiu96: Updated!

jinseopkim0 (approved):

PR #13549, opened 2026-06-24
chore(bigquery-jdbc): fix flaky testSetTimeout test

logachev:
Seems like BQ can cache results of existing query, so timeout is not reliable. This query is guaranteed to run for 300 seconds, so it will ensure job is killed by timeout.

### DIFF
--- a/java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITStatementTest.java
+++ b/java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITStatementTest.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.bigquery.jdbc.it;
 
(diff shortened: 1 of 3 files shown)


java-bigquery-jdbc/src/test/java/com/google/cloud/bigquery/jdbc/it/ITStatementTest.java
```diff
@@ -388,14 +377,6 @@ public void testRangeSelectDataset() throws SQLException {
...
   }
 
-  int getSizeOfResultSet(ResultSet resultSet) throws SQLException {
-    int count = 0;
-    while (resultSet.next()) {
-      count++;
-    }
-    return count;
-  }
-
```
keshavdandeva, line 398: Tests are failing because of this removal

keshavdandeva (approved):

PR #13552, opened 2026-06-24
chore: update librarian to v0.22.0

JoeWang1127:


blakeli0 (approved):

JoeWang1127:
The missing line after header in grafeas was caused by https://github.com/googleapis/librarian/pull/6481

We will investigate the reason and reformat the file in the next version of librarian.

PR #13556, opened 2026-06-25
refactor(bigquery-jdbc): migrate metadata thread management to connection-scoped executor

keshavdandeva:
b/520400589

This PR optimizes thread management for JDBC metadata methods, implementing a deadlock-free, bipartite thread pool design and unifying resource lifecycle management.

### Key Changes

* **Bipartite Thread Pool Architecture (Deadlock Prevention)**:
  * Redirected all outer, orchestrating fetcher tasks (parent tasks) to the connection’s cached query executor (`connection.getExecutorService()`). Because the cached pool scales dynamically as needed, this structurally prevents pool-induced deadlocks (thread starvation).
  * Kept all inner, highly concurrent parallel network API calls (child tasks) on the connection's fixed-size metadata executor (`connection.getMetadataExecutor()`). This bounds parallel network traffic to prevent overwhelming the BigQuery backend.
  * *Result*: The driver is fully protected against pool-induced deadlocks, even under high concurrency or if `metadataFetchThreadCount` is configured to `1`.

* **Connection-Scoped Shared Executor**:
  * Migrated all asynchronous metadata operations (`getSchemas`, `getTables`, `getColumns`, `getProcedures`, `getProcedureColumns`, `getFunctions`, and `getFunctionColumns`) from spawning raw threads or short-lived local pools to connection-managed executors.
  * Configured the shared metadata pool to be lazily instantiated on demand and gracefully shut down with the connection lifecycle. Idle threads are automatically reclaimed after 60 seconds.

* **Lower CPU Overhead (Sequential Row-Building)**:
  * Completely removed secondary CPU-bound thread pools (such as `routineProcessorExecutor`, `processArgsExecutor`, `tableProcessorExecutor`, and `processParamsExecutor`) previously used for row-building.
  * Refactored row-building and parameter-processing tasks to execute sequentially on the background fetcher threads, removing context-switching overhead and preventing nested-pool deadlock risks.

* **Teardown of Legacy Bridges**:
  * Deleted the obsolete `wrapThread(Thread)` compatibility adapter in `BigQueryDatabaseMetaData.java`.
  * Removed obsolete `testWrapThread_*` cases from `BigQueryDatabaseMetaDataTest.java`. Metadata queries now track and cancel background tasks natively via standard `Future<?>` handles.

* **Test Suite Refactoring**:
  * Updated `BigQueryDatabaseMetaDataTest` setup to inject real, connection-scoped cached executor services, ensuring that background metadata tasks execute deterministically during testing.
  * Added `@AfterEach` teardown logic to prevent test-suite thread leaks.
  * Refactored argument-processing tests to verify row-building synchronously without mock executors.

### DIFF
--- a/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
+++ b/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
@@ -217,7 +217,7 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
   Boolean reqGoogleDriveScope;
   private final Properties clientInfo = new Properties();
   private boolean isReadOnlyTokenUsed = false;
-  private final ExecutorService metadataExecutor;
+  private ExecutorService metadataExecutor;
   private final ExecutorService queryExecutor;
 
   BigQueryConnection(String url) throws IOException {
@@ -362,11 +362,6 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
 
       this.headerProvider = createHeaderProvider();
       this.bigQuery = getBigQueryConnection();
-      // Fixed thread pool queues tasks to limit concurrent metadata calls and prevent API
-      // throttling.
-      this.metadataExecutor =
-          BigQueryJdbcMdc.newFixedThreadPool(
-              String.format("BQ-Metadata-%s", connectionId), metadataFetchThreadCount);
       // Cached pool executes queries immediately without queueing and reclaims all idle threads
       // when inactive, minimizing resources.
       this.queryExecutor =
@@ -1067,7 +1062,12 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
     return this.queryExecutor;
   }
 
-  ExecutorService getMetadataExecutor() {
+  synchronized ExecutorService getMetadataExecutor() {
+    if (this.metadataExecutor == null) {
+      this.metadataExecutor =
+          BigQueryJdbcMdc.newFixedThreadPool(
+              String.format("BQ-Metadata-%s", connectionId), metadataFetchThreadCount);
+    }
     return this.metadataExecutor;
   }
 
--- a/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
+++ b/java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
@@ -949,7 +949,7 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
    * @see Connection#close()
    */
   @Override
-  public void close() throws SQLException {
+  public synchronized void close() throws SQLException {
     if (isClosed()) {
       return;
     }
@@ -1063,11 +1063,13 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
   }
 
   synchronized ExecutorService getMetadataExecutor() {
-    if (this.metadataExecutor == null) {
-      this.metadataExecutor =
-          BigQueryJdbcMdc.newFixedThreadPool(
-              String.format("BQ-Metadata-%s", connectionId), metadataFetchThreadCount);
+    if (this.metadataExecutor != null) {
+      return this.metadataExecutor;
     }
+    checkClosed();
+    this.metadataExecutor =
+        BigQueryJdbcMdc.newFixedThreadPool(
+            String.format("BQ-Metadata-%s", connectionId), metadataFetchThreadCount);
     return this.metadataExecutor;
   }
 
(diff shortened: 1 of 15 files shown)


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -217,7 +217,7 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
   Boolean reqGoogleDriveScope;
   private final Properties clientInfo = new Properties();
   private boolean isReadOnlyTokenUsed = false;
-  private final ExecutorService metadataExecutor;
+  private ExecutorService metadataExecutor;
```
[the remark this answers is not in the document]
keshavdandeva: Already covered. There are explicit null checks in `closeImpl()` before interacting with `metadataExecutor` at both stages of teardown:

- Initial shutdown check: `if (this.metadataExecutor != null) { this.metadataExecutor.shutdown(); }`
- Termination await check: `if (this.metadataExecutor != null) { ... }`

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -934,21 +906,17 @@ public ResultSet getProcedures(
...
-            shutdownExecutor(routineProcessorExecutor);
             LOG.info("Procedure fetcher thread finished.");
           }
         };
 
-    Thread fetcherThread = new Thread(procedureFetcher, "getProcedures-fetcher-" + catalog);
+    Future<?> fetcherFuture = connection.getMetadataExecutor().submit(procedureFetcher);
     BigQueryJsonResultSet resultSet =
-        BigQueryJsonResultSet.of(resultSchema, -1, queue, null, wrapThread(fetcherThread));
+        BigQueryJsonResultSet.of(resultSchema, -1, queue, null, new Future<?>[] {fetcherFuture});
```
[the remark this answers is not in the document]
keshavdandeva: Done

Neenu1995 (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -362,11 +362,6 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
 
       this.headerProvider = createHeaderProvider();
       this.bigQuery = getBigQueryConnection();
-      // Fixed thread pool queues tasks to limit concurrent metadata calls and prevent API
```
logachev, line 365: Why changing this to lazy init? Init is cheap & it won't start threads unless required

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -1187,30 +1115,23 @@ public ResultSet getProcedureColumns(
...
             LOG.info("Procedure column fetcher thread finished for catalog: " + catalogParam);
           }
         };
 
-    Thread fetcherThread =
-        new Thread(procedureColumnFetcher, "getProcedureColumns-fetcher-" + catalog);
+    Future<?> fetcherFuture = connection.getExecutorService().submit(procedureColumnFetcher);
     BigQueryJsonResultSet resultSet =
-        BigQueryJsonResultSet.of(resultSchema, -1, queue, null, wrapThread(fetcherThread));
+        BigQueryJsonResultSet.of(resultSchema, -1, queue, null, new Future<?>[] {fetcherFuture});
```
logachev: nit: add override accepting single Future & constructing list under the cover?

logachev (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
keshavdandeva: I was thinking to optimize it for the cases when connections only execute SQL statements (and never run metadata queries) so it takes absolutely zero memory or object allocations for the metadata pool.

But yeah you are right, I missed `Init is cheap & it won't start threads unless required`. Reverted.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
keshavdandeva: Done

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryConnection.java
```diff
@@ -362,15 +362,13 @@ public class BigQueryConnection extends BigQueryNoOpsConnection {
...
-      this.metadataExecutor =
-          BigQueryJdbcMdc.newFixedThreadPool(
-              String.format("BQ-Metadata-%s", connectionId), metadataFetchThreadCount);
       // Cached pool executes queries immediately without queueing and reclaims all idle threads
       // when inactive, minimizing resources.
       this.queryExecutor =
           BigQueryJdbcMdc.newCachedThreadPool(String.format("BQ-Query-%s", connectionId));
+      this.metadataExecutor =
+          BigQueryJdbcMdc.newFixedThreadPool(
+              String.format("BQ-Metadata-%s", connectionId), this.metadataFetchThreadCount);
```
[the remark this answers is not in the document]
keshavdandeva: No

PR #13557, opened 2026-06-25
chore: expose the user set executor on TransportProvider

igorbernstein2:
This allows service clients to reuse that executor for other transport related things in addition to gax


mutianf (approved):

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -70,6 +71,17 @@ public interface TransportChannelProvider {
...
   boolean needsExecutor();
 
+  /**
+   * @return the user provided executor. This can be null if the user didn't override the executor
+   *     and the TransportProvider is using its internal executor.
+   * @throws UnsupportedOperationException if the TransportProvider doesn't support this method.
+   */
+  @Nullable
+  default Executor getExecutor() {
+    throw new UnsupportedOperationException(
```
blakeli0: I'm leaning towards returning `null` instead of throwing exceptions.  Because `null` is a valid return value for this method, which means that the user didn't provide an external executor. Also we don't have to override it for FixedTransportChannelProvider anymore.

igorbernstein2: I was a bit on the fence about this. I originally was thinking to return null. But it seems like spanner has a an impl of TransportChannelProvider and i wasnt sure about their executor semantics, so I thought it would be safer for the new method to throw unsupported. But if you feel strongly. I can return null

blakeli0: Yes, let's return null then.

igorbernstein2: updated

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/FixedTransportChannelProvider.java
```diff
@@ -57,6 +58,12 @@ public boolean needsExecutor() {
     return false;
   }
 
+  @Nullable
+  @Override
+  public Executor getExecutor() {
```
blakeli0: This override is not needed anymore.

igorbernstein2: fixed

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/FixedTransportChannelProvider.java
```diff
@@ -36,6 +36,7 @@
 import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
+import javax.annotation.Nullable;
```
blakeli0: This import is not needed either.

igorbernstein2: oops, removed

sdk-platform-java/gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -70,6 +71,16 @@ public interface TransportChannelProvider {
   @Deprecated
   boolean needsExecutor();
 
+  /**
+   * @return the user provided executor. This can be null if the user didn't override the executor
+   *     and the TransportProvider is using its internal executor.
+   * @throws UnsupportedOperationException if the TransportProvider doesn't support this method.
```
blakeli0: I might missed this earlier, Java doc needs to be updated as well.

blakeli0 (approved):

PR #13558, opened 2026-06-25
feat(gapic-generator): Add Nullable annotation to generated classes

nnicolee:
This PR updates the gapic-generator-java to add the JSpecify @Nullable annotation to all generated class declarations. This is the second phase in onboarding the generated client libraries to compile-time safety validation, see design doc for more details: [go/sdk:java-jspecify-null-annotations-gapic](http://goto.google.com/sdk:java-jspecify-null-annotations-gapic)

Classes Annotated:

- Client Classes: AbstractServiceClientClassComposer
- Stub Classes: AbstractServiceStubClassComposer and ServiceStubClassComposer
- Resource Names: ResourceNameHelperClassComposer

Implementation Changes:

- Built the parameter variable expression in equals(Object o) as a reference initially, and attached the @Nullable annotation during its method argument declaration (with isDecl = true). This prevents IllegalStateException crashes in the generator's strict AST validation engine.
- Leveraged a VaporReference type name-hacking trick to render generic type-use annotations (such as List<@Nullable ResourceName>) without needing disruptive changes to the core AST reference rendering engine.
- Modified two owlbot.py scripts to reflect the signature changes: ‎java-bigquerydatatransfer/owlbot.py & java-bigquerydatatransfer/owlbot.py

Verification/Testing:
- Please refer to this doc: https://docs.google.com/document/d/1h126pbTGqSwtJyl35E2_vQjM9gNtmWETYCM7PObcvmQ/edit?tab=t.0#heading=h.s4q06jxocx8b


java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
```diff
@@ -398,6 +399,7 @@ protected ComplianceClient(ComplianceStub stub) {
     this.stub = stub;
   }
 
+  @Nullable
```
lqiu96: I think this needs to be next to the type right?

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
```diff
@@ -360,7 +361,7 @@
 @BetaApi
 @Generated("by gapic-generator-java")
 public class ComplianceClient implements BackgroundResource {
-  private final ComplianceSettings settings;
+  @Nullable private final ComplianceSettings settings;
```
lqiu96: I think this also needs to be next to the type

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
```diff
@@ -1580,7 +1582,8 @@ public static class ListLocationsFixedSizeCollection
           ListLocationsPage,
           ListLocationsFixedSizeCollection> {
 
-    private ListLocationsFixedSizeCollection(List<ListLocationsPage> pages, int collectionSize) {
+    private ListLocationsFixedSizeCollection(
+        @Nullable List<ListLocationsPage> pages, int collectionSize) {
```
lqiu96, line 1585: Can you double check this one? Do we need to add an annotation here for paging?

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/MessagingClient.java
```diff
@@ -630,8 +632,7 @@ public final UnaryCallable<CreateRoomRequest, Room> createRoomCallable() {
    * @throws com.google.api.gax.rpc.ApiException if the remote call fails
    */
   public final Room getRoom(RoomName name) {
-    GetRoomRequest request =
-        GetRoomRequest.newBuilder().setName(name == null ? null : name.toString()).build();
+    GetRoomRequest request = GetRoomRequest.newBuilder().setName(name.toString()).build();
```
lqiu96: qq, I don't think we should change this. Perhaps better to mark the param as nullable?

java-showcase/proto-gapic-showcase-v1beta1/src/main/java/com/google/showcase/v1beta1/BlurbName.java
```diff
@@ -188,6 +189,7 @@ public static String formatRoomLegacyRoomBlurbName(String room, String legacyRoo
         .toString();
   }
 
+  @Nullable
```
lqiu96: same here, should be type-use annotation here.

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/MessagingClient.java
diegomarquezp: +1 the [new logic](https://github.com/googleapis/google-cloud-java/blob/783d6b68ddc26b7e02d52764aa5fa3f18028dcae/sdk-platform-java/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientClassComposer.java#L1763-L1788) may affect this, which is a good way to catch additional cases where `@Nullable` is needed.

diegomarquezp: Discussed with @nnicolee, looks like `setName` is a proto class, which we don't have control over, and is being marked by NullAway with:

```
[ERROR] /usr/local/google/home/nicolejlee/google-cloud-java/java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/MessagingClient.java:[1822,64] error: [NullAway] passing @Nullable parameter 'parent == null ? null : parent.toString()' where @NonNull is required
```

If it was possible to configure ErrorProne to ignore proto classes in these cases that would be great. That could be a good first step.
On the other hand, looks like Proto classes actually can take nulls (since the code's been like this), which can be worked around by adding a warning suppression as Nicole suggested.

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
nnicolee: Updated! Yes!

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
nnicolee: Updated, yes!

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
nnicolee: Hi! Yes, we need to add an annotation here because two places call this function (createCollection and createEmptyCollection (on line 1590-1) which calls the function with null in the first parameter (which is pages). Verified this annotation with Nullaway!

java-showcase/proto-gapic-showcase-v1beta1/src/main/java/com/google/showcase/v1beta1/BlurbName.java
diegomarquezp: Hey, I think the body of the function can return null. IIUC `@Nullable` at the top means the result of the function can be null?
If true, maybe it makes sense to keep both this line and also annotate the parameter with `@Nullable`?

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
lqiu96: I think this one is still next to private and not the type

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
lqiu96: Can you double check this one again

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
nnicolee: pushed the updated tests!

sdk-platform-java/gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
```diff
@@ -479,6 +479,9 @@ private void variableExpressions(List<VariableExpr> expressions) {
   }
 
   private void handleReference(Reference reference) {
+    if (reference.isNullable()) {
+      addImport("org.jspecify.annotations.Nullable");
+    }
```
lqiu96, line 484: qq, should this be at the top of this method? Would it make sense to add this at the bottom after all the checks for early returns?

sdk-platform-java/gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -222,6 +222,10 @@ public void visit(ConcreteReference reference) {
       buffer.append(DOT);
     }
 
+    if (reference.isNullable()) {
+      buffer.append("@Nullable ");
```
lqiu96: nit: Can you make `@Nullable ` a constant?

sdk-platform-java/gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
nnicolee: placing it at the top because if we're using a type that doesn't require an import (like something from java.lang or the current package), we still need to ensure that org.jspecify.annotations.Nullable is imported if that type is annotated. moving it to bottom would cause the method to return early before the @Nullable import is added!

sdk-platform-java/gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
nnicolee: yes updated!

sdk-platform-java/gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
lqiu96: > we still need to ensure that org.jspecify.annotations.Nullable is imported if that type is annotated.

Can you explain this part a bit more? What do you mean by we need to ensure Nullable is annotated if the type is annotated?

If I have something like ComplianceStub as the reference, we still need to add it?

nnicolee: yes! we need to ensure Nullable is imported if the type is annotated by adding it to the top because if the type being annotated does not require an import (because it's in Java.lang or the current package), handleReference exits early via the package check.

for example: @Nullable String flow:
1. the generator visits the String type reference
2. it calls handleReference(String)
3. since String is from Java.lang, it matches the package check and hits the early return
4. result: the generated file has @Nullable String, but no import statements which would cause compilation to fail

if we have something like ComplianceStub as the reference, we still need to import if a type in the same package is annotated!

for example: if we're generating Compliance client.java and it references ComplianceStub with @Nullable (ex. @Nullable ComplianceStub stub), ComplianceStub itself does not need to be imported since it shares the current package, but the file still need to import Nullable for the annotation to compile

lqiu96 (approved):

java-showcase/proto-gapic-showcase-v1beta1/src/main/java/com/google/showcase/v1beta1/BlurbName.java
nnicolee: Thank you!  I refactored this to annotate the parameter with @Nullable, since upon double-checking, I think only the parameter can be nullable

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceClient.java
nnicolee: Yes updated! @Nullable is next to type

java-showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/MessagingClient.java
nnicolee: Yes, currently we're ignoring this error

PR #13560, opened 2026-06-26
refactor(bigquery-jdbc): optimize thread management and unify discovery logic in `DatabaseMetaData` methods

keshavdandeva:
b/520407325
b/520406763

This PR refactors and optimizes the database metadata retrieval methods in the BigQuery JDBC driver. It resolves thread management inefficiencies, eliminates duplicate API blocks, ensures consistent project discovery support, and implements strict, fail-fast error propagation across all asynchronous metadata methods.

### 1. Catalog-Based Routing in `getSchemas`
*   **Synchronous Single-Catalog Path**: If a specific catalog (project) is requested, `getSchemas` now executes completely synchronously on the calling thread, eliminating background executor and queue overhead.
*   **Parallel Multi-Catalog Path**: If no catalog is specified, the query executor runs parallel scans across all accessible projects (primary, additional, and discovered) in the background.

### 2. Unified Dataset Discovery & Deduplication
*   Introduced a single, shared helper method `fetchMatchingDatasets` to serve as the sole entry point for listing and filtering datasets.
*   Deduplicated dataset-listing logic across **7 metadata methods** (`getSchemas`, `getTables`, `getColumns`, `getProcedures`, `getProcedureColumns`, `getFunctions`, and `getFunctionColumns`), ensuring consistent support for project discovery and SQL wildcard matching across all of them.

### 3. Strict Fail-Fast Error & Interruption Handling
*   **Preventing Silent Truncation**: Background metadata fetchers used to silently swallow `InterruptedException`s and `ExecutionException`s, returning partial or empty results to the user (which severely violates the JDBC contract during query timeouts or connection drops).
*   **Consistent Propagation**: All 7 background methods now strictly push an `SQLException` to the client's queue whenever an interruption or runtime error occurs, ensuring the user gets an explicit failure when calling `rs.next()`.
*   **Boilerplate Elimination**: Extracted the complex error-handling and queue-finalization boilerplate into centralized helper methods (`handleFetcherException` and `finalizeFetcher`). This removed over 50 lines of duplicate code, deleted 14+ redundant `isInterrupted()` checks, and established a clean, unified standard for all asynchronous metadata threads.


java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -3505,20 +3458,37 @@ public ResultSet getSchemas(String catalog, String schemaPattern) throws SQLExce
...
+        Comparator<FieldValueList> comparator = defineGetSchemasComparator(resultSchemaFields);
+        sortResults(collectedResults, comparator, "getSchemas", LOG);
+        populateQueue(collectedResults, queue, resultSchemaFields);
+      } catch (Throwable t) {
+        LOG.severe("Unexpected error in synchronous getSchemas: " + t.getMessage());
+        writeErrorToQueue(queue, t);
+      } finally {
+        signalEndOfData(queue, resultSchemaFields);
+      }
+      return BigQueryJsonResultSet.of(resultSchema, -1, queue, null);
```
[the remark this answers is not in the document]
keshavdandeva: There is no compilation error here because a 4-parameter overload of BigQueryJsonResultSet.of()  exists

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -4950,38 +4899,45 @@ private void signalEndOfData(
...
-        LOG.warning("Executor did not terminate gracefully after 10s, forcing shutdownNow().");
-        List<Runnable> droppedTasks = executor.shutdownNow();
-        LOG.warning(
-            "Executor shutdownNow() initiated. Dropped tasks count: " + droppedTasks.size());
-        if (!executor.awaitTermination(10, TimeUnit.SECONDS)) {
-          LOG.severe("Executor did not terminate even after shutdownNow().");
+  private List<Dataset> fetchMatchingDatasets(
+      String catalog, String schemaPattern, Pattern schemaRegex) throws SQLException {
+    List<String> projects =
+        (catalog != null) ? Collections.singletonList(catalog) : getAccessibleCatalogNames();
```
logachev, line 4843: `catalog=""` is a special case, I think in BQ it should return an empty list since in BQ all datasets belong to some project.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -4950,38 +4899,45 @@ private void signalEndOfData(
...
+                schemaRegex,
+                LOG);
+        if (datasets != null) {
+          allDatasets.addAll(datasets);
         }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        break;
+      } catch (Exception e) {
+        if (catalog != null) {
```
logachev: Why special case for this one? I don't see anything about it in spec, so 2 scenarios in my mind:

- Service account has some access to projectA, but no BQ access to list datasets.. 
- call getSchemas(null, null) -> returns empty result, no exception
- call getSchemas(projectA, null) -> throws exception.. 

So I think we should unify the behavior. 
Spec is saying only "[SQLException](https://docs.oracle.com/javase/8/docs/api/java/sql/SQLException.html) - if a database access error occurs").. So I think we can throw exception unless we were able to successfully fetch all data.

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
keshavdandeva: Yeah, this is handled in all metadata methods already. We check `if ((catalog != null && catalog.isEmpty())` and return empty resultset. So, any catalog with `""` would not reach here

keshavdandeva: On a side note, while confirming this, I found `determineEffectiveCatalogAndSchema` method that is used to help the connection property `isFilterTablesOnDefaultDataset` is not exactly optimized/working as it should. Will refactor it in separate PR

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
keshavdandeva: Yeah, so the reason I did it like this was for the case when projectDiscovery is enabled. Basically:

*   **Targeted queries** (`catalog != null`): Fail-fast and throw a `SQLException` on failure.
*   **Discovery queries** (`catalog == null`): Resilient/best-effort. We skip individual project failures (e.g. permission/billing issues on sandbox/test projects) with a warning log, and return results from other accessible projects. This prevents a single inactive project from breaking the entire metadata call when using user credentials (ADC).

To unify this, which approach would you prefer we take?

1.  **Fully Unified (Fail-Fast)**: Always throw `SQLException` if any project fails during the scan, regardless of whether it was a targeted or discovery query. (as you suggested above)
2.  **Conditional Hybrid (Project Discovery Only)**: Only use the best-effort model (log warning and skip) if `EnableProjectDiscovery=true` If discovery is disabled, any failure on the primary catalog or manually configured `AdditionalProjects` will throw a `SQLException`

keshavdandeva: Implemented the unified approach to throw `SQLException` if any project fails during the scan + removed swalloqing of exceptions for other methods as well

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -5026,4 +4904,14 @@ static String readSqlFromFile(String filename) {
...
+
+  private void writeErrorToQueue(BlockingQueue<BigQueryFieldValueListWrapper> queue, Throwable t) {
+    Exception ex = (t instanceof Exception) ? (Exception) t : new Exception(t);
+    try {
+      queue.put(BigQueryFieldValueListWrapper.ofError(ex));
+    } catch (InterruptedException ie) {
+      LOG.warning("Failed to put exception to queue due to interruption.");
+      Thread.currentThread().interrupt();
+    }
+  }
```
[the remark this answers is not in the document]
keshavdandeva: Fixed

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -3504,21 +3403,34 @@ public ResultSet getSchemas(String catalog, String schemaPattern) throws SQLExce
     final FieldList resultSchemaFields = resultSchema.getFields();
 
     final BlockingQueue<BigQueryFieldValueListWrapper> queue =
```
logachev: very nit: can create that queue within relevant `if`/after if`

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
```diff
@@ -4937,51 +4792,88 @@ private void populateQueue(
...
-        LOG.warning(
-            "Executor shutdownNow() initiated. Dropped tasks count: " + droppedTasks.size());
-        if (!executor.awaitTermination(10, TimeUnit.SECONDS)) {
-          LOG.severe("Executor did not terminate even after shutdownNow().");
+  private List<Dataset> fetchMatchingDatasets(
+      String catalog, String schemaPattern, Pattern schemaRegex) throws SQLException {
+    List<String> projects =
+        (catalog != null) ? Collections.singletonList(catalog) : getAccessibleCatalogNames();
+    List<Dataset> allDatasets = new ArrayList<>();
+    for (String project : projects) {
```
logachev, line 4845: Do you plan to parallelize that later?

logachev (approved):

java-bigquery-jdbc/src/main/java/com/google/cloud/bigquery/jdbc/BigQueryDatabaseMetaData.java
keshavdandeva: I have some more optimization/refactoring things, will address this and discuss separately

PR #13567, opened 2026-06-29
fix: fallback on VPC

mutianf:


sushanb:
I think in case of  PERMISSION_DENIED exception is caused by a VPC Service Controls,
 if i understand correctly, it will never send the request to AFE so isALTS should be false, so it should fallback to regular transport.

java-bigtable/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/dp/ClassicDirectAccessChecker.java
```diff
@@ -74,6 +74,15 @@ public boolean check(Channel channel) {
     }
   }
 
+  /** Checks if the exception is due to a VPC Service Controls policy violation. */
+  private boolean isAllowed(StatusRuntimeException e) {
```
sushanb: add isAllowedFromVPCServiceControls

java-bigtable/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/dp/ClassicDirectAccessChecker.java
```diff
@@ -74,6 +74,15 @@ public boolean check(Channel channel) {
...
+    String description = e.getStatus().getDescription();
+    String message = e.getMessage();
+    return (description != null
+            && description.contains("Request is prohibited by organization's policy"))
```
sushanb: can you see if this is the authorative string?

I see "request is prohibited by " as well.

sushanb (approved):

igorbernstein2 (approved):

